PR #257, opened 2021-05-12
Signature map expiry: 1 min

nomeata:
This time span needs to be long enough to in the typical case cover
these actions:
An update call is executed.
The polling client notices (via state read)
The client performs a query calls.

Probably 10s would be enough in most cases and 1min is still plenty.


roman-kashitsyn (approved):
LGTM, thanks!

PR #259, opened 2021-05-12
[FOLLOW-134] fix assigned user range on upgrades

roman-kashitsyn:
The default canister user range used to be 8M, which is way too large
for a single canister.

This change implements post-upgrade logic that sets the range to a
more sensible value.


src/internet_identity/src/main.rs
```diff
@@ -31,6 +31,9 @@ const DEFAULT_SIGNATURE_EXPIRATION_PERIOD_NS: u64 = secs_to_nanos(600);
 // 5 mins
 const POW_NONCE_LIFETIME: u64 = secs_to_nanos(300);
 
+/// The maximum number of users we can store in a single canister.
+const MAX_USERS_PER_CANISTER: u64 = 2_000_000_000;
```
nomeata, line 35: I wonder if we should calculate this value (based on record size and memory size).

Or, maybe better, in `storage.set_user_number_range` and `Storage::new` trap if we’d need more storage than is available in stable memory (i.e. if the last byte of the the last record is out of bounds).

Or, in any case, something like
```suggestion
// This number is derived from the stable memory size (4GB) and the user record size (2kb)
// leaving a few pages in the end as spares
const MAX_USERS_PER_CANISTER: u64 = 2_000_000_000;
```

roman-kashitsyn: Very good point. I changed the code to compute the max number of users based on the canister capacity and realized that the upper bound I specified is 1000x higher than it should be (it should be 2M, not 2B, *facepalm*).
The computed upper bound is `10_000 + 2_097_151`.

src/internet_identity/src/storage.rs
```diff
@@ -12,6 +12,11 @@ const HEADER_SIZE: u32 = 512;
 const DEFAULT_ENTRY_SIZE: u16 = 2048;
 const EMPTY_SALT: [u8; 32] = [0; 32];
 const WASM_PAGE_SIZE: u32 = 65536;
+const STABLE_MEMORY_SIZE: u64 = 1 << 32;
+
+/// The maximum number of users this canister can store.
+pub const DEFAULT_RANGE_SIZE: u64 =
+    (STABLE_MEMORY_SIZE - HEADER_SIZE as u64) / DEFAULT_ENTRY_SIZE as u64;
```
nomeata, line 19: Shall we reserve some space in the stable memory for posteriorty? 10%? Maybe we need to add features that need additional space later.

roman-kashitsyn: I like that idea, done.

nomeata (approved):

PR #264, opened 2021-05-12
[IC-232] use a hash set for nonce cache

roman-kashitsyn:
This change improves the algorithmic complexity of the NonceCache by
introducing a HashSet storing all known (timestamp, nonce) pairs.

This way we don't have to do a linear search over the whole queue in
contains().


nomeata (approved):
Oh, that looked easy :-). Thanks!

PR #265, opened 2021-05-14
Client auth flow: pass maxTimeToLive (FOLLOW-136)

nomeata:
We were not passing `maxTimeToLive` through to the backend. This fixes
this.

It also refactors the selenium test suite mildly and adds tests that
exercise setting the default maxTimeToLive, 1min, 1day and 1 month.

This needs a released update in `auth-client` for the tests to work, so
we’ll have to wait for https://github.com/dfinity/agent-js/pull/417


nomeata:
Harsh review of my TypeScript code is welcome :-)

What is the difference between `bigint` and `BigInt` in typescript? Should we indeed be using both?

demos/sample-javascript/src/main.js
```diff
@@ -8,27 +9,54 @@ const hostUrlEl = document.getElementById("hostUrl");
...
   authClient = await AuthClient.create();
-  principalEl.innerText = await authClient.getIdentity().getPrincipal();
 
-  // Redirect to the identity provider
+  const updateView = () => {
+    const identity = authClient.getIdentity();
+
+    principalEl.innerText = identity.getPrincipal();
+    if (identity instanceof DelegationIdentity) {
+      delegationEl.innerText = JSON.stringify(identity.getDelegation().toJSON());
```
kritzcreek, line 27:
```suggestion
      delegationEl.innerText = JSON.stringify(identity.getDelegation().toJSON(), undefined, 2);
```

If you'd like it to pretty print the JSON with some spacing and newlines

demos/sample-javascript/src/index.html
```diff
@@ -13,10 +13,24 @@ <h2>Sign In</h2>
...
+	  placeholder="(use default)"
+        />
+      </div>
       <button id="signinBtn">Sign In</button>
       <button id="signoutBtn">Sign Out</button>
-      <h2>Principal:</h2>
+      <h3>Principal:</h3>
       <div id="principal"></div>
+      <h3>Delegation:</h3>
+      <div id="delegation"></div>
```
kritzcreek, line 31: I'd maybe make this a `pre` (so it's monospaced)

demos/sample-javascript/src/main.js
```diff
@@ -8,27 +9,54 @@ const hostUrlEl = document.getElementById("hostUrl");
...
+
+      const nextExpiration =
+        identity.getDelegation().delegations
+         .map(d => d.delegation.expiration)
+         .reduce((current, next) => next < current ? next : current);
```
kritzcreek, line 32:
```suggestion
        Math.min(...identity.getDelegation().delegations.map(d => d.delegation.expiration));
```

Might be a little easier for the non FP folks

src/frontend/src/selenium.test.ts
```diff
@@ -387,6 +416,30 @@ async function run_in_browser_common(
...
+  const elem = await driver.findElement(By.id(id));
+  elem.sendKeys(Key.CONTROL + "a");
+  elem.sendKeys(Key.DELETE);
+  elem.sendKeys(text);
+}
+
+async function switchToPopup(driver: ThenableWebDriver) {
+  let handles = await driver.getAllWindowHandles();
+  expect(handles.length).toBe(2);
+  await driver.switchTo().window(handles[1]);
```
kritzcreek, line 429: Is this order always deterministic? Could we maybe do something like finding the popup via its window title?

kritzcreek (changes requested):
Looks great! `BigInt` is the constructor, while `bigint` is the type... You can also use `BigInt` as the type, but that causes problems with inference so just stick to `bigint` for types. Hans explained that to me at some point, but I don't remember the details.

demos/sample-javascript/src/main.js
nomeata: I actually copied that from the `dfinity_wallet` code :-)

src/frontend/src/selenium.test.ts
nomeata: So far it works 🤷🏻

nomeata:
I found an existing use of `BigInt` in a type, and replaced that with `bigint`, and thus could remove one conversion. hope that’s fine!

src/frontend/src/selenium.test.ts
kritzcreek: Allright, good enough for me. If it ever starts failing the test will fail local enough, that we can figure it out.

kritzcreek (approved):

PR #271, opened 2021-05-17
Bump the dfinity JS libraries, fetch root key on dev instances

nomeata:
This bumps to the latest JS libraries, unblocking #265. The new versions will use the official root IC key by default, so we have to call `fetchRootKey()` in development mode. I opted for a dynamic detection of development environments (so that we can run identical `.wasm` on testnets and mainnet), with a small, hopefully reasonably safe, conditions for when we fetch the root key.


src/frontend/src/utils/iiConnection.ts
```diff
@@ -172,6 +164,26 @@ export class IIConnection {
...
+    delegationIdentity: DelegationIdentity
+  ): Promise<ActorSubclass<_SERVICE>> {
+    const agent = new HttpAgent({ identity: delegationIdentity });
+    // try to recognize development environments
+    // this is conservative, and defaults to _not_ fetching the root key
+    if (
+      window.location.origin.startsWith("http://localhost:") ||
+      window.location.origin.endsWith(".dfinity.network")
+    ) {
+      await agent.fetchRootKey();
```
ielashi, line 178: Question: what's wrong with modifying `fetchRootKey` to first check the status endpoint of the replica it's talking to? If the root key is available (in case of local dev), then use that, otherwise, use the IC mainnet root key.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -172,6 +164,26 @@ export class IIConnection {
...
   // Create an actor representing the backend
+
+  static async createActor(
+    delegationIdentity: DelegationIdentity
+  ): Promise<ActorSubclass<_SERVICE>> {
+    const agent = new HttpAgent({ identity: delegationIdentity });
+    // try to recognize development environments
+    // this is conservative, and defaults to _not_ fetching the root key
+    if (
+      window.location.origin.startsWith("http://localhost:") ||
```
ielashi, line 175: Maybe add 127.0.0.1 as well?

src/frontend/src/utils/iiConnection.ts
nomeata: Two reasons
 * we still _do_ serve the root key from the IC mainnet
 * even if we woudn’t, a MITM attacker could, and the whole thing breaks apart

src/frontend/src/utils/iiConnection.ts
nomeata:
```suggestion
      window.location.origin.startsWith("http://localhost:") ||
      window.location.origin.startsWith("http://127.0.0.1:") ||
```

src/frontend/src/utils/iiConnection.ts
roman-kashitsyn: What stops a rogue replica from serving it's own root key on the mainnet?

nomeata: Nothing, that’s why we shouldn’t ever fetch the key in non-dev situations

ielashi: > we still do serve the root key from the IC mainnet

Is that something that we should change soon? It's a fairly easy change.

nomeata: Yes, but one that will break many applications; the changes to the agents hardcoding the IC root key are rather recent.

kritzcreek (approved):
I'm happy with this now, but someone should review my change ;)

nomeata:
LGTM

PR #272, opened 2021-05-17
canister: export max number of users metric

roman-kashitsyn:
This change adds new metrics: the range of user numbers a canister can
serve.  It's useful for adding alerts when the canister is close to
being full.


src/internet_identity/src/main.rs
```diff
@@ -411,6 +411,11 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
             s.storage.borrow().user_count() as f64,
             "Number of users registered in this canister.",
         )?;
+        w.encode_gauge(
+            "internet_identity_max_users",
+            s.storage.borrow().max_users() as f64,
```
nomeata, line 416: Wouldn’t it be better to return the `hi-lo`?

nomeata (changes requested):

src/internet_identity/src/main.rs
roman-kashitsyn: Yes, that gives us more information.

nomeata: I guess my point was that we want to get the correct number, even if the canister is configured differently during init. But yeah, we can do the subtraction in grafana if the concrete bounds are useful (which they may be later)

nomeata (approved):

PR #273, opened 2021-05-17
Allow adding new devices from the management page

kritzcreek:
TODO:
- [x] Link to documentation on how to add additional devices that aren't connected to the current machine
- [x] Make the `Cancel` button when picking a device alias go back to the management page
- [ ] Check whether the new device is already registered


nomeata:
BTW, the `.create()` API call allows you to list a number of credentials, and will prevent the devices with these credentials from being added again. This might require beefing up `@dfinity/authentication` (or maybe not), and we should look into that (either here, or maybe note it in a ticket)
https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-excludecredentials

kritzcreek:
We're not guarding against adding the same device twice anywhere else yet, so I'm tempted to do that as a follow-up task using the `excludeCredentials` options you've outlined above. Does that sounds reasonable, or is the probability of people doing it in this flow too high?

src/frontend/src/flows/manage.ts
```diff
@@ -26,6 +28,9 @@ const pageContent = (userNumber: bigint) => html`<style>
     </p>
     <label>User Number</label>
     <div class="highlightBox">${userNumber}</div>
+    <button id="addAdditionalDevice" type="button">
+      Add an additional device
+    </button>
```
nomeata, line 33: I would have intuitively put this at the end of the list (mentally, I’d look at the list, notice what’s there, and after reading it might think there is something missing, and then be glad to see the button there). Up to you, though!

src/frontend/src/flows/manage.ts
```diff
@@ -50,6 +55,58 @@ export const renderManage = (
...
+          If you're trying to add a device that is not attached to this machine
+          try following the instructions at<br />
+          <a
+            target="_blank"
+            href="https://sdk.dfinity.org/docs/ic-identity-guide/auth-how-to.html#_add_a_device"
+            >https://sdk.dfinity.org/docs/ic-identity-guide/auth-how-to.html#_add_a_device</a
+          >
+        `,
+        detail: error.message,
+        primaryButton: "Try again",
```
nomeata, line 80: Does this really try again? Based on reading the code (I might be wrong) this seems to just go back to the manage view

src/frontend/src/flows/manage.ts
```diff
@@ -50,6 +55,58 @@ export const renderManage = (
...
+      });
+      return renderManage(userNumber, connection);
+    }
+    const deviceName = await pickDeviceAlias();
+    if (deviceName === null) {
+      return renderManage(userNumber, connection);
+    }
+    // TODO check whether newDevice is already registered
```
nomeata, line 88:
```suggestion
    // TODO check whether newDevice is already registered,
    // or better, pass existing devices to `.create` so that they
    // cannot be added again
```

nomeata (approved):
> Does that sounds reasonable, or is the probability of people doing it in this flow too high?

sounds reasonable to me.

Do you want to add selenium tests with this PR?

src/frontend/src/flows/manage.ts
kritzcreek: I wanted to make sure the button always was in view, even if the list got long enough to cause vertical scroll.

src/frontend/src/flows/manage.ts
kritzcreek: My idea was that if they wanted to try again they could initiate the process again from the management page.. Would you rather have this button say `Back to manage`?

kritzcreek:
> Do you want to add selenium tests with this PR?

I thought about it... can I even get two separate `VirtualAuthenticator`s in the same browser window?

PR #276, opened 2021-05-17
Display warning about single device lockout after login

kritzcreek:


nomeata:
Making the test pass (e.g. insert the click on continue) should be not hard? Especially when moving fast knowing that we have CI to help is is something I wouldn't give up easily.

kritzcreek:
> Making the test pass (e.g. insert the click on continue) should be not hard? Especially when moving fast knowing that we have CI to help is is something I wouldn't give up easily.

Good call. I hope I made the appropriate changes.

kritzcreek:
@nomeata I'm getting a couple of:
`TimeoutError: Waiting for element to be located By(css selector, *[id="confirmRedirect"])`

Which suggests that the `register` helper isn't getting past the warning, but in local testing that seems to work just fine. I can't get the selenium tests to run locally for me, could you maybe give it a try?

src/frontend/src/flows/displaySingleDeviceWarning.ts
```diff
@@ -0,0 +1,70 @@
...
+      your Internet Identity and at <em>least</em> one actual security key. You
+      can see how to add more devices here:
+    </p>
+    <a
+      target="_blank"
+      href="https://sdk.dfinity.org/docs/ic-identity-guide/auth-how-to.html"
+      >https://sdk.dfinity.org/docs/ic-identity-guide/auth-how-to.html</a
+    >
+    <p>
+      For now, the best recovery strategy is to add extra devices or a Yubikey
```
ggreif, line 47:
```suggestion
      For now, the best recovery strategy is to add extra devices or a YubiKey
```
Maybe we shouldn't single out one product?

nomeata (approved):

src/frontend/src/flows/displaySingleDeviceWarning.ts
Eve832: +1, we don't want to use a brand name in our UI. There could be legal repercussions.

src/frontend/src/flows/displaySingleDeviceWarning.ts
```diff
@@ -0,0 +1,70 @@
...
+  </style>
+  <div id="warningContainer" class="container">
+    ${warningIcon}
+    <h1 id="warningHeading">Warning</h1>
+    <p>
+      If you have your Internet Identity tied to only one device and you lose
+      that one device, you will be locked out.
+    </p>
+    <p>
+      Additionally on iOS and MacOS clearing your Safari browser data will
```
Eve832, line 33: comma after "Additionally"

src/frontend/src/flows/displaySingleDeviceWarning.ts
```diff
@@ -0,0 +1,70 @@
...
+      <em>delete</em> your logins from the device.
+    </p>
+    <p>
+      Hence as a best practice we always recommend you add multiple devices to
```
Eve832, line 37: No need for "Hence"

src/frontend/src/flows/displaySingleDeviceWarning.ts
Eve832: Also, I'd replace "or" with "and"

Eve832: Full suggestion: 
"Currently the best strategy to ensure you can log in is to add extra devices and at least one security key..."

PR #283, opened 2021-05-19
Describes how recovery works API and flow wise

kritzcreek:
This is a wishlist at this point. It's incomplete, in particular no updates to the backend storage format have been made yet. It also doesn't describe that recovery devices may not be used for the regular login flow.


docs/internet-identity-spec.adoc
```diff
@@ -536,6 +539,23 @@ This flow is the boring default
...
+2. Depending on their choice
+ a) If they choose the seed phrase it is generated and displayed to them with a copy button and an explanation of having to keep this phrase secure
+ b) 👆 If they choose the security key they generate new credentials for the touched device
+3. The device is added to the users account with the `is_recovery` flag set
+
+=== Flow: Use recovery
+1. On the login page the user selects "Recover my account"
+2. The user is prompted for their user number
+3. If no `DeviceData with` `is_recovery` is found for the user number an error is displayed
+4. The user is asked to provide the seed phrase or 👆 their recovery security key
```
kritzcreek, line 556: I'm not sure what to do here when a user has configured multiple recovery devices yet.

docs/internet-identity-spec.adoc
```diff
@@ -536,6 +539,23 @@ This flow is the boring default
...
 
+=== Flow: Setup recovery
+1. The user is offered two options for recovery
+  * A security key
+  * A BIP-39 seed phrase
+  * (Skip?)
+2. Depending on their choice
+ a) If they choose the seed phrase it is generated and displayed to them with a copy button and an explanation of having to keep this phrase secure
+ b) 👆 If they choose the security key they generate new credentials for the touched device
+3. The device is added to the users account with the `is_recovery` flag set
```
nomeata, line 550: Would they be asked for an alias?

docs/internet-identity-spec.adoc
nomeata: Presumably they’d have names, and the frontend would ask which one they want to use?

docs/internet-identity-spec.adoc
```diff
@@ -520,6 +522,7 @@ This flow is the boring default
 3. User sees their management screen. In particular
 +
 - Their user number
+- A button to add additional devices
 - The list of their devices, with device aliases, and a button to remove
```
nomeata, line 526:
```suggestion
- The list of their devices, with device aliases, a symbol marking recovery devices, and a button to remove
```

kritzcreek: In the longterm I'd like to show recovery devices in a different list/menu, but you're right for now this seems like the simplest thing.

docs/internet-identity-spec.adoc
kritzcreek: If they had configured multiple security keys we could do the same thing we do on login right now, but you're right picking from a list of names might be the simplest for now.

nomeata: > If they had configured multiple security keys we could do the same thing we do on login right now

No, I don’t think so: Once you have non-WebAuthn keys (e.g. recovery codes), this will not work anymore: You can’t present those to the browser and make it do the right thing anymore.

docs/internet-identity-spec.adoc
nomeata: If we want to expand the scope of this ticket, I would solve this by creating two ways of adding (normal, non-recovery) devices:

 * Add this laptop/phone as an authentication option.
 * Add a portable security device as an authentication option.

The only effect is that we set the `cross-platform` or the other on creation, else we do the same thing. Maybe store this choice with the device data, so that we can maybe show a suitable icon.

The benefit is that user’s have more guidance and more control, and less OS-specific dialogs.

And once we have that, we can allow users to designate their devices as recovery devices, by only allowing that for cross-platform devices.

docs/internet-identity-spec.adoc
nomeata: Yeah, or that. As long as they are visible and distinguishable.

docs/internet-identity-spec.adoc
kritzcreek: We can tell recovery codes from Webauthn devices by whether or not they have a CredentialId. That's how I'm planning to prompt the user to ☝️ or insert their seed phrase without them having to tell me which of the two options they chose for recovery.

nomeata: That might be short-sighted: If I use dfx to add a key, it will not have a CredentialId, but it also wouldn't be a recovery code. Better keep that information explicitly in the backend if you need to use it.

kritzcreek: Right... so an extra enum: `{ #platformAuth; #crossPlatformAuth; #recoveryCode; }` and that would keep it extensible and backwards compatible? Or would you make the full `DeviceData` a variant?

nomeata: Yes, extra enum, wrapped in `opt` so that we can decode existing fields that don't have it yet.

And I would maybe distinguish between form and purpose, i.e.
```
type = { #platformAuth, #crossPlatformAuth; #bipCodes; #plain } // plain for, e.g., dfx or other external tools
purpuse = { #hot; #recovery }
```
Because I may want to use a `crossPlatformAuth` for recovery.

docs/internet-identity-spec.adoc
```diff
@@ -197,6 +197,8 @@ type DeviceData = record {
   pubkey : DeviceKey;
   alias : text;
   credential_id : opt CredentialId;
+  purpose : { #recovery; #authentication };
+  type : { #unknown; #platform; #crossPlatform; #seedPhrase };
```
dsarlis, line 201: These should be wrapped in `opt`, no?

dsarlis, line 201: `type` is a keyword in Rust. Can we name this `key_type` or something like that?

nomeata: We don’t have to be backward-compatible towards the frontend, so here we don’t.

We need to wrap them in `opt` when we serialize/deserialize to stable memory, because there we want backward compatibility.

kritzcreek:
PTAL @dsarlis

docs/internet-identity-spec.adoc
```diff
@@ -197,6 +197,8 @@ type DeviceData = record {
   pubkey : DeviceKey;
   alias : text;
   credential_id : opt CredentialId;
+  purpose : { #recovery; #authentication };
+  key_type : { #unknown; #platform; #cross_platform; #seed_phrase };
```
dsarlis, line 201: Is this actually valid Candid? If yes, what's the difference in using `variant` instead? Saving on representation bytes?

dsarlis (commented):
LGTM. I'll leave approval to Joachim who knows better the flows.

docs/internet-identity-spec.adoc
kritzcreek: It's not. Sorry I just used the Motoko syntax for these variants, but I see you fixed it in your backend PR so we can just merge it with that, right?

kritzcreek:
@dsarlis @nomeata Think this is good to go in as well?

docs/internet-identity-spec.adoc
```diff
@@ -536,6 +540,23 @@ This flow is the boring default
 +
 (This could include a button to go to the management screen, or maybe this _is_ the management screen with a info box.)
 
+=== Flow: Setup recovery
+1. The user is offered two options for recovery
+  * A security key
+  * A BIP-39 seed phrase
+  * (Skip?)
```
nomeata, line 547: Well, let’s decide.

I’m in favor of Skip. If only so that we can test stuff more easily :-) But also to be less patronizing.

docs/internet-identity-spec.adoc
```diff
@@ -536,6 +540,23 @@ This flow is the boring default
...
+2. Depending on their choice
+ a) If they choose the seed phrase it is generated and displayed to them with a copy button and an explanation of having to keep this phrase secure
+ b) 👆 If they choose the security key they generate new credentials for the touched device
```
nomeata, line 550: Do we prevent using the same security here (using `excludeCredentials`)? Or is that future work?

docs/internet-identity-spec.adoc
```diff
@@ -536,6 +540,23 @@ This flow is the boring default
...
+3. The device is added to the users account with the purpose `#recovery` set
+
+=== Flow: Use recovery
+1. On the login page the user selects "Recover my account"
+2. The user is prompted for their user number
+3. If no `DeviceData with` the purpose `#recovery` is found for the user number an error is displayed
```
nomeata, line 556:
```suggestion
3. If no `DeviceData` with the purpose `#recovery` is found for the user number an error is displayed
```

docs/internet-identity-spec.adoc
```diff
@@ -520,7 +523,8 @@ This flow is the boring default
 3. User sees their management screen. In particular
 +
 - Their user number
-- The list of their devices, with device aliases, and a button to remove
+- A button to add additional devices
```
nomeata, line 526: Or two buttons, one to add login devices and one to add recovery devices?

nomeata (commented):
Got a few more comments/questions that might be nice to nail down.

docs/internet-identity-spec.adoc
kritzcreek: I did add a skip button, so yeah I'll update this one.

docs/internet-identity-spec.adoc
kritzcreek: It's a little more.. complicated by now:
grafik

The nag is only shown when no recovery has been configured so far, otherwise you get:
grafik

nomeata: Great, well done!

docs/internet-identity-spec.adoc
kritzcreek: We do now. I've implemented it both here and for the `Add device` button on the management page.

docs/internet-identity-spec.adoc
```diff
@@ -197,6 +197,8 @@ type DeviceData = record {
   pubkey : DeviceKey;
   alias : text;
   credential_id : opt CredentialId;
+  purpose : variant { recovery; authentication };
+  key_type : variant { unknown; platform; cross_platform; seed_phrase };
```
nomeata, line 201: The actual `.did` in the repo has type aliases for these variants. Either is fine.

nomeata (approved):
I guess this is now just documenting what we have built anyways? Rubber stamping.

docs/internet-identity-spec.adoc
kritzcreek: I'll leave it as is. For documentation purposes I think it's easier with less indirections.

PR #284, opened 2021-05-20
Frontend for the recovery flow

kritzcreek:
Implements the recovery flow. We allow the user to register both a recovery key as well as a recovery seedphrase.


src/frontend/src/components/icons.ts
```diff
@@ -165,3 +165,32 @@ export const warningIcon = html`
...
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M42.75 23.5625H39.2969V12.1875C39.2969 8.59727 36.3871 5.6875 32.7969 5.6875H20.2031C16.6129 5.6875 13.7031 8.59727 13.7031 12.1875V23.5625H10.25C9.35117 23.5625 8.625 24.2887 8.625 25.1875V44.6875C8.625 45.5863 9.35117 46.3125 10.25 46.3125H42.75C43.6488 46.3125 44.375 45.5863 44.375 44.6875V25.1875C44.375 24.2887 43.6488 23.5625 42.75 23.5625ZM17.3594 12.1875C17.3594 10.6184 18.634 9.34375 20.2031 9.34375H32.7969C34.366 9.34375 35.6406 10.6184 35.6406 12.1875V23.5625H17.3594V12.1875ZM40.7188 42.6562H12.2812V27.2188H40.7188V42.6562ZM25.0781 35.5977V38.2891C25.0781 38.5125 25.2609 38.6953 25.4844 38.6953H27.5156C27.7391 38.6953 27.9219 38.5125 27.9219 38.2891V35.5977C28.3411 35.2967 28.6539 34.8705 28.8155 34.3804C28.977 33.8903 28.9789 33.3616 28.8208 32.8704C28.6627 32.3791 28.3529 31.9508 27.9358 31.6468C27.5188 31.3429 27.016 31.1792 26.5 31.1792C25.984 31.1792 25.4812 31.3429 25.0642 31.6468C24.6471 31.9508 24.3373 32.3791 24.1792 32.8704C24.0211 33.3616 24.023 33.8903 24.1845 34.3804C24.3461 34.8705 24.6589 35.2967 25.0781 35.5977Z"
+      fill="#292A2E"
+    />
+  </svg>
```
nomeata, line 195: Tangential, would it be worth putting these actual svgs into the repo embed them from there (using `declare module` or something with webpack)?

kritzcreek: I'm not sure if that works... type wise these are `TemplateResult`s as opposed to strings, so we don't end up reparsing them with the template library over and over.

kritzcreek:
```
          <details class="displayErrorDetail">
              <summary>Error details</summary>
              <pre><!---->The user attempted to register an authenticator that contains one of the credentials already registered with the relying party.<!----></pre>
            </details>
```
Looks like the "don't add the same device twice" check trips over CI. Interestingly enough I'm not getting the error locally.

Something to investigate, but it doesn't have to block this PR.

kritzcreek:
Nice! Selenium tests passed and screenshots were updated.

src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -0,0 +1,49 @@
...
+      font-size: 1rem;
+    }
+  </style>
+  <div class="container">
+    <h1>Seedphrase</h1>
+    <p>Your seed phrase makes it easy to recover your account.</p>
+    <div class="warningBox">
+      <span class="warningIcon">${warningIcon}</span>
+      <div class="warningMessage">
+        Do <b>NOT</b> forget to save this seed phrase. Save a backup on a
```
Dfinity-Bjoern, line 17: I think we may want to add a warning that it should be kept secret.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -161,8 +174,42 @@ export class IIConnection {
...
-    return baseActor.lookup(userNumber);
+  static async fromSeedPhrase(
+    userNumber: bigint,
+    seedPhrase: string
+  ): Promise<LoginResult> {
+    const identity = await fromMnemonicWithoutValidation(
+      seedPhrase,
+      IC_DERIVATION_PATH
+    );
+    const delegationIdentity = await requestFEDelegation(identity);
```
Dfinity-Bjoern, line 185: As discussed in Zoom, this may not be necessary. But it also does not hurt, so if it makes the code easier it can stay here.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -279,3 +328,40 @@ const requestFEDelegation = async (
...
+    excludeCredentials: exclude.flatMap((device) =>
+      device.credential_id.length === 0
+        ? []
+        : {
+            id: new Uint8Array(device.credential_id[0]),
+            type: "public-key",
+          }
+    ),
+    attestation: "direct",
+    challenge: Uint8Array.from("<ic0.app>", (c) => c.charCodeAt(0)),
```
Dfinity-Bjoern, line 350: This does not matter now since we do not validate the attestation anywhere, but it may change in the future at least for the registration flow.

Dfinity-Bjoern (commented):
Overall looks great, thanks a lot for this work @kritzcreek !

Dfinity-Bjoern (approved):
Looks great!

src/frontend/src/utils/iiConnection.ts
kritzcreek: I just copied what we had been using as the defaults given from "@dfinity/identity". Want me to just pass an empty array here? Or will some keys complain about that?

PR #285, opened 2021-05-20
Support for mnemonics + deriving Ed25519 identities via Slip 10.

ielashi:


Dfinity-Bjoern (commented):
Looks good to me! I'd also like @kritzcreek to have a look.

src/frontend/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Create an Ed25519 based on a mnemonic phrase according to SLIP 0010:
+ * https://github.com/satoshilabs/slips/blob/master/slip-0010.md
+ *
+ * The derivation path is an array that is always interpreted as a hardened path.
+ * e.g. to generate m/44'/223’/0’/0’/0' the derivation path should be [44, 223, 0, 0, 0]
+ */
+export async function fromMnemonic(
+  mnemonic: string,
+  derivationPath?: number[]
```
kritzcreek, line 42: Could we put our derivation path somewhere as a constant?

src/frontend/src/crypto/mnemonic.ts
```diff
@@ -0,0 +1,11 @@
+import { blobFromUint32Array } from "@dfinity/agent";
+import { entropyToMnemonic } from "bip39";
+
+/**
+ * @returns A random BIP39 mnemonic with 256 bits of entropy.
+ */
+export function generateMnemonic(): string {
+  const entropy = new Uint32Array(32);
+  crypto.getRandomValues(entropy);
+  return entropyToMnemonic(blobFromUint32Array(entropy).toString("hex"));
```
kritzcreek, line 10: I think we should explicitly set the word list here, I'm not sure if it picks something locale dependent by default?

kritzcreek (changes requested):
API looks great, I think we'll have to update the build system so we don't include _all_ the wordlists: https://github.com/bitcoinjs/bip39#removing-wordlists-from-webpackbrowserify

But I can do that in a follow-up PR.

src/frontend/src/crypto/mnemonic.ts
ielashi: I believe it's English by default, but I made it explicit to make it clearer.

src/frontend/src/crypto/ed25519.ts
ielashi: This is meant to be a generic API and shouldn't be coupled to a specific use-case. If there's a flow that needs a specific path, then I'd argue that the definition of that path should be part of that flow, not here.

kritzcreek: Got it. We'll find a good spot in the codebase for it then.

kritzcreek (approved):

PR #288, opened 2021-05-21
Clarify why we don’t use `Candid.candidFile`

nomeata:


kritzcreek (approved):

nomeata:
Ups, this contains more than it should. sorry.

PR #289, opened 2021-05-21
Optionally verify that a mnemonic is valid.

ielashi:
When generating an ed25519 from a mnemonic, we didn't validate that the mnemonic
is correct. This PR adds this validation, with an option to skip it. The reason I added
an option to skip it is because of the developer guidelines from https://www.npmjs.com/package/bip39:

> Please remember to allow recovery from mnemonic phrases that have invalid checksums (or that you don't have the wordlist)
>
> When a checksum is invalid, warn the user that the phrase is not something generated by your app, and ask if they would like to use it anyway. This way, your app only needs to hold the wordlists for your supported languages, but you can recover phrases made by other apps in other languages.
>
> However, there should be other checks in place, such as checking to make sure the user is inputting 12 words or more separated by a space. ie. phrase.trim().split(/\s+/g).length >= 12


Dfinity-Bjoern (commented):
Thanks @ielashi! Makes sense to me; not approving yet since I'd like @kritzcreek to have a look as well.

src/frontend/src/crypto/ed25519.ts
```diff
@@ -32,13 +32,20 @@ export async function fromSeedWithSlip0010(
...
+ * @param skipValidation if true, validation checks on the mnemonics are skipped.
  */
 export async function fromMnemonic(
   mnemonic: string,
-  derivationPath?: number[]
+  derivationPath: number[] = [],
+  skipValidation = false
 ): Promise<Ed25519KeyIdentity> {
+  if (!skipValidation && !validateMnemonic(mnemonic)) {
+    throw new Error(`Invalid mnemonic: ${mnemonic}`);
```
kritzcreek, line 46: Could we not do this as an exception, but rather as a return value? I don't think this is exceptional behaviour, it's a situation we should expect callers to handle.

The easy way would be to return `Ed25519KeyIdentity | null` unless we see more possible error conditions, in which case we should use a variant.

kritzcreek (changes requested):
Thanks, that's going to really help the UX!

src/frontend/src/crypto/ed25519.ts
kritzcreek: Not that I think about it, we should probably split this into two functions, so the return type on the version that never fails can remain correct. WDYT?

ielashi: Agreed and done.

kritzcreek (approved):

PR #292, opened 2021-05-22
Prefixes the seed phrase with the user number

kritzcreek:


Dfinity-Bjoern (approved):
Thanks! While it isn't clear to me that we want to do this long-term, I'll take @jancamenisch 's opinion as more important for now. Also we should be able to change this later if we want to.

PR #294, opened 2021-05-23
Catch failure to set up a recovery mechanism

kritzcreek:
This can happen because the user ran out of space for additional devices, or for other network/unforseeable reasons.


Dfinity-Bjoern (approved):
Looks good!

PR #297, opened 2021-05-26
Disallow removing the last device

kritzcreek:


Dfinity-Bjoern (approved):
That appears to be an effective measure :-)

PR #304, opened 2021-05-27
Fix selenium test upon removing last device

gobengo:
I noticed this part of the selenium tests were failing when for my other README changes, so tried to fix it.

## Description

I think the tests had not been updated upon https://github.com/dfinity/internet-identity/pull/297

The old test asserted that you could remove the last device after an alert, and then would be sent back to the Welcome screen.
Now it asserts you can't, and you stay on the same page with the device *not* removed.

## Motivation and Context

* moving towards green tests
* notably, there's still a selenium test that fails

## How Has This Been Tested?
* github ci + selenium

## Types of changes

- [x] test fix

## Checklist:

### DIFF
--- a/src/frontend/src/selenium.test.ts
+++ b/src/frontend/src/selenium.test.ts
@@ -779,14 +779,13 @@ test("Screenshots", async () => {
     expect(alertText1).toBe(
       "This will remove your current device and you will be logged out"
     );
+    await driver.wait(until.alertIsPresent());
     const alert2 = driver.switchTo().alert();
     const alertText2 = await alert2.getText();
     await alert2.accept();
-    expect(alertText2).toBe(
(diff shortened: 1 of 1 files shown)


gobengo:
Timeboxed this but couldn't get it fixed. :/

gobengo:
Searched web as last resort, and looks like this 'back to back alert' case is quirky for others too https://stackoverflow.com/questions/48082926/how-to-handle-back-to-back-alerts-on-selenium-webdriver-automation

gobengo:
I fixed one test but another still fails (and was there before). I think this is an improvement regardless.

ggreif (approved):
Thanks for doing this, it was quite distracting  to me when doing #298.

LGTM!

I looked at https://github.com/dfinity/internet-identity/pull/304/checks?check_run_id=2688005907#step:12:408 and it seems to appear in the context of a different flow.

PR #309, opened 2021-06-02
dfx.json: Don’t say “mainnet”, say “mercury”

nomeata:
as it seems to be the “more proper” name.


jplevyak:
We agreed to call this mainnet.

jplevyak:
@samuelburri was involved in that naming discussion

samuelburri:
@nomeata @jplevyak the term `mainnet` was agreed with Brendan and Eric. Background, Eric expects that one day we might have other 'nets', e.g. in China.

nomeata:
Thanks for the clarification!

PR #311, opened 2021-06-03
Discourage user verification

Dfinity-Bjoern:
The PR changes the use of Web Authentication so that User Verification is not used anymore.

## Description

Change `userVerification` from `preferred` to `discouraged`.

## Motivation and Context
When using web authentication, discourage user verification. Note that there are [two different but related concepts](https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/User_Presence_vs_User_Verification.html) in web authentication:

- User Presence: means the user has to tickle the Yubikey
- User Verification: means that the user has to enter a PIN to use the Yubikey

Most users have no PIN set up on their Yubikey so our previous setting of `preferred` would mean the same as not requesting the PIN. But for people using Windows 10 (helpful discussion [here](https://github.com/dfinity/internet-identity/pull/311#issue-660741581)), the OS automatically configures a PIN on the Yubikey so that with the setting of `preferred`, the PIN will actually be requested.

While User Verification is actually discouraged for 2FA use cases, it is actually *encouraged* for use cases such as ours where no additional password is required. So why deactivate it now by changing the setting to `discouraged`? The reason is that whether User Verification has been performed is included only in the `authenticatorData`. This, however, is never actually checked on the replica side: a fraudulent use of the front end could thus always set User Verification to `discouraged`. Setting it to `preferred` in our front end thus only bugs the honest users and does not prevent attacks.

On a longer time frame, we should record `User Verification` as a property of the records and check the `authenticatorData` in the canister. Since that requires significant changes, I created a ticket for it here: https://dfinity.atlassian.net/browse/II-1

## How Has This Been Tested?

So far test has only been done locally. Once we deployed it to test net, I will ask @Eve832 to test as well since she was affected by the previous behavior.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:


nomeata (approved):
LGTM

PR #325, opened 2021-06-08
Show principal during delegation

Dfinity-Bjoern:
Shows the user's principal as observed by the application prior to the redirect.

## Description
Introduces a new query function `get_principal` on the canister, which computes the user principal given the user number and the front end URL. As this endpoint allows linking user principals across multiple front ends, it check authorization via any key associated with that user number.

Upon requesting user authorization to generate the delegation, displays the principal to the user. This requires the new `get_principal` API as we should not request the delegation prior to getting user consent.

## Motivation and Context
Currently it seems difficult for users to understand that every front end application they use on the IC with II will perceive them under a different principal. The first attempt for mitigating this is presenting the actual principal that will be used upon redirecting the user. Addresses [II-13](https://dfinity.atlassian.net/browse/II-13).

## How Has This Been Tested?
Backend tests for failing authentication, and for correctness of computed principals. Manual front-end test with a local deployment.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


JensGroth (approved):

kritzcreek (approved):
Code looks good.

PR #329, opened 2021-06-11
Change in terminology

Dfinity-Bjoern:
This PR changes the wording as follows:

- An "account" on the canister is referred to as an *identity*. One *user* (i.e. real world person) can have multiple *identities* (i.e. accounts on the canister), each of which is specified by its *ID number*, and can be associated with multiple application-specific *pseudonyms*.
- We also remove the wording of Internet Identity *Service*.


kritzcreek:
How did we decide on the new wording? Changing it now will cause confusion and means we definitely can't change it again anytime soon, for example when we go through a redesign,

Dfinity-Bjoern:
> How did we decide on the new wording? Changing it now will cause confusion and means we definitely can't change it again anytime soon, for example when we go through a redesign,

See mail from Brendan earlier today, linking this document: https://docs.google.com/document/d/1Lxu7PpvLHElnUoA3d0xiXoCrXq69emn62R782j21HIA/edit

nomeata (approved):
I am sure you are feeling very productive after such a change :-)

imageoracle:
I think removing the word service is a good idea too .  It's an ii layer

PR #330, opened 2021-06-11
Selenium tests: Use elem.clear()

nomeata:
TIL that the WebDriver API provides `.clear()`, which is a cleaner way
to clear an input box than Ctrl-A Delete, and should work also on OSX.


nomeata:
This would have been useful to @Dfinity-Bjoern earlier this week, I presume.

kritzcreek (approved):

Issue #331, opened 2021-06-11
Unable to recover account using first four letters of words in seed phrase


paulyoung:
The source code for Internet Identity suggests that the seed phrase is a BIP39 mnemonic.

Some backup solutions only allow storing the first 4 letters in accordance with the BIP39 protocol. My understanding is that the first-four-letter combinations are unique and sufficient to recover the full word.

However, when I try recovering an account using only the first four letters of each word in the seed phrase I get the "invalid seedphrase" message.

nomeata:
Did you also shorten the ID number in the beginning? Not sure if the II frontend checks that number.

paulyoung:
I tried both shortening that number and omitting the number.

nomeata:
The README of https://www.npmjs.com/package/bip39 very prominently says

> **Please remember to allow recovery from mnemonic phrases that have invalid checksums (or that you don't have the wordlist)**
> 
> When a checksum is invalid, warn the user that the phrase is not something generated by your app, and ask if they would like to use it anyway. This way, your app only needs to hold the wordlists for your supported languages, but you can recover phrases made by other apps in other languages.
> 
> However, there should be other checks in place, such as checking to make sure the user is inputting 12 words or more separated by a space. ie. phrase.trim().split(/\s+/g).length >= 12

So I guess we should do that :-)

paulyoung:
It seems like someone already did: https://github.com/dfinity/internet-identity/pull/289

kritzcreek:
> Did you also shorten the ID number in the beginning? Not sure if the II frontend checks that number.

You'll have to omit the number probably. It's not part of the phrase and the frontend just strips it when reading the seedphrase.

> It seems like someone already did: #289

We're strictly checking that validation passes in the frontend. If we want to allow this "stripped" seed phrase to work, I think the best way of going about this would be to let the user continue even if we have validation failures here: https://github.com/dfinity/internet-identity/blob/17e5c867ab0c8d8500721e45ab8b87d7ae8e75ef/src/frontend/src/flows/recovery/inputSeedPhrase.ts#L53-L59

Something like a "Continue anyway" button.

That way we can still give the user an early indication that they might have pasted an incomplete seed phrase.

paulyoung:
If I implement this do I get to submit the proposal to the NNS? 🙂

nomeata:
Why not? :-)

paulyoung:
[...]

PR #335, opened 2021-06-12
Allow users to continue even if the seed phrase can't be validated

paulyoung:
## Description
* Add an optional secondary button to the "display error" component
* Update the "input seed phrase" flow to allow users to continue even if the seed phrase can't be validated
* Fix some outdated documentation

## Motivation and Context
This intends to address https://github.com/dfinity/internet-identity/issues/331

> Some backup solutions only allow storing the first 4 letters in accordance with the BIP39 protocol. My understanding is that the first-four-letter combinations are unique and sufficient to recover the full word.
>
> However, when I try recovering an account using only the first four letters of each word in the seed phrase I get the "invalid seedphrase" message.

## How Has This Been Tested?
I tested this locally by creating a new identity and generating a recovery seed phrase. I then tried to recover the account using the seed phrase as it was generated and also with the first 4 letters of each word in the recovery seed phrase (omitting the leading ID number)

Screen Shot 2021-06-12 at 3 54 48 PM

## Types of changes

## Checklist:


paulyoung:
@nomeata @kritzcreek I've opened this as draft since I'm getting the following error despite having done everything I set out to do.

Screen Shot 2021-06-12 at 3 55 10 PM

I think this now requires the attention of someone who understands the cryptography involved to determine if the `bip39` package doesn't support entering the seed phrase in the first-four-letters format and/or what other changes are required to retrieve an identity that way.

imageoracle:
I think a solution would be to somehow make it so the bounded identity can look and not touch components at least. a button for broken bits so to speak .  Just an idea .  This would dampen some of the frustrations and provide a curious entrepreneur with a peek instead of error messages. If this is a bad idea , let me know . I'm brainstorming .

nomeata:
I wouldn't add buttons for this use case. UX complexity is a very scarce resource, and our “normal” users probably wouldn't even expect that they may truncate the seed phrase (I certainly didn't, and who is more normal than me???).

And the advanced user can, in the worst case, check the source, find the word list, and use that to “reinflate” the seed phrase.

Or write a CLI took to add a new device.

paulyoung:
Should I close this and #331 then?

I think it’s reasonable to figure it out from a word list like this one: https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt

kritzcreek:
Given that the bip39 library authors don't want to implement this one: https://github.com/bitcoinjs/bip39/issues/158

I'd say we don't do it here either, but we do know that we can help someone that ends up with only the leading 4 characters if they contact us.

PR #338, opened 2021-06-27
Adds Windows Hello support by enabling RSA signatures

kritzcreek:
Enables RSA keys, which in turns enables Windows Hello. Needs updates to the replica to be able to verify RSA signatures.

TODO:

- [x] Use the DER wrapping from this PR in agent-js
- [x] Get dfinity/ic-hs#26 merged


kritzcreek:
@robin-kunzler I've changed the tests so that one of the two "test devices" uses an RSA key instead of an ECDSA one. Does that suffice for testing that RSA keys works, or would you prefer more explicit tests? (Really ic-hs is mostly testing itself here, because our canister doesn't actually parse these public keys, it just stores them as bytes)

Depends on https://github.com/dfinity/ic-hs/pull/26

robin-kunzler:
@kritzcreek : 

> @robin-kunzler I've changed the tests so that one of the two "test devices" uses an RSA key instead of an ECDSA one. Does that suffice for testing that RSA keys works, or would you prefer more explicit tests? (Really ic-hs is mostly testing itself here, because our canister doesn't actually parse these public keys, it just stores them as bytes)

If I understand correctly: 
* switching the device2 keys to RSA shows that the canister is able to deal with devices that use RSA keys. As you say, the canister just stores the bytes. 
* the replica must accept RSA signatures, but since the replica is ic-ref here, this part only tests ic-ref.

This looks good to me in terms of coverage on that level. 

These tests only test the canister, right? Is it possible to write a selenium test for the UI that uses RSA/Windows hello?

kritzcreek:
Ready when you are ;)

src/frontend/src/utils/iiConnection.ts
```diff
@@ -353,7 +353,6 @@ export const creationOptions = (
 ): PublicKeyCredentialCreationOptions => {
   return {
     authenticatorSelection: {
-      userVerification: "preferred",
```
robin-kunzler, line 356: Can you briefly explain why you remove the following? 
```
userVerification: "preferred",
attestation: "direct",
```
* It seems `userVerification: "preferred"` is the default, so removing it does not have an effect, correct? Also, is there a reason for not using the seemingly more secure "required"? 
* Removing attestation seems to now default to "none". So this means we are just not interested in attestation here?

robin-kunzler (commented):
@kritzcreek : Looks good to me, thanks! I only have one, probably naive question, please see below.

src/frontend/src/utils/iiConnection.ts
kritzcreek: > It seems userVerification: "preferred" is the default, so removing it does not have an effect, correct? Also, is there a reason for not using the seemingly more secure "required"?

It doesn't have an effect, right. This is what causes users to have to enter the additional PIN for yubikeys on Windows, but there's no setting which actually disables it on creation. I'm just going to revert this change, there's no point.

> Removing attestation seems to now default to "none". So this means we are just not interested in attestation here?

To make `attestation` useful, we'd have to verify it on the other end, which we're not doing. We have some plans for it in the future, but it would be a large undertaking. @Dfinity-Bjoern has some more details if you're curious.
I just figured we shouldn't make the authenticator produce data we don't end up using, and potentially even exclude using security keys that can't produce attestion.

robin-kunzler: Thanks for providing the context. Sounds good to me. 

Regarding "Also, is there a reason for not using the seemingly more secure "required"?" 
I assume we actually don't want to ask the users to enter a pin when they use their Yubikey.

robin-kunzler (approved):
@kritzcreek: LGTM, thanks for the PR and for answering my questions!

Dfinity-Bjoern (approved):
Looks good, thanks a lot @kritzcreek !

PR #339, opened 2021-06-28
Use a different loading animation

kritzcreek:
Edges look pretty rough, I've made the image smaller to make it less obvious. What do you think? @Dfinity-Bjoern


Dfinity-Bjoern:
Good question. This seems pretty far from perfect. Maybe we can set up a test deployment where we can have Dom and Jonathan try it out? I think we may be able to convince them that, with the current size restrictions, we should opt for a different type of graphics that can be represented more compactly. What do you think?

kritzcreek:
This is already using the `webp` encoded animation, so it's not about the format at this point. It adds about 100kb compared to the old spinner which is reasonable, but I think the quality is too low now.

dsarlis (approved):

PR #341, opened 2021-07-12
Teminology updates

kritzcreek:
User Number -> Identity Anchor
Login -> Authenticate

### DIFF
--- a/src/frontend/src/flows/confirmRedirect.ts
+++ b/src/frontend/src/flows/confirmRedirect.ts
@@ -1,7 +1,7 @@
 import { html, render } from "lit-html";
 import { FrontendHostname } from "../../generated/internet_identity_types";
 
-const pageContent = (hostName: string, principal: string) => html`
+const pageContent = (hostName: string) => html`
   <style>
     #confirmRedirectHostname {
       font-size: 0.875rem;
@@ -16,19 +16,16 @@ const pageContent = (hostName: string, principal: string) => html`
     <h1>Authorize Login</h1>
     <p>Proceed to login in to:</p>
     <div id="confirmRedirectHostname" class="highlightBox">${hostName}</div>
-    <p>With application-specific pseudonym:</p>
-    <div id="confirmRedirectPrincipal" class="highlightBox">${principal}</div>
     <button id="confirmRedirect" class="primary">Proceed</button>
     <button id="cancelRedirect">Cancel</button>
   </div>
 `;
 
 export const confirmRedirect = async (
-  hostName: FrontendHostname,
-  userPrincipal: string
+  hostName: FrontendHostname
 ): Promise<boolean> => {
   const container = document.getElementById("pageContent") as HTMLElement;
-  render(pageContent(hostName, userPrincipal), container);
+  render(pageContent(hostName), container);
   return init();
 };
 
--- a/src/frontend/src/flows/confirmRedirect.ts
+++ b/src/frontend/src/flows/confirmRedirect.ts
@@ -13,8 +13,8 @@ const pageContent = (hostName: string) => html`
     }
   </style>
   <div class="container">
-    <h1>Authorize Login</h1>
-    <p>Proceed to login in to:</p>
+    <h1>Authorize Authentication</h1>
+    <p>Proceed to authenticate with:</p>
     <div id="confirmRedirectHostname" class="highlightBox">${hostName}</div>
     <button id="confirmRedirect" class="primary">Proceed</button>
     <button id="cancelRedirect">Cancel</button>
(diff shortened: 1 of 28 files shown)


src/frontend/src/flows/confirmRedirect.ts
```diff
@@ -1,7 +1,7 @@
 import { html, render } from "lit-html";
 import { FrontendHostname } from "../../generated/internet_identity_types";
 
-const pageContent = (hostName: string, principal: string) => html`
+const pageContent = (hostName: string) => html`
```
robin-kunzler, line 4: Here I am assuming that it was decided to remove the pseudonym from the redirect confirmation, right?

src/frontend/src/flows/displaySingleDeviceWarning.ts
```diff
@@ -29,34 +29,35 @@ const pageContent = () => html`
...
-      <em>delete</em> your logins from the device.
+      <em>delete</em> your authentication keys from the respective device,
+      disabling it.
     </p>
     <p>
-      As a best practice we always recommend you add multiple devices to your
-      Internet Identity and at <em>least</em> one actual security key. You can
-      see how to add more devices here:
+      As a best practice we always recommend you add multiple devices to an
+      Identity Anchor and at <em>least</em> one portable authenticator (usb
```
robin-kunzler, line 42: Nit: maybe use 'USB' in upper case letters?

src/frontend/src/utils/userIntent.ts
```diff
@@ -15,13 +15,25 @@ export const intentFromUrl = (url: URL): UserIntent => {
...
   switch (intent.kind) {
     case "addDevice":
-      return "add your new device to your Internet Identity";
+      return "Authenticate to add your new device";
     case "auth":
-      return "authenticate using your Internet Identity";
+      return "Authenticate using Internet Identity";
     case "manage":
-      return "manage your Internet Identity";
+      return "Authenticate using Internet Identity";
```
robin-kunzler, line 25: Did something go wrong here? Should it say "Manage your Internet Identity"?

src/frontend/src/utils/userIntent.ts
```diff
@@ -15,13 +15,25 @@ export const intentFromUrl = (url: URL): UserIntent => {
...
+  }
+};
+
+// TODO: Remove me
```
robin-kunzler, line 29: Should this code be removed or should the TODO be associated to a Jira ticket?

robin-kunzler (approved):
@kritzcreek : Thanks a lot for all the terminology changes, this looks good to me. I only have a few minor comments, see below. 

I did not cross check against Slack or the doc. I just read the texts and checked that the texts look reasonable to me.

src/frontend/src/flows/confirmRedirect.ts
kritzcreek: Yes, we decided to not roll that change into the terminology changes, because we're still debating whether it's a good idea.

src/frontend/src/utils/userIntent.ts
kritzcreek: This is the bit I talked about where previously it said "Authenticate to manage your Internet Identity". Now we want the "manage your Internet Identity" bit to disappear but just saying "Authenticate to <nothing>" doesn't work so I had to change things around a bit.

src/frontend/src/utils/userIntent.ts
kritzcreek: We should probably make a JIRA ticket to deal with the follow-up changes and cleanup we need to do after the terminology changes. We should mention this TODO in that ticket.

src/frontend/src/flows/displaySingleDeviceWarning.ts
kritzcreek:
```suggestion
      Identity Anchor and at <em>least</em> one portable authenticator (USB
```

src/frontend/src/utils/userIntent.ts
robin-kunzler: I created [II-36](https://dfinity.atlassian.net/browse/II-36) for this. Ideally, IMO all TODOs that we commit in the future should have the form
```
// TODO (II-??): Blah
```
and refer to a Jira ticket. This also conforms with the [guidelines](https://docs.google.com/document/d/1AQQGnvUU05kPg8_MahJtbyJ2KlPIPYbFblHNR4B3I9A/edit) we had to follow for shipshaping code for launch.

src/frontend/src/utils/userIntent.ts
robin-kunzler: Ok, thanks for explaining. I also added this point to [II-36](https://dfinity.atlassian.net/browse/II-36) in case we want to address this again when cleaning this file up.

PR #342, opened 2021-07-15
Adds coordinates for our test net

kritzcreek:


canister_ids.json
```diff
@@ -1,6 +1,6 @@
 {
   "internet_identity": {
-    "messaging": "rdmx6-jaaaa-aaaaa-aaadq-cai",
+    "identity": "rdmx6-jaaaa-aaaaa-aaadq-cai",
```
robin-kunzler, line 3: Another naive question: Why was this called messaging before? Was this just wrong?

dfx.json
```diff
@@ -18,8 +18,8 @@
       "bind": "127.0.0.1:8000",
       "type": "ephemeral"
     },
-    "messaging": {
-      "providers": ["http://10.12.36.13:8080"],
+    "identity": {
+      "providers": ["http://[2a00:fb01:400:42:5000:60ff:fed5:8464]:8080"],
```
robin-kunzler, line 22: probably naive from me, but shouldnt this be 
```
https://rdmx6-jaaaa-aaaaa-aaadq-cai.identity.dfinity.network/
```
instead of the dashboard page?

robin-kunzler (commented):
@kritzcreek : Thanks for adapting the configurations, I have two minor questions, see below.

canister_ids.json
kritzcreek: We used the `messaging` testnet as our testnet for a while, and the configuration remained afterwards.

dfx.json
kritzcreek: I don't know exactly what this points to, but when it's set up like this you can use `dfx` to make calls against the running canister on our testnet.

robin-kunzler: Ok, if that works that looks good to me!

robin-kunzler (approved):
@kritzcreek : Thanks for answering the questions, LGTM!

PR #343, opened 2021-07-20
II-45: Changes capitalization to 'Identity Anchor'

kritzcreek:


src/frontend/src/flows/loginUnknown.ts
```diff
@@ -39,19 +39,19 @@ const pageContent = (userIntent: UserIntent) => html` <style>
...
       id="registerUserNumber"
-      placeholder="Enter Identity anchor"
+      placeholder="Enter Identity Anchor"
     />
     <button type="button" id="loginButton" class="primary">Authenticate</button>
     ${userIntent.kind === "addDevice"
       ? html`<div class="spacer"></div>`
       : html`<div class="textLink" id="registerSection">
-            New user?
+            New?
```
robin-kunzler, line 54: Just because this seems to be an unrelated change: Do you actually want to change "New user" to "New"?

robin-kunzler (approved):
@kritzcreek : Thanks for adapting the capitalization! Looks good to me. 

* I also did a search through the code base and did not find any other mentions of identity anchor where the capitalization is still wrong. 
* Only have one minor question, see below.

src/frontend/src/flows/loginUnknown.ts
kritzcreek: Good spot. Jan asked me to make that change so I added it but it doesn't really fit the topic of this PR. Would you prefer I move it to a later terminology PR?

robin-kunzler: I'm completely fine with leaving it in this PR, just asked to make sure that this change was intentional!

kritzcreek: Great! Understood.

PR #346, opened 2021-07-21
II-38: Updates npm dependencies to their latest versions

kritzcreek:
Also fixes a NPE related to window.opener not being defined in certain scenarios.

Most of this work was first done in #338 but it makes sense to land these changes before then individually so the branch doesn't bitrot. Also addresses the `npm audit` findings, even if they just applied to our devDependencies.


package.json
```diff
@@ -13,51 +13,52 @@
...
     "stream-browserify": "3.0.0",
-    "style-loader": "^2.0.0",
-    "terser-webpack-plugin": "5.1.1",
-    "ts-jest": "^26.5.4",
-    "ts-loader": "8.1.0",
-    "ts-node": "^9.1.1",
-    "typescript": "^4.2.4",
+    "style-loader": "^3.2.1",
+    "terser-webpack-plugin": "5.1.4",
+    "ts-jest": "^27.0.3",
```
robin-kunzler, line 42: nit: Seems this could even be updated to 27.0.4, since 18 hours ago ;-) 
npm outdated
```
Package  Current  Wanted  Latest  Location              Depended by
ts-jest   27.0.3  27.0.4  27.0.4  node_modules/ts-jest  internet-identity
```

package.json
```diff
@@ -13,51 +13,52 @@
...
+    "ts-loader": "9.2.3",
+    "ts-node": "^10.1.0",
+    "typescript": "^4.3.5",
     "url-loader": "^4.1.1",
-    "util": "0.12.3",
-    "webpack": "5.24.4",
-    "webpack-cli": "4.5.0",
+    "util": "0.12.4",
+    "webpack": "5.45.1",
+    "webpack-cli": "4.7.2",
```
robin-kunzler, line 49: `webpack-dev-server` depends on vulnerable version of chokidar -> glob-parent 5.1.2. The fix suggested by npm audit is downgrading to 1.16.5, which is probably not an option. Also, given that this is only a dev dependency, I suggest we don't act on this now.

robin-kunzler (approved):
@kritzcreek : Great, thanks for updating the dependencies and fixing the bugs. I only have two minor comments, please see below.

package.json
kritzcreek: That'll just have to be included in the next round of updates ;)

PR #348, opened 2021-07-22
Adapt terminology in ic-identity-guide/what-is-ic-identity

robin-kunzler:
This adapts the what-is-ic-identity documentation to the new terminology.

Also moves the introduction from the how-to here and replaces it by a link since that is duplicated at the moment.


docs/modules/ic-identity-guide/pages/what-is-ic-identity.adoc
```diff
@@ -6,37 +6,30 @@
...
 
-Unlike most authentication services, your Internet Identities do not require you to set and manage passwords or provide any personal identifying information to applications. Instead, you use the authentication profiles that you create such as facial recognition from a smartphone, your computer unlock password, or a security key.
+Unlike most authentication methods, Internet Identity does not require you to set and manage passwords or provide any personal identifying information to dapps or to Internet Identity.
```
robin-kunzler, line 11: Note that this part is simply moved from the how-to above.

docs/modules/ic-identity-guide/pages/what-is-ic-identity.adoc
```diff
@@ -6,37 +6,30 @@
...
+Internet Identity enables you to use security modules built into your smart phone, your laptop or security keys plugged into the USB port of your computer to authenticate and authorize access to dapps running on the {IC}.
 
-Applications that integrate with Internet Identity prompt you to register or log in with a previously-registered Identity Anchor. 
-Registering a device is a simple process that generates a unique cryptographic key for you that is stored on the Internet Computer blockchain. You can then use your Identity Anchor to register additional devices, so that you can access applications across all of your devices.
+Applications that integrate with Internet Identity prompt you to authenticate using an Identity Anchor. If you don't have an Identity Anchor yet, creating one is a simple process that generates a unique cryptographic key for you, whose public part is stored on the Internet Computer blockchain. You can then use your Identity Anchor to add additional devices, so that you can access applications across all of your devices.
```
kritzcreek, line 19: I know this hasn't changed, but the "a unique cryptographic key for you", wording sounds like "unique cryptographic key _identifiying_ you" which is a bit unfortunate.

After all the cryptographic _keys_(!!) are generated by the Yubikeys and TPM's and can be swapped out at any point. The only thing the process really generates is the number for the Identity Anchor.

kritzcreek (approved):
Great changes, thanks!

docs/modules/ic-identity-guide/pages/what-is-ic-identity.adoc
robin-kunzler: Thanks @kritzcreek . I agree we should improve this. I rewrote parts of this paragraph to:
- Make it more clear that the keys are associated to the devices that are added to an anchor (and not the anchor directly)
- Clarify that a pair of keys is generated. 
Could you take another look?

robin-kunzler (commented):
@kritzcreek : Thanks for the review. I addressed your comment, see below. Could you take another look?

kritzcreek:
> I addressed your comment, see below. Could you take another look?

My review was actually in response to your recent changes ;) Concurrency is tricky :D

Changes look great!

PR #349, opened 2021-07-22
Christoph/selenium fixes

kritzcreek:
I've worked around the problem where trying to add a second authenticator would fail most of the time because the first authenticator was still present by removing the first one. Unfortunately there's no API to set one or the other authenticator to be the `active` one.

This is a bit crude, because it means we can't test that adding a second credential to an anchor still allows the first to work, but I'm also not sure how to support that kind of a test with the current webdriver API.

I'm still expecting some amount of flakyness because of screenshot timeouts, but a restart here or there should let us get past those for now. Splitting the screenshot job up is still worth it, but I think this PR in its current state should give us a chance of merging PRs with green CI checkmarks again.


src/frontend/src/selenium.test.ts
```diff
@@ -546,7 +546,7 @@ test("Register new identity and add additional device", async () => {
     await on_Main(DEVICE_NAME1, driver);
     await on_Main(DEVICE_NAME2, driver);
 
-    await await on_Main_Logout(driver);
+    await on_Main_Logout(driver);
```
robin-kunzler, line 549: naive question: why is only one await needed here, but in other places (like line 526) two awaits are used?

src/frontend/src/selenium.test.ts
```diff
@@ -533,8 +533,8 @@ test("Register new identity and add additional device", async () => {
     await driver.get(II_URL);
     const userNumber = await registerNewIdentity(driver);
     await on_Main(DEVICE_NAME1, driver);
+    await removeVirtualAuthenticator(driver, firstAuthenticator);
```
robin-kunzler, line 536: Maybe add a comment explaining that the first authenticator is removed as a workaround to flakyness, since this is not really expected / obvious?

demos/sample-javascript/webpack.config.js
```diff
@@ -47,6 +47,11 @@ module.exports = {
     },
   },
   devtool: 'source-map',
+  devServer: {
+    client: {
```
robin-kunzler, line 51: Another naive question: where do I find the docs for this? I looked [here](https://webpack.js.org/configuration/dev-server/), but this does not list a 'client' option. 

What does setting `overlay: false` do exactly?

robin-kunzler (approved):
Thanks @kritzcreek , looks good to me. I have three minor comments/questions, see below.

demos/sample-javascript/webpack.config.js
kritzcreek: > Another naive question: where do I find the docs for this?

Not a naive question at all... as this is JavaScript you must never trust any documentation, because it's all out-of-date by the time it's published. I ended up finding the right config after trial-and-error. 

> What does setting overlay: false do exactly?

It disables a `webpack-dev-server` (that hosts the html/js for the sample JS app) feature, where it puts a screen-wide overlay over your app to display any warnings/errors that occurred during JS bundling. We don't want that for our selenium tests, as the overlay blocks access to the fields/buttons we actually want to fill in/click.

src/frontend/src/selenium.test.ts
kritzcreek: Double await is never *needed* (because of JS's questionable design to auto-flatten Promises) it's basically a No-Op. I'll find the other spots this is happening and correct them as well.

demos/sample-javascript/webpack.config.js
robin-kunzler: > I ended up finding the right config after trial-and-error.

Haha, ok I see. 

> What does setting overlay: false do exactly?
It disables a webpack-dev-server (that hosts the html/js for the sample JS app) feature, where it puts a screen-wide overlay over your app to display any warnings/errors that occurred during JS bundling. We don't want that for our selenium tests, as the overlay blocks access to the fields/buttons we actually want to fill in/click.

I see, thanks for explaining!

PR #352, opened 2021-07-26
Updates instructions on how to deploy to our testnet

kritzcreek:
Updates outdated deployment instructions for our testnet, found when deploying the RSA branch with @robin-kunzler last week.

### DIFF
--- a/docs/internet-identity-spec.adoc
+++ b/docs/internet-identity-spec.adoc
@@ -603,8 +603,6 @@ This section needs to describe aspects like
 
 You will need the `ic-admin` tool. You have various options
 
-* Download `ic-admin` for your platform from https://blobules.dfinity.systems/dfinity-ci-build.dfinity/ic-admin/0.1.0/[blobules]. If you are on macOS you will likely need to right-click on `ic-admin` and then select `Open` (to avoid having it considered malware).
-
 * You can build it from source. In a checkout of `dfinity`, run:
 +
 [source,bash]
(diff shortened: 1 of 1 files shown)


docs/internet-identity-spec.adoc
```diff
@@ -716,14 +698,39 @@ dfx canister --network mainnet --no-wallet info internet_identity
...
+----
+
+This will create the Wasm file you want to use for the following deployment steps at `./internet_identity.wasm`.
+
+==== Installing on testnets
+
+Submit the proposal to install the canister on our `identity` testnet (replace for other testnets as appropriate):
+[source,bash]
+----
+ic-admin --nns-url "http://[2a00:fb01:400:42:5000:60ff:fed5:8464]:8080/" propose-to-change-nns-canister --test-neuron-proposer --canister-id rdmx6-jaaaa-aaaaa-aaadq-cai --mode reinstall --wasm-module-path internet_identity.wasm --arg arg.in
```
robin-kunzler, line 723: I compared to the command I used in our session. I used `--mode install` here. Does that matter or is `reinstall` fine too?

robin-kunzler (commented):
Thanks for adapting the docs @kritzcreek ! Looks good, I just have one question, see below.

docs/internet-identity-spec.adoc
kritzcreek: > I guess reinstall is just more general in that it reinstalls in case one is installed already?

Yes. It'll wipe all state in the process (compared to doing an upgrade). I think it's preferable to document this command, because it works in more situations, and on the testnet we're not too worried about wiping existing state.

robin-kunzler (approved):
LGTM, thanks @kritzcreek !

PR #356, opened 2021-08-03
Show Application Principal ID during auth

nmattia:
## Description

This re-adds the functionality introduced by
https://github.com/dfinity/internet-identity/pull/325 but makes the
Principal ID less visible.

**NOTE**: I've added a tooltip. This is what I would have needed to understand the process the first time I authenticated with II as a user.

https://user-images.githubusercontent.com/6930756/128024280-79391604-f25d-460f-8c60-146de41fc8c3.mov

This works by calling `get_principal()` during the auth process and
thread the resulting principal ID to the authentication confirmation
screen.

## Motivation and Context

Currently it seems difficult for users to understand that every front end application they use on the IC with II will perceive them under a different principal. The first attempt for mitigating this is presenting the actual principal that will be used upon redirecting the user.

## How Has This Been Tested?

Manual front-end test with a local deployment.

Screenshot 2021-08-03 at 15 33 52
Screenshot 2021-08-03 at 15 34 19

## Types of changes

### DIFF
--- a/src/frontend/src/auth.ts
+++ b/src/frontend/src/auth.ts
@@ -88,7 +88,10 @@ async function handleAuthRequest(
   request: AuthRequest,
   hostname: FrontendHostname
 ): Promise<AuthResponse> {
-  if (!(await confirmRedirect(hostname))) {
+
+  const userPrincipal = await connection.getPrincipal(userNumber, hostname);
+
+  if (!(await confirmRedirect(hostname, userPrincipal.toString()))) {
     return {
       kind: "authorize-client-failure",
       text: `User did not grant access to ${hostname}.`,
--- a/src/frontend/src/auth.ts
+++ b/src/frontend/src/auth.ts
@@ -88,7 +88,6 @@ async function handleAuthRequest(
   request: AuthRequest,
   hostname: FrontendHostname
 ): Promise<AuthResponse> {
-
   const userPrincipal = await connection.getPrincipal(userNumber, hostname);
 
   if (!(await confirmRedirect(hostname, userPrincipal.toString()))) {
--- a/src/frontend/src/auth.ts
+++ b/src/frontend/src/auth.ts
@@ -88,7 +88,9 @@ async function handleAuthRequest(
   request: AuthRequest,
   hostname: FrontendHostname
 ): Promise<AuthResponse> {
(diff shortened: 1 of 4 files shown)


src/frontend/src/auth.ts
```diff
@@ -88,7 +88,9 @@ async function handleAuthRequest(
   request: AuthRequest,
   hostname: FrontendHostname
 ): Promise<AuthResponse> {
-  if (!(await confirmRedirect(hostname))) {
+  const userPrincipal = await connection.getPrincipal(userNumber, hostname);
```
kritzcreek, line 91: Do we need to cover this with a spinner? Query calls are fast, but depending on your connection you might still end up with an "unresponsive" app.

kritzcreek (changes requested):
Thanks, changes look great in general, and I love the tooltip.

Just have a couple of nits.

src/frontend/src/auth.ts
nmattia: Are you thinking the "astronaut" kind of spinner or a new kind?

Generally this makes sense though!

kritzcreek: Yeah the astronaut. I do the same when loading the data for the `manage` page with a query

src/frontend/src/flows/confirmRedirect.ts
```diff
@@ -1,15 +1,51 @@
 import { html, render } from "lit-html";
 import { FrontendHostname } from "../../generated/internet_identity_types";
 
-const pageContent = (hostName: string) => html`
+// We show a tooltip giving more information about the "application-specific"
+// nature of pseudonyms. This is the text that gets displayed when hovering.
+const textTooltip = `Internet Identity issues a unique "pseudonym" for each application requesting
+authentication. Since this pseudonym is unique for each application you
+authenticate for, it cannot be used to track you.
```
nmattia, line 8:
```suggestion
authenticate for, applications cannot use it to track you.
```

Suggestion from @Dfinity-Bjoern

src/frontend/src/auth.ts
nmattia: Done! PTAL.

kritzcreek (approved):

PR #357, opened 2021-08-03
Testnet App to assist in testing

kritzcreek:


demos/testnet_app/src/testnet_app_assets/src/index.js
```diff
@@ -0,0 +1,35 @@
+import { Actor, HttpAgent } from '@dfinity/agent';
+import { AuthClient } from "@dfinity/auth-client";
+import { idlFactory as testnet_app_idl, canisterId as testnet_app_id } from 'dfx-generated/testnet_app';
+
+
+const testnetII = "https://rdmx6-jaaaa-aaaaa-aaadq-cai.identity.dfinity.network/"
```
nmattia, line 6: maybe add a note in the README that this is hardcoded? do we want to be able to run this anywhere else?

_EDIT:_ I see this is an autofill. Worth adding a note about this here.

demos/testnet_app/webpack.config.js
```diff
@@ -0,0 +1,105 @@
...
+const TerserPlugin = require("terser-webpack-plugin");
+const dfxJson = require("./dfx.json");
+
+// List of all aliases for canisters. This creates the module alias for
+// the `import ... from "@dfinity/ic/canisters/xyz"` where xyz is the name of a
+// canister.
+const aliases = Object.entries(dfxJson.canisters).reduce(
+  (acc, [name, _value]) => {
+    // Get the network name, or `local` by default.
+    const networkName = process.env["DFX_NETWORK"] || "local";
```
nmattia, line 13: Should this be added to the README?

demos/testnet_app/webpack.config.js
```diff
@@ -0,0 +1,105 @@
...
+    // front-end development, add module loaders to the default
+    // webpack configuration. For example, if you are using React
+    // modules and CSS as described in the "Adding a stylesheet"
+    // tutorial, uncomment the following lines:
+    // module: {
+    //  rules: [
+    //    { test: /\.(ts|tsx|jsx)$/, loader: "ts-loader" },
+    //    { test: /\.css$/, use: ['style-loader','css-loader'] }
+    //  ]
+    // },
```
nmattia, line 82: I guess this was copy pasted, probably not worth duplicating here

demos/testnet_app/webpack.config.js
```diff
@@ -0,0 +1,105 @@
...
+      new webpack.ProvidePlugin({
+        Buffer: [require.resolve('buffer/'), 'Buffer'],
+        process: require.resolve('process/browser'),
+      }),
+    ],
+  };
+}
+
+// If you have additional webpack configurations you want to build
+//  as part of this configuration, add them to the section below.
```
nmattia, line 98: Same as above

nmattia (approved):

demos/testnet_app/webpack.config.js
kritzcreek: All of this is the dfx generated standard project. I left it here in the hope that this will make it easier to update for newer breaking dfx releases, since the migration instructions they put out will correspond closely to our setup.

demos/testnet_app/src/testnet_app_assets/src/index.js
kritzcreek: I moved the binding into the event handler. Hopefully that makes it easier to see its small scope.

I also added a comment describing what we're doing.

PR #360, opened 2021-08-06
Adds a note about flaky Selenium tests

kritzcreek:
Better late than never. Responding to https://github.com/dfinity/internet-identity/pull/349#discussion_r674807200


robin-kunzler (approved):
@kritzcreek : Great, thanks for adding the comment!

PR #361, opened 2021-08-06
Adapt documentation for Windows Hello authentication

robin-kunzler:
This PR adapts the how-to guide to explain the new Windows Hello support.


docs/modules/ic-identity-guide/pages/auth-how-to.adoc
```diff
@@ -94,6 +94,8 @@ From here, you can add and remove authentication methods, and set up additional
 
 The workflow for adding a device can vary depending on what devices you’ve already added to an Identity Anchor. For example, if you first authorized your computer to create the Identity Anchor, and you’d like to add your phone as a second authentication method, you must be able to authenticate your phone on the authorized computer. You must always be able to authorize the device you want to add by using a device that is already authorized.
 
+NOTE: If you start the add device flow on a Windows device that supports Windows Hello authentication, the browser first asks you to add Windows Hello as the new authentication method. If you have registered the device with Windows Hello already and would like to add e.g. a security key instead, you need to cancel the Windows Hello prompt. Then the browser lets you choose a different authentication method, such as a security key.
```
kritzcreek, line 97: I think it's actually the same behaviour for Touch/FaceId on OSX, but I don't have a device to test right now. Maybe you could?
If that's the case I think we should add those two to this paragraph

robin-kunzler: @kritzcreek : I tried the add device flow with my touch ID on OSX. 
I get this: 
Screen Shot 2021-08-09 at 12 10 49 PM

Then I choose "This Device" and then I can use touch id on this prompt: 
Screen Shot 2021-08-09 at 12 12 31 PM

Assuming other macs would behave in the same way, I guess the doc here does not extend to Mac and there would be nothing to add? Agreed? 

(Tested on Chrome, Firefox does not seem to let me add my mac with TouchId.)

kritzcreek: > I guess the doc here does not extend to Mac and there would be nothing to add? Agreed?

Agreed, thanks for checking!

kritzcreek (approved):
Thanks!

PR #369, opened 2021-08-17
Add FAQ page

nmattia:
## Description

This adds a FAQ page that is linked to from
1. The navbar
2. Anywhere where relevant (like the "confirm request" page to clarify what "application-specific" pseudonym means)

https://user-images.githubusercontent.com/6930756/129713084-d19e82cd-4a20-49fe-9389-1d2c71fd1bae.mov

## Notes

Our test framework (jest) made some of this more complicated than intended. It fails to run tests on a file that imports `lit-html`. Two workaround had to be put in place: first, the "questions" are split into a separate file (we're testing that anchors are unique). Second, the links are not inlined in the answers (but this turns out to look quite nice).

## Types of changes

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -8,5 +8,6 @@ export default {
   moduleNameMapper: {
     "dfx-generated/internet_identity": internet_identity,
   },
-  setupFiles: [`<rootDir>/src/frontend/test-setup.ts`]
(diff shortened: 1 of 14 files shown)


nmattia:
Note: this supersedes #link

PR #372, opened 2021-08-19
Add FAQ entry about pseudonyms

nmattia:
## Description
Add a FAQ entry about pseudonyms.

### DIFF
--- a/src/frontend/src/flows/faq/questions.ts
+++ b/src/frontend/src/flows/faq/questions.ts
@@ -62,7 +62,7 @@ export const questions = {
       "Does Internet Identity share my anchor or personal information with DApps?",
     anchor: "share-ii-anchor",
     answer:
-      'No. Internet Identity uses a different Principal ID (a "pseudonym") for each DApp that you authenticate for using Internet Identity. Since the pseudonyms Internet Identity generates for you are different for each DApp, and because each pseudonym is random, in principle the DApps cannot use pseudonyms to track you across DA',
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/faq/questions.ts
```diff
@@ -57,4 +57,12 @@ export const questions = {
       },
     ],
   },
+  shareIIAnchor: {
+    question:
+      "Does Internet Identity share my anchor or personal information with DApps?",
+    anchor: "share-ii-anchor",
+    answer:
+      'No. Internet Identity uses a different Principal ID (a "pseudonym") for each DApp that you authenticate for using Internet Identity. Since the pseudonyms Internet Identity generates for you are different for each DApp, and because each pseudonym is random, in principle the DApps cannot use pseudonyms to track you across DA',
```
Dfinity-Bjoern, line 65: Overall looks good, just a few comments:

- it should be "principal" not "principal id" 
- isn't it "authenticate to" instead of "for"?

And I think I would replace
> Since the pseudonyms Internet Identity generates for you are different for each DApp, and because each pseudonym is random, in principle the DApps cannot use pseudonyms to track you across DA
by
> Since the pseudonyms Internet Identity generates for you are different for each DApp, DApps cannot use them to track you outside of their realm.
or something similar. I don't think "random" is completely correct, since they are in the end computed deterministically. And I did not understand what you meant by "across DA".

nmattia: > And I did not understand what you meant by "across DA".

Looks like I had a brainfart and forgot to type the last few characters of "across DApps". Thanks for the feedback!

nmattia:
@Dfinity-Bjoern PTAL!

Dfinity-Bjoern (approved):
Thanks @nmattia!

PR #379, opened 2021-09-07
Extract views and flows from selenium E2E test file 

mminder:
## Description & Motivation
Separate views from test file to have a better overview over tests without scrolling a bunch first.

## How Has This Been Tested?
Tests have been executed locally and should now also run on CI.

## Types of changes

## Checklist:


nmattia (commented):
I'll admit that for a newbie like me, passing the driver around explicitly made things really clear; I wonder if the (class) indirection hurts clarity. Happy to go with it if you think it's a net improvement though! I'll let @kritzcreek be the judge

src/frontend/src/test-e2e/flows.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const FLOWS = {
+  registerNewIdentity: async (
+    deviceName: string,
+    driver: ThenableWebDriver
+  ): Promise<string> => {
+    const welcomeView = new WelcomeView(driver);
+    const registerView = new RegisterView(driver);
+    const singleDeviceWarningView = new SingleDeviceWarningView(driver);
+    const recoveryMethodSelector = new RecoveryMethodSelector(driver);
```
kritzcreek, line 18: I'd prefer initializing these right before usage instead

kritzcreek (changes requested):
Looks good in general 👍 I'd like to move the initialization of the view objects closer to their usage, like we're already doing it in the smaller tests.

src/frontend/src/test-e2e/flows.ts
mminder: done

kritzcreek (approved):

PR #383, opened 2021-09-10
replace selenium with webdriverio & extract util functions from selenium test 

mminder:
## Description
- Replace selenium-webdriver with webdriver.io as test script framework
- Extract setup and util functions into separate file to have better overview of actual tests
- Implement recovery use case test

## Motivation and Context
Webdriver.io has some advantages over selenium-webdriver, e.g.:
  * Improved error logging: If a test fails, the log contains the actual lines of test code that failed. This enables faster bugfixing and is just convenient.
  * Better documentation: Webdriver.io docs (https://webdriver.io/docs/gettingstarted) provide extensive documentation such as setup guides, good practices and configuration options. 
  * Flexibility for future testing needs: Webdriver.io can switch between webdriver-protocol and devTools protocol (chrome/puppeteer). It can also be used to write Appium tests, if in the future we decide to automate android/iOS ui tests.
  * API is a bit more concise (e.g. 'browser.$("#registerButton").click();'  vs. 'driver.findElement(By.id("registerButton")).click();' )

## How Has This Been Tested?
local and CI

## Types of changes

## Checklist:


src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -55,453 +41,368 @@ const DEMO_APP_URL = "http://localhost:8080/";
...
-    await alert1.accept();
+    const alertText1 = await browser.getAlertText();
     expect(alertText1).toBe(
       "This will remove your current device and you will be logged out"
     );
-    await driver.wait(until.alertIsPresent());
-    const alert2 = driver.switchTo().alert();
-    const alertText2 = await alert2.getText();
-    await alert2.accept();
+    await browser.acceptAlert();
```
nmattia, line 387: Shouldn't this be done after we checked the alter text?

src/frontend/src/test-e2e/util.ts
```diff
@@ -0,0 +1,209 @@
...
+  if (outer) {
+    webdriverProcess = await startWebdriver();
+  }
+
+  const browser = await remote({
+    capabilities: {
+      browserName: "chrome",
+      "goog:chromeOptions": {
+        args: [
+          "--no-sandbox",
```
nmattia, line 56: why disable the sandbox?

nmattia (commented):
Very cool! Out of my league here so I'll let @kritzcreek review as well.

src/frontend/src/test-e2e/util.ts
mminder: I copied that from somewhere but it doesn't seem necessary. Good point, I'll remove it :-)

src/frontend/src/test-e2e/selenium.test.ts
mminder: I kept the logic as it was before. There are two alerts in succession and it seems to sufficient to wait for the second one

kritzcreek (approved):
Looks great! Nice that the async usage has been made to look nicer than the docs would suggest.

PR #389, opened 2021-09-13
Prettify the FAQ page

nmattia:
Screenshot 2021-09-14 at 11 43 03

## Description

Add a pretty new FAQ page


src/frontend/src/flows/faq/index.ts
```diff
@@ -8,69 +9,91 @@ export { questions } from "./questions";
...
+      animation-duration: 600ms;
     }
   </style>
-  <div class="container" id="faq">
-    <h1>FAQ</h1>
-    <ul>
-      ${Object.values(questions).map((faq) => renderQuestion(faq))}
-    </ul>
+  <div class="faq__container">
+    <div>
```
kritzcreek, line 74: I don't think we need this extra div

src/frontend/src/flows/faq/index.ts
```diff
@@ -8,69 +9,91 @@ export { questions } from "./questions";
...
+      <h1 class="faq__title">FAQ</h1>
+      <div class="faq__title-underline"></div>
+      <ul class="faq__questions">
+        ${Object.values(questions).map((faq) => renderQuestion(faq))}
```
kritzcreek, line 78: I'd rather not rely on the ordering of `Object.values` here. I think we care about the order in which questions are displayed here, so we can put the most important ones at the top, right?

In that case we should probably use an Array here. Was the reason for using an object to make the access type safe? I think we can find other ways of achieving that if so.

src/frontend/src/flows/faq/index.ts
```diff
@@ -8,69 +9,91 @@ export { questions } from "./questions";
...
+      </ul>
+    </div>
   </div>
 `;
 
+const openAnchor = (): void => {
+  const hash = location.hash.substring(1);
+  if (hash) {
```
kritzcreek, line 86: The linter will complain that this isn't using a boolean condition.

src/frontend/src/flows/faq/index.ts
```diff
@@ -8,69 +9,91 @@ export { questions } from "./questions";
...
   <style>
+    html,
+    body,
+    main {
+      height: max-content;
+      overflow-x: hidden;
+    }
+
+    body {
+      position: relative;
```
kritzcreek, line 53: I don't think this is necessary. What is this doing?

src/frontend/src/flows/faq/index.ts
```diff
@@ -8,69 +9,91 @@ export { questions } from "./questions";
...
 }
 
 // The FAQ page
 const pageContent = html`
   <style>
+    html,
+    body,
+    main {
+      height: max-content;
+      overflow-x: hidden;
```
kritzcreek, line 49: I'm not sure what's going on here. Why are we hiding overflow?

src/frontend/src/flows/faq/index.ts
```diff
@@ -8,69 +9,91 @@ export { questions } from "./questions";
...
+    <details
+    id=${faq.anchor} >
+    <summary class="faq__question-summary">
+      ${faq.question}
+    <div class="faq__question-underline"></div>
+    </summary>
+    <div>
+      <p class="faq__answer">${faq.answer}</p>
+      <ul class="faq__answer-links">
+        ${Object.values(faq.links).map(
```
kritzcreek, line 24: Again I think we probably want to control the order in which these links appear.

src/frontend/src/flows/faq/index.ts
```diff
@@ -8,69 +9,91 @@ export { questions } from "./questions";
...
+  >
+    <details
+    id=${faq.anchor} >
+    <summary class="faq__question-summary">
+      ${faq.question}
+    <div class="faq__question-underline"></div>
+    </summary>
+    <div>
+      <p class="faq__answer">${faq.answer}</p>
+      <ul class="faq__answer-links">
```
kritzcreek, line 23: Let's only add this `<ul>` if we actually have links to display

kritzcreek (changes requested):
Nice! I added a first couple of comments.

src/frontend/src/flows/faq/index.ts
nmattia: agreed!

src/frontend/src/flows/faq/index.ts
nmattia: Agreed! Removing.

src/frontend/src/flows/faq/index.ts
nmattia: Agreed, will change.

src/frontend/src/flows/faq/index.ts
nmattia: Sure, will do!

nmattia: Done!

src/frontend/src/flows/faq/index.ts
nmattia: Done!

src/frontend/src/flows/faq/index.ts
nmattia: Fixed! and fixed the linter too: https://github.com/dfinity/internet-identity/pull/391

nmattia:
@kritzcreek PTAL!

src/frontend/src/flows/faq/index.ts
nmattia: Done! Added priority and then sorting alphabetically.

src/frontend/src/flows/faq/index.ts
```diff
@@ -38,39 +56,45 @@ const pageContent = html`
...
   </div>
 `;
 
+// Open the anchor with id="foo" if the page hash is "#foo"
+const openAnchor = (): void => {
+  const hash = location.hash.substring(1);
+
+  if (hash !== "") {
+    const details = document.getElementById(hash);
+    console.log(details);
```
kritzcreek, line 85:
```suggestion
```

Leftover debug log here

src/frontend/src/flows/faq/index.ts
```diff
@@ -38,39 +56,45 @@ const pageContent = html`
...
+
+    if (details) {
+      details.setAttribute("open", "");
+    }
```
kritzcreek, line 89:
```suggestion
    details?.setAttribute("open", "");
```

Is a little nicer here, WDYT?

src/frontend/src/flows/faq/index.ts
```diff
@@ -38,39 +56,45 @@ const pageContent = html`
...
+};
+
 export const faqView = (): void => {
+  document.title = "FAQ | Internet Identity";
   const container = document.getElementById("pageContent") as HTMLElement;
   render(pageContent, container);
+  openAnchor(); // needs to happen after DOM was rendered
 };
+
+faqView();
```
kritzcreek, line 100: Shouldn't this be called from `index.ts`? I'm not sure what it means to have the call at the top-level on the module

kritzcreek (changes requested):
Looks great! I spotted a few more places I'm unclear about

src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -400,5 +401,13 @@ test("Screenshots", async () => {
     const compatabilityNoticeView = new CompatabilityNoticeView(browser);
     await compatabilityNoticeView.waitForDisplay();
     await screenshot("16-compatibility-notice", browser);
+
+    // Compatibility notice page
```
kritzcreek, line 405:
```suggestion
    // FAQ page
```

src/frontend/src/flows/faq/index.ts
```diff
@@ -1,32 +1,50 @@
 import { html, render } from "lit-html";
+import "../../styles/main.css";
```
kritzcreek, line 2:
```suggestion
```

Since the faq page is rendered from `frontend/src/index.ts` the css should already be around.

src/frontend/src/flows/faq/index.ts
kritzcreek: @nmattia Once you've removed this call, I'm happy to approve.

nmattia: yup, just double checking what this is about!

nmattia: I know. I used a custom index at some point. Removing now.

kritzcreek (approved):
Looks great, thank you!

PR #391, opened 2021-09-16
Use strict boolean checks for strings (and numbers)

nmattia:
## Description

This sets the `strict-boolean-expressions` options for `allowString` and
`allowNumber` to `false`, making sure all empty strings and zeros are
checked explicitly. Some occurrences of lax checks were updated.

### DIFF
--- a/src/frontend/src/utils/userIntent.ts
+++ b/src/frontend/src/utils/userIntent.ts
@@ -8,7 +8,7 @@ export type AddDeviceIntent = { kind: "addDevice" };
 export const intentFromUrl = (url: URL): UserIntent => {
   if (url.hash == "#authorize") {
     return { kind: "auth" };
-  } else if (url.hash?.split("device=")[1]) {
+  } else if (url.hash?.split("device=")[1] !== undefined) {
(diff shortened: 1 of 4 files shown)


src/frontend/src/utils/userIntent.ts
```diff
@@ -8,7 +8,7 @@ export type AddDeviceIntent = { kind: "addDevice" };
 export const intentFromUrl = (url: URL): UserIntent => {
   if (url.hash == "#authorize") {
     return { kind: "auth" };
-  } else if (url.hash?.split("device=")[1]) {
+  } else if (url.hash !== ""?.split("device=")[1]) {
```
kritzcreek, line 11: This looks weird. The `?` here is not the ternary operator, but a part of the null-safe access operator `?.`.

nmattia: Oh damn you're right!

nmattia: PTAL!

kritzcreek (approved):

PR #393, opened 2021-09-16
decrese selenium/webdriver loglevel

mminder:
CI log is way to long, WARN should be sufficient

## Types of changes

## Checklist:


nmattia (commented):
A case can definitely be made for having INFO logs when encountering a failure. Can we log everything to a file and add that as an artefact? And then only have WARN in the log output on CI?

mminder:
> A case can definitely be made for having INFO logs when encountering a failure. Can we log everything to a file and add that as an artefact? And then only have WARN in the log output on CI?

Excellent idea. I just pushed a commit to try it out

mminder:
@nmattia Looks good. I also added the screenshots as build artifact and also I'm taking another screenshot on error. Could be useful, what do you think?

src/frontend/src/test-e2e/util.ts
```diff
@@ -57,6 +57,9 @@ export async function runInBrowserCommon(
     },
     automationProtocol: "webdriver",
     path: "/wd/hub",
+    logLevel: "info",
+    // outputDir pipes all webdriver log output into ./wdio.log
```
nmattia, line 61:
```suggestion
    // outputDir pipes all webdriver log output into ./wdio.log
    // stdout only contains errors on test failures
```

nmattia (approved):
Looking great! One tiny suggestion.

src/frontend/src/test-e2e/util.ts
mminder: oh no, I just missed that comment! Will add it to a future PR

PR #395, opened 2021-09-16
Automatically increase screenshot count

nmattia:
This automatizes the screenshot counting ("01-foo", "02-bar", etc).
Screenshots were getting mixed up (because some screenshots were added
to already existing, old "stories") and screenshot numbers were being
duplicated (because everyone doesn't go through each screenshot to
figure out what the smallest new available number is).

### DIFF
--- a/src/frontend/src/test-e2e/selenium.test.ts
+++ b/src/frontend/src/test-e2e/selenium.test.ts
@@ -205,7 +205,6 @@ test("Recover access, after registration", async () => {
 
 test("Screenshots", async () => {
   await runInBrowser(async (browser: WebdriverIO.Browser) => {
-
     const screenshots = new Screenshots("screenshots");
 
     await addVirtualAuthenticator(browser);
--- a/src/frontend/src/test-e2e/selenium.test.ts
+++ b/src/frontend/src/test-e2e/selenium.test.ts
@@ -16,11 +16,11 @@ import {
 } from "./views";
 import { FLOWS } from "./flows";
 import {
+  Screenshots,
   addVirtualAuthenticator,
   removeVirtualAuthenticator,
   runInBrowser,
   runInNestedBrowser,
-  screenshot,
   switchToPopup,
   waitForFonts,
   waitToClose,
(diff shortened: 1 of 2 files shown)


src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -41,6 +41,15 @@ const DEMO_APP_URL = "http://localhost:8080/";
 const DEVICE_NAME1 = "Virtual WebAuthn device";
 const DEVICE_NAME2 = "Other WebAuthn device";
 
+const screenshots = new Screenshots("screenshots");
+
+async function screenshot(
```
mminder, line 46: do we really need this function? we could just use 'screenshots.take' directly

src/frontend/src/test-e2e/util.ts
```diff
@@ -155,11 +155,24 @@ export async function removeVirtualAuthenticator(
...
-): Promise<void> {
-  await browser.saveScreenshot(`screenshots/${name}.png`);
+// 'Screenshots' objects are used to make sure all screenshots end up in the
+// same directory, each with a different (increasing) number prefixed in the
+// filename.
+export class Screenshots {
+  private directory: string;
+  private count: number;
+
+  constructor(dir: string) {
```
mminder, line 165: you could use "constructor(private directory:string)", then you don't need to specify the member on line 162 and assign it on 166

src/frontend/src/test-e2e/util.ts
```diff
@@ -155,11 +155,24 @@ export async function removeVirtualAuthenticator(
...
-  name: string,
-  browser: WebdriverIO.Browser
-): Promise<void> {
-  await browser.saveScreenshot(`screenshots/${name}.png`);
+// 'Screenshots' objects are used to make sure all screenshots end up in the
+// same directory, each with a different (increasing) number prefixed in the
+// filename.
+export class Screenshots {
+  private directory: string;
+  private count: number;
```
mminder, line 163: this could be initialized like "private count: number = 0;", then you don't need to do anything in the constructor

src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -41,6 +41,15 @@ const DEMO_APP_URL = "http://localhost:8080/";
 const DEVICE_NAME1 = "Virtual WebAuthn device";
 const DEVICE_NAME2 = "Other WebAuthn device";
 
+const screenshots = new Screenshots("screenshots");
```
mminder, line 44: We could construct this inside the screenshot test, because the numbering is tied to that test. Maybe in the future there will be an additional test that also takes screenshots in a different folder with its own ordering

mminder (changes requested):
cool idea! I just have some ideas we could discuss :-)

src/frontend/src/test-e2e/selenium.test.ts
nmattia: I tried keeping the diff size low :)

src/frontend/src/test-e2e/selenium.test.ts
nmattia: Ah yes, makes sense!

mminder (approved):

PR #398, opened 2021-09-17
start selenium before tests and increase screenshot test waits

mminder:
## Description
- start selenium server outside of test script to avoid killing and restarting it for each "test" method
- add some temporary explicit waits in screenshot test to increase stability. 

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,11 @@
     "generate:types": "didc bind ./src/internet_identity/internet_identity.did -t ts > src/frontend/generated/internet_identity_types.d.ts",
     "generate:js": "didc bind ./src/internet_identity/internet_identity.did -t js > src/frontend/generated/internet_identity_idl.js",
     "build": "NODE_ENV=production webpack",
-    "test": "jest --verbose",
+    "test": "npm run start-selenium-server > /dev/null 2>&1 & jest --verbose",
     "lint": "eslint --cache --cache-location node_modules/.cache/eslint 'src/frontend/**/*.ts*'",
     "format": "prettier --write src/frontend",
-    "install-webdrivers" : "selenium-standalone install"
+    "install-webdrivers": "selenium-standalone install",
+    "start-selenium-server": "selenium-standalone start"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.14.1",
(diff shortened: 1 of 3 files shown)


src/frontend/src/test-e2e/util.ts
```diff
@@ -78,12 +47,6 @@ export async function runInBrowserCommon(
       "An error occurred during e2e test execution. Logs can be found in the wdio.log file and an additional error screenshot was saved under screenshots/error. On Github Actions you can find the log and screenshots under 'Artifacts'."
     );
     throw e;
-  } finally {
-    if (outer) {
-      // only close outer session
-      await browser.deleteSession();
```
mminder, line 84: it looks like the "deleteSession()" can lead to some weird timeout issues where following tests overtake the failed one. I'd like to remove it for now. I think Selenium Server can handle a couple of old dangling sessions

src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -291,8 +291,10 @@ test("Screenshots", async () => {
       await welcomeBackView.waitForDisplay();
       await welcomeBackView.fixup();
       await screenshot("08-new-device-login", browser);
+      await browser.pause(3000);
```
mminder, line 294: The flaky issue occurs, because we click to early. So it's no use to wait longer in " await singleDeviceWarningView.waitForDisplay();", we have to wait a bit before clicking the link.

package.json
```diff
@@ -8,10 +8,11 @@
     "generate:types": "didc bind ./src/internet_identity/internet_identity.did -t ts > src/frontend/generated/internet_identity_types.d.ts",
     "generate:js": "didc bind ./src/internet_identity/internet_identity.did -t js > src/frontend/generated/internet_identity_idl.js",
     "build": "NODE_ENV=production webpack",
-    "test": "jest --verbose",
+    "test": "npm run start-selenium-server > /dev/null 2>&1 & jest --verbose",
```
nmattia, line 11: I'm uneasy about the server outliving the tests

src/frontend/src/test-e2e/selenium.test.ts
nmattia: In my opinion pauses in tests are rarely a solution. Why do we click too early? Does that mean the user may also click too early?

mminder:
not needed in this form

PR #401, opened 2021-09-17
Restyle about page

nmattia:
This restyles the about page, very much in the style of the new FAQ
page.

Screenshot 2021-09-17 at 15 03 11

https://user-images.githubusercontent.com/6930756/133786964-26fd66b5-f452-48dd-ae64-eea9592a4ff6.MP4


src/frontend/src/components/compatibilityChart.ts
```diff
@@ -1,22 +1,29 @@
 import { html } from "lit-html";
 
-export const compatibilityChart = html`<p>
-    Browser support for WebAuthentication is constantly evolving. Your preferred
-    browser may not support WebAuthentication, or may only support it using a
-    security key. If you run into issues, please try again with one of our
-    recommended browsers.
-  </p>
+export const compatibilityData = {
```
nmattia, line 3: Instead of raw HTML this is now "data" that can be used and styled by the page requesting it.

src/frontend/src/styles/main.css
```diff
@@ -405,5 +394,120 @@ hr {
   background-color: rgba(243, 244, 246, 1);
   max-width: 42rem;
   height: 100%;
-  position: fixed;
+}
```
nmattia, line 397: Here I removed `position: fixed`, which I should never have merged to start with. The content grows beyond the parent element now on the FAQ and About page, but it's much better that not being able to scroll. Will need to fix this separately.

kritzcreek (approved):

PR #403, opened 2021-09-19
feat: use 64 bit stable memory API

roman-kashitsyn:
This change updates the Storage implementation to take advantage of 64
bit memory API that was recently introduced by the Execution team.

This should buy us some time to plan multi-canister scaling as we can
now afford storing twice as many users in a single canister.


nomeata:
Isn't that experimental? Or just the concrete API, but not the feature itself?

roman-kashitsyn:
> Isn't that experimental? Or just the concrete API, but not the feature itself?

It's spec-ed out and fully implemented in replica, I believe Björn has some issues with releasing the new version of the spec.
It's not implemented in ic-ref yet though...

nomeata:
Ok, I think on the forums [it was said that it's experimental](https://forum.dfinity.org/t/increased-canister-storage/6148/17?u=nomeata) and one shouldn't use it for critical applications. But maybe internally different rules apply, and the II developers will likely get a heads up if this interface gets removed again.

dsarlis:
> Ok, I think on the forums [it was said that it's experimental](https://forum.dfinity.org/t/increased-canister-storage/6148/17?u=nomeata) and one shouldn't use it for critical applications. But maybe internally different rules apply, and the II developers will likely get a heads up if this interface gets removed again.

@nomeata I think the intent of that comment was to simply give a heads up that this is a new api and as such there are always risks involved (just like with any new software). Not really meant to prevent someone from using the API. On the contrary, we'd like to see usages of it so we can solicit feedback.

Also, re "getting a heads up if this interface gets removed", we would never remove an interface without making sure we don't break existing use cases. We've started applying some basic tracking (with a good sense) of which apis canisters are using and will make sure no uses exist before we deprecate an api. And that goes for any api. (E.g. we really want to drop `call_simple` but there are canisters in a subnet using it, so we need to work with them to migrate to the builder api for calls).

nomeata:
Thanks, that’s good insight, also for other people who might be considering using “experimental” interfaces

backend-tests/backend-tests.hs
```diff
@@ -759,7 +759,11 @@ tests wasm_file = testGroup "Tests" $ upgradeGroups $
 
     when should_upgrade $ doUpgrade cid
     s <- queryII cid dummyUserId #stats ()
-    lift $ s .! #assigned_user_number_range @?= (100, 103)
+    -- The storage updates the upper bound on upgrade if it doesn't use the
+    -- full capacity. This is a hack that have to go away when we start using
```
nmattia, line 763:
```suggestion
    -- full capacity. This is a hack that has to go away when we start using
```

roman-kashitsyn:
> LGTM although I don't understand what's happening in `backend-tests`, didn't know they existed until now.

I modified backend tests because the canister auto-expands the range of users assigned to it on upgrade so that we could use more storage.

PR #405, opened 2021-09-20
Document and demystify the management page

nmattia:
This extracts some of the style and error code and adds some comments
about what's happening in the management page.

### DIFF
--- a/src/frontend/src/flows/manage.ts
+++ b/src/frontend/src/flows/manage.ts
@@ -268,7 +268,7 @@ const init = async (
     "#addAdditionalDevice"
   ) as HTMLButtonElement;
 
-  addAdditionalDeviceButton.onclick = async () =>
+  addAdditionalDeviceButton.onclick = () =>
     addAdditionalDevice(userNumber, connection, devices);
 
   // Add recovery
--- a/src/frontend/src/flows/manage.ts
+++ b/src/frontend/src/flows/manage.ts
@@ -17,86 +17,138 @@ import { WebAuthnIdentity } from "@dfinity/identity";
 import { setupRecovery } from "./recovery/setupRecovery";
 import { hasOwnProperty } from "../utils/utils";
 
-const pageContent = (userNumber: bigint, devices: DeviceData[]) => html`<style>
-    .nagBox {
-      display: flex;
-      align-items: center;
-      gap: 1rem;
-      padding: 1rem;
-      margin-bottom: 2rem;
-      box-sizing: border-box;
-      border-style: double;
-      border-width: 2px;
-      border-radius: 4px;
-      border-image-slice: 1;
-      outline: none;
-      border-image-source: linear-gradient(
-        270.05deg,
-        #29abe2 10.78%,
-        #522785 22.2%,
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/manage.ts
```diff
@@ -17,86 +17,138 @@ import { WebAuthnIdentity } from "@dfinity/identity";
...
-      font-weight: 600;
-    }
-    .labelAction::before {
-      content: "+";
-      margin-right: 3px;
-      color: #387ff7;
-    }
-  </style>
+const displayFailedToAddTheDevice = (error: Error) => {
+  displayError({
```
kritzcreek, line 40: You need to either `return` the Promise or remove the surrounding block here, or you can't `await` the user clicking "Back to manage"

src/frontend/src/flows/manage.ts
```diff
@@ -17,86 +17,138 @@ import { WebAuthnIdentity } from "@dfinity/identity";
...
+    title: "Failed to add the new device",
+    message:
+      "We failed to add the new device to this Identity Anchor. Please try again",
+    detail: error.message,
+    primaryButton: "Back to manage",
+  });
+};
+
+const displayFailedToListDevices = (error: Error) => {
+  displayError({
```
kritzcreek, line 50: Same as above

src/frontend/src/flows/manage.ts
```diff
@@ -150,94 +203,82 @@ export const renderManage = async (
...
+    await displayFailedToAddNewDevice(error);
+    return renderManage(userNumber, connection);
+  }
+  const deviceName = await pickDeviceAlias();
+  if (deviceName === null) {
+    return renderManage(userNumber, connection);
+  }
+  // TODO check whether newDevice is already registered,
+  // or better, pass existing devices to `.create` so that they
+  // cannot be added again
```
kritzcreek, line 237: We do pass the existing devices to `.create` in line 225, so I don't think there's anything we can do here?

src/frontend/src/flows/manage.ts
```diff
@@ -150,94 +203,82 @@ export const renderManage = async (
...
-    } catch (error) {
-      await displayError({
-        title: "Failed to add the new device",
-        message:
-          "We failed to add the new device to this Identity Anchor. Please try again",
-        detail: error.message,
-        primaryButton: "Back to manage",
-      });
-    }
+  addAdditionalDeviceButton.onclick = async () =>
```
kritzcreek, line 271:
```suggestion
  addAdditionalDeviceButton.onclick = () =>
```

No need for this to be async, since we're not awaiting in here.

kritzcreek (changes requested):
Looks good in general, just a couple nits

src/frontend/src/flows/manage.ts
nmattia: I didn't write that :)

kritzcreek: In that case, even better!

```suggestion
```

kritzcreek (approved):

PR #408, opened 2021-09-21
start selenium server before tests

mminder:
## Description
Selenium server can be started before all tests and stopped after all tests. The previous logic (re-starting it for every test case) was complicated and could sometimes lead to orphaned selenium server processes.

## Types of changes

## Checklist:

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -8,5 +8,5 @@ export default {
   moduleNameMapper: {
     "dfx-generated/internet_identity": internet_identity,
   },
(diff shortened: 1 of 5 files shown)


nmattia (approved):
LGTM! Now that all tests share the same server instance, are there any risks of having a test pollute the environment?

mminder:
@nmattia The server instance opens the relevant browser for each test and then passes the commands through, I don't think there is any state in there that could influence test cases

PR #421, opened 2021-10-05
Fix webpack.config.js for testnet_app

nmattia:
The `dfx` executable stores the canister IDs in their sanitized
(alphanum) form, so we need to index the `canisters` with the sanitized
IDs too.


kritzcreek (approved):
Ideally we'd get this into the default project dfx generates, so we don't override the fix the next time we regenerate the webpack config

nmattia:
Yup, will port it here: https://github.com/dfinity/sdk/pull/1841

PR #425, opened 2021-10-08
Re-enable whoami check without ic-ref

nmattia:
This is a horrible hack. The 'whoami' canister does not respond when running with the emulator, meaning we had to disable one check in our test suite. It is taking longer than expected to figure out why, so for the time being we make sure to at least run the full test suite when _not_ using the emulator.

### DIFF
--- a/.github/workflows/selenium.yml
+++ b/.github/workflows/selenium.yml
@@ -66,8 +66,19 @@ jobs:
 
       - run: rm -v -f screenshots/*.png
       - run: npm test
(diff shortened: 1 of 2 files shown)


roman-kashitsyn:
This is indeed very weird, we should create an issue to get rid of this hack.

> but first I need to get ic-ref, which is proving surprisingly hard

Isn't it shipped with each version of DFX? You can find it in `~/.cache/dfinity/versions/$DFX_VERSION/ic-ref`...

PR #431, opened 2021-10-18
Add warning when seed phrase used for wrong anchor

nmattia:
This adds a big red warning when the first digits of the seed phrase do
not match the anchor. That number is stripped anyway and used only to
match the anchor to the seed phrase.

https://user-images.githubusercontent.com/6930756/137704954-7121d3ad-6c83-4a9d-b8d8-5c1a15974885.mov

### DIFF
--- a/src/frontend/src/flows/recovery/inputSeedPhrase.ts
+++ b/src/frontend/src/flows/recovery/inputSeedPhrase.ts
@@ -51,14 +51,11 @@ const init = (userNumber: bigint): Promise<string | null> =>
       "warningBadAnchor"
     ) as HTMLParagraphElement;
     inputSeedPhraseInput.oninput = () => {
-      const lun = leadingUserNumber(inputSeedPhraseInput.value);
-      console.log(leadingUserNumber(inputSeedPhraseInput.value));
+      const lun = leadingUserNumber(inputSeedPhraseInput.value.trim());
       if (lun !== null && lun !== userNumber) {
         warningBadAnchor.innerHTML = `<strong>warning</strong><br/>
                     it looks like you are using a seed phrase for anchor <strong>${lun}</strong> but the anchor you are trying to recover is <strong>${userNumber}</strong>
             `;
-
-        console.log("not ok...");
       } else {
         warningBadAnchor.innerHTML = "";
       }


src/frontend/src/flows/recovery/inputSeedPhrase.ts
```diff
@@ -13,27 +13,56 @@ const pageContent = () => html`
...
+    ) as HTMLParagraphElement;
+    inputSeedPhraseInput.oninput = () => {
+      const lun = leadingUserNumber(inputSeedPhraseInput.value);
+      console.log(leadingUserNumber(inputSeedPhraseInput.value));
+      if (lun !== null && lun !== userNumber) {
+        warningBadAnchor.innerHTML = `<strong>warning</strong><br/>
+                    it looks like you are using a seed phrase for anchor <strong>${lun}</strong> but the anchor you are trying to recover is <strong>${userNumber}</strong>
+            `;
+
+        console.log("not ok...");
```
bitdivine, line 61: Do you want to keep this `console.log`?  And if you keep it do you want `console.warn()` or similar instead?  If kept the text should probably say what is not ok. :-)

src/frontend/src/flows/recovery/inputSeedPhrase.ts
```diff
@@ -69,3 +98,12 @@ const dropLeadingUserNumber = (s: string): string => {
     return s;
   }
 };
+
+const leadingUserNumber = (s: string): bigint | null => {
+  const i = s.indexOf(" ");
```
bitdivine, line 103: Do you want to trim() first, so that leading whitespace does not matter?  Something like this maybe?
```
let [num_str] = s.trim().split(" ", 1) // get first word.
if ("" !== num_str) {
  return parseUserNumber(num_str);
...
```
I know I sometimes copy and paste passphrases and catch an additional whitespace at the beginning and it can be really confusing if the password is "incorrect" but I can see no problem.  With passwords one is rather more attuned to every character mattering, so is more careful.

nmattia:
> Is the identity required to be at the beginning of the seed phrase? I primarily use cli tools but e.g. my keysmith passphrase does not start with my account ID.

Not at all. The javascript code actually strips it, maybe your CLI tools do the same?

src/frontend/src/flows/recovery/inputSeedPhrase.ts
nmattia: Ok I've addressed this a bit differently. The code here stays the same (mirroring `dropLeadingUserNumber` above) but I trim the value before calling `leadingUserNumber`. Thanks!

Issue #381, opened 2021-09-08
Error Compiling on Mac M1 - 'could not compile wabt-sys'


gschian0:
**I run this command on macbook pro M1**

`II_ENV=development dfx deploy --no-wallet --argument '(null)'`

**Hangs up on  this for a long while**

`Building [=======================>   ] 68/75: wabt-sys(build), binaryen-sys(build)  `

**error message produced after hanging up**

`error: failed to add native library /var/folders/z8/06mhg00s4lzd_dzykt3q0s6m0000gn/T/cargo-installNkJZeH/release/build/wabt-sys-0c5d3f90b9057634/out/build/libwabt.a: file too small to be an archive

error: aborting due to previous error

error: could not compile `wabt-sys`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
    Building [========================>  ] 71/75: binaryen-sys(build)        `

gschian0:
**Following these instructions I was able to compile and deploy**

[https://github.com/dfinity/internet-identity/blob/713a8d8ec27fd0474f64bceb3571d8022cdd1455/src/internet_identity/build.sh#L14](url)

**however now when I attempt to create an ID I get this error message**
`We couldn't reach Internet Identity
We failed to call the Internet Identity service, please try again.
Error details
Server returned an error:
  Code: 404 (Not Found)
  Body: Requested canister has no wasm module
Try again`

gschian0:
**when executing command**

`dfx canister call internet_identity init_salt`

`dfx canister call internet_identity init_salt
The replica returned an HTTP Error: Http Error: status 404 Not Found, content type "", content: Requested canister has no wasm module`

gschian0:
**Now... I've made some progress and I'm getting this error.**

`Call was rejected:
  Request ID: 5c0f48ff860be6ef8c0461a243c10150af51396075ce816d79bfb337dcd9b408
  Reject code: 
  Reject text: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai has no update method 'register'`

nmattia:
Hi, can you check the latest `main` and see if the issue persists? I've removed the step that installed `ic-cdk-optimizer` which seemed to have been causing the issue. You will need to install `ic-cdk-optimizer`:

``` bash
$ cargo install ic-cdk-optimizer --version 0.3.1
```

wewei:
Verified on MBP-M1 with macOS 12.0.1, deployed successfully!
[...]

PR #434, opened 2021-10-23
fix: build on Apple M1 chip

plitzenberger:
## Description

when running `II_ENV=development dfx deploy --no-wallet --argument '(null)'` on apple M1 chip the build fails with:

Check #381 

```
error: failed to add native library /var/folders/z8/06mhg00s4lzd_dzykt3q0s6m0000gn/T/cargo-installNkJZeH/release/build/wabt-sys-0c5d3f90b9057634/out/build/libwabt.a: file too small to be an archive

error: aborting due to previous error

error: could not compile wabt-sys
```

## Motivation and Context

This PR is primarily intended as an opening to discussions on how to fix the above build issue in general. By applying this change, the build script installs `ic-cdk-optimizer` to your home dir `~/.cargo/bin/ic-cdk-optimizer` instead of the local target folder. For whatever reason this works. I'm not a rust expert.

## How Has This Been Tested?

It works on my machine :)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/internet_identity/build.sh
+++ b/src/internet_identity/build.sh
@@ -11,13 +11,13 @@ TARGET="wasm32-unknown-unknown"
 cargo build --manifest-path "$II_DIR/Cargo.toml" --target $TARGET --release -j1
 
 # keep version in sync with Dockerfile
-cargo install ic-cdk-optimizer --version 0.3.1 --root "$II_DIR"/../../target
+cargo install ic-cdk-optimizer --version 0.3.1
 STATUS=$?
 
 if [ "$STATUS" -eq "0" ]; then
-      "$II_DIR"/../../target/bin/ic-cdk-optimizer \
-      "$II_DIR/../../target/$TARGET/release/internet_identity.wasm" \
-      -o "$II_DIR/../../target/$TARGET/release/internet_identity.wasm"
+  ic-cdk-optimizer \
+    ./target/$TARGET/release/internet_identity.wasm \
+    -o ./target/$TARGET/release/internet_identity.wasm
 
   true
 else


bitdivine:
The `II_DIR` refers to `src/internet_identity`, so in theory the path construction could make sense, however on line 6 of `build.sh` we have `npm run build` which assumes that the script is being run from the root of the git dir.  Options:
* If we want to make the script strongly location independent we can start the script with `cd "$(dirname "$(realpath "$0")")/../.."` but note that the original realpath spec was broken, so most Linux distributions dropped it and used `readlink -f ...` instead.  Recent Ubuntu has a semi-sane realpath installed by default (>20.04 if I recall correctly), but earlier versions of Ubuntu and many other distros will reject the code at this point.
* We can assume that there are no soft links in place and do `cd "$(dirname "$0")/../.."`.  We don't have soft links so this should work out of the box for everyone who doesn't customize their setup.
* Or we can just assume that we are running from the git root.

Mert75:
I'm also on the M1 chip and ran into the same problem.

Running ```cargo install ic-cdk-optimizer --version 0.3.1``` followed by the solution in the PR solved it for me 👍

krpeacock:
@nmattia @bitdivine bumping this

nmattia:
@krpeacock unfortunately @bitdivine and very snowed under at the moment; hopefully in two or three weeks we'll have a few spare cycles. In the meantime feel free to have a look! This is definitely on our radar.

wewei:
Hi, any updates on this one?
The forum link is https://forum.dfinity.org/t/setup-internet-identity-locally-failed-macbook-m1/7684

letmejustputthishere:
I can confirm this solves the problem on a M1, thanks @plitzenberger 

you really have to run `cargo install ic-cdk-optimizer --version 0.3.1` first though.

PR #446, opened 2021-11-01
Pause for 0.5s before taking screenshot

nmattia:
We have some flakiness in screenshots that can't really be explained.
The best hypothesis thus far is that the screenshot is taken after the
first rendering pass, but before it has settled on exactly how to render
everything, leading to a few (8) pixels having different colors. If this
doesn't solve the issue, then we'll need to hack and change how screenshots are
committed, and only commit if the image was modified beyond some
threshold.

Here are the culprits:

https://user-images.githubusercontent.com/6930756/139709344-477ddd6f-d196-4f52-ae28-a6fcaef24bbf.mov


bitdivine (commented):
OK by me.  I know that arguably one should use an event look.  I see that it is possible to wait for the page to load with something like:

bitdivine (approved):

PR #454, opened 2021-11-18
Update candid in backend-tests

nmattia:
This updates the candid definition used by the backend-tests, which was a bit old. It also updates `ic-hs` and `haskell-candid` to include the following fixes:

* https://github.com/nomeata/haskell-candid/pull/17
* https://github.com/dfinity/ic-hs/pull/58

This also includes a fix for stabilizing the value of `webauthPK` which is checked in tests:
* https://github.com/dfinity/ic-hs/issues/59

Finally the backend-tests workflow now uses `cabal build/run` instead of `cabal install`, because `cabal install` moves the code to a different location, which makes breaks the relative path to `internet_identity.did` used to quote the file.


nmattia:
@nomeata any red flags?

nomeata (approved):

backend-tests/backend-tests.hs
```diff
@@ -71,107 +71,7 @@ import Prometheus hiding (Timestamp)
...
-
-  prepare_delegation : (UserNumber, FrontendHostname, SessionKey, opt nat64) -> (UserKey, Timestamp);
-  get_delegation: (UserNumber, FrontendHostname, SessionKey, Timestamp) -> (GetDelegationResponse) query;
-
-  http_request: (request: HttpRequest) -> (HttpResponse) query;
-}
-  |]
+type IIInterface m = [Candid.candidFile|../src/internet_identity/internet_identity.did|]
 
 -- Names for some of these types. Unfortunately requires copying
```
bitdivine, line 76: I wonder why?  This is not the topic of this PR but if you happen to know please share.

bitdivine (approved):

backend-tests/backend-tests.hs
nomeata: Haskell-candid allows importing .did files wholesome to get thr the service type, but not that type definitions therein. Maybe adding that would be a nice new feature for a leisurely hacking afternoon

nomeata: working on it : https://github.com/nomeata/haskell-candid/pull/18

PR #463, opened 2021-11-25
Replace certified_map with library version

frederikrothenberger:
Remove the code in src/certified_map and src/hashtree and use the library ic-certified-map instead. The library contains a more recent version of that code, which in particular removes uses of unsafe code. Currently there is no official release of that version therefore commit 129d1a8c595525c6f4c54e73bafa779143e568ce is referenced directly as a stopgap until one is available.


ggreif:
dfinity/ic uses a GitHub commit, not "0.2.0". I have suggested a new release here: https://github.com/dfinity/cdk-rs/issues/176, feel free to thumbs-up it.

src/internet_identity/src/main.rs
```diff
@@ -782,12 +781,12 @@ fn get_signature(
...
 
-    let tree = hashtree::fork(
-        HashTree::Pruned(hashtree::labeled_hash(
+    let tree = ic_certified_map::fork(
+        HashTree::Pruned(ic_certified_map::labeled_hash(
             LABEL_ASSETS,
             &asset_hashes.root_hash(),
         )),
-        hashtree::labeled(&LABEL_SIG[..], witness),
+        labeled(&LABEL_SIG[..], witness),
```
robin-kunzler, line 789: nit: is there a reason to import `labeled` directly instead of using `ic_certified_map::labeled` as in the other places?

src/internet_identity/Cargo.toml
```diff
@@ -5,9 +5,8 @@ edition = "2018"
 
 [dependencies]
 base64 = "0.13.0"
-certified_map = { path = "../certified_map" }
+ic-certified-map = "0.2.0"
```
robin-kunzler, line 8: This version is the newest on crates.io, but it would be great to move to the one where the unsafe code is removed. Are you planning to do this before merging?

robin-kunzler: nit: we usually sort the dependencies in alphabetical order. Could you do that too?

robin-kunzler (commented):
Thanks @frederikrothenberger, looks very good. I have a few comments / questions, see below.  

Also, would it make sense to be a bit more explicit in the PR descriptions? E.g. change to: 

> Remove the code in src/certified_map and src/hashtree and use the library `ic-certified-map` from crates.io instead. The library contains a more recent version of that code, which in particular removes uses of unsafe code.

roman-kashitsyn:
The change looks good to me.
It's better to use the latest version of ic-certified-map (either by specifying the commit hash or releasing a new version), but it's not super-critical.

selenium-standalone.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   drivers: {
     chrome: {
       // This version needs to match the chrome version on GitHub Actions
-      version: '94.0.4606.41',
+      version: '96.0.4664.45',
```
nmattia, line 5: probably best to rebase instead of merging to get rid of this

src/internet_identity/Cargo.toml
nmattia: out of curiosity, which one is the one without the unsafe code? is that on GitHub but not yet released?

re. alphabetical order: would be worth checking this and doing it automatically! wdyt?

src/internet_identity/src/main.rs
```diff
@@ -826,9 +825,9 @@ fn make_asset_certificate_header(
         trap("data certificate is only available in query calls");
     });
     let witness = asset_hashes.witness(asset_name.as_bytes());
-    let tree = hashtree::fork(
-        hashtree::labeled(LABEL_ASSETS, witness),
-        HashTree::Pruned(hashtree::labeled_hash(LABEL_SIG, &sigs.root_hash())),
+    let tree = ic_certified_map::fork(
+        labeled(LABEL_ASSETS, witness),
```
nmattia, line 829: same here 

> nit: is there a reason to import labeled directly instead of using ic_certified_map::labeled as in the other places?

src/internet_identity/src/signature_map.rs
```diff
@@ -1,16 +1,17 @@
 //! Maintains user signatures and expirations.
-use certified_map::{AsHashTree, RbTree};
-use hashtree::{leaf_hash, Hash, HashTree};
+use ic_certified_map::{leaf_hash, AsHashTree, Hash, HashTree, RbTree};
+use std::borrow::Cow;
 use std::collections::BinaryHeap;
 
+#[derive(Default)]
```
nmattia, line 6: what is this for?

nmattia (commented):
For my own understanding, can someone clarify the relationship between `ic_certified_map` and `hashtree`? and did they start their lives inside this repo?

selenium-standalone.config.js
frederikrothenberger: done

src/internet_identity/src/main.rs
frederikrothenberger: No, changed to ic_certified_map::labeled

src/internet_identity/Cargo.toml
frederikrothenberger: Sorted imports. Referenced latest version via commit hash.

> re. alphabetical order: would be worth checking this and doing it automatically! wdyt?

This would be great.

src/internet_identity/src/signature_map.rs
frederikrothenberger: The library version of rb_tree requires Default.

robin-kunzler (approved):
LGTM, thanks @frederikrothenberger for addressing the comments!

nmattia (approved):
Looking great!

PR #474, opened 2021-11-30
Show tool versions on CI

nmattia:
This helps with debugging.


frederikrothenberger (approved):

frederikrothenberger:
It would be even better to add a .node-version file specifying the required version and referencing that from GitHub actions and the readme.

nmattia:
Right, though still useful to have the actual versions printed out!

Issue #417, opened 2021-09-25
Unable to deploy, Object is of type 'unknown'


spencerbug:
I just updated from dfx verion 0.7.1 to 0.8.1
I deleted the internet-identity repo and started from scratch. 
I'm compiling the latest version of internet identity from main
rustc version 1.52.1
When I call `II_ENV=development dfx deploy --no-wallet --argument '(null)'`
I get the following errors:
`
Deploying all canisters.
Creating canisters...
Creating canister "internet_identity"...
"internet_identity" canister created with canister id: "rwlgt-iiaaa-aaaaa-aaaaa-cai"
Building canisters...
Executing 'src/internet_identity/build.sh'
Compiling frontend assets

> internet_identity_assets@0.1.0 build
> NODE_ENV=production webpack

assets by status 1.21 MiB [cached] 4 assets
orphan modules 275 KiB [orphan] 43 modules
runtime modules 1.16 KiB 6 modules
javascript modules 1 MiB
  modules by path ./node_modules/ 894 KiB 61 modules
  modules by path ./src/frontend/ 132 KiB 36 modules
  crypto (ignored) 15 bytes [built] [code generated]
  util (ignored) 15 bytes [built] [code generated]
  util (ignored) 15 bytes [built] [code generated]
optional modules 159 KiB [optional] 10 modules

ERROR in /Users/sneilan/workspace/blockchain/dfinity/internet-identity/src/frontend/src/flows/addDevice.ts
./src/frontend/src/flows/addDevice.ts 97:18-23
[tsl] ERROR in /Users/sneilan/workspace/blockchain/dfinity/internet-identity/src/frontend/src/flows/addDevice.ts(97,19)
      TS2571: Object is of type 'unknown'.
 @ ./src/frontend/src/index.ts 9:20-48

ERROR in /Users/sneilan/workspace/blockchain/dfinity/internet-identity/src/frontend/src/flows/addDeviceUserNumber.ts
./src/frontend/src/flows/addDeviceUserNumber.ts 81:18-23
[tsl] ERROR in /Users/sneilan/workspace/blockchain/dfinity/internet-identity/src/frontend/src/flows/addDeviceUserNumber.ts(81,19)
      TS2571: Object is of type 'unknown'.
 @ ./src/frontend/src/flows/loginUnknown.ts 10:30-62
 @ ./src/frontend/src/flows/login.ts 13:23-48
 @ ./src/frontend/src/index.ts 7:16-40

[...]

PR #476, opened 2021-11-30
Support for typescript4.4, needs catch(error) to be specified of type Error:any to be printed (defaults to type unknown)

spencerbug:
## Description
https://github.com/dfinity/internet-identity/issues/417

## Motivation and Context
I needed this change to compile internet-identity

## How Has This Been Tested?
This has not gone through any standard regression testing for internet identity
but I have been running this locally with no issues so far. 

## Types of changes
compilation fix

## Checklist:
 
- [ ]  My change requires a change to the documentation.
- [ ]  I have updated the documentation accordingly.
- [ ]  I have added tests to cover my changes.
- [ ]  All new and existing tests passed.


src/frontend/src/flows/login.ts
```diff
@@ -61,12 +61,12 @@ export const login = async (
...
       }
     }
-  } catch (err) {
+  } catch (err:any) {
     await displayError({
       title: "Something went wrong",
       message:
-        "An unexpected error occurred during authentication. Please try again",
-      detail: err,
+        "An unexpected error occured during authentication. Please try again",
```
nmattia, line 68: Mh, I don't think that's correct: 

> Spelling words isn’t always easy, especially when they contain double letters. In this case, the correct spelling is occurred. How can you remember it when you need it?
> 
> The best way to remember how to spell occurred is to remember its double set of double consonants. In English, the final letter is doubled when a word of two or more syllables has stress on the final syllable. Occur fits the rule, so there are two Cs and two Rs in occurred.

https://www.grammarly.com/blog/occurred-occured-ocurred/

nmattia (commented):
One comment; thanks for the PR!

src/frontend/src/flows/login.ts
spencerbug: Woops, I think that happened from a merge conflict. Updated

nmattia:
@spencerbug looks like CI tried to fix the formatting for you but failed to push because you've created a PR from your fork; would you mind running `npm run format` locally? Thanks!

spencerbug:
> 

@nmattia I updated the PR with the linter. Sorry for the delay

nmattia:
Hi @spencerbug , no worries, thanks for your contribution!

nmattia (approved):
LGTM!

PR #478, opened 2021-12-03
Add missing security headers

frederikrothenberger:
Add security headers to make use of browser security features.


src/internet_identity/src/main.rs
```diff
@@ -597,6 +601,41 @@ fn http_request(req: HttpRequest) -> HttpResponse {
     }
 }
 
+fn security_headers() -> Vec<HeaderField> {
```
nmattia, line 604: Can you clarify where you got this list from and what it achieves?

backend-tests/backend-tests.hs
```diff
@@ -462,6 +464,42 @@ validateHttpResponse cid asset resp = do
 
     _ -> lift $ assertFailure $ "Could not parse header: " <> show h
 
+validateSecurityHeaders :: HasCallStack => HttpResponse -> M ()
+validateSecurityHeaders resp = do
+  validateStaticHeader resp "X-Frame-Options" "DENY"
+  validateStaticHeader resp "X-Content-Type-Options" "nosniff"
+  validateStaticHeader resp "Referrer-Policy" "same-origin"
+  validateStaticHeader resp "Permissions-Policy" $ CI.mk $ T.intercalate "," [
```
nmattia, line 472: nitpick: in the rust code you used a multiline string, might be cleaner to do the same here

https://stackoverflow.com/questions/22918837/how-can-i-write-multiline-strings-in-haskell

backend-tests/backend-tests.hs
```diff
@@ -462,6 +464,42 @@ validateHttpResponse cid asset resp = do
...
+    "picture-in-picture=()",
+    "publickey-credentials-get=(self)",
+    "screen-wake-lock=()",
+    "sync-xhr=(self)",
+    "usb=()",
+    "web-share=()",
+    "xr-spatial-tracking=()"
+    ]
+
+validateStaticHeader :: HasCallStack => HttpResponse -> CI T.Text -> CI T.Text -> M ()
```
nmattia, line 495: Why "static"?

nmattia (commented):
Few questions!

backend-tests/backend-tests.hs
frederikrothenberger: Because the values are static (e.g. no regex pattern matching, etc.).

src/internet_identity/src/main.rs
```diff
@@ -535,21 +535,24 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
 #[query]
 fn http_request(req: HttpRequest) -> HttpResponse {
     let parts: Vec<&str> = req.url.split('?').collect();
+    let mut default_headers = security_headers();
```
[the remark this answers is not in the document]

src/internet_identity/src/main.rs
frederikrothenberger: Done

backend-tests/backend-tests.hs
frederikrothenberger: Done

nmattia (approved):
LGTM, thanks!

src/internet_identity/src/main.rs
```diff
@@ -800,20 +841,28 @@ fn stats() -> InternetIdentityStats {
...
                     ContentEncoding::GZip => {
                         vec![("Content-Encoding".to_string(), "gzip".to_string())]
                     }
                 };
+                let content_type = match content_type {
+                    ContentType::HTML => "text/html".to_string(),
+                    ContentType::JS => "text/javascript".to_string(),
+                    ContentType::ICO => "image/vnd.microsoft.icon".to_string(),
+                    ContentType::WEBP => "image/webp".to_string(),
+                };
```
robin-kunzler, line 864: Would it be more elegant to implement this mapping directly on `ContentType`? E.g. like this: 
```
impl ContentType {
    fn to_mime_type_string(&self) -> String {
        match self {
            ContentType::HTML => "text/html".to_string(),
            ContentType::JS => "text/javascript".to_string(),
            ContentType::ICO => "image/vnd.microsoft.icon".to_string(),
            ContentType::WEBP => "image/webp".to_string(),
        }
    }
}
```

src/internet_identity/src/main.rs
```diff
@@ -786,6 +789,44 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
+             payment=(),\
+             picture-in-picture=(),\
+             publickey-credentials-get=(self),\
+             screen-wake-lock=(),\
+             sync-xhr=(self),\
+             usb=(),\
+             web-share=(),\
+             xr-spatial-tracking=()"
+                .to_string(),
+        ),
```
robin-kunzler, line 826: Thanks for adding this! Generally looks good. I saw that some features are missing in the list. Btw. [permissionspolicy.com](https://www.permissionspolicy.com/) is a nice tool to generate these. 

If I disable everything, also proposed and experimental features, I get: 
```
Permissions-Policy: accelerometer=(), ambient-light-sensor=(), autoplay=(), battery=(), camera=(), cross-origin-isolated=(), display-capture=(), document-domain=(), encrypted-media=(), execution-while-not-rendered=(), execution-while-out-of-viewport=(), fullscreen=(), geolocation=(), gyroscope=(), keyboard-map=(), magnetometer=(), microphone=(), midi=(), navigation-override=(), payment=(), picture-in-picture=(), publickey-credentials-get=(self), screen-wake-lock=(), sync-xhr=(self), usb=(), web-share=(), xr-spatial-tracking=(), clipboard-read=(), clipboard-write=(), gamepad=(), speaker-selection=(), conversion-measurement=(), focus-without-user-activation=(), hid=(), idle-detection=(), interest-cohort=(), serial=(), sync-script=(), trust-token-redemption=(), window-placement=(), vertical-scroll=()
```
WDYT, would it make sense to extend it?

src/internet_identity/src/main.rs
```diff
@@ -786,6 +789,44 @@ fn http_request(req: HttpRequest) -> HttpResponse {
     }
 }
 
+/// List of recommended security headers as per https://owasp.org/www-project-secure-headers/
+/// These headers enable browser security features (like limit access to platform apis and set
+/// iFrame policies, etc.).
+fn security_headers() -> Vec<HeaderField> {
+    vec![
```
robin-kunzler, line 796: I also checked identity.raw.ic0.app on securityheaders.com and noticed we don't see _any_ security headers. See [here](https://securityheaders.com/?q=identity.raw.ic0.app&followRedirects=on). 

* Should we add also HSTS, X-Frame-Options, X-Content-Type-Options here?

backend-tests/backend-tests.hs
```diff
@@ -468,6 +470,40 @@ validateHttpResponse cid asset resp = do
 
     _ -> lift $ assertFailure $ "Could not parse header: " <> show h
 
+validateSecurityHeaders :: HasCallStack => HttpResponse -> M ()
```
robin-kunzler, line 473: A general question: Did you check that the headers arrive in the browser also through the boundary nodes nginx? In the worst case some of the headers would be removed by some of the proxies and not arrive in the browser end to end. 

This could be checked using a testnet or we just merge this and then check on prod. Ideally we could write a test, but I'm not sure we can do that, as I don't think we have tests going through the boundary nodes, right?

robin-kunzler (commented):
Thanks a lot for the good work adding the security headers, @frederikrothenberger ! 

Please find a few comments / questions below.

src/internet_identity/src/main.rs
frederikrothenberger: Done

src/internet_identity/src/main.rs
frederikrothenberger: Yes, thanks.

src/internet_identity/src/main.rs
frederikrothenberger: X-Frame-Options, X-Content-Type-Options are both there?
HSTS I missed because I did not check the raw page. Thanks!

backend-tests/backend-tests.hs
frederikrothenberger: If they are not there, then this would be **highly** misleading to developers and imho a bug in the platform.
We know that nginx overwrites certain headers. But this will be resolved later, when we loosen the nginx config (after we have certified headers).
I'm pretty sure other headers are passed through, otherwise the gzip encoding header would also not make it to the client (and this works in production).

robin-kunzler: Sounds good, thanks! Maybe let's still quickly check once this is rolled out, just to be sure ;-)

src/internet_identity/src/main.rs
robin-kunzler: > X-Frame-Options, X-Content-Type-Options are both there?

Oh sure, sorry I got confused ;-) Thanks for adding HSTS!

robin-kunzler (approved):
@frederikrothenberger : LGTM, thanks for addressing the comments!

PR #480, opened 2021-12-03
Audit dependencies periodically

bitdivine:
## Description
Add a github action to audit npm and rust dependencies daily.

## Motivation and Context
`cargo audit` and `npm audit` look for known vulnerabilities in the dependency tree.  We should be aware if such vulnerabilities appear.

## How Has This Been Tested?
- The rust github action is identical to the clause I use in other projects.
- The npm command is a one-liner that I have run and verified that it finds vulnerable dependencies.  I have not run it as  a github cron job in this project; for that I would have to merge it.

## Types of changes
- [ ] New feature (non-breaking change which adds functionality)  **<- Does adding tests count as adding functionality?  Not usually...**

## Checklist:
- [ ] I have added tests to cover my changes.  **<--- These changes are tests**
- [ ] All new and existing tests passed. **<-- No, we have vulnerable dependencies.**


nmattia (commented):
LGTM, can you just clarify how we get notified about this?

bitdivine:
Hello @nmattia.  Any user or org (dfinity or otherwise) watching the repo can configure their own notifications.  See: https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#github-actions-notification-options

bitdivine:
@nmattia I note that some pre-existing emulator tests are timing out.  The failing code and tests are identical to main so this warrants some investigation, even if the outcome is just to increase the timeout.

.github/workflows/audit.yaml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+jobs:
+  audit:
+    name: Rust dependency audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/audit-check@v1
```
nmattia, line 12: quick note: these are apparently unofficial actions, and this one in particular is one big minified JS file, which we give access to our GH actions token. Maybe we can just run `cargo audit`?

nmattia (approved):
> Any user or org (dfinity or otherwise) watching the repo can configure their own notifications. 

Gotcha! So the plan is to get an email if this one fails?

> I note that some pre-existing emulator tests are timing out.

Yes, flakiness in the tests. Needs to be looked at when I have a second.

.github/workflows/audit.yaml
```diff
@@ -0,0 +1,19 @@
+name: Security audit
+on:
+  schedule:
```
nmattia, line 3: I'm guessing this runs every day? Can you add a comment about this?

Re. only `cron` and `workflow_dispatch`; I'm guessing this is intentionally not run on PRs, since this would cause failures unrelated to the PR (for instance: vulnerability DB was updated)? Can you add a comment about that?

.github/workflows/audit.yaml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/audit-check@v1
+  eslint:
+    name: Npm dependency audit
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Audit packages
+      run: git ls-files | grep -E '/package.json$' | xargs -I{} bash -xc 'cd "$(dirname "{}")" && npm audit'
```
nmattia, line 19: Does this match the top level `package.json`?

nmattia: Note: maybe you could use `find . -name 'package.json' -execdir npm audit \;` ?

nmattia:
Closing because it looks a bit outdated; feel free to reopen!

PR #484, opened 2021-12-06
Add cargo fmt to CI

frederikrothenberger:
Refactor workflows to no longer duplicate rust installation and add cargo fmt to CI pipeline.


nmattia (approved):
LGTM!

nmattia (commented):
What happened to having everything in one file?

frederikrothenberger:
> What happened to having everything in one file?

I felt, that having it in backend-tests is not really fitting. So I moved it. If one file is preferred, I'd rather add it to the cargo-tests one (as it is a cargo command as well).

nmattia (commented):
My 2cts: let's move it to `cargo-tests.yml` as a separate job (`cargo-fmt`) which only runs on e.g. ubuntul-latest. WDYT?

frederikrothenberger:
> My 2cts: let's move it to `cargo-tests.yml` as a separate job (`cargo-fmt`) which only runs on e.g. ubuntul-latest. WDYT?

Agreed. I also refactored the rust installation to not duplicate the rust install commands. I hope this is ok.

nmattia (approved):
LGTM, thanks!

PR #485, opened 2021-12-08
Add CSP Header

frederikrothenberger:
Adds CSP header to HTTP responses as a defense-in-depth mechanism against XSS.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -813,6 +813,23 @@ fn security_headers() -> Vec<HeaderField> {
     vec![
         ("X-Frame-Options".to_string(), "DENY".to_string()),
         ("X-Content-Type-Options".to_string(), "nosniff".to_string()),
+
+        // Note: trusted types cannot be used -> FF issue
+        (
+            "Content-Security-Policy".to_string(),
+            "object-src 'none';\
+             script-src 'sha256-Jv0sAUkSG6MPh2x4LKtLcTk/t4cxuu/fPen13STeEsw=' 'unsafe-inline' 'strict-dynamic' https: http:;\
+             base-uri 'none';\
+             default-src 'self';\
+             child-src 'none';\
+             frame-ancestors 'none';\
+             form-action 'none';\
+
+             style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;\
+             style-src-elem 'unsafe-inline' https://fonts.googleapis.com;\
+             font-src https://fonts.gstatic.com"
+                .to_string()
+        ),
         (
             "Strict-Transport-Security".to_string(),
             "max-age=31536000 ; includeSubDomains".to_string(),
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -820,7 +820,7 @@ fn security_headers() -> Vec<HeaderField> {
             "Content-Security-Policy".to_string(),
             "default-src 'none';\
              img-src 'self';\
-             script-src 'sha256-Jv0sAUkSG6MPh2x4LKtLcTk/t4cxuu/fPen13STeEsw=' 'unsafe-inline' 'strict-dynamic' https: http:;\
+             script-src 'sha256-syYd+YuWeLD80uCtKwbaGoGom63a0pZE5KqgtA7W1d8=' 'unsafe-inline' 'strict-dynamic' https: http:;\
              base-uri 'none';\
              frame-ancestors 'none';\
              form-action 'none';\
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -815,8 +815,6 @@ fn security_headers() -> Vec<HeaderField> {
         ("X-Content-Type-Options".to_string(), "nosniff".to_string()),
 
         // Content Security Policy
-        // General overview: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
-        // Additional resources: https://csp.withgoogle.com/docs/index.html and https://csp-evaluator.withgoogle.com/
         //
         // The sha256 hash matches the inline script in index.html. This inline script is a workaround
         // for Firefox not supporting SRI (recommended here https://csp.withgoogle.com/docs/faq.html#static-content).
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -823,8 +823,7 @@ fn security_headers() -> Vec<HeaderField> {
         // script-src 'unsafe-eval' is required because agent-js uses a WebAssembly module for the
         // validation of bls signatures.
         // There is currently no other way to allow execution of WebAssembly modules with CSP.
-        // See https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md and
-        // https://dfinity.atlassian.net/browse/FOLLOW-383.
+        // See https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md.
         //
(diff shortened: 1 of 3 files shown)


src/internet_identity/src/main.rs
```diff
@@ -813,6 +813,31 @@ fn security_headers() -> Vec<HeaderField> {
...
+        // See https://bugzilla.mozilla.org/show_bug.cgi?id=1409200.
+        //
+        // script-src 'unsafe-eval' is required because agent-js uses a WebAssembly module for bls signatures.
+        // There is currently no other way to allow execution of WebAssembly modules with CSP.
+        // See https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md and
+        // https://dfinity.atlassian.net/browse/FOLLOW-383.
+        //
+        // script-src 'unsafe-inline' https: are only there for backwards compatibility and ignored
+        // by modern browsers.
+        (
```
nmattia, line 827: Can you clarify why we need those headers, and what they do? Where did you get this list from? Is this also https://owasp.org/www-project-secure-headers/ ?

src/internet_identity/src/main.rs
```diff
@@ -813,6 +813,31 @@ fn security_headers() -> Vec<HeaderField> {
...
+            "Content-Security-Policy".to_string(),
+            "default-src 'none';\
+             connect-src 'self';\
+             img-src 'self' data:;\
+             script-src 'sha256-syYd+YuWeLD80uCtKwbaGoGom63a0pZE5KqgtA7W1d8=' 'unsafe-inline' 'unsafe-eval' 'strict-dynamic' https:;\
```
nmattia, line 832: How did you come up with the hash? Doesn't it depend on the content of the compiled javascript, i.e. don't we have to update this on every change?

---

Actually, is this the hash of the inline script above?

``` js
      const scripts = ["index.js"];
       scripts.forEach(function (scriptUrl) {
         let s = document.createElement("script");
         s.async = false;
         s.src = scriptUrl;
         document.head.appendChild(s);
       });
```

and if so, doesn't this make the whole CSP header moot? I thought we wanted to bake in the hashes of the loaded javascript files. Sorry, noob here 😅

src/internet_identity/src/main.rs
frederikrothenberger: A good overview over CSP directives is provided here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
As additional resources, I used https://csp.withgoogle.com/docs/index.html and https://csp-evaluator.withgoogle.com/.
In general, they restrict which resources can be used within the execution context of Internet Identity.

src/internet_identity/src/main.rs
frederikrothenberger: Yes, it is the hash of this inline script (so we do not have to update for every change). This does **not** make the CSP moot for the following reasons:
* The CSP still disallows all scripts that are 'parser-inserted'. See here: https://www.w3.org/TR/CSP3/#strict-dynamic-usage
* In our case, we have additional guarantees as those scripts are certified by IC.
* The CSP also has directives unrelated to scripts which improve security as well.

This workaround is recommended [here](https://csp.withgoogle.com/docs/faq.html#static-content) for static pages that cannot use nonces (such as II). But I agree, it would be nicer if we could just rely on SRI. But unfortunately Firefox does not (yet) support this.

bitdivine (approved):
LGTM, thanks!

src/internet_identity/src/main.rs
```diff
@@ -813,6 +813,36 @@ fn security_headers() -> Vec<HeaderField> {
...
+        // This also prevents use of trusted-types. See https://bugzilla.mozilla.org/show_bug.cgi?id=1409200.
+        //
+        // script-src 'unsafe-eval' is required because agent-js uses a WebAssembly module for the
+        // validation of bls signatures.
+        // There is currently no other way to allow execution of WebAssembly modules with CSP.
+        // See https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md and
+        // https://dfinity.atlassian.net/browse/FOLLOW-383.
+        //
+        // script-src 'unsafe-inline' https: are only there for backwards compatibility and ignored
+        // by modern browsers.
```
robin-kunzler, line 830: Thanks for adding the detailed explanations, they are really important to have here.

src/internet_identity/src/main.rs
```diff
@@ -813,6 +813,36 @@ fn security_headers() -> Vec<HeaderField> {
...
+        //
+        // The sha256 hash matches the inline script in index.html. This inline script is a workaround
+        // for Firefox not supporting SRI (recommended here https://csp.withgoogle.com/docs/faq.html#static-content).
+        // This also prevents use of trusted-types. See https://bugzilla.mozilla.org/show_bug.cgi?id=1409200.
+        //
+        // script-src 'unsafe-eval' is required because agent-js uses a WebAssembly module for the
+        // validation of bls signatures.
+        // There is currently no other way to allow execution of WebAssembly modules with CSP.
+        // See https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md and
+        // https://dfinity.atlassian.net/browse/FOLLOW-383.
```
robin-kunzler, line 827: Since this is open source we should not link to the internal FOLLOW issue. Could you remove that link?

src/internet_identity/src/main.rs
```diff
@@ -813,6 +813,36 @@ fn security_headers() -> Vec<HeaderField> {
...
+            "Content-Security-Policy".to_string(),
+            "default-src 'none';\
+             connect-src 'self';\
+             img-src 'self' data:;\
+             script-src 'sha256-syYd+YuWeLD80uCtKwbaGoGom63a0pZE5KqgtA7W1d8=' 'unsafe-inline' 'unsafe-eval' 'strict-dynamic' https:;\
+             base-uri 'none';\
+             frame-ancestors 'none';\
+             form-action 'none';\
+             style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;\
+             style-src-elem 'unsafe-inline' https://fonts.googleapis.com;\
```
robin-kunzler, line 841: Would it make sense to add a comment explaining why `unsafe-inline` styles are needed? 

As discussed, removing these would probably be a major change and I agree that this can't be addressed in this PR.

src/internet_identity/src/main.rs
```diff
@@ -813,6 +813,36 @@ fn security_headers() -> Vec<HeaderField> {
...
+             font-src https://fonts.gstatic.com;\
```
robin-kunzler, line 842: Could we avoid loading fonts from Google and instead host them on the canister? This doesn't have to be done in this issue/PR, but maybe create a ticket and add a TODO here? 

Also, why is `https://fonts.googleapis.com` needed in style-src? Are styles (not just fonts) loaded from there?

robin-kunzler (approved):
@frederikrothenberger : Thanks a lot for adding the CSP, this looks good to me!

Please find a few minor comments below. 

Also, did you do some testing (looking into the console) with different browsers (also maybe old ones not supporting CSP3) to make sure things are working as expected?

src/internet_identity/src/main.rs
frederikrothenberger: We also load css from google (in main.css):
`@import url("https://fonts.googleapis.com/css2?family=Montserrat:wght@400;500;600;700&display=swap");`

`https://fonts.googleapis.com` is required on style-src by Firefox and on style-src-elem by Chrome.
Ticket created: https://dfinity.atlassian.net/browse/II-149

src/internet_identity/src/main.rs
frederikrothenberger: Done.

src/internet_identity/src/main.rs

PR #486, opened 2021-12-09
Also send security headers in the metrics error case

frederikrothenberger:
Sets security headers also on the /metrics error response.


nmattia (approved):

robin-kunzler:
Thanks @frederikrothenberger !

PR #488, opened 2021-12-10
Don't ask for confirmation on webauthn failure

nmattia:
Before this change, and since the introduction of CAPTCHA, canceling the
identity creation would display the "Error" screen _but_ we would then
show the anchor creation confirmation form nonetheless as soon as the
CAPTCHA would be ready.

This makes sure we do either, or.

### DIFF
--- a/src/frontend/src/flows/register.ts
+++ b/src/frontend/src/flows/register.ts
@@ -59,18 +59,19 @@ const init = (): Promise<LoginResult | null> =>
       try {
         const pendingIdentity = WebAuthnIdentity.create({
           publicKey: creationOptions(),
-        }).catch((error) => {
-          resolve(apiResultToLoginResult({ kind: "authFail", error }));
-          // We can never get here, but TS doesn't understand that
-          return 0 as unknown as WebAuthnIdentity;
         });
         await tick();
         // Kick-start both the captcha creation and the identity
         const captcha = makeCaptcha();
-        const identity = await pendingIdentity;
-        await captcha;
-        const result = await confirmRegister(captcha, identity, alias);
-        resolve(result);
+        pendingIdentity
+          .catch((error) => {
+            resolve(apiResultToLoginResult({ kind: "authFail", error }));
+            // We can never get here, but TS doesn't understand that
+            return 0 as unknown as WebAuthnIdentity;
+          })
+          .then((identity) => {
+            confirmRegister(captcha, identity, alias).then(resolve);
+          });
       } catch (err) {
         reject(err);
       }
--- a/src/frontend/src/flows/register.ts
+++ b/src/frontend/src/flows/register.ts
@@ -62,16 +62,18 @@ const init = (): Promise<LoginResult | null> =>
         });
         await tick();
         // Kick-start both the captcha creation and the identity
-        const captcha = makeCaptcha();
-        pendingIdentity
-          .catch((error) => {
+        Promise.all([
+          makeCaptcha(),
+          pendingIdentity.catch((error) => {
             resolve(apiResultToLoginResult({ kind: "authFail", error }));
             // We can never get here, but TS doesn't understand that
             return 0 as unknown as WebAuthnIdentity;
-          })
-          .then((identity) => {
-            confirmRegister(captcha, identity, alias).then(resolve);
-          });
+          }),
+        ]).then(([captcha, identity]) => {
+          confirmRegister(Promise.resolve(captcha), identity, alias).then(
+            resolve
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/register.ts
```diff
@@ -59,18 +59,19 @@ const init = (): Promise<LoginResult | null> =>
...
         await tick();
         // Kick-start both the captcha creation and the identity
         const captcha = makeCaptcha();
-        const identity = await pendingIdentity;
-        await captcha;
-        const result = await confirmRegister(captcha, identity, alias);
-        resolve(result);
+        pendingIdentity
+          .catch((error) => {
+            resolve(apiResultToLoginResult({ kind: "authFail", error }));
```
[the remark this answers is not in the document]
nmattia: Thanks for the suggestions! I'm not sure I understood everything though, I think I lack hands on async experience...

> To transform the error, why not `catch(error => throw apiResultToLoginResult({ kind: "authFail", error }))`? That is the equivalent of `map_err(...)` in Rust and is more idiomatic.

But here the behavior is quite different, correct? In your suggestion, the error isn't `catch`ed and will bubble up to the console. I've tried not to modify the original behavior (written by someone else) where the promise always returns a `LoginResult`, be it success or failure (which is why we `catch` in the first place).

> For running the two promises in parallel, why not use: ...

Here, the behavior is different I believe. It would wait until the captcha has been created before showing the confirmation page, whereas now it kickstarts both but only waits for the identity to be ready. I might be wrong, I'm not very fluent with promises!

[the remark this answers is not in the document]
nmattia: While the behavior is a bit different, I think `Promise.all` is much cleaner, so I followed your suggestion. PTAL!

src/frontend/src/flows/register.ts
```diff
@@ -59,18 +60,20 @@ const init = (): Promise<LoginResult | null> =>
...
-        const result = await confirmRegister(captcha, identity, alias);
-        resolve(result);
+        Promise.all([makeCaptcha(), pendingIdentity])
+          .catch((error) => {
+            resolve(apiResultToLoginResult({ kind: "authFail", error }));
+            // We can never get here, but TS doesn't understand that
+            return 0 as unknown as [Challenge, WebAuthnIdentity];
+          })
+          .then(([captcha, identity]) => {
+            confirmRegister(Promise.resolve(captcha), identity, alias).then(
```
bitdivine, line 73: The captcha has already been resolved by virtue of being an output of promise.all.  If all were successful, it has been resolved, if one failed, the others will have been aborted.

src/frontend/src/flows/register.ts
```diff
@@ -59,18 +60,20 @@ const init = (): Promise<LoginResult | null> =>
...
+              resolve
+            );
+          });
```
bitdivine, line 76: Does this do what you want?
- It skips the confirmRegister if any of makeCaptcha or pendingIdentity fail.
- The entire chain is infallible, due to the last line.
- No resolve.
```
        Promise.all([makeCaptcha(), pendingIdentity])
          .catch(error => throw apiResultToLoginResult({ kind: "authFail", error }))
          .then(([captcha, identity]) => confirmRegister(captcha, identity, alias))
          .catch(e => e);
```
Note that any failures in the confirmRegister might not be LoginResult errors unless you map those as well.  It looks as if there are several different error types that can happen there.  I feel tempted to think that if re-rendering fails something is really messed up and maybe the right thing to do is to call reload(), effectively making the confirmRegister infallible.  It will never report anything but success.  Alternatively, you could map all errors to login result errors with:
```
        Promise.all([makeCaptcha(), pendingIdentity])
          .then(([captcha, identity]) => confirmRegister(captcha, identity, alias))
          .catch(error => apiResultToLoginResult({ kind: "authFail", error }));
```

bitdivine (approved):

src/frontend/src/flows/register.ts
nmattia: I know, but I need it to typecheck :)

src/frontend/src/flows/register.ts
nmattia: Does this do what you want?

It skips the confirmRegister if any of makeCaptcha or pendingIdentity fail.
The entire chain is infallible, due to the last line.
No resolve.

---

Yes, this is what I want; I believe this is what the original author meant to do.


> Alternatively, you could map all errors to login result errors with:

I like this. Let me edit.

PR #489, opened 2021-12-13
Disable registration on unexpected origins

frederikrothenberger:
Disable registrations on unexpected origins, most notably https://identity.raw.ic0.app.
Using unexpected origins is a security risk and breaks the integration with dApps authenticating via Internet Identity.


frederikrothenberger:
Screenshot of info message:
Screenshot 2021-12-15 at 14 59 03
The 'here' link points to the deprecation announcement here: https://forum.dfinity.org/t/internet-identity-deprecation-of-account-creation-on-all-origins-other-than-https-identity-ic0-app/9694

src/frontend/src/flows/loginUnknown.ts
```diff
@@ -99,6 +100,14 @@ export const loginUnknown = async (
   });
 };
 
+function isRegistrationAllowed() {
```
nmattia, line 103: a note here and some comments in `registerDisabled.ts` would be great!

src/frontend/src/flows/loginUnknown.ts
```diff
@@ -99,6 +100,14 @@ export const loginUnknown = async (
...
+  return (
+    /^(https:\/\/identity\.ic0\.app$)|(https:\/\/[a-zA-Z0-9-]+\.[a-z]+\.dfinity\.network$)/.test(
```
nmattia, line 105: There is exactly one endpoint (namely `identity.raw.ic0.app`) that you want to disable registration on. Why not simply check for that?

dubovitskaya (approved):
Thanks Frederik! The text looks good!
Just a small question - maybe we dont need the Logout button in this case?

src/frontend/src/flows/loginUnknown.ts
frederikrothenberger: Done. AFAIK we have to check against 3 origins:
https://identity.raw.ic0.app, https://rdmx6-jaaaa-aaaaa-aaadq-cai.ic0.app, https://rdmx6-jaaaa-aaaaa-aaadq-cai.raw.ic0.app

Are you sure there are not more?

nmattia: I am not! I actually didn't even think of rdmx6-..., good catch! If we find more, then let's add them. Thanks!

frederikrothenberger:
> Thanks Frederik! The text looks good! Just a small question - maybe we dont need the Logout button in this case?

Thanks. I removed the logout button.

src/frontend/src/flows/loginUnknown.ts
frederikrothenberger: Done.

nmattia (approved):
LGTM!

frederikrothenberger:
Note: Not merging until 22.12.2021 due to deprecation period.

frederikrothenberger:
Manual test of this feature on the identity test successful.

ggreif:
Just a drive-by comment, does this code also apply to the testnets? We have https://rdmx6-jaaaa-aaaaa-aaadq-cai.peopleparty01.dfinity.network, will that be disabled too?

frederikrothenberger:
@ggreif: No. This will only affect https://rdmx6-jaaaa-aaaaa-aaadq-cai.ic0.app and anything using .raw.

PR #493, opened 2021-12-16
Clarify credentials creation options

### LINKED ISSUE #491: why is the publicKeyCredentialCreationOptions challenge not random
From [webauthn.guide](https://webauthn.guide/#registration):

> challenge: The challenge is a buffer of cryptographically random bytes generated on the server, and is needed to prevent "replay attacks".

From the [specs](https://w3c.github.io/webauthn/#sctn-cryptographic-challenges):

> As a cryptographic protocol, Web Authentication is dependent upon randomized challenges to avoid replay attacks. Therefore, the values of both PublicKeyCredentialCreationOptions.challenge and PublicKeyCredentialRequestOptions.challenge MUST be randomly generated by Relying Parties in an environment they trust (e.g., on the server-side), and the returned challenge value in the client’s response MUST match what was generated. This SHOULD be done in a fashion that does not rely upon a client’s behavior, e.g., the Relying Party SHOULD store the challenge temporarily until the operation is complete. Tolerating a mismatch will compromise the security of the protocol. 
> In order to prevent replay attacks, the challenges MUST contain enough entropy to make guessing them infeasible. Challenges SHOULD therefore be at least 16 bytes long.

But the implementation is not random:

[`createOptions`](https://github.com/dfinity/internet-identity/blob/86e867acce5a0c0e4b35a8dcf9269a8014283c7f/src/frontend/src/utils/iiConnection.ts#L389) used in [`register`](https://github.com/dfinity/internet-identity/blob/86e867acce5a0c0e4b35a8dcf9269a8014283c7f/src/frontend/src/flows/register.ts#L62)

nmattia:
Fixes #491 

This explains some of the reasoning behind the creation options used for
the WebAuthn credentials.

### DIFF
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -369,6 +369,23 @@ const requestFEDelegation = async (
   return DelegationIdentity.fromDelegation(sessionKey, chain);
 };
 
+// The options sent to the browser when creating the credentials.
+// Credentials (key pair) creation is signed with a private key that is unique per device
+// model, as an "attestation" that the credentials were created with a FIDO
+// device. In II we discard this attestation because we only care about the key
+// pair that was created and that we use later. Discarding the attestation
+// means we do not have to care about attestation checking security concerns
+// like setting a server-generated, random challenge.
+//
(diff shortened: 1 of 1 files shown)


frederikrothenberger (approved):
LGTM.

PR #496, opened 2022-01-03
Allow application to copy seedphrase to clipboard

frederikrothenberger:
Allow internet identity to write the seed phrase to the clipboard using the Clipboard API.

Note: this is just future-proofing (not a bugfix) as currently the document.execCommand('copy') is being used, which is not governed by the premissions-policy.


nmattia (approved):
LGTM!

nmattia (commented):
Two questions: 

* what is `self`, and where can I read up on how this works? The [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API) MDN page doesn't mention `self`.
* The `Clipboard.write` [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) mentions that `clipboard-write` is granted when the tab is active. Does that mean your change here only applies to inactive tabs? In this case, is it still necessary?

frederikrothenberger:
>     * what is `self`, and where can I read up on how this works? The [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API) MDN page doesn't mention `self`.

The 'self' allow-list entry is explained here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy
Note: Feature-Policy has now been renamed to Permissions-Policy.

 
>     * The `Clipboard.write` [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) mentions that `clipboard-write` is granted when the tab is active. Does that mean your change here only applies to inactive tabs? In this case, is it still necessary?

Apparently, it is only relevant in the context of iFrames. See this specification explainer: https://docs.google.com/document/d/1ZHU2M8hGX3yUtE_SRx35ZpYk33gnBJLSFO-hgUCvShI/edit#

I would opt for keeping this change anyway as we _are_ using the clipboard and specifying () is misleading.

nmattia (approved):
LGTM, thanks for the clarification!

PR #503, opened 2022-01-13
Update agent-js to v0.10.2

frederikrothenberger:
Update agent-js to newest version.


src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -118,7 +118,7 @@ test("Log into client application, after registration", async () => {
 
     // default value
     const exp = await browser.$("#expiration").getText();
-    expect(Number(exp) / (30 * 60_000_000_000)).toBeCloseTo(1);
+    expect(Number(exp) / (24 * 60 * 60_000_000_000)).toBeCloseTo(1);
```
nmattia, line 121: didn't you want to change this?

nmattia (approved):
LGTM, though would be even better to have @krpeacock or someone very familiar with agent-js have a look as well

src/frontend/src/test-e2e/selenium.test.ts
nmattia: I see, this will be a(security) breaking change for many apps though, I'm not sure how to communicate that... but arguably it's a problem for `agent-js`

krpeacock: You can have the sample client dapp set whichever `maxTimeToLive` you like. There will need to be another refactor once I develop the idle timeout flow for agent-js 0.11.0, however

frederikrothenberger: This is a test for the default value. We have other tests that do set maxTimeToLive.

PR #505, opened 2022-01-14
Align actor creation

frederikrothenberger:
Align actor creation so that the root_key is (not) fetched consistently.


nmattia (approved):
LGTM! Didn't know the identity was optional

PR #506, opened 2022-01-14
Correctly display error message on registration failure

frederikrothenberger:
Correctly display error message on registration failure. Added special treatment for the 'registerNoSpace' case.

### DIFF
--- a/src/frontend/src/flows/confirmRegister.ts
+++ b/src/frontend/src/flows/confirmRegister.ts
@@ -67,20 +67,26 @@ const tryRegister = (
       confirmParagraph.innerHTML =
         "The value you entered is incorrect. A new challenge is generated.";
       requestCaptcha();
-    } else {
-      // Currently if something goes wrong we only tell the user that
+    } else if (result.kind == "registerNoSpace") {
+      // Currently, if something goes wrong we only tell the user that
       // something went wrong and then reload the page.
+      displayError({
+        title: "No Space Left",
+        message:
+          "We could not create an identity anchor because Internet Identity is at maximum capacity. Click 'ok' to reload.",
+        primaryButton: "Ok",
+      }).then(() => {window.location.reload();})
+    } else {
       displayError({
         title: "Something went wrong",
         message:
           "We could not create an identity anchor. You will find the full error message below. Click 'ok' to reload.",
         primaryButton: "Ok",
-        detail: JSON.stringify(result),
+        detail: JSON.stringify(result.error, Object.getOwnPropertyNames(result.error)),
       })
         .then(() => {
           window.location.reload();
-        })
-        .then(() => requestCaptcha());
+        });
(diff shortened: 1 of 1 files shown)


nmattia (commented):
This is odd, I'm pretty sure I've seen an error message that was mentioning capacity when I stress tested the size. Did you actually try to fill up a canister and then got an unknown error?

frederikrothenberger:
> This is odd, I'm pretty sure I've seen an error message that was mentioning capacity when I stress tested the size. Did you actually try to fill up a canister and then got an unknown error?

Yes, `{"kind":"registerNoSpace"}` are shown even with the current master. The main advantage of having a separate `if` branch is that typescript can infer, that all variants in the `else` branch have an error field. So I can use `result.error` without any casting / conversions.

I also changed the PR description to make the intent more clear (which is showing `result.error` correctly).

src/frontend/src/flows/confirmRegister.ts
```diff
@@ -67,20 +67,30 @@ const tryRegister = (
...
         primaryButton: "Ok",
-        detail: JSON.stringify(result),
-      })
-        .then(() => {
-          window.location.reload();
-        })
-        .then(() => requestCaptcha());
+        detail: JSON.stringify(
+          result.error,
+          Object.getOwnPropertyNames(result.error)
```
nmattia, line 89: What's this for?

src/frontend/src/flows/confirmRegister.ts
```diff
@@ -67,20 +67,30 @@ const tryRegister = (
...
         title: "Something went wrong",
         message:
           "We could not create an identity anchor. You will find the full error message below. Click 'ok' to reload.",
         primaryButton: "Ok",
-        detail: JSON.stringify(result),
-      })
-        .then(() => {
-          window.location.reload();
-        })
-        .then(() => requestCaptcha());
```
nmattia, line 83: Looks like we're not requesting a new captcha in case of failure, where did that go?

src/frontend/src/flows/confirmRegister.ts
frederikrothenberger: It forces `stringify` to also print out the fields which are not `enumerable` by explicitly listing the properties of `error` we want to get printed.

src/frontend/src/flows/confirmRegister.ts
frederikrothenberger: I removed it because the `window.location.reload()` makes it obsolete. On reload, we are on the welcome page again -> when we chose try again the captcha will be requested then.

nmattia: Mh, I really wonder what that was doing there. Bad copy paste I guess?

nmattia (approved):

PR #513, opened 2022-01-25
Change workflows to run on push

nmattia:
This makes sure the GHA workflows run on any push. This means:

* The actual commit is used during the tests (as opposed to
  pull_request which uses a merge commit). This is simpler and anyway we
enforce that the branch must be up to date with master.
* The Actions tab show the commit name for the workflows
* Tests are run on branches too, meaning one doesn't have to create a
  draft PR just for the sake of running tests anymore

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -2,9 +2,6 @@ name: Backend tests
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


frederikrothenberger (approved):
LGTM!

PR #522, opened 2022-02-01
Fix docker-build comment

nmattia:
It wasn't up-to-date; the builds now take ~3mn and run on every push.


frederikrothenberger (approved):
Thanks for updating that

PR #523, opened 2022-02-01
Compute asset hashes in post_upgrade

nmattia:
This removes the build-time hash computation and make it happen when
`for_each_asset` is called, effectively in `post_upgrade` (and `init`).

The interface is still a bit cumbersome (e.g. `for_each_asset` is still
callback based) but this makes sure that `main.rs` is not changed; the
focus here is the asset hashes. This is the first step in removing
`CANISTER_ID` from the build process; we want to inject it in the assets
directly, and this will change `main.rs` and `for_each_asset`.

I've tested this locally and the certificate appears to be the same. What's the correct way of checking that everything is OK?

NOTE: needs https://github.com/dfinity/internet-identity/pull/524


roman-kashitsyn:
> This is the first step in removing CANISTER_ID from the build process

I'm not sure I see the connection: `build.rs` doesn't use `CANISTER_ID` in any way.
We also don't need to know the canister id to certify the assets.

The reason I implemented hashing at build time is to make sure we never have to do a lot of work on an upgrade, so we can't run out of cycles.

src/internet_identity/src/assets.rs
```diff
@@ -1 +1,89 @@
...
+    }
+}
+
+
+// Hash the content of an asset in an `ic_certified_map` friendly way
+fn hash_content(bytes: &[u8]) -> [u8; 32] {
+    let mut hasher = sha2::Sha256::new();
+    hasher.update(&bytes);
+    hasher.finalize().into()
+}
```
roman-kashitsyn, line 89: I believe this can be just `sha2::Sha256::digest(bytes).into()`.

roman-kashitsyn:
> I'm not sure I see the connection: `build.rs` doesn't use `CANISTER_ID` in any way.

I think I figured it out: we want to substitute CANISTER_ID in assets during the upgrade, this means we don't know it at build time. Makes sense.

src/internet_identity/src/assets.rs
nmattia: Thanks, updating

nmattia:
@roman-kashitsyn exactly. If we ever run into the limit, we can always pre-compute some hashes (and add some tests that ensure they are kept up to date).

PR #525, opened 2022-02-02
Allow https://ic0.app in CSP

frederikrothenberger:
Fixes the CSP to allow connecting to ic0.app. This change is required by the new version of agent-js.


nmattia (approved):
Thanks, also I guess technically this should only be enabled on mainnet? And this assumes the domain name won't change?

frederikrothenberger:
> Thanks, also I guess technically this should only be enabled on mainnet? And this assumes the domain name won't change?

I think, it is a reasonable assumption that the domain won't change (or if it does, that we can do another release). Also, I'm now working on the PR to make this change also necessary for the development environment (selenium tests) so I would not hide it behind a dev/prod flag.

PR #527, opened 2022-02-03
Use origin https://identity.ic0.app in selenium tests

frederikrothenberger:
Use the maninnet hostname for selenium tests in order to get accurate feedback for features depending on the hostname (e.g. CSP).


reverse_proxy/nginx.conf
```diff
@@ -0,0 +1,54 @@
+worker_processes  1;
+
+#error_log  logs/error.log;
+#error_log  logs/error.log  notice;
+#error_log  logs/error.log  info;
+#pid        logs/nginx.pid;
+
```
nmattia, line 7:
```suggestion
```

src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -37,10 +37,11 @@ import canister_ids2 from "../../../../demos/whoami/.dfx/local/canister_ids.json
 const IDENTITY_CANISTER = canister_ids1.internet_identity.local;
 const WHOAMI_CANISTER = canister_ids2.whoami.local;
 
-const REPLICA_URL = "http://localhost:8000";
-const II_URL = `http://localhost:8000/?canisterId=${IDENTITY_CANISTER}`;
-const FAQ_URL = `http://localhost:8000/faq?canisterId=${IDENTITY_CANISTER}`;
-const ABOUT_URL = `http://localhost:8000/about?canisterId=${IDENTITY_CANISTER}`;
+const REPLICA_URL = "https://ic0.app";
```
nmattia, line 40: That's gonna break the tests locally unfortunately. Can you make this conditional somehow?

src/frontend/src/test-e2e/util.ts
```diff
@@ -48,6 +48,7 @@ export async function runInBrowserCommon(
       "goog:chromeOptions": {
         args: [
           "--headless",
+          "--ignore-certificate-errors",
```
nmattia, line 51: Can you add a note about why this is needed? Also I reckon this is only needed/wanted in the github actions runs, not locally

dfx.json
```diff
@@ -14,7 +14,7 @@
   },
   "networks": {
     "local": {
-      "bind": "127.0.0.1:8000",
+      "bind": "0.0.0.0:8000",
```
nmattia, line 17: What's the issue with 127.0.0.1? Guessing this may have something to do with being accessible from Docker. If that's the case, can you create a ticket about reverting this once/if we find a better setup?

nmattia (commented):
Nice! Can you add a README in `reverse_proxy/`? Should include why we do this, an overview of how this works, how the certs were generated, etc.

src/frontend/src/test-e2e/util.ts
frederikrothenberger: No, this is always required, because we use self-signed certificates.

src/frontend/src/test-e2e/selenium.test.ts
frederikrothenberger: Can we keep it for now? We'll move to docker-compose ASAP anyway.

dfx.json
frederikrothenberger: Yes, excactly. The requests coming out of docker do apparently not use `127.0.0.1`. This will be reverted once we use docker-compose.

nmattia (approved):
Thanks!

src/frontend/src/test-e2e/selenium.test.ts
frederikrothenberger: Added a switch

PR #528, opened 2022-02-03
Read canister ID from canister

nmattia:
This includes a few changes that together make the frontend read the canister ID from the canister. This means:
* The same wasm module can be used anywhere (since the canister ID isn't baked in anymore)
* The webpack config is a lot saner and doesn't rely on assumptions and hard coded values
* You can now `npm run build` without having to first start the replica + call dfx canister create (or after having wiped `.dfx/`).

This includes the following changed:

* The `<script>` tag in `index.html` is now a dummy; the real content is injected in the backend. The real content includes the canister ID. Notes:
  * By doing this -- as opposed to e.g. fetching a config file also served by the canister -- we don't need to deal with promises and the FE code doesn't need to change
  * This means the SRI hash (for CSP) is now dynamic, so we compute it on startup (init/post_upgrade)
* Iterating over the assets (to insert the hashes in the certification map/tree) isn't callback-based anymore; instead all assets have a static lifetime. This also ensures that assets are created once and (in the case of index.html) aren't duplicated.
* The webpack config changed a bit. In order to retain the live reload feature (when using `npm run start`) we make sure that _only_ the `index.html` is fetched from the backend; that way, the index.js is still local. The webpack config doesn't assume that the canister has been deployed anymore; now it only tried reading canister IDs when the server is started; before, it would do so at build time as well, meaning you couldn't build the FE if your canister hadn't been deployed (ok, created). I think this is much, much saner.


src/internet_identity/src/assets.rs
```diff
@@ -19,15 +21,43 @@ pub enum ContentType {
...
+pub fn get_assets() -> [ (&'static str, &'static [u8], ContentEncoding, ContentType); 8]  {
+    let index_html: &[u8] = INDEX_HTML_STR.as_bytes();
+    [
          ("/",
-            index_html,
-            ContentEncoding::Identity,
-            ContentType::HTML,
+             index_html,
+             ContentEncoding::Identity,
+             ContentType::HTML,
```
bitdivine, line 60: Is a formatter run as part of CI?  I would not expect to see formatting-only changes in a PR.

bitdivine:
How has this code been tested?

src/internet_identity/src/assets.rs
```diff
@@ -19,15 +21,43 @@ pub enum ContentType {
...
+    };
 
-    let assets: [ (&str, &[u8], ContentEncoding, ContentType); 8] = [
+    // The full content of the index.html, after the canister ID (and script tag) have been
+    // injected
+    static ref INDEX_HTML_STR: String = {
+        let index_html = include_str!("../../../dist/index.html");
+        let foo: String = INDEX_HTML_SETUP_JS.to_string();
+        let index_html = index_html.replace(
+            "<script id='setupJs'></script>",
```
peterpeterparker, line 45: is it intended that the `script` is not set as `lazy` or meant for`type="module"`?

randomly noticed this, just in case it would be useful.

webpack.config.js
```diff
@@ -43,13 +38,58 @@ function generateWebpackConfigForCanister(name, info) {
...
+              pathRewrite: (pathAndParams, req) => {
+                  let queryParamsString = `?`;
+
+                  const [path, params] = pathAndParams.split("?");
+
+                  if (params) {
+                      queryParamsString += `${params}&`;
+                  }
+
+                  queryParamsString += `canisterId=${canisterId}`;
```
frederikrothenberger, line 80: This is so that we can access the dev server on an url without the parameter right?
But shouldn't there also be changes in the `selenium.test.ts` file? Otherwise we have the canisterID twice.

src/internet_identity/src/assets.rs
nmattia: I didn't even remember we did, so I created this: https://dfinity.atlassian.net/browse/L2-230

@bitdivine no, the first item was off so I aligned it, didn't realize it wasn't like that before. Happy to revert if you feel strongly about it

src/internet_identity/src/assets.rs
```diff
@@ -19,15 +21,43 @@ pub enum ContentType {
...
+            &format!("<script id='setupJs'>{foo}</script>").to_string()
```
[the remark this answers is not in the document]
nmattia: oh yeah, will change the name

src/internet_identity/src/assets.rs
nmattia: Thanks! This was already here, so I left it as is. I'll create a ticket to look into it.

webpack.config.js
```diff
@@ -43,13 +38,58 @@ function generateWebpackConfigForCanister(name, info) {
...
+          const dfxJson = './dfx.json';
+          let replicaHost;
+
+          try {
+              replicaHost = require(dfxJson).networks.local.bind;
+          } catch (e) {
+              throw Error(`Could get host from ${dfxJson}: ${e}`);
+          }
+
+          if(!replicaHost.startsWith("http://")) {
```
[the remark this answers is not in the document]
nmattia: Because `localhost:8000` doesn't work, it needs to be prefixed with `http://`. Will add a comment.

webpack.config.js
nmattia: Good point!

src/internet_identity/src/assets.rs
nmattia: https://dfinity.atlassian.net/browse/L2-233

webpack.config.js
nmattia: Ah, actually no. Looks like the selenium tests directly connect to the replica. We can homogenize everything at some point (I'd like a setup like the [nns-dapp](https://github.com/dfinity/nns-dapp/tree/main/e2e-tests))

nmattia:
> How has this code been tested?
> - @bitdivine 

This has been tested manually locally and automatically by selenium and the backend-tests on CI. This will also allow us to test the wasm on mainnet with a different canister ID than the official II.

frederikrothenberger (approved):

PR #531, opened 2022-02-07
Use build.sh in Dockerfile

nmattia:
This makes sure the docker build uses the same commands as `dfx build`.
We're now specifying `ic-cdk-optimizer` as a dependency -- as it should
be -- potentially trivially resolving https://github.com/dfinity/internet-identity/issues/381

### DIFF
--- a/.github/workflows/selenium.yml
+++ b/.github/workflows/selenium.yml
@@ -40,6 +40,25 @@ jobs:
           rustup default "$RUSTC_VERSION"
           rustup target add wasm32-unknown-unknown
 
+      # Cache the ic-cdk-optimizer
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.local/bin/ic-cdk-optimizer
+            target
+          key: ${{ runner.os }}-ic-cdk-optimizer-${{ env.IC_CDK_OPTIMIZER_VERSION }}
+
+      - name: Install ic-cdk-optimizer
+        run: |
+          install_dir="$HOME/.local/bin/"
+          mkdir -p "$install_dir"
+          if [ ! -x "$install_dir/ic-cdk-optimizer" ]
+          then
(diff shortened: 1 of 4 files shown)


peterpeterparker (approved):

.github/workflows/selenium.yml
```diff
@@ -40,6 +40,25 @@ jobs:
...
+        run: |
+          install_dir="$HOME/.local/bin/"
+          mkdir -p "$install_dir"
+          if [ ! -x "$install_dir/ic-cdk-optimizer" ]
+          then
+            cargo install --version "0.3.1" ic-cdk-optimizer
+            cp $HOME/.cargo/bin/ic-cdk-optimizer "$install_dir"
+            echo "$install_dir" >> $GITHUB_PATH
+          fi
+
```
bitdivine, line 61: I am unfamiliar with this cacher but it seems odd that the caching happens before the optimizer is installed.  Is the former clause fetching FROM the cache?  That would make sense, followed by "if it wasn't in the cache, make it".

bitdivine (approved):
LGTM, thanks.  One quibble about caching but that's probably just because I don't know enough about it.

.github/workflows/selenium.yml
nmattia: Fair point!

Calling the action does two things, where `actions/cache@v2` is called:

1. checks the remote cache for an entry with the given key
2. register a "delayed" step to happen _after_ the action has completed

so in this case this is exactly what we want. On the first ever run, there will be a cache miss; the following step (`Install ic-cdk-optimizer`) will install ic-cdk-optimizer and the delayed step will actually populate the cache.

Note that if the `actions/cache@v2` ran after the ic-cdk-optimizer step we'd be installing (i.e. building) ic-cdk-optimizer on every run.

nmattia: For more info: https://github.com/actions/cache

PR #537, opened 2022-02-11
Add USE_DUMMY_AUTH build that bypasses webauthn

nmattia:
This introduces a build-time option, `USE_DUMMY_AUTH`, that will bypass all uses of webauthn. The resulting build is useful for external developers wishing to test their dapp with an II login.

It's particularly tricky to test webauthn flows, in particular only chrome seems to support testing it (through its virtual authenticator) and by far not all test frameworks support it. In this repo we use selenium (as opposed to some lighterweight, more recent framework) because we need to test the webauthn flow.

To make life simpler for other people, we generate this build. It goes without saying that this build should never land in production.

**About this PR**:

This adds the `USE_DUMMY_AUTH` opt-in option. When enabled, a known key pair (generated with a known, 32-byte long `\NUL` seed) is used both for creating the anchor and for authenticating. The `dummy-auth-tests` test directory (which accounts for most of the diff) adds a single [webdriver.io](https://webdriver.io) test that serves two purposes. First, it ensures that the build indeed doesn't require webauthn. Then, it serves as an example for external developers wishing to test their II integration.

**TODOs for the next PR**:

* [Sanitize the build options](https://dfinity.atlassian.net/browse/L2-254); in particular I'd want to always show a red banner on the landing page if any of the build options are enabled (fetch root key, dummy captcha, dummy auth, etc)
* [Improve dfx-proxy](https://dfinity.atlassian.net/browse/L2-255) and move most of the code in `dummy-auth-tests/test.js` out and into `dfx-proxy` itself.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -92,7 +92,9 @@ jobs:
           name: internet_identity_dummy_captcha.wasm
           path: out/internet_identity.wasm
 
-  # TODO: document
+  # A version of II where a known keypair is used for every anchor registration
+  # and authentication. This is meant for testing. This also includes the dummy
+  # captcha.
   docker-build-dummy-auth:
     runs-on: ubuntu-latest
     needs: docker-build-base
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -92,6 +92,36 @@ jobs:
           name: internet_identity_dummy_captcha.wasm
           path: out/internet_identity.wasm
 
+  # TODO: document
+  docker-build-dummy-auth:
+    runs-on: ubuntu-latest
+    needs: docker-build-base
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up docker buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build internet identity
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: Dockerfile
+          build-args: |
+            USE_DUMMY_CAPTCHA=1
+            USE_DUMMY_AUTH=1
+            II_ENV=development
+          cache-from: type=gha,scope=cached-stage
+          # Exports the artefacts from the final stage
+          outputs: ./out
+
+      - run: sha256sum out/internet_identity.wasm
+      - name: 'Upload dummy auth wasm'
+        uses: actions/upload-artifact@v2
+        with:
+          name: internet_identity_dummy_auth.wasm
+          path: out/internet_identity.wasm
+
   #####################
   # The backend tests #
   #####################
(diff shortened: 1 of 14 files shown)


dummy-auth-tests/wdio.conf.ts
```diff
@@ -0,0 +1,36 @@
+import { existsSync, mkdirSync } from "fs";
+
+console.log(process.env);
+
+export const config: WebdriverIO.Config = {
```
lmuntanerDfinity, line 5: Is `WebdriverIO` a global variable?

dummy-auth-tests/wdio.conf.ts
```diff
@@ -0,0 +1,36 @@
+import { existsSync, mkdirSync } from "fs";
+
+console.log(process.env);
```
lmuntanerDfinity, line 3: I guess you plan on removing this log?

dummy-auth-tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+// npm run start # start the proxy
+if (true) {
+  /*
+   * Start the tests
+   */
+  const wdio = child_process.spawn("npm", ["run", "wdio"], {
+    env: { ...process.env, II_DAPP_URL: II_DAPP_URL },
+  });
+
+  wdio.stdout.on("data", (data) => {
```
lmuntanerDfinity, line 86: why not also `stderr`?

dummy-auth-tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+ */
+const II_DAPP_PORT = 8086;
+const II_DAPP_URL = `http://localhost:${II_DAPP_PORT}`;
+
+const child_process = require("child_process");
+
+const proxy = child_process.spawn("proxy", [
+  "--replica-host",
+  replica_host,
+  `${canister_id}:${II_DAPP_PORT}`,
```
lmuntanerDfinity, line 60: Why this? Does this look like `xxxxx-xxxxx...:8086`?

dummy-auth-tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+ * Any port would do here, it just needs to be the same the test runner uses,
+ * hence we set it as the `II_DAPP_URL` environment variable which is read by
+ * wdio.conf.js..
+ */
+const II_DAPP_PORT = 8086;
+const II_DAPP_URL = `http://localhost:${II_DAPP_PORT}`;
+
+const child_process = require("child_process");
+
+const proxy = child_process.spawn("proxy", [
```
lmuntanerDfinity, line 57: This proxy is then forwarding the requests to a running replica?

mstrasinskis (commented):
JS LGTM

.github/workflows/canister-tests.yml
```diff
@@ -252,6 +282,60 @@ jobs:
...
+
+      - name: Run tests
+        working-directory: dummy-auth-tests
+        run: |
+          npm ci
+          npm run test
+
+      - name: Stop replica
+        run: |
+          dfx stop
```
bitdivine, line 337: Just out of curiosity, does the job keep running as long as dfx is running?  I would have expected the VM to be killed regardless.  This is not a change request.  Having a stop certainly doesn't harm, is safe and good.  I'm seeking more insights into how GH actions work.

dummy-auth-tests/wdio.conf.ts
nmattia: Injected by the test runner (`wdio`)

dummy-auth-tests/wdio.conf.ts
nmattia: So THAT's where that came from 🤣 thanks!

dummy-auth-tests/test.js
nmattia: yes, exactly! see https://github.com/nmattia/dfx-proxy#usage

dummy-auth-tests/test.js
nmattia: yep that's it!

mstrasinskis (approved):

lmuntanerDfinity (approved):

lmuntanerDfinity:
Thanks for the explanations @nmattia !

PR #539, opened 2022-02-11
Simplify docker build

nmattia:
As per @bitdivine's suggestion this uses `buildx` to perform the build,
meaning we can actually set `--output`. I tried building with `docker
buildx build` but got errors locally. This seems to work.


frederikrothenberger (approved):
LGTM

PR #547, opened 2022-02-15
Upload screenshots on branch runs

nmattia:
We recently made the workflows run on all pulls and not only PRs. What I
missed was that the uploader on runs on "PR" runs.

This change ensures that the screenshots are correctly uploaded.


peterpeterparker (approved):
LGTM

PR #550, opened 2022-02-15
Clean up and document build flavors

nmattia:
This homogenizes the build "flavors" by prefixing everything with `II_` (instead of `USE_`) and renaming the `II_ENV` to `II_FETCH_ROOT_KEY` which clarifies the intent. I've also added some documentation for each build flavor and cleaned up the docker build script.


frederikrothenberger (approved):
LGTM!

nmattia:
Official build sha256 before: d1f4eef1f2c93c162b947935f97be120ac8abb037cb4512368d66cfe7b703031  https://github.com/dfinity/internet-identity/runs/5202484579?check_suite_focus=true#step:5:4
Official build sha256 after: d1f4eef1f2c93c162b947935f97be120ac8abb037cb4512368d66cfe7b703031 https://github.com/dfinity/internet-identity/runs/5203461573?check_suite_focus=true#step:5:4

(because webpack simply removes the build time variables leaving us with the same code as before)

PR #551, opened 2022-02-15
Add build flavors warning

nmattia:
This adds a red banner when any of the custom build flavors are enabled, i.e. only the official build does not have the banner. The banner serves as a guard so that users know they are not using the real Internet Identity.

I've removed the banner from screenshots; except for one, to make sure we keep track of what the banner looks like.

---

Screenshot 2022-02-16 at 12 02 00


src/frontend/src/flavors.ts
```diff
@@ -0,0 +1,83 @@
...
+        color: ${white};
+      }
+
+      .flavors-warning-btn:hover,
+      .flavors-warning-btn:focus {
+        background: ${white};
+        color: ${red};
+      }
+    </style>
+    This Internet Identity is insecure because it was built with flavors
```
frederikrothenberger, line 60: I would change the wording to:
"This is an __insecure__ development version of Internet Identity. Do not use in production!"
And then have the links.

src/frontend/src/flavors.ts
```diff
@@ -0,0 +1,83 @@
...
+    <a
+      class="flavors-warning-btn"
+      target="_blank"
+      rel="noopener noreferrer"
+      href="https://github.com/dfinity/internet-identity#build-flavors"
+      >more</a
+    >
+    <a class="flavors-warning-btn flavors-warning-btn-close" href="#">close</a>
```
frederikrothenberger, line 68: The banner should not be closable IMHO.

src/frontend/src/flavors.ts
nmattia: I think "Do not use in production!" is a bit too late, right? Whomever sees this probably isn't the person deploying II

src/frontend/src/flavors.ts
nmattia: That's what I thought at first, then realized it might get be very annoying because some content might get hidden behind it

src/frontend/src/flavors.ts
nmattia: Addressed; left out the "do not use in production" bit

src/frontend/src/flavors.ts
frederikrothenberger: How about making the banner not overlapping but just at the top?

nmattia: Removed the close button! And moved the banner to the top of the page as opposed to overlaying the content

frederikrothenberger (approved):
Nit: I think `featureFlags` or `buildFlags` would have communicated the intent better than `flavors`.

PR #553, opened 2022-02-16
Add mainnet test canister id to canister_ids.json

frederikrothenberger:
Add canister ID for mainnet test canister.


nmattia (commented):
It's a bit confusing to have both `ic` and `mainnet`; also would be great to have details somewhere on which one to use when. Is `mainnet` every used?

canister_ids.json
```diff
@@ -1,6 +1,5 @@
 {
   "internet_identity": {
-    "identity": "rdmx6-jaaaa-aaaaa-aaadq-cai",
-    "mainnet": "rdmx6-jaaaa-aaaaa-aaadq-cai"
+    "ic": "fgte5-ciaaa-aaaad-aaatq-cai"
```
bitdivine, line 3: To match the environments as defined by inf it might be worth calling this mainnet.  See e.g. the list of environments here:  https://gitlab.com/dfinity-lab/private/ic-stopgap/-/tree/master/boundary_node/env  Mainnet is also the term used in nns-dapp and extensively in the main ic-repo.  I realize it might seem annoying but the consistency will probably pay off.

bitdivine: BTW, is there a reason for removing the identity environment?  I would quite like to add nnsdapp as I have to add that every time I deploy the nnsdapp testnet, and it uses a different canister ID.

dfx.json
```diff
@@ -16,14 +16,6 @@
     "local": {
       "bind": "0.0.0.0:8000",
       "type": "ephemeral"
```
bitdivine, line 18: What is the motivation behind removing these?

bitdivine (approved):
I don't mind this PR but I don't understand the motivation either.  Please proceed if you would find it useful, we can always adapt this again if needed, but I would consider leaving in the dfx.json.

canister_ids.json
frederikrothenberger: As discussed, kept identity testnet values and added nnsdapp values.

PR #558, opened 2022-02-25
L2-8: New Device Add Flow

frederikrothenberger:
New two-step device addition flow which makes adding new devices more convenient, easier to understand and more secure.


nmattia (commented):
Thanks a lot for this! I love the flow and the new pages look really nice. I've had a look mostly at the rust code so far; sending some feedback already before I tackle the frontend code. The main point is about the datatypes used here; I think it might be worth rethinking how we store the inflight device registrations, for legibility, security and correctness.

src/internet_identity/src/main.rs
```diff
@@ -255,6 +329,206 @@ async fn init_salt() {
...
+    })
+}
+
+async fn generate_device_verification_code() -> DeviceVerificationCode {
+    let res: Vec<u8> = match call(Principal::management_canister(), "raw_rand", ()).await {
+        Ok((res,)) => res,
+        Err((_, err)) => trap(&format!("failed to get randomness: {}", err)),
+    };
+    let rand = u32::from_be_bytes(res[..4].try_into().unwrap());
+    (rand % 1_000_000).to_string()
```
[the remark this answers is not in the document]

nmattia (commented):
A note about the naming. In general, concise (but clear) names are ideal. A concise name means the person reading the code can know what they're looking at without spending cycles actually reading the name. Also, the longer the name, the harder it is to differentiate two functions or variables:
* `check_tentative_device_code_counting_failed_attempts`
* `check_tentative_device_registration_prerequisites`

That being said, different projects (and in my experience, languages) tend to have different notions of what is "the right size". In general we should really strive for consistency. In the project so far, names were kept fairly short and concise, and I don't believe that came at the cost of clarify -- although I know there was a time cost when coming up with the names.

In this PR, there's a very different trend; this PR introduces 6 of the 7 longest function names:

```
 check_tentative_device_code_counting_failed_attempts
 check_tentative_device_registration_prerequisites
 prune_expired_tentative_device_registrations
 generate_device_verification_code
 __get_candid_interface_tmp_hack
 enter_device_registration_mode
 exit_device_registration_mode
```

<details>
<summary>Full list, not very scientific</summary>
<pre><code>
~/internet-identity$ cat src/internet_identity/src/main.rs | grep fn | sed 's/.*fn//' | sed 's/<.*//' | sed 's/(.*//' | awk '{ print length, $0 }' | sort -n -r | cut -d" " -f2-
 check_tentative_device_code_counting_failed_attempts
 check_tentative_device_registration_prerequisites
 prune_expired_tentative_device_registrations
 generate_device_verification_code
 __get_candid_interface_tmp_hack
 enter_device_registration_mode
 exit_device_registration_mode
 delegation_signature_msg_hash
 der_encode_canister_sig_key
 trap_if_not_authenticated
 prune_expired_signatures
 verify_tentative_device
 check_frontend_length
 add_tentative_device
 check_proof_of_work
 check_proof_of_work
 prepare_delegation
 check_entry_limits
 check_nonce_cache
 check_nonce_cache
 update_root_hash
 create_challenge
 get_anchor_info
 ensure_salt_set
 check_challenge
 get_delegation
 create_captcha
 create_captcha
 calculate_seed
 secs_to_nanos
 retrieve_data
 random_string
 get_signature
 get_principal
 add_signature
 http_request
 init_salt
 register
 make_rng
 default
 remove
 lookup
 stats
 main
 init
 from
 from
 add
</code></pre>
</details>

I think it's worth taking a second to wonder why that is, and if we could do something in the direction of consistency!

backend-tests/backend-tests.hs
```diff
@@ -481,6 +484,67 @@ tests wasm_file = testGroup "Tests" $ upgradeGroups $
...
+    lift $ anchor_info .! #devices @?= V.fromList [device1]
+    lift $ anchor_info .! #tentative_device @?= Just device2
+    lift $ anchor_info .! #device_registration_mode_expiration @?= Just expiration
+
+  , withoutUpgrade $ iiTest "reject tentative device if not in registration mode" $ \cid -> do
+    user_number <- register cid webauth1ID device1 >>= mustGetUserNumber
+    V.IsJust (V.Label :: Label "device_registration_mode_off") _ <- callII cid anonymousID #add_tentative_device (user_number, device2)
+    pure ()
+
+  , withoutUpgrade $ iiTest "reject tentative device if registration mode is disabled again" $ \cid -> do
```
nmattia, line 519:
```suggestion
  , withoutUpgrade $ iiTest "reject tentative device after leaving registration mode" $ \cid -> do
```

backend-tests/backend-tests.hs
```diff
@@ -481,6 +484,67 @@ tests wasm_file = testGroup "Tests" $ upgradeGroups $
...
+    user_number <- register cid webauth1ID device1 >>= mustGetUserNumber
+    _ <- callII cid webauth1ID #enter_device_registration_mode user_number
+    _ <- callII cid webauth1ID #exit_device_registration_mode user_number
+    V.IsJust (V.Label :: Label "device_registration_mode_off") _ <- callII cid anonymousID #add_tentative_device (user_number, device2)
+    pure ()
+ 
+  , withoutUpgrade $ iiTest "reject tentative device if registration mode is expired" $ \cid -> do
```
nmattia, line 526:
```suggestion
  , withoutUpgrade $ iiTest "reject device verification if registration mode is expired" $ \cid -> do
```

src/internet_identity/src/main.rs
```diff
@@ -244,6 +303,211 @@ async fn init_salt() {
...
+                    user_number, err
+                ))
+            });
+        trap_if_not_authenticated(entries.iter().map(|e| &e.pubkey));
+
+        prune_expired_tentative_device_registrations(state);
+
+        let mut device_registration_state = state.tentative_device_registrations.borrow_mut();
+        match device_registration_state.remove(&user_number) {
+            None | Some(DeviceRegistrationModeActive { .. }) => Err(WrongCode { retries_left: 0 }),
```
nmattia, line 434: Was looking at the backend tests and got surprised by this. Do we reply with "wrong code" for security reasons? If so can you add a note?

backend-tests/backend-tests.hs
```diff
@@ -481,6 +484,67 @@ tests wasm_file = testGroup "Tests" $ upgradeGroups $
...
+    user_number <- register cid webauth1ID device1 >>= mustGetUserNumber
+    expiration <- callII cid webauth1ID #enter_device_registration_mode user_number
+    V.IsJust (V.Label :: Label "added_tentatively") response <- callII cid anonymousID #add_tentative_device (user_number, device2)
+    lift $ expiration @?= response .! #device_registration_timeout
+    setCanisterTimeTo cid (fromIntegral expiration) -- registration mode is expired at time expiration
+    V.IsJust (V.Label :: Label "wrong_code") result <- callII cid webauth1ID #verify_tentative_device (user_number, response .! #verification_code)
+    lift $ result .! #retries_left @?= 0
+
+  , withoutUpgrade $ iiTest "reject wrong code on verify remote device" $ \cid -> do
```
nmattia, line 535: nit: we could verify that:
* The device can be verified
* The device can be verified after N retries, with N smaller than MAX_ATTEMPS

src/frontend/src/flows/addDevice/manage/addLocalDevice.ts
```diff
@@ -0,0 +1,70 @@
+// Add a new device (i.e. a device connected to the device the user is
```
nmattia, line 1:
```suggestion
// Add a new device (i.e. a device connected to the browser the user is
```

src/frontend/src/flows/addDevice/manage/addLocalDevice.ts
```diff
@@ -0,0 +1,70 @@
...
+    });
+  } catch (error: unknown) {
+    await displayFailedToAddNewDevice(
+      error instanceof Error ? error : unknownError()
+    );
+    return renderManage(userNumber, connection);
+  }
+  const deviceName = await pickDeviceAlias();
+  if (deviceName === null) {
+    return renderManage(userNumber, connection);
```
nmattia, line 47:
```suggestion
    // user clicked "cancel", so we go back to "manage"
    return renderManage(userNumber, connection);
```

src/frontend/src/flows/addDevice/manage/addLocalDevice.ts
```diff
@@ -0,0 +1,70 @@
...
+        deviceName,
+        { unknown: null },
+        { authentication: null },
+        newDevice.getPublicKey().toDer(),
+        newDevice.rawId
+      )
+    );
+  } catch (error: unknown) {
+    await displayFailedToAddTheDevice(
+      error instanceof Error ? error : unknownError()
```
nmattia, line 62: how about converting the error to a string if it's not an `Error`?

src/internet_identity/src/main.rs
```diff
@@ -1,20 +1,30 @@
...
-use internet_identity::signature_map::SignatureMap;
 use rand_chacha::rand_core::{RngCore, SeedableRng};
 use serde::Serialize;
 use serde_bytes::ByteBuf;
-use std::cell::{Cell, RefCell};
-use std::collections::HashMap;
-use std::convert::TryInto;
+use AddTentativeDeviceResponse::DeviceRegistrationModeOff;
+
+use crate::AddTentativeDeviceResponse::AnotherDeviceTentativelyAdded;
```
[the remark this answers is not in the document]
frederikrothenberger: This shortens the usage of `AddTentativeDeviceResponse::AnotherDeviceTentativelyAdded` to `AnotherDeviceTentativelyAdded`.

src/internet_identity/src/main.rs
```diff
@@ -533,6 +797,57 @@ fn lookup(user_number: UserNumber) -> Vec<DeviceData> {
     })
 }
 
+#[update] // this is an update call because queries are not (yet) certified
+fn get_anchor_info(user_number: UserNumber) -> IdentityAnchorInfo {
```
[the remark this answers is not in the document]
frederikrothenberger: So, as we agreed:
* lookup will be deprecated
* get_anchor_info will be kept as is and requires authentication and gives you all the anchor information
* get_credential_ids will be introduced in another PR and will replace the current usages of lookup.

src/internet_identity/src/main.rs
frederikrothenberger: Yes, we discussed this earlier. But actually, I just realized that it is useless as this is an authenticated call. I'll revert to giving descriptive errors.

src/frontend/src/flows/addDevice/manage/addLocalDevice.ts
nmattia: Makes sense! We can always revisit if we see unknown errors

src/frontend/src/flows/addDevice/manage/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    .flowOption:focus {
+      outline: none;
+      box-shadow: none;
+    }
+    .flowIcon {
+      height: 52px;
+    }
+    .flowOptionTitle {
+      font-weight: 500;
+      margin: 0.5rem;
```
nmattia, line 38: nit: there's a lot of `rem`s in this `<style>`; for margin especially might be interesting to use `em` instead so that the margins are scaled with the font size, if we ever decide to make it bigger or smaller.

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+const pageContent = (userNumber: bigint, endTimestamp: bigint) => html`
+  <div class="container">
+    <h1>Add New Remote Device</h1>
+    <p>
+      Device registration mode enabled for Identity Anchor
+      <b>${userNumber}</b>. Please follow these steps to add your new device:
+    </p>
+    <ol class="instruction-steps">
+      <li>
+        On your <i>new</i> device:<br />
```
nmattia, line 25: nit: there's a lot of `<i>` here and in other files. I think `<em>` is more appropriate since you really want to _emphasize_. Same goes with `<b>` (and `<strong>`).

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+  userNumber: bigint,
+  connection: IIConnection
+): Promise<void> => {
+  await withLoader(async () => {
+    const [timestamp, userInfo] = await Promise.all([
+      connection.enterDeviceRegistrationMode(userNumber),
+      connection.lookupAnchorInfo(userNumber),
+    ]);
+    const tentative_device = getTentativeDevice(userInfo);
+    if (tentative_device !== undefined) {
```
nmattia, line 64:
```suggestion
    if (tentative_device) {
```

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+    await renderManage(userNumber, connection);
+  };
+};
+
+const getTentativeDevice = (
+  userInfo: IdentityAnchorInfo
+): DeviceData | undefined => {
+  if (
+    userInfo.device_registration.length === 1 &&
+    userInfo.device_registration[0].tentative_device.length === 1
```
nmattia, line 129: This length checking is just awkwardness inherited from how agent-js encodes `Opt`s with lists, correct?

src/frontend/src/utils/countdown.ts
```diff
@@ -0,0 +1,78 @@
+import { render } from "lit-html";
```
nmattia, line 1: Some doc would be much appreciated here! Esp. re. usage.

src/frontend/src/utils/countdown.ts
```diff
@@ -0,0 +1,78 @@
...
+      Math.max(0, this.interval - drift)
+    );
+  }
+}
+
+export const setupCountdown = (
+  timestamp: bigint,
+  continueFunc: () => Promise<void>
+): Countdown => {
+  const timer = document.getElementById("timer") as HTMLElement;
```
nmattia, line 44: How about passing the element as an argument instead? Using global IDs like this is really brittle/dangerous, especially given the amount of usage doc in this file :)))

Alternatively, a standalone [component](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) might encapsulate the logic better

src/frontend/src/utils/countdown.ts
```diff
@@ -0,0 +1,78 @@
...
+  return new Countdown(
+    () => render(formatRemainingTime(timestamp), timer),
+    1000,
+    async () => {
+      await displayError({
+        title: "Timeout Reached",
+        message:
+          'The timeout has been reached. For security reasons the "add device" process has been aborted.',
+        primaryButton: "Ok",
+      });
```
nmattia, line 54: The behavior on "error" should probably be customizable; it's really odd having a "util" module that references a feature like "add device" directly

src/frontend/src/utils/countdown.ts
```diff
@@ -0,0 +1,78 @@
...
+import { displayError } from "../components/displayError";
+
+export class Countdown {
+  private timeoutHandle;
+  private expected;
+  private readonly endTimestamp;
+
+  constructor(
+    private updateFunc: () => void,
+    private interval: number,
```
nmattia, line 11: is this ever not `1000`?

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+        On your <i>new</i> device:<br />
+        Choose an alias for your new device
+      </li>
+    </ol>
+    <p>
+      This page will automatically refresh after completing the above steps.
+    </p>
+    <p>
+      Time remaining:
+      <span id="timer">${formatRemainingTime(endTimestamp)}</span>
```
nmattia, line 48: let's avoid the id if we can (don't forget to update main.css)

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+const startPolling = (
+  connection: IIConnection,
+  userNumber: bigint,
+  timerUpdate: Countdown,
+  endTimestamp: bigint
+): number => {
+  const pollingHandle = window.setInterval(async () => {
+    const userInfo = await connection.lookupAnchorInfo(userNumber);
+    const tentative_device = getTentativeDevice(userInfo);
+    if (tentative_device !== undefined) {
```
nmattia, line 84:
```suggestion
    if (tentative_device) {
```

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+  return pollingHandle;
+};
+
+const init = (
+  userNumber: bigint,
+  endTimestamp: bigint,
+  connection: IIConnection
+) => {
+  const countdown = setupCountdown(endTimestamp, () =>
+    renderManage(userNumber, connection)
```
nmattia, line 104: is the (polling) interval cleared if the countdown reaches zero?

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+    <button id="cancelAddRemoteDevice" class="linkStyle">Cancel</button>
+  </div>
+`;
+
+export const pollForTentativeDevice = async (
+  userNumber: bigint,
+  connection: IIConnection
+): Promise<void> => {
+  await withLoader(async () => {
+    const [timestamp, userInfo] = await Promise.all([
```
nmattia, line 59: Note: some funky stuff would happen if the browser's clock is slower than the canister's. Since you're anyway polling the canister every 2 seconds, why not return how much time is left and use that as the displayed value (+ some less reliable countdown in between requests) ? Then you'd be sure there'd be no drift at all

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+};
+
+const startPolling = (
+  connection: IIConnection,
+  userNumber: bigint,
+  timerUpdate: Countdown,
+  endTimestamp: bigint
+): number => {
+  const pollingHandle = window.setInterval(async () => {
```
nmattia, line 81: As far as I can tell you'll have several simultaneous inflight requests if the update call to "enter registration mode" takes longer than 2 seconds, which is very likely. Would it make sense to wait for the result of a request before issuing the next one? That'll also reduce the load on the canister.

nmattia (commented):
In general it would be really nice to have a line or two explaining what modules do, esp. the logic-heavy modules that handing polling and counting down!

Also any idea why the time is not shown in this screenshot? `screenshots/09-new-device-instructions-mobile.png`

src/frontend/src/test-e2e/views.ts
```diff
@@ -153,7 +153,7 @@ export class MainView extends View {
   async waitForDisplay(): Promise<void> {
     await this.browser
       .$("//h1[string()='Anchor Management']")
-      .waitForDisplayed({ timeout: 5_000 });
+      .waitForDisplayed({ timeout: 10_000 });
```
nmattia, line 156: did things get much slower?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -73,10 +76,20 @@ if (typeof canisterId !== undefined) {
 }
 
 export const canisterIdPrincipal: Principal = Principal.fromText(canisterId);
-export const baseActor = Actor.createActor<_SERVICE>(internet_identity_idl, {
-  agent: new HttpAgent({}),
-  canisterId,
-});
+
+async function createBaseActor(): Promise<ActorSubclass<_SERVICE>> {
```
nmattia, line 80: as far as I can tell, the only difference between this and `createActor()` (line 349) is that is basically `createActor(null)`; is there a reason for having both? Did I miss a subtlety?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -283,7 +296,8 @@ export class IIConnection {
   }
 
   static async lookupAll(userNumber: UserNumber): Promise<DeviceData[]> {
-    return await baseActor.lookup(userNumber);
+    const actor = await baseActor;
+    return await actor.lookup(userNumber);
```
nmattia, line 300: Does this make a difference?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -296,14 +310,36 @@ export class IIConnection {
   static async lookupAuthenticators(
     userNumber: UserNumber
   ): Promise<DeviceData[]> {
-    const allDevices = await baseActor.lookup(userNumber);
+    const actor = await baseActor;
+    const allDevices = await actor.lookup(userNumber);
```
nmattia, line 314: Does this make a difference?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -296,14 +310,36 @@ export class IIConnection {
...
+        : [],
+      key_type: keyType,
+      purpose,
+    });
+  }
+
   static async lookupRecovery(userNumber: UserNumber): Promise<DeviceData[]> {
-    const allDevices = await baseActor.lookup(userNumber);
+    const actor = await baseActor;
+    const allDevices = await actor.lookup(userNumber);
```
nmattia, line 342: Does this make a difference?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -345,6 +381,35 @@ export class IIConnection {
     return this.actor;
   }
 
+  lookupAnchorInfo = async (
```
nmattia, line 384: shouldn't this be `getAnchorInfo` for consistency with the canister method naming?

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
frederikrothenberger: yes, exactly.

src/frontend/src/test-e2e/views.ts
frederikrothenberger: A bit, yes. This is because we switched from a query call to an update call.

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: No, you're right. I initially stumbled across this problem, because the baseActor did not fetch root keys correctly. But the fix I made was of course totally redundant to the existing createActor() function.

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: See above. This did not fetch root keys correctly (which does not matter for queries, but it does for updates. And for consistency I think it should be fixed everywhere.

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: See above.

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: See above.

src/frontend/src/utils/iiConnection.ts
nmattia: I'm confused still. Aren't the two following bits equivalent?

``` ts
return await foo.bar();
```

``` ts
const tmp = await foo;
return tmp.bar();
```

Is there something in `const actor =...` that makes it fetch the root key correctly? I might also just be missing something obvious

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
```diff
@@ -0,0 +1,131 @@
...
+  ) as HTMLButtonElement;
+
+  cancelButton.onclick = () => {
+    countdown.stop();
+    // TODO L2-309: do this without reload
+    window.location.reload();
+  };
+};
+
+function credentialIdEqual(
```
nmattia, line 124: note: the [`sign()` function in `multiWebAuthnIdentity.ts`](https://github.com/dfinity/internet-identity/blob/2ae97d1a79b991f7b47aca4b24c8dea6149bb235/src/frontend/src/utils/multiWebAuthnIdentity.ts#L88) does something similar, albeit with a different implementation. Not saying they _should_ be the same, just thought you might want to have a look.

src/frontend/src/utils/countdown.ts
```diff
@@ -0,0 +1,102 @@
...
+    endTimestamp: bigint,
+    private timeoutFunc?: () => Promise<void>
+  ) {
+    this.expected = Date.now() + this.interval;
+    this.endTimestamp = Number(endTimestamp / BigInt("1000000"));
+    this.timeoutHandle = window.setTimeout(() => this.step()); // execute the first step immediately
+  }
+
+  /**
+   * Stops the countdown. The timout callback is not called.
```
nmattia, line 31:
```suggestion
   * Stops the countdown. The timeout callback is not called.
```

src/frontend/src/utils/countdown.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @param timeoutFunc Optional callback function to be called when the timeout is reached.
+ * @return the initialized {@link Countdown}
+ */
+export const setupCountdown = (
+  endTimestamp: bigint,
+  timerElement: HTMLElement,
+  timeoutFunc?: () => Promise<void>
+): Countdown => {
+  return new Countdown(
+    () => render(formatRemainingTime(endTimestamp), timerElement),
```
nmattia, line 79: nit: this performs an actual DOM update every second, which might be a _bit_ slow, and is quite annoying when you have the dev tools open. Have you considered using a CSS variable and with the [CSS content attribute](https://www.w3schools.com/cssref/pr_gen_content.asp)? It might work though my CSS knowledge isn't great

nmattia (approved):
This is looking really good! Few more comments but I think we're almost there

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
frederikrothenberger: Hm, I think this implementation is more straight forward than converting to buffer first. I'll keep it as is for now.

src/frontend/src/utils/countdown.ts
frederikrothenberger: Well, mine isn't either. Let's file this under front-end improvements for later.

PR #564, opened 2022-03-01
Fail CI on bad ts formatting

nmattia:
This fails the CI checks if the ts code isn't properly formatted. Before
this, the CI would commit the formatted files. This was annoying
because:
* It would push to the branch, meaning developers had to pull or
  force-push
* The commit would not trigger a CI run due to GHA limitations


frederikrothenberger (approved):
LGTM

PR #569, opened 2022-03-03
Run audit and update lockfiles

nmattia:
This runs `npm audit fix` in the `demos/`, and makes sure we use a
recent node version everywhere.

Note about the testnet lockfile: it was bumped by node. Looks like node v16 is much more verbose in its lockfiles.

---

EDIT: I've also removed the `setupSeleniumServer()` function that used to set up selenium from jest. This has caused many issues over the past few months where the selenium server couldn't be killed. This completely broke with node 16, as jest just wouldn't return anymore. The setup is now done outside of the test running.


frederikrothenberger (approved):
thanks

PR #570, opened 2022-03-03
Remove Proof of Work

nmattia:
This removes the Proof of Work that was required when generating the
CATPCHA. The idea is that the Proof of work was very expensive to
compute in a browser using javascript (would sometimes take tens of
seconds) but really isn't a great deterrent for someone computing with
low-level code running on a multi-core laptop.

In particular this removes:

* The canister functions related to the proof of work check
* The TypeScript functions computing the proof of work
* The "DUMMY_POW" build flavor and references in the build files
* References to the Proof of work in the doc, README and spec
* Test code (backend tests) and test test helpers (pow.ts)


nomeata:
Yay, removing code!

frederikrothenberger (approved):

PR #572, opened 2022-03-08
Add release script and release on tag push

nmattia:
This adds a script, `scripts/release`, that creates a release based on a tag:

```
~/internet-identity$ ./scripts/release --help
Release Internet Identity

Usage:
  release --tag TAG [-- FILE...]

Options:
  --tag TAG     specify the tag to release

Environment:
  GITHUB_TOKEN  (required) Personal Access Token for GitHub


Any files specified after '--' will be uploaded as release assets.

NOTE: This does not generate release notes. A link will be displayed when the
release is created. Visit that link to write release notes and click "Auto-generate release notes"
to add contributions and commit information.
```

This also adds a `release` job that creates a release (with all three builds generated on CI) when tags starting with `release-` are pushed to GitHub. There are existing GitHub actions allowing us to do that, here are the advantages from a custom script:
* More flexibility when creating the release
* Script can be run locally as opposed to just on CI
* We don't need to review (potentially compiled) code that has access to our github token

Also the script is just two `curl` calls with lots of comments around it.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -87,7 +87,8 @@ jobs:
         with:
           # name is the name used to display and retrieve the artifact
           name: ${{ matrix.name }}
-          # path is the name used as the file to upload, and the created when downloaded
+          # path is the name used as the file to upload and the name of the
+          # downloaded file
           path: ${{ matrix.name }}
 
   #####################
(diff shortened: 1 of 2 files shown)


frederikrothenberger (approved):
LGTM, thanks

PR #574, opened 2022-03-10
Add CI badges for our GHA checks

nmattia:
This way other users can know if the repo because unhealthy, and we can
spot issues faster.

Screenshot 2022-03-10 at 10 04 04


frederikrothenberger (approved):
Thanks

PR #577, opened 2022-03-15
Update specification for new device flow

frederikrothenberger:
Update documentation to explain the new add device flow.


docs/internet-identity-spec.adoc
```diff
@@ -179,53 +179,9 @@ This section describes the interface that the backend canister provides.
...
-  register : (DeviceData) -> (UserNumber);
-  add : (UserNumber, DeviceData) -> ();
-  remove : (UserNumber, DeviceKey) -> ();
-  lookup : (UserNumber) -> (vec DeviceData) query;
-  get_principal : (UserNumber, FrontendHostname) -> (principal) query;
-
-  prepare_delegation : (UserNumber, FrontendHostname, SessionKey, maxTimeToLive : opt nat64) -> (UserKey, Timestamp);
-  get_delegation: (UserNumber, FrontendHostname, SessionKey, Timestamp) -> (GetDelegationResponse) query;
-}
+include::../src/internet_identity/internet_identity.did[]
```
nmattia, line 184: nice

docs/internet-identity-spec.adoc
```diff
@@ -260,12 +216,44 @@ It is the responsibility of the frontend UI to protect the user from doing these
...
+
+Tentatively adds a new device to the supplied identity anchor and returns a verification code. This code has to be used with the `verify_tentative_device` method to verify this device. If the flow is aborted or not completed within 15 minutes, the tentative device is discarded.
+
+Tentatively added devices cannot be used to login into the management view or authorize authentications for other dApps.
+
+*Authorization*: Anyone can call this
+
+=== The `verify_tentative_device` method
+
+This method, if called with a valid verification code, adds the tentative device to the user permanently and exits registration mode.
```
nmattia, line 241:
```suggestion
For an anchor in device registration mode: if called with a valid verification code, adds the tentative device as a regular device to the anchor and exits registration mode. The registration flow is aborted if this method is called five times with invalid verification codes.

Returns an error if called for a device not in registration mode.
```

docs/internet-identity-spec.adoc
```diff
@@ -260,12 +216,44 @@ It is the responsibility of the frontend UI to protect the user from doing these
...
 === The `lookup` query method
 
-Fetches all data associated with a user.
+Fetches all device data associated with a user.
 
 *Authorization*: Anyone can call this
 
+=== The `get_anchor_info` method
+
+Fetches all data associated with a user including registration mode and tentatively registered devices.
```
nmattia, line 253:
```suggestion
Fetches all data associated with an anchor including registration mode and tentatively registered devices.
```

docs/modules/ic-identity-guide/pages/auth-how-to.adoc
```diff
@@ -102,26 +102,17 @@ If you are adding a new device, such as a new security key, or a new browser pro
...
-. In the browser on the computer that has already been authorized, open the above link, enter your Identity Anchor, click *Authenticate* and authenticate using an existing authentication method.
-. Link your phone to your identity. 
-+
-If you’re sure that the link you pasted in the browser came from you, click *Yes, add new device*.
-+
-WARNING: Never add authentication methods with links that were given to you or that you are not sure you can trust. If you add an attacker's device, they will be able to take over this Identity Anchor.
-. Give the device profile a name and click *Add Device*.
-+
-Your phone will be redirected to the login page, and you can now use it with your Identity Anchor to authenticate.
+. On your existing computer: Open the Internet Identity web page and log in.
```
nmattia, line 105: Existing computer versus imaginary phone? 😉

nmattia (approved):
Very nice, thank you!

docs/internet-identity-spec.adoc
nomeata: Hmm, it prevents being able to update the spec independently of the code (e.g. design a feature in one PR and implement it later) or to keep legacy or convenience stuff (e.g. monitoring) in the implementation that's not in the spec. At least that was the reasoning back then, maybe less of a problem now.

nomeata (commented):
I guess the sections “Identity design and data model” and “Internal data model and data structures used” are yet to be updated?

And of course the “The Internet Identity Service frontend” flows

frederikrothenberger:
Thanks for reminding me about the flows in the spec.

> I guess the sections “Identity design and data model” and “Internal data model and data structures used” are yet to be updated?

So far the non-stable data structures are not discussed anywhere. We should probably add that, but not in this PR.

nomeata:
I guess they could still be outlined in the “Identity design and data model”. I think it’s intentional that the _implementation_ of the non-stable data types (e.g. the precise rust declaration) isn’t in the spec, but it would be worth adding the conceptual types.

Also worth pointing that that it’s intentional that an upgrade of the Internet Identity will interrupt the device adding flow.

PR #581, opened 2022-03-15
Remove old screenshots before downloading new ones

nmattia:
This ensures that old screenshots (for instance, deleted) are not
re-uploaded.


nmattia:
Some flakiness going on here, will need to figure out what's happening. Anchors keep going back and forth between 1008 and 1009.

PR #582, opened 2022-03-15
Clarify build flavors section of release

nmattia:
The current text ("Read more in the build flavors section of the
README") doesn't make much sense on its own.


lmuntanerDfinity (approved):
Always in favor of improve the Readmes

PR #583, opened 2022-03-15
Add examples on how to use dev build

nmattia:


demos/using-dev-build/README.md
```diff
@@ -0,0 +1,16 @@
+# Integrating with Internet Identity
```
nmattia, line 1: NOTE: link to this from README once https://github.com/dfinity/internet-identity/pull/580 is merged

PR #585, opened 2022-03-17
Make better use of space in README table

nmattia:
The column names caused the table in the flavors section to look really
odd, where tickmark-columns would take up a lot of space. This inlines
the list of features in the description, making everything much easier
on the eyes.

**before:**

Screenshot 2022-03-17 at 14 50 40

**after:**
Screenshot 2022-03-17 at 14 50 44


lmuntanerDfinity (approved):
👍

PR #584, opened 2022-03-17
New alias screenshot: wait for loader to go

nmattia:
Otherwise the screenshot might be taken too early: https://github.com/dfinity/internet-identity/pull/582/commits/eea037f8be5b19f39e1f5d78390aa652e9d61a06

PR #586, opened 2022-03-17
Fix screenshot upload

nmattia:
This includes a few things:
* Update the `add-and-commit` github action that sometimes didn't upload
* Fix the selenium tests to actually wait for the loader to disappear (I
  had a brainfart in
  https://github.com/dfinity/internet-identity/pull/584 and put the
  waiting code in the wrong spot)
* Rename the screenshots (actually done by CI)


lmuntanerDfinity (approved):
Ah, ok, so it wasn't working well before?

nmattia:
@lmuntanerDfinity the issue doesn't _always_ happen (it's a matter of timing and luck) so I couldn't be sure but just assumed it was fixed

PR #587, opened 2022-03-18
Add examples on how to use dev build

nmattia:
This moves and renames `dummy-auth-build` to `demos/using-dev-build`,
which is a demo on how to use the development build. The demo's README
is updated to give basic build and run instructions. The test runner is
also patched to actually report test failures on CI (throws error).


.github/workflows/canister-tests.yml
```diff
@@ -296,21 +296,28 @@ jobs:
...
+          set -euo pipefail
 
-          dfx canister --no-wallet create --all
-          dfx canister --no-wallet install --mode install internet_identity --argument '(null)'
+          # Create a fake curl so that we use this build's wasm; also create a "witness"
+          # file to make sure the fake curl was run (rm $witness will fail otherwise)
+          curl_dir=$(mktemp -d); witness=$(mktemp)
+          cat > "$curl_dir/curl" << EOF
+          #!/usr/bin/env bash
+          cp ../../internet_identity_dev.wasm ./internet_identity.wasm && touch $witness
```
lmuntanerDfinity, line 309: I didn't know how to do that, this is amazing.

demos/using-dev-build/dfx.json
```diff
@@ -0,0 +1,22 @@
+{
+  "canisters": {
+    "internet_identity": {
+      "type": "custom",
+      "candid": "internet_identity.did",
+      "wasm": "internet_identity.wasm",
+      "build": "curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity_dev.wasm -o internet_identity.wasm"
```
lmuntanerDfinity, line 7: So, it's official this is the best way to get other canisters, I guess I could do the same for Ledger and Governance, right?

lmuntanerDfinity (approved):
LGTM!

demos/using-dev-build/dfx.json
nmattia: I don't know if this is the official best way, but I've created a [ticket](https://github.com/dfinity/sdk/issues/2085) in the SDK repo. Re. Ledger and Governance, yes, I don't see why that wouldn't work!

PR #591, opened 2022-03-22
Add 'Getting Help' section to README

nmattia:
This adds a few useful links so that readers now where to get help. Will
also add a link to the Slack/Discord as soon as I've clarified with
@alexabsmith which one we should use.


frederikrothenberger (approved):
Thanks

PR #595, opened 2022-03-24
Don't allow adding too many devices

### LINKED ISSUE #588: Not getting front-end error when device count limit is reached
Error in console:

```javascript
Uncaught (in promise) Error: Call was rejected:
  Request ID: ...
  Reject code: 5
  Reject text: Canister rdmx6-jaaaa-aaaaa-aaadq-cai trapped explicitly: at most 10 authentication information entries are allowed per user
```
Expected to show a proper UI to frontend, and before adding device not after the process

nmattia:
Fixes #588 

This adds a FE check to make sure that users don't try to add more
devices than the canister will allow.

* Disable "ADD DEVICE" button if 8 authenticators are already registered
  (8 + 2 recovery = 10) and add tooltip on hover
* Show number of devices and total allowed number

I tried to limit the changes to the frontend, as opposed to letting users actually add a device and modify the canister to return a response (as opposed to just trapping). The CSS is a bit awkward (most notably, everything is in `px`) but that seems to be the case everywhere in that file anyway. The whole flex + label mechanism is also a tad odd so I had to shoehorn this in the best I could, and would love to have some experienced FE eyes on this.

https://user-images.githubusercontent.com/6930756/160122449-3adc1f5b-8720-4539-9147-6fd138bac557.mov


src/frontend/src/flows/manage.ts
```diff
@@ -45,16 +54,44 @@ const style = () => html`<style>
     width: auto;
     margin: 0;
     cursor: pointer;
-    color: #387ff7;
+  }
+
+  .labelActionText {
+    float: right;
```
peterpeterparker, line 60: Just FYI: I actually never use `float: right`. No particular reason.

I would probably have use `display: flex` on the container and `justify-content: space-between`.

src/frontend/src/flows/manage.ts
```diff
@@ -75,9 +112,23 @@ const pageContent = (userNumber: bigint, devices: DeviceData[]) => html`
...
     <div class="labelWithAction">
-      <label id="deviceLabel">Added devices</label>
-      <button class="labelAction" id="addAdditionalDevice">
-        ADD NEW DEVICE
+      <label class="labelAddedDevices" id="deviceLabel"
+        ><span class="tooltip"
+          >You can register up to ${MAX_AUTHENTICATORS} authenticator devices
+          (recovery devices excluded)</span
+        >Added devices
+        (${numAuthenticators(devices)}/${MAX_AUTHENTICATORS})</label
```
peterpeterparker, line 120: amount of data is small so perf are good but `numAuthenticators` is iterated twice. just in case you rather like to use a state that is assigned once.

src/frontend/src/flows/manage.ts
```diff
@@ -87,7 +138,7 @@ const pageContent = (userNumber: bigint, devices: DeviceData[]) => html`
           <div class="labelWithAction">
             <label id="deviceLabel">Recovery mechanisms</label>
             <button class="labelAction" id="addRecovery">
```
peterpeterparker, line 140: you shouldn't disable the recovery button if there is already two recovery devices?

peterpeterparker:
Visually I noticed two things, if it can be interesting.

Capture d’écran 2022-03-24 à 14 52 52

peterpeterparker (approved):
Just a couple of NTH comments. LGTM, thx

src/frontend/src/flows/manage.ts
nmattia: Done! thanks

src/frontend/src/flows/manage.ts
```diff
@@ -45,16 +54,44 @@ const style = () => html`<style>
...
     content: "+";
     margin-right: 3px;
-    color: #387ff7;
+  }
+
+  .labelAction:disabled .labelActionText {
+    color: var(--grey-100);
+  }
+
+  .tooltip {
```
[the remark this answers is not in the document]
nmattia: Looks really nice! Thank you

src/frontend/src/flows/manage.ts
nmattia: Right, but I couldn't figure out a way to save that in a variable without turning the whole `html` bit into a full blown function with body; that'll do

src/frontend/src/flows/manage.ts
nmattia: I just noticed an issue with the recovery bit; will address separately

nmattia:
Thanks for the review! I've made the `(1/1)` part smaller, and I've switched to `flex` so that everything aligns nicely.

nmattia:
Ok, screenshots show some issue, esp on mobile. Will have another look.

frederikrothenberger (approved):
LGTM Thanks!

PR #604, opened 2022-03-25
Fix hasRecoveryDevice

nmattia:
The name was wrong; it did the exact opposite of what the name
suggested.


frederikrothenberger (approved):
👍

PR #605, opened 2022-03-25
Use invite link for Discord

nmattia:
The previous link linked to the channel directly, which would break
discord in case the user hadn't already been invited to the channel.

Courtesy of GLdev and Sea Snake on discord.


peterpeterparker (approved):
Bravo

PR #612, opened 2022-04-04
Improve release sha table

nmattia:
This makes the sha table a little less wide by linking to the CI Run
from the sha itself, instead of having a column that adds no information
except for the link.


frederikrothenberger (approved):
LGTM

PR #613, opened 2022-04-04
Show reason if feature check fails

nmattia:
This displays a reason for the feature check (compatibility check) in
the compatibility notice, which should help users understand why their
browser is not supported.


src/frontend/src/utils/featureDetection.ts
```diff
@@ -1,11 +1,17 @@
-export const hasRequiredFeatures = async (url: URL): Promise<boolean> => {
-  if (window.PublicKeyCredential === undefined) return false;
-  if (url.hash === "#compatibilityNotice") return false;
+export const checkRequiredFeatures = async (
+  url: URL
+): Promise<true | string> => {
+  if (window.PublicKeyCredential === undefined)
+    return "window.PublicKeyCredential is not defined";
+  if (url.hash === "#compatibilityNotice")
+    return "Remove #compatibilityNotice from the URL and try again.";
```
frederikrothenberger, line 7: Why do we even have the `#compatibilityNotice`? Do we ever link to that? I'd rather remove the whole thing instead of prompting users to fiddle with the URL. Or at the least provide a button / link to help the user.

nmattia: I'm guessing we're using this to trigger the page in CI to get a screenshot of it

nmattia: Confirmed https://github.com/dfinity/internet-identity/pull/228:

```
commit 48f68c7af2817d5020c9c9080737afd840b19f8d
Author: Kyle Peacock <kylpeacock@gmail.com>
Date:   Fri May 7 05:57:36 2021 -0700

    chore: slightly improved compatibilityNotice (#228)
    
    * chore: slightly improved compatibilityNotice
    Adds mobile info
    Check for browser compatibility checks if hardware can sign without security key for mobile devices
    
    * updating compatibility chart
    
    * extracts feature detection logic
    
    also adds a hash route for directly displaying the compatibility
    notice
    
    * Take a screenshot of the compatibility notice
    
    * whitespace nit
    
    * Updating selenium screenshots
    
    * fix whitespace in selenium tests
    
    Co-authored-by: Christoph <christoph.hegemann1337@gmail.com>
    Co-authored-by: Screenshot Committer <nobody@example.com>
```

frederikrothenberger: Still, why not make the `Remove #compatibilityNotice from the URL and try again.` to a link back to the main page?

nmattia: Because right now we return a `string` from the feature check and returning a link would mean more machinery (because then we need to return HTML). Also no one should ever land on that page. Happy to do it if you feel strongly about it.

frederikrothenberger (approved):

PR #618, opened 2022-04-08
Show banner if origin isn't identity.ic0.app

nmattia:
We should make sure that unsuspecting users do not land on testing
builds of II. While we can't ensure that, we can at least let them know
that something is wrong. So far we've displayed a banner if the build
included features; this adds a banner if the origin is not
identity.ic0.app.


frederikrothenberger (approved):
I think it is good to show the warning for all URLs that the user should _not_ be using (rather than making an exception for https://rdmx6-jaaaa-aaaaa-aaadq-cai.ic0.app just because it is technically the official one).

PR #620, opened 2022-04-12
Add FAQ entry about invalid seedphrase

nmattia:
Closes https://github.com/dfinity/internet-identity/issues/331

[A PR](https://github.com/dfinity/internet-identity/pull/289) purported to actually allow recovering anchors using only the first 4 letters of the words; looking at the implementation it did not. 

Following the trail from #331, it looks like we've decided not to go that route -- with reasons.

Instead, this PR here gives some pointers on what to do when faced with "Invalid
Seedphrase". It also cleans up some documentation that was out of date.


peterpeterparker (approved):
LGTM

PR #622, opened 2022-04-14
More comments in example and expanded readme

frederikrothenberger:
Add more explanations to the using dev build example. Extract some parts of `index.js` to the Readme so that they are visible without leaving https://beta.smartcontracts.org/docs/current/developer-docs/functionality/internet-identity/integrate-identity.


demos/using-dev-build/README.md
```diff
@@ -31,6 +31,41 @@ At this point, the replica (for all practical matters, a local version of the In
...
+DFINITY provides an easy-to-use library to log in with Internet Identity: [agent-js](https://github.com/dfinity/agent-js)
+
+These are the steps required to log in and use the obtained identity for canister calls:
+```
+// First we have to create and AuthClient.
+const authClient = await AuthClient.create();
+
+// Call authClient.login(...) to login with Internet Identity. This will open a new tab
+// with the login prompt. The code has to wait for the login process to complete.
+// We can either use the callback functions directly or wrap the in a promise as
```
lmuntanerDfinity, line 45: Typo: "or wrap it in a promise."

I would also remove the "as we have done here".

demos/using-dev-build/README.md
```diff
@@ -31,6 +31,41 @@ At this point, the replica (for all practical matters, a local version of the In
...
+// we have done here.
+await new Promise((resolve, reject) => {
+  authClient.login({
+    onSuccess: resolve,
+    onError: reject
+  });
+});
+
+// At this point, we're authenticated with II.
```
lmuntanerDfinity, line 54: I would separate this into another code block. Maybe mention before:

"Once we have been authenticated with II we have access to the identity:"

```
// Get the identity from the library
const identity = authClient.getIdentity();
// Using the identity obtained from the auth client, we can create an agent to interact with the IC
const agent = new HttpAgent({ identity });
// Using the interface description of our webapp, we create an Actor that we use to call the service methods
const webapp = Actor.createActor(webapp_idl, {
  agent,
  canisterId: webapp_id
});
// Call whoami which returns the principal (user id) of the current user.
const principal = await webapp.whoami();
```

demos/using-dev-build/webapp/index.js
```diff
@@ -17,38 +17,48 @@ export const init = ({ IDL }) => {
...
+  // First we have to create and AuthClient.
   const authClient = await AuthClient.create();
 
-  // This API is just weird
+  // Find out which URL should be used for login.
+  const iiUrl = document.getElementById("iiUrl").value;
+
+  // Call authClient.login(...) to login with Internet Identity. This will open a new tab
+  // with the login prompt. The code has to wait for the login process to complete.
+  // We can either use the callback functions directly or wrap the in a promise as
```
lmuntanerDfinity, line 41: ditto

demos/using-dev-build/webapp/index.js
```diff
@@ -17,38 +17,48 @@ export const init = ({ IDL }) => {
...
     authClient.login({
       identityProvider: iiUrl,
       onSuccess: resolve,
-      onError: reject,
+      onError: reject
     });
   });
 
-  // At this point, we're authenticated with II and we can query the whoami canister
+  // At this point, we're authenticated with II.
```
lmuntanerDfinity, line 51: Decide whether you want periods on all the comments or on none. 

It seems that you have a mix now.

lmuntanerDfinity (approved):
Since these are subjective comments, I approve and leave it to you to implement them or not.

I like the changes and adding the explanation in the Readme.

demos/using-dev-build/webapp/index.js
```diff
@@ -17,38 +17,47 @@ export const init = ({ IDL }) => {
...
+
+  // Call authClient.login(...) to login with Internet Identity. This will open a new tab
+  // with the login prompt. The code has to wait for the login process to complete.
+  // We can either use the callback functions directly or wrap in a promise.
   await new Promise((resolve, reject) => {
     authClient.login({
       identityProvider: iiUrl,
       onSuccess: resolve,
-      onError: reject,
+      onError: reject
```
nmattia, line 46: please avoid aesthetic changes like removing commas (here and below); in particular, commas on last elements minify the diff: adding an element only adds a line, without modifying the previously-last element

demos/using-dev-build/webapp/index.js
```diff
@@ -17,38 +17,47 @@ export const init = ({ IDL }) => {
...
-  document.getElementById("iiUrl").value = testnetII;
+  let iiUrl;
+  if (process.env.DFX_NETWORK === "local") {
+    iiUrl = `http://localhost:8000/?canisterId=${process.env.II_CANISTER_ID}`;
+  } else if (process.env.DFX_NETWORK === "ic") {
+    iiUrl = `https://${process.env.II_CANISTER_ID}.ic0.app`;
+  } else {
+    iiUrl = `https://${process.env.II_CANISTER_ID}.dfinity.network`;
+  }
+  document.getElementById("iiUrl").value = iiUrl;
```
nmattia, line 28: did anything change here beyond the variable name?

demos/using-dev-build/README.md
```diff
@@ -31,6 +31,42 @@ At this point, the replica (for all practical matters, a local version of the In
 
 If the IC actually lets the call (request) through to the `whoami` canister, it means that everything checked out, and the `whoami` canister just responds with the information the IC adds to requests, namely your identity (principal).
 
+### Using the Auth-Client Library To Log In With Internet Identity
+
+DFINITY provides an easy-to-use library to log in with Internet Identity: [agent-js](https://github.com/dfinity/agent-js)
+
+These are the steps required to log in and use the obtained identity for canister calls:
```
nmattia, line 38: you mention an "identity" here and below, but it's unclear what that actually is. Two paragraphs above we say that "principals are basically the 'identity' [...]" which might lead the reader to believe that the identity you refer to here is a principal (which I don't think is correct).

Note that I don't think this document is the right place to introduce delegations, unless it can be made really simple to understand.

demos/using-dev-build/README.md
```diff
@@ -31,6 +31,42 @@ At this point, the replica (for all practical matters, a local version of the In
 
 If the IC actually lets the call (request) through to the `whoami` canister, it means that everything checked out, and the `whoami` canister just responds with the information the IC adds to requests, namely your identity (principal).
 
+### Using the Auth-Client Library To Log In With Internet Identity
+
+DFINITY provides an easy-to-use library to log in with Internet Identity: [agent-js](https://github.com/dfinity/agent-js)
```
nmattia, line 36: Here and below "(... is available here: [client auth protocol specification](#))" links are given as "this is the document: link". In the rest of the document I tried to include the links in the flow of the text to prevent it looking like a list of links; can something similar be done here?

nmattia (commented):
Few comments; would also be nice to keep the

demos/using-dev-build/README.md
frederikrothenberger: I tried. What do you think?

PR #623, opened 2022-04-19
Don't commit small pixel differences

nmattia:
This makes sure that screenshot differences of 8 pixels or less are not
committed. This works around some flakiness in the screenshot taking,
where the same 8 pixels often change, which might be due to aliasing or
just browser bugs.


frederikrothenberger (approved):
Thanks!

PR #631, opened 2022-05-03
Support WebAuthn Apple Passkeys

dostro:


nmattia:
Hi @dostro, thanks for the contribution! Can you give some context here? I can actually use "iCloud Keychain" with Internet Identity as it is now. Is this what you mean by "WebAuthn Apple Passkeys"?

nmattia:
Note: it looks like [`attestation` used to be set to `direct`](https://github.com/dfinity/internet-identity/pull/338/files#diff-f7c49a174f5c76f4cf2b4b99163ad65f188b0b3694803839b160d7bbf8007ff5L367), but not anymore. I'm going to do some spelunking in the git history and try to figure out what's happened.

PR #633, opened 2022-05-04
Fix wrong TTL in test

frederikrothenberger:
The error was introduced in https://github.com/dfinity/internet-identity/pull/632 where the actual TTL set by the test was not increased.


nmattia (commented):
I'm confused, does the value here actually matter? Since the tests were still green after #632 I'm guessing not that much

frederikrothenberger:
No, but then the test is not testing what it says it does. The test should check that requesting 2 months gets capped at 1 month. Now the test is requesting 1 month and gets 1 month. So the cap is not actually being tested.

nmattia (approved):

PR #635, opened 2022-05-12
Use HTML in FAQ answers

nmattia:
I got frustrated that I couldn't use HTML in the FAQ answers while adding some recovery phrase related information, so I fixed a long-standing issue without our jest setup.

---

This changes the `jest.config.ts` to support importing `lit-html`.

Prior to this, jest would choke on test files that included `lit-html`
(transitively). This prevented us from using HTML in the FAQ questions,
since those questions are used in the test suite.

This includes some other changes:

* Since the questions can now be in a file that imports `lit-html`, they
  are moved to `faq/index.ts` to simplify the file tree,
* The Windows Hello now has a direct (HTML) link, as opposed to only
  using the `links` property of `Question`,
* The `Question` type has added doc,
* The behavior of creating a safe `<a>` tag is extracted in
  `mkExternalLink` and reused where applicable in the FAQ.


src/frontend/src/flows/faq/index.ts
```diff
@@ -1,10 +1,107 @@
...
+  },
+  invalidSeedphrase: {
+    priority: 10,
+    question:
+      'Why do I get "Invalid Seedphrase" when I try to recover my Identity Anchor?',
+    anchor: "invalid-seedphrase",
+    answer: `
+    Most likely, it means that the recovery phrase (also known as seed phrase) is not BIP39 compatible.
+
+    A recovery phrase generated by Internet Identity will contain an anchor number followed by some words. For instance, "135033-squirrel-soccer-...". The number is the anchor for which the recovery phrase is generated; note that Internet Identity will strip the number before trying to recover an Identity, meaning the number is only meant for humans!
```
frederikrothenberger, line 86: >135033-squirrel-soccer-

NIT: Not changed in this PR, but these should be spaces.

frederikrothenberger (approved):

src/frontend/src/flows/faq/index.ts
nmattia: Thanks! #636

PR #641, opened 2022-05-16
Converting II-Spec to Markdown

aisconnolly:
Hi II folks! 
With the new dev portal, we are converting the docs from adoc to md. Of course we would like the reference documentation to live in their home repo. Here I'm pushing a markdown version of the current II Spec. This is the file that will be pulled into the portal repo, so it would be great if further updates could be made in the md version. 
If you've any questions or comments, feel free to shout. Thanks for the patience as we transition to the new docs :)


nmattia (commented):
Hi @aisconnolly!

How about replacing the `.adoc` version with this? The two are bound to get out of sync otherwise!

aisconnolly:
Indeed, it's the hope to replace the adoc. I didn't rm it yet incase anyone wanted to check or compare. But I agree, the sooner the better so they don't get out of sync!

nmattia:
@aisconnolly then go ahead and `rm` it :) We can still find it in the git history if necessary; also, by removing it we can actually compare the diff on the PR!

Out of curiosity, how did you convert the doc from adoc to md? Pandoc or manually?

aisconnolly:
Hi @nmattia! I rm-ed the adoc files now. And yes the conversion was done with asciidoc and pandoc (and some tweaking!)

Just as FYI - we are adding Candid syntax highlighting, so we'll be able to add the .did files as codeblocks soon

aisconnolly:
Thank you thank you! Looks like I still need a review to merge, but then we should be good to go!

bitdivine (approved):

PR #642, opened 2022-05-17
Unify authentication and authorize authentication screens

frederikrothenberger:
Instead of first the usual login screen followed by the approve authorization screen, there is now a single screen unifying both. As a consequence, the recovery nag is now shown after authorizing the login.

Summary of the changes:
* Created the `authenticate` directory within flows
  * `index.ts` for the view
  * `fetchDelegation.ts` for the related interactions with the back-end
  * `postMessageInterface.ts` for code related to the window post messages
* Removed `auth.ts` and `confirmRedirect.ts` which covered logic for the old flow
* Pulled out `registrationLink.ts` from `unknownAnchor.ts` because (some of) the same logic is used in the new flow

This version is deployed here:
* Demo App: https://6zm7l-dyaaa-aaaaa-qaaka-cai.identity.dfinity.network/
* II: https://6cjdo-zaaaa-aaaaa-qaaiq-cai.identity.dfinity.network/


src/frontend/src/styles/main.css
```diff
@@ -250,7 +244,7 @@ main {
 }
 
 .hidden {
-  visibility: hidden;
```
nmattia, line 253: I suspect this will visually break the warning box in the recovery phrase; `display: none` removes the element from the flow, which means that when warnings show up, the page and buttons will move (not ideal). I'm happy to add a screenshot test for that actually

bitdivine (approved):
LGTM, thanks!

src/frontend/src/styles/main.css
frederikrothenberger: Yeah, it also had other visual impacts on the device add flow -> reverted.

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+import { displayError } from "../../components/displayError";
+import { useRecovery } from "../recovery/useRecovery";
+import { initRegistration } from "../../components/registrationLink";
+import waitForAuthRequest, { AuthContext } from "./postMessageInterface";
+import { toggleErrorMessage } from "../../utils/errorHelper";
+import { fetchDelegation } from "./fetchDelegation";
+
+const pageContent = (
+  hostName: string,
+  maxTimeToLive: bigint | undefined,
```
nmattia, line 25:
```suggestion
  maxTimeToLive?: bigint,
```

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+      border-color: var(--grey-050);
+      margin: 1rem 0;
+    }
+
+    .not-displayed {
+      display: none;
+    }
+  </style>
+  <div class="container">
+    <h1>Authorize private login</h1>
```
nmattia, line 108: "private" as opposed to what?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+          Manage your Identity Anchor
+        </button>
+      </div>
+      <hr />
+    </div>
+    ${navbar}
+  </div>
+  ${footer}`;
+
+export class AuthSuccess {
```
nmattia, line 170: Why a class? Maybe we can move this to the "interface" module?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  constructor(
+    public userNumber: bigint,
+    public connection: IIConnection,
+    public sendDelegationMessage: () => void
+  ) {}
+}
+
+/**
+ * Shows the authorize application view and returns information about the authenticated user and a callback, which sends
```
nmattia, line 179: ~~What's the "authorize application view"? I'd assume it's the first page that asks if you authorize the dapp to authenticate, but I don't see any of this happening in the code, so I'm guessing you mean something else?~~

---

Ah, I see. We first wait for the initial request, and _then_ load up the view.

src/frontend/src/flows/authenticate/postMessageInterface.ts
```diff
@@ -21,14 +23,67 @@ export interface AuthResponseSuccess {
...
+  // Send a message to indicate we're ready.
+  // NOTE: Because `window.opener.origin` cannot be accessed, this message
+  // is sent with "*" as the target origin. This is safe as no sensitive
+  // information is being communicated here.
+  if (window.opener !== null) {
+    window.opener.postMessage(READY_MESSAGE, "*");
+  } else {
+    // If there's no `window.opener` a user has manually navigated to "/#authorize".
+    // Signal that there will never be an authentication request incoming.
+    return Promise.resolve(null);
```
nmattia, line 86: Could we start the function with that? i.e. 

``` ts
export default async function waitForAuthRequest(): Promise<AuthContext | null> {
  if(window.opener === null) {
    return Promise.resolve(null);
  }

  ... 
} 
```

at least it'd be clear why we need `| null`, and then you don't have to read the big promise block above to get to that. It's basically "failing early".

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+export const authenticate = async (): Promise<AuthSuccess> => {
+  return new Promise((resolve) => {
+    withLoader(async () => {
+      const authContext = await waitForAuthRequest();
+      if (authContext === null) {
+        // The user has manually navigated to "/#authorize".
+        redirectToWelcomeScreen();
+        return;
+      }
+      resolve(Promise.resolve(init(authContext)));
```
nmattia, line 192: Is the inner `Promise.resolve` necessary? or should this be
``` ts
init(...).then(resolve)
```

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  recoverButton.onclick = () => {
+    const userNumber = readUserNumber(pageMode);
+    if (userNumber !== undefined) {
+      return useRecovery(userNumber);
+    }
+    toggleErrorMessage("userNumberInput", "invalidAnchorMessage", true);
+  };
+};
+
+type PageMode = { mode: "existingUserNumber" | "newUserNumber" };
```
nmattia, line 346:
```suggestion
type PageMode = "existingUserNumber" | "newUserNumber";
```

---

EDIT: actually reading the code, I suspect you're relying on the fact that `mode` is a reference

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  ) as HTMLButtonElement;
+  manageButton.onclick = () => redirectToWelcomeScreen();
+  initRecovery(pageMode);
+
+  const editAnchorButton = document.getElementById(
+    "editAnchorButton"
+  ) as HTMLButtonElement;
+  editAnchorButton.onclick = () => {
+    pageMode.mode = "newUserNumber";
+    setMode(pageMode);
```
nmattia, line 215:
```suggestion
    setMode("newUserNumber");
```


... and others

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,390 @@
...
+    public sendDelegationMessage: () => void
+  ) {}
+}
+
+/**
+ * Shows the authorize application view and returns information about the authenticated user and a callback, which sends
+ * the delegation to the application window. After having received the delegation the application will close the
+ * Internet Identity window.
+ */
+export const authenticate = async (): Promise<AuthSuccess> => {
```
nmattia, line 183: could be an `export default` since it's the only thing you export

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: Yes, exactly. Is this ok?

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: Yes, exactly.

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: This was suggested by Jan as a hint to the per application principals. I think it's meant as private in the sense of privacy-preserving, as opposed to public / linkable.

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    hr {
+      border-color: var(--grey-050);
+      margin: 1rem 0;
+    }
+  </style>
+  <div class="container">
+    <h1>Authorize Authentication</h1>
+    <p class="sectionTitle">Application URL</p>
```
nmattia, line 105: nit: should probably be an h2 or something, same for the other ones

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,355 @@
...
+      <hr />
+    </div>
+    ${navbar}
+  </div>
+  ${footer}`;
+
+const existingAnchorSection = (userNumber: bigint) => html` <div
+  class="modeContainer"
+>
+  <div class="childContainer">
```
nmattia, line 140: nit: the two top-level divs (modeContainer, childContainer) could be lifted out of here. Also is there a way we could reuse the `input` (and maybe `disable` it) and just hide the `invalidAnchorMessage`? all that in order to modify the DOM as little as possible

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,355 @@
...
+ * Shows the authorize application view and returns information about the authenticated user and a callback, which sends
+ * the delegation to the application window. After having received the delegation the application will close the
+ * Internet Identity window.
+ */
+export default async (): Promise<AuthSuccess> => {
+  return new Promise((resolve) => {
+    withLoader(async () => {
+      const authContext = await waitForAuthRequest();
+      if (authContext === null) {
+        // The user has manually navigated to "/#authorize".
```
nmattia, line 179: the promise is never resolved here; it _shouldn't_ matter but just FYI

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,355 @@
...
+          e.preventDefault();
+          authenticateButton.click();
+        }
+      };
+    } else {
+      const editAnchorButton = document.getElementById(
+        "editAnchorButton"
+      ) as HTMLButtonElement;
+      editAnchorButton.onclick = () =>
+        init(authContext, true, userNumber).then(resolve);
```
nmattia, line 222: (paranoia) here there's an assumption that the `userNumber` never changes; but I can see us in the future adding a "tickmark" button that disables editMode, in which case if the user (1) lands on the page (2) edits the anchor (3) clicks the tickmark (4) re-tries to edit the anchor, the anchor will be reverted to the value from (1).

Instead of having state duplicated (once in the input, and once in js) we could use the value in the `input` as source of truth?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,355 @@
...
+): Promise<AuthSuccess> => {
+  displayPage(authContext.requestOrigin, editMode, userNumber);
+  const manageButton = document.getElementById(
+    "manageButton"
+  ) as HTMLButtonElement;
+  manageButton.onclick = () => redirectToWelcomeScreen();
+  initRecovery(editMode);
+
+  return new Promise((resolve) => {
+    const authenticateButton = document.getElementById(
```
nmattia, line 202:
```suggestion
    const authorizeButton = document.getElementById(
```

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,355 @@
...
+        }
+      };
+    } else {
+      const editAnchorButton = document.getElementById(
+        "editAnchorButton"
+      ) as HTMLButtonElement;
+      editAnchorButton.onclick = () =>
+        init(authContext, true, userNumber).then(resolve);
+      authenticateButton.focus();
+      authenticateButton.onkeypress = (e) => {
```
nmattia, line 224: I'm a little unfamiliar with `onkeypress` for anything other than text inputs. What happens here? does focusing the button + hitting enter not trigger "onclick"?

src/frontend/src/components/registrationLink.ts
```diff
@@ -0,0 +1,36 @@
...
+ **/
+function isRegistrationAllowed() {
+  return !/(^https:\/\/rdmx6-jaaaa-aaaaa-aaadq-cai\.ic0\.app$)|(.+\.raw\..+)/.test(
+    window.origin
+  );
+}
+
+export const initRegistration = async (): Promise<LoginFlowResult | null> => {
+  const registerButton = document.getElementById(
+    "registerButton"
```
nmattia, line 28: this is quite dangerous, the function is expecting this element to be on the page but does not advertise this anywhere. how about only exporting the following, and let the caller bind it to any element/action they like?

```
isRegistrationAllowed() ? register() : registerDisabled()
```

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: Yes, it does. I'll remove the `onkeypress` again.

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: I lifted the `modeContainer` at least.

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,363 @@
...
+  try {
+    if (userNumber === undefined) {
+      toggleErrorMessage("userNumberInput", "invalidAnchorMessage", true);
+      return null;
+    }
+    const result = await withLoader(() => IIConnection.login(userNumber));
+    const loginResult = apiResultToLoginFlowResult(result);
+    if (loginResult.tag === "ok") {
+      return await handleAuthSuccess(loginResult, authContext);
+    }
```
nmattia, line 285: nit: shouldn't this also be in `withLoader`?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,363 @@
...
+    if (userNumber !== undefined) {
+      return useRecovery(userNumber);
+    }
+    toggleErrorMessage("userNumberInput", "invalidAnchorMessage", true);
+  };
+};
+
+const redirectToWelcomeScreen = () => {
+  window.location.hash = "";
+  window.location.reload();
```
nmattia, line 348: nit: I would inline those 2 lines, it would make it clear that the promise is not just dropped and forgotten about, but that the page is thrashed proper

src/frontend/src/test-e2e/util.ts
```diff
@@ -176,6 +181,98 @@ export async function addCustomCommands(
       }
     )
   );
+
+  await browser.addCommand(
```
nmattia, line 185:
```suggestion
  // This retrieves previously created credentials, see https://www.w3.org/TR/webauthn-2/#sctn-automation-get-credentials
  await browser.addCommand(
```

I wouldn't think of looking in the object to find the link, wdyt?

nmattia: Out of curiosity, what are all those "variables" and "parameters" and "refs" etc? Are they actually read? Or just used in error messages?

src/frontend/src/test-e2e/util.ts
```diff
@@ -191,6 +288,33 @@ export async function removeVirtualAuthenticator(
...
+    authId,
+    rpId,
+    credential.credentialId,
+    credential.isResidentCredential,
+    credential.privateKey,
+    credential.signCount
+  );
+}
+
+export function originToRpId(origin: string): string {
```
nmattia, line 314: what does RpId stand for?

src/frontend/test-setup.ts
```diff
@@ -12,6 +19,19 @@ declare global {
...
+      addWebauthnCredential: (
+        authenticatorId: string,
+        rpId: string,
+        credentialId: string,
+        isResidentCredential: boolean,
+        privateKey: string,
+        signCount: number,
+        userHandle?: string,
+        largeBlob?: string
+      ) => Promise<void>;
```
nmattia, line 34:
```suggestion
      addWebauthnCredential: (WebAuthnCredential & { rpId: string, userHandle?: string, largeBlob?: string, authenticatorId: string }) => Promise<void>;
```

src/frontend/src/test-e2e/util.ts
frederikrothenberger: Relying Party Id (as per WebAuthn standard).

src/frontend/src/index.ts
```diff
@@ -31,23 +31,27 @@ const init = async () => {
...
   const userIntent = intentFromUrl(url);
+
   switch (userIntent.kind) {
     // Authenticate to a third party service
     case "auth": {
-      return auth(userNumber, connection);
+      // show the application authorization screen. The user can authenticate, create a new anchor or jump to other pages to recover and manage.
+      const authSuccess = await authenticate();
+      // show the recovery wizard before sending the window post message, otherwise the II window will be closed
+      await recoveryWizard(authSuccess.userNumber, authSuccess.connection);
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, `login` is the current welcome screen (for login into management) and authenticate is used for authorizing authentication. I'll do some renaming.

src/frontend/src/test-e2e/util.ts
frederikrothenberger: Yes, parameters and variables are read by the selenium driver. What it essentially does is create a HTTP request out of this information. `description` and `ref` is for humans only, but I figured the additional information might be useful (and also following the existing example).

src/frontend/test-setup.ts
frederikrothenberger: This does not work, as the function takes separate arguments only and not an object.

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,356 @@
...
+    <div class="highlightBox hostName">${hostName}</div>
+    <h2 class="sectionTitle">Identity Anchor</h2>
+    <div class="modeContainer">
+      <div class="childContainer">
+        <input
+          class="anchorText"
+          type="text"
+          id="userNumberInput"
+          placeholder="Enter anchor"
+          value="${userNumber !== undefined ? userNumber?.toString() : ""}"
```
nmattia, line 115: I'm curious, why do you need `?.` and `toString()`? I would assume that ts infers `bigint` (and not `bigint | undefined`) and that `.toString()` would be implicit

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,356 @@
...
+  userNumberInput.onkeypress = (e) => {
+    if (e.key === "Enter") {
+      // authenticate if user hits enter
+      e.preventDefault();
+      authorizeButton.click();
+    }
+  };
+
+  if (userNumber === undefined) {
+    editAnchorButton.classList.toggle("hidden", true);
```
nmattia, line 204: If I understand the doc correctly, using `toggle` with `force = true` or `force = false` is just `add()` and `remove()` respectively?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -0,0 +1,356 @@
...
+    registerSection.classList.toggle("hidden", true);
+    userNumberInput.disabled = true;
+    authorizeButton.focus();
+  }
+  editAnchorButton.onclick = () => {
+    registerSection.classList.toggle("hidden", false);
+    editAnchorButton.classList.toggle("hidden", true);
+    userNumberInput.classList.toggle("highlightBox", false);
+    userNumberInput.disabled = false;
+    userNumberInput.select();
```
nmattia, line 217: ❤️

src/frontend/src/index.ts
```diff
@@ -31,23 +31,27 @@ const init = async () => {
...
-  // one. The exact flow depends on the device they use.
-  await recoveryWizard(userNumber, connection);
-
   const userIntent = intentFromUrl(url);
+
   switch (userIntent.kind) {
     // Authenticate to a third party service
     case "auth": {
-      return auth(userNumber, connection);
+      // show the application 'authorize authentication' screen. The user can authenticate, create a new anchor or jump to other pages to recover and manage.
```
nmattia, line 39: ❤️

nmattia (approved):
Looking really good, thanks for the hard work!

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: Yes, switched them out for clarity.

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: You're right. I think I had to be more explicit in some other context and this is just a left-over artifact of that -> decluttered.

PR #648, opened 2022-05-20
Add trailing commas to prettier config

frederikrothenberger:
Configure prettier to add trailing commas.


nmattia (commented):
Isn't this the [default value](https://prettier.io/docs/en/options.html#trailing-commas)?

frederikrothenberger:
Ah, right.

PR #650, opened 2022-05-20
Split and move auth.ts to flows/authenticate

frederikrothenberger:
Splits and moves the contents related to postMessages and fetching of delegations out of `auth.ts` in preparation for #642.


nmattia (approved):
lgtm!

PR #655, opened 2022-05-30
Remove cubehash

nmattia:
This removes both the canister and javascript implementations of
"cubehash".

I suspect cubehash was used in `certified_map` et al. and wasn't removed
when said crates were removed.


Dfinity-Bjoern (approved):
Yay!

nomeata:
> I suspect cubehash was used in certified_map et al. and wasn't removed

I was used for the “proof of work” captcha, removed in a67392642af192ea196e1050feb0d054697432b6

nmattia:
Ah, thanks Joachim!

PR #660, opened 2022-06-03
Fix invalid anchors showing up as previously used

frederikrothenberger:
II will no longer display invalid anchors as previously used. The user input will be kept after returning to the "authorize authentication" screen after acknowledging the error message.


nmattia (approved):
lgtm!

PR #665, opened 2022-06-07
Revert "Converting II-Spec to Markdown"

nmattia:
Reverts dfinity/internet-identity#641. The markdown does not render well and links are broken. Will fix and resubmit change.


nomeata:
Yeah, the problem with markdown: If you want any non-trivial markup feature, you end up using some tool’s dialect (I guess in this case Docusaurus), and on other tools things look broken…

PR #666, opened 2022-06-08
Add Rust canister tests

nmattia:
This adds some simple rust canister tests (install, upgrade/downgrade).
Eventually these tests will completely replace the Haskell tests. Rare
are the Haskellers among us these days.

This includes a few changes:

* The types definitions for types used in the II interface were moved to
  a new crate, `internet_identity_interface`, which is used by both the
  main crate and the new canister tests,
* The canister tests are implemented in a new crate, `canister_tests`,
  that only contains test code. Building it without `--tests` is a
  no-op,
* The `cargo-tests` workflow was moved to the `canister-tests` workflow
  because it needs the built II Wasm module.

### DIFF
--- a/src/canister_tests/src/api.rs
+++ b/src/canister_tests/src/api.rs
@@ -18,13 +18,13 @@ pub fn create_challenge(env: &StateMachine, canister_id: CanisterId) -> types::C
 pub fn register(env: &StateMachine, canister_id: CanisterId, device_data: types::DeviceData, challenge_attempt: types::ChallengeAttempt) -> types::RegisterResponse {
     match framework::call_candid_as(env, canister_id, framework::some_principal(), "register", (device_data, challenge_attempt)) {
         Ok((r,)) => r,
-        Err(e) => panic!("Oh no! {:?}", e),
+        Err(e) => panic!("Failed to register: {:?}", e),
     }
 }
 
 pub fn lookup(env: &StateMachine, canister_id: CanisterId, user_number: types::UserNumber) -> Vec<types::DeviceData> {
     match framework::call_candid_as(env, canister_id, framework::some_principal(), "lookup", (user_number,)) {
         Ok((r,)) => r,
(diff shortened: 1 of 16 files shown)


nomeata (commented):
Haven't seen `ic_state_machine_tests` yet. A kind of `drun` but scriptable from rust instead of stdin? Quite neat.

src/canister_tests/src/api.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub fn create_challenge(env: &StateMachine, canister_id: CanisterId) -> types::Challenge {
+    let (c,) = framework::call_candid(env, canister_id, "create_challenge", ()).unwrap();
+    c
+}
+
+pub fn register(env: &StateMachine, canister_id: CanisterId, device_data: types::DeviceData, challenge_attempt: types::ChallengeAttempt) -> types::RegisterResponse {
+    match framework::call_candid_as(env, canister_id, framework::some_principal(), "register", (device_data, challenge_attempt)) {
+        Ok((r,)) => r,
+        Err(e) => panic!("Oh no! {:?}", e),
```
frederikrothenberger, line 21: I would prefer more explicit error message, i.e. `register failed: {:?}`

src/canister_tests/src/api.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+pub fn lookup(env: &StateMachine, canister_id: CanisterId, user_number: types::UserNumber) -> Vec<types::DeviceData> {
+    match framework::call_candid_as(env, canister_id, framework::some_principal(), "lookup", (user_number,)) {
+        Ok((r,)) => r,
+        Err(e) => panic!("Oh no! {:?}", e),
```
frederikrothenberger, line 28: dito

frederikrothenberger (commented):
Very cool. I like it. 👍 
Left a few small comments.

nmattia:
@frederikrothenberger thanks, addressed!

PR #667, opened 2022-06-08
Show confirm dialog on device deletions for all devices

frederikrothenberger:
Previously, there was only a dialog when the same device that was used to authenticate was deleted. Now there is a confirmation dialog for all device deletions.


src/frontend/src/flows/manage.ts
```diff
@@ -328,19 +327,26 @@ const bindRemoveListener = (
     if (isOnlyDevice) {
       return alert("You can not remove your last device.");
     } else {
+      let shouldProceed;
```
nmattia, line 330: nit: this suggests `shouldProceed` is mutable whereas really it's just that we don't know how to initialize it here. Also, using the `let` there's a risk that someone will add a branch to the `if/else` and will forget to assign `shouldProceed`.

How about making sure `shouldProceed` is always set correctly?

```
const shouldProceed = sameDevice ? confirm("this be yar last device") : confirm("ya sure, pirate?");
...

PR #671, opened 2022-06-08
Fix Docker cached dependencies

nmattia:
For reasons unclear, now that we have more than 1 crate involved in the
build, it makes a difference to cargo whether we specify the manifest
path or not during the first dependency build.

After specifying the manifest path, we now also need to `touch` the
updated, non-dummy `lib.rs`s; unclear why this wasn't needed before.


frederikrothenberger (approved):
Thanks, that's a massive speed-up! :-)

PR #672, opened 2022-06-09
Rename sample-javascript to selenium-test-app

frederikrothenberger:
Renames the sample-javascript app to better convey the intent of it's usage. This change is made in preparation for further changes that will extend this test app to add functionality only required for testing.


nmattia (approved):
LGTM, thanks! Did you grep for all occurrences of `sample-javascript`?

frederikrothenberger:
Yes I did. There is more in the `package.json` but I will touch that on the next PR where I will also update dependencies.

PR #673, opened 2022-06-09
Update selenium-test-app dependencies

frederikrothenberger:
Also includes other minor changes in package.json:
- renamed the package in accordance with the previous rename
- removed the version as it is not used and never actually updated

### DIFF
--- a/src/frontend/src/test-e2e/selenium.test.ts
+++ b/src/frontend/src/test-e2e/selenium.test.ts
@@ -244,7 +244,7 @@ test("Log into client application, after registration", async () => {
 
     // default value
     const exp = await browser.$("#expiration").getText();
-    expect(Number(exp) / (24 * 60 * 60_000_000_000)).toBeCloseTo(1);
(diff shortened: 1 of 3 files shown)


src/frontend/src/test-e2e/selenium.test.ts
```diff
@@ -244,7 +244,7 @@ test("Log into client application, after registration", async () => {
 
     // default value
     const exp = await browser.$("#expiration").getText();
-    expect(Number(exp) / (24 * 60 * 60_000_000_000)).toBeCloseTo(1);
+    expect(Number(exp) / (8 * 60 * 60_000_000_000)).toBeCloseTo(1);
```
nmattia, line 247: what's this? updated version of something?

frederikrothenberger: Yes, the updated Agent-js changed the session validity default.

nmattia (approved):

PR #675, opened 2022-06-10
Add more tests for WindowPostMessage interactions

frederikrothenberger:
This PR extends the selenium-test-app to include more capabilities useful for tests with non-standard windowPostMessages.

The following testcases are added:
- assert 'authorize-ready' is actually sent by II
- test that invalid data won't crash II

Sending an incomplete message now results in an ugly error. A fix for this will be submitted in a follow-up PR.


demos/selenium-test-app/src/main.js
```diff
@@ -16,49 +27,148 @@ const iiUrlEl = document.getElementById("iiUrl");
...
-      delegationEl.innerText = "Current identity is not a DelegationIdentity";
-      expirationEl.innerText = "N/A";
+function addMessageElement(message, received) {
+  const messageContainer = document.createElement("div");
+  messageContainer.classList.add("postMessage");
+  const messageTitle = document.createElement("div");
+  messageTitle.classList.add("postMessage-title");
+  const messageContent = document.createElement("div");
+  messageContent.innerText = JSON.stringify(message, (_, v) =>
+    typeof v === "bigint" ? v.toString() : v
```
nmattia, line 62: nice

nmattia (approved):
Really great work, thanks! If we were to add any more code I think we should port this to typescript, but for now LGTM.

PR #677, opened 2022-06-10
Wait for badge to be loaded in selenium tests

frederikrothenberger:
Since the IC currently does not support caching, the IC badge is loaded again on all screens that display it. To avoid flakyness, the selenium tests were changed to wait for it to be loaded.


nmattia (commented):
how about waiting for all images to be loaded? then no need to add the ID and we don't have to repeat this for every new image: https://github.com/dfinity/nns-dapp/blob/0449da36fd20eb9bb5d712d78aea8879cb51ec8e/e2e-tests/common/waitForImages.ts

src/frontend/src/test-e2e/views.ts
```diff
@@ -7,6 +9,7 @@ export class WelcomeView extends View {
     await this.browser
       .$("#registerUserNumber")
       .waitForDisplayed({ timeout: 10_000 });
+    await waitForImages(this.browser);
```
nmattia, line 12: extra nit: I guess this will always be useful for screenshots. Should we add this to the [screenshot taking function](https://github.com/dfinity/internet-identity/blob/59c06f609955eabb136e2968be7070bedd511ddb/src/frontend/src/test-e2e/util.ts#L328)?

nmattia (approved):
Nice, thanks!

src/frontend/src/test-e2e/views.ts
frederikrothenberger: 👍

PR #679, opened 2022-06-11
Port register public key mismatch unit test to rust

frederikrothenberger:


src/canister_tests/src/tests.rs
```diff
@@ -24,16 +27,20 @@ fn ii_upgrade_retains_anchors() {
...
     let challenge = api::create_challenge(&env, canister_id);
-    let user_number = match api::register(
-        &env,
+    let env_argument = &env;
+    let device_data = framework::some_device_data();
+    let challenge_attempt = types::ChallengeAttempt {
+        chars: "a".to_string(),
+        key: challenge.challenge_key,
+    };
+    let user_number = match register(
```
nmattia, line 36: how about `.unwrap()` or `.expect()`? in this test case I really don't think we ever expect an error, right?

nmattia (commented):
Thanks! few questions

src/canister_tests/src/tests.rs
```diff
@@ -24,16 +27,20 @@ fn ii_upgrade_retains_anchors() {
     let env = StateMachine::new();
     let canister_id = framework::install_ii_canister(&env, framework::II_WASM_PREVIOUS.clone());
     let challenge = api::create_challenge(&env, canister_id);
-    let user_number = match api::register(
-        &env,
+    let env_argument = &env;
```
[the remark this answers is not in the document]
frederikrothenberger: Ah, this is a refactoring artifact.

src/canister_tests/src/tests.rs
frederikrothenberger: 👍 Still have to match the variant though.

src/canister_tests/src/tests.rs
```diff
@@ -52,3 +59,23 @@ fn ii_canister_can_be_upgraded_and_rolled_back() {
...
+#[test]
+fn registration_with_mismatched_sender_fails() {
+    let env = StateMachine::new();
+    let canister_id = framework::install_ii_canister(&env, framework::II_WASM_PREVIOUS.clone());
+    let challenge = api::create_challenge(&env, canister_id);
+    let result = register(
+        &env,
+        canister_id,
+        other_principal(),
+        framework::some_device_data(),
```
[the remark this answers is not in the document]
frederikrothenberger: Agreed. But I'd rather do numbered sets (i.e. `pubkey_1`, `device_data_1`, `principal_1`) because we also need the different devices to test device registration, etc. And it does not make a lot of sense for alice to add bobs device to her account.

src/canister_tests/src/api.rs
```diff
@@ -18,18 +19,19 @@ pub fn create_challenge(env: &StateMachine, canister_id: CanisterId) -> types::C
 pub fn register(
     env: &StateMachine,
     canister_id: CanisterId,
+    sender: PrincipalId,
     device_data: types::DeviceData,
     challenge_attempt: types::ChallengeAttempt,
-) -> types::RegisterResponse {
+) -> Result<types::RegisterResponse, CallError> {
```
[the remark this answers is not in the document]
frederikrothenberger: As discussed, we leave it as is for now.

src/canister_tests/src/tests.rs
```diff
@@ -24,23 +28,27 @@ fn ii_upgrade_retains_anchors() {
     let env = StateMachine::new();
     let canister_id = framework::install_ii_canister(&env, framework::II_WASM_PREVIOUS.clone());
     let challenge = api::create_challenge(&env, canister_id);
-    let user_number = match api::register(
-        &env,
+    let challenge_attempt = types::ChallengeAttempt {
```
nmattia, line 31: nit: unclear why this moved , was it for clarity? should we keep it consistent with the other test and inline as well?

src/canister_tests/src/tests.rs
```diff
@@ -52,3 +60,24 @@ fn ii_canister_can_be_upgraded_and_rolled_back() {
...
+#[test]
+fn registration_with_mismatched_sender_fails() {
+    let env = StateMachine::new();
+    let canister_id = framework::install_ii_canister(&env, framework::II_WASM_PREVIOUS.clone());
+    let challenge = api::create_challenge(&env, canister_id);
+    let result = register(
+        &env,
+        canister_id,
+        principal_2(),
+        device_data_1(),
```
nmattia, line 73: I think this improves readability + understanding a lot, wdyt?

src/canister_tests/Cargo.toml
```diff
@@ -14,4 +15,5 @@ internet_identity_interface = { path = "../internet_identity_interface" }
 candid = "0.7"
 ic-cdk = "0.5"
 ic-types = "0.3"
+ic-error-types = { git = "https://github.com/dfinity/ic", rev = "46b3de985fef2cc9e326ed191c266b610512443b" }
```
nmattia, line 18: I _think_ that if you just add `ic-error-types = "*"` it will pull it from the git checkout and then there's no risk of updating one rev but not the other

nmattia (approved):
looking great, thanks!

src/canister_tests/src/tests.rs
frederikrothenberger: Hm, I need to be more careful with IDE powered refactorings

-> reverted.

src/canister_tests/src/tests.rs
frederikrothenberger: Yes, agreed 👍

PR #681, opened 2022-06-13
Stop suggesting 'docker build .'

nmattia:
Being the only code blog in that section, users would go straight to
that instead of using the `docker-build` script.


frederikrothenberger (approved):
Thanks! 👍

PR #682, opened 2022-06-14
Port remote device registration flow tests to rust

frederikrothenberger:


src/canister_tests/src/api.rs
```diff
@@ -51,3 +51,86 @@ pub fn lookup(
...
+    match framework::call_candid_as(
+        env,
+        canister_id,
+        sender,
+        "enter_device_registration_mode",
+        (user_number,),
+    ) {
+        Ok((r,)) => Ok(r),
+        Err(e) => Err(e),
+    }
```
nmattia, line 82:
```suggestion
    framework::call_candid_as(
        env,
        canister_id,
        sender,
        "enter_device_registration_mode",
        (user_number,)).map(|(x,)| x)
```

and same above and below

src/canister_tests/src/framework.rs
```diff
@@ -175,3 +187,13 @@ where
         Err(user_error) => Err(CallError::UserError(user_error)),
     }
 }
+
+pub fn expect_user_error_with_message<T>(
+    result: Result<T, CallError>,
+    error_code: ErrorCode,
+    message: &str,
```
nmattia, line 194: nit: how about passing a `Regex` object? that'll make it clear that you can actually use regexes here

src/canister_tests/src/tests.rs
```diff
@@ -75,7 +58,306 @@ fn registration_with_mismatched_sender_fails() {
...
+
+        assert!(matches!(
+            api::verify_tentative_device(
+                &env,
+                canister_id,
+                principal_1(),
+                user_number,
+                "invalid code".to_string()
+            )?,
+            types::VerifyTentativeDeviceResponse::WrongCode { retries_left: 2 }
```
nmattia, line 184: nit: this will break if we ever bump the number of retries

src/canister_tests/src/tests.rs
```diff
@@ -75,7 +58,306 @@ fn registration_with_mismatched_sender_fails() {
...
+    fn reject_tentative_device_if_registration_mode_is_expired() -> Result<(), CallError> {
+        const REGISTRATION_MODE_EXPIRATION: Duration = Duration::from_secs(900);
+        let env = StateMachine::new();
+        let canister_id = framework::install_ii_canister(&env, framework::II_WASM_PREVIOUS.clone());
+        let user_number = flows::register_anchor(&env, canister_id);
+
+        api::enter_device_registration_mode(&env, canister_id, principal_1(), user_number)?;
+        env.set_time(
+            SystemTime::UNIX_EPOCH + REGISTRATION_MODE_EXPIRATION + Duration::from_secs(1),
+        );
```
nmattia, line 266: would be great if there was a way to modify the time instead of setting the time; here for instance we have no guarantee that we aren't going back in time (although unlikely). @roman-kashitsyn thoughts on that?

nmattia (approved):
Looking great! a few nitpicks.

Any thoughts on why porting those tests removed 100 lines but added 400+ lines?

frederikrothenberger (commented):
> Any thoughts on why porting those tests removed 100 lines but added 400+ lines?

I think there are a few reasons:
- Rust is a less dense language than Haskell
- we focus more on readability in the Rust tests and added a few comments
- no Haskell infrastructure has been removed yet (only tests) but Rust infrastructure was added.

But I much prefer having more readable tests even if it takes 4 times the lines.

PR #683, opened 2022-06-14
Add clean-build check on main

nmattia:
This adds a full build from scratch on 6 platforms (3 Ubuntu, 3 macOS) to
make sure that builds are succesful on as many platforms as possible.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -534,3 +534,18 @@ jobs:
           # populated by GitHub Actions
           # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  clean-build:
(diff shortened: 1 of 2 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -534,3 +534,18 @@ jobs:
           # populated by GitHub Actions
           # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  clean-build:
+    runs-on: ${{ matrix.os }}
+    if: ${{ github.ref == 'refs/heads/main' }}
```
frederikrothenberger, line 540: Is there a CI run that executed this action (e.g. before the condition was introduced) that you could link to?

nmattia: There you go: https://github.com/dfinity/internet-identity/actions/runs/2495144575

It's not exactly the same, but very similar; also note that the workflow got canceled after I force pushed the commit

PR #684, opened 2022-06-14
Enable Ubuntu reproducibility checks

nmattia:
This checks the Ubuntu-generated modules' checksums against the docker-generated module's checksum. This asks rustc to map the path to sources to a standard (across our builds) `/cargo`, to avoid the user home's popping up in our builds.

First commit adds the failing check, second commit makes the builds reproducible.


frederikrothenberger (approved):
LGTM

PR #685, opened 2022-06-14
Don't cache ic-cdk-optimizer

nmattia:
Because we build across many different Ubuntu versions, it can be that a
recent Ubuntu (with recent glibc) builds ic-cdk-optimizer, which then
can't be used on older Ubuntus, but older Ubuntus will still get a cache
hit.

What we could do is add the current version of Ubuntu/macOS to the cache
key, but really this only runs on 'main', so for now it's acceptable to
have an extra 4mn in the build.


frederikrothenberger (approved):
Rubber stamp

PR #687, opened 2022-06-16
Pass RUSTFLAGS in dependency build

nmattia:
This instructs the Dockerfile to use the official build script, which
now takes an extra option for building only rust dependencies.

[When introducing `--remap-path-prefix`](https://github.com/dfinity/internet-identity/pull/684) I forgot to set the build flag
during the dependency build, the actual build in Docker would rebuild
most of the crates. Now the Dockerfile uses the build script as well.

I've moved the build script to `./scripts/build`, which is more standard than `src/internet_identity/build.sh`. See first commit for actual changes to the script.


frederikrothenberger (approved):
LGTM

PR #690, opened 2022-06-17
Fix add-and-commit warning

nmattia:
The `add-and-commit` action was updated to a version that deprecated
some action inputs. This fixes the inputs and makes the message more
consistent with the selenium screenshot committer.


frederikrothenberger (approved):
LGTM

PR #697, opened 2022-06-20
Port asset certification tests to rust

frederikrothenberger:
The check for 404 responses was not ported as it does not conform to the specification (and does not work in practice).


src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,145 @@
+// adapted from https://github.com/dfinity/icx-proxy/blob/0cd1a22f717b56ac550a3554a25a845878bfb4e8/src/main.rs#L611
+// TODO: certificate validation should be it's own library
```
nmattia, line 2:
```suggestion
// TODO: certificate validation should be its own library
```

do we need to track this somewhere?

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,145 @@
...
+pub enum ValidationError {
+    MalformedCertificate { message: String },
+    CertificateValidationFailed { inner: AgentError },
+    WitnessLookupFailed { inner: AgentError },
+    TreeHashCertifiedDataMismatch,
+    AssetPathLookupFailed,
+    AssetHashMismatch,
+}
+
+pub fn validate_certification(
```
nmattia, line 36: lot of stuff happening here. Is this following a spec? Can we give some context to help the reader understand what's happening?

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,145 @@
...
+    ic_certificate: &str,
+    canister_id: CanisterId,
+    uri_path: &str,
+    body: &[u8],
+    encoding: Option<&str>,
+    root_key: ThresholdSigPublicKey,
+) -> Result<(), ValidationError> {
+    let cert = parse_header_value(ic_certificate, 1)?;
+    let tree: HashTree = parse_header_value(ic_certificate, 2)?;
+    let agent = initialize_agent(&root_key);
```
nmattia, line 46: what is the role of the "agent"? is there any chance we can steal some code from ic-agent instead of tying us to a new dependency?

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,165 @@ fn registration_with_mismatched_sender_fails() {
...
+        let assets: Vec<(&str, Option<&str>)> = vec![
+            ("/", None),
+            ("/index.html", None),
+            ("/index.js", Some("gzip")),
+            ("/loader.webp", None),
+            ("/favicon.ico", None),
+            ("/ic-badge.svg", None),
+        ];
+        let env = StateMachine::new();
+        let canister_id = framework::install_ii_canister(&env, framework::II_WASM_PREVIOUS.clone());
```
nmattia, line 92: do you really want "II_WASM_PREVIOUS" here?

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,165 @@ fn registration_with_mismatched_sender_fails() {
...
+
+        for (asset, encoding) in assets {
```
nmattia, line 94:
```suggestion
        // for each asset, fetch the asset, check the HTTP status code, headers and certificate.
        for (asset, encoding) in assets {
```

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,165 @@ fn registration_with_mismatched_sender_fails() {
...
+                env.root_key(),
+            )
+            .expect(&format!("validation for asset \"{}\" failed", asset));
+
+            verify_security_headers(&http_response.headers)
+        }
+        Ok(())
+    }
+
+    fn verify_security_headers(headers: &Vec<HeaderField>) {
```
nmattia, line 144: It'd be nice to keep this file as short as possible and only contain tests so that they're easy to read and get an idea of what's getting testing. Do you think we could put these helpers somewhere else?

src/canister_tests/src/tests.rs
frederikrothenberger: No, thanks. Also, I have just realized, this is wrong in a lot of other places too --> PR incoming.

src/canister_tests/src/certificate_validation.rs
frederikrothenberger: Added link to spec 👍

src/canister_tests/src/certificate_validation.rs
```diff
@@ -33,6 +33,8 @@ pub enum ValidationError {
     AssetHashMismatch,
 }
 
+/// Validates asset certification accoriding to the HTTP gateway specification:
```
nmattia, line 36:
```suggestion
/// Validates asset certification according to the HTTP gateway specification:
```

src/canister_tests/src/certificate_validation.rs
frederikrothenberger: I tried, but it's worse than having a dependency on `agent-rs`. The reason is that `agent-rs` includes a _modified_ copy of https://github.com/miracl/core. And the reason for that, in turn, is that the library can only be configured by copying and modifying it. 🙃

backend-tests/backend-tests.hs
```diff
@@ -707,15 +707,6 @@ tests wasm_file = testGroup "Tests" $ upgradeGroups $
     assertMetric metrics "internet_identity_signature_count" 1.0
     assertMetric metrics "internet_identity_users_in_registration_mode" 1.0
 
-  , withUpgrade $ \should_upgrade -> testGroup "HTTP Assets"
-    [ iiTest asset $ \cid -> do
-      when should_upgrade $ doUpgrade cid
-      r <- queryII cid dummyUserId #http_request (httpGet asset)
-      validateHttpResponse cid asset r
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, removed

backend-tests/backend-tests.hs
```diff
@@ -707,15 +707,6 @@ tests wasm_file = testGroup "Tests" $ upgradeGroups $
...
-      validateSecurityHeaders r
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, removed

src/canister_tests/src/certificate_validation.rs
nmattia: 😭

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,147 @@
...
+    };
+
+    if *witness != tree.digest() {
+        return Err(TreeHashCertifiedDataMismatch);
+    }
+
+    let asset_path = ["http_assets".into(), uri_path.into()];
+    let tree_sha = match tree.lookup_path(&asset_path) {
+        LookupResult::Found(v) => v,
+        _ => match tree.lookup_path(&["http_assets".into(), "/index.html".into()]) {
```
nmattia, line 73: > Else, if it does not exist, ["http_assets","/index.html"] must exist and be a leaf.

definitely useful to have here

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,147 @@
...
+            LookupResult::Found(v) => v,
+            _ => {
+                return Err(AssetPathLookupFailed);
+            }
+        },
+    };
+
+    let body_sha = decode_body_to_sha256(body, encoding).unwrap();
```
nmattia, line 81: maybe add a comment about how II doesn't follow the spec since gzipped content has the hash of the gzipped content?

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,147 @@
...
+use regex::Regex;
+use serde::Deserialize;
+use sha2::{Digest, Sha256};
+use std::io::Read;
+
+// The limit of a buffer we should decompress ~10mb.
+const MAX_CHUNK_SIZE_TO_DECOMPRESS: usize = 1024;
+const MAX_CHUNKS_TO_DECOMPRESS: u64 = 10_240;
+
+const CERTIFICATE_REGEX: &'static str = "^certificate=:([^:]*):, tree=:([^:]*):$";
```
nmattia, line 24:
```suggestion
const CERTIFICATE_REGEX: &'static str = "^certificate=:([^:]*):, tree=:([^:]*):$";
```

nit: must there be exactly one whitespace between the values or can there be an arbitrary number?

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,147 @@
...
+                message: format!("no regex match for capture group {}", capture_group),
+            })?
+            .as_str(),
+    )
+    .map_err(|err| MalformedCertificate {
+        message: format!("failed to decode base64 value: {:?}", err),
+    })?;
+    serde_cbor::from_slice(&cert_blob).map_err(|err| MalformedCertificate {
+        message: format!("failed to decode cbor value: {:?}", err),
+    })
```
nmattia, line 124: here it looks like you're parsing something like this:

`certificate=:<cert>:, tree=:<tree>:`

and returning either `cert` if `capture_group` is `1`, and `tree` if `capture_group` is `2` (that's what I gathered from a quick read).

Few questions:

* what happens if I pass `capture_group: 3`?
* do you ever call `parse_header_value` to use _only_ either `tree` or `cert`? i.e. does the caller really need to specify which one they need?

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,82 @@ fn registration_with_mismatched_sender_fails() {
...
+            )?;
+
+            assert_eq!(http_response.status_code, 200);
+
+            // check the appropriate Content-Encoding header is set
+            if let Some(enc) = encoding {
+                let (_, content_encoding) = http_response
+                    .headers
+                    .iter()
+                    .filter(|(name, _)| name == "Content-Encoding")
```
nmattia, line 114: note: http headers should be case insensitive

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,82 @@ fn registration_with_mismatched_sender_fails() {
...
+                    content_encoding, enc,
+                    "unexpected \"Content-Encoding\" header value"
+                );
+            }
+
+            let (_, ic_certificate) = http_response
+                .headers
+                .iter()
+                .filter(|(name, _)| name == "IC-Certificate")
+                .exactly_one()
```
nmattia, line 127: case insensitive, find

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,82 @@ fn registration_with_mismatched_sender_fails() {
...
+                .unwrap();
+
+            validate_certification(
+                ic_certificate,
+                canister_id,
+                asset,
+                &http_response.body,
+                None, // should really be `encoding`, but we need to solve L2-722 first
```
nmattia, line 135:
```suggestion
                None, // should really be `encoding`, but cannot use it because ..., see L2-722 for details
```

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,147 @@
...
+    uri_path: &str,
+    body: &[u8],
+    encoding: Option<&str>,
+    root_key: ThresholdSigPublicKey,
+) -> Result<(), ValidationError> {
+    let cert = parse_header_value(ic_certificate, 1)?;
+    let tree: HashTree = parse_header_value(ic_certificate, 2)?;
+    let agent = initialize_agent(&root_key);
+
+    if let Err(err) = agent.verify(&cert, canister_id.get().0, false) {
```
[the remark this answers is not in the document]
frederikrothenberger: I added the comment explaining which step of the spec the agent is supposed to check.

src/canister_tests/src/certificate_validation.rs
frederikrothenberger: Yeah, I restructured the code to be more clear (hopefully).

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,147 @@
...
+        Some("gzip") => {
+            let mut decoder = GzDecoder::new(body);
+            for _ in 0..MAX_CHUNKS_TO_DECOMPRESS {
+                let bytes = decoder.read(&mut decoded).ok()?;
+                if bytes == 0 {
+                    return Some(sha256.finalize().into());
+                }
+                sha256.update(&decoded[0..bytes]);
+            }
+            if decoder.bytes().next().is_some() {
```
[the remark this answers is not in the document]
frederikrothenberger: I added a comment explaining it. The limits are there, because I took the code from `icx-proxy`.

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+        Ok(witness) => witness,
+        Err(err) => {
+            return Err(WitnessLookupFailed { inner: err });
+        }
+    };
+
+    // 4. The tree must be a hash tree as per Encoding of certificates.
+    let tree: HashTree = decode_value(encoded_tree)?;
+    // 5. The root hash of that tree must match the canister's certified data.
+    if *witness != tree.digest() {
```
nmattia, line 72:
```suggestion
    if witness != tree.digest() {
```

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+        )
+        .build()
+        .unwrap();
+    agent
+        .set_root_key(public_key_to_der(&root_key.into_bytes()).unwrap())
+        .expect("setting root key failed");
+    agent
+}
+
+fn decode_value<T>(encoded_value: &str) -> Result<T, ValidationError>
```
nmattia, line 133: nit: for once I wouldn't mind a name that's a little longer; `decode_value` could be decoding anything. Or maybe something like `decode_cert`?

nmattia (approved):
Nice, thanks a lot!

src/canister_tests/src/certificate_validation.rs
```diff
@@ -0,0 +1,166 @@
...
+        return Err(AssetHashMismatch);
+    }
+    Ok(())
+}
+
+fn parse_header(ic_certificate: &str) -> Result<(&str, &str), ValidationError> {
+    let captures = Regex::new("^certificate=:([^:]*):,\\s*tree=:([^:]*):$")
+        .unwrap()
+        .captures(ic_certificate)
+        .ok_or(MalformedCertificate {
```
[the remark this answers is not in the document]
frederikrothenberger: I'd rather keep the error, as it allows to expect specific validation errors.

PR #701, opened 2022-06-21
Clarify dev banner

nmattia:
The dev builds of II are being used and shared more and more. This
clarifies to users that they should not use their regular II anchor.


peterpeterparker (approved):
LGTM.

Being said, I would personally shorten the msg, something like `Test only. Do not use your regular anchor!`

PR #707, opened 2022-06-23
Faq summary text indentation

wackyleo459:
- Summary text (span) when runs over 1 line, does not indent automatically outside summary::marker.

Before:
Screen Shot 2022-06-23 at 1 33 37 PM

Changed: 
Screen Shot 2022-06-24 at 12 10 30 PM


nmattia:
Thanks a lot, looks much nicer when aligned!

In [this](https://github.com/dfinity/internet-identity/pull/637) PR I tried to add some spacing between the marker and the text, do you know why it disappeared here?

wackyleo459:
@nmattia Thank you for pointing that out! Since I had changed the span text to `display: list-item` the left-margin was not working. But with left-padding, the spacing should look as you expected!!

nmattia (approved):
It's looking great, thanks @wackyleo459 !

PR #708, opened 2022-06-24
Port device management tests to rust

frederikrothenberger:
These are tests for the following canister methods: add, remove, lookup and get_anchor_info.
Compared to the haskell tests, test cases have been split to be more specific.

### DIFF
--- a/src/canister_tests/src/api.rs
+++ b/src/canister_tests/src/api.rs
@@ -87,17 +87,34 @@ pub fn lookup(
     env: &StateMachine,
     canister_id: CanisterId,
     user_number: types::UserNumber,
-) -> Vec<types::DeviceData> {
-    match framework::call_candid_as(
+) -> Result<Vec<types::DeviceData>, CallError> {
+    framework::query_candid(env, canister_id, "lookup", (user_number,)).map(|(x,)| x)
+}
+
+pub fn add(
+    env: &StateMachine,
+    canister_id: CanisterId,
+    sender: PrincipalId,
+    user_number: types::UserNumber,
+    device_data: types::DeviceData,
+) -> Result<(), CallError> {
+    framework::call_candid_as(env, canister_id, sender, "add", (user_number, device_data))
+}
+
+pub fn remove(
+    env: &StateMachine,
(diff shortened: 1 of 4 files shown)


src/canister_tests/src/api.rs
```diff
@@ -87,17 +87,34 @@ pub fn lookup(
...
-) -> Vec<types::DeviceData> {
-    match framework::call_candid_as(
+) -> Result<Vec<types::DeviceData>, CallError> {
+    framework::query_candid(env, canister_id, "lookup", (user_number,)).map(|(x,)| x)
+}
+
+pub fn add(
+    env: &StateMachine,
+    canister_id: CanisterId,
+    sender: PrincipalId,
```
nmattia, line 97: question: do we have any unauthenticated tests for `add`? how would that work?

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,246 @@ fn registration_with_mismatched_sender_fails() {
...
+
+        api::add(
+            &env,
+            canister_id,
+            principal_1(),
+            user_number,
+            device_data_2(),
+        )?;
+        let devices = api::lookup(&env, canister_id, user_number)?;
+        assert_eq!(devices.len(), 2);
```
nmattia, line 97: nit: could also compare sorted arrays

src/canister_tests/src/tests.rs
```diff
@@ -66,6 +66,246 @@ fn registration_with_mismatched_sender_fails() {
...
+        expect_user_error_with_message(
+            result,
+            CanisterCalledTrap,
+            Regex::new("[a-z\\d-]+ could not be authenticated.").unwrap(),
+        );
+    }
+
+    /// Verifies that the same device cannot be added twice.
+    #[test]
+    fn should_not_add_existing_device() -> Result<(), CallError> {
```
nmattia, line 192: nit: how about having all the `add` tests together? otherwise going back and forth between remove and add is a bit confusing

nmattia (approved):
LGTM!

PR #709, opened 2022-06-24
Port get_principal tests to rust

frederikrothenberger:
Migrates tests for the get_principal query call to rust. The tests were added to the delegation test module as the functionality is closely related.
Changes to the haskell tests:
- a test was added to verify that the result of get_principal matches the prepare_delegation principal
- tests were added to make sure that the principals are unique per user_number and frontend hostname

### DIFF
--- a/src/canister_tests/src/tests.rs
+++ b/src/canister_tests/src/tests.rs
@@ -302,15 +302,19 @@ mod device_management_tests {
     }
 }
 
-/// Tests related to prepare_delegation and get_delegation II canister calls.
+/// Tests related to prepare_delegation, get_delegation and get_principal II canister calls.
 #[cfg(test)]
 mod delegation_tests {
-    use crate::framework::{expect_user_error_with_message, principal_1, principal_2, CallError};
+    use crate::framework::{
+        device_data_1, device_data_2, expect_user_error_with_message, principal_1, principal_2,
+        CallError,
+    };
     use crate::{api, flows, framework};
     use ic_error_types::ErrorCode::CanisterCalledTrap;
     use ic_state_machine_tests::StateMachine;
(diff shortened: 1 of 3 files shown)


src/canister_tests/src/tests.rs
```diff
@@ -761,6 +765,121 @@ mod delegation_tests {
...
+
+        let dapp_principal_2 = api::get_principal(
+            &env,
+            canister_id,
+            principal_1(),
+            user_number,
+            frontend_hostname_2.to_string(),
+        )?;
+
+        assert_ne!(dapp_principal_1, dapp_principal_2);
```
nmattia, line 824: I feel like I saw this exact check somewhere else.. but I think I'm dreaming

frederikrothenberger: There is also `should_issue_different_principals` for the `prepare_delegation` call. But I think both have their place because the implementation could be broken independently.

src/canister_tests/src/tests.rs
```diff
@@ -761,6 +765,121 @@ mod delegation_tests {
...
+    }
+
+    /// Verifies that get_principal returns different principals for different anchors.
+    #[test]
+    fn should_return_different_principals_for_different_users() -> Result<(), CallError> {
+        let env = StateMachine::new();
+        let canister_id = framework::install_ii_canister(&env, framework::II_WASM.clone());
+        let user_number_1 =
+            flows::register_anchor_with(&env, canister_id, principal_1(), device_data_1());
+        let user_number_2 =
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, it does. Thanks 👍

nmattia (approved):
LGTM!

PR #710, opened 2022-06-24
Ports the metrics endpoint tests to rust

frederikrothenberger:
Migrates the haskell test for the /metrics http endpoint to rust.
The single test is split into multiple tests for each individual metric as well as a test for completness & sanity check regarding metrics time.
Tests were added for metrics that were not tested so far (memory pages, inflight captchas, users in registration mode and min/max user number).


src/canister_tests/src/tests.rs
```diff
@@ -955,6 +956,262 @@ mod http_tests {
...
+        let env = StateMachine::new();
+        let canister_id = framework::install_ii_canister(&env, framework::II_WASM.clone());
+
+        for count in 0..2 {
+            let metrics = flows::get_metrics(&env, canister_id);
+
+            let (user_count, _) = framework::parse_metric(&metrics, "internet_identity_user_count");
+            assert_eq!(user_count, count);
+
+            flows::register_anchor(&env, canister_id);
```
nmattia, line 1018: nit: how about asserting that `user_count` has been incremented here? I think it would clarify the intent a bit

src/canister_tests/src/tests.rs
```diff
@@ -955,6 +956,262 @@ mod http_tests {
...
+            user_number,
+            frontend_hostname.to_string(),
+            ByteBuf::from("last session key"),
+            None,
+        )?;
+
+        let metrics = flows::get_metrics(&env, canister_id);
+        let (signature_count, _) =
+            framework::parse_metric(&metrics, "internet_identity_signature_count");
+        assert_eq!(signature_count, 1); // old ones pruned and a new one created
```
nmattia, line 1064: nit: how about checking each individual? i.e. first `advance_time` to prune, check metric, then `prepare_delegation`, check metric

src/canister_tests/src/tests.rs
```diff
@@ -955,6 +956,262 @@ mod http_tests {
...
+            assert_eq!(user_count, count);
+
+            flows::register_anchor(&env, canister_id);
+        }
+        Ok(())
+    }
+
+    /// Verifies that the signature count metric is updated correctly.
+    #[test]
+    fn metrics_signature_count_should_increase_after_prepare_delegation() -> Result<(), CallError> {
```
nmattia, line 1025:
```suggestion
    fn metrics_signature_count() -> Result<(), CallError> {
```

since you're also checking for pruning, you're not only checking that it gets increases, but also decreased

src/canister_tests/src/tests.rs
```diff
@@ -955,6 +956,262 @@ mod http_tests {
...
+
+    /// Verifies that the stable memory pages count metric is updated correctly.
+    #[test]
+    fn metrics_stable_memory_pages_should_increase_with_more_users() -> Result<(), CallError> {
+        let env = StateMachine::new();
+        let canister_id = framework::install_ii_canister(&env, framework::II_WASM.clone());
+
+        let metrics = flows::get_metrics(&env, canister_id);
+        let (signature_count, _) =
+            framework::parse_metric(&metrics, "internet_identity_stable_memory_pages");
```
nmattia, line 1076: I think this could be abstracted away with this if necessary (since this gets repeated as lot)

```
framework::assert_metric(&env, canister_id, "internet_identity_stable_memory_pages", 1);
```

src/canister_tests/src/tests.rs
```diff
@@ -955,6 +956,262 @@ mod http_tests {
...
+            assert_eq!(
+                upgrade_timestamp,
+                env.time()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap()
+                    .as_nanos() as u64
+            );
+
+            env.advance_time(Duration::from_secs(300)); // the state machine does not advance time on its own
+            framework::upgrade_ii_canister(&env, canister_id, framework::II_WASM.clone());
```
nmattia, line 1113: here again I'm a bit confused by the construct of "first check metric, then perform operation that we expect to change the metrics". Do you mean to check the metrics after the op? if not, what does upgrading the canister give you on the last loop?

nmattia (commented):
Wow that's thorough, thanks!

src/canister_tests/src/tests.rs
frederikrothenberger: `advance_time` is not enough, we also need to do an update call that prunes expired signatures. There following update calls prune signatures: `add`, `create_challenge`, `prepare_delegation`, `register`, `remove`

I chose to call `prepare_delegation` again, because this is the least likely one of those calls to be changed to stop pruning signatures (as it is the only one that actually executes code related to signatures).

I changed the comment a little bit to make it more clear.

src/canister_tests/src/tests.rs
frederikrothenberger: I wanted to avoid having 2 asserts. But I agree, this is confusing.

src/canister_tests/src/tests.rs
frederikrothenberger: I wanted to avoid having two assert metrics, so I figured I could just loop (and disregard the last loop execution). But this is confusing and for this test I have now removed the loop altogether.

nmattia (approved):
LGTM, thanks!

PR #712, opened 2022-06-28
Fix typo in canister_tests build command

nmattia:
The command is suggested when the wasm cannot be found, and it had a
typo which led to the build not enabling the dummy captcha feature when
copy/pasted.


frederikrothenberger (approved):
Thanks

PR #715, opened 2022-06-28
Adds rust tests for the registration flow

frederikrothenberger:
The registration flow consistes of two canister calls: create_challenge and register.
As part of testing these canister calls the inteface of api::create_challenge has been changed to return a Result<Challenge, CallError> to make error behaviour testable.
Related to registration, this PR includes tests related to the intialization of the user range (ported from haskell).
The test for the user range widening has been moved to the http test module (together with other metrics) since it is unrelated to user registration.


src/canister_tests/src/tests.rs
```diff
@@ -43,27 +40,184 @@ fn ii_canister_can_be_upgraded_and_rolled_back() {
...
+        Ok(())
+    }
+
+    /// Tests that there is a maximum number of captchas that can be created in a given timeframe.
+    #[test]
+    fn should_limit_captcha_creation() -> Result<(), CallError> {
+        let env = StateMachine::new();
+        let canister_id = framework::install_ii_canister(&env, framework::II_WASM.clone());
+
+        for _ in 0..500 {
```
frederikrothenberger, line 209: This is probably too slow: https://github.com/dfinity/internet-identity/runs/7093616311?check_suite_focus=true#step:7:978

@nmattia : What do you think?

peterpeterparker (approved):
Rubber stamp

src/canister_tests/src/tests.rs
frederikrothenberger: Created ticket L2-791 to track this.

PR #717, opened 2022-06-28
Fix with_candid error message

nmattia:
The previous error message had its arguments mixed up (was showing
infered type as "bytes"). This also updates the error message to show
the actual received bytes as well as a UTF8 representation of said
bytes.


frederikrothenberger (approved):
👍

PR #724, opened 2022-07-01
Add specification for canister chosen alternative domains

frederikrothenberger:
This PR adds the specification for the upcoming feature that allows canisters to specify alternative origins for their frontends.
These origins will then receive the same principals as the existing canister frontend urls https://<canister_id>.ic0.app or
https://<canister_id>.raw.ic0.app.


docs/internet-identity-spec.adoc
```diff
@@ -170,9 +173,20 @@ The client application frontend should support delegation chains of length more
...
 --
 
+[#alternative-frontend-origins]
+=== Alternative Frontend Origins
+To allow flexibility regarding the canister frontend URL, the client may choose to provide the canonical canister frontend URL (`https://<canister_id>.ic0.app` or `https://<canister_id>.raw.ic0.app`) as the `derivationOrigin` (see <<client-auth-protocol>>). This means that Internet Identity will issue the same principals to the frontend (which uses a different origin) as it would if it were using one of the canonical URLs.
+
+*Note:* `https://<canister_id>.ic0.app` and `https://<canister_id>.raw.ic0.app` do _not_ issue the same principals by default. However, this feature can also be used to map `https://<canister_id>.raw.ic0.app` to `https://<canister_id>.ic0.app` principals or vice versa.
+
+In order for Internet Identity to accept the `derivationOrigin` the corresponding canister must list the frontend origin in the https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-canister-information[_public_ canister metadata] with name `ii_alternative_origins`. The value of this metadata must be candid encoded `(Vec<Text>)`, where each element is an allowed alternative frontend URL.
```
nomeata, line 186: Ah, that feature comes again.

Abstractly, we need a trusted way for the II frontend to learn from derivationOrigin  which alternative origins are allowed.

I would strongly advise against using a IC-specific meachanism to implement this mapping. We deliberately designed the Internet Identity protocol so that it is completely oblivious as to where a certain client frontend is hosted, whether on the Internet Identity. Instead, we relied only on Web and HTTP related technology. Right now, I can host a webpage on `https://foo.nomeata.de`, and allow users to log in using II, and if I later move that page onto the Internet Computer (once it supports custom URLs, which I fail to understand why it still doesn’t), I can take my users with me.

I think throwing that property out would be an unnecessary layer violation that bars certain applications.

So what would be an alternative? If all we have is an URL, we have to use HTTP? I’d go for a https://en.wikipedia.org/wiki/Well-known_URI at the derivationOrigin. For non-raw canister URLs that will secure thanks to the service workers. And raw urls really shouldn’t be used, just like http:// URLs shouldn’t…

docs/internet-identity-spec.adoc
```diff
@@ -125,12 +126,14 @@ where `msg` is a value with of type
...
    sessionPublicKey: Uint8Array;
    maxTimeToLive?: bigint;
+   derivationOrigin?: string;
  }
 
 where
 
 * the `sessionPublicKey` contains the public key of the session key pair.
 * the `maxTimeToLive`, if present, indicates the desired time span (in nanoseconds) until the requested delegation should expire. The Identity Provider frontend is free to set an earlier expiry time, but should not create a larger.
+* the `derivationOrigin`, if present, indicates an origin that should be used for principal derivation instead of the client origin. Values must match the following regular expression: `^https:\/\/[\w-]*(\.raw)?\.ic0\.app$`. Internet Identity will only accept values that are also listed in the https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-canister-information[_public_ custom WASM section] `ii_alternative_domains` of the corresponding canister (see <<alternative-frontend-origins>>).
```
nomeata, line 136: Shouldn’t this be alternative _origins_, rather than URLs? I expect most applications want to require `https` here. And probably not just any port, but only specific ones.

docs/internet-identity-spec.adoc
```diff
@@ -528,13 +542,18 @@ This flow is the boring default
...
 4. The `event.data` should be a message as per our <<client-auth-protocol>>.
-5. The `event.origin` is used as the Application Frontend’s hostname
-6. The front end calls `get_principal()` to obtain the user- and front-end-specific principal.
-7. The user is asked if they want to log into the client application, showing the client application frontend’s hostname and the used principal.
+5. If the `message` contains a value for `derivationOrigin`:
+  . If the `derivationOrigin` value does not match the format specified in the section  <<client-auth-protocol>>, the process flow is aborted with a failure message.
+  . The frontend calls `/api/v2/canister/<effective_canister_id>/read_state` to retrieve the `ii_alternative_origins` metadata of the canister corresponding to the `derivationOrigin`
+  . If there is no such metadata in the state tree, the flow is aborted with a failure message.
+  . The frontend decodes the retrieved metadata as per encoding specified in section <<alternative-frontend-origins>>.
+  . If the `event.origin` is listed in the decoded metadata the `derivationOrigin` will be used as the Application Frontend’s hostname.
```
nomeata, line 550: Ah, here you write ``ii_alternative_origins`. Better this way.

nomeata:
There is another reason why I would suggest not to use the custom WASM sections for this:

Custom WASM sections are a hard an mostly immutable part of the Canister module, i.e. that part that is the same across all instances of a service. But the names of possible alternative domains is certainly metadata that tends to vary between different instances of the same service. It would be odd if an admin of a service would have to recompile or otherwise patch the binary just to configure who that particular instance works.

Imagine a future world where there are plenty of off-the-shelve apps you can just fetch in pre-compiled form and deploy. You’d expect you can configure them dynamically, or at least via the init parameter, but not by changing the binary.

This is also very important for open services where users expect to be able to make sense of the wasm hash reported by the Internet Computer: All deployments of the same code better report the same wasm hash. This would be broken if the users would have to change the wasm.

We don’t really have a market place of apps yet, but note that there is at least one DFINITY provided app where developers regularly download a pre-compiled wasm and use: The Internet Identity dev-build! So this team in particular sees value in creating canister modules that are reusable for many instances.

More generally, I am under the impression that we often forget about the distinction between an app _developer_ (writes codes, creates a wasm), and the app _admin_ (creates canisters, uploads code, configures etc.). Often in small project they are the they same, but in general, these roles can be different people, and it is a design smell if admin-concerns affect the wasm. It is not a surprise that this isn’t always high on people’s mind – they are both on the “other” side of the Interface Spec – but we can at least try :-)

docs/internet-identity-spec.adoc
dostro: How does an app developer verify ownership of any of their `alternative-frontend-origins`? Seems like they could be dishonest about which origins they own if they're allowed to add any in their metadata without verification.

nomeata: I think that is not a problem: If nomeata.de lists dostro.com as their alternative-frontend-origins, then that just means that nomeata.de announces it trusts dostro.com.  This on its own does affect dostro.com’s security at all!

Only when the frontend served from dostro.com also sets the derivationOrigin to nomeata.de something will happen. So there is opt-in from both sides.

dostro: I missed the double opt-in bit :) thank you for the clarification!

frederikrothenberger: @nomeata: Thanks for the input! I agree that it would be good to keep II independent of canister infrastructure. I was initially worried about using HTTP requests because currently headers and status codes are not covered by certification. Thus any malicious replica or boundary node could redirect to a malicious endpoint not covered by certification. But the `fetch` API allows to suppress redirects, which would make it secure against this attack.

So the following caveats would apply:
* The URL _must not_ issue a redirect. This breaks with standard semantics of HTTP requests, but other precedents exist: e.g. service workers also cannot be installed from an URL that returns a redirect.
* Canisters that have used a `.raw` domain so far will be stuck in a worse place until we enforce certification on `.raw`, because any malicious replica or boundary node could issue a valid mapping for the given `.raw` well-known URL. Such applications do exist (e.g. DSCVR).
* Apparently, asset certification is hard to do in Motoko at the moment. This seems like a tooling issue, but I'm not sure. @roman-kashitsyn : Could you elaborate on this point?

@bjoernek: What do you think? Given the points above, should we move to using a well-known HTTP URL?

docs/internet-identity-spec.adoc
Dfinity-Bjoern: Isn't the pattern implicitly describing the origin `(https, [\w-]*(\.raw)?\.ic0\.app, 443)`?

nomeata: Yes, I am referring to the name. But below ii_alternative_origins is used, so maybe just a typo.

frederikrothenberger: @nomeata: from the perspective of the canister, there can be multiple alternative origins, thus the custom section (or well-known URL) will be called `alternative_origins`. But this message is sent by the front-end hosted on one of the `alternative_origins` requesting a derivation using the (single) original one, so `derivationOrigin` seems appropriate for me?

nomeata: Yes, that’s all appropriate! I am merely commenting on the `ii_alternative_domains` in the diff :-)

frederikrothenberger: Ah, I missed that. Will update in a bit :-)  Thanks

docs/internet-identity-spec.adoc
frederikrothenberger: Switched spec to use HTTP requests. Pending security review and confirmation by the SDK team that the CORS headers can be implemented within a reasonable timeframe.

frederikrothenberger: Update regarding CORS headers: The SDK team should be able to provide this within the required timeframe.

PR #726, opened 2022-07-04
Ports the stable memory restore tests to rust

frederikrothenberger:
The haskell test has been split into multiple test cases in the rust version.
Also, the expected principal has been modified to reflect the fact, that the II canister has a different canister Id on the rust test infrastructure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,6 +405,7 @@ dependencies = [
  "base64 0.13.0",
  "candid",
  "flate2",
+ "hex",
  "ic-cdk",
  "ic-certified-vars",
  "ic-crypto-internal-basic-sig-iccsa",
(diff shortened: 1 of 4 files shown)


bitdivine (approved):
Looks good but do I understand correctly that this just verifies that a given state snapshot can always be loaded?  Should there also be a test that other state, e.g. state produced by the last prod release, can be recovered?

frederikrothenberger:
> Looks good but do I understand correctly that this just verifies that a given state snapshot can always be loaded? Should there also be a test that other state, e.g. state produced by the last prod release, can be recovered?

Yes, this is just the port of the haskell test, verifying an old memory layout still works. There will be added more, see next PR. 😉

PR #728, opened 2022-07-05
Improve wording related to the new device settings

frederikrothenberger:
It was unclear, which device was being protected. Also slightly changed the message explaining why the last device cannot be removed.


peterpeterparker (approved):
Thanks!

PR #729, opened 2022-07-05
Fix error on deleting device used for login

frederikrothenberger:
This fixes a bug showing an error loading the list of devices after deleting the device that was used to authenticate the current session with.


peterpeterparker (approved):
Rubber stamp, thanks

PR #730, opened 2022-07-08
Extends the selenium test app with functionality for alterantive canister chosen origins

frederikrothenberger:
Additionally makes the selenium test app a proper canister so that it can be used for testing on mainnet.
This allows to remove the dependency of the selenium tests to the whoami motoko backend of the using-dev-build demo project.


demos/selenium-test-app/README.md
```diff
@@ -5,6 +5,8 @@ This app is used by the Internet Identity Selenium tests. It contains additional
 ## Getting Started
```
lmuntaner, line 5: Why is it called `selenium-test-app`?

For what I can see it's a normal dapp, isn't it?

demos/selenium-test-app/canister_ids.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
frederikrothenberger: Because we have a mainnet deployment of this and we want to keep this canister id: https://kvusz-kaaaa-aaaad-aabwa-cai.ic0.app/

demos/selenium-test-app/package.json
```diff
@@ -1,13 +1,13 @@
...
+  "name": "selenium_test_app",
   "private": true,
   "dependencies": {
-    "@dfinity/agent": "^0.11.3",
-    "@dfinity/auth-client": "^0.11.3",
-    "@dfinity/authentication": "^0.11.3",
-    "@dfinity/candid": "^0.11.3",
-    "@dfinity/identity": "^0.11.3",
-    "@dfinity/principal": "^0.11.3",
+    "@dfinity/agent": "^0.12.1-beta.0",
```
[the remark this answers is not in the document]
frederikrothenberger: This is required because only the beta version supports `derivationOrigin`. This is required to test the fast-tracked feature https://dfinity.atlassian.net/browse/L2-712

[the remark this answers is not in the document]
lmuntaner: 👍

demos/selenium-test-app/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub enum AlternativeOriginsMode {
+    CertifiedContent,
+    UncertifiedContent,
+    Redirect { location: String },
+}
+
+#[query]
+pub fn http_request(req: HttpRequest) -> HttpResponse {
```
nmattia, line 94: isn't this better implemented as an asset canister? why re-implement the asset serving logic?

demos/selenium-test-app/README.md
nmattia: How exactly do we tamper with the behavior? Also that's a big-ish design decision (for this test app), can we record that somewhere in the README?

demos/selenium-test-app/lib.rs
nmattia: How exactly do we tamper with the behavior? Also that's a big-ish design decision (for this test app), can we record that somewhere in the README?

demos/selenium-test-app/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub struct HttpResponse {
+    pub status_code: u16,
+    pub headers: Vec<HeaderField>,
+    pub body: Cow<'static, Bytes>,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub enum AlternativeOriginsMode {
```
nmattia, line 87: What does this do?

demos/selenium-test-app/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+    api::set_certified_data(&prefixed_root_hash[..]);
+}
+
+lazy_static! {
+    // The full content of the index.html, after the canister ID (and script tag) have been
+    // injected
+    static ref INDEX_HTML_STR: String = {
+        let canister_id = api::id();
+        let index_html = include_str!("dist/index.html");
+        let index_html = index_html.replace(
```
nmattia, line 208: copy pasta?

demos/selenium-test-app/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+        (
+            "/.well-known/ii-alternative-origins",
+            b"{\"alternativeOrigins\":[]}",
+            ContentType::JSON,
+        ),
+        (
+            "/.well-known/evil-alternative-origins",
+            b"{\"alternativeOrigins\":[\"https://evil.com\"]}",
+            ContentType::JSON,
+        ),
```
nmattia, line 237: Can you explain these two values? Why do we return an empty list of origins, and when would `evil-...` be queried?

demos/selenium-test-app/lib.rs
frederikrothenberger: It switches the behavior from the `/.well-known/ii-alternative-origins` asset:
you can either have
* the content with valid certification
* the content without certification
* a redirect

demos/selenium-test-app/lib.rs
frederikrothenberger: Yes, it has been copied from II but it is in use.

demos/selenium-test-app/lib.rs
frederikrothenberger: The empty list is just the default. The value of this asset can be updated by calling `update_alternative_origins`.
`/.well-known/evil-alternative-origins` is just the default value for the `redirect` mode explained in the previous comment.

demos/selenium-test-app/lib.rs
nmattia: And this is all for testing, correct? Can we have this mentioned in the code?

demos/selenium-test-app/lib.rs
nmattia: So II's index.html has a `script` tag with id `setupJs` which gets looked up and replaced.

In this test app I can't see a script tag in the index.html. Is this using a `script` tag generated by webpack? Why does the new value have `id="setupJs"`?

demos/selenium-test-app/lib.rs
nmattia: I see. Say someone were to discover the codebase for the first time, how easy do you reckon it would be to figure all of this out? Should we add some notes about this somewhere, for instance here where you define the assets?

demos/selenium-test-app/lib.rs
frederikrothenberger: > Is this using a script tag generated by webpack?
Yes, exactly.

> Why does the new value have id="setupJs"?
This is indeed a copy paste artifact.

demos/selenium-test-app/lib.rs
frederikrothenberger: Yes, let me do a clean-up pass through the test app. I have created L2-885 to track this.

demos/selenium-test-app/lib.rs
frederikrothenberger: Yes, see #751.

PR #731, opened 2022-07-08
Html5 semantics

wackyleo459:
## Motivation : to improve user accessibility

Changes
- Any banners, such as notification message that is shown on top during development or testing purpose, are given a `role` of banner. This is to account for any community users who rely on assistive technology.

- Color contrast: the "ADD NEW DEVICE" and "ADD RECOVERY MECHANISM" button text color was adjusted to have appropriate color contrast ratio of 4.5:1. Color differentiation is not apparent.
Screen Shot 2022-07-07 at 2 02 58 PM
Screen Shot 2022-07-08 at 1 14 22 PM
Screen Shot 2022-07-08 at 1 14 17 PM

- Icon Settings button given aria-label

- During Copy of seed phrase, copy button changes to checked icon. Added title label.
Screen Shot 2022-07-08 at 12 56 54 PM

Fix
- typo in FAQ section
- id label `deviceLabel` changed to class label, as id label only is for one unique element.


src/frontend/src/flows/manage/index.ts
```diff
@@ -139,7 +139,7 @@ const pageContent = (userNumber: bigint, devices: DeviceData[]) => html`
     <label>Identity Anchor</label>
     <div class="highlightBox">${userNumber}</div>
     <div class="labelWithAction">
-      <label class="labelAddedDevices" id="deviceLabel"
+      <label class="labelAddedDevices" class="deviceLabel"
```
frederikrothenberger, line 142: Nit: I don't think there is any styling attached to the `deviceLabel` class. You could also just remove it?

frederikrothenberger (approved):
LGTM, thanks!

PR #732, opened 2022-07-09
Run Selenium tests within docker compose

frederikrothenberger:
This gives the following benefits:
- chrome driver and chrome are always in sync and decoupled from the github actions runners
- docker compose simplifies networking (i.e. we no longer have to use the canisterId url parameter)


.gitignore
```diff
@@ -19,3 +19,6 @@ wdio.log
 
 # Cargo build output
 target/
+
+# selenium docker setup filse
```
lmuntaner, line 23: typo: "filse" --> "files" 👀

lmuntaner (approved):
rubber stamp

docker-test-env/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+    volumes:
+      - "./dfx-docker/canisters:/canisters:ro"
+    healthcheck:
+      test: [ "CMD", "curl", "-f", "http://localhost:8000?canisterId=rrkah-fqaaa-aaaaa-aaaaq-cai" ]
+      interval: 30s
+      timeout: 10s
+      retries: 5
+    extra_hosts:
+      - "rrkah-fqaaa-aaaaa-aaaaq-cai.localhost:127.0.0.1"
+      - "ryjl3-tyaaa-aaaaa-aaaba-cai.localhost:127.0.0.1"
```
nmattia, line 16: where do these values come from?

docker-test-env/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
+version: "3.9"
```
nmattia, line 1: Lots happening throughout this file, few comments would go a long way

nmattia (commented):
I'm confused, can I still run the selenium tests without docker? How would I go about running tests with non-headless?

Also I'm struggling to understand the architecture. What's running where now? Is the test driver still running locally?

docker-test-env/docker-compose.yml
frederikrothenberger: The test driver is still running locally. Everything else is within docker compose:
so test driver (locally) -> selenium container ->  nginx container -> dfx container. I have added a readme explaining this in a later PR.

frederikrothenberger: The values are hardcoded to just the canisterIds that `dfx` happens to assign those canisters (deterministically).

docker-test-env/docker-compose.yml
frederikrothenberger: I created L2-858 to track this.

PR #733, opened 2022-07-11
Implement canister chosen alternative origins for principal derivation

frederikrothenberger:
This feature allows applications to move to other origins while keeping the original user principals. See the specification in docs/internet-identity-spec.adoc for more details.


src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+    return {
+      result: "invalid",
+      message:
+        'derivationOrigin does not match regex "^https:\\/\\/([\\w-]*)(\\.raw)?\\.ic0\\.app$"',
+    };
+  }
+
+  try {
+    const canisterId = Principal.fromText(matches[1]); // verifies that a valid canister id was matched
+    const alternativeOriginsUrl = `https://${canisterId.toText()}.ic0.app/.well-known/ii-alternative-origins`;
```
lmuntaner, line 33: Should this match: `derivationOrigin + '/.well-known/ii-alternative-origins'`?

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+    const response = await fetch(
+      // always fetch non-raw
+      alternativeOriginsUrl,
+      // fail on redirects
+      { redirect: "error" }
+    );
+
+    if (!response.ok) {
```
lmuntaner, line 41: `response.ok` checks the 2xx range.

According to the spec only 200 is accepted. Could this be an issue?

lmuntaner (commented):
A couple of questions.

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
frederikrothenberger: Hm, good point. I don't think this is a problem regarding security. But I'll restrict it anyway. 👍

lmuntaner (approved):

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -175,26 +189,61 @@ export interface AuthSuccess {
...
     });
+  }
+
+  const validationResult = await withLoader(
+    async () =>
+      await validateDerivationOrigin(
+        authContext.requestOrigin,
+        authContext.authRequest.derivationOrigin
+      )
+  );
```
nmattia, line 208: nit: isn't this going to flash/restart the loader half way through, since we have to `withLoader`s back to back? or is there some user action that I'm missing?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -175,26 +189,61 @@ export interface AuthSuccess {
...
+  if (validationResult.result === "invalid") {
```
nmattia, line 209: Would be nice to also check for `valid` explicitly and/or make sure the match is exhaustive in case more alternatives are added, see this for instance: https://github.com/dfinity/internet-identity/blob/1939b306a307e62a2b0f2e76ea74cea983a0e6d4/src/frontend/src/flows/recovery/recoverWith/device.ts#L54-L65

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
+import { Principal } from "@dfinity/principal";
```
nmattia, line 1: Some doc would be nice

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export type ValidationResult =
+  | { result: "valid" }
+  | { result: "invalid"; message: string };
+
+export const validateDerivationOrigin = async (
+  authRequestOrigin: string,
+  derivationOrigin?: string
```
nmattia, line 9: I'm confused; what is the use case for `derivationOrigin === undefined`? why even call this function?

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (
+    derivationOrigin === undefined ||
+    derivationOrigin === authRequestOrigin
+  ) {
+    // this is the default behaviour -> no further validation necessary
+    return { result: "valid" };
+  }
+
+  // check format of derivationOrigin
+  const matches = /^https:\/\/([\w-]*)(\.raw)?\.ic0\.app$/.exec(
```
nmattia, line 20: What I gathered is that we only allow ic0.app origins for now, but that we may extend that to all origins. Would be nice to have a note about this here!

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+    derivationOrigin
+  );
+  if (matches === null) {
+    return {
+      result: "invalid",
+      message:
+        'derivationOrigin does not match regex "^https:\\/\\/([\\w-]*)(\\.raw)?\\.ic0\\.app$"',
```
nmattia, line 27: nit: extract the regex and use `regex.toString()` here to avoid mismatch

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+    };
+  }
+
+  try {
+    const canisterId = Principal.fromText(matches[1]); // verifies that a valid canister id was matched
```
nmattia, line 32: does this mean that all valid principals are valid canister IDs?

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+    // check for expected property
+    if (!Array.isArray(alternativeOriginsObj?.alternativeOrigins)) {
+      return {
+        result: "invalid",
+        message: `resource ${alternativeOriginsUrl} has invalid format: received ${alternativeOriginsObj}`,
+      };
+    }
+
+    // check allowed alternative origins
+    if (!alternativeOriginsObj.alternativeOrigins.includes(authRequestOrigin)) {
```
nmattia, line 61: looks like we're trying to handle malformed content. Should also display a nice error if the content is not JSON?

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+      result: "invalid",
+      message:
+        'derivationOrigin does not match regex "^https:\\/\\/([\\w-]*)(\\.raw)?\\.ic0\\.app$"',
+    };
+  }
+
+  try {
+    const canisterId = Principal.fromText(matches[1]); // verifies that a valid canister id was matched
+    const alternativeOriginsUrl = `https://${canisterId.toText()}.ic0.app/.well-known/ii-alternative-origins`;
+    const response = await fetch(
```
nmattia, line 34: does this automatically set `Accept` to `application/json`?

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (!alternativeOriginsObj.alternativeOrigins.includes(authRequestOrigin)) {
+      return {
+        result: "invalid",
+        message: `"${authRequestOrigin}" is not listed in the list of allowed alternative origins. Allowed alternative origins: ${alternativeOriginsObj.alternativeOrigins}`,
+      };
+    }
+  } catch (e) {
+    return {
+      result: "invalid",
+      message: `An error occurred while validation the derivationOrigin "${derivationOrigin}": ${e.message}`,
```
nmattia, line 70:
```diff
-message: `An error occurred while validation the derivationOrigin "${derivationOrigin}": ${e.message}`,
+message: `An error occurred while validating the derivationOrigin "${derivationOrigin}": ${e.message}`,
```

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
frederikrothenberger: This allows to not having to check whether `derivationOrigin` is `undefined` before calling.

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
frederikrothenberger: No, but only valid principals could potentially be canister ids. If the matched value is not even a principal it is certainly not a valid canister id. This is here to give developers early (and more specific) feedback if they have a malformed `derivationOrigin` (e.g. due to a typo).

I agree, the comment is misleading. I will update it when I revisit this code.

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
frederikrothenberger: I think the current error message is clear enough.

Example for the content `not-json`:
```
JSON.parse: unexpected keyword at line 1 column 1 of the JSON data
```

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
frederikrothenberger: Maybe I'm blind, but what's the difference here?

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
frederikrothenberger: No, it does not. Added to my notes, of things to refactor here. 👍

src/frontend/src/flows/authenticate/postMessageInterface.ts
```diff
@@ -23,6 +24,11 @@ export interface AuthResponseSuccess {
   userPublicKey: Uint8Array;
 }
 
+interface AuthResponseFailure {
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, otherwise `flows/authenticate/index.ts` would run into a typing error when sending back the error message.

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: Yes, it might. I can refactor that. 👍

src/frontend/src/flows/authenticate/index.ts
frederikrothenberger: Ah, right. Will do 👍

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
frederikrothenberger: Nvm, I'm blind 😅

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
nmattia: Heh, but then it still needs to be checked after calling right? And the value needs to be threaded through. Also in terms of separation of concerns I wonder if this function really cares about this. But 🤷

PR #737, opened 2022-07-12
Remove --emulator selenium runs

frederikrothenberger:
Currently, the --emulator flags are broken, because it does not actually use the emulator.
To fix it, we would have to change the whole docker compose setup again to allow for unpredictable canister ids.
Since the benefit of running the selenium tests on emulator is questionable anyway, I think the pragmatic solution is just to no longer run the --emulator selenium tests.


lmuntaner (approved):
rubber stamp

nmattia (approved):

PR #738, opened 2022-07-12
Build the test app in a separate CI step

frederikrothenberger:
This allows building II and the test app in parallel cutting CI times by about 3 minutes.


.github/workflows/canister-tests.yml
```diff
@@ -121,6 +121,34 @@ jobs:
             exit 1
           fi
 
+  test-app-build:
```
lmuntaner, line 124: maybe rename to `test-app-build-upload`?

lmuntaner (approved):
Cool! Just a comment being picky...

.github/workflows/canister-tests.yml
frederikrothenberger: I'd rather not, because it would make it inconsistent with `docker-build`.

PR #739, opened 2022-07-12
Tighten glob on canister tests cache

frederikrothenberger:
The previous value would also match the demos/test-app/cargo.lock leading to unnecessary cache misses.


lmuntaner (approved):
LGTM

PR #742, opened 2022-07-13
replace ClipboardJS with clipboard API

wackyleo459:
# Replacing clipboardJs library with async [clipboard API ](https://www.w3.org/TR/clipboard-apis/#async-clipboard-api)
- Motivation: Apparently ClipboardJS uses execCommand("copy") which is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)
- Changes: 
     - used navigator.clipboard.writeText async method
     - replicated the auto selection method that clipboardJs used - which selects the seed phrase (you see text highlighted in blue)
Screen Shot 2022-07-12 at 6 38 27 PM

### DIFF
--- a/src/frontend/src/flows/recovery/displaySeedPhrase.ts
+++ b/src/frontend/src/flows/recovery/displaySeedPhrase.ts
@@ -1,4 +1,3 @@
-import ClipboardJS from "clipboard";
 import { html, render } from "lit-html";
 import { checkmarkIcon, warningIcon } from "../../components/icons";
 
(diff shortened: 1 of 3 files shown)


frederikrothenberger (approved):
Thanks, LGTM!

src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -43,10 +42,31 @@ const init = (): Promise<void> =>
...
-    new ClipboardJS(seedCopy).on("success", () => {
-      const seedCopy = document.getElementById("seedCopy") as HTMLButtonElement;
-      displaySeedPhraseContinue.classList.toggle("hidden", false);
-      render(checkmarkIcon, seedCopy);
-      seedCopy.title = "copied";
+    const seedPhrase = document.getElementById("seedPhrase")
+      ?.innerText as string;
+
+    const selectText = (element: HTMLElement) => {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
nmattia, line 49: Why disable the eslint check?

frederikrothenberger: This is suppressing the not-null assertion. Essentially it is the same as writing `as HTMLElement` (like we do everywhere else). It is just more explicit about suppressing a check.

src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -43,10 +42,31 @@ const init = (): Promise<void> =>
...
+      selection.removeAllRanges();
+      selection.addRange(range);
+    };
+
+    seedCopy.addEventListener("click", () => {
+      navigator.clipboard
+        .writeText(seedPhrase)
+        .then(() => {
+          const seedPhraseDiv = document.getElementById("seedPhrase");
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
frederikrothenberger: See above. I think we would have to do way more `null` checking everywhere. But we can of course decide to change our stance on that...

PR #744, opened 2022-07-14
Add wasm files to gitignore

frederikrothenberger:


lmuntaner (approved):

nmattia (commented):
It was actually decided to not ignore the wasm files because (internal & external) people would run the build script and then wonder why no wasm was created. Can you clarify why we need to ignore the wasm files now?

PR #745, opened 2022-07-14
Remove headless option for selenium tests and improves information on test failures

frederikrothenberger:
The reason is that the clipboard permission is denied in headless mode. To keep screenshot sizes and aspect ratio the same, chrome is run in mobile emulation mode for the mobile tests.

The log output on selenium test failures has been optimized to be more concise and provide additional information in files. This was done, because the clipboard issue would have been found much quicker if the error output had been better.


.github/workflows/canister-tests.yml
```diff
@@ -289,12 +289,19 @@ jobs:
...
-        if: ${{ always() }}
+      - name: Upload screenshots
         uses: actions/upload-artifact@v2
         with:
           name: e2e-screenshots-${{ matrix.device }} ${{ matrix.start-flag }}
-          path: screenshots/**/*.png
+          path: screenshots/*.png
+
+      - name: Archive test failures
+        if: ${{ always() }}
```
lmuntaner, line 299: why is this needed and what does it do?

src/frontend/src/test-e2e/util.ts
```diff
@@ -50,13 +66,24 @@ export async function runInBrowser(
...
+    const printableLogs = browserLogs.reduce(
+      (accumulator, entry) => accumulator + "\n" + JSON.stringify(entry),
+      ""
     );
+    await fsasync.writeFile(`test-failures/${testName}.log`, printableLogs);
+    await browser.saveScreenshot(`test-failures/${testName}.png`);
+    console.error(e);
     console.log(
-      "An error occurred during e2e test execution. Logs can be found in the wdio.log file and an additional error screenshot was saved under screenshots/error. On Github Actions you can find the log and screenshots under 'Artifacts'."
+      "An error occurred during e2e test execution. WebDriver logs can be found in the wdio.log file and an error information (screenshot, console logs, page source) was saved in the test-failures folder. On Github Actions you can find the log and screenshots under 'Artifacts'."
```
lmuntaner, line 86: I'd love something like this in our e2e tests as well

lmuntaner (approved):
LGTM, just one question

.github/workflows/canister-tests.yml
frederikrothenberger: So I have changed the file structure of the selenium errors:
It previously used to be `screenshots/<name>.png` for normal screenshots and `screenshots/error/<time>.png` for errors. Now its `screenshots/<name>.png` for normal screenshots. And a new artifact that collects a screenshot, the page source of the page that failed and the browser logs at that moment in `test-failures/<test_name>.png|html|log`.

And this is why I had to touch the github actions workflow.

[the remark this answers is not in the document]
frederikrothenberger: see above

lmuntaner: but why the `if: always()`? This looks like `if (true)` when I read it. That's why I'm asking.

frederikrothenberger: Ah, normally github actions just aborts the workflow if it fails. But for test-failure artifacts we absolutely want the files anyway. So it's `always` as opposed to the default which is only on success.

lmuntaner: aaahhh, ok, got it. Thanks!

src/frontend/src/test-e2e/util.ts
```diff
@@ -23,17 +29,27 @@ export async function runInBrowser(
...
+  } else {
+    chromeOptions.args?.push(
+      `--window-size=${runConfig.screenConfiguration.windowSize}`
+    );
+  }
+
   const browser = await remote({
     capabilities: {
       browserName: "chrome",
-      "goog:chromeOptions": {
```
nmattia, line 29: why did this move? also do we now run with mobile emulation? probably worth mentioning in the PR description

src/frontend/src/test-e2e/util.ts
```diff
@@ -1,17 +1,23 @@
...
+    width: 360,
+    height: 640,
+  },
 };
 
 // desktop resolution is used when env variable SCREEN=desktop is set
 const DESKTOP_SCREEN: ScreenConfiguration = {
   screenType: "desktop",
-  windowSize: "1920,1080",
+  windowSize: "1928,1209", // increased from 1920,1080 when the headless flag was removed to account for the GUI border of 8,129.
```
nmattia, line 20: future contributors won't care that there was a headless flag; also what about the GUI border? why trim it?

src/frontend/src/test-e2e/util.ts
```diff
@@ -371,6 +412,16 @@ export async function switchToPopup(
   return await addVirtualAuthenticator(browser);
 }
 
+/**
+ * Switches to the first window of the given browser. Explicit window focusing is required when using the mobile emulation mode of chrome.
```
nmattia, line 416: Do you know why focusing is required? Seems odd that we need to explicitly focus

nmattia (commented):
Can we link to some doc about why the clipboard permission can not be used in headless mode?

src/frontend/src/test-e2e/util.ts
frederikrothenberger: Agreed, the idea initially was to keep the diff smaller. But apparently there are minor changes in font rendering when not using headless mode. So this point is moot. I Will create a PR reverting this to 1920*1080.

src/frontend/src/test-e2e/util.ts
frederikrothenberger: This moved in order to more easily set different options depending on the run configuration. The PR description already mentions that mobile emulation is turned on.

src/frontend/src/test-e2e/util.ts
frederikrothenberger: Yes, the mobile emulation comes with it's own sets of weird quirks. If this is removed, the browser just gets stuck on the wrong screen. I'm not sure why the chrome developer chose this behavior.

frederikrothenberger:
> Can we link to some doc about why the clipboard permission can not be used in headless mode?

Unfortunately, I did not find any documentation of this not being possible, only a that it does not work in practice and other people having the same issues: e.g. https://github.com/DevExpress/testcafe/issues/6982#issuecomment-1101475783

src/frontend/src/test-e2e/util.ts
frederikrothenberger: #750

PR #747, opened 2022-07-15
Add usage warning to the alternative origins feature

frederikrothenberger:
Additionally, the validation regex now ensures a non empty canister id.

### DIFF
--- a/src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
+++ b/src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
@@ -17,18 +17,24 @@ export const validateDerivationOrigin = async (
   }
 
   // check format of derivationOrigin
-  const matches = /^https:\/\/([\w-]*)(\.raw)?\.ic0\.app$/.exec(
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -17,18 +17,24 @@ export const validateDerivationOrigin = async (
   }
 
   // check format of derivationOrigin
-  const matches = /^https:\/\/([\w-]*)(\.raw)?\.ic0\.app$/.exec(
+  const matches = /^https:\/\/([\w-]+)(\.raw)?\.ic0\.app$/.exec(
```
nmattia, line 20: unrelated to this PR, but shouldn't `(\.raw)` be a non-capturing group? or is that used somewhere?

nmattia (approved):

src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
nmattia: just checking, no need! thanks for the clarification

PR #750, opened 2022-07-19
Revert desktop screen size to 1920*1080

frederikrothenberger:
Initially changed to keep desktop screenshots out of the diff. But due to minor changes in font rendering this did not work.


nmattia (approved):
LGTM but why was it changed in the first place?

frederikrothenberger:
> LGTM but why was it changed in the first place?

It was initially changed to keep screenshot size the same (to reduce the diff).

Issue #651, opened 2022-05-21
Request to include candid file in release assets


paulyoung:
If the `.did` file was also a release asset it would make setting up internet identity as a remote canister much simpler.

The current releases require 2 separate downloads; one for the Wasm and another for the source archive, which also requires unpacking it.

paulyoung:
To illustrate:

```nix
{
  internetIdentity = options: pkgs.runCommand "internet-identity-${options.flavor}-${options.release}" {} ''
    mkdir $out
    cp -R ${internetIdentityDid options.release options.tarGzSha256}/. $out
    cp -R ${internetIdentityWasm options.release options.wasmSha256 options.flavor}/. $out
  '';

  internetIdentityWasm = release: sha256: flavor: pkgs.stdenv.mkDerivation rec {
    inherit release;
    name = "internet-identity-${flavor}-wasm";
    src = pkgs.fetchurl {
      inherit sha256;
      url = "https://github.com/dfinity/internet-identity/releases/download/${release}/internet_identity_${flavor}.wasm";
    };
    phases = [ "installPhase" ];
    installPhase = ''
      mkdir $out
      cp $src $out/internet_identity_${flavor}.wasm
    '';
  };

  internetIdentityDid = release: sha256: pkgs.stdenv.mkDerivation rec {
    inherit release;
    name = "internet-identity-did";
    src = pkgs.fetchzip {
      inherit sha256;
      url = "https://github.com/dfinity/internet-identity/archive/refs/tags/${release}.tar.gz";
    };
    installPhase = ''
      mkdir $out
      cp internet-identity-${release}/src/internet_identity/internet_identity.did $out
    ''
    ;
  };
}
```

nmattia:
Hey @paulyoung,

yeah that'd make sense. To clarify, you want an archive with both the candid file and the wasm module(s) so that you can use a single sha256, is that right?

We could add the candid as an asset, as as well a tarball that includes all three wasm modules + the candid description. WDYT?

paulyoung:
I was thinking that adding it as another release asset would make things easier but I guess a single archive would be most convenient.

nmattia:
I think that adding it as another release asset makes most sense, but then I thought that wouldn't help for your workflow because you'd still need to add another sha256. Did I misunderstand?

paulyoung:
I guess it depends if other workflows want to download the assets directly rather than having to unpack them.

e.g. https://github.com/dfinity/sdk/issues/2085

[...]

PR #754, opened 2022-07-19
Don't symlink to candid file from demo

### LINKED ISSUE #725: Cannot run using-dev-build demo as a standalone project
I tried running the demo as standalone:

1. I copied the folder [using-dev-build](https://github.com/dfinity/internet-identity/blob/main/demos/using-dev-build) in a separate directory.
2. I run the commands in the [README.md](https://github.com/dfinity/internet-identity/blob/main/demos/using-dev-build/README.md).
3. I got this error after the command `dfx deploy --no-wallet --argument '(null)'`:
  ```
  Error: Failed while trying to deploy canisters.
  Caused by: Failed while trying to deploy canisters.
    Failed to build call canisters.
      Failed while trying to build all canisters.
        The post-build step failed for canister 'rwlgt-iiaaa-aaaaa-aaaaa-cai' (internet_identity) with an embedded error: No such file or directory (os error 2) 
  ```

I found out that the problem was due to the symbolic link of `internet_identity.did` which is pointing to a file that doesn't exist anymore (as I'm using the using-dev-build as standalone). To fix this I substituted it with a copy of the [file](https://github.com/dfinity/internet-identity/blob/main/src/internet_identity/internet_identity.did) pointed by the symbolic link.

I suggest adding instruction to copy the file `internet_identity.did` to the README.md so that you can actually run the demo as standalone.

nmattia:
Fixes #725 

Before this the candid file was symlinked from `using-dev-build`. We do
however suggest to copy the directory out of the repo for users to start
their II-enabled projects, which would lead to a broken symlink.

Instead of suggesting to copy the target file as well, we simply put the
candid file in the release (as also suggested in https://github.com/dfinity/internet-identity/issues/651) and download it
like we download the wasm file.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -344,27 +344,48 @@ jobs:
           path: .
 
       - name: Deploy II and run tests
-        working-directory: demos/using-dev-build
         run: |
           set -euo pipefail
 
+          # Copy example to make sure it does not rely on living inside the II repo
+          builddir=$(mktemp -d)
+          cp -r ./demos/using-dev-build/. "$builddir"
+          pushd "$builddir"
+
+          # Install npm deps
           npm ci
 
-          # Create a fake curl so that we use this build's wasm; also create a "witness"
-          # file to make sure the fake curl was run (rm $witness will fail otherwise)
-          curl_dir=$(mktemp -d); witness=$(mktemp)
+          # Create a fake curl so that we use this build's wasm and did files; also create "witness"
+          # files to make sure the fake curl was run (rm $witness_foo will fail otherwise)
+          curl_dir=$(mktemp -d); witness_wasm=$(mktemp)
+          curl_dir=$(mktemp -d); witness_did=$(mktemp)
           cat > "$curl_dir/curl" << EOF
           #!/usr/bin/env bash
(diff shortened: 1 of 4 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -344,27 +344,48 @@ jobs:
...
+
+          # Install npm deps
           npm ci
 
-          # Create a fake curl so that we use this build's wasm; also create a "witness"
-          # file to make sure the fake curl was run (rm $witness will fail otherwise)
-          curl_dir=$(mktemp -d); witness=$(mktemp)
+          # Create a fake curl so that we use this build's wasm and did files; also create "witness"
+          # files to make sure the fake curl was run (rm $witness_foo will fail otherwise)
+          curl_dir=$(mktemp -d); witness_wasm=$(mktemp)
```
frederikrothenberger, line 360: How does this witness mechanism work exactly? `witness_wasm=$(mktemp)` creates the file already, so having executed this line `rm $witness_wasm` will not fail, even if the `touch $witness_wasm` is not executed. Or am I missing something?
Also you are creating two `curl_dir`s. This is not necessary, right?

nmattia: That's absolutely correct! Good catch

frederikrothenberger (approved):
LGTM, thanks!

PR #761, opened 2022-07-21
Remove webauthn compatibilitiy check when DUMMY_AUTH is enabled

frederikrothenberger:
The check unnecessarily restricts devs when using II in their dev environment.


nmattia (approved):
Makes sense! Out of curiosity, how did this come up?

frederikrothenberger:
> Makes sense! Out of curiosity, how did this come up?

We were notified about it on slack: see details in L2-710.

PR #762, opened 2022-07-22
Limit the number of alternative origins to 10

frederikrothenberger:
This should not impact developers just moving from one domain to another, but will prevent merging principal spaces of many unrelated applications
(which this feature is not intended for).


src/frontend/src/flows/authenticate/validateDerivationOrigin.ts
```diff
@@ -1,6 +1,7 @@
 import { Principal } from "@dfinity/principal";
 
 const ORIGIN_VALIDATION_REGEX = /^https:\/\/([\w-]+)(?:\.raw)?\.ic0\.app$/;
+const ALTERNATIVE_ORIGINS_LIMIT = 10;
```
nmattia, line 4: nit:

```suggestion
const MAX_ALTERNATIVE_ORIGINS = 10;
```

to clarify it's a maximum; also more consistent with other constants we have

nmattia (approved):

PR #763, opened 2022-07-22
Fix candid spec upload

nmattia:
I had a brainfart and asked our release script to build/prepare the
candid interface. We just need to upload it, there's nothing to prepare.


frederikrothenberger:
So the effect of this was, that it was listed in the release notes table?

nmattia:
The effect was that the release script tried to figure out the build step that produced this and failed

frederikrothenberger (approved):

PR #764, opened 2022-07-22
Update spec to reflect alternative origins limit

frederikrothenberger:
The limit was introduced in #762.


nmattia (approved):
Good thinking

PR #765, opened 2022-07-22
Include glibc version in cache key

nmattia:
This includes the glibc version in the bootstrap cache key, to avoid
issues like this: https://github.com/dfinity/internet-identity/actions/runs/2717048520

While it's unclear why the issue actually occurred, the cache key is now
human readable for easier debugging in the future.


mstrasinskis (approved):
Rubber stamp

PR #767, opened 2022-07-25
Fix path to candid description in release

nmattia:
We added the candid description as a released asset, however there were
two issues:

* The (relative) path to the `.did` was incorrect,
* The release script didn't support files outside of the top-level
  directory.


peterpeterparker (approved):
Rubber stamp

PR #769, opened 2022-07-26
Pin ic-cdk-optimizer version

mraszyk:
This pins the `ic-cdk-optimizer` version to ensure builds are optimized consistently, in the name of reproducibility. This also upgrades `ic-cdk-optimizer` to `0.3.4` in the build and bootstrap scripts.


scripts/build
```diff
@@ -61,15 +61,9 @@ do
...
-# because of https://github.com/dfinity/cdk-rs/issues/181
-# Once the issue is fixed, we can ensure that the correct version is installed
-if ! command -v ic-cdk-optimizer
+if [[ ! "$(command -v ic-cdk-optimizer)" || "$(ic-cdk-optimizer --version)" != "ic-cdk-optimizer 0.3.4" ]]
 then
-    echo could not find ic-cdk-optimizer
-    echo "ic-cdk-optimizer version 0.3.1 is needed, please run the following command:"
-    echo "  cargo install ic-cdk-optimizer --version 0.3.1"
-    exit 1
+    cargo install ic-cdk-optimizer --version 0.3.4
```
nmattia, line 66: This will install the tool globally for the user, let's just let them know how they can install it instead

scripts/bootstrap
```diff
@@ -23,8 +23,8 @@ echo "using rust version '$rust_version'"
 run curl --fail https://sh.rustup.rs -sSf | run sh -s -- -y --default-toolchain "none" --no-modify-path
 
 echo "looking for ic-cdk-optimizer"
-if ! command -v ic-cdk-optimizer
+if [[ ! "$(command -v ic-cdk-optimizer)" || "$(ic-cdk-optimizer --version)" == "ic-cdk-optimizer 0.3.4" ]]
 then
     echo "installing ic-cdk-optimizer"
-    run cargo install ic-cdk-optimizer --version 0.3.1
+    run cargo install ic-cdk-optimizer --version 0.3.4
```
nmattia, line 29: Here the message still says `looking for/installing ic-cdk-optimizer`, I think it'd be worth mentioning the version (otherwise the user may get confused as to why the script is re-installing the tool)

nmattia (commented):
Thanks!

scripts/build
mraszyk: I thought the build script is intended to be used inside a Docker container and not directly on the user's host machine.

scripts/bootstrap
mraszyk: Done.

scripts/build
nmattia: The build script is first and foremost meant to be called from by `dfx` and hence also on the user's host machine

mraszyk: The README file already points to using Docker for the build and this could become the recommended way of running the build.

mraszyk: I'm not installing ic-cdk-optimizer in the user build script anymore.

nmattia (approved):

scripts/bootstrap
```diff
@@ -22,9 +22,9 @@ echo "using rust version '$rust_version'"
 # here we set the toolchain to 'none' and rustup will pick up on ./rust-toolchain.toml
 run curl --fail https://sh.rustup.rs -sSf | run sh -s -- -y --default-toolchain "none" --no-modify-path
 
-echo "looking for ic-cdk-optimizer"
-if ! command -v ic-cdk-optimizer
+echo "looking for ic-cdk-optimizer 0.3.4"
+if [[ ! "$(command -v ic-cdk-optimizer)" || "$(ic-cdk-optimizer --version)" == "ic-cdk-optimizer 0.3.4" ]]
```
nmattia, line 26: won't this try to install v0.3.4 if it's already installed, and do nothing if the version is different than 0.3.4?
```suggestion
if [[ ! "$(command -v ic-cdk-optimizer)" || "$(ic-cdk-optimizer --version)" != "ic-cdk-optimizer 0.3.4" ]]
```

nmattia (changes requested):

scripts/bootstrap
mraszyk: Good catch! Your patch is what `scripts/build` looks like.

nmattia (approved):
Thanks!

PR #772, opened 2022-07-28
Don't thread userNumber for IIConnection

nmattia:
This lets IIConnection store the `userNumber` (which never changes) as
opposed to specifying it for every call.

The `getPrincipal` function was also removed because it wasn't being
used.


mstrasinskis (approved):
LGTM, nice improvement!

PR #773, opened 2022-07-28
Script "setupJs" with async as true

wackyleo459:
Have the `<script id=SetupJs>` changed to async, to enable lazy loading behavior.

Note: Setting to type="module" causes the error of `canisterId is undefined`.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attributes

Screen Shot 2022-07-27 at 10 30 07 PM


nmattia (commented):
Thanks!

@peterpeterparker I think you were the one who reported this. Is `async` better than `defer`? Here the script is loaded at the end of the page, so I don't think what we do is very important, but:
* IIUC with neither `async` nor `defer` we run the risk of the JS kicking in a little late and the page will feel unresponsive
* With `async` we could run the risk that some (HTML) elements may not be parsed/rendered yet before the JS kicks in, meaning the JS could fail (although here the only relevant elements seem to appear before `<script>`
* With `defer` we'd be safer since the JS wouldn't be evaluated before the HTML has finished loading.

@wackyleo459 @peterpeterparker am I getting this correctly? Wdyt?

peterpeterparker:
this gist is interesting regarding `async`, `defer` or `module` 👉 https://gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6

> IIUC with neither async nor defer we run the risk of the JS kicking in a little late and the page will feel unresponsive
> With async we could run the risk that some (HTML) elements may not be parsed/rendered yet before the JS kicks in, meaning the JS could fail (although here the only relevant elements seem to appear before <script>

generally speaking I rather like UI that are not blocked by JS so I use `module` or `async` but yes, in such case, the UI should amend the fact that the script might not be loaded when rendered. So not sure if it is the case here or not, I miss a bit of context.

> With defer we'd be safer since the JS wouldn't be evaluated before the HTML has finished loading.

not sure that would work, i don't know when above script is added. e.g. 

```
<body>
  ...
  <script defer />

!==

<body>
  ...
   <script>
      const script = document.createScript
      script.defer = true
      head.append(script)
```

But again, I probably miss some context.

nmattia (approved):

wackyleo459:
As per discussion, current solution would suffice at this time.

PR #774, opened 2022-07-28
set width for container div

wackyleo459:
Issue: Inside device setting screen: depending on the length of device name, the overall container width changes (including buttons). Sometimes it's too small. Moreover, the name would overflow the container.

Changes : 
- .container div has a standard width 365px (or if screen is smaller, the full width of screen).
- h1 width is controlled with: `word-break: break-word`
- if device name gets long , it will expand to multiple lines.

Changes

 
Before any changes


nmattia (commented):
Thanks for the PR!

If I remember correctly, the issue was that those buttons would grow in width with the size of the device name, i.e. a device named "foo" would have tiny buttons, and a device named "foo bar baz quux" would have wide buttons.

Here in practice the issue is fixed, because the buttons are always about 347px wide. One question though, why set a min-width instead of a width then? Also, do you think it would make sense to use something other than pixels?

There's also something odd: it looks like the page itself got wider than! The `.container` appears to have a min-width as well, set to `300px`. Does this mean the buttons drive the size of the page now? 😅

Curious what you and @mstrasinskis think!

Also worth noting: I think @meodai will take a stab at cleaning up the CSS more generally, not sure if he got started yet!

meodai:
@nmattia I will yes! In I would probably have missed the issue when the text is too long for the button. I think there are a few possible ways to address this: 

a) Set a `width` and use 
```css
text-overflow: ellipsis;
overflow: hidden;
white-space: nowrap;
``` 
to make sure the button nothing funky happens if the contained text is too long. (maybe I would be nice to set the full name as a title prop as well, so it can still be read by a desktop user at least)

or

b) Just accept that the button can be higher if something does not fit the width, using something like: `word-break: break-all;`


In every case I would set the `width: 100%;` and not to a fixed pixel value, since it would be nice not having to touch every single element if we ever change the container width.

wackyleo459:
Since the button width change is due to a larger problem that the enclosing container width changes depending on its content (such as h1 element), first solution was have the container have a fixed width. 

- main .container div has fixed width 365px (side padding was increased slightly, due to too little padding during mobile view, which does not significantly change other elements)
- h1 width is controlled with: word-break: break-all
- device will have auto x-scroll if name overflows.

Padding before and after
Smaller padding
N1

mstrasinskis (changes requested):
Horizontal scrollbars are not so usual nowadays. What do you think about always displaying the whole name in multiple lines? 

image

mstrasinskis (approved):
LGTM. Thanks a lot 👍

PR #775, opened 2022-07-28
submit button contained in form

wackyleo459:
New Device Name screen:

Issue: When typing the device name, cannot submit by keyboard "enter" button

Resolution: 
 - Wrapped  input and buttons within a form, enabling auto submit for key enter
 - Fixed: Even when there was no name given, submit button would fire which resulted in an empty device name. Now if no text is given form will not be submitted.


src/frontend/src/styles/main.css
```diff
@@ -651,3 +651,7 @@ nav {
 button[data-action="protect"] {
   margin-bottom: 0;
 }
+
+input {
+  line-height: 1em;
```
nmattia, line 656: is this related to the PR?

nmattia (commented):
Thanks!


In [other places](https://github.com/dfinity/internet-identity/blob/5ac2973fefd2c0a673509f38445b74c5652ae8a1/src/frontend/src/flows/register.ts#L57) in the codebase we `preventDefault()` when submitting the form. Is that necessary? Or can you just perform the action in `onsubmit` like you do here?

src/frontend/src/styles/main.css
wackyleo459: Yes it is, the input line-height decreases by default when the input gets contained inside <form>. 
I created a custom class name, and applied styling to just this specific input [here](https://github.com/dfinity/internet-identity/pull/775/commits/ee92e6ceb80895529cca024296d53645e124e5ef)

wackyleo459:
> Thanks!
> 
> In [other places](https://github.com/dfinity/internet-identity/blob/5ac2973fefd2c0a673509f38445b74c5652ae8a1/src/frontend/src/flows/register.ts#L57) in the codebase we `preventDefault()` when submitting the form. Is that necessary? Or can you just perform the action in `onsubmit` like you do here?

We discussed it's safer to still include preventDefault(), although it works without because there is no actual html form data submission.

nmattia (approved):
Looking great, thanks!

PR #777, opened 2022-07-29
Revert "Remove CODEOWNERS"

nmattia:
Reverts dfinity/internet-identity#759

For visibility we still want to have the (empty) CODEOWNERS so that people know how to reach us.


lmuntaner (approved):
Does that mean we can do the same in all the repos?

nmattia:
@lmuntaner unclear... let's wait and see

PR #778, opened 2022-07-29
Show version info in console

nmattia:
This bakes the commit sha (if possible) and shows it in the browser
console for easily figuring out what version is running.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,6 +7,34 @@ const HttpProxyMiddlware = require("http-proxy-middleware");
 const dfxJson = require("./dfx.json");
 require("dotenv").config();
 
+const execSync = require('child_process').execSync;
+
+/**
+ * Try to come up with meaningful version info based on Git status.
+ *
+ * This returns either:
+ *  * 'unknown' on error
+ *  * 'dirty' if dirty or not on branch 'main'
+ *  * the commit sha1 otherwise
+ */
+function versionInfo() {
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -7,6 +7,34 @@ const HttpProxyMiddlware = require("http-proxy-middleware");
...
+ *  * 'dirty' if dirty or not on branch 'main'
+ *  * the commit sha1 otherwise
+ */
+function versionInfo() {
+    try {
+        // Grab the branch and trim newline; on failure (for instance not a git repo, git not installed, etc)
+        // just say "unknown"
+        const branch = execSync('git rev-parse --abbrev-ref HEAD', { encoding: 'utf-8'} ).replace(/^\s+|\s+$/g, '');
+        try {
+            if (branch !== "main") { throw Error('not on main'); }
```
peterpeterparker, line 26: "not on main" will be reported as "dirty"...just in case it's not expected

peterpeterparker (approved):
LGTM

webpack.config.js
nmattia: Yes, that's expected! I thought about putting the branch name or something but often the branch will later be deleted etc, so I thought it'd be simpler to say "this is not a clean, "main" build". Good catch though!

nmattia:
Just realized there's no git info in the (official) docker build, will need to inject that

PR #780, opened 2022-07-30
prepare cover build

mraszyk:
Update to make internet_identity build compatible with cover: https://mzsvw-zqaaa-aaaam-qafqq-cai.ic0.app/

Note that dfx version is set to 0.10.0 in the cover builder to prevent postprocessing done by dfx after executing script/build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,7 +70,7 @@ RUN touch src/canister_tests/src/lib.rs
 RUN npm ci
 
 RUN ./scripts/build
-RUN sha256sum /internet_identity.wasm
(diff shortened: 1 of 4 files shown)


nmattia (commented):
Hi @mraszyk ! Thanks for the contribution.

A cover build is not very important to us right now, and this adds a bunch of complexity and changes -- including for instance a new canister ID for mainnet. Is this something you actually intend to merge? If so, can you give some context and motivation, and if not can you close the PR?

Thanks!

mraszyk:
I've reverted the canister ID on the mainnet - it was an oversight on my side to change it.

mraszyk:
We're going to keep the cover build setup in a separate fork for now.

PR #783, opened 2022-08-02
Improve HTML semantics

meodai:
- improves the use of HTML5 tags and accessibility. 
- some CSS needed to be edited, since some of the rules used an element selector


src/frontend/src/flows/manage/index.ts
```diff
@@ -130,50 +152,57 @@ const style = () => html`<style>
 // recovery devices.
 const pageContent = (userNumber: bigint, devices: DeviceData[]) => html`
   ${style()}
-  <div class="container">
+  <section class="container">
```
wackyleo459, line 155: Is this purposeful, having the `container` div be a section sometimes, and sometimes an article as in [here](https://github.com/dfinity/internet-identity/blob/51968c98bb6c68e20bbecc68e9af484edff9840f/src/frontend/src/flows/addDevice/manage/index.ts#L45)? Is it because of different purpose in the flows? 
I initially thought it would make sense to keep as one consistent type.

src/frontend/src/flows/manage/index.ts
```diff
@@ -130,50 +152,57 @@ const style = () => html`<style>
...
-      >
-      <button
-        ?disabled=${numAuthenticators(devices) >= MAX_AUTHENTICATORS}
-        class="labelAction"
-        id="addAdditionalDevice"
-      >
-        <span class="tooltip"
-          >You can register up to ${MAX_AUTHENTICATORS} authenticator devices.
-          Remove a device before you can add a new one.</span
+    <aside>
```
wackyleo459, line 161: Curious about the role of aside here, how it's different from role of an article? (I'm still quite new with semantics and would like to understand your approach!)

wackyleo459 (approved):
Thanks for going throughly over the semantics usage. I left a few observations/questions.
Other than that everything looks great!

src/frontend/src/flows/manage/index.ts
meodai: Yes it is. Its to transport the meaning of the information it contains

> Represents a generic document or application section. In this context, a section is a thematic grouping of content, typically with a header, possibly with a footer. Examples include chapters in a book, the various tabbed pages in a tabbed dialog box, or the numbered sections of a thesis. A web site's home page could be split into sections for an introduction, news items, contact information. [Section HTML5 Doctor](http://html5doctor.com/element-index/#s)

The article is more of a single widget, that could be anywhere else on the website. I thought it was more appropriate in this case. But I think its not a clear cut in this case. Since it contains two things that are quite different semantically I thought it would be nice to group them loosely.

I hope this answers your question. Don't hesitate to ping me if not.

src/frontend/src/flows/manage/index.ts
meodai: An aside is used whenever an information does not make much sense on its own. In other words: You need the context (the article/section around) to understand its content.

meodai:
I had to recreate this pull-request because otherwise the CI would run on my own hours :D https://github.com/dfinity/internet-identity/pull/789

PR #789, opened 2022-08-04
Improve HTML semantics

meodai:
- improves the use of HTML5 tags and accessibility.
- some CSS needed to be edited, since some of the rules used an element selector


nmattia (approved):
Nice, thanks!

meodai:
I've shrunken some of the labels. But I think its still works nicely. I change them a bit again on the CSS rework

src/frontend/src/flows/manage/index.ts
```diff
@@ -48,19 +49,19 @@ const style = () => html`<style>
     text-align: right;
   }
 
-  .labelWithAction label {
+  .labelWithAction .labelWithAction {
```
wackyleo459, line 52: Was this intended? Wasn't sure if this was typo.

wackyleo459: @meodai

PR #790, opened 2022-08-04
Update II spec links

nmattia:
The spec has moved, and the official page is the one on
[internetcomputer.org](https://internetcomputer.org/docs/current/references/ii-spec). The code still lives in this repo.

EDIT: for context, see also https://github.com/dfinity/internet-identity/tree/main/docs#specification


frederikrothenberger (approved):
LGTM

PR #791, opened 2022-08-04
Deploy main build automatically

nmattia:
This adds automated deploys on the main branch. This deploys to our test
canister on mainnet.


frederikrothenberger (approved):
We might want to start having a second mainnet canister then, where we can deploy branches, specific versions etc. manually.

nmattia:
Agreed, although we don't deploy manually that often. Let's see how it works in practice.

PR #795, opened 2022-08-04
Show version info in browser console

nmattia:
This adds a browser console log showing the version of the deployed canister. In particular:

* A new endpoint, `/version`, is added to the canister's HTTP interface. What this endpoint returns is set by the deployment arguments to the canister.
* The webapp fetches `/version` on startup and displays its content asynchronously (@peterpeterparker can you double check the frontend just in case I did something stoopid?)
* The `deploy` job bakes in the commit when deploying.

### DIFF
--- a/src/frontend/src/index.ts
+++ b/src/frontend/src/index.ts
@@ -10,8 +10,23 @@ import { recoveryWizard } from "./flows/recovery/recoveryWizard";
 import { showWarningIfNecessary } from "./banner";
 import authorizeAuthentication from "./flows/authenticate";
 
+/**
+ * Shows some version information, retrieved from /version
+ */
+const showVersion = async () => {
+  const response = await fetch("/version");
+  if (response.ok) {
+    const versionInfo = await response.text();
+    console.log(`version: ${versionInfo}`);
+  } else {
+    throw Error(`Could not read version info from /version: ${response}`);
+  }
+};
+
 const init = async () => {
   const url = new URL(document.URL);
+
+  showVersion().catch((e) => console.log(e));
 
   // If the build is not "official", show a warning
   // https://github.com/dfinity/internet-identity#build-features
--- a/src/frontend/src/index.ts
+++ b/src/frontend/src/index.ts
@@ -26,6 +26,8 @@ const showVersion = async () => {
 const init = async () => {
   const url = new URL(document.URL);
 
+  // Note: we don't await anything here because we don't want to block anything while
+  // waiting for the version information to be shown
   showVersion().catch((e) => console.log(e));
 
   // If the build is not "official", show a warning
--- a/src/internet_identity/src/http.rs
+++ b/src/internet_identity/src/http.rs
@@ -58,13 +58,14 @@ pub fn http_request(req: HttpRequest) -> HttpResponse {
         // Technically this should only accept requests with 'accept: text/plain' but we don't
         // check headers anywhere else so neither do we here.
         "/version" => {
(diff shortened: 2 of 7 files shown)


src/internet_identity/src/http.rs
```diff
@@ -53,6 +53,28 @@ pub fn http_request(req: HttpRequest) -> HttpResponse {
                 },
             }
         }
+
+        // the /version endpoint returns information about the version.
+        // Technically this should only accept requests with 'accept: text/plain' but we don't
+        // check headers anywhere else so neither do we here.
+        "/version" => {
+            let version_info =
+                crate::VERSION_INFO.with(|version_info| version_info.borrow().clone());
```
nmattia, line 62: I'm not super happy with this `.clone()` but the `body` of HttpResponse either needs a `'static` (which is not ideal here) or for the data to be owned. I have no idea why the type is set up as is right now and will need to do some digging (independently of the PR).

src/internet_identity/src/main.rs
```diff
@@ -175,6 +175,9 @@ struct Challenge {
 thread_local! {
     static STATE: State = State::default();
     static ASSETS: RefCell<HashMap<&'static str, (Vec<HeaderField>, &'static [u8])>> = RefCell::new(HashMap::default());
+    /// This build's version info, provided on startup through arguments.
```
frederikrothenberger, line 178: Is there a specific reason why you chose to allow setting the version at startup / upgrade? Because this allows to mismatch the `wasm` file and the version. Could we not bake the version into the `wasm` during the build? I think, in all contexts where II gets built we do have access to the git hash.

src/frontend/src/index.ts
```diff
@@ -10,9 +10,26 @@ import { recoveryWizard } from "./flows/recovery/recoveryWizard";
 import { showWarningIfNecessary } from "./banner";
 import authorizeAuthentication from "./flows/authenticate";
 
+/**
+ * Shows some version information, retrieved from /version
+ */
+const showVersion = async () => {
+  const response = await fetch("/version");
```
frederikrothenberger, line 17: This won't work in practice on mainnet. The `/version` has to be certified otherwise the service worker will reject the payload. Since asset hashes are also recalculated on upgrade and install, just certifying the version there as well should not be a problem, right?
(Note: just using raw does not help, as the service worker will require certification for calls to raw on pages loaded on non-raw domains.)

If you use `agent-js` to make the query call instead of using fetch, you could also bypass the certification requirement. Not a huge fan of that approach though.

Alternatively, we could just not print the version in the front-end code. And if somebody is interested in what the version is, they could just manually visit https://identity.raw.ic0.app/version (i.e. the same approach as for /metrics).

src/internet_identity/src/http.rs
frederikrothenberger: I don't think cloning such a small value is a problem. But yeah, we should definitely find out and document the reasoning behind the typing.

frederikrothenberger (commented):
Left a few comments. Having the deployed version available would be great. 👍

src/internet_identity/src/main.rs
nmattia: In general it's not advised, because the build should not depend on being run in Git. In particular, we do not have access to `.git` in docker.

src/frontend/src/index.ts
nmattia: I think I'll just bake in the version in the assets, that was my first idea but our asset types are a bit silly. Will give it another try.

nmattia:
I'm closing this for now, will revisit when I have time!

Issue #785, opened 2022-08-02
CSS reset


wackyleo459:
Due to possible inconsistent styling across different browsers, suggestion is to have css reset styling. For our purposes, having a custom separate css reset stylesheet with needed normalized styling (such as lists) should do the trick.

Guidelines:
- https://www.webfx.com/blog/web-design/css-tip-1-resetting-your-styles-with-css-reset/
- https://meyerweb.com/eric/tools/css/reset/index.html

Possible tool/option:
https://necolas.github.io/normalize.css/

meodai:
I think we agreed on using a reset not a normalize. I usually just use that one https://github.com/elad2412/the-new-css-reset

wackyleo459:
https://github.com/dfinity/internet-identity/pull/796

PR #796, opened 2022-08-04
Css reset

wackyleo459:
Issue discussed here: https://github.com/dfinity/internet-identity/issues/785

# Changes
- Reset styling for different browser styles, by using "the-new-css-reset" stylesheet that gets copied into main.css.

- `main.css` has a few added styling to restore some large discrepancies from previous style.


meodai:
Nice job! I like that you choose to add it as a dependency.

src/frontend/src/styles/main.css
```diff
@@ -156,9 +157,9 @@ button {
   margin-bottom: 1rem;
   padding: 0.6rem 1rem;
   border-radius: 4px;
-
   line-height: 1.25rem;
   letter-spacing: 0.005em;
+  font-family: Arial;
```
meodai, line 162: I know that it was `Arial` before. But I think it look nicer if its the same font then the rest of the site no? I think it was an omission before.

meodai:
I have chosen to add the reset within the `main.css` in my branch:
https://github.com/dfinity/internet-identity/blob/feature/css-cleanup/src/frontend/src/styles/main.css

There are some reasons why I think this make sense: 
- I don't see an other way to follow the ITCSS order: https://csswizardry.com/2018/11/itcss-and-skillshare/ (to reconsider if we ever introduce SCSS or/and PostCSS)
- security: its one less npm dependency that could potentially cause harm to the end-user.

Do you think that would also be a good idea in this branch?

src/frontend/src/styles/main.css
wackyleo459: Yup, sure it seems to look great with same font. I might have to adjust certain button font-size, but that should be simple enough!

<image src="https://user-images.githubusercontent.com/38238542/182981797-ea9c3612-4302-40fc-91b8-4a971cbf5f67.jpeg" width=400>

wackyleo459:
Yes I have followed your suggestion to include the reset inside the main.css. However, my format is a bit different in that I added the reset styling before everything else. I hope when you eventually merge your branch it will be somewhat straight forward!
@meodai

wackyleo459:
@frederikrothenberger I noticed selenium tests kept failing, and it seemed to be coming from an extra space for a test. Once I fixed that, the tests all passed!

frederikrothenberger:
> @frederikrothenberger I noticed selenium tests kept failing, and it seemed to be coming from an extra space for a test. Once I fixed that, the tests all passed!

It seems fine to remove it. Do you know, why this space was there previously (and now isn't)?

frederikrothenberger:
This has broken the layout on the remote device flow instructions, e.g. here: [screenshots/09-new-device-instructions-mobile.png](https://github.com/dfinity/internet-identity/pull/796/files#diff-aa730327da063680d3a9c8a61abdd87d9e5cb9d068e88b3326e0b16b69041f3a)

wackyleo459:
> > @frederikrothenberger I noticed selenium tests kept failing, and it seemed to be coming from an extra space for a test. Once I fixed that, the tests all passed!
> 
> It seems fine to remove it. Do you know, why this space was there previously (and now isn't)?

The space seemed to always be there, since when you first created the test. I'm not sure why tests were passing before and now were failing. 
I've updated the styling again, and styling should be consistent regarding numbers and bullets!

nmattia (commented):
Thanks a lot! 

I like having the reset in the `main.css` as well, for the reasons mentioned by @meodai .

I noticed a few things that broke:

* The "about" compatibility box layout is now vertical whereas it was horizontal before
* Links in both "about" and "faq" aren't styled anymore
* Headers on the compatibility page have lost their weight

Would be great if we could at least fix the headers and links!

wackyleo459:
> Thanks a lot!
> 
> I like having the reset in the `main.css` as well, for the reasons mentioned by @meodai .
> 
> I noticed a few things that broke:
> 
> * The "about" compatibility box layout is now vertical whereas it was horizontal before
> * Links in both "about" and "faq" aren't styled anymore
> * Headers on the compatibility page have lost their weight
> 
> Would be great if we could at least fix the headers and links!

@nmattia These issues have been addressed and fixed in the latest changes. If all tests seem to pass, I hope this PR will go through! =)

nmattia (approved):
This is great, thanks for addressing those few issues!

Issue #784, opened 2022-08-02
Change in-text buttons to links


wackyleo459:
There are some elements that don't appear as buttons (such as in-line texts), but are created as buttons.
These elements would be better suited as links. These buttons are located inside `<div class="textLink">` and `<div class="nav-links">`

Landing Page elements to address:
- [x] Create an Internet Identity Anchor.
- [x] but using a new device?
- [x] and want to recover?
- [x] ABOUT
- [x] FAQ

wackyleo459:
This PR https://github.com/dfinity/internet-identity/pull/797 resolves changing buttons to links
@meodai @nmattia

PR #797, opened 2022-08-05
Change inline text buttons to anchor tags on landing page.

wackyleo459:
This addresses issue 784: https://github.com/dfinity/internet-identity/issues/784

Inside each <li class="textLink> following are changed to anchor elements:
- Create an Internet Identity Anchor
- but using a new device?
- and want to recover?


nmattia (approved):
Thanks! LGTM. Great that you could keep the style exactly the same. @meodai you brought this up, looks good to you?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -163,12 +163,12 @@ const pageContent = (
     <div class="textLink">
       <hr />
       <div class="buttonContainer">
-        <button id="recoverButton" class="linkStyle">Lost access?</button>
+        <a id="recoverButton" class="linkStyle">Lost access?</a>
```
mstrasinskis, line 166: According to the documentation, the `<a>` tag should have `href` attribute. In this case, there is an action bound to the element. The best solution probably could be to have a proper `href` if only navigation occurs after click or leave the `button` tag.

https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#the-a-element
> If the [a](https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#elementdef-a) element has no [href](https://www.w3.org/TR/2016/REC-html51-20161101/links.html#element-attrdef-a-href) attribute, then the element [represents](https://www.w3.org/TR/2016/REC-html51-20161101/dom.html#represent) a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element’s contents.

mstrasinskis (commented):
Not sure about this replacing. But probably I just don't know the initial intention.

PR #798, opened 2022-08-06
Scope org package.json name

peterpeterparker:
# Motivation

To follow the same pattern - for consistency reason only - as the libs deployed on npm, scope the `package.json` name to the org.

Follow-up of my answer on [Slack](https://dfinity.slack.com/archives/C01JY0BT49Y/p1659767343345499?thread_ts=1659731522.066729&cid=C01JY0BT49Y).


nmattia (approved):
Thanks! Although to clarify, this is just for consistency, and isn't necessary since we don't publish this anyway, correct?

peterpeterparker:
Yes exactly, just for consistency

nmattia:
@peterpeterparker should we merge this?

PR #802, opened 2022-08-09
Fix /about in devServer

nmattia:
There was a typo on the `/about` route in local development which would
lead to a 404 when trying to get the page


peterpeterparker (approved):
LgTm

PR #803, opened 2022-08-11
Don't read canister ID as global

nmattia:
Before this change, the `canisterId` was set in the `index.html` as a global value and the `iiConnection.ts` module would read it upon initialization. This was done so as to minimize code changes when we started injecting the canister ID from the canister.

The major drawback was that `iiConnection.ts` would fail if there was no global `canisterId`, and would fail _during_ initialization, meaning whenever the module was imported (and not necessarily used).

The `IIConnection` class was used to issue IC calls to the canister. Some `static` methods would read the (global value) canister ID and would be used to talk to the canister. Once the user was authenticated, an actual instance of the `IIConnection` class would be used (which then contained the user's Identity object used to authenticate).

The major drawback here was that the entire codebase would use those `static` methods directly, making it impossible to mock the canister behavior.

This PR changes a few dependent things, which prepare the ground for mocking `IIConnection`/`Connection` (i.e. the interface with the canister):
* The canister ID is now added as an HTML tag dataset (as opposed to inlined as a global js value) and read on startup. A `Connection` is now created using this canister ID and is used instead of calling the `static` methods from `IIConnection`. This new object is threaded to all pages that need to talk to the IC.
* The `IIConnection` (the instantiated one) was renamed to `AuthenticatedConnection` to better reflect the difference with `Connection`. This object is _still_ threaded to all pages that need to issue _authenticated_ requests to the IC.
* I took the liberty to homogenize the order of `connection` and `userNumber` arguments to always have them in the same order: first `userNumber`, then `connection`. That's the convention that most of the codebase uses but recent PRs have introduced inconsistencies.

### DIFF
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -38,7 +38,6 @@ import { Principal } from "@dfinity/principal";
 import { MultiWebAuthnIdentity } from "./multiWebAuthnIdentity";
 import { hasOwnProperty } from "./utils";
 import * as tweetnacl from "tweetnacl";
-import { displayError } from "../components/displayError";
 import { fromMnemonicWithoutValidation } from "../crypto/ed25519";
 import { features } from "../features";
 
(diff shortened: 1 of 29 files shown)


src/frontend/src/utils/iiConnection.ts
```diff
@@ -358,11 +341,39 @@ export class IIConnection {
...
     const actor = Actor.createActor<_SERVICE>(internet_identity_idl, {
       agent,
-      canisterId: canisterId,
+      canisterId: this.canisterId,
     });
     return actor;
-  }
+  };
 
+  requestFEDelegation = async (
```
nmattia, line 349: this isn't new code; just moved to the class before it needs to read the canisterId

peterpeterparker (approved):
LGTM

nmattia:
FTR: @peterpeterparker discussed various alternative approaches offline (global stores, etc) and decided to go with this nonetheless

PR #805, opened 2022-08-12
Remove aboutOld.ts

nmattia:
I'm not sure when it was last used, but it's definitely not used
anymore.


lmuntaner (approved):
Rubber stamp

PR #806, opened 2022-08-12
Clean up webpack configuration

nmattia:
This updates our webpack config a bit. The main improvement is that requests that do not need to be routed to the canister can be made with `npm run start` _without_ having to spin up `dfx` first (which would originally crash because there was no local dfx state to read the canister ID from). This is in preparation for the static pages showcase which should greatly improve CSS hacking workflows.

See individual commits for more information.


peterpeterparker (approved):
Rubber stamp

PR #809, opened 2022-08-15
Remove more cruft from webpack config

nmattia:
This was automatically added by dfx and isn't needed.


peterpeterparker (approved):
Thx

nmattia:
Looks like the app needs those!

PR #810, opened 2022-08-15
Invalidate boostrap cache

nmattia:
We've bumped the ic-cdk-optimizer version and the GitHub actions cache
still contain the old version. Instead of carrying the old version
around forever, we invalidate the cache.


peterpeterparker (approved):
Rubber stamp

Issue #794, opened 2022-08-04
Extract "Warning" component


nmattia:
We have some code repetition, esp. with "warning"s that are pretty much all copy pasted. Instead we should have a separate component that can be reused.

meodai:
```HTML
  <aside class="warnBox">
    <div class="warnIcon">${warningIcon}</div>
    <div class="warnContent">
      <h2 class="warnTitle">Security Warning</h2>
      <div class="warnMessage">
        You are in the process of adding a new device. Any device added here
        will have <strong>full control over your identity</strong>. Only
        continue the process if you want to add a new device that you
        <em>personally own</em>.
      </div>
    </div>
  </aside>
```

could be something like 

```typescript
import { html, TemplateResult } from "lit-html";
import {warningIcon} from "./somewhere";

export const wranBox = (
  props = {
    title: "Warning Warning!",
    message: "Somewhere unsafe string that can also contain html <b>Evil HTML</b>",
    icon: warningIcon
  }
): TemplateResult => html`<aside class="warnBox">
    <div class="warnIcon">${icon}</div>
    <div class="warnContent">
      <h2 class="warnTitle">${title}</h2>
      <div class="warnMessage">
        ${message}
      </div>
    </div>
  </aside>`;
```

wackyleo459:
Resolved here: https://github.com/dfinity/internet-identity/pull/812

PR #812, opened 2022-08-15
Sueann/warning component

wackyleo459:
Resolves https://github.com/dfinity/internet-identity/issues/794

Warning Component is extracted with custom:
- title
- warning message
- choice between `div` or `aside` as type of container
- option to have custom button


src/frontend/src/components/warnBox.ts
```diff
@@ -0,0 +1,36 @@
...
+  message: string | TemplateResult;
+  componentType: string;
+  slot?: TemplateResult;
+}
+
+export const warnBox = (
+  props: warnBoxProps = {
+    title: "Warning",
+    message: html`Somewhere unsafe string that can also contain html
+      <b>Evil HTML</b>`,
```
nmattia, line 15: What exactly is this?

src/frontend/src/components/warnBox.ts
```diff
@@ -0,0 +1,36 @@
+import { html, TemplateResult } from "lit-html";
+import { warningIcon } from "./icons";
+
+interface warnBoxProps {
```
nmattia, line 4: nit: I don't think we use "props" terminology anywhere else

src/frontend/src/components/warnBox.ts
```diff
@@ -0,0 +1,36 @@
...
+  title: string;
+  message: string | TemplateResult;
+  componentType: string;
+  slot?: TemplateResult;
+}
+
+export const warnBox = (
+  props: warnBoxProps = {
```
nmattia, line 12: do we ever need these defaults?

src/frontend/src/components/warnBox.ts
```diff
@@ -0,0 +1,36 @@
...
+    title: "Warning",
+    message: html`Somewhere unsafe string that can also contain html
+      <b>Evil HTML</b>`,
+    componentType: "div",
+    slot: html`<button></button>`,
+  }
+): TemplateResult =>
+  html`${props.componentType === "div"
+    ? html` <div class="warnBox">
```
nmattia, line 21: the whole inner template is duplicated, which is not ideal. How about extracting that?

src/frontend/src/components/warnBox.ts
```diff
@@ -0,0 +1,36 @@
+import { html, TemplateResult } from "lit-html";
+import { warningIcon } from "./icons";
+
+interface warnBoxProps {
+  title: string;
+  message: string | TemplateResult;
+  componentType: string;
```
nmattia, line 7: shouldn't this be `"div" | "aside"`? Side note: do we really want to support both `aside` and `div`? CC @meodai

src/frontend/src/flows/manage/index.ts
```diff
@@ -217,18 +218,15 @@ const deviceListItem = (device: DeviceData) => html`
...
-      <div class="warnTitle">Recovery Mechanism</div>
-      <div class="warnMessage">
-        Add a recovery mechanism to help protect this Identity Anchor.
-      </div>
-      <button id="addRecovery" class="primary warnButton">Add Recovery</button>
-    </div>
-  </div>
-`;
+const recoveryNag = () =>
+  html`${warnBox({
```
nmattia, line 222: does this need to be wrapped in `html` or could you just have `warnBox`?

src/frontend/src/flows/manage/index.ts
```diff
@@ -217,18 +218,15 @@ const deviceListItem = (device: DeviceData) => html`
...
+    title: "Recovery Mechanism",
+    message: html`Add a recovery mechanism to help protect this Identity Anchor.`,
```
nmattia, line 224:
```suggestion
    message: "Add a recovery mechanism to help protect this Identity Anchor.",
```

src/frontend/src/components/warnBox.ts
```diff
@@ -0,0 +1,36 @@
...
+  slot?: TemplateResult;
+}
+
+export const warnBox = (
```
nmattia, line 11: Looking at the uses of this, and considering the question about about `div` vs. `aside` I wonder if it wouldn't be nicer/cleaner to have e.g. `warnBoxDiv` and `warnBoxAside` for instance, or even always have the `warnBox` be a `div` and let the caller wrap it in an `<aside></aside>` if necessary

src/frontend/src/components/warnBox.ts
```diff
@@ -0,0 +1,36 @@
+import { html, TemplateResult } from "lit-html";
```
nmattia, line 1: note about the file: this is a pretty straightforward component but for good measure would still be nice to have a line or two (short) lines about what this is and how to use it.

nmattia (commented):
Thanks!

Note: this is still a draft but [this](https://github.com/dfinity/internet-identity/issues/794#issuecomment-1215565103) looked like you were awaiting review so I just went ahead; apologies if I jumped the gun!

wackyleo459:
Thank you for the approval. I was waiting on David's pr to be completed. But if it's alright with everyone could we merge this first? This should not cause much issue, as it's simple implementation.

src/frontend/src/components/warnBox.ts
wackyleo459: It is a default dummy template, which we don't really need. Will do away with it!

src/frontend/src/components/warnBox.ts
wackyleo459: I like the idea of wrapping it into an aside. If it's alright with @meodai I can go ahead and make that change.

src/frontend/src/flows/manage/index.ts
wackyleo459: Yes it works without wrapping, thank you!

src/frontend/src/components/warnBox.ts
wackyleo459: Actually, having the warnBox **always** be an aside component makes sense to me. @meodai I think that may have been your intent as well? In your previous HTML semantics pull, you've changed a warnBox into an aside on this page [here](https://github.com/dfinity/internet-identity/blob/main/src/frontend/src/flows/addDevice/manage/index.ts#L59). But since the warnBox component was replicated many other places (as a div by default), I tried to include that option between div and an aside.

@nmattia 
I will make the change for the warnBox to be only an aside, because being encased in a div does not help with accessibility as you can see below:

Just as a div component: information looks fragmented in accessibility tree.

Screen Shot 2022-08-16 at 7 22 09 PM

As an aside component: the accessibility tree is much better.

Screen Shot 2022-08-16 at 7 20 53 PM

meodai: I think its a good idea to pass in the `componentType` in some occasions, it will be annoying to wrap if in something else. I would just call it htmlElement or something like that and then interpolate it in the template instead of using an if statement.

`<${htmlElement}></${htmlElement}>`

`htmlElement: 'aside'`, is probably a good default though. ()

wackyleo459: > I think its a good idea to pass in the `componentType` in some occasions, it will be annoying to wrap if in something else. I would just call it htmlElement or something like that and then interpolate it in the template instead of using an if statement.
> 
> `<${htmlElement}></${htmlElement}>`
> 
> `htmlElement: 'aside'`, is probably a good default though. ()

I've given the option to wrap between aside (default) and a div. Unfortunately lit-html does not allow string interpolation inside a tag for its tag type. I think that's the best we can do, unless we explicitly specify another element type. 

Screen Shot 2022-08-17 at 4 37 11 PM

nmattia (approved):
Thanks @wackyleo459 , this looks great! I too think it makes sense to merge this, so please go ahead!

PR #813, opened 2022-08-17
Device Alias form has 3 validation factors:

wackyleo459:
- Max length of 30 characters
- Name can not start with a space
- Name should not contain special characters (only alphanumeric characters, underscore character, and hyphen or space in the middle are allowed)


src/frontend/src/flows/addDevice/manage/addDevicePickAlias.ts
```diff
@@ -51,4 +53,26 @@ const init = (): Promise<string | null> =>
...
     };
+    const deviceInput = document.getElementById(
+      "deviceAlias"
+    ) as HTMLInputElement;
+
+    deviceInput.addEventListener("invalid", () => {
+      const firstChar = deviceInput.value[0];
+      const lastChar = deviceInput.value.slice(-1);
+
+      if (firstChar === " ") {
```
nmattia, line 64: the `pattern` above (`^[A-Za-z0-9_]+((-|\\s)*[A-Za-z0-9_])*$`) already seems to preclude starting with a space; which validity check takes precedence here? Same question below with the checks below.

src/frontend/src/flows/addDevice/manage/addDevicePickAlias.ts
```diff
@@ -13,6 +13,8 @@ const pageContent = () => html`
         placeholder="Device alias"
         type="text"
         required
+        maxlength="30"
+        pattern="^[A-Za-z0-9_]+((-|\\s)*[A-Za-z0-9_])*$"
```
nmattia, line 17: do you think it'd be worth having a comment here explaining what this matches? maybe an example or two? I always find it hard to tell with confidence what a regex actually matches

src/frontend/src/flows/addDevice/manage/addDevicePickAlias.ts
```diff
@@ -51,4 +53,26 @@ const init = (): Promise<string | null> =>
       event.preventDefault();
       resolve(deviceAlias.value);
     };
+    const deviceInput = document.getElementById(
+      "deviceAlias"
+    ) as HTMLInputElement;
+
+    deviceInput.addEventListener("invalid", () => {
+      const firstChar = deviceInput.value[0];
+      const lastChar = deviceInput.value.slice(-1);
```
nmattia, line 62: Do you think it'd be clearer to check with `startsWith` and `endsWith`? or more confusing?

nmattia (commented):
Thanks! I'm not entirely sure how custom validity works. Do the `maxlength` and `pattern` cause the input to become red as well? If the pattern isn't matched (e.g. if I write "###") what will the user see?

src/frontend/src/flows/addDevice/manage/addDevicePickAlias.ts
wackyleo459: @nmattia I've made the changes, with validation order of precedence

1. If no entries (validity.valueMissing)

Following all fall under validity.patternMismatch

2. If name starts with space
3. If it ends with space or hyphen 
4.  If there are special characters


https://user-images.githubusercontent.com/38238542/185503386-c73e5839-fd9f-40de-aad4-e0519716d189.mov

src/frontend/src/flows/addDevice/manage/addDevicePickAlias.ts
```diff
@@ -1,5 +1,9 @@
 import { html, render } from "lit-html";
 import { initLogout, logoutSection } from "../../../components/logout";
+// Regex Pattern for input: All characters, including starting(^) single character or more(+),
```
nmattia, line 3: nit: `^` is not the starting character, it's the anchor at the start of the string, but I'm really being picky here...

src/frontend/src/flows/addDevice/manage/addDevicePickAlias.ts
```diff
@@ -1,5 +1,9 @@
...
+// must be alphabet, number or underscore. Can have hyphen(s) or space(s) in the middle.
```
nmattia, line 4: any particular reason we allow starting with an underscore but not a hyphen? how about treating underscores like we do whitespaces and hyphens?

src/frontend/src/flows/addDevice/manage/addDevicePickAlias.ts
```diff
@@ -51,4 +58,30 @@ const init = (): Promise<string | null> =>
...
+        if (deviceInput.value.startsWith(" ")) {
+          message = "Name can't start with a space.";
+        } else if (
+          deviceInput.value.endsWith(" ") ||
+          deviceInput.value.endsWith("-")
+        ) {
+          message = "Name can't end with a space or hyphen.";
+        } else {
+          message = "Name can't contain special characters.";
+        }
```
nmattia, line 79: how about extracting this into a function? I.e.:

```ts
const foo = (validity: ..., input: string): string => {
    if (...) { ... }
    else if (...) { return "Can't end with a space or hyphen"; }
}

...

    const message = foo(deviceInput.validity, deviceInput.value);
```

it'll clarify the code a bit (those nested ifs are a bit daunting) and you may also be able to use it in `register.rs`. What do you think is best/more clear?

nmattia (commented):
Thanks a lot! Tried it out locally, works great.

As expected, clicking the submit button doesn't take me to the next page when the validity check is failing. I'm guessing this is form behavior; what would happen if we used `.onclick` instead of `.onsubmit`? Is this something we need to worry about?

wackyleo459:
> Thanks a lot! Tried it out locally, works great.
> 
> As expected, clicking the submit button doesn't take me to the next page when the validity check is failing. I'm guessing this is form behavior; what would happen if we used `.onclick` instead of `.onsubmit`? Is this something we need to worry about?

We don't need to worry about this. Even if the .onclick gets triggered, the parent form will not allow it to be submitted if validation fails. I tried to simulate the behavior (with mouse click and the console) and will not work.

I've extracted the function, so it looks cleaner.

nmattia (approved):
Nice, thank you!

PR #815, opened 2022-08-18
Improve selenium infrastructure

frederikrothenberger:
The following improvements are contained in this PR:
- Ability to run on Apple devices with M1 chip
- Support hot reloading of the II front-end
- Support redeploying the II backend and test-app
- No reliance on hard-coded canister ids

These improvements were achieved by running dfx and the II dev server natively on the host instead of dockerized.


meodai:
Oh so cool! Im excited about this, it will make it much easier for me to work with II

frederikrothenberger:
> Oh so cool! Im excited about this, it will make it much easier for me to work with II. Thanks so much

Sorry for the broken setup that was there before! If you have any issues with the setup or further suggestions on how to improve it I'm happy to make further changes.

.github/workflows/canister-tests.yml
```diff
@@ -221,9 +221,7 @@ jobs:
       # Make sure that one failing test does not cancel all other matrix jobs
       fail-fast: false
     env:
-      DFX_VERSION: 0.8.3
-      REPLICA_URL: https://ic0.app
-      II_ORIGIN: https://identity.ic0.app
+      DFX_VERSION: 0.11.1
```
nmattia, line 224: do we actually need the newer version to run this? should we do this separately and update `using-dev-build` as well?

docker-test-env/docker-compose.yml
```diff
@@ -2,45 +2,33 @@ version: "3.9"
...
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
     networks:
       ic:
         aliases:
           - ic0.app
-          # internet identity
-          - rrkah-fqaaa-aaaaa-aaaaq-cai.ic0.app
+          # internet identity, II_CANISTER_ID is substituted by the start-selenium-env script
+          - II_CANISTER_ID.ic0.app
```
nmattia, line 18: is this file ever used as is or is this just a template?

scripts/start-selenium-env
```diff
@@ -0,0 +1,60 @@
...
+# Make sure we always run from the root
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPTS_DIR/.."
+
+#########
+# USAGE #
+#########
+
+function title() {
+    echo "Launch docker based Selenium test infrastructure (selenium and nginx containers)"
```
nmattia, line 13: We used to have `with-selenium-standalone` which ran a given command with the environment up, meaning you didn't have to worry about shutting anything down. Do you think it'd make sense to have the same here?

scripts/start-selenium-env
```diff
@@ -0,0 +1,60 @@
...
+            title
+            usage
+            help
+            exit 0
+            ;;
+        *)
+            echo "ERROR: unknown argument $1"
+            usage
+            echo
+            echo "Use 'start-selenium-env --help' for more information"
```
nmattia, line 48: nit: you use `start-selenium-env` here and `$0` in the usage function

scripts/start-selenium-env
```diff
@@ -0,0 +1,60 @@
...
+            exit 1
+            ;;
+    esac
+done
+
+II_CANISTER_ID=$( jq -r .internet_identity.local .dfx/local/canister_ids.json )
+TEST_APP_CANISTER_ID=$( jq -r .test_app.local demos/test-app/.dfx/local/canister_ids.json )
+
+cd "docker-test-env"
+COMPOSE_CONIFIG=$(sed -u "s/II_CANISTER_ID/$II_CANISTER_ID/g; s/TEST_APP_CANISTER_ID/$TEST_APP_CANISTER_ID/g" docker-compose.yml)
```
nmattia, line 58:
```suggestion
COMPOSE_CONFIG=$(sed -u "s/II_CANISTER_ID/$II_CANISTER_ID/g; s/TEST_APP_CANISTER_ID/$TEST_APP_CANISTER_ID/g" docker-compose.yml)
```

nmattia: What is `-u` for?

.github/workflows/canister-tests.yml
```diff
@@ -250,29 +260,39 @@ jobs:
...
 
-      - name: Prepare Canister WASM Files
+      - name: Deploy Internet Identity
         run: |
-          mv internet_identity_test.wasm docker-test-env/dfx-docker/canisters/internet_identity.wasm
-          cp src/internet_identity/internet_identity.did docker-test-env/dfx-docker/canisters/internet_identity.did
-          mv test_app.wasm docker-test-env/dfx-docker/canisters/test_app.wasm
-          cp demos/test-app/test_app.did docker-test-env/dfx-docker/canisters/test_app.did
+          dfx canister create --all
+          dfx canister install internet_identity --argument '(null)' --wasm internet_identity_test.wasm
```
nmattia, line 268:
```suggestion
          dfx canister install internet_identity --wasm internet_identity_test.wasm
```

this might or might not work; for `dfx deploy` at least and since 0.10 you can omit null args

nmattia (approved):
Nice! I'm not a big fan of the sed-templating, but this is a big improvement so let's gooo!

scripts/start-selenium-env
frederikrothenberger: Yes, but I will do that in a separate PR. 👍

docker-test-env/docker-compose.yml
frederikrothenberger: It can be used as is to shut down the docker compose environment.

.github/workflows/canister-tests.yml
frederikrothenberger: If I leave out the `null` deployment argument as suggested below then we need it, right?
I would suggest to leave the version here and I will create a PR shortly also updating `using-dev-build` to the same version.

nmattia: sounds good. At the same time if you feel like creating an action for setting up dfx... that'd be amazing 😬 In particular it would fix two issues:
* `dfx` could be cached, which would prevent somewhat regular transient issues when the sdk download URL returns 503 (and might even save a few seconds, although that's secondary)
* We'd have a single place where the version is defined

frederikrothenberger:
> I'm not a big fan of the sed-templating, but this is a big improvement so let's gooo!

I'm not a big fan either. But this is due to a restriction of docker-compose, which does not allow setting a build context outside of the file hierarchy the compose file is placed in. I will try to come up with a better way once this iteration of improvements is merged.

PR #816, opened 2022-08-18
Rename ic-certified-vars dependency

frederikrothenberger:
The crate `ic-certified-vars` crate was renamed to `ic-certification`.


nmattia (approved):
LGTM, thanks! And this includes the IC update, correct?

frederikrothenberger:
> LGTM, thanks! And this includes the IC update, correct?

Yes, so the next run of the update job should be fine again.

PR #817, opened 2022-08-18
Send notification on ic update and rust update failures

nmattia:
This extracts the slack notification mechanism from the release build checks into an action and uses this newly created action in all our scheduled builds (release checks, ic update, rust update). This way, we avoid missing failures on scheduled jobs (notifications are only sent to whomever last edited the cron job).


.github/actions/slack/action.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Slack'
+description: 'A simple action that send a slack message to a channel.'
+inputs:
+  message:
+    description: 'The message to send.'
```
frederikrothenberger, line 5: Maybe add some doc that a link to the run will be included in the message?

frederikrothenberger (approved):
Thanks!

.github/actions/slack/action.yml
nmattia: That's a very good point, thanks!

PR #819, opened 2022-08-19
Don't serve about on #about

nmattia:
A year ago, when revamping the about page, we decided to serve it on
`/about` instead of `#about`, yet kept `#about` just in case. The about
page not being very important, I think it's safe to remove the old
`#about` route.

Note: I've had to update the part of the workflow that skips uploading screenshots if changes are too small to accommodate for deleted screenshots.


frederikrothenberger (approved):
Thanks for also improving the CI job 👍

PR #820, opened 2022-08-19
Run CI selenium tests without hot-reloading

frederikrothenberger:
This removes an unnecessary component from CI speeding it up by a few seconds.


nmattia (approved):
Makes a ton of sense. Thanks!

PR #822, opened 2022-08-22
Add script that allows executing a command with the docker compose env

frederikrothenberger:
The script encapsulates starting and stopping the docker environment and executes the supplied command while the environment is up.

### DIFF
--- a/scripts/with-docker-env
+++ b/scripts/with-docker-env
@@ -39,5 +39,5 @@ ret=0
 echo "'$*' returned with $ret"
 
 cd docker-test-env
-docker compose down
+docker compose down || echo "docker compose failed to shut down!"
 exit "$ret"
--- a/scripts/with-docker-env
+++ b/scripts/with-docker-env
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Make sure we always run from the root
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPTS_DIR/.."
+
+#########
+# USAGE #
+#########
+
+function title() {
+    echo "Execute a command between starting and stopping the docker compose environment"
+}
+
+function usage() {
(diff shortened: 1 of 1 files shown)


scripts/with-docker-env
```diff
@@ -0,0 +1,43 @@
...
+
+scripts/start-selenium-env
+
+ret=0
+"$@" || ret="$?"
+
+echo "'$*' returned with $ret"
+
+cd docker-test-env
+docker compose down
```
nmattia, line 42: what happens if docker compose is not up? Could this already return with non-zero?

nmattia (approved):
nice! do we need both scripts? we could for instance have one script with two (very similar) behaviors:

``` bash
$ ./scripts/with-docker-env my-command # runs until my-command has returned, and then tear down env
$ ./scripts/with-docker-env # runs until users hits Ctrl+C, i.e. uses something like `sleep infinity` as the command
```

wdyt?

scripts/with-docker-env
frederikrothenberger: If `docker compose` fails, you'll get the error output from docker and a non-zero exit code. E.g.:
```
$ scripts/with-docker-env npm run test:e2e-desktop
[+] Building 0.0s (0/0)
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?: driver not connecting
```

frederikrothenberger: Ah, and yes, failure of `docker compose down` will cause non-zero exit code.

nmattia: meaning the script will exit and you won't know what `$ret` was, correct? is it worth also "catching" the return code of `docker compose down` and then show both return codes before exiting?

frederikrothenberger: `echo "'$*' returned with $ret"` shows the return code?

frederikrothenberger: Added an explicit warning when `docker compose` fails to shut down properly.

PR #823, opened 2022-08-22
Add authenticate page to showcase

nmattia:
This adds the `authorizeAuthentication` page/function to the showcase.
Unfortunately the page is very static as button logic does somewhat
depend on extra input like local storage; if we want button logic (like
clicking the edit icon to edit the anchor) we'll have the refactor
things a little first.


meodai:
LGTM

meodai (approved):

PR #826, opened 2022-08-22
Strip down logo XML to a minimum

meodai:
- The logo no is just the logo it self (it had whitespace around it before, what would make it odd to size and center)
- Uses [SVGO](https://github.com/svg/svgo) to make minify the SVG output a bit (mostly simplifies gradients and paths in this case)


nmattia (approved):
Thanks!

PR #828, opened 2022-08-23
Adds metrics for anchor operations and delegation creation

frederikrothenberger:
These additional metrics will give a clearer picture regarding the usage of II. This data is relevant to decide on a suitable design for the anchor operations logging feature.


nmattia (approved):
Looks great, thanks!

PR #831, opened 2022-08-23
Fix layout of recovery with passphrase page

meodai:
The page was full with but it was not supposed to. a few elements were update to HTML5 since this pages was missed in the semantics update.


nmattia (approved):
Nice, thanks a lot for jumping in!

PR #832, opened 2022-08-23
Improve canister_tests caching

nmattia:
This makes the canister_tests job faster by improving caching. On main,
tests can take up to 40minutes to run, and are now down to 5 or 10mn
depdending on the cache state.

In particular, this adds more file to be cached (including files
recently introduced by cargo which most likely broke the caching) and
splits some steps into more steps for better visibility re. where time
is spent.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -157,13 +157,18 @@ jobs:
   # (note: this runs _all_ cargo tests)
   canister-tests:
     runs-on: ${{ matrix.os }}
-    needs:
-      - docker-build
+    needs: docker-build
     strategy:
       matrix:
         os: [ ubuntu-latest, macos-latest ]
     steps:
       - uses: actions/checkout@v2
+
+      - uses: ./.github/actions/bootstrap
+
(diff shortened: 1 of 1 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -203,8 +213,7 @@ jobs:
           # PRs that used to be green may become red (if the new release broke something). While this is not CI best practice, it's
           # a relatively small price to pay to make sure PRs are always tested against the latest release.
           curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity_test.wasm -o internet_identity_previous.wasm
-          # the "rand" features are needed for the captcha crate to compile
```
frederikrothenberger, line 206: Why is this no longer needed?

frederikrothenberger (approved):
LGTM. Just one question.

.github/workflows/canister-tests.yml
nmattia: Meant to leave a note about this, my bad. We don't need to build the CAPTCHA crate at all when building the canister tests, I'm really unsure why this was there before. Maybe at some point the canister tests did depend on the canister build (though don't think so) or it was a case of copy pasta

PR #834, opened 2022-08-23
Create a grid out of the showcase

meodai:
Makes the showcase a grid of iFrames. 

Makes it easier to see if something breaks and someone who is not familiar with a the view titles yet can easily work with it. Screenshot 2022-08-23 at 14 22 38


nmattia (approved):
Looks great! And to clarify, we'll stick to inline styles here because we don't want to bloat the main.css with stuff that won't be used in production, right?

meodai:
@nmattia yes. Those are only needed for the showcase and should not end up in the clients CSS.

PR #835, opened 2022-08-23
Skip canister_tests build if necessary

nmattia:
This greatly improves cache reuse for canister tests build on GHA.

This adds a step that sets the mtime of all files relevant in the
canister tests builds to the date they were last edited, according to
the git history. This means that in most cases (e.g. when the canister
tests' code is not changed) the build is skipped entirely (though tests
are still run).

In order to achieve this the checkout step is changed to fetch the
entire git history, which takes 4s instead of 3s for a time gain of
about 5mn.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -163,6 +163,9 @@ jobs:
         os: [ ubuntu-latest, macos-latest ]
     steps:
       - uses: actions/checkout@v2
+        with:
+          # Pull a deep copy so that we can set meaningful timestamps across builds, based on the commit history
+          fetch-depth: '0'
 
       - uses: ./.github/actions/bootstrap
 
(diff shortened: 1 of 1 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -184,22 +187,25 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('src/**/Cargo.lock', 'rust-toolchain.toml') }}-2
 
-      - name: Create fake assets
```
nmattia, line 187: Just like the CAPTCHA related code in #832 , I suspect (though can't quite remember) that we used to build the II code in this workflow; this is not the case now anyway and this step is not necessary anymore.

frederikrothenberger (approved):
Fantastic improvement! 😄

PR #836, opened 2022-08-24
Add github action to setup dfx

frederikrothenberger:
The dfx setup action installs dfx and then caches it. If the same version is required, it will be restored from cache.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -249,18 +249,18 @@ export const renderManage = async (
   }
 };
 
-export const displayManage = async (
+export const displayManage = (
   userNumber: bigint,
   connection: AuthenticatedConnection,
   devices: DeviceData[]
-): Promise<void> => {
+): void => {
   const container = document.getElementById("pageContent") as HTMLElement;
   render(pageContent(userNumber, devices), container);
   init(userNumber, connection, devices);
 };
 
 // Initializes the management page.
-const init = async (
+const init = (
   userNumber: bigint,
   connection: AuthenticatedConnection,
   devices: DeviceData[]


.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    # is why we run this only on Linux for now
+    - name: Install DFX
+      shell: bash
+      env:
+        DFX_VERSION: 0.11.1
+      run: |
+        if [ -f ~/dfx/dfx ]
+        then
+          echo "DFX restored from cache"
+          ~/dfx/dfx --version
```
frederikrothenberger, line 22: Whether or not the cache is worth it, is very much dependent on the download speed of dfx. Sometimes there seems to be some github caching kicking in for the download, making restoring from cache slower than installation.
But I expect restoring from cache to be more consistent as the time for installation seems to vary quite a bit:
* Fast example (1s download): https://github.com/dfinity/internet-identity/runs/7992688821?check_suite_focus=true
* Slow(er) example (35s download): https://github.com/dfinity/internet-identity/runs/7992485707?check_suite_focus=true

Restoring from cache takes about 18 seconds.

.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,31 @@
...
+description: Setup DFX
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/cache@v2
+      with:
+        path: |
+          ~/dfx
+          ~/.cache/dfinity
+        key: dfx-cache-0.11.1-2
```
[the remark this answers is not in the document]
nmattia: Yep, you can use the [`env`](https://docs.github.com/en/actions/learn-github-actions/contexts#env-context) context, though you'll probably need to lift the var up to the job (and not the step)

.github/actions/setup-dfx/action.yml
nmattia: Another reason for wanting caching here is that the dfx download regularly errors out with 503s and the download script itself doesn't have a retry option

.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,32 @@
...
+    # This step hangs on Github actions on Darwin for some reason, that
+    # is why we run this only on Linux for now
+    - name: Install DFX
+      shell: bash
+      env:
+        DFX_VERSION: 0.11.1
+      run: |
+        if [ -f ~/dfx/dfx ]
```
nmattia, line 19: Is there a conventional install dir we could use?

.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,32 @@
...
+        then
+          echo "DFX restored from cache"
+          ~/dfx/dfx --version
```
nmattia, line 22: nit: move this out of the "true" branch so that it also shows the version when installing without the cache

nmattia, line 22: Does the full path need to be specified here? isn't dfx in the PATH already? If so, let's show the version when calling it from PATH (should be the same, but will prevent confusion if that's not the case)

.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,32 @@
...
+        else
+          # sent env variable here so that '~' gets expanded
+          export DFX_INSTALL_ROOT=~/dfx
+          echo DFX not restored from cache, running install script:
+          echo DFX Version: "$DFX_VERSION"
+          echo DFX location: "$DFX_INSTALL_ROOT"
+          (yes || true) | sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
```
nmattia, line 29: guessing pipefail kicks in and fails everything? any idea why this wasn't the case before?

.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,32 @@
...
+      shell: bash
+      env:
+        DFX_VERSION: 0.11.1
+      run: |
+        if [ -f ~/dfx/dfx ]
+        then
+          echo "DFX restored from cache"
+          ~/dfx/dfx --version
+        else
+          # sent env variable here so that '~' gets expanded
```
nmattia, line 24:
```suggestion
          # set env variable here so that '~' gets expanded
```

here as opposed to where? `~` should get expanded everywhere (as long as you don't wrap in it quotes or double quotes)

.github/actions/setup-dfx/action.yml
frederikrothenberger: Yes, exactly: https://github.com/dfinity/internet-identity/runs/8015828437?check_suite_focus=true

Apparently, the behavior changes when moved to a composite action.

.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,27 @@
...
+  using: "composite"
+  steps:
+    - uses: actions/cache@v2
+      with:
+        path: |
+          /usr/local/bin/dfx
+          ~/.cache/dfinity
+        key: dfx-cache-0.11.1-3
+    # This step hangs on Github actions on Darwin for some reason, that
+    # is why we run this only on Linux for now
```
nmattia, line 13: I'm being a little pedantic here but actions don't specify a platform to run on, so I think this comment is just confusing here. We could throw and error on Darwin; that being said, did you confirm that this still breaks on Darwin? I'd be happy to just remove the comment for now; we're not installing dfx on darwin anyway

.github/actions/setup-dfx/action.yml
```diff
@@ -0,0 +1,27 @@
...
+description: Setup DFX
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/cache@v2
+      with:
+        path: |
+          /usr/local/bin/dfx
+          ~/.cache/dfinity
+        key: dfx-cache-0.11.1-3
```
nmattia, line 11: didn't you want to us `${{ env.DFX_VERSION }}` in the cache key?

nmattia (approved):
Looks great, thanks!

PR #837, opened 2022-08-24
Remove no longer needed compatibility shim

frederikrothenberger:
Originally introduced when device protection was added to the candid interface.


src/canister_tests/src/api.rs
```diff
@@ -114,83 +114,6 @@ pub fn lookup(
     framework::query_candid(env, canister_id, "lookup", (user_number,)).map(|(x,)| x)
 }
 
-/// A "compatibility" module for older versions of II that don't necessarily have a
-/// `protection`. Can be removed once the latest release includes `protection`.
-pub mod compat {
```
nmattia, line 119: We'll definitely need this again in the future, do you think we should keep the module around w/ the comment?

nmattia (approved):
Thanks!

PR #838, opened 2022-08-25
Add screenshots screens to showcase

nmattia:
This adds all screens recorded on screenshots to the showcase. Not only
does this mean that most if not all screens can be tweaked from the
showcase, but this prepare the showcase code to take over the screenshot
tests.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -232,8 +232,6 @@ export const renderManage = async (
   userNumber: bigint,
   connection: AuthenticatedConnection
 ): Promise<void> => {
-  const container = document.getElementById("pageContent") as HTMLElement;
-
   let anchorInfo: IdentityAnchorInfo;
   try {
     anchorInfo = await withLoader(() => connection.getAnchorInfo());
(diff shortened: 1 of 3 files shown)


frederikrothenberger (approved):
This showcase is really cool. 👍 

I think there is now a version of every screen, but not every variation of those screens.
I just noticed the following still missing:
* Management screen with recovery nag
* Start session screen with different derivationOrigin
* Start session screen with prefilled (grey) anchor

nmattia:
Yes good catch! I'm not sure what the best way to deal with this is, I'm not sure having _n_ variations of each pas as a standalone page is useful. I'll give it a think and address when I move the screenshots over!

src/frontend/src/flows/manage/index.ts
```diff
@@ -247,11 +245,20 @@ export const renderManage = async (
...
-    init(userNumber, connection, anchorInfo.devices);
+    displayManage(userNumber, connection, anchorInfo.devices);
   }
 };
 
+export const displayManage = async (
+  userNumber: bigint,
+  connection: AuthenticatedConnection,
+  devices: DeviceData[]
+): Promise<void> => {
```
peterpeterparker, line 256: either this function doesn't necessary have to be a promise or `render` or `init` need to be awaited.

randomly noticed it, just in case it can be useful.

nmattia: Good catch, let me check!

nmattia: https://github.com/dfinity/internet-identity/pull/839

PR #840, opened 2022-08-26
Don't hardcode versions in Dockerfile

nmattia:
This removes the last hard coded versions and adds a `.node-version` for
node. This files is read natively by a few tools; for the ones that do
not read it like nvm we update the nvm setup code.

The node setup in GHA is also extracted into an action that can be
reused where the version is set only once, like we do for dfx.


frederikrothenberger (approved):
Thanks, LGTM 👍 
One question though: shouldn't the `frontend-checks` job also use this action? Apparently, we use whatever is provide by `ubuntu:latest` there.

nmattia:
Good catch, updated!

PR #845, opened 2022-08-29
Add node autoupdate

nmattia:
This adds a weekly workflow that updates node to the latest v16,
similar to what we have for rust.


frederikrothenberger (approved):
LGTM, thanks

PR #854, opened 2022-08-30
Mention II in release text

nmattia:
This adds a clear mention of "Internet Identity" in the release text.
This text is also used in proposals, where it's unclear what's being
updated.


frederikrothenberger (approved):
👍

PR #855, opened 2022-08-31
Fix class attr duplication and missing button classes

mstrasinskis:
A couple of minor class fixes.

# Changes

- remove `class` attribute duplication (not valid HTML)
- add missing button classes


meodai (approved):
LGTM thanks so much.

nmattia (approved):
Thanks a lot!

This is still a draft I see, is this ready to go nonetheless?

One note: can we horizontally center these as well?

Screenshot 2022-09-01 at 10 10 58

mstrasinskis:
> Thanks a lot!
> 
> This is still a draft I see, is this ready to go nonetheless?
> 
> One note: can we horizontally center these as well?
> 
> Screenshot 2022-09-01 at 10 10 58

Good catch. Updated:
image

Issue #856, opened 2022-09-01
Fix "Warning" borders on iOS


nmattia:
![55FD0096-5BF8-46AF-B397-517DD4B3971E](https://user-images.githubusercontent.com/6930756/187870902-a562b6e8-b4cf-4bf9-9d64-0c3db27b89f4.jpeg)

As reported [here](https://forum.dfinity.org/t/proposal-78214-internet-identity-upgrade-its-about-to-get-pretty/15186/2?u=nmattia), the warn box needs some margin on iOS.

PR #858, opened 2022-09-01
Fix warning borders on iOS and small devices

nmattia:
Fixes https://github.com/dfinity/internet-identity/issues/856

The warning boxes were too big on small devices which caused the warning
icon to leak out of the page. All warning messages were moved to
`warnBox` and wrapped in a highligh-card.

Additionally, the "details" caret was moved right a bit to prevent it
overlapping with the left border.


mstrasinskis (approved):
LGTM, thanks.

PR #859, opened 2022-09-01
Bump threshold for updating screenshots

nmattia:
Before uploading new screenshots, we use imagemagick to check how many
pixels have changed. Until now the threshold for considering an image
changed was 8 pixels. It looks like this isn't enough however as new
pixels with irrelevant 26 pixel changes are being updated. This bumps
the threashold to 30 pixels.


peterpeterparker (approved):
Rubber stamp

PR #862, opened 2022-09-01
fix issue with the blurred background in safari

meodai:
Forcing the browser to render the blur background using the GPU


nmattia (approved):
Thanks!

PR #863, opened 2022-09-02
Wrap words

nmattia:
This ensures that longs inputs (like a long URL when authenticating) are
wrapped.

Before: 
Screenshot 2022-09-02 at 10 38 51

After:

Screenshot 2022-09-02 at 10 39 10

### DIFF
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -677,7 +677,7 @@ a:hover,
   background: var(--rc-input);
   color: var(--rc-onInput);
   border: calc(var(--vs-line) * 2) solid var(--rc-line);
-  word-break: break-all;
+  word-break: break-word;
 }
 
 .c-input.has-error,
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
Thx

lmuntaner (commented):
I assume you tried `break-word;` instead of `break-all`? Can't we rely that the work will be short enough to fit?

peterpeterparker:
`break-all`
Capture d’écran 2022-09-02 à 12 09 50

`break-word`
Capture d’écran 2022-09-02 à 12 08 27

nmattia:
@lmuntaner @peterpeterparker is one better than the other here?

src/frontend/src/styles/main.css
```diff
@@ -677,6 +677,7 @@ a:hover,
   background: var(--rc-input);
   color: var(--rc-onInput);
   border: calc(var(--vs-line) * 2) solid var(--rc-line);
+  word-break: break-all;
```
nmattia, line 680:
```suggestion
  word-break: break-word;
```

PR #865, opened 2022-09-02
Use browser.execute to click

nmattia:
In some cases, `elem.click()` doesn't seem to be enough, for unknown
reasons.


lmuntaner (approved):
Wow, that is a very weird problem...

nmattia:
For some odd reason tests are now green, without the need of this PR. The plot thickens.

PR #874, opened 2022-09-05
Add darkmode

meodai:
@mstrasinskis @nmattia thats the CSS dark-branch.

I will still work on the style, but we can go ahead and add the JS logic to save the user settings.


meodai:
I am also building a button now. But the more I work on it, the more I hate it. It feels like adding a "Print this page" / "Back to Top" or "Add this page to your bookmarks" buttons. I don't like the idea of recreating things that are usually controlled on a browser / os level.

nmattia:
@meodai what would you suggest instead? How would we deal with situations where someone is using dark mode on their system, but needs light mode in II e.g. for accessibility reasons, or simply personal preference? Not allowing a toggle will unconditionally force everyone using darkmode on their system to use II in dark mode.

I've looked around a bit, all platforms I could think of either provide only a light mode, or allow changing. Even GitHub does not use the system settings by default:

Screenshot 2022-09-06 at 13 24 24

PR #875, opened 2022-09-06
Add more pages to showcase

nmattia:
This updates the anchor management page to have more devices and adds the "recovery with device" page.


meodai (approved):
Not a huge fan of numbering things when naming, but since its only for the show-case I don't mind

nmattia:
@meodai interesting, what would you suggest?

nmattia:
Updated with an array

Issue #776, opened 2022-07-29
using II for local development in Safari fails


peterpeterparker:
# Motivation

As a developer I want to be able to develop my dapp locally with II installed locally as well. This works well with Chrome or Firefox but unfortunately, fails in Safari.

e.g. given II local path `http://127.0.0.1:8000/?canisterId=rwlgt-iiaaa-aaaaa-aaaaa-cai` I get an error "Fail to load resource through https"

# Reproduction

```
npm init ic
follow the prompt, create a project with II, start dfx and deploy
open the provided II link in Safari
```

# Screenshots

Safari:

<img width="1498" alt="Capture d’écran 2022-07-29 à 09 47 41" src="https://user-images.githubusercontent.com/16886711/181722317-794c20ea-759c-42ab-a9d8-c96fb3f2e397.png">

Chrome for comparison:

<img width="1510" alt="Capture d’écran 2022-07-29 à 09 47 52" src="https://user-images.githubusercontent.com/16886711/181722347-0d5cc09c-3c1d-41d3-96b1-a474ba9d4f1a.png">

nmattia:
Thanks for the report, I can reproduce this. Will get on it asap.

nmattia:
Actually gave this a thought; while `npm run start` will now work with Safari (because we're using webpack to serve the HTML) I fear the problem may persist when using the canister independently (since there CSP headers still comme straight from the canister). Thoughts?

nmattia:
Yeah, can still reproduce this. I think the best way forward is to add a new build feature to [disable CSP](https://github.com/dfinity/internet-identity#features) and have the dev version use that. @peterpeterparker or @frederikrothenberger wanna give this a try?

PR #877, opened 2022-09-06
#776: fix local development CSP issues in Safari

nathanosdev:
This fixes: https://github.com/dfinity/internet-identity/issues/776

Safari seems to be taking a different approach to the CSP headers on localhost compared to other browsers. Chrome for example, ignores the upgrade-insecure-requests directive on localhost, but Safari respects it and forces index.js to be loaded over https, but it's not being served over https by webpack dev server.

Handling this in webpack only, means we don't need to change any application code to work around the issue.

### DIFF
--- a/demos/using-dev-build/webpack.config.js
+++ b/demos/using-dev-build/webpack.config.js
@@ -1,8 +1,6 @@
 const path = require("path");
 const webpack = require("webpack");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
-const TerserPlugin = require("terser-webpack-plugin");
-const CopyPlugin = require("copy-webpack-plugin");
 
 let canisters;
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,32 +1,9 @@
 const path = require("path");
 const webpack = require("webpack");
 const CopyPlugin = require("copy-webpack-plugin");
-const TerserPlugin = require("terser-webpack-plugin");
 const CompressionPlugin = require("compression-webpack-plugin");
-const HttpProxyMiddlware = require("http-proxy-middleware");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
-const dfxJson = require("./dfx.json");
 require("dotenv").config();
-
-/** Read the replica host from dfx.json */
-function readReplicaHost() {
-  const dfxJson = "./dfx.json";
-  let replicaHost;
-
-  try {
-    replicaHost = require(dfxJson).networks.local.bind;
-  } catch (e) {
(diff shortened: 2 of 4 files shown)


demos/using-dev-build/webpack.config.js
```diff
@@ -1,8 +1,6 @@
 const path = require("path");
 const webpack = require("webpack");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
-const TerserPlugin = require("terser-webpack-plugin");
```
nathanosdev, line 4: These were not used, unrelated to the PR.

webpack.config.js
```diff
@@ -1,33 +1,10 @@
...
 const CopyPlugin = require("copy-webpack-plugin");
-const TerserPlugin = require("terser-webpack-plugin");
 const CompressionPlugin = require("compression-webpack-plugin");
-const HttpProxyMiddlware = require("http-proxy-middleware");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
-const dfxJson = require("./dfx.json");
 require("dotenv").config();
 
-/** Read the replica host from dfx.json */
-function readReplicaHost() {
```
nathanosdev, line 12: Not necessary anymore with these changes

webpack.config.js
```diff
@@ -45,6 +22,27 @@ function readCanisterId() {
   return canisterId;
 }
 
+// This plugin emulates the behaviour of http.rs while using webpack dev server locally
+// so we don't need to proxy to the backend canister for the index.html file.
+// This overcomes some issues with Safari and the CSP headers set in http.rs.
+class InjectCanisterIdPlugin {
```
nathanosdev, line 28: This could also be handled by using HtmlWebpackPlugin's templating features, but that means we would either have one messy index.html or maintain two separate ones. I think it's nicer to keep this complexity in the webpack config. But I'm happy to take the alternative approach if anyone disagrees.

webpack.config.js
nmattia: I _guess_ we could still use it in the [/api proxy](https://github.com/dfinity/internet-identity/pull/877/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fL116), but it wasn't used there until now, so I'm happy removing it

demos/using-dev-build/webpack.config.js
nmattia: Gotcha! Next time don't hesitate to create another PR

nmattia (approved):
Really great work, thanks a lot! Merge at will

PR #879, opened 2022-09-06
Slim down authenticate page

nmattia:
This slims down the "authenticate" page to make important/commonly used
bits (anchor number, start button) more prominent.

Before:

Screenshot 2022-09-06 at 16 33 41

After (s/originally/an alias of):

Screenshot 2022-09-06 at 16 33 16


src/frontend/src/flows/authenticate/index.ts
```diff
@@ -30,65 +30,57 @@ const pageContent = (
...
-    <div class="c-input t-vip t-vip--small c-input--readonly">${hostName}</div>
-    ${derivationOrigin !== undefined && derivationOrigin !== hostName
-      ? derivationOriginSection(derivationOrigin)
-      : ""}
-    <p class="t-paragraph">Use Identity Anchor:</p>
+    <p class="t-lead">Authenticating to <span class="t-strong">${hostName}</span> ${
+  derivationOrigin !== undefined && derivationOrigin !== hostName
+    ? html`(an alias of <span class="t-strong">${derivationOrigin}</span>)`
+    : ""
+} using <span class="t-strong">Internet Identity</h1>.</p>
```
meodai, line 40: you are opening a span here but closing an h1

meodai (changes requested):

src/frontend/src/flows/authenticate/index.ts
nmattia: Good catch, thanks. Will need to work on this a bit more anyway.

PR #880, opened 2022-09-07
Setup log canister

frederikrothenberger:
Adds the structure to build the log canister.

Screenshot changes are due to the log canister being deployed as well which influences the canister id of the test app.


src/internet_identity_log/internet_identity_log.did
```diff
@@ -0,0 +1,54 @@
...
+
+  // Returns the latest up to 1000 log entries. If an index is given, entries starting from the given index are returned.
+  //
+  // Parameters:
+  // 1. optional index into the list of log entries
+  // 2. optional limit of the number of entries. If no limit is given, at most 1000 entries are returned.
+  get_logs : (opt nat64, opt nat16) -> (Logs) query;
+
+  // Writes a log entry. Only the Internet Identity canister is authorized to call this function.
+  write_entry : (UserNumber, Timestamp, blob) -> ();
```
nomeata, line 50: Will the II canister use one-shot calls to write to the log, to be always upgradable etc?

In that case (but also otherwise) I wonder if the log writes should also contain a sequence number (globally or per user), so that failed writes at least leave a trace (even if the data is lost).

Would it be useful to add a continuous

frederikrothenberger: No, the logging calls will not be `oneway`, as this feature only exists in candid and does not actually have an effect (yet). We might reconsider once the feature is actually ready.

We plan on introducing a logging reliability metric on the II side.

nomeata: I'm not referring to the candid annotation, but rather the way of calling a canister without registering a callback, so that you stay upgradable, as discussed in https://www.joachim-breitner.de/blog/789-Zero-downtime_upgrades_of_Internet_Computer_canisters.

I think @roman-kashitsyn added support for that to the Rust CDK.

I don't see much value in waiting for the response from a log canister - what do you do if it fails anyways?

frederikrothenberger: Hi @nomeata
I have now looked into that feature and I think we will use it. Thanks for pointing it out. 👍

frederikrothenberger:
Superseded by #905.

PR #881, opened 2022-09-07
feat: link footer with ic homepage

peterpeterparker:
Link "100% on chain" footer with the IC homepage and add `rel` attributes.


nmattia (approved):
Thank you!

PR #884, opened 2022-09-10
style: link hover

peterpeterparker:
# Motivation

I had an hour train to kill so I had fun with the `:hover` effects since we discussed this briefly with @nmattia this week.

# Note

Absolutely no string attached, you can trash and reject this PR. I won't be sad if so 😉.

Found the "text background effect" in this article https://css-tricks.com/css-link-hover-effects/

# Changes

- replace `text-underline: none` hover effect for link with text color animation and bottom border
- update svg to inherit current color
- update footer link to standard class system
- hover primary box-shadow update to inherit primary color ("black-black-pink" instead of "black-white-pink")
- replace warning color with warning color
- use primary color for interaction

# Screenshot

link-hover


src/frontend/src/styles/main.css
```diff
@@ -387,10 +385,21 @@ body {
...
+    var(--vc-brand-pink),
+    var(--vc-brand-pink) 50%,
+    #000 50%
+  );
+  background-size: 200% 100%;
+  background-position: -100%;
+  position: relative;
+  -webkit-background-clip: text;
+  -webkit-text-fill-color: transparent;
+  transition: all 0.2s ease-in-out;
```
meodai, line 401: I would specify what changes. All is quite CPU/GPU intensive and some times comes with side-effects.

src/frontend/src/styles/main.css
```diff
@@ -387,10 +385,21 @@ body {
...
+  background-image: linear-gradient(
+    to right,
+    var(--vc-brand-pink),
+    var(--vc-brand-pink) 50%,
+    #000 50%
+  );
+  background-size: 200% 100%;
+  background-position: -100%;
+  position: relative;
+  -webkit-background-clip: text;
```
meodai, line 399: Since background-clip only works in webkit. What happens in other browser?

src/frontend/src/styles/main.css
```diff
@@ -421,6 +439,7 @@ a:hover,
   height: calc(1em * var(--vs-line-height));
   aspect-ratio: 1;
   vertical-align: middle;
+  transition: color 0.2s ease-out;
```
meodai, line 442: I think transitions on color and opacity should be linear, especially when that short.

meodai (changes requested):
I like the changes, thanks! But I am not a fan of the hover effect. I think it's a bit much.
 But if people like it, I don't feel strongly about it.

peterpeterparker:
@meodai no strong feeling neither, like I said, just had fun while killin some times 😉

you would rather like a simple `color` change of the text with short `animation` - same as for the icon?

src/frontend/src/styles/main.css
peterpeterparker: background-clip text is supported cross browser ([source](https://caniuse.com/?search=background-clip)).

src/frontend/src/styles/main.css
peterpeterparker: sure

src/frontend/src/styles/main.css
peterpeterparker: sure

peterpeterparker:
Was fun but not a must so I close this.

PR #887, opened 2022-09-14
Share index.html between app and showcase

nmattia:
This shares a single index.html between the actual app and the showcase, which has the following benefits:
* HTML is deduplicated
* Webpack config is simplified
* By using `HtmlWebpackPlugin` in production too, the HTML gets minimized


frederikrothenberger (approved):
LGTM!

PR #888, opened 2022-09-14
Extract CSS to file for production

nmattia:
This stops using "style-loader" in production, and instead uses a separate CSS file served by the canister.

This is the recommended "production"-way by "style-loader" itself: https://github.com/webpack-contrib/style-loader#recommend

> For production builds it's recommended to extract the CSS from your bundle being able to use parallel loading of CSS/JS resources later on. This can be achieved by using the mini-css-extract-plugin, because it creates separate css files. For development mode (including webpack-dev-server) you can use style-loader, because it injects CSS into the DOM using multiple <style></style> and works faster.

This updates the HTML template to read the CSS from the canister in production; the devServer workflow stays the same. Additionally, the `dist/` directory is wiped before build, as this sometimes caused issues hard to debug (`output: clean: true`).

Note: the CSP rules had to be updated to allow loading the stylesheet from the canister.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
         "html-webpack-plugin": "^5.3.2",
         "jest": "^27.0.6",
         "jest-cli": "^27.0.6",
+        "mini-css-extract-plugin": "^2.6.1",
         "prettier": "2.3.2",
         "process": "0.11.10",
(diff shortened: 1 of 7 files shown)


frederikrothenberger (approved):
LGTM, thanks :-)

PR #890, opened 2022-09-15
Rework authenticate page and change color

nmattia:
This is the result of some experimentation on the "authenticate" page, as well as some general design changes:

* The buttons and links on the "authenticate" page were moved a bit
* A lot of text was removed from "authenticate", and the (optional) alternative origin URL is hidden in a tooltip
* Buttons and cards in general were rounded
* Color scheme was changed from pink/orange to purple
* Anchor number input in "authenticate" was updated to only allow digits
* Showcase was updated to show a alternative origin URL
* "Internet Identity" was removed from titles and added to logo instead

CC @meodai


peterpeterparker (approved):
Rubber stamp

meodai (approved):
we will need to adapt some design tokens, but lgtm

PR #892, opened 2022-09-16
Reorganize rust canister tests to be colocated with the canister code

frederikrothenberger:
Additionally removes the II dependency on `ic-types` as it was no longer needed (and clashing with the monorepo `ic-types` needed in the `dev-dependencies`).
Some imports were reorganized as well to remove clutter.


src/internet_identity/tests/api/mod.rs
```diff
@@ -1,9 +1,10 @@
 /** The functions here are derived (manually) from Internet Identity's Candid file */
-use crate::framework;
-use crate::framework::CallError;
-use candid;
 use candid::Principal;
-use ic_state_machine_tests::{CanisterId, PrincipalId, StateMachine};
+use canister_tests::framework::{
+    call_candid, call_candid_as, query_candid, query_candid_as, CallError,
```
nmattia, line 4: Why do all these suddenly need to be imported explicitly?

.github/workflows/update-ic.yml
```diff
@@ -16,22 +16,35 @@ jobs:
...
         id: update
         run: |
-          current_ic_commit=$(cat ./src/canister_tests/Cargo.toml | sed -n 's/.*{\s*git\s*=\s*"https:\/\/github\.com\/dfinity\/ic",\s*rev\s*=\s*"\([a-z0-9]*\)" }.*/\1/p' | head -n 1)
-          echo "current ic commit '$current_ic_commit'"
-          latest_ic_commit=$(curl --silent -H 'Accept: application/vnd.github.v3+json' https://api.github.com/repos/dfinity/ic/commits/master | jq -cMr .sha)
-          echo "latest ic commit '$latest_ic_commit'"
-
-          if [ "$current_ic_commit" != "$latest_ic_commit" ]
-          then
+          shopt -s globstar
```
nmattia, line 19: After squinting at those changes for a bit, I'm guessing this is now performing the update in every cargo toml, since we now have more than one that depends on IC deps. Is that right?

.github/workflows/update-ic.yml
```diff
@@ -16,22 +16,35 @@ jobs:
...
+            echo Checking "$filename"
+            [ -e "$filename" ] || continue
+            if ! grep -q https://github\.com/dfinity/ic "$filename"; then
+              echo "$filename" does not contain ic dependecy, skipping
+              continue
+            fi
+          
+            current_ic_commit=$(cat "$filename" | sed -n 's/.*{\s*git\s*=\s*"https:\/\/github\.com\/dfinity\/ic",\s*rev\s*=\s*"\([a-z0-9]*\)" }.*/\1/p' | head -n 1)
+            echo "current ic commit '$current_ic_commit'"
+            latest_ic_commit=$(curl --silent -H 'Accept: application/vnd.github.v3+json' https://api.github.com/repos/dfinity/ic/commits/master | jq -cMr .sha)
```
nmattia, line 32: what happens if a new commit is pushed to the ic repo as we're iterating?

nmattia (commented):
I was a bit confused why `src/canister_tests` didn't move; it _looks_ like `canister_tests` itself is just the framework then? Why is that left behind when all tests have moved to `internet_identity/`?

.github/workflows/update-ic.yml
frederikrothenberger: Ah, good point. I'll move this out of the loop.

frederikrothenberger:
> I was a bit confused why `src/canister_tests` didn't move; it _looks_ like `canister_tests` itself is just the framework then? Why is that left behind when all tests have moved to `internet_identity/`?

Yes, exactly. This change is in preparation for the archive canister that will use some of the same infrastructure (i.e. even installing II to test for interoperability). We can also rename the `canister_tests` crate to better convey that it is infrastructure. Maybe `test_utils`?

src/canister_tests/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "canister_tests"
 version = "0.1.0"
 edition = "2021"
 
-[dev-dependencies]
+[dependencies]
```
nmattia, line 6: why change this? do we expect anything to be built except for the tests? or is that necessary for crates that depend on this?

src/internet_identity/tests/api/mod.rs
```diff
@@ -1,9 +1,8 @@
 /** The functions here are derived (manually) from Internet Identity's Candid file */
-use crate::framework;
-use crate::framework::CallError;
-use candid;
 use candid::Principal;
-use ic_state_machine_tests::{CanisterId, PrincipalId, StateMachine};
+use canister_tests::framework::*;
```
nmattia, line 3: let's avoid global imports, they really hurt readability

src/internet_identity/tests/tests.rs
```diff
@@ -1,29 +1,42 @@
-use crate::framework::device_data_1;
-use crate::{api, flows, framework};
+use canister_tests::framework::*;
```
nmattia, line 1: Same, let's avoid global imports

nmattia (commented):
Nice; couple of nits. In general would be great to avoid renaming imports, it makes the diff much bigger and less focused.

src/internet_identity/tests/api/mod.rs
nmattia: meh, actually I don't care

nmattia (approved):

src/canister_tests/Cargo.toml
frederikrothenberger: Yes, this is required because `#[cfg(test)]` does not propagate to dependencies.

src/internet_identity/tests/api/mod.rs
frederikrothenberger: Well, I have now reverted it anyways. In my opinion, having wildcard imports for _local_ dependencies is fine.

Maybe we can revisit this topic some other time.

PR #895, opened 2022-09-19
CSS cleanup

meodai:
Since we quickly hacked together the new design. I wanted to take some more time to clean things up a bit.

I have reintroduced the concept of the line color, instead of using the value.

+ the concept of `l-section` is not that fitting anymore so I renamed it to `l-stack`.
When coming up with it, I assumed it would be used to visually separate a section, but in practice it is only used to add a space to a top of things.


nmattia (approved):
LGTM, thanks! Merge away

PR #898, opened 2022-09-20
Use lit instead of lit-html

nmattia:
This upgrades our version of `lit`, as well as import from `lit` instead of `lit-html` (now the correct way to import it).

Using `lit` also gives us access to some extra
[directives](https://lit.dev/docs/templates/directives/) which will come in handy when factoring out (non-Web) components.


peterpeterparker (approved):
LGTM, thanks

PR #899, opened 2022-09-20
Add jest.config.ts to prettier

nmattia:
Makes sure the config is consistent with the rest of the codebase.


peterpeterparker (approved):
LGTM

Note, another option: format `.` (everything) and add `.prettierignore` in the repo for the folders to ignore

nmattia:
> Note, another option: format . (everything) and add .prettierignore in the repo for the folders to ignore

Yeah, I'm close to the tipping point, but... just not yet

PR #900, opened 2022-09-20
Use @click in logoutButton

nmattia:
"lit" can specify event listeners, which is safer than querying the element and adding the listener manually.


peterpeterparker (approved):
LGTM

PR #904, opened 2022-09-20
Introduce TemplateRef and make authenticate more robust

nmattia:
This introduces a new helper type, `TemplateRef`, that represents a "lit" template that contains references.

This new module contains some helpers that bring element lookup closer to the template description. By using "lit" `Ref`s, DOM elements can be bound without needing `id`s. Then, when the `TemplateRef` is rendered using `renderTemplateRef`, references are resolved and returned.

This means that, assuming references are given the proper types, element creation is completely type-safe.

The `authenticate` module is partly rewritten to avoid usage of unsafe ids.

### DIFF
--- a/src/frontend/src/flows/authenticate/index.ts
+++ b/src/frontend/src/flows/authenticate/index.ts
@@ -123,7 +123,7 @@ const pageContent = (
 
   return {
     template,
-    keys: { authorizeButton, userNumberInput, registerButton },
+    refs: { authorizeButton, userNumberInput, registerButton },
   };
 };
 
--- a/src/frontend/src/flows/authenticate/index.ts
+++ b/src/frontend/src/flows/authenticate/index.ts
@@ -1,4 +1,4 @@
-import { html, render } from "lit";
+import { html } from "lit";
 import { icLogo, attentionIcon } from "../../components/icons";
 import { footer } from "../../components/footer";
 import {
@@ -8,6 +8,8 @@ import {
 } from "../../utils/userNumber";
 import { withLoader } from "../../components/loader";
 import { AuthenticatedConnection, Connection } from "../../utils/iiConnection";
+import { TemplateRef, renderTemplateRef } from "../../utils/templateRef";
+import { ref, createRef } from "lit/directives/ref.js";
 import {
   apiResultToLoginFlowResult,
   LoginFlowSuccess,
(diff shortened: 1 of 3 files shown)


src/frontend/src/utils/templateRef.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  // XXX: there may be a way to describe what's happening to tsc but I
+  // haven't found it.
+  // eslint-disable-next-line
+  const ret = {} as any;
+
+  for (const i in tpl.keys) {
+    const val = tpl.keys[i].value;
+    if (val === null) {
+      throw Error("Bad reference found in template");
```
nmattia, line 33: @peterpeterparker what do you reckon is the best way to handle errors?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -24,72 +26,106 @@ import {
...
+        </ul>
+      </div>
     </div>
-  </div>
-  ${footer}`;
+    ${footer}`;
+
+  return {
+    template,
+    keys: { authorizeButton, userNumberInput, registerButton },
```
peterpeterparker, line 126: is the reference static or dynamic?

e.g. if a consumer renders the template and gets a ref, if that particular ref gets dynamically removed within the component, is the consumer informed the ref is not valid anymore?

I mean something like the React `forwardRef` pattern.

src/frontend/src/utils/templateRef.ts
```diff
@@ -0,0 +1,40 @@
...
+export function renderTemplateRef<T>(
+  tpl: TemplateRef<T>,
+  container: HTMLElement
+): T {
+  render(tpl.template, container);
+
+  // XXX: there may be a way to describe what's happening to tsc but I
+  // haven't found it.
+  // eslint-disable-next-line
+  const ret = {} as any;
```
peterpeterparker, line 28: TypeScript utility `Record<something>`?

src/frontend/src/utils/templateRef.ts
peterpeterparker: we have a beautiful assertion checker in `utils-js` if it's what you are looking for?

https://github.com/dfinity/ic-js/blob/main/packages/utils/src/utils/asserts.utils.ts

nmattia: No, I am looking for the best way to report the error. Here, if we do throw an error, then we lose track of all references. An alternative would be to not fail completely here and instead return e.g. useless elements.

src/frontend/src/flows/authenticate/index.ts
nmattia: The actual ref is dynamic, but after it's been extracted (in `renderTemplateRef`, with `.value`) then it's static. It's not ideal, but the behavior is similar to what we'd have before with `document.getElementById("...")` where the returned elements may not be attached to the DOM forever.

src/frontend/src/utils/templateRef.ts
nmattia: Anything more specific? 😅

src/frontend/src/utils/templateRef.ts
peterpeterparker: Don't really have a best approach to speak frankly, always depends a bit on how critical the error is I would say. Generally, maybe, when it's a DOM / components related error I try to handle it as close as I can. Like displayed an invalid entry in an input next to the input but, if it's a functional issue then it's often throw error and catched to be displayed in some sort of toasts or something I guess. Does that answer it?

Being said, try the assert of our utils, it's nice type safety.

src/frontend/src/flows/authenticate/index.ts
peterpeterparker: I see. Do you actually need to expose these ref or can you bind function / method to the template? If yes would it be cleaner to do so?

```
@Stencil
@tag 'my-yolo'
export class YoloComponent {

  ref!: HTMLDivElement;
  
  @Method()
  doSomething() {
      this.ref.letsDoIt();
  }

   render <div ref=el />
}

<html>
   <script>
      document.querySelector('my-yolo').doSomething();
   </script>
   
   <my-yolo />
</html>
```

Honestly just speaking out at loud since you said it's a bit an experiment. Totally fine with ref.

src/frontend/src/utils/templateRef.ts
peterpeterparker: `Record<string, unknow>`? `unknow` or the "developer does not know exactly the types but it's probably less ugly than any" 😉

src/frontend/src/flows/authenticate/index.ts
nmattia: Mh, the template itself isn't a component, just (basically) a string to be rendered. Are you suggesting bypassing lit's `render`? I'm not sure I understand

peterpeterparker: Ah alright, somehow the custom element discussion sticked in my head. So all clear.

peterpeterparker, line 126: maybe rename `keys` to `refs`?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -24,72 +26,106 @@ import {
...
+  const onRecoverClick = () => useRecovery(connection, readUserNumber());
 
-      <p
-        id="invalidAnchorMessage"
-        class="anchor-error-message is-hidden t-paragraph t-strong"
-      >
-        The Identity Anchor is not valid. Please try again.
-      </p>
-    </div>
+  const authorizeButton = createRef();
```
peterpeterparker, line 48: would it make sense to postfix the variables with the keyword `Ref`?

e.g. `authorizeButtonRef`. verbose but clear.

nmattia: I actually did this at first, but then on the `return` I was just renaming all keys (authorizeButtonRef -> authorizeButton) so I just dropped the suffix altogether

peterpeterparker: rename also where you use it 🤷‍♂️

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/authenticate/index.ts
nmattia: no, because where it's used it's not a ref anymore; that's the translation `TemplateRef` does at the type level; alternatively I could use that step to rename the keys at type level as well but this is getting a little crazy

PR #905, opened 2022-09-21
Add archive canister

frederikrothenberger:
Adds the archive canister to keep a record of all the anchor operations made on II.

Note that this canister does not yet interact with II. Integration with II will be added in a separate PR.


src/archive/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+ic-cdk = "0.5"
+ic-cdk-macros = "0.5"
+ic-stable-structures = "0.1"
+# other
+serde = "1"
+serde_bytes = "0.11"
+
+[dev-dependencies]
+canister_tests = { path = "../canister_tests" }
+ic-state-machine-tests = { git = "https://github.com/dfinity/ic", rev = "50c53c8e996e7b00b88c2f43120c46534cc499f6" }
```
nmattia, line 24: how about specifying `ic-types = "*"` (same for the other ic crates) and specify a [`[patch]`](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) section in the workspace manifest?

that'll reduce duplication a bit

scripts/build-archive-canister
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
frederikrothenberger: Agreed, I have now changed the `build` script and removed the extra one.

src/archive/archive.did
```diff
@@ -0,0 +1,143 @@
+type HeaderField = record { text; text; };
```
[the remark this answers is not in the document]
frederikrothenberger: There would be from the candid perspective. Unfortunately, it is not yet supported by `dfx`: https://dfinity.atlassian.net/browse/SDK-746

We can revisit this once `import` is properly supported.

Dockerfile
```diff
@@ -72,8 +75,22 @@ RUN touch src/canister_tests/src/lib.rs
 RUN touch src/metrics_encoder/src/lib.rs
 RUN npm ci
 
-RUN ./scripts/build
+RUN ./scripts/build --single-canister internet_identity
 RUN sha256sum /internet_identity.wasm
 
-FROM scratch AS scratch
-COPY --from=build /internet_identity.wasm /
+FROM deps as build_archive
```
nmattia, line 81: out of curiosity, do you know if the two builds run in parallel?

scripts/build
```diff
@@ -11,31 +11,33 @@ cd "$SCRIPTS_DIR/.."
...
-    echo "Build Internet Identity"
+    echo "Builds Internet Identity and Archive Canisters"
 }
 
 function usage() {
     cat << EOF
 
 Usage:
-  $0 [--only-dependencies]
+  $0 [--only-dependencies] [--single-canister CANISTER]
```
nmattia, line 21: nit: how about `--internet-identity` to enable the II build, and `--archive` to enable the archive build? shorter to type and more flexible; would default to `--internet-identity --archive`. Would also simplify the logic around `$CANISTER`

src/archive/archive.did
```diff
@@ -0,0 +1,143 @@
...
+type Logs = record {
+  entries: vec blob;
+  next_idx: opt nat64 // next index to fetch the next page of log entries (if any)
+};
+
+type ArchiveInit = record {
+  ii_canister: principal;
+};
+
+service : (opt ArchiveInit) -> {
```
nmattia, line 119: really nitpicky here, but how about bringing the `service` to the top, or at least hiding all those HTTP types at the button, or making different sections with stuff actually relevant, and stuff needed because candid is what is it?

```

// Our lovely types

type Logs = ...


// Stuff that shouldn't even be here

type AwfulHttpStreamingType = ...
```

src/archive/archive.did
```diff
@@ -0,0 +1,143 @@
...
+
+  // Returns the logs for the given user number. If a timestamp is given, only the logs starting from that timestamp are
+  // returned. Use the Cursor to skip to later log entries.
+  // This function can be called anonymously.
+  //
+  // Parameters:
+  // 1. user number to fetch the logs for
+  // 2. optional cursor to specify which logs to fetch
+  // 3. optional limit of the number of entries. If no limit is given, at most 1000 entries are returned.
+  get_user_logs : (UserNumber, opt Cursor, opt nat16) -> (UserLogs) query;
```
nmattia, line 129: Out of curiosity, why are both functions offered? In particular, do you think `get_logs` will be useful? Or is that a side effect of testing?

src/archive/archive.did
```diff
@@ -0,0 +1,143 @@
...
+  next_token: opt blob // cursor to fetch the next page of log entries (if any)
+};
+
+type Logs = record {
+  entries: vec blob;
+  next_idx: opt nat64 // next index to fetch the next page of log entries (if any)
+};
+
+type ArchiveInit = record {
+  ii_canister: principal;
```
nmattia, line 116: how about not actually mentioning II here or in the `write_entry` doc below?

I'm thinking like this: someone who sees `ii_canister` may assume that somehow the archive canister needs to authenticate or call the II canister to function; if the attribute name was e.g. `owner` or `writer` or something more descriptive, then it would be clearer

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +173,92 @@ pub struct HttpResponse {
...
+    UpdateDevice {
+        updated_device: PublicKey,
+        changed_data: DeviceDataUpdate,
+    },
+    #[serde(rename = "remove_device")]
+    RemoveDevice { removed_device: PublicKey },
+}
+
+#[derive(Eq, PartialEq, Clone, Debug, CandidType, Deserialize)]
+pub struct LogEntry {
```
nmattia, line 197: nit: now that the canister is called "archive", should this just be an "entry"? Same for `Logs` and where applicable

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +173,92 @@ pub struct HttpResponse {
...
+    pub credential_id: Option<CredentialId>,
+    pub purpose: Option<Purpose>,
+    pub key_type: Option<KeyType>,
+    pub protection: Option<DeviceProtection>,
+}
+
+// Placeholder for information that has been hidden for privacy reasons.
+#[derive(Eq, PartialEq, Clone, Debug, CandidType, Deserialize)]
+pub enum Hidden {
+    HiddenForPrivacyReasons,
```
nmattia, line 230: how about "Private"?

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +173,92 @@ pub struct HttpResponse {
...
+pub enum Cursor {
+    // timestamp of the next entry not included in this response, if any
+    #[serde(rename = "timestamp")]
+    Timestamp { timestamp: Timestamp },
+    // index of the next entry not included in this response, if any
+    #[serde(rename = "next_token")]
+    NextToken { next_token: ByteBuf },
+}
+
+/// Init arguments of the II archive canister.
```
nmattia, line 259:
```suggestion
/// Init arguments of the archive canister.
```

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +173,92 @@ pub struct HttpResponse {
...
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub enum Cursor {
+    // timestamp of the next entry not included in this response, if any
+    #[serde(rename = "timestamp")]
+    Timestamp { timestamp: Timestamp },
+    // index of the next entry not included in this response, if any
+    #[serde(rename = "next_token")]
+    NextToken { next_token: ByteBuf },
```
nmattia, line 256: out of curiosity, why "token"? also should the `next_token` be an opaque type like `Index` or something, or does it need to be a `ByteBuf` (because somehow read and interpreted by the client)?

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +173,92 @@ pub struct HttpResponse {
...
+    #[serde(rename = "register_anchor")]
+    RegisterAnchor {
+        initial_device: DeviceDataWithoutAlias,
+    },
+    #[serde(rename = "add_device")]
+    AddDevice { new_device: DeviceDataWithoutAlias },
+    #[serde(rename = "update_device")]
+    UpdateDevice {
+        updated_device: PublicKey,
+        changed_data: DeviceDataUpdate,
```
nmattia, line 190: I'll admit that when reading `changed_data` I'm not sure if it's the data that was changed (i.e. the old data) or the current/new data

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +173,92 @@ pub struct HttpResponse {
 pub struct InternetIdentityInit {
     pub assigned_user_number_range: (UserNumber, UserNumber),
 }
+
+// Archive specific types
+
+#[derive(Eq, PartialEq, Clone, Debug, CandidType, Deserialize)]
+pub enum OperationType {
```
nmattia, line 180: throughout this type it may be nice to not repeat the operation name in each `..._device` field, but instead simply use `device: ...`, which may also be much nicer to use for any client written in a language with row polymorphism

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+use std::cell::RefCell;
+
+type Memory = RestrictedMemory<DefaultMemoryImpl>;
+type StableLog = Log<VirtualMemory<Memory>, VirtualMemory<Memory>>;
+type ConfigCell = StableCell<ArchiveConfig, Memory>;
+type UserIndex = StableBTreeMap<VirtualMemory<Memory>, UserIndexKey, ()>;
+
+const GIB: u64 = 1 << 30;
+const WASM_PAGE_SIZE: u64 = 65536;
+/// The maximum number of Wasm pages that we allow to use for the stable storage.
```
nmattia, line 24: Guessing `8 * GIB` is the canister max memory?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+        let buf = candid::encode_one(&self).expect("failed to encode log config");
+        Cow::Owned(buf)
+    }
+
+    fn from_bytes(bytes: Vec<u8>) -> Self {
+        candid::decode_one::<ArchiveConfig>(&bytes).expect("failed to decode log options")
+    }
+}
+
+impl Default for ArchiveConfig {
```
nmattia, line 107: I'm uneasy about this implementation:

* The canister ID really should be configurable. If it isn't set, it should clearly be so (e.g. `None`); not some default incorrect value that we can't detect
* The `max_entries_per_call` don't seem to be set anywhere, and seem to always be set to `1000` (making it not a default but a constant value)
* The `last_upgrade_timestamp` is clearly wrong; the canister was never upgraded in Jan 1970

Is this just a matter of making the compiler happy? I'm guessing that after the first init, we should have meaningful values for the canister ID and the last upgrade timestamp

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+struct ArchiveConfig {
+    /// This canister will accept entries only from this principal.
+    ii_canister: Principal,
+    /// The maximum number of entries returned in a single read canister call.
+    max_entries_per_call: u16,
+    /// Timestamp of the last install / upgrade of this canister.
+    last_upgrade_timestamp: Timestamp,
+}
+
+impl Storable for ArchiveConfig {
```
nmattia, line 96: might be a silly question: can't this be derived?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+    MEMORY_MANAGER.with(|cell| f(&*cell.borrow()))
+}
+
+/// A helper function to access the log.
+fn with_log<R>(f: impl FnOnce(&StableLog) -> R) -> R {
+    LOG.with(|cell| f(&*cell.borrow()))
+}
+
+/// A helper function to access the user index.
+fn with_user_index<R>(f: impl FnOnce(&mut UserIndex) -> R) -> R {
```
nmattia, line 81:
```suggestion
fn with_user_index_mut<R>(f: impl FnOnce(&mut UserIndex) -> R) -> R {
```

nit

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    /// Index to efficiently filter entries by user number.
+    static USER_INDEX: RefCell<UserIndex> = with_memory_manager(|memory_manager| {
+        RefCell::new(StableBTreeMap::new(memory_manager.get(USER_INDEX_MEMORY_ID), USER_INDEX_KEY_LENGTH as u32, 0))
+    });
+}
+
+/// Creates a memory region for the configuration stable cell.
+fn config_memory() -> Memory {
+    RestrictedMemory::new(DefaultMemoryImpl::default(), 0..1)
```
nmattia, line 57: I'm a little uneasy about those bare indices for ranges, here and below. Are they related to the `..._MEMORY_ID`s that you define above?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
+use candid::{CandidType, Deserialize, Principal};
```
nmattia, line 1: Would be nice to have a few words about the memory layout, I'm struggling to wrap my head around it

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+const WASM_PAGE_SIZE: u64 = 65536;
+/// The maximum number of Wasm pages that we allow to use for the stable storage.
+const MAX_WASM_PAGES: u64 = 8 * GIB / WASM_PAGE_SIZE;
+
+const LOG_INDEX_MEMORY_ID: MemoryId = MemoryId::new(0);
+const LOG_DATA_MEMORY_ID: MemoryId = MemoryId::new(1);
+const USER_INDEX_MEMORY_ID: MemoryId = MemoryId::new(2);
+
+/// Length of the user index key. Changing this value requires a stable memory migration.
+const USER_INDEX_KEY_LENGTH: usize = 24;
```
nmattia, line 32: out of curiosity, where does `24` come from?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+use internet_identity_interface::*;
+use metrics_encoder::MetricsEncoder;
+use serde_bytes::ByteBuf;
+use std::borrow::Cow;
+use std::cell::RefCell;
+
+type Memory = RestrictedMemory<DefaultMemoryImpl>;
+type StableLog = Log<VirtualMemory<Memory>, VirtualMemory<Memory>>;
+type ConfigCell = StableCell<ArchiveConfig, Memory>;
+type UserIndex = StableBTreeMap<VirtualMemory<Memory>, UserIndexKey, ()>;
```
nmattia, line 20: Where I saw `UserIndex` below I though a `UserIndex` would be used to somehow index users, whereas this is like an index (collection,mapping) of users, right? Do you think we could clarify that?

src/archive/archive.did
nmattia, line 129: What is a "user"? do you mean "anchor"? "user number" was dropped in favor of "anchor", though all occurrences of the term "user number" haven't be replaced yet

(applies here and below and in main.rs)

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+        for idx in start_idx..start_idx + num_entries {
+            let entry = match log.get(idx) {
+                None => break,
+                Some(entry) => entry,
+            };
+            entries.push(Some(
+                candid::decode_one(&entry).expect("failed to decode log entry"),
+            ))
+        }
+        Logs { entries, next_idx }
```
nmattia, line 206: This requires building a big object containing all the entries, which may be really heavy on the heap. Have you considered a streaming approach?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+            Some(idx) => idx as usize,
+        };
+
+        let next_idx = if start_idx + num_entries < length {
+            Some((start_idx + num_entries) as u64)
+        } else {
+            None
+        };
+
+        let mut entries = Vec::with_capacity(log.len().min(num_entries));
```
nmattia, line 196: Why build the `Vec` with capacity `min(log_size, num_entries)`?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+    USER_INDEX.with(|cell| f(&mut *cell.borrow_mut()))
+}
+
+/// Configuration of the archive canister.
+#[derive(Clone, Debug, CandidType, Deserialize)]
+struct ArchiveConfig {
+    /// This canister will accept entries only from this principal.
+    ii_canister: Principal,
+    /// The maximum number of entries returned in a single read canister call.
+    max_entries_per_call: u16,
```
nmattia, line 91: is this not reflected in the candid type?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+            last_upgrade_timestamp: 0,
+        }
+    }
+}
+
+/// Index key for the user index.
+#[derive(Debug)]
+struct UserIndexKey {
+    user_number: UserNumber,
+    timestamp: Timestamp,
```
nmattia, line 121: I'm a little confused; why is there a timestamp in the key? _EDIT: see my question around the `UserIndex` type instead_

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +173,92 @@ pub struct HttpResponse {
...
+
+// Placeholder for information that has been hidden for privacy reasons.
+#[derive(Eq, PartialEq, Clone, Debug, CandidType, Deserialize)]
+pub enum Hidden {
+    HiddenForPrivacyReasons,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub struct Logs {
+    // make this a vec of options to keep LogEntry extensible
```
nmattia, line 235: why does this keep `LogEntry` extensible? if it's a matter of wrapping it in an enum, can we make a single-constructor enum with an appropriate name?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+        limit
+            .map(|l| l.min(config.max_entries_per_call))
+            .unwrap_or(config.max_entries_per_call) as usize
+    })
+}
+
+#[init]
+#[post_upgrade]
+fn init(maybe_arg: Option<ArchiveInit>) {
+    if let Some(arg) = maybe_arg {
```
nmattia, line 264: this is a `match`, right?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+                .iter()
+                .map(|(_, entry)| candid::decode_one(&entry).expect("failed to decode log entry"))
+                .collect();
+
+            UserLogs { entries, cursor }
+        })
+    })
+}
+
+fn sanitize_limit(limit: Option<u16>) -> usize {
```
nmattia, line 253: nit: I find it odd to have an `Option` as argument for something that's supposed to be "sanitizing" the input. I'd only expected a number or string to be sanitized. How about renaming (`limit_or_default` or I don't know what) or move everything unrelated to sanitization to the call site?

src/archive/src/main.rs
```diff
@@ -0,0 +1,394 @@
...
+}
+
+#[query]
+fn get_user_logs(user_number: u64, cursor: Option<Cursor>, limit: Option<u16>) -> UserLogs {
+    let num_entries = sanitize_limit(limit);
+
+    with_user_index(|index| {
+        let iterator = match cursor {
+            None => index.range(user_number.to_le_bytes().to_vec(), None),
+            Some(Cursor::NextToken { next_token }) => index.range(next_token.into_vec(), None),
```
nmattia, line 217: do these implicitly rely on the `Storable` implementation?

src/archive/src/main.rs
nmattia, line 20: I've been squinting at the index code for a while and still don't feel confident about my understanding of how it works/what it does.

What does it do? How does it achieve this? Why are all values of the BTree empty?

src/archive/archive.did
frederikrothenberger: So far it has been convention that `service` is the very last definition of a candid file (mostly because it requires all previous definitions, at least transitively), Internet Identity does it this way, but also the management canister and all canisters in the IC repo. I would not change this here.

But I agree, that adding a comment on the types is helpful. I will do that.

src/archive/archive.did
frederikrothenberger: Yes, I think `get_logs` is useful, especially to diagnose errors, e.g. if a message gets lost to pin-point the timestamp where `sequence_number` went out of sync.

src/archive/archive.did
frederikrothenberger: 👍

src/archive/src/main.rs
frederikrothenberger: I refactored it, so that the code can now correctly express the uninitialized state.

src/archive/src/main.rs
frederikrothenberger: No, unfortunately not.

src/archive/src/main.rs
frederikrothenberger: Currently, yes.

src/internet_identity_interface/src/lib.rs
frederikrothenberger: I took the `token` terminology from the Twitter pagination api.

Yes the `next_token` should be treated as an opaque blob (which `ByteBuf` represents). It needs correct typings (i.e. instead of candid `reserved`) because the client needs to decode it as part of the response an re-encode it as part of the next request.

src/internet_identity_interface/src/lib.rs
frederikrothenberger: I changed the field name to `new_values` which should hopefully make it more clear.

src/archive/src/main.rs
frederikrothenberger: These bare indices just separate out the first memory page from the managed memory. I added documentation which explains this.

src/archive/src/main.rs
frederikrothenberger: It is the size of the byte representation of the `AnchorIndexKey` struct. But I have replaced it now with `std::mem::size_of::<AnchorIndexKey>()`.

src/archive/src/main.rs
frederikrothenberger: Well, it's and index for entries by user/anchor. I added a comment to clarify.

src/archive/src/main.rs
frederikrothenberger: We need to candid encode the result. Candid encoding currently does not support streaming.

src/archive/src/main.rs
frederikrothenberger: This is there to allow finding entries for an anchor based on the `timestamp` as well (which might be useful for anchors with a lot of activity or just after using it for a long time).

Added a clarifying comment.

src/archive/src/main.rs
frederikrothenberger: You're right, this was missing --> added.

src/archive/src/main.rs
frederikrothenberger: I added a comment explaining this as a top level module doc.

src/archive/src/main.rs
frederikrothenberger: Yes, they are tied to the `AnchorIndexKey`. The `next_token` is directly the byte representation obtained by the `Storable` implementation.

src/archive/src/main.rs
frederikrothenberger: This was there for the case when there are less than `num_entries` in the log total. But since this will only be true for a very short time (and even then it's just an irrelevant optimization), I have removed it again.

Dockerfile
frederikrothenberger: No, because docker will build only one target at a time.

scripts/build
```diff
@@ -11,31 +11,34 @@ cd "$SCRIPTS_DIR/.."
...
-    echo "Build Internet Identity"
+    echo "Builds Internet Identity and Archive Canisters"
 }
 
 function usage() {
     cat << EOF
 
 Usage:
-  $0 [--only-dependencies]
+  $0 [--only-dependencies] [CANISTER...]
```
nmattia, line 21: nit: CLI flags like `--internet-identity` or `--archive` are usually spelled out; `CANISTER` would be a argument that is free (i.e. read as a string) (also I've rarely seen underscores like `--internet_identity` (while it's more consistent with the internal naming of our canister, it's less consistent with CLI conventions 🙃 ))

src/archive/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+                TryFrom::try_from(&bytes[16..]).expect("failed to read log_index"),
+            ),
+        }
+    }
+}
+
+#[update]
+fn write_entry(anchor: Anchor, timestamp: Timestamp, entry: ByteBuf) {
+    with_config(|config| {
+        if config.ii_canister != caller() {
```
nmattia, line 205: out of curiosity, is there an `api` call to figure out the controller of this canister?

src/archive/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+
+        // the only way this expect can trigger is when the key size is wrong.
+        // On other failures (e.g. no more stable memory available) the underlying StableBTreeMap will panic directly.
+        index.insert(key, ()).expect("bug: key size mismatch");
+    })
+}
+
+#[query]
+fn get_entries(index: Option<u64>, limit: Option<u16>) -> Entries {
+    let num_entries = limit_or_default(limit);
```
nmattia, line 232: this is still a `limit` and not necessarily the number of entries returned, right? why not `let limit = limit_or_default(limit)`?

src/archive/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+
+    with_log(|log| {
+        let length = log.len();
+        let start_idx = match index {
+            None => length.saturating_sub(num_entries),
```
nmattia, line 237: If `index` is not specified, we only return the _last_ `limit` entries, is that correct? Why not return `0 .. limit`? Also would be good to mention this in the API description:

> If no limit is given, at most the configured number (see ArchiveInit) entries are returned.

src/archive/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+            Some(idx) => idx as usize,
+        };
+
+        let next_idx = if start_idx + num_entries < length {
```
nmattia, line 241: If I understand correctly you check that the log _has_ a `next_idx`, i.e. whether the log has more entries. Is that useful information? Won't `next_idx` exist eventually when more entries are added?

src/archive/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+                })
+                .collect();
+
+            let cursor = if entries.len() > num_entries {
+                entries.pop().map(|(key, _)| Cursor::NextToken {
+                    next_token: ByteBuf::from(key.to_bytes()),
+                })
+            } else {
+                None
+            };
```
nmattia, line 292: Wouldn't it be simpler to first extract `num_entries` and then read `iterator.nth(num_entries + 1)`? Looks like the index [only implements `next()`](https://docs.rs/ic-stable-structures/latest/src/ic_stable_structures/btreemap/iter.rs.html#89), so I'm guessing this is an optimization to avoid iterating twice?

src/internet_identity/tests/README.md
```diff
@@ -4,7 +4,7 @@ This is a test suite for the Internet Identity canister. To run the tests:
 
 ``` bash
 # Make sure II is built with the "test" flavor
-II_DUMMY_CAPTHA=1 ./scripts/build
+II_DUMMY_CAPTHA=1 ./scripts/build --internet_identity
```
nmattia, line 7: thinking about it now, should it just build II by default?

src/internet_identity_interface/src/lib.rs
```diff
@@ -173,3 +174,90 @@ pub struct HttpResponse {
...
+}
+
+// Placeholder for information that has been hidden for privacy reasons.
+#[derive(Eq, PartialEq, Clone, Debug, CandidType, Deserialize)]
+pub enum Private {
+    HiddenForPrivacyReasons,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub struct Entries {
```
nmattia, line 233: `Entries` itself (and `AnchorEntries`) is never serialized to stable memory, right? why do we care about keeping this extensible?

nmattia (commented):
Couple of comments, but this is looking really good! thank you so much!

src/archive/src/main.rs
frederikrothenberger: There is. One could call the `ManagementCanister` for the `canister_status` and extract it from there. But I think separating canister ownership from authorization to write entries is sensible, even if we keep them the same for now.

src/archive/src/main.rs
frederikrothenberger: The problem is that `collect` consumes the iterator and iterating twice seems wasteful. Are you ok with leaving this as is? Or have a suggestion to structure this more nicely?

nmattia: > iterating twice seems wasteful

Ok, that's what I wanted to clarify.

I'm completely fine with it as is! I wanted to make sure I understood what was happening, so I think it's at least worth a comment.

Another approach would be to use `.next()` directly and explicitly; might be a little more verbose but the intent would be clearer. Happy with current implementation though.

src/archive/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+fn limit_or_default(limit: Option<u16>) -> usize {
+    with_config(|config| {
+        limit
+            .map(|l| l.min(config.max_entries_per_call))
+            .unwrap_or(config.max_entries_per_call) as usize
+    })
+}
+
+#[init]
+#[post_upgrade]
```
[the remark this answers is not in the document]
nmattia: > The interface for init and post_upgrade must be the same.

Interesting! How come? Is that a canister interface limitation, or a silly candid spec limitation?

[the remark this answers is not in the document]
frederikrothenberger: I think it's a bit of both:
* Candid cannot express the differing interfaces because it uses only one signature for services: `service: () -> {}`
* Some component in the stack relies on there being only one component, as I get a candid type error if I try to make the interfaces different.

src/archive/archive.did
```diff
@@ -0,0 +1,148 @@
...
+    cross_platform;
+    seed_phrase;
+};
+
+type DeviceProtection = variant {
+    protected;
+    unprotected;
+};
+
+type OperationType = variant {
```
nmattia, line 24: nit: this is not really an `OperationType` but a full blown operation, right?

src/archive/src/main.rs
```diff
@@ -0,0 +1,469 @@
...
+    timestamp: Timestamp,
+    log_index: u64,
+}
+
+impl AnchorIndexKey {
+    /// Returns bytes corresponding only to the timestamp and log_index component of the anchor index key.
+    /// This is useful as an offset when doing a range scan with the anchor component.
+    fn to_anchor_offset(&self) -> Vec<u8> {
+        let mut buf =
+            Vec::with_capacity(std::mem::size_of::<Timestamp>() + std::mem::size_of::<u64>());
```
nmattia, line 183: How about making the log index a type alias, just for clarity?

src/archive/src/main.rs
```diff
@@ -0,0 +1,469 @@
...
+            return ConfigState::Uninitialized;
+        }
+        ConfigState::Initialized(
+            candid::decode_one::<ArchiveConfig>(&bytes).expect("failed to decode archive config"),
+        )
+    }
+}
+
+/// Index key for the anchor index.
+/// Changing the (serialized) size of this value requires a stable memory migration.
```
nmattia, line 170: What would happen without migration? Would this be caught by the type system or by tests?

src/archive/src/main.rs
```diff
@@ -0,0 +1,469 @@
...
+            Vec::with_capacity(std::mem::size_of::<Timestamp>() + std::mem::size_of::<u64>());
+        buf.extend(&self.timestamp.to_be_bytes());
+        buf.extend(&self.log_index.to_be_bytes());
+        buf
+    }
+}
+
+/// Storable implementation for the index key.
+/// Note: byte ordering is very important as the keys are sorted on a byte level (lower to higher index)
+/// --> use big endian to ensure that the most significant bytes are compared first
```
nmattia, line 192: If I understand correctly, the following entry would be encoded like this:
* anchor `10042` at time `123456`, at index `1`: `100421234561`

This way, by scanning/fetching indices lexicographically, we ensure that all anchor operations will be listed together, ordered first by timestamp, then by index. We can then deserialize the key and retrieve the index (last 8 bytes), used to index the log.

Is that correct?

* This wasn't trivial to figure out; how about adding a line or two about this?
* It looks like we rely on each field (anchor, timestamp) to have a known, fixed sized. What if the size changes (for instance changing the type of `Anchor`, do we fail silently? Does that invalidate the entire index?

src/archive/src/main.rs
nmattia: I think that for simplicity's sake we may want to default to this when integrating with II (passing arguments is just another point of failure) although I'm guessing this'll stay necessary for testing, maybe?

src/archive/src/main.rs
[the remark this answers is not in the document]
nmattia: Gotcha! I think it could be interesting to create discussion on the SDK repo

src/canister_tests/src/framework.rs
```diff
@@ -32,7 +32,21 @@ lazy_static! {
...
+    /** The Wasm module for the current archive build, i.e. the one we're testing */
+    pub static ref ARCHIVE_WASM: Vec<u8> = {
+        let def_path = path::PathBuf::from("..").join("..").join("archive.wasm");
+        let err = format!("
+        Could not find Archive Wasm module for current build.
+
+        I will look for it at {:?}, and you can specify another path with the environment variable ARCHIVE_WASM (note that I run from {:?}).
+
+        In order to build the Wasm module, please run the following command:
+            ./scripts/build --archive
```
nmattia, line 44: Should we release the canister? Or alternatively, should we suggest a command to run everything but the archive canister tests?

nmattia (approved):
This is great, thanks a LOT for doing this!

src/archive/src/main.rs
frederikrothenberger: I expanded the documentation with this example. 👍

src/archive/src/main.rs
frederikrothenberger: I added unit tests that catch unexpected changes.

src/archive/src/main.rs
frederikrothenberger: I just realized that in `init` and `post_upgrade` we cannot do a call to the management canister. So in order to find out the controller, we would have to delay initialization of the config into the first `update` call, which imho adds more complexity than simply passing in the argument.

I will leave this arg as it is for now and we can revisit when we do the integration with II.

PR #906, opened 2022-09-22
Add missing pages to showcase

nmattia:
This adds two pages that are present in the screenshot tests, but which weren't yet shown on the showcase. This means that all screenshot pages can now be browsed on the showcase.


frederikrothenberger (approved):
LGTM!

PR #907, opened 2022-09-22
Extract anchor input component

nmattia:
This extracts the anchor input component (with input filter) from the authenticate page, which allows reuse in both the login and anchor prompt pages.

Also removes unnecessary anchor input CSS.


src/frontend/src/components/anchorInput.ts
```diff
@@ -0,0 +1,84 @@
+import { html } from "lit";
+import { createRef, ref } from "lit/directives/ref.js";
+import { TemplateRef } from "../utils/templateRef";
+
+/** A component for inputting an anchor number */
+export const mkAnchorInput = (
```
peterpeterparker, line 6: `mk...` like `mock`?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -46,8 +47,8 @@ const pageContent = (
   const onRecoverClick = () => useRecovery(connection, readUserNumber());
 
   const authorizeButton = createRef();
-  const userNumberInput = createRef();
   const registerButton = createRef();
+  const anchorInput = mkAnchorInput("userNumberInput", userNumber);
```
peterpeterparker, line 51: for fun

```
const {template, refs: {userNumberInput}} = mkAnchorInput("userNumberInput", userNumber);
```

peterpeterparker (approved):
LGTM

Note: while reviewing this I just assumed that what's new is the component and new export of the template and refs - i.e. that the rest was just code moving around

src/frontend/src/components/anchorInput.ts
nmattia: ah no, like `make`! what do you suggest instead?

src/frontend/src/flows/authenticate/index.ts
nmattia: I think `ts` would then complain about `template` being bound again...

nmattia:
> the rest was just code moving around

Close! The `setInputFilter` was changed to use "lit" event binder, and the `flash-error` animation was removed, but in essence yes it's just code moved around

meodai (approved):
Looks good, cant wait for that!

src/frontend/src/components/anchorInput.ts
peterpeterparker: oki doki, all good, thanks

nmattia: clarified offline

PR #909, opened 2022-09-23
Update and showcase registerDisabled

nmattia:
The "Registration Disabled" screen was missed during the addition of CSS Reset and was missing from the showcase.

This updates the code -- both CSS and JS (to using better lit code) -- and adds the screen to the showcase.

### DIFF
--- a/src/frontend/src/flows/registerDisabled.ts
+++ b/src/frontend/src/flows/registerDisabled.ts
@@ -14,14 +14,17 @@ const pageContent = (onCancel: () => void) => html`
         message: html`<p class="t-paragraph t-lead">
             You are <b class="t-strong">not</b> browsing this website on the
             expected URL:
-            <a class="t-link" href="https://identity.ic0.app">https://identity.ic0.app</a>. For
-            security reasons creation of new Internet Identity anchors is
+            <a class="t-link" href="https://identity.ic0.app"
+              >https://identity.ic0.app</a
+            >. For security reasons creation of new Internet Identity anchors is
             disabled on this origin.
           </p>
           <p class="t-paragraph">
             Please switch to
-            <a class="t-link" href="https://identity.ic0.app">https://identity.ic0.app</a> to
-            register a new Internet Identity anchor.
+            <a class="t-link" href="https://identity.ic0.app"
+              >https://identity.ic0.app</a
+            >
+            to register a new Internet Identity anchor.
           </p>
           <p class="t-paragraph">
             If you were redirected here by another website, please inform the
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/registerDisabled.ts
```diff
@@ -1,47 +1,55 @@
...
+            developers. More information is provided
+            <a
+              class="t-link"
+              href="https://forum.dfinity.org/t/internet-identity-proposal-to-deprecate-account-creation-on-all-origins-other-than-https-identity-ic0-app/9760"
+              >here</a
+            >.
+          </p>`,
+      })}
+    </div>
+    <div class="l-stack">
```
peterpeterparker, line 37: don't know well the css of II but not sure you need that extra `div` for styling purpose

src/frontend/src/flows/registerDisabled.ts
```diff
@@ -1,47 +1,55 @@
...
-
-const init = (): Promise<null> =>
-  new Promise((resolve) => {
-    const deviceAliasCancel = document.getElementById(
-      "deviceAliasCancel"
-    ) as HTMLButtonElement;
-    deviceAliasCancel.onclick = () => {
-      resolve(null);
-    };
+export const registerDisabled = (): Promise<null> => {
```
peterpeterparker, line 45: `Promise<null>` is correct, you do something with the `null` result or it should actually be `Promise<void>`?

peterpeterparker (approved):
LGTM

src/frontend/src/flows/registerDisabled.ts
nmattia: It's actually expecting a `null`, which means "user canceled the flow". Agree it's a bit ugly that it trickles down here and the caller should probably translate `null` into `void` but that'd involve some bigger changes.

peterpeterparker: If you would review the PR, you would probably suggest to add a comment then 😉. 
Thanks for the feedback, gotcha

src/frontend/src/flows/registerDisabled.ts
nmattia: Without the `div` the button is too close to the text, even if `l-stack` is set on the button. Not 100% sure why but the `div` doesn't really bother me

peterpeterparker: Can live with it, was just comment, PR is already accepted.

Being said, if I had to bet, I would bet this `div` is not needed.

PR #910, opened 2022-09-23
Update addDevice welcome page

nmattia:
This updates the "welcome view" of the add device flow to use lit `@click` and the new anchor input component for consistency.


src/frontend/src/flows/login/unknownAnchor.ts
```diff
@@ -145,7 +145,7 @@ const initLinkDevice = (connection: Connection) => {
       "registerUserNumber"
     ) as HTMLInputElement;
 
-    const userNumber = parseUserNumber(userNumberInput.value);
-    await addRemoteDevice(userNumber, connection);
+    const userNumber = parseUserNumber(userNumberInput.value) ?? undefined;
+    await addRemoteDevice(connection, userNumber);
```
peterpeterparker, line 149: note: for these kind of stuffs I find object params for function more reliable `addRemoteDevice({connection, userNumber})` - order of the params becomes irrelevant

src/frontend/src/flows/addDevice/welcomeView/index.ts
```diff
@@ -44,24 +42,15 @@ const pageContent = (userNumber: bigint | null) => html`
  * This shows a prompt to enter the identity anchor to add this new device to.
  */
 export const addRemoteDevice = async (
```
peterpeterparker, line 44: this function doesn't need to be `async` if I am not wrong

peterpeterparker (approved):
LGTM

src/frontend/src/flows/login/unknownAnchor.ts
nmattia: agreed, but the rest of the codebase doesn't do that and I'd rather stay consistent

PR #917, opened 2022-09-26
Clean up deviceSettings code

nmattia:
This factors out the button logic and uses lit `@click` instead of an `init` method, which clarifies the code. Logic wasn't changed.


frederikrothenberger (approved):
Thanks, lgtm!

PR #918, opened 2022-09-26
Make anchor inputs more prominent

meodai:
https://user-images.githubusercontent.com/608386/192300105-621800c3-8114-43f0-86cc-969a9fffa27e.mov


meodai:
Question: should it stay "fancy" once the anchor is in the field? `:focus` + `:not(:empty)`?

src/frontend/src/flows/displayUserNumber.ts
```diff
@@ -14,7 +14,7 @@ const pageContent = (userNumber: bigint) => html`
     })}
     <h2 class="t-title">Identity Anchor:</h2>
 
-    <div class="c-input c-input--vip c-input--readonly t-vip" data-usernumber>
+    <div class="c-input c-input--vip c-input--readonly t-vip c" data-usernumber>
```
nmattia, line 17:
```suggestion
    <div class="c-input c-input--vip c-input--readonly t-vip" data-usernumber>
```

typo?

nmattia (commented):
Looking really really nice!

One question: the halo stays the same colour even on bad key press, whereas before it'd turn blue-ish (bruised). Is that on purpose?

Also I've noticed that a solid white rectangle appears behind the input on bad input:

https://user-images.githubusercontent.com/6930756/192302443-e41064c8-4abc-45ad-ad34-acc5859b63e8.mov

nmattia (approved):
Looks great, thanks!

PR #920, opened 2022-09-26
Use lit-html instead of lit

nmattia:
I realized too late that `lit-html` also supports `ref` directives, meaning there is not reason to use the heavier `lit` package.


meodai (approved):
good to know! lgtm

PR #921, opened 2022-09-27
Provide meta information for SEO and social platforms

peterpeterparker:
Provide meta information (title, description, social image, more favicons) for SEO and social platforms.

Note: In NNS-dapp we use the `raw` domain to provide icons and social images because this was the only way until today to provide such meta data. However, this should now be resolved. That's why here I provide these using the main domain.


src/frontend/assets/favicons/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<browserconfig>
+    <msapplication>
+        <tile>
+            <square150x150logo src="/assets/mstile-150x150.png"/>
+            <TileColor>#522785</TileColor>
```
nmattia, line 6: This value appears in 3 different files and will probably eventually get out of sync. This is bound to get out of sync. Could we generate these files and/or set the value during the build?

nmattia (commented):
Thanks a lot!

Couple of thoughts:

* For the description (`og:description`, manifest.json) what about staying consistent with the description in the repo? Happy to go with either, the other one will need to change though. Same goes for `social-image.png`. "Internet Identity, a blockchain authentication system for the Internet Computer"
* Unless you prepared each image manually, can you share the script you used to generate all the sizes?
* If you actually want those assets to be served, you'll need to add them [here](https://github.com/dfinity/internet-identity/blob/e16def95440f783b5f1fc7d59f483d1af7eca8b3/src/internet_identity/src/assets.rs#L79) (and update the size of the array, currently `9`)

src/frontend/assets/favicons/browserconfig.xml
peterpeterparker: - Consitant is a good idea, I guess then the repo should be updated (description here comes from Pete or ping Slack channel). Note: I don't have enough privileges to update the repo description

- Images copied from NNS-dapp. Source IC logo. Probably generated manually back then or with [realfavicongenerator](https://realfavicongenerator.net/)

- I extended the `assets.ts`. Being said, not sure I use the correct "includer" (`include_bytes`) for JSON and XML and also not sure if it supports sub-folder in the path (`""/favicons/browserconfig.xml"`). Let me know what need to be modified.

- regarding generating files or values during build? yes. Will I provide it in this PR? no. Being said, not against, feel free to automate the process.

nmattia:
Note: this will blow up the canister size, and we'll need some better way of handling assets. Pausing this until this is resolved.

nmattia:
@peterpeterparker I'm cleaning up a bit but feel free to reopen; should be simpler now that the canister simply bundles the `dist/`

PR #922, opened 2022-09-27
Remove 'TemplateRef' abstraction

nmattia:
This removes the `TemplateRef` abstraction which turned out to be really inconvenient to use. Since introducing `TemplateRef` I also realized that `Ref<T>` actually carries phantom type data, making most of the `TemplateRef` benefits moot.


src/frontend/src/components/anchorInput.ts
```diff
@@ -37,6 +37,10 @@ export const mkAnchorInput = (
         @contextmenu=${inputFilter(isDigits, onBadInput)}
         @drop=${inputFilter(isDigits, onBadInput)}
         @focusout=${inputFilter(isDigits, onBadInput)}
+        @keypress=${onKeyPress ??
```
peterpeterparker, line 40: doesn't lit don't bind the keypress if undefined is provided - i.e. a fallback is really needed?

src/frontend/src/flows/login/unknownAnchor.ts
```diff
@@ -100,39 +109,41 @@ const initRecovery = (connection: Connection) => {
...
 
-  userNumberInput.onkeypress = (e) => {
-    // submit if user hits enter
-    if (e.key === "Enter") {
-      e.preventDefault();
-      loginButton.click();
-    }
-  };
+  withRef(userNumberInput, (userNumberInput) => {
+    userNumberInput.onkeypress = (e) => {
```
peterpeterparker, line 120: isn't this code duplicated? worth extracting a function?

src/frontend/src/utils/utils.ts
```diff
@@ -1,3 +1,20 @@
+import { Ref } from "lit-html/directives/ref.js";
```
peterpeterparker, line 1: not sure about this import, you might have issue someday if you switch to a module based bundler - just a feeling though

src/frontend/src/flows/login/unknownAnchor.ts
nmattia: yes, will do in a separate PR

src/frontend/src/utils/utils.ts
```diff
@@ -1,3 +1,20 @@
...
+
+// Read a "lit-html" ref, showing an error message (in the console) in case the
+// element is not available.
+export function withRef<A, B>(ref: Ref<A>, f: (val: A) => B): B | undefined {
+  const value = ref.value;
+
+  if (value === undefined) {
```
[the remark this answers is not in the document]
nmattia: no, the `value` can only be `undefined` (I think `null` would show up in the types)

src/frontend/src/components/anchorInput.ts
nmattia: good question, will check

peterpeterparker (approved):
LGTM

src/frontend/src/flows/login/unknownAnchor.ts
nmattia: you're right! updated

PR #924, opened 2022-09-27
Move canister test readme

frederikrothenberger:
This file was erronously not moved when implementing #892.


nmattia (approved):
Good catch, thanks!

PR #932, opened 2022-09-30
Always display error details

### LINKED ISSUE #928: Error details caret is misaligned
The system's summary/details caret is currently hidden and replaced with a `before` element, meaning it has no box and had to be shifted to the left. This means it is misaligned with the rest of the text, potentially overlapping with the warnbox border.

Screenshot 2022-09-29 at 18 05 52

nmattia:
Fixes #928 

Before this we'd hide error details behind a `<details>` element, but all error messages I've seen so far were very short, meaning they'd fit perfectly well on the page.

These changes always show the error message, which makes error pages more consistent and avoid styling issues re. the more dynamic page.

Screenshot 2022-09-30 at 13 00 41

### DIFF
--- a/src/frontend/src/components/displayError.ts
+++ b/src/frontend/src/components/displayError.ts
@@ -13,17 +13,13 @@ const pageContent = (options: ErrorOptions) => html`
     ${warnBox({
       title: options.title,
       message: options.message,
+      htmlElement: "div",
       slot:
         options.detail !== undefined
-          ? html`<details class="displayErrorDetail">
-              <summary class="c-summary">
-                <span class="c-summary__link t-link">Error details</span>
-              </summary>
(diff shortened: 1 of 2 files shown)


frederikrothenberger:
The error detail could also be `Resource https://5s2ji-faaaa-aaaaa-qaaaq-cai.ic0.app/.well-known/ii-alternative-origins has too many entries: To prevent misuse at most 10 alternative origins are allowed.` or exception texts. So I'm not so sure about them always being short.

Also, the detail is usually technical so for most users the content might not be understandable.

But I don't feel strongly about showing it by default. It might also help developers if users take screenshots of the error messages.

frederikrothenberger (approved):

PR #934, opened 2022-10-03
Generate screenshots from showcase

nmattia:
This extracts the screenshot taking infrastructure out of the selenium tests, simplifies it, and adapts it to use the showcase instead. There are a couple differences:

* The screenshots are not numbered anymore, just given a name representative of what they are
* The screenshots are split between `screenshots/mobile/` and `screenshots/desktop/`, which makes browsing screenshots much nicer
* Some pages (esp. on mobile) are too tall; this is due to https://github.com/webdriverio/webdriverio/issues/8903. I've spent quite some time trying to work around it but haven't find a way yet to have the correct height with the new setup.
* The desktop pages are not as huge as they used to which speeds up the screenshots a fair bit and greatly reduces the size of the files.


meodai:
LGTM. But I am no expert. I'd feel more confortable if someone else would review this.

screenshots/README.md
```diff
@@ -1,20 +1,13 @@
...
-    npm run test:e2e
+``` bash
+npm run showcase&
+npm run screenshots
+```
 
-That will override screenshots, but not delete files that are no longer
-generated by the test; in that case, run `rm -f screenshots/**/*.png` before
-running this script.
+That will generate screenshots in `screenshots/custom/`.
```
frederikrothenberger, line 10: Maybe directly ignore this folder so that people do not accidentally commit their `custom` screenshots?

frederikrothenberger (approved):
Just two nits, thanks!

src/frontend/screenshots.ts
```diff
@@ -0,0 +1,141 @@
...
+ * * https://github.com/webdriverio/webdriverio/issues/8903
+ */
+function readScreenshotsConfig(): {
+  mobileEmulation?: {
+    deviceMetrics: { width: number; height: number };
+  };
+} {
+  const screenshotsType = process.env["SCREENSHOTS_TYPE"];
+  switch (screenshotsType) {
+    case "mobile":
```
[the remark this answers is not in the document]
nmattia: Yes, it's still necessary because the e2e tests use it. I'm not sure how much signal we get from the mobile tests, but they do tend to fail more often than the desktop ones, so _some_ signal I guess...

screenshots/README.md
frederikrothenberger: Neither do I, but I have a slight preference for adding it to the ignore file, just because I'm sure at some point this directory will show up in one of my PR diffs accidentally ... 😉

nmattia: mh I have a slight preference to not add it to the ignore file, because really it's not something that should live there for long (unlike for instance node_modules, etc). Will leave it out for now!

PR #937, opened 2022-10-05
Update chrome driver version in using-dev-build

frederikrothenberger:
The chrome driver is no longer compatible with current stable chrome leading to CI failures.


frederikrothenberger:
Apparently not all the runners get the `chrome 106` version yet (as this one here did: https://github.com/dfinity/internet-identity/actions/runs/3188512268/jobs/5201316370#step:3:38).

PR #938, opened 2022-10-06
Update typescript and webdriverio

nmattia:
This updates the webdriverio dependency to include https://github.com/webdriverio/webdriverio/pull/8945, which simplifies our screenshot code and makes sure all mobile screenshots have the same size (which does not depend on the size of the webpage).

In order to do this, `ts-loader` had to be updated, which led to a few other typescript-related package updates. In turn (and in order to please typescript & eslint) some bits of code had to be made more robust (yay).


frederikrothenberger (approved):
Thanks, lgtm!

PR #943, opened 2022-10-10
Move heap state handling into a separate module

frederikrothenberger:
The heap state is now only accessible via the state module. Various helper methods allow reading / modifying the state without having to deal with the `RefCell`s.

Some statements in the various functions of `main.rs` were reordered or slightly modified to:
- reduce nesting
- reduce parameter lists
- fix bugs related to initialization of the salt / and trap_if_not_authenticated
  (`caller()` must not be accessed after an `async` function invocation)

This is a preparation step to split the main file into further smaller modules handling different concerns.

### DIFF
--- a/src/internet_identity/src/assets.rs
+++ b/src/internet_identity/src/assets.rs
@@ -2,7 +2,7 @@
 //
 // This file describes which assets are used and how (content, content type and content encoding).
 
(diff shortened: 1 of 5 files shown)


nmattia (approved):
LGTM, thanks a lot!

PR #948, opened 2022-10-11
Makes the chasm fit the design system

meodai:
Also cleans up the "arrow" icon a bit


nmattia (approved):
Ah yes, that button is looking slick now!

PR #949, opened 2022-10-11
Update bash for mac os clean build workflows

frederikrothenberger:
The build script uses bash features that are not available on mac os default bash (v3). This adds step to update bash for mac os runners.


frederikrothenberger:
CI run of the new bash step: https://github.com/dfinity/internet-identity/actions/runs/3227339588/jobs/5282496303#step:4:11

.github/workflows/canister-tests.yml
```diff
@@ -535,6 +535,14 @@ jobs:
         with:
           name: internet_identity_production.wasm
           path: .
+      - run: |
```
nmattia, line 538:
```suggestion
      # Install recent bash (default bash 3 is lacking important features)
      - run: |
```

.github/workflows/canister-tests.yml
```diff
@@ -535,6 +535,14 @@ jobs:
...
+          /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+          brew update
+          brew install bash
+          echo "/usr/local/bin" >> $GITHUB_PATH
```
nmattia, line 542: is this necessary? how come `bash --version` below works already if bash is not in the PATH?

nmattia (commented):
Nice!

.github/workflows/canister-tests.yml
frederikrothenberger: Yes, I think so.
`brew` updates the the current `PATH` (I think), but this would be lost in the next step: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path

nmattia (approved):
Thanks!

.github/workflows/canister-tests.yml
nmattia: Gotcha, thought so, just checking!

PR #953, opened 2022-10-13
Allow building the archive using the docker-build script

frederikrothenberger:
This allows to build the archive using docker the same way as II.


scripts/docker-build
```diff
@@ -70,31 +78,48 @@ function build() {
...
+        --archive)
+            CANISTER+=("archive")
+            shift
+            ;;
         *)
             echo "ERROR: unknown argument $1"
             usage
             echo
-            echo "Use 'release --help' for more information."
+            echo "Use 'docker-build --help' for more information"
```
nmattia, line 111:
```suggestion
            echo "Use '$0 --help' for more information"
```

might as well prevent future copy pasta

scripts/docker-build
```diff
@@ -70,31 +78,48 @@ function build() {
...
             exit 1
             ;;
     esac
 done
 
-build
+# build everything by default
```
nmattia, line 117: this only builds II though, right?

frederikrothenberger: Ah, yes. I have also fixed the same error in the `build` script.

scripts/docker-build
```diff
@@ -70,31 +78,48 @@ function build() {
...
-    cp "$tmp_outdir/internet_identity.wasm" .
+    cp "$tmp_outdir/$canister.wasm" .
 
     echo "Removing $tmp_outdir"
     rm -rf "$tmp_outdir"
 }
 
 # ARGUMENT PARSING
 
+CANISTER=()
```
nmattia, line 89: nit: this should probably be `CANISTERS`

nmattia (approved):

PR #954, opened 2022-10-14
Integrate internet identity with the archive canister

frederikrothenberger:
The following features are implemented:
- II spawns an archive if the correct wasm is uploaded to it
- II upgrades existing archives if the installed module is outdated
- once an archive has been spawned, II will record anchor operations

Spawning an archive is optional. The archive feature is behind a feature flag and _not_ enabled in the production build.


src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,179 @@
...
+        if archive_state == ArchiveState::NotCreated {
+            // lock archive creation because of async operation
+            persistent_state.archive_info.state = ArchiveState::CreationInProgress;
+        }
+        archive_state
+    });
+
+    match archive_state {
+        ArchiveState::NotCreated => create_and_install_archive(wasm.into_vec()).await,
+        ArchiveState::CreationInProgress => DeployArchiveResult::CreationInProgress,
```
nmattia, line 33: just confirming my assumption about semantics here: if `CreationInProgress` is returned, it means that _some other call_ triggered a deploy, and there is no particular reason why the wasm module being deployed is the same as `wasm`, correct?

src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+}
+
+async fn create_and_install_archive(wasm: Vec<u8>) -> DeployArchiveResult {
+    match create_archive().await {
+        Err(e) => {
+            state::persistent_state_mut(|persistent_state| {
+                // unlock archive creation again
+                persistent_state.archive_info.state = ArchiveState::NotCreated
```
nmattia, line 48: I'm a bit uneasy about the control flow here; if for any reason (bug, canister trapping, etc) we fail before this line then we'll forever be stuck in `CreationInProgress`, right? Also, having the state being set somewhere (in `deploy_archive` and reverted elsewhere (here) is not particularly clear and may lead to some nasty bugs down the road

src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,179 @@
...
+            match install_archive(archive_canister_id, wasm).await {
+                Ok(()) => DeployArchiveResult::Success(archive_canister_id),
+                Err(err) => UpgradeFailed(err),
+            }
+        }
+    }
+}
+
+async fn create_archive() -> CallResult<ArchiveData> {
+    let (result,) = create_canister(CreateCanisterArgument { settings: None }).await?;
```
nmattia, line 67: Might have missed something, but it looks like we're creating the canister before even checking the hash. Does that mean that an II canister with no hash specified will still be creating an archive canister if anyone hits the `deploy_archive` endpoint?

src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,179 @@
...
+            state::persistent_state_mut(|persistent_state| {
+                // unlock archive creation again
+                persistent_state.archive_info.state = ArchiveState::NotCreated
+            });
+            return CreationFailed(format!("failed to create archive: {:?}", e));
+        }
+        Ok(data) => {
+            let archive_canister_id = data.archive_canister;
+            state::persistent_state_mut(|persistent_state| {
+                // safe archive info permanently
```
nmattia, line 55:
```suggestion
                // save archive info permanently
```

src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,179 @@
...
+    let () = notify(
+        archive_data.archive_canister,
+        "write_entry",
+        (anchor, timestamp, encoded_entry),
+    )
+    .expect("failed to send archive entry notification");
+    state::increment_archive_seq_nr();
+}
+
+fn verify_wasm_hash(wasm_module: &Vec<u8>) -> Result<[u8; 32], String> {
```
nmattia, line 138: out of curiosity, did you check the computation cost of this function and if spamming `deploy_archive` would have a noticeable impact on II (DoS)?

src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,179 @@
...
+    let archive_data = state::archive_data(); // traps if archive is not available
+    let timestamp = time();
+    let entry = Entry {
+        anchor,
+        operation,
+        timestamp,
+        caller,
+        sequence_number: archive_data.sequence_number,
+    };
+    let encoded_entry = candid::encode_one(entry).expect("failed to encode archive entry");
```
nmattia, line 125: out of curiosity: I'm really surprised that _encoding_ may fail. Do you know why that might be the case?

src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,179 @@
...
+        .unwrap_or(false)
+    {
+        // Don't do anything further if the archive canister has the given module already installed
+        return Ok(());
+    }
+
+    let mode = match module_hash {
+        None => Install,
+        Some(_) => Upgrade,
+    };
```
nmattia, line 96: As far as I can tell this creates an `Option<bool>` with the following semantics (I've renamed `module_hash` to `current_hash` and `expected_hash` to `new_hash` for clarity):

* if `current_hash` is not set, return `None` -> Install
* if `current_hash` is the same as `new_hash`, then return `Some(true)` -> Upgrade
* if `current_hash` is _not_ the same as `new_hash`, then return `Some(false)` -> Upgrade


Couple questions:
* Why even upgrade if the hash is the same? is that in case of new arguments?
* Do you think explicitly matching on `current_hash` (`module_hash`) might be simpler to read/understand?

src/archive/tests/tests.rs
```diff
@@ -629,6 +632,96 @@ mod metrics_tests {
...
+        const TIMESTAMP: Timestamp = 1620328630000000000;
+        const ANCHOR: UserNumber = 10_000;
+        let env = StateMachine::new();
+        let canister_id = install_archive_canister(&env, EMPTY_WASM.clone());
+
+        restore_compressed_stable_memory(&env, canister_id, "stable_memory/archive_v1.bin.gz");
+        upgrade_archive_canister(&env, canister_id, ARCHIVE_WASM.clone());
+
+        let entries = api::get_entries(&env, canister_id, None, None).unwrap();
+        assert_eq!(entries.entries.len(), 4);
```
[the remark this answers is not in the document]
frederikrothenberger: This is due the underlying memory manager that is used in the archive. For efficiency reasons, the memory manager allocates virtual memory in buckets of 1000 wasm pages (64MB). And the archive uses 3 virtual memory objects. And then there is some overhead to all the data structures and some config memory as well, so we end up with an uncompressed stable memory size of about ~200MB. So compressing that down to 192KB is quite a high compression ratio. 😉

src/internet_identity/internet_identity.did
```diff
@@ -122,10 +122,22 @@ type GetDelegationResponse = variant {
 type InternetIdentityStats = record {
   users_registered: nat64;
   assigned_user_number_range: record { nat64; nat64; };
+  archive_info: ArchiveInfo;
+};
+
+// Information about the archive.
+type ArchiveInfo = record {
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, and both are valid states:
* `archive_canister: none, expected_wasm_hash: some`: the archive has not yet been deployed, but it could be, by calling `deploy_archive` with a wasm module matching the hash.
* `archive_canister: some, expected_wasm_hash: none`: although currently unsupported (due to the way init arguments are set up) this would just mean that there is an archive, but we do not allow any changes to it (thus there is no valid wasm hash for redeployment).

src/internet_identity/internet_identity.did
```diff
@@ -122,10 +122,22 @@ type GetDelegationResponse = variant {
...
+    // Expected hash of the archive wasm module. II will accept wasm modules hashing to this hash in the canister call
+    // deploy_archive. The actual installed archive wasm hash might be different if the archive has not yet been upgraded.
+    // When empty, deployment of the archive is disabled and II will not accept any wasm module.
+    expected_wasm_hash: opt blob;
 };
 
 type InternetIdentityInit = record {
-  assigned_user_number_range : record { nat64; nat64; };
+  assigned_user_number_range : opt record { nat64; nat64; };
+  archive_module_hash : opt blob;
```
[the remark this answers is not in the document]
frederikrothenberger: Then we just keep the previous configuration. Supplying `None` just has the semantics of "I don't want to change things".

src/internet_identity/internet_identity.did
```diff
@@ -145,6 +157,18 @@ type IdentityAnchorInfo = record {
     device_registration: opt DeviceRegistrationInfo;
 };
 
+type DeployArchiveResult = variant {
+  // The archive was deployed successfully and the supplied wasm module has been installed. The principal of the archive
+  // canister is returned.
+  success: principal;
```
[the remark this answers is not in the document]
frederikrothenberger: This is because you cannot create the archive but not deploy it. So it is either:
* create and deploy
* just deploy

Thus, having a single success result at the end of all actions made sense to me...

src/internet_identity/internet_identity.did
```diff
@@ -145,6 +157,18 @@ type IdentityAnchorInfo = record {
...
+  // Initial archive creation is already in progress.
+  creation_in_progress;
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, you can poll `stats`, which will return the principal of the archive once creation is complete.

src/internet_identity/src/archive.rs
frederikrothenberger: > just confirming my assumption about semantics here: if `CreationInProgress` is returned, it means that _some other call_ triggered a deploy, and there is no particular reason why the wasm module being deployed is the same as `wasm`, correct?

Yes, exactly. Also note, that the lock on archive operation is during archive creation _only_, to avoid creating multiple archive canisters due to race conditions.

But if you were to upgrade II in rapid succession to change the `wasm` hash for the archive and deploy then deploy with different hashes, the order in which the deploy messages get executed is up to the platform. Although practically, I don't think you can upgrade II fast enough, given the archive and II are on the same subnet, so that the deploy messages could cross and get reordered.

src/internet_identity/src/archive.rs
frederikrothenberger: Hm, I agree. Let me reorganize things to make it clearer. Also, there is an inherent risk of trapping during canister creation because we need to send a message which might fail for a whole host of reasons. So I'll add a timestamp to the `CreationInProgress` lock, which will unlock new archive canister creation after some reasonably long time (24 hours?).

src/internet_identity/src/archive.rs
frederikrothenberger: Yes, that is true. I will change that.

src/internet_identity/src/archive.rs
frederikrothenberger: As explained by @roman-kashitsyn:
> All serde-based encoders can fail. One reason is that serde tries to serialize values under a mutex, and grabbing a mutex might fail. Not a great design decision IMO; serialization should be infallible.

src/internet_identity/src/archive.rs
frederikrothenberger: The `if` clause above checks the hash and exits early if `current_hash` is the same as `new_hash`. But I agree, I will reshuffle things to make it more clear. 👍

src/internet_identity/src/archive.rs
frederikrothenberger: So I did some quick profiling and indeed hashing takes quite a few cycles:
* Hashing the actual archive.wasm: 72'438'283 cycles
* Submit bogus payload close to message size limit: 243'362'366 cycles

For comparison:
* `lookup`: 1'824'387 cycles
* `prepare_delegation`: 2'198'374 cycles

So roughly, one bogus `deploy_archive` call is about 100 times more expensive than an authentication process. We could limit payload size to not check the hash if it is outside the expected range which would make it only 35 times more expensive.

frederikrothenberger: I've talked to @roman-kashitsyn about this: I'll first check the installed `wasm` module and if it is already matching the configured one, I'll reject the payload right away, as there is nothing to do anyway. This will make it super cheap, once the archive has been deployed.

frederikrothenberger: Ok, I have now refactored the code as outlined above and the results are as follows:
* After deploying the archive and having cached the archive status a call to `deploy_archive` uses now `8'747'312` cycles of which about >99%(!) can be attributed to just decoding the incoming message.

We could further optimize the cycles usage by not decoding the argument, but I would only do this if it indeed turns out to be a problem since it requires more low level argument handling.

src/internet_identity/src/anchor_management.rs
```diff
@@ -127,23 +146,32 @@ pub async fn update(user_number: UserNumber, device_key: DeviceKey, device_data:
...
 
+    let caller = caller(); // caller is only available before await
     state::ensure_salt_set().await;
     delegation::prune_expired_signatures();
 
-    mutate_device_or_trap(&mut entries, device_key, None);
+    let operation = mutate_device_or_trap(&mut entries, device_key, None);
     write_anchor_data(user_number, entries);
+
+    if state::archive_ready() {
```
nmattia, line 172: this pattern is recurrent, why not call `archive_operation` and let it decide what to do? this will simplify the code, separate concerns, and reduce the risk of calling `archive_operation` if the archive is not ready

src/internet_identity/src/archive.rs
```diff
@@ -0,0 +1,247 @@
...
+                    // we already have an archive with the expected module and don't need to do anything
+                    return Success(archive_data.archive_canister);
+                }
+                Some(_) => (archive_data.archive_canister, Upgrade, false),
+            }
+        }
+    };
+
+    // creating the archive verifies the wasm hash as well
+    if !wasm_hash_verified {
```
nmattia, line 66: This blind boolean is asking for trouble; why do we verify the hash in two difference places? Also, can we use types to signal whether the hash was verified? something like `newtype` would do nicely

src/internet_identity/src/state.rs
```diff
@@ -229,6 +313,49 @@ pub fn anchor_devices(anchor: UserNumber) -> Vec<DeviceDataInternal> {
...
 }
 
+pub fn archive_state() -> ArchiveState {
+    STATE.with(|s| s.persistent_state.borrow().archive_info.state.clone())
+}
+
+pub fn archive_ready() -> bool {
+    STATE.with(|s| match s.persistent_state.borrow().archive_info.state {
+        ArchiveState::Created(_) => true,
+        _ => false,
```
nmattia, line 323: nit: slightly dangerous if we ever add e.g. `ArchiveState::Installed`

src/internet_identity/src/storage.rs
```diff
@@ -324,6 +336,69 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned, M: Memory> Storage<T,
...
         Ok(record_number)
     }
+
+    /// Returns the address of the first byte not yet allocated to a user.
+    /// This address exists even if the max user number has been reached, because there is a memory
+    /// reserve at the end of stable memory.
+    fn unused_memory_start(&self) -> u64 {
+        let record_number = self.header.num_users as u64;
+        RESERVED_HEADER_BYTES + record_number * self.header.entry_size as u64
+    }
```
nmattia, line 346: why don't we simply always return the value after the 8 gigs allocated to users? In production we have many more users than on test canisters; by always assuming the canister may be full we reduce the risk of issues in production that we can't trigger with test canisters with fewer users

src/internet_identity/src/storage.rs
```diff
@@ -256,20 +271,17 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned, M: Memory> Storage<T,
     }
 
     /// Make sure all the required metadata is recorded to stable memory.
-    pub fn flush(&self) {
-        if self.memory.size() < 1 {
```
nmattia, line 260: I'm not very familiar with this code; why don't we gros the memory anymore?

EDIT: I see, we're now using a writer. Why does the archive require that?

src/internet_identity/internet_identity.did
```diff
@@ -122,10 +122,24 @@ type GetDelegationResponse = variant {
...
+    // deploy_archive. The actual installed archive wasm hash might be different if the archive has not yet been upgraded.
+    // When empty, deployment of the archive is disabled and II will not accept any wasm module.
+    expected_wasm_hash: opt blob;
 };
 
 type InternetIdentityInit = record {
-  assigned_user_number_range : record { nat64; nat64; };
+  assigned_user_number_range : opt record { nat64; nat64; };
+  archive_module_hash : opt blob;
+  canister_creation_cycles_cost : opt nat64;
```
nmattia, line 142: Can you add a line or two about how this is supposed to be used? There's some rationale in `create_canister` but I wouldn't know how to use it

nmattia (approved):

PR #956, opened 2022-10-17
Bake .select() in anchor input

nmattia:
This makes the `anchorInput` component call `.select()` on itself when mounted in the DOM, so that callers do not have to do it themselves.


src/frontend/src/components/anchorInput.ts
```diff
@@ -140,3 +149,38 @@ const inputFilter = (inputFilter: (c: string) => boolean, onBad: () => void) =>
...
+      // disconnect is to avoid leaking the observer to keep observing the
+      // DOM if the element was created but never mounted.
+      //
+      // Note: it would be much easier to use the "DOMNodeInsertedIntoDocument"
+      // event on the element itself, but the API is deprecated and Firefox
+      // does not support it.
+
+      const observer = new MutationObserver(() => {
+        try {
+          callback(e);
```
peterpeterparker, line 178: you can maybe iterate on the nodes that are added and check if the Element that is passed is one of those

something like:

```
const observer = new MutationObserver((mutations: MutationRecord[]) => {
    
     const addedNodes: Node[] = mutations
    .filter(({addedNodes}: MutationRecord) => addedNodes?.length > 0)
    .reduce((acc: Node[], {addedNodes}: MutationRecord) => [...acc, ...Array.from(addedNodes)], []);

   const nodeAdded: boolean = addedNodes.find((addedNode: Node) => addedNode.isEqualNode(e)) === undefined);
   
   if (!nodeAdded) {
     return;
   }
   
   try {
     ....
     
```

nmattia, line 178: TODO: ensure that element was added to DOM before calling `callback`

peterpeterparker: assuming `e` is a direct reference - not a clone - to the `input` and that the reference is preserved there is maybe two option.

1. check that the nodes are added contains the `e`. this works only if above statement is correct - i.e. `e` needs to be in the dom

```
const observer = new MutationObserver((mutations: MutationRecord[]) => {
    const addedNodes: Node[] = mutations.reduce(
      (acc: Node[], {addedNodes}: MutationRecord) => [...acc, ...Array.from(addedNodes)],
      []
    );
    
    const isElement: boolean = addedNodes.find((addedNode: Node) => addedNode.isEqualNode(removedNode)) !== undefined;
    
    // Some other nodes were added in the DOM
    if (!isElement) {
       return;
    }
    
    try {
       ...
```

2. if indeed we have a ref that is preserved, maybe even simplier

 ```
 const observer = new MutationObserver(() => {
    if (!e.isConnected) {
       // The element is not yet parsed in the DOM
       return;
    }
    
    try {
      ...
```

nmattia: Did the latter! Thanks for the suggestion

peterpeterparker (approved):
LGTM, thx

PR #957, opened 2022-10-17
Work around empty arrays for bash <4.4

nmattia:
This adds a workaround for accessing an empty array in bash versions older than 4.4, as well as removing the bash upgrade on CI.


frederikrothenberger (approved):
Thanks!

PR #958, opened 2022-10-18
Update test app dependencies and fix canister id bug

frederikrothenberger:
This PR updates the rust dependencies to be in line with the rest of the Internet Identity repository. Additionally it fixes a small bug related to the canisterId replacement in the `index.html`.

### DIFF
--- a/demos/test-app/lib.rs
+++ b/demos/test-app/lib.rs
@@ -216,7 +216,7 @@ lazy_static! {
 
         // the string we are replacing here is inserted by webpack during the front-end build
         let index_html = index_html.replace(
-            r#"<script defer="defer" src="bundle.js"></script>"#,
+            r#"<script defer src="bundle.js"></script>"#,
             &format!(r#"<script>var canisterId = '{canister_id}';</script><script defer="defer" src="bundle.js"></script>"#).to_string()
         );
         index_html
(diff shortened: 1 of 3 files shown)


demos/test-app/lib.rs
```diff
@@ -216,7 +216,7 @@ lazy_static! {
 
         // the string we are replacing here is inserted by webpack during the front-end build
         let index_html = index_html.replace(
-            r#"<script defer="defer" src="bundle.js"></script>"#,
```
nmattia, line 219: I'm a little confused, how was the js injected?

frederikrothenberger: I'm confused now, too. Isn't the comment above this line exactly answering the question?
`// the string we are replacing here is inserted by webpack during the front-end build`

frederikrothenberger: The diff is there, because apparently webpack changed behavior and is now injecting something different (i.e. just `defer` instead of `defer="defer"`).

nmattia: > Isn't the comment above this line exactly answering the question?

It is... kind of 🙃 but I was wondering _how_ it was being injected if the string didn't match. Don't we use the test app in CI? If so, how does it work?

frederikrothenberger: Ok, I took another look. It works for the selenium tests because the value is not used for anything critical. This being broken just results in a silly default value of `[object HTMLInputElement]` in the canisterId input box.

But our test always fill in the canisterId explicitly so they are not affected by the default.

PR #959, opened 2022-10-18
Fix misleading information about II derived principals

frederikrothenberger:
This PR fixes the issue of the spec not referring to the DER encoded public key when describing principal derivation.


Dfinity-Bjoern (approved):
Thanks for fixing this, @frederikrothenberger

PR #964, opened 2022-10-24
Layer GitHub Actions cache

nmattia:
This splits the GHA cache into three layers:
* `~/.cargo`
* external dependencies
* canister code

This means that a commit that only modifies the JS code won't cause a rebuild at all, and a commit modifiying the `archive` canister won't cause a dependency rebuild.

More layers can be added (e.g. `internet_identity/`) if necessary.

Note: macOS is still rebuilding a little too much because `~/.cargo` is not cached properly, see https://github.com/actions/cache/issues/966.


.github/workflows/canister-tests.yml
```diff
@@ -195,26 +195,120 @@ jobs:
...
+      - name: Retrieve commit information
+        id: git_info
+        run: |
+          # Last time dependencies or toolchain were updated; used for ~/cargo and dependencies
+          commit_deps=$(git log -1 --pretty=format:%H -- Cargo.* src/*/Cargo.toml rust-toolchain.toml)
+          echo "::set-output name=commit_deps::$commit_deps"
+          echo "deps: $commit_deps"
+
+          # Last time the canister code was modified
+          commit_canister=$(git log -1 --pretty=format:%H -- src/[^frontend]*)
```
frederikrothenberger, line 212: I think `[...]` makes it character based? At least in my quick test, this command included commits that touched stuff in `src/frontend` _only_.

Shouldn't this be `git log  -- . 'src/**' ':^src/frontend/**'`?

Also, if you really only want canister changes you could also additionally exclude `canister_tests` and `src/*/tests/**` (integration tests), i.e. `git log  -- . 'src/**' ':^src/frontend/**' ':^src/canister_tests/**' ':^src/*/tests/**'`

.github/workflows/canister-tests.yml
```diff
@@ -195,26 +195,120 @@ jobs:
...
+      - name: Cargo cache (build)
+        run: cargo fetch --locked
+
+      - name: Cargo cache (save to cache)
+        if: steps.cache-dot-cargo.outputs.cache-hit != 'true'
+        run: rsync -av "$HOME/.cargo/" "/tmp/dot-cargo/"
+
+      # Caching external dependencies
+
+      - uses: actions/cache@v3
```
frederikrothenberger, line 262: This cache action is also dependent on `steps.git_info.outputs.commit_deps`, so in theory it could be merged with `cache-dot-cargo` and using multiple folders, right?

.github/workflows/canister-tests.yml
```diff
@@ -195,26 +195,120 @@ jobs:
...
+
+      - name: Canister code (save to cache)
+        if: steps.cache-canister.outputs.cache-hit != 'true'
+        run: rsync -av --compare-dest="/tmp/target-deps/" "$PWD/target/" "/tmp/target-canister/"
+
+      # Build the tests
+      - name: Build
+        run: |
+          git checkout ${{ steps.git_info.outputs.commit_now }}
+          cargo build --tests --release --frozen
```
frederikrothenberger, line 311: Why do we need to run this multiple times? I would assumed, that if we restored all the files correctly (including timestamps), we could just run `cargo test` at the end?

.github/workflows/canister-tests.yml
```diff
@@ -195,26 +195,120 @@ jobs:
...
 
+      # Figure out which commit modified what. Each commit defined here is used as a cache layer.
+      # We use those commit shas as cache keys. Since cargo uses timestamp-based comparison,
+      # each cache layer is retrieved and timestamped when the particular commit is checked out,
+      # so that checking out a later commit tells cargo which files are newer (since git will
+      # update the mtime of newer files)
+      - name: Retrieve commit information
+        id: git_info
+        run: |
+          # Last time dependencies or toolchain were updated; used for ~/cargo and dependencies
```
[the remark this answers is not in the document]
nmattia: Indeed; feel free to use ` ```suggestion` next time

.github/workflows/canister-tests.yml
```diff
@@ -195,26 +195,120 @@ jobs:
...
+          echo "deps: $commit_deps"
+
+          # Last time the canister code was modified
+          commit_canister=$(git log -1 --pretty=format:%H -- src/[^frontend]*)
+          echo "::set-output name=commit_canister::$commit_canister"
+          echo "canister: $commit_canister"
+
+          # This commit
+          commit_now=$(git log -1 --pretty=format:%H)
+          echo "::set-output name=commit_now::$commit_now"
```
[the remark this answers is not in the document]
nmattia: No, let's instead move everything at once to avoid mixing `set-output` and the env file

.github/workflows/canister-tests.yml
```diff
@@ -195,26 +195,120 @@ jobs:
...
         with:
-          # note on paths: we include the whole ~/.cargo/. (as opposed to just .cargo/git, etc.) because cargo stores crate
-          # data as .crates(2?).{json,toml} without which the whole cache is invalidated. We may accumulate unnecessary
-          # stuff but the cache is anyway invalidated on every toolchain update.
-          # We also include all Cargo tomls and the lockfile
-          path: |
-            ~/.cargo
-            target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock', 'Cargo.toml', 'src/**/Cargo.toml', 'rust-toolchain.toml') }}-3
+          path: /tmp/dot-cargo
```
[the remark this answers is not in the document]
nmattia: Caches are only uploaded at the end of the workflow run; I want to make sure we only upload what's downloaded at that checkout but happy to do it differently if there are good reasons?

.github/workflows/canister-tests.yml
nmattia: True! However I've separated things as much as possible for legibility and also in case we make those steps dynamic at some point (based on a list of directories/cache layers for instance)

.github/workflows/canister-tests.yml
nmattia: We could; though just like we have currently it's clearer to split build and tests so that we can quickly see what takes time

.github/workflows/canister-tests.yml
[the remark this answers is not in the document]
frederikrothenberger: No, keep it. I was just curious. Maybe it would be good to have a comment that explains the reasoning there.

.github/workflows/canister-tests.yml
nmattia: > I think [...] makes it character based?

You're right! Looks like globs are a little more like regexes than I thought.

> Shouldn't this be `git log -- . 'src/**' ':^src/frontend/**'`?

Looks like this works, though I don't quite understand what it does... also I didn't know git itself supports globbing! What does `:` mean here?

frederikrothenberger: Its part of the git `pathspec` syntax: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-exclude

.github/workflows/canister-tests.yml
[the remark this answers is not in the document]
nmattia: Added a comment!

.github/workflows/canister-tests.yml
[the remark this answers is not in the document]
nmattia: Updated

frederikrothenberger (approved):
LGTM, thanks!

PR #965, opened 2022-10-24
Add localStorage tests

nmattia:
This adds a few tests for storing/reading "user numbers" (anchors) as well as tests checking the serialized form of user numbers in localStorage.

### DIFF
--- a/src/frontend/src/utils/userNumber.test.ts
+++ b/src/frontend/src/utils/userNumber.test.ts
@@ -77,5 +77,3 @@ function readLocalStorage(): LocalStorage {
 
   return ls;
 }
-
-function populateLocalStorage(content: any) {}
--- a/src/frontend/src/utils/userNumber.test.ts
+++ b/src/frontend/src/utils/userNumber.test.ts
@@ -1,0 +1,81 @@
+import { getUserNumber, setUserNumber } from "./userNumber";
+
+testLocalStorage("user number can be set", () => {
+  expect(getUserNumber()).toBe(undefined);
+  setUserNumber(BigInt(123456));
+  expect(getUserNumber()).toBe(BigInt(123456));
+});
+
+testLocalStorage("user number can be cleared", () => {
+  setUserNumber(BigInt(123456));
+  expect(getUserNumber()).toBe(BigInt(123456));
+  setUserNumber(undefined);
+  expect(getUserNumber()).toBe(undefined);
+});
+
+testLocalStorage(
+  "user number localStorage format",
+  () => {
+    setUserNumber(BigInt(123456));
+  },
+  { after: { userNumber: "123456" } }
+);
+
+testLocalStorage(
+  "user number read from known localStorage format",
+  () => {
+    expect(getUserNumber()).toStrictEqual(BigInt(123456));
+  },
+  { before: { userNumber: 123456 } }
+);
+
+/** Run a test that makes use of localStorage. Local storage is always cleared after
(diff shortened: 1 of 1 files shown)


src/frontend/src/utils/userNumber.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    const key = localStorage.key(i);
+    if (key !== null) {
+      ls[key] = localStorage.getItem(key);
+    }
+  }
+
+  return ls;
+}
+
+function populateLocalStorage(content: any) {}
```
peterpeterparker, line 81: still need this?

src/frontend/src/utils/userNumber.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/// Set the entire localStorage (does NOT clear localStorage)
+function setLocalStorage(ls: LocalStorage) {
+  for (const key in ls) {
+    localStorage.setItem(key, JSON.stringify(ls[key]));
+  }
+}
+
+function readLocalStorage(): LocalStorage {
+  let ls: LocalStorage = {};
```
peterpeterparker, line 70: was curious if it can be inlined, so just in case

```
Object.entries(localStorage).map(([key, value]) => ({[key]: value}))
```

src/frontend/src/utils/userNumber.test.ts
```diff
@@ -0,0 +1,81 @@
+import { getUserNumber, setUserNumber } from "./userNumber";
+
+testLocalStorage("user number can be set", () => {
+  expect(getUserNumber()).toBe(undefined);
+  setUserNumber(BigInt(123456));
```
peterpeterparker, line 5: `BigInt(123456)` is repeated few don't know if you rather like to have a common const value for your test?

peterpeterparker (approved):
LGTM, just few super minor comments because you explicitely asked for a review 😉

src/frontend/src/utils/userNumber.test.ts
nmattia: Yeah was thinking about this, but I'll wait until I have to repeat it a few more times :)

src/frontend/src/utils/userNumber.test.ts
nmattia: Ah I don't! Good catch

src/frontend/src/utils/userNumber.test.ts
nmattia: Also, doesn't this map to an array of `{key: value}`s, i.e. `[{key1: value1}, {key2: value2}, ...]`?

nmattia: either way, I'm curious now :) but it's less clear, so I'll probably stick to the current solution

peterpeterparker: > The Object.entries() method returns an array of a given object's own enumerable string-keyed property [key, value] pairs. [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries)

`getItem` and `setItem` are no properties but functions.

> doesn't this map to an array

You are right. It would need a `reduce`

```
Object.entries(localStorage).reduce((acc, [key, value]) => ({...acc, [key]: value}), {})
```

Seems to work but yes current solution is fine, was curiosity too from my side.

Capture d’écran 2022-10-25 à 10 29 03

PR #966, opened 2022-10-25
Remove unused values in dfx.json

nmattia:
These were added for testing or workaround and are no longer used.


bitdivine (approved):
Note: The entire networks block can be removed.  `dfx start` is no longer attached to a project, so it doesn't really make sense to have local here.  You can keep it; it is expected to be supported until early 2023 IIRC.

PR #968, opened 2022-10-26
Refactor storage module to be testable

frederikrothenberger:
This PR makes `Storage` testable by extending it to use an abstractions of stable memory. It also adds a unit test suite for `Storage` and replaces low level memory interactions with implementations of the `stable_structures` crate.
More documentation about the memory layout is included as well.


src/internet_identity/src/storage.rs
```diff
@@ -1,31 +1,83 @@
+//! This module implements all the stable memory interactions of Internet Identity.
+//! It uses the [Reader] and [Writer] implementations of the `stable_structures` crate.
+//!
+//! ## Stable Memory Layout
```
nmattia, line 4: This is great!

src/internet_identity/src/storage/tests.rs
```diff
@@ -0,0 +1,175 @@
+use crate::storage::{Header, StorageError};
```
nmattia, line 1: (about the whole module) if I understand correctly, these tests make use of a dummy vector-based implementation for stable memory, correct? Also, those tests are compiled and run as x86/aarch, right? Would there be any reason why memory/serializing would behave different when compiled to Wasm?

src/internet_identity/src/storage/tests.rs
```diff
@@ -0,0 +1,175 @@
...
+    memory.write(512, &(buf.len() as u16).to_le_bytes());
+    memory.write(514, &buf);
+
+    let read_from_storage = storage.read(123).unwrap();
+    assert_eq!(read_from_storage, device_vec);
+}
+
+#[test]
+fn should_deserialize_subsequent_record_at_expected_memory_location() {
+    const EXPECTED_RECORD_OFFSET: u64 = 204_800;
```
nmattia, line 131: Where does this value come from?

src/internet_identity/src/storage/tests.rs
```diff
@@ -0,0 +1,175 @@
...
+    let memory = VectorMemory::default();
+    memory.grow(4);
```
nmattia, line 133: I'm not super familiar with the implementation. What happens here?

src/internet_identity/src/storage.rs
```diff
@@ -106,7 +159,7 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned> Storage<T> {
                 &mut header as *mut _ as *mut u8,
                 std::mem::size_of::<Header>(),
             );
-            stable64_read(0, slice);
```
nmattia, line 109: out of curiosity: is the address size abstracted away now or do we pass that in as configuration somewhere?

src/internet_identity/src/storage.rs
```diff
@@ -140,46 +194,46 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned> Storage<T> {
...
-        if pages > current_size {
-            let pages_to_grow = pages - current_size;
-            let result = stable64_grow(pages - current_size);
-            if result.is_err() {
-                trap(&format!(
-                    "failed to grow stable memory by {} pages",
-                    pages_to_grow
-                ))
-            }
-        }
```
nmattia, line 165: Guess all the growing business is now hidden away in `Memory`, correct?

src/internet_identity/src/storage.rs
```diff
@@ -140,46 +194,46 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned> Storage<T> {
...
-        stable64_write(stable_offset, &(buf.len() as u16).to_le_bytes());
-        stable64_write(stable_offset + std::mem::size_of::<u16>() as u64, &buf);
+        let mut writer = BufferedWriter::new(
+            self.header.entry_size as usize,
+            Writer::new(&mut self.memory, stable_offset),
```
nmattia, line 210: How come we don't write directly to the `Writer` here but wrap it in a `BufferedWriter`?

src/internet_identity/src/storage.rs
```diff
@@ -140,46 +194,46 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned> Storage<T> {
...
+        );
+        writer
+            .write(&(buf.len() as u16).to_le_bytes())
+            .expect("memory write failed");
+        writer.write(&buf).expect("memory write failed");
```
nmattia, line 215: Just to make sure I understand, this will move the pointer whenever data is written right?

src/internet_identity/tests/tests.rs
```diff
@@ -1896,18 +1896,20 @@ mod http_tests {
...
-        let (signature_count, _) = parse_metric(&metrics, "internet_identity_stable_memory_pages");
+        let (stable_memory_pages, _) =
+            parse_metric(&metrics, "internet_identity_stable_memory_pages");
         // empty II has some metadata in stable memory which requires at least one page
-        assert_eq!(signature_count, 1);
+        assert_eq!(stable_memory_pages, 1);
 
-        // a wasm page is 64kb and a single user takes up 2kb -> 32 users require a complete wasm page
-        for _ in 0..32 {
+        // a wasm page is 64kb and a single user takes up  2kb -> 33rd anchor record starts on the second wasm page
```
nmattia, line 1904:
```suggestion
        // a wasm page is 64kb and a single user takes up 2kb -> 33rd anchor record starts on the second wasm page
```

src/internet_identity/tests/tests.rs
```diff
@@ -1896,18 +1896,20 @@ mod http_tests {
...
+        for _ in 0..33 {
```
nmattia, line 1905: Can you explain why this had to be bumped? I'm guessing 32 anchors just fit in one page; why did the test pass before?

src/internet_identity/src/storage/tests.rs
frederikrothenberger: Yes, exactly. The only potential difference I see is that for Wasm we have a 32bit architecture and on x86/aarch we have 64bit. This means that the size of `usize` is 8 instead of 4 in the tests. But since the `Memory` trait uses `u64` exclusively, this should not be an issue.

src/internet_identity/src/storage/tests.rs
frederikrothenberger: It's just 100 times the anchor size of 2048 bytes. I'll add a comment.

src/internet_identity/src/storage/tests.rs
frederikrothenberger: This grows the stable memory to fit the address `512 + EXPECTED_RECORD_OFFSET` such that we don't get an out of bounds error on `memory.write`. Memory has to be allocated explicitly in 64kb chunks (called pages). And this address happens to lie in chunk 4.

src/internet_identity/tests/tests.rs
frederikrothenberger: The reason is, that by using stable structures `Writer`, we now only grow the memory on write, where beforehand we would always grow the memory to fit the whole 2048 bytes an anchor record might use (regardless of the actual size).

And because the anchor does not actually use the whole 2048 bytes, this caused it to still fit into the previous memory page. So we had to bump it in order to cause a write to page 2.

src/internet_identity/src/storage.rs
frederikrothenberger: `stable_structures` supports 64bit addresses only. Which has the nice benefit that now one cannot get it wrong accidentally. 🙂

src/internet_identity/src/storage.rs
frederikrothenberger: Stable memory writes are expensive. By using `BufferedWriter` we have only one actual write to stable memory.

src/internet_identity/src/storage.rs
frederikrothenberger: Yes, exactly. This is the benefit of using the `Writer`. No more fiddling with incremental offsets.

src/internet_identity/src/storage.rs
frederikrothenberger: It's hidden in `Writer`, but yes.

src/internet_identity/src/storage.rs
nmattia: Ah, I see! Can you add a comment about this?

src/internet_identity/src/storage/tests.rs
```diff
@@ -0,0 +1,175 @@
...
+    let mut buf = vec![0; HEADER_SIZE];
+    memory.read(0, &mut buf);
+    assert_eq!(buf, hex::decode("49494301000000000100000000000000020000000000000000080505050505050505050505050505050505050505050505050505050505050505").unwrap());
+}
+
+#[test]
+fn should_recover_header_from_memory() {
+    let memory = VectorMemory::default();
+    memory.grow(1);
+    memory.write(0, &hex::decode("494943010500000040e2010000000000f1fb09000000000000084343434343434343434343434343434343434343434343434343434343434343").unwrap());
```
nmattia, line 38: I'm guessing that somewhere in there there's the storage range 123456 - 654321. Did you run the test and then copy out the hex data? Lots of these tests seem to be testing a particular encoding of the header, and sometimes (like here) a particular field value. What do these tests try to prevent in terms of bugs?

nmattia (approved):
Nice; how are we doing in terms of test duration? Guessing this doesn't take much time since it's all native code?

src/internet_identity/src/storage/tests.rs
frederikrothenberger: Yes. These tests detect if the memory location shifts or the data layout changes, which are both things that must not happen by accident.

frederikrothenberger:
>how are we doing in terms of test duration?

` test result: ok. 26 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s`

I don't think we have to worry about that 😉

PR #969, opened 2022-10-27
Unify login pages

nmattia:
This replaces the two different landing pages we have currently with a
single one. What is displayed depending on whether this is a first time
or returning user.


meodai (approved):

src/frontend/src/components/anchorInput.ts
```diff
@@ -86,21 +92,21 @@ export const mkAnchorInput = (props: {
 
   // Select the input when the component is rendered (and value is empty)
   const selectInput = (elem: Element): void => {
-    if (elem instanceof HTMLInputElement && elem.value === "") {
+    if (elem instanceof HTMLInputElement) {
```
peterpeterparker, line 95: if `&& elem.value === ""` has been on purpose deleted, then the comment should probably be updated

src/frontend/src/components/anchorInput.ts
```diff
@@ -86,21 +92,21 @@ export const mkAnchorInput = (props: {
...
     }
   };
 
-  const template = html` <div class="l-stack c-input--anchor">
+  const template = html` <div class="c-input--anchor">
     <label class="c-input--anchor__wrap" aria-label="Identity Anchor">
       <input
         ${ref(userNumberInput)}
-        ${mount(selectInput)}
+        ${focus ? mount(selectInput) : ""}
```
peterpeterparker, line 104: not sure with lit but in React or Stencil when nothing need to be rendered commonly `undefined` (rather than `""`) is used. 

would `${focus ? mount(selectInput) : ""}` work?

or even more fancy `${focus && mount(selectInput)}`?

nmattia: > when nothing need to be rendered commonly undefined (rather than "") is used.

Mh, good question

> would `${focus ? mount(selectInput) : ""}` work?

That's what's there now right? 

> more fancy `${focus && mount(selectInput)}`?

Linter will definitely complain about that (mixing booleans and falsy/truthy)


---

Actually looks like `undefined`, `null` and `""` are all ok: https://lit.dev/docs/v1/lit-html/writing-templates/#rendering-nothing

> Sometimes, you may want to render nothing at all. The values undefined, null and the empty string ('') in a text binding all render an empty text node.

src/frontend/src/components/anchorInput.ts
nmattia: Oh good catch, thanks

src/frontend/src/flows/login/index.ts
```diff
@@ -15,7 +38,38 @@ export const login = async (
...
   try {
-    const x = await tryLogin(connection);
+    const userNumber = getUserNumber();
+    const clearCache = () => {
+      const message = `This will forget your anchor number but your credentials will still be stored. To clear your credentials please check your device's documentation on how to clear WebAuthn credentials.`;
+      const confirmed = confirm(message);
+      if (confirmed) {
+        window.localStorage.clear();
+        window.location.reload();
+      }
```
peterpeterparker, line 48: should we had a `return;` after `reload` just in case for an unknown reason reload would not work? most probably never gonna happen but just in case?

I guess not related to this refactoring

peterpeterparker (approved):
LGTM.

Note: I am understanding that the logic is refactored so did not went deep down.

PR #972, opened 2022-10-28
Move lit-html utils to module

nmattia:
This extracts `mount` and `withRef` into a `lit-html`-specific utils module.


peterpeterparker (approved):
LGTM

PR #975, opened 2022-10-31
Add anchor picker to showcase

nmattia:
This introduces a new component (an anchor _picker_) that allows picking one of many anchors. The picker is not yet integrated with any of the pages; this PR only adds the picker code, the tests, and an integration in the showcase.

Texts still have to be agreed on; current text is only placeholder.

https://user-images.githubusercontent.com/6930756/198989572-7bf70fba-66dd-4731-8684-88940e7fa541.mov


peterpeterparker (approved):
LGTM

PR #979, opened 2022-11-01
Shard browser tests

nmattia:
This shards our browser tests to speed up CI runs. Test setup is repeated for each shard but is small compared to actual test duration.

The monolithic `selenium.test.ts` is split into multiple files, since jest will always assign an entire file to a shard.

This also adds a custom sequencer to make sure that two long tests do not end up on the same shard.


package.json
```diff
@@ -16,8 +16,8 @@
     "test:e2e-desktop": "SCREEN=desktop npm run test:e2e",
     "test:e2e-mobile": "SCREEN=mobile npm run test:e2e",
     "lint": "eslint --max-warnings 0 --cache --cache-location node_modules/.cache/eslint 'src/frontend/**/*.ts*'",
-    "format": "prettier --write src/frontend .eslintrc.json webpack.config.js jest.config.ts",
-    "format-check": "prettier --check src/frontend .eslintrc.json webpack.config.js jest.config.ts"
+    "format": "prettier --write src/frontend tsconfig.json .eslintrc.json webpack.config.js jest.config.ts",
```
peterpeterparker, line 19: looking forward to the day you gonna finally switch to `.` here 😉

src/frontend/custom-sequencer.ts
```diff
@@ -0,0 +1,29 @@
+const Sequencer = require("@jest/test-sequencer").default;
+
+/** A sequencer that ensures that ensures that the slowest tests
+ * all land on different shards. */
+class CustomSequencer extends Sequencer {
+  shard(tests, { shardIndex, shardCount }) {
```
peterpeterparker, line 6: not typed, I guess no need to?

src/frontend/src/test-e2e/alternativeOrigin.test.ts
```diff
@@ -0,0 +1,236 @@
...
+  await runInBrowser(async (browser: WebdriverIO.Browser) => {
+    const authenticatorId1 = await addVirtualAuthenticator(browser);
+    await browser.url(II_URL);
+    const userNumber = await FLOWS.registerNewIdentityWelcomeView(
+      DEVICE_NAME1,
+      browser
+    );
+    await FLOWS.addRecoveryMechanismSeedPhrase(browser);
+    const credentials = await getWebAuthnCredentials(browser, authenticatorId1);
+    expect(credentials).toHaveLength(1);
```
peterpeterparker, line 32: would it be worth to also check the content?

src/frontend/src/test-e2e/alternativeOrigin.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    );
+    expect(await errorView.getErrorDetail()).toEqual(
+      '"https://nice-name.com" is not listed in the list of allowed alternative origins. Allowed alternative origins:'
+    );
+  });
+}, 300_000);
+
+test("Should not issue delegation when derivationOrigin is malformed", async () => {
+  await runInBrowser(async (browser: WebdriverIO.Browser) => {
+    const authenticatorId1 = await addVirtualAuthenticator(browser);
```
peterpeterparker, line 65: it isn't the same block of code as above? I know just a test, just saying in case you would like to space some lines...

src/frontend/src/test-e2e/alternativeOrigin.test.ts
nmattia: ah this was just moved

src/frontend/src/test-e2e/alternativeOrigin.test.ts
nmattia: ah this was just moved

src/frontend/custom-sequencer.ts
nmattia: it's actually a pain to set up because it gets picked up in the actual app build so we need to ts-ignore it. Will need you help eventually to split the tsconfig...

src/frontend/custom-sequencer.ts
```diff
@@ -0,0 +1,29 @@
...
+    // Tests by order of increasing duration. Update by looking at test duration.
+    // Does not need to be exact.
+    const testFilesByDuration = [
```
peterpeterparker, line 9: would it be worth to build this array dynamically? List files in ordner x with NodeJS?

tsconfig.json
```diff
@@ -17,5 +17,10 @@
     "isolatedModules": true,
     "types": ["node", "jest", "webdriverio/async"]
   },
-  "exclude": ["src/**/*.test.ts", "src/frontend/generated/*"]
+  "exclude": [
+    "src/**/*.test.ts",
+    "src/frontend/src/test-e2e/*.ts",
+    "src/frontend/generated/*",
+    "src/frontend/custom-sequencer.ts"
```
peterpeterparker, line 24: to make more obvious that the custom-sequencer for test, would it be worth renaming it `test-custom-sequencer.ts`?

peterpeterparker (approved):
LGTM, minor comments. Would love to have sharding in NNS-dapp!

tsconfig.json
nmattia: Very good point. Will rename `jest-custom-sequencer.ts`!

src/frontend/custom-sequencer.ts
nmattia: Actually I don't need a list of all the files, I just need the files that take longer to run. What's important here is the order of the files.

PR #982, opened 2022-11-01
Refactor authenticate page

nmattia:
This refactors the `authenticate` page (`authenticate/index.ts`), disentangling the code (fewer params, promises, etc passed around) and making use of more idiomatic lit constructs (`@click`, etc).

This also renames `Manage Anchor` to `Home` to clarify the intent (redirect to `/` as opposed to just reloading the page; either way not directly managing an anchor).

This is all preparation work for the upcoming revamp of the landing page.


src/frontend/src/flows/authenticate/index.ts
```diff
@@ -22,33 +22,28 @@ import {
 import { unreachable } from "../../utils/utils";
 import { withRef } from "../../utils/lit-html";
 
-type PageElements = {
-  authorizeButton: Ref<HTMLButtonElement>;
-  registerButton: Ref<HTMLLinkElement>;
+type PageProps = {
+  origin: string;
+  onContinue: (arg: bigint) => void;
```
peterpeterparker, line 27: shouldn't `onContinue` return a `Promise<void>` according how you use it?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -115,19 +109,22 @@ const pageContent = (
...
               >Lost Access?</a
             >
           </li>
+        </ul>
+        <ul class="c-list--flex">
           <li>
-            <a @click="${onManageClick}" class="t-link" id="manageButton"
-              >Manage Anchor</a
-            >
+            <a href="/" class="t-link" id="manageButton">Home</a>
```
peterpeterparker, line 127: there is no state or store in memory that needs to be reseted on navigation right?

Just asking because we used to reload the window which woudl had for effect to reset potential stores, just in case though better asking

peterpeterparker: but maybe current routing acts also as a reload anyway right

nmattia: Nope, nothing! At least for now

src/frontend/src/flows/authenticate/index.ts
nmattia: `onContinue` can be anything, but it's not `await`ed or anything anywhere. It's really just an action that happens when the user clicks.

peterpeterparker (approved):
Thanks for the explanation in a chat, LGTM

PR #986, opened 2022-11-03
Update using-dev-build dependencies

frederikrothenberger:
This PR updates the npm dependencies of the using-dev-build demo project.


nmattia (approved):
LGTM, thanks! Any idea why the lockfile grew so much?

frederikrothenberger:
> LGTM, thanks! Any idea why the lockfile grew so much?

I'm not sure. First I thought it was because I missed updating one dependency leading to duplication. But fixing it did not make the diff smaller. So there seem to be just a lot more transitive dependencies in the newer versions...

PR #988, opened 2022-11-03
Ensure canister code cache is invalidated when dependencies change

frederikrothenberger:
This PR makes sure that the check for last canister code change matches strictly more files than the dependencies check. Otherwise the `commit_canister` could be a commit older than `commit_deps` leading to a subsequent error when telling cargo to build with `--frozen`.


nmattia (approved):
Makes sense, thanks!

PR #993, opened 2022-11-07
Split authenticate HTML for legibility

nmattia:
This splits the authenticate page's HTML to make it more readable. 99% of this PR is moving code into functions; the remaining 1% is flattening a return type.


peterpeterparker (approved):
LGTM

PR #996, opened 2022-11-08
Refactor authentication protocol

nmattia:
This refactors the code responsible for authentication in the webapp and includes some improvements.

In particular:

* The whole `postMessage` protocol logic is moved to a single function which can easily be checked against the spec.
* Progress updates are shown to the user; in particular window is never closed but if for some reason the caller fails to close the window, user is prompted to do it.
* Event listener is removed after use when listening for incoming messages

### DIFF
--- a/src/frontend/src/flows/authenticate/index.ts
+++ b/src/frontend/src/flows/authenticate/index.ts
@@ -2,9 +2,10 @@ import { html, render, TemplateResult } from "lit-html";
 import { icLogo, caretDownIcon } from "../../components/icons";
 import { footer } from "../../components/footer";
 import { getUserNumber, setUserNumber } from "../../utils/userNumber";
+import { unreachable } from "../../utils/utils";
 import { withLoader } from "../../components/loader";
 import { mkAnchorInput } from "../../components/anchorInput";
-import { AuthenticatedConnection, Connection } from "../../utils/iiConnection";
+import { Connection } from "../../utils/iiConnection";
 import { ref, createRef, Ref } from "lit-html/directives/ref.js";
 import {
   apiResultToLoginFlowResult,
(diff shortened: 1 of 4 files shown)


src/frontend/src/flows/authenticate/index.ts
```diff
@@ -322,11 +255,79 @@ async function handleAuthSuccess(
...
+        message: `It looks like you were sent here for authentication, but no dapp requested authentication.`,
+        primaryButton: "Home",
+      });
+
+      location.hash = "";
+      window.location.reload();
+      break;
+    case "failure":
+      render(
+        html`<h1>
```
frederikrothenberger, line 316: Why is this not using `displayError`?

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -322,11 +255,79 @@ async function handleAuthSuccess(
...
+        message: `"${result.authContext.authRequest.derivationOrigin}" is not a valid derivation origin for "${result.authContext.requestOrigin}"`,
+        detail: result.message,
+        primaryButton: "Continue",
+      }),
+
+    onProgress: (status) => {
+      switch (status) {
+        case "waiting":
+          render(
+            html`<h1>waiting for authentication data from dapp...</h1>`,
```
frederikrothenberger, line 289: I would call it `service` instead of `dapp`. This is also what we had on the authentication page previously: "connect to service"

src/frontend/src/flows/authenticate/index.ts
```diff
@@ -322,11 +255,79 @@ async function handleAuthSuccess(
...
+        html`<h1>
+          Something went wrong during authentication. Authenticating dapp was
+          notified and you may close this page.
+        </h1>`,
+        container
+      );
+      break;
+    case "success":
+      render(
+        html`<h1>Authentication sucessful. You may close this page.</h1>`,
```
frederikrothenberger, line 325: Would it make sense to have screenshots of those pages too?

frederikrothenberger (approved):
Left a few minor comments. LGTM!

src/frontend/src/flows/authenticate/index.ts
nmattia: These will almost never be seen, hence the very simplistic html... I think it's not worth the extra code for extracting the pages + adding them to the showcase

src/frontend/src/flows/authenticate/index.ts
nmattia: Makes sense!

nmattia: Done

PR #997, opened 2022-11-08
Move away from CommonJS

nmattia:
This changes our ts and project config to use es as opposed to commonjs. This will make future changes (moving to vite, SSR support) simpler.


webpack.config.js
```diff
@@ -1,21 +1,27 @@
...
+import { config } from "dotenv";
 
-require("dotenv").config();
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
+
+config();
 
 /** Read the II canister ID from dfx's local state */
-function readCanisterId() {
+async function readCanisterId() {
```
peterpeterparker, line 17: not sure `async` was actually ever needed, is it?

nmattia: Good catch, thanks

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "lib": ["ES2019", "DOM"],
-    "module": "commonjs",
+    "module": "esnext",
```
peterpeterparker, line 9: does it works with `es2020` as well? maybe we can use it to be futurstic by using modules but be a bit preservative at the same time by using a well supported standard?

webpack.config.js
```diff
@@ -1,12 +1,17 @@
...
+import CopyPlugin from "copy-webpack-plugin";
+import CompressionPlugin from "compression-webpack-plugin";
+import HtmlWebpackPlugin from "html-webpack-plugin";
+import MiniCssExtractPlugin from "mini-css-extract-plugin";
+import CssMinimizerPlugin from "css-minimizer-webpack-plugin";
+import { fileURLToPath } from "url";
+import { config } from "dotenv";
 
-require("dotenv").config();
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
```
peterpeterparker, line 12: well played 👍

peterpeterparker (approved):
LGTM, thx

webpack.config.js
nmattia: 100% stackoverflow 😉

tsconfig.json
nmattia: Done!

PR #998, opened 2022-11-08
Deploy archive script and automatic deploy to staging on release

frederikrothenberger:
This PR adds an archive deployment script and adds a manual workflow to deploy the archive to the production II instance.

It extends the release deployment to the staging environment to:
* include the init argument to update the II config with the new archive hash
* deploy the archive


scripts/deploy-archive
```diff
@@ -0,0 +1,104 @@
...
+
+tmp_arg_file=$(mktemp)
+
+echo -n '(blob "' > "$tmp_arg_file"
+hexdump -ve '1/1 "%.2x"' "$wasm" | sed 's/../\\&/g' >> "$tmp_arg_file"
+echo -n '")' >> "$tmp_arg_file"
+
+dfx canister "${network_arg[@]}" call "$canister_id" deploy_archive --argument-file "$tmp_arg_file"
+
+rm "$tmp_arg_file"
```
nmattia, line 104: I'm not sure what's happening here, couple of questions:

* Why use `--argument-file` as opposed to `--argument`? I remember discussion about `--argument` being broken sometimes, is that what's happening here? Why does `--argument-file` work differently? Is there a ticket you can share?
* Why the temporary file business?

Since I don't understand what's happening I'm not sure this is relevant, but here's a snippet that I think encodes the argument like you do but without tmp file:

```bash
echo "(blob \"$(hexdump -ve '1/1 "%.2x"' ./archive.wasm | sed 's/../\\&/g')\")"
cat <(echo "(blob \"$(hexdump -ve '1/1 "%.2x"' ./archive.wasm | sed 's/../\\&/g')\")")

```

both give `(blob "\66...")`, once used as argument, once as file argument. And speaking of response files, doesn't `dfx` support `@...` response file syntax? (like gcc, curl, etc)

frederikrothenberger: * `--argument-file` is used because the full wasm is quite big and resulted in an error (`argument list too long`) when running the script locally. Using a file solves the issue.
* The scripts creates a temporary file so that people don't get their environment cluttered when running the script.
* AFAIK `dfx` does not support response files

nmattia (approved):
LGTM, thanks!

PR #999, opened 2022-11-08
Refactor auth flow further

nmattia:
Bunch of cleanup that I missed in
https://github.com/dfinity/internet-identity/pull/996. In particular, generally the authentication page does not care about the delegation; only the postMessage protocol.

### DIFF
--- a/src/frontend/src/test-e2e/misc.test.ts
+++ b/src/frontend/src/test-e2e/misc.test.ts
@@ -31,6 +31,37 @@ test("Authorize ready message should be sent immediately", async () => {
   });
 }, 300_000);
 
+test("Should allow valid message", async () => {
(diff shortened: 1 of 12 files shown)


peterpeterparker (approved):
Half rubber stamp, half review, 100% approved

src/frontend/src/test-e2e/misc.test.ts
```diff
@@ -31,6 +31,37 @@ test("Authorize ready message should be sent immediately", async () => {
   });
 }, 300_000);
 
+test("Should allow valid message", async () => {
```
nmattia, line 34: added this test because the next test was failing, but had nothing to do with what it was actually testing. This helped clarify what was happening.

PR #1002, opened 2022-11-10
Update test-app config

nmattia:
This updates the canister ID (was out of date) and uses another port (8080 may already be taken by the webpack config in the root).


peterpeterparker (approved):
Rubber stamp

PR #1015, opened 2022-11-15
Fix cancel in add device flow

nmattia:
This makes sure the callback actually calls `window.location.reload()`. Before this, the callback was bound to `window.location.reload`, which may change over time, causing the "Cancel" button to call an invalid reference.


frederikrothenberger (approved):
LGTM!

PR #1018, opened 2022-11-15
Center progress messages on mobile

nmattia:
The progress messages weren't centered vertically on mobile because we disable vertical centering in a media query generally. This bypasses the usual flexbox centering and positions the messages manually.

### DIFF
--- a/src/frontend/src/flows/authenticate/index.ts
+++ b/src/frontend/src/flows/authenticate/index.ts
@@ -43,6 +43,15 @@ export const authenticationFlow = async (
 ): Promise<void> => {
   const container = document.getElementById("pageContent") as HTMLElement;
   render(html`<h1>starting authentication</h1>`, container);
+  const showMessage = (msg: string) =>
+    render(
+      html`<h1
+        style="position: absolute; max-width: 100%; top: 50%; transform: translate(0, -50%);"
+      >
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/authenticate/index.ts
```diff
@@ -43,6 +43,15 @@ export const authenticationFlow = async (
 ): Promise<void> => {
   const container = document.getElementById("pageContent") as HTMLElement;
   render(html`<h1>starting authentication</h1>`, container);
+  const showMessage = (msg: string) =>
+    render(
+      html`<h1
+        style="position: absolute; max-width: 100%; top: 50%; transform: translate(0, -50%);"
```
frederikrothenberger, line 49: Wouldn't it be nicer to extract the styling and add a class for that?

frederikrothenberger (commented):
Have you deployed this somewhere for testing?

src/frontend/src/flows/authenticate/index.ts
nmattia: these are transient screens, not really part of the design system, so I thought it'd be nicer to keep it self contained

frederikrothenberger (approved):
LGTM

PR #1019, opened 2022-11-15
Migrate anchors in stable memory

frederikrothenberger:
This PR adds a stable memory migration for anchors. The anchors will have twice as much space and a more flexible candid schema.

The migration must be explicitly started using upgrade arguments. The migration piggy backs on anchor modifying operations and can also be paused.

Todo still:
- [ ] testing
- [ ] clean up / refactoring


src/internet_identity/src/storage.rs
```diff
@@ -56,69 +87,115 @@
...
+    // version 3: post-migration layout
     version: u8,
     num_users: u32,
     id_range_lo: u64,
     id_range_hi: u64,
     entry_size: u16,
     salt: [u8; 32],
+    entry_size_new: u16,       // post-migration entry size
+    new_layout_start: u32,     // all records < this value are still in the old layout
+    migration_batch_size: u32, // batch size for incremental anchor migration
```
nmattia, line 160: guessing you checked that this all still fits? do the new values just get initialized at 0?

src/internet_identity/src/storage.rs
```diff
@@ -56,69 +87,115 @@
...
+}
+
+struct RecordMeta {
```
nmattia, line 163: what's this?

src/internet_identity/src/storage.rs
```diff
@@ -167,6 +246,22 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned, M: Memory> Storage<T,
         self.flush();
     }
 
+    pub fn migration_state(&self) -> MigrationState {
+        match self.header.version {
+            1 => MigrationState::NotStarted,
+            2 if self.header.migration_batch_size == 0 => MigrationState::Paused {
+                anchors_left: self.header.new_layout_start as u64,
+                batch_size: self.header.migration_batch_size as u64,
+            },
```
nmattia, line 255: Do we even need a "Paused" state? Isn't this the same as having `migration_batch_size == 0`?

src/internet_identity/src/storage.rs
```diff
@@ -56,69 +87,115 @@
...
-    // version 3+: invalid
+    // version 2: migration in progress genesis -> post-migration layout, must have persistent state
+    // version 3: post-migration layout
     version: u8,
     num_users: u32,
     id_range_lo: u64,
     id_range_hi: u64,
     entry_size: u16,
     salt: [u8; 32],
+    entry_size_new: u16,       // post-migration entry size
```
nmattia, line 158: I'm actually a little confused (not 100% related to this PR): why do we store the entry size in the header?

src/internet_identity/src/storage.rs
nmattia: Can we have `layout: LayoutV1, entry_size: DEFAULT_ENTRY_SIZE_V3`?

src/internet_identity/src/storage.rs
```diff
@@ -56,69 +87,115 @@
...
+            layout: Layout::V3,
+            offset: RESERVED_HEADER_BYTES_V3 + record_number as u64 * entry_size as u64,
+            entry_size,
+        }
+    }
+
+    pub fn candid_size_limit(&self) -> usize {
+        // u16 is the length of candid before the actual candid starts
+        self.entry_size as usize - std::mem::size_of::<u16>()
+    }
```
nmattia, line 189: Can you clarify what this is? does this mean we have `size - 2 bytes` for candid, because we prefix the record with the size of the serialized candid structured and hence here return the actual size available for the serialized candid?

src/internet_identity/src/storage.rs
```diff
@@ -129,25 +206,27 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned, M: Memory> Storage<T,
...
         Self {
             header: Header {
                 magic: *b"IIC",
-                version: 1,
+                version: 3,
                 num_users: 0,
                 id_range_lo,
                 id_range_hi,
-                entry_size: DEFAULT_ENTRY_SIZE,
+                entry_size: DEFAULT_ENTRY_SIZE_V3,
```
nmattia, line 223: shouldn't this be v1? any idea why no test broke?

src/internet_identity/src/storage.rs
```diff
@@ -129,25 +206,27 @@ impl<T: candid::CandidType + serde::de::DeserializeOwned, M: Memory> Storage<T,
...
                 salt: EMPTY_SALT,
+                entry_size_new: DEFAULT_ENTRY_SIZE_V3,
+                new_layout_start: 0,
```
nmattia, line 226: I think I misunderstood this. Shouldn't this start at the end of stable memory?

src/internet_identity/src/storage.rs
frederikrothenberger: It was this way before. But I would assume for the following reasons:
* storage used to be generic, so it made sense to store the entry size in the data structure itself
* reading the data from stable memory makes sure that it matches the rest of the data structure. I.e. reading it from a constant could lead to bugs where the constant changes without migrating the data.

src/internet_identity/src/storage.rs
frederikrothenberger: Yes, it does. And data is always allocated as all 0s. But even if it is not 0, as long as the header version is 1, we will not read those values. And when we configure the migration, these values are written --> we never read values we did not write.

src/internet_identity/src/storage.rs
frederikrothenberger: This is a small record that packages the info that `read_internal` and `write_internal` operations need to know to do their job.

> Can we have layout: LayoutV1, entry_size: DEFAULT_ENTRY_SIZE_V3

`read_internal` and `write_internal` could handle it, but in practice it does not occur.

src/internet_identity/src/storage.rs
frederikrothenberger: Yes, that's it. I'll make the comment clearer.

src/internet_identity/src/storage.rs
frederikrothenberger: No, this is `Storage::new`, which initializes `Storage` directly as `v3`, since there is no previous data.

src/internet_identity/src/storage.rs
frederikrothenberger: This is the state after the migration being initialized directly.

src/internet_identity/src/storage.rs
frederikrothenberger: Yes, it's the same. I just thought it was nicer to have it explicitly. `inProgress` without making further progress feels weird, so I would rename it to just `started` then...

frederikrothenberger:
Superseded by #1044.

PR #1020, opened 2022-11-16
Disallow webauthn credentials without credential id

frederikrothenberger:
Currently, II allows registering credentials without credential id that are also not recovery phrases. These credentials are useless in the context of II. No additional credentials of this type should be created.

### DIFF
--- a/src/archive/tests/tests.rs
+++ b/src/archive/tests/tests.rs
@@ -674,7 +674,13 @@ mod stable_memory_tests {
         let add_entry = Entry {
             anchor: ANCHOR,
             operation: Operation::AddDevice {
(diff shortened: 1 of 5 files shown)


frederikrothenberger:
Apparently, it is a valid use-case for NFID to create devices without credential ids.
It is probably better to abandon this idea here entirely and instead create support for generic public key credentials (that do not have the webauthn properties like `key_type`).

Dfinity-Bjoern:
Please don't do this in the backend – there are other use cases as well such as adding a key for CLI tooling.

frederikrothenberger:
> Please don't do this in the backend – there are other use cases as well such as adding a key for CLI tooling.

Yes, we will keep support for (non-webauthn) generic public keys.

PR #1021, opened 2022-11-16
Fix error in persistent state reading

frederikrothenberger:
This PR fixes issues with regards to persistent state reading when none has been written. I.e. it adds additional checks to make sure that the code does not attempt to read memory that it should not.


nmattia (approved):
Rubber stamp

PR #1022, opened 2022-11-17
Don't nest auth prompts

nmattia:
This changes the error/cancelation handling in the authentication flow. Previously, a top-level function would be called, and each error would nest a new function call. This makes sure we have only a single function call and use a loop instead to check for the result.

### DIFF
--- a/src/frontend/src/components/authenticateBox.ts
+++ b/src/frontend/src/components/authenticateBox.ts
@@ -45,12 +45,33 @@ type ChasmOpts = {
 
 /** Authentication box component which authenticates a user
  * to II or to another dapp */
-export const authenticateBox = (
+export const authenticateBox = async (
   connection: Connection,
   templates: AuthTemplates
 ): Promise<LoginData> => {
-  const retryOnError = async (result: LoginFlowResult): Promise<LoginData> => {
+  const promptAuth = () =>
+    new Promise<LoginFlowResult>((resolve) => {
+      authenticateBoxTemplate({
+        templates,
+        addDevice: () => addRemoteDevice(connection),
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/authenticateBox.ts
```diff
@@ -45,52 +45,48 @@ type ChasmOpts = {
...
+        register: () => {
+          resolve(registerIfAllowed(connection));
+        },
+        recoverAnchor: (userNumber) => useRecovery(connection, userNumber),
+        userNumber: getUserNumber(),
+      });
+    });
+
+  // Retry until user has successfully authenticated
+  for (;;) {
```
nmattia, line 69: equivalent to `while(true)` but has the advantage of being accepted by eslint by default: https://github.com/eslint/eslint/issues/5477

src/frontend/src/components/authenticateBox.ts
```diff
@@ -45,52 +45,48 @@ type ChasmOpts = {
...
-        return authenticateBox(connection, templates);
-      case "ok":
-        return result;
+        break;
       case "canceled":
-        return authenticateBox(connection, templates);
+        break;
       default:
         unreachable(result);
         break;
```
peterpeterparker, line 87: shouldn't  this `break` becomes now a `return` otherwise it might continue looping if `default` is reached which was not the case before?

nmattia: `default` is actually never reached, that's what `unreachable` ensures. A `return` here would actually break the flow and make the app non-responsive because the promise is never resolved

peterpeterparker (approved):
LGTM, thx

PR #1024, opened 2022-11-17
Use anchor picker in authenticate

nmattia:
This integrates the anchor picker component in the authenticate flow/prompt.

The component supports selecting from multiple anchors, but for simplicity we only support a single anchor for now (to avoid messing with the local storage schema just yet).


src/frontend/src/components/anchorPicker.ts
```diff
@@ -64,7 +71,7 @@ function elements(props: PickerProps): TemplateResult[] {
         // replace the "use another anchor" button with actual menu when ready (clicked)
         new Promise<void>((resolve) => {
           otherAnchorOpen = resolve;
-        }).then(() => Promise.resolve(otherAnchorMenu({ pick: props.pick }))),
+        }).then(() => Promise.resolve(otherAnchorMenu(props))),
```
nmattia, line 74: I know, will try to get rid of this `.then()` later

src/frontend/src/components/anchorPicker.ts
```diff
@@ -115,22 +122,47 @@ function otherAnchorMenu(props: { pick: PickCB }): TemplateResult {
...
-        <button class="c-button c-button--secondary">
-          Create a new anchor
-        </button>
-      </div>
+      ${mkLinks(props)}
     </div>
   `;
 }
+
+const mkLinks = ({
```
nmattia, line 140: just moved

frederikrothenberger (approved):
I like the new box and it works well. 👍

However, I think we should add screenshots for the no anchor selected / "more options" clicked variant of the box.

nmattia:
> However, I think we should add screenshots for the no anchor selected / "more options" clicked variant of the box.

Good call, will do this in a separate PR

PR #1028, opened 2022-11-18
Safely handle authentication requests

nmattia:
Until now, an authentication request with `kind: "authorize-client"` but nothing else would break II, since it'd get accepted as an "AuthRequest" (subtyped from `event.data: any`).

This first turns `event.data` into an `unknown` and then narrows down the type correctly.

### DIFF
--- a/src/frontend/src/flows/authenticate/postMessageInterface.ts
+++ b/src/frontend/src/flows/authenticate/postMessageInterface.ts
@@ -4,6 +4,7 @@ import { Principal } from "@dfinity/principal";
 import { fetchDelegation } from "./fetchDelegation";
 import { LoginData } from "../login/flowResult";
 import { validateDerivationOrigin } from "./validateDerivationOrigin";
+import { hasOwnProperty } from "../../utils/utils";
 
 export interface Delegation {
   delegation: {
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/authenticate/postMessageInterface.ts
```diff
@@ -36,6 +37,60 @@ export interface AuthRequest {
...
+  }
+
+  if (msg === null) {
+    return undefined;
+  }
+
+  // Some extra conversions to take typescript by the hand
+  // eslint-disable-next-line
+  const tmp: {} = msg;
+  const obj: Record<string, unknown> = tmp;
```
frederikrothenberger, line 53: Why not just `const obj = msg as Record<string, unknown>;`

frederikrothenberger (approved):

src/frontend/src/flows/authenticate/postMessageInterface.ts
nmattia: Because in this case it's not assigning anymore, but it's casting, meaning you could even do that:

```ts
const obj = msg as AuthRequest;
```

PR #1030, opened 2022-11-18
Enforce persistent state being present for storage layouts 2 and up

frederikrothenberger:
This allows being mores strict on the persistent state being present in the future.


nmattia (approved):
LGTM, thanks!

PR #1032, opened 2022-11-21
Add hover effect to anchors

meodai:
Added a very subtle hover effect to the anchor list.


meodai:
https://user-images.githubusercontent.com/608386/203060570-d20229c4-8ae4-421c-81fc-ce99d01d23c0.mov

PR #1033, opened 2022-11-21
Refactor registration flow

nmattia:
This cleans up the registration code and changes the module structure a bit. In particular the device alias picker is made to use idiomatic lit, most `.then()`s are replaced in favor of more consistent `async/await` and some tests are added.


src/frontend/src/flows/register/alias.test.ts
```diff
@@ -0,0 +1,36 @@
+import { promptDeviceAliasPage } from "./alias";
+import { render } from "lit-html";
```
frederikrothenberger, line 2: Unused import

src/frontend/src/flows/register/index.ts
```diff
@@ -0,0 +1,42 @@
+import { Connection } from "../../utils/iiConnection";
+import { unknownToString } from "../../utils/utils";
+import { confirmRegister } from "./captcha";
+import { makeCaptcha } from "./captcha";
+import { LoginFlowResult, cancel } from "../login/flowResult";
+import { constructIdentity } from "./construct";
+import { promptDeviceAlias } from "./alias";
+
+/** Registration (anchor creation) flow for new users */
+export const register = async ({
```
frederikrothenberger, line 10: This is so much nicer and clearer to follow! 👍

src/frontend/src/flows/register/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (alias === null) {
+      return cancel;
+    }
+
+    const [captcha, identity] = await Promise.all([
+      makeCaptcha(connection),
+      constructIdentity(),
+    ]);
+
+    const result = await confirmRegister(
```
frederikrothenberger, line 26: (nit, because was this way previously) The captcha module still contains the part about congratulating the user. I would have expected it here, as it has nothing to do with captcha.

frederikrothenberger (approved):
LGTM, thanks!

PR #1036, opened 2022-11-21
Add spinner for auth progress

nmattia:
This adds an animated spinner for authentication progress and changes the auth wording to avoid confusing terms like "delegation".


src/frontend/src/flows/authenticate/index.ts
```diff
@@ -45,11 +46,12 @@ export const authenticationFlow = async (
...
   const showMessage = (msg: string) =>
     render(
-      html`<h1
-        style="position: absolute; max-width: 100%; top: 50%; transform: translate(0, -50%);"
-      >
-        ${msg}
-      </h1>`,
+      html`
+        <div class="l-container c-card c-card--highlight t-centered">
+          <div class="c-spinner">${spinner}</div>
```
meodai, line 51: should we provide the wrapping div whit the icon? So it comes with everything you need to make it work?

nmattia: Up to you, I'm happy to go with this and potentially add it later if we find ourselves adding it everywhere

meodai (approved):
LGTM

PR #1046, opened 2022-11-23
Store multiple anchors in local storage

nmattia:
This updates our local storage handling to actually store up to 10 anchors, which we then all display on authentication pages.

UX note: We keep "infinitely" many anchors because we want to visually discourage users from creating too many anchors (but really we keep only 10 for practical reasons); we also keep as many anchors as possible because this is the only place where the user will see all their anchors.

Implementation note: we make the anchor reading/parsing as non-destructive as possible in case we e.g. add a field in the future and then need to rollback for some reason but want to keep the field.


src/frontend/src/utils/userNumber.ts
```diff
@@ -1,18 +1,186 @@
...
+  anchors.sort();
+
+  return anchors;
+};
+
+/** Set the most recently used anchor */
+const setAnchorUsed = (anchor: bigint) => {
+  const anchors = readAnchors();
+  for (const ix in anchors) {
+    anchors[ix].unusedCount++;
```
peterpeterparker, line 55: not sure what's the goal here, if you can explain me in a chat...

src/frontend/src/utils/userNumber.ts
```diff
@@ -1,18 +1,186 @@
...
+
+  arr.sort((a, b) => b.unusedCount - a.unusedCount);
+
+  return arr[0]?.ix;
+};
+
+/* "Low-level" serialization functions to read and write anchors to local storage */
+
+const readAnchors = (): Anchors => {
+  const blob = localStorage.getItem("anchors");
```
peterpeterparker, line 125: by reading `blob` I kind of except it to be a `Blob` while actually it's a `string`. Renaming the variable to e.g. `text` would be a bit more easy to read and intepret.

src/frontend/src/utils/userNumber.ts
```diff
@@ -1,18 +1,186 @@
...
+  migrate();
+  return { getAnchors, setAnchorUsed };
+};
+
+/** Read saved anchors from local storage, returning at most
+ * MAX_SAVED_ANCHORS */
+const getAnchors = (): bigint[] => {
+  const anchors = [];
+  const data = pruneAnchors(readAnchors());
+  for (const ix in data) {
```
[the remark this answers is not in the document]
nmattia: `for ... in` returns the indices of the object being iterated, here meaning the keys of the record, which are the actual anchors/user numbers; this just constructs an array of all anchor numbers. This is probably better done with `Object.keys()`, will update!

peterpeterparker (approved):
LGTM, thx

PR #1047, opened 2022-11-23
Allow the archive to use the full 32GB stable memory

frederikrothenberger:
The archive was developed when the stable memory limit was still 8GB. This PR updates the limit to 32 GB.


nmattia (approved):
nothing else to change?

frederikrothenberger:
> nothing else to change?

No, this just changes the limit how far the `MemoryManager` is allowed to hand out memory. Everything else can grow dynamically. 🙂

PR #1049, opened 2022-11-23
Revert "Safely handle authentication requests (#1028)"

nmattia:
This reverts commit 425d586f460f0ace01d31f09b4e7bfcb33f417ca.

This seems to be causing some issues with some dapps due to the fact that we're less lenient about what we accept.


frederikrothenberger:
@nmattia: Does this block certain versions of the `authClient` or only implementations unknown to us?

nmattia:
@frederikrothenberger still figuring that out, but I suspect it only blocks implementations unknown to us!

nmattia:
Superseded by #1051

Issue #1050, opened 2022-11-24
Ensure maxTimeToLive is bigint and not number


nmattia:
It looks like some clients (in particular (taggr)[https://taggr.top]) use a `number` `maxTimeToLive` in the authentication request, whereas [Internet Identity expects a `bigint`](https://internetcomputer.org/docs/current/references/ii-spec#client-authentication-protocol).

For now we'll be lenient and accept `bigint` as well as `number` but we should ensure all our client libraries enforce `bigint`.

PR #1051, opened 2022-11-24
Accept number maxTimeToLive

nmattia:
This makes our postMesasge interface a bit more lenient (see https://github.com/dfinity/internet-identity/issues/1050 for details) and adds console logs to ease debugging of bad authentication requests.


frederikrothenberger (approved):
Maybe having tests for this interface would help with refactoring in the future.

nmattia:
> would help with refactoring in the future.

Just to be clear, we didn't break anything here! Testing against the interface wouldn't have helped us prevent the issue on Taggr.

That being said, agreed to test! Will submit in a separate PR

PR #1052, opened 2022-11-24
Clarify meaning of authenticate

nmattia:
This renames and restructures flows and clarifies the meaning of "authenticate" which we've been using left and right:
* "Authorize"/"Authorization" now refers to the process of authorizing the use of an anchor/device for another dapp
* "Authenticate"/"authentication" now refers to the actual process of authentication (e.g. using WebAuthn etc) for both "manage" and "authorize" flows

The code in `flows/login` was moved to either utils, manage, or authenticate.


peterpeterparker (approved):
Rubber stamp

PR #1053, opened 2022-11-24
Expose archive canister status

frederikrothenberger:
This PR adds a public method to retrieve the archive canister status. Since II is the only controller of the archive there is currently no way of knowing the cycles balance of the archive canister. Adding a public status method fixes this.

### DIFF
--- a/src/archive/archive.did
+++ b/src/archive/archive.did
@@ -126,7 +126,7 @@ type StreamingStrategy = variant {
 
 // management canister types: https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-candid
 
-type DefiniteCanisterCettings = record {
+type DefiniteCanisterSettings = record {
     controllers : vec principal;
     compute_allocation : nat;
     memory_allocation : nat;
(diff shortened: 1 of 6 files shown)


src/archive/archive.did
```diff
@@ -124,6 +124,28 @@ type StreamingStrategy = variant {
     };
 };
 
+// management canister types: https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-candid
+
+type DefiniteCanisterCettings = record {
```
nmattia, line 129: Cettings?

also the interface spec uses snake case, should we use that too?

frederikrothenberger: I'd rather be consistent with what we have and the maybe change everything to snake case in a later PR.

nmattia: Looks like `snake_case` is the preferred way: https://github.com/dfinity/candid/blob/master/docs/modules/candid-guide/pages/candid-concepts.adoc#reusing-complex-types

Works for me, let's change everything!

nmattia (approved):

PR #1057, opened 2022-11-25
Limit the size of variable fields across all devices

frederikrothenberger:
This PR introduces a limit on total size of variable fields across all devices. This limit is equal to the current total anchor size so it will not impact any anchors before the stable memory migration.

After the memory migration, this limit makes sure that anchors do not fill up the additional space with more device data. In the future, the additional space will be uesful when adding additional fields and / or changing the candid type of anchors in stable memory.


nmattia (approved):
Rubber stamp

PR #1058, opened 2022-11-25
Use a fixed release in stable memory migration integration tests

frederikrothenberger:
This PR removes the assumption that II_WASM_PREVIOUS initializes storage in layout v1 from the stable memory migration tests.


nmattia (approved):
Rubber stamp

PR #1063, opened 2022-12-01
Remove automatic changing of user range

frederikrothenberger:
This PR removes code that automatically changes the user range when
 storage capacity changes.
This is no longer desired, as other configuration changes are explicit
 using init arguments.


nmattia (approved):
Thanks! The user range is persisted now right?

frederikrothenberger:
> Thanks! The user range is persisted now right?

The range has always been persisted, but the upper end was modified based on this now removed logic.

PR #1067, opened 2022-12-01
Use typescript for webpack config

nmattia:
This changes the webpack config to use TypeScript.

This makes the config easier to change, and already helped uncover two issues.

Webpack itself and two plugins have been updated (new versions include type definitions).

A new `npm` script is added which sets proper `NODE_OPTIONS` for our typescript scripts.


peterpeterparker (approved):
LGTM, thx

PR #1070, opened 2022-12-01
Simplify webpack config

nmattia:
This cleans up the dev server by

* removing the port setting (was set to default)
* disabling "magicHtml" which tries to serve js files like `index.js` on `/index` with basic HTMl, but which is just confusing in practice
* explicitly disabling `static` assets serving, since we serve everything from our webpack pipeline
* Resetting allowed hosts to the default `"auto"` because custom values weren't used anymore


frederikrothenberger (approved):
Tried it and it still works 😉👍

PR #1071, opened 2022-12-02
Rework archive metrics according to best practices

frederikrothenberger:
This PR improves the archive metrics to better follow [Prometheus best practices](https://prometheus.io/docs/practices/naming/).

### DIFF
--- a/src/archive/src/main.rs
+++ b/src/archive/src/main.rs
@@ -428,45 +428,42 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
         )
     })?;
     with_log(|log| {
-        w.encode_gauge(
-            "ii_archive_entries_count{source=\"log\"}",
-            log.len() as f64,
-            "Number of log entries stored in this canister.",
-        )?;
-        w.encode_gauge(
-            "ii_archive_log_bytes_total{memory=\"data\"}",
-            log.data_size_bytes() as f64,
-            "Total size of all logged entries in bytes.",
-        )?;
-        w.encode_gauge(
-            "ii_archive_log_bytes_total{memory=\"index\"}",
-            log.index_size_bytes() as f64,
-            "Total size of the log index in bytes.",
-        )
-    })?;
-    with_anchor_index_mut(|index| {
-        w.encode_gauge(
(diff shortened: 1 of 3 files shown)


nmattia (approved):
Rubber stamp

src/archive/src/main.rs
```diff
@@ -421,54 +422,49 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
             "Timestamp of the last upgrade of this canister.",
         )
     })?;
     with_log(|log| {
         w.encode_gauge(
-            "ii_archive_log_entries_count",
+            "ii_archive_entries_count{source=\"log\"}",
             log.len() as f64,
             "Number of log entries stored in this canister.",
         )?;
```
alin-at-dfinity, line 435: The latest version of `metrics_encoder` has [explicit support for labeled metrics](https://gitlab.com/dfinity-lab/public/ic/-/blob/86d4fe8bc8e62a31776a42de0aebb26a745ce32a/rs/monitoring/metrics_encoder/src/lib.rs#L161-178) and [checks against non-alphanumeric characters in metric names](https://gitlab.com/dfinity-lab/public/ic/-/blob/86d4fe8bc8e62a31776a42de0aebb26a745ce32a/rs/monitoring/metrics_encoder/src/lib.rs#L229-246), meaning this change would start failing as soon as you update the `metrics_encoder` dependency, even though it will work for now.

I would pull in the latest version of `metrics_encoder` and use `w.gauge_vec` here instead.

src/archive/src/main.rs
```diff
@@ -421,54 +422,49 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
         w.encode_gauge(
-            "ii_archive_log_entries_size",
-            log.log_size_bytes() as f64,
-            "Total size of all logged entries in bytes, not counting data structure overhead.",
+            "ii_archive_log_bytes_total{memory=\"data\"}",
```
alin-at-dfinity, line 437: `_total` is a suffix that denotes counters (monotonically increasing metrics that keep track of the number of occurrences of an event). This is (and should continue to be) a gauge, it should not end in `_total`.

Same everywhere else.

src/archive/src/main.rs
frederikrothenberger: Thanks!

PR #1072, opened 2022-12-02
Serve static pages statically

nmattia:
This updates our webpack config to pre-render the FAQ and about pages, which are then served as HTML from the canister (and then hydtated in the browser). This means that `/about` and `/faq` can be browsed (and more importantly scraped/indexed) without JavaScript. A side-effect is faster page load.

The process for generating static pages is a little manual right now, though we may automate it/abstract it if we decide to have more static pages:
1. Use `lit-ssr` to render a string version of the page during the webpack build
1. Inject that content in the HTML page served by the canister
1. Hydrate the page before the first `render()`


webpack.config.ts
```diff
@@ -31,27 +36,72 @@ function readCanisterId() {
...
+  pageName: string,
+  pageContent: TemplateResult
+): HtmlReplacePlugin =>
+  new HtmlReplacePlugin((html) => {
+    const content = Array.from(render(pageContent)).reduce(
+      (acc, v) => acc + v,
+      ""
+    );
+    return html.replace(
+      '<main id="pageContent" class="l-wrap" aria-live="polite"></main>',
```
peterpeterparker, line 85: if someone replaces the style class or the tag `main` one day, this replace won't work anymore. would it be more handy to replace a placeholder within the tag?

```
// config
return html.replace(
      '<!-- I_WILL_BE_REPLACED -->',
      
// effective html      
<main id="pageContent" class="l-wrap" aria-live="polite">
   <!-- I_WILL_BE_REPLACED -->
</main>
```

src/internet_identity/src/assets.rs
```diff
@@ -73,23 +92,24 @@ pub fn init_assets() {
 // prepared only once (like injecting the canister ID).
 fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType); 8] {
     let index_html: &[u8] = INDEX_HTML_STR.as_bytes();
+    let about_html: &[u8] = ABOUT_HTML_STR.as_bytes();
```
peterpeterparker, line 95: out of curiosity, any particular reason you created extra variables instead an hashmap (or else) of static HTML assets?

peterpeterparker (approved):
LGTM, thanks. Minor comments and questions.

src/internet_identity/src/assets.rs
nmattia: Just for consistency; first I copy pasted `index_html` then I decided to inline it and then I reverted (for consistency with `index_html`). Don't feel particularly strongly about it though

webpack.config.ts
nmattia: That's a good point. I decided to go with this because in the rest of the app we explicitly use `#pageContent` as a container where we `render()` from, so I wanted to make sure this is consistent (since the prerendered version needs to be inserted at the same spot). I'll think a bit more.

src/internet_identity/src/assets.rs
```diff
@@ -23,29 +23,48 @@ pub enum ContentType {
...
-        let index_html = include_str!("../../../dist/index.html");
-        let canister_id = api::id();
-        let setup_js: String = INDEX_HTML_SETUP_JS.to_string();
-        let index_html = index_html.replace(
-            r#"<script id="setupJs"></script>"#,
-            &format!(r#"<script data-canister-id="{canister_id}" id="setupJs">{setup_js}</script>"#).to_string()
-        );
-        let index_html = index_html.replace("<meta replaceme-with-csp/>", &format!(r#"<meta http-equiv="Content-Security-Policy" content="{}" />"#,&http::content_security_policy_meta() ));
-        index_html
+        fixup_html(include_str!("../../../dist/index.html"))
```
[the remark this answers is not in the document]
nmattia: I think the macro will only accept a string literal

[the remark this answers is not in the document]
nmattia: (this happens at build time)

webpack.config.ts
nmattia: @peterpeterparker thinking about this more, I realize I have no idea what's best here. I'd suggest we go with this for now, and adapt as we go along (i.e. when/if more pages are added). Wdyt?

peterpeterparker: I can live with it, I already accepted the PR, sure go on

PR #1075, opened 2022-12-05
Fix dead links

nmattia:
A lot of pages moved from https://sdk.dfinity.org to https://internetcomputer.org. The links were still redirected but to pages that didn't exist. This fixes those links.


peterpeterparker (approved):
Thanks

PR #1076, opened 2022-12-06
Catch error on key type update

nmattia:
This is a workaround for unexpected authenticator attachments.

On Windows the attachment may be `null`, which we now treat as `undefined`. In general this also updates the error handling to ensure the key type update does not throw an exception, and updates the "Impossible happened" error message.

### DIFF
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -243,7 +243,11 @@ export class Connection {
 
     const attachmentInfo = identity.getAuthenticatorAttachment();
     if (attachmentInfo !== undefined) {
-      this.updateKeyTypeIfNecessary(devices, attachmentInfo, connection);
+      try {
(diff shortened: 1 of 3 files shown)


frederikrothenberger (approved):
Thanks!

Olayemi10:
I'm unable to access my dscvr with Internet identity ...please help me

nmattia:
hi @Olayemi10 , feel free to reach to support for any issues: https://support.dfinity.org/hc/en-us you can also ask on the [forum](http://forum.dfinity.org)

peterpeterparker:
@Olayemi10 already asked the same question on the forum and I spotted it. I forward the info to the support team which responded there too.

https://forum.dfinity.org/t/the-impossible-has-happened/17080/34?u=peterparker

Olayemi10:
Ok thanks...I've actually been communicating with someone on here since
yesterday

PR #1077, opened 2022-12-06
Update to typescript 4.9

nmattia:
In particular, the `in` operator now better narrows types and allows us to get rid of most `hasOwnProperty` calls.

* https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#in-narrowing


src/frontend/src/flows/addDevice/manage/verifyTentativeDevice.ts
```diff
@@ -2,7 +2,6 @@ import { html, render } from "lit-html";
 import { AuthenticatedConnection } from "../../../utils/iiConnection";
 import { withLoader } from "../../../components/loader";
 import { renderManage } from "../../manage";
-import { hasOwnProperty } from "../../../utils/utils";
```
frederikrothenberger, line 5: Is there a reason to not delete the `hasOwnProperty` function?

frederikrothenberger (approved):
Nice!

package.json
```diff
@@ -52,7 +52,7 @@
     "ts-jest": "~29",
     "ts-loader": "9.4.1",
     "ts-node": "^10.8.1",
-    "typescript": "4.8.4",
+    "typescript": "~4",
```
peterpeterparker, line 55: I would keep a version pinned to be sure there isn't unexpected build issue in the ci or suddently locally

src/frontend/src/flows/addDevice/manage/verifyTentativeDevice.ts
nmattia: Yeah, I couldn't get rid of it in the `utils` module just yet

package.json
nmattia: How would that happen since we have the lockfile?

peterpeterparker: If you run any `npm i` or `npm i another_lib` it might be bumped my understanding. So yeah less chance in the CI, more chance locally. Maybe not tomorrow, not after tomorrow, but some day.

peterpeterparker (approved):
LGTM

PR #1080, opened 2022-12-06
Move registration screens out of captcha

nmattia:
Before this, the captcha module dealt with general login error handling (duplicating the work of `apiResultToLoginFlowResult`) and other registration matters like showing the user their anchor.

These are now moved to `register/index.ts`.


frederikrothenberger (approved):
This makes the error pages a little bit less readable. But we should never hit those anyway.

PR #1085, opened 2022-12-09
Temporarily disable screenshots

frederikrothenberger:
The latest version of chrome triggered a bug so that screenshots on mobile now have the wrong aspect ratio and produce flaky results.

Upgrading webdriverio to version 8 would fix the issue, however this then breaks compatibility with jest.

Since it is a bigger change to make the screenshot infrastructure work again, this PR disables the screenshots (temporarily) so that we can deal with fixing the infrastructure properly at a later stage.


nathanosdev (approved):

frederikrothenberger:
@peterpeterparker: The issue is that old chrome binaries are not freely available. Google only provides the latest. Of course we could use one from a third party source, but I'm not sure how reliable they are.

But I agree, any proper solution to this issue involves building something that also pins the chrome version.

PR #1093, opened 2022-12-14
Add candid schema migration

frederikrothenberger:
This PR introduces the possibility to migrate the anchors to the new record based stable memory layout. This is the second part of the stable memory migration in preparation of the domain migration.


src/internet_identity/src/storage.rs
```diff
@@ -511,6 +539,97 @@ impl<M: Memory> Storage<M> {
...
+            self.header.version, 4,
+            "unexpected header version during migration finalization {}",
+            self.header.version
+        );
+        assert_eq!({ self.header.new_layout_start }, 0, "cannot finalize migration when not all anchors were migrated! Remaining anchors to migrate: {}", { self.header.new_layout_start });
+
+        self.header.version = 5;
+        // clear now unused header fields
+        self.header.migration_batch_size = 0;
+        self.header.migration_batch_size = 0;
```
nathanosdev, line 631: `self.header.migration_batch_size = 0` is set twice here, is the 2nd one meant to be `self.header.new_layout_start = 0`?

`layout_start` is already 0 at this point anyway

src/internet_identity/src/storage/tests.rs
```diff
@@ -94,11 +106,28 @@ fn should_serialize_first_record() {
 }
 
 #[test]
-fn should_serialize_subsequent_record_to_expected_memory_location() {
+fn should_serialize_first_record_v5() {
```
nathanosdev, line 109: I really like the approach of writing to storage with method you're testing and then reading directly from memory to make your assertions so the method under test is isolated, likewise for reading.

But I think it might also be useful to have a sanity check test where both storage read and write methods are used, so you can be sure that they are compatible with one another. Otherwise I think it's possible that the direct read/writes in the tests can deviate from the `Storage` implementation of reading and writing.

nathanosdev: On second thoughts, this is already covered in `should_keep_anchors_intact_when_migrating` in `src/internet_identity/tests/schema_migration_tests.rs`

frederikrothenberger: Thanks for pointing it out! This is a property that should not only be accidentally tested --> added tests for that.

nathanosdev (approved):

PR #1098, opened 2022-12-19
Use precompiled state machine binary to run canister tests

frederikrothenberger:
This PR introduces the precompiled state machine binary to run canister integration tests. This greatly speeds up the CI pipeline.

Note: the crate `state_machine_client` will be moved to a separate repository in the future and published on crates.io.

### DIFF
--- a/src/canister_tests/src/framework.rs
+++ b/src/canister_tests/src/framework.rs
@@ -1,13 +1,11 @@
 use crate::api;
-use candid::utils::{decode_args, encode_args, ArgumentDecoder, ArgumentEncoder};
 use candid::Principal;
 use flate2::read::GzDecoder;
 use flate2::{Compression, GzBuilder};
+use ic_cdk::api::management_canister::main::CanisterId;
 use ic_crypto_iccsa::types::SignatureBytes;
 use ic_crypto_iccsa::{public_key_bytes_from_der, verify};
-use ic_state_machine_tests::{
-    CanisterId, ErrorCode, PrincipalId, StateMachine, UserError, WasmResult,
-};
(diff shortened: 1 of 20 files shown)


src/state_machine_client/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
nmattia, line 1: can we also get a README in this directory?

src/canister_tests/src/framework.rs
```diff
@@ -135,14 +134,39 @@ fn get_wasm_path(env_var: String, default_path: &path::PathBuf) -> Option<Vec<u8
...
 
-/* Here are a few useful helpers for writing tests */
-pub fn install_empty_canister(env: &StateMachine) -> CanisterId {
-    env.install_canister(EMPTY_WASM.clone(), vec![], None)
-        .expect("failed to install empty canister.")
+/** The Wasm module for the current II build, i.e. the one we're testing */
+pub static STATE_MACHINE_BINARY: &str = "../../ic-test-state-machine";
+
+pub fn env() -> StateMachine {
+    let path = match env::var_os("STATE_MACHINE_BINARY") {
```
nmattia, line 141: shouldn't `state_machine_client` deal with this?

nmattia (approved):
mostly rubber stamping

src/canister_tests/src/framework.rs
frederikrothenberger: Hm, I'm not convinced. The library should not keep where to find the binary or which `env` variable can be supplied insted...

frederikrothenberger: We can still move this code later, if you feel strongly about it.

PR #1099, opened 2022-12-19
Pin chromium version and re-enable screenshots

nmattia:
This introduces a step before starting the screenshots machinery that downloads and caches chromium.

The download script is based on
https://github.com/juliangruber/download-chromium but simplified and ported to TypeScript.


src/frontend/download-chrome.ts
```diff
@@ -0,0 +1,112 @@
...
+
+function getConfig(): Config {
+  const platform = os.platform();
+
+  if (platform !== "linux" && platform !== "darwin") {
+    throw new Error(`Platform ${platform} is not supported`);
+  }
+
+  // Follow steps here to update: https://www.chromium.org/getting-involved/download-chromium/
+  const revision = platform === "darwin" ? "1036822" : "1036826";
```
frederikrothenberger, line 75: What version does this correspond to? 106?

frederikrothenberger (approved):
Thanks a lot!

src/frontend/download-chrome.ts
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env node
+import * as os from "os";
```
peterpeterparker, line 2: even though it is not that super important since not bundled, I would avoid importing `*` anyway as a best practice

src/frontend/download-chrome.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const stat = promisify(fs.stat);
+const unlink = promisify(fs.unlink);
+
+const installPath = `${os.homedir()}/.local-chromium`;
+
+type Config = { platform: "linux" | "darwin"; revision: string };
+
+/* This downloads Chromium from the Chromium build archives
+ * For more info on finding a particular version of chromium, see https://www.chromium.org/getting-involved/download-chromium/
```
peterpeterparker, line 20: maybe link also the tool of Cypress which ease the process of finding a version: https://chromium.cypress.io/

src/frontend/download-chrome.ts
```diff
@@ -0,0 +1,112 @@
...
+import { promisify } from "util";
+import * as fs from "fs";
+import got from "got";
+import stream from "node:stream";
+import extract from "extract-zip";
+import { unreachable } from "./src/utils/utils";
+
+const pipeline = promisify(stream.pipeline);
+
+const stat = promisify(fs.stat);
```
peterpeterparker, line 12: which version of NodeJs is used? you cannot import the promises "directly"?

```
import {stat} from 'fs/promises';
```

btw. isn't `stat` deprecated?

```
import {access} from 'fs/promises';

try {
   await access(moduleExecutablePath)
   ...
```

src/frontend/download-chrome.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  // Follow steps here to update: https://www.chromium.org/getting-involved/download-chromium/
+  const revision = platform === "darwin" ? "1036822" : "1036826";
+
+  return { platform, revision };
+}
+
+/** The chromium download URL for the given platform and revision */
+function downloadURL(config: Config) {
+  if (config.platform === "darwin") {
```
peterpeterparker, line 82: nth:

- you can deconstruct `{revision, platform}`
- `https://www.googleapis.com/storage/v1/b/chromium-browser-snapshots/o/` seems to be the same for the two urls

src/frontend/download-chrome.ts
```diff
@@ -0,0 +1,112 @@
...
+    return `https://www.googleapis.com/storage/v1/b/chromium-browser-snapshots/o/Mac%2F${config.revision}%2Fchrome-mac.zip?alt=media`;
+  } else if (config.platform === "linux") {
+    return `https://www.googleapis.com/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F${config.revision}%2Fchrome-linux.zip?alt=media`;
+  }
+}
+
+/** Name of the archive (depends on platform) */
+function archiveName(config: Config) {
```
peterpeterparker, line 90: as I just notice it's a `.ts` file, the return types of the functions is missing

```
function archiveName(config: Config): string | undefined {
...
```

src/frontend/screenshots.ts
```diff
@@ -66,10 +67,12 @@ async function withChrome<T>(
 ): Promise<T> {
   // Screenshot image dimension, if specified
   const { mobileEmulation } = readScreenshotsConfig();
+  const chromePath = await downloadChrome();
```
peterpeterparker, line 70: should it be wrapped with a try / catch?

src/frontend/download-chrome.ts
```diff
@@ -0,0 +1,112 @@
...
+  try {
+    fs.mkdirSync(installPath);
+  } catch (_) {
+    /* noop */
+  }
+  const folderPath = getFolderPath(config, installPath);
+  const zipPath = `${folderPath}.zip`;
+
+  await pipeline(
+    await got.stream(downloadURL(config)),
```
peterpeterparker, line 56: would a download function such as https://github.com/peterpeterparker/create-ic/blob/c7f1359e7a8771c71a86517bbb4dbafeeef66e3d/src/utils/download.utils.ts#L10 work as well? not sure it supports big download but if it does, maybe you can spare a dependency

src/frontend/download-chrome.ts
nmattia: indeed, added comment

src/frontend/download-chrome.ts
nmattia: oh this is NICE!

nmattia: Added

src/frontend/download-chrome.ts
nmattia: agreed; copied over from original script. Will change

src/frontend/download-chrome.ts
nmattia: thanks, updated to fs/promises! imports were copied over from original script. will keep `stat` for now, just because it works

src/frontend/download-chrome.ts
nmattia: ah, that's awesome! thanks!

PR #1100, opened 2022-12-20
Encapsulate anchor in its own module

frederikrothenberger:
This makes the anchor and device constraints unit testable and separates the logic of validating data from the other tasks such as authentication or housekeeping.

### DIFF
--- a/src/internet_identity/src/storage/anchor.rs
+++ b/src/internet_identity/src/storage/anchor.rs
@@ -1,0 +1,391 @@
+use candid::{CandidType, Deserialize, Principal};
+use internet_identity_interface::archive::DeviceDataWithoutAlias;
+use internet_identity_interface::*;
+use std::{fmt, iter};
+
+#[cfg(test)]
+mod tests;
+
+/// Internal representation of the anchor.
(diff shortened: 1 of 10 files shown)


src/internet_identity/src/storage/anchor.rs
```diff
@@ -0,0 +1,391 @@
...
+            .iter()
+            .find(|e| e.pubkey == device.pubkey)
+            .is_some()
+        {
+            return Err(AnchorError::DuplicateDevice {
+                device_key: device.pubkey,
+            });
+        }
+        check_device(&device)?;
+        check_invariants(&self.devices.iter().chain(iter::once(&device)).collect())?;
```
nathanosdev, line 85: I think I would have understood the intention behind these functions a little easier if they were named `check_device_invariants` and `check_anchor_invariants`.

nathanosdev (approved):
This is a really nice refactoring!

src/internet_identity/src/storage/anchor.rs
frederikrothenberger: good suggestion 👍 --> renamed

PR #1102, opened 2022-12-20
Update rust dependencies and unpin IC repo

frederikrothenberger:
This PR retires pinning a specific commit of the IC repository and the scheduled update of the given dependencies repo since we only depend on stable apis.
The scheduled update job is changed to update dependencies in general via the cargo update command.


nmattia (approved):
LGTM, thanks!

PR #1104, opened 2022-12-22
Remove support for layout v4 and lower

frederikrothenberger:
The candid layout migration was completed successfully so the migration code can now be removed.


nmattia (approved):
🎉

PR #1111, opened 2022-12-30
Add job to auto-update state machine

nmattia:
This adds a GitHub action job that checks for IC updates weekly, so that we always get the latest state machine executable.

### DIFF
--- a/.github/workflows/update-state-machine.yml
+++ b/.github/workflows/update-state-machine.yml
@@ -1,0 +1,88 @@
+# A GitHub Actions workflow that regularly creates a pull request to update the IC artefacts
+name: IC Artefacts Update
+
+on:
+  schedule:
+    # create a new pull request every monday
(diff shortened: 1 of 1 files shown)


.github/workflows/update-state-machine.yml
```diff
@@ -0,0 +1,88 @@
+# A GitHub Actions workflow that regularly creates a pull request to update the IC artefacts
+name: IC Artefacts Update
+
+on:
+  schedule:
+    # create a new pull request every monday
+    - cron:  '0 0 * * MON'
```
peterpeterparker, line 7: TIL `MON` can be used instead of numbers. Thanks, good to know.

peterpeterparker (approved):
LGTM / Rubber stamp (e.g. I don't know well `jq` or never build a `while` in a GH action)

PR #1115, opened 2023-01-05
Also update state machine download hint with latest ic commit sha

frederikrothenberger:
The framework module also contains the IC commit sha in an error message. This makes sure, that this message does not get out of sync with what's happening on CI.


nmattia (approved):
thanks!

PR #1117, opened 2023-01-05
Add some checks when using release API

nmattia:
The GitHub API has been returning some bad data, though not reproducibly. This adds some logging and aborts the rust update early (before writing the new toolchain version).


frederikrothenberger (approved):
Thanks!

PR #1118, opened 2023-01-05
Refactor CAPTCHA and improve UX

nmattia:
This refactors the captcha code a bit and changes the UX to clarify what is expected of the user.

In particular, the captcha page is now more reactive. `Chan`s are used, which are `asyncReplace`d in the template on state changes (see code for actual states).

In terms of UX, a new `retry` button is introduced and the page prompts the user to retry on failure.

Before:

https://user-images.githubusercontent.com/6930756/210792599-a0916f98-668b-4bbc-a414-837d99293bca.mov

After:

https://user-images.githubusercontent.com/6930756/210792651-f3c9224b-73a9-4f0b-92f0-135836ec3057.mov


peterpeterparker:
Since we do some UX improvements, would it be useful to disable autosuggestion for the input field?

From your video:

Capture d’écran 2023-01-05 à 15 00 01

peterpeterparker:
The layout shift when a msg with two lines is displayed, is there a way to fix this for all screens or will this happens not that often and therefore is a-ok?

Capture d’écran 2023-01-05 à 15 01 49

Capture d’écran 2023-01-05 à 15 01 30

src/frontend/src/flows/register/captcha.ts
```diff
@@ -1,180 +1,225 @@
...
+  // The "next" button behavior
+  const next = new Chan<((e: SubmitEvent) => void) | undefined>();
+  const nextDisabled = next.map((f) => f === undefined);
+
+  // The "retry" button behavior
+  const retry = new Chan<(() => void) | undefined>();
+  const retryDisabled = retry.map((f) => f === undefined);
+
+  // The various states the component can inhabit
+  type State =
```
peterpeterparker, line 47: looks like Rust no? isn't the type in JS/TS:

```
type State = {status: "requesting" | "prompting" | "verifying" | "bad"; challenge?: Challenge }
```

src/frontend/src/flows/register/captcha.ts
```diff
@@ -1,180 +1,225 @@
...
+    const challenge = await props.requestChallenge();
+    update({ status: "prompting", challenge });
+  };
+
+  // On verification, check the chars and either continue (on good challenge)
+  // or go to "bad" state
+  const doVerify = async (challenge: Challenge) => {
+    update({ status: "verifying" });
+    withRef(input, async (input) => {
+      const res = await props.verifyChallengeChars({
```
peterpeterparker, line 65: nth: I would find more readable to deconstruct the props. Here I searched for few seconds what was the "props" and where was that defined since it is not part of the function. If I would have just read the function name, I would have spared these seconds.

src/frontend/src/flows/register/captcha.ts
nmattia: no, because only one state (`prompting`) has a `challenge`

src/frontend/src/flows/register/captcha.ts
```diff
@@ -1,180 +1,225 @@
...
+    });
   });
 };
+
+// Returns a function that returns `first` on the first call,
+// and values returned by `f()` from the second call on.
+function precomputedFirst<T>(first: T, f: () => T): () => T {
+  let firstTime = true;
+
+  return () => {
```
peterpeterparker, line 217: nth: not sure, is the return type maybe missing here?

src/frontend/src/flows/register/captcha.ts
```diff
@@ -1,180 +1,225 @@
...
+    chars: string;
+    challenge: Challenge;
+  }) => Promise<T | typeof badChallenge>;
+  onContinue: (result: T) => void;
+}) => {
+  // We define a few Chans that are used to update the page in a
+  // reactive way; see template returned by this function
+
+  // The text and image shown
+  const text = new Chan<TemplateResult>();
```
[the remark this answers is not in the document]
nmattia: Was thinking the same; we _could_ send the `state` through a chan, and then have all the other values derived from a single chan (through `.map()`). I'll give it a try!

src/frontend/src/flows/register/captcha.ts
```diff
@@ -1,180 +1,225 @@
...
+  const update = (state: State) => {
+    switch (state.status) {
+      case "requesting":
+        text.send(html`<p>Generating new challenge...</p>`);
+        img.send(html`<div class="c-spinner">${spinner}</div> `);
+        next.send(undefined);
+        retry.send(undefined);
+        break;
+      case "prompting":
+        text.send(html`<p>Type the characters you see</p>`);
```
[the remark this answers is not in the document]
nmattia: That CAPTCHA really shouldn't be there at all, it was a stop-gap measure until People Parties...

src/frontend/src/flows/register/captcha.ts
nmattia: Can you clarify the question? Missing from where?

src/frontend/src/flows/register/captcha.ts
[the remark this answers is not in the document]
nmattia: Ok it's a little more complicated than I thought, may revisit in the future. WDYT?

src/frontend/src/flows/register/captcha.ts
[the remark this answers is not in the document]
peterpeterparker: Ok

src/frontend/src/flows/register/captcha.ts
peterpeterparker: I meant should it be following (adding `: T`)?

```
  return (): T => {
```

not sure...

src/frontend/src/flows/register/captcha.ts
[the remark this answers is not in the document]
peterpeterparker: Ok

peterpeterparker (approved):
LGTM, thanks

PR #1119, opened 2023-01-05
Refactor persistent state in preparation for archive pull integration

frederikrothenberger:
This PR refactors the archive state handling and prepares the persistent state for the additional configuration parameters required when the archive integration is switch to pull entries from II (instead of II pushing them to the archive).

This PR introduces changes that require a persistent state migration which can be triggered by setting the corresponding flag in the II init arguments.


src/internet_identity/internet_identity.did
```diff
@@ -140,14 +140,31 @@ type InternetIdentityStats = record {
...
+    // Configured value will be discarded for storage_layout_version < 6.
+    entries_buffer_limit: nat64;
+    // The maximum number of entries to be transferred to the archive per call.
+    // Currently unused: in preparation of switching the archive integration to pull.
+    // Configured value will be discarded for storage_layout_version < 6.
+    entries_fetch_limit: nat16;
+    // Polling interval to fetch new entries from II (in nanoseconds).
+    // Currently unused: in preparation of switching the archive integration to pull.
+    // Configured value will be discarded for storage_layout_version < 6.
+    polling_interval_ns: nat64;
```
nmattia, line 159: What value do you have in mind here? nanoseconds seems very precise

src/internet_identity/internet_identity.did
```diff
@@ -158,13 +175,15 @@ type InternetIdentityInit = record {
         nat64;
         nat64;
     };
-    // Set the module hash of the archive canister
-    archive_module_hash : opt blob;
+    // Archive config parameters to use for the next archive deployment.
+    archive_config: opt ArchiveConfig;
```
nmattia, line 179: what does "next archive deployment" mean?

Also, if `archive_config: null` then no archive is deployed right? (just double checking)

src/internet_identity/src/archive.rs
```diff
@@ -14,27 +14,40 @@ use internet_identity_interface::*;
...
-    NotCreated,
-    CreationInProgress(Timestamp), // timestamp when creation was initiated
-    Created(ArchiveData),
+    /// Not configured and not created.
+    NotConfigured,
+    /// Configured but not created.
+    Configured {
+        /// Archive related configuration
+        config: ArchiveConfig,
+    },
```
nmattia, line 31: What is configured? The II persistent state or an existing archive canister? Also I'm a little confused why this is needed

src/internet_identity/src/main.rs
```diff
@@ -221,19 +222,37 @@ fn post_upgrade(maybe_arg: Option<InternetIdentityInit>) {
...
+        if let Some(new_config) = arg.archive_config {
+            update_archive_config(new_config);
         }
         if let Some(cost) = arg.canister_creation_cycles_cost {
             state::persistent_state_mut(|persistent_state| {
                 persistent_state.canister_creation_cycles_cost = cost;
             })
         }
+        if let Some(flag) = arg.upgrade_persistent_state {
+            if flag {
```
nmattia, line 234: beware of boolean blindness with this naked boolean; also probably simpler to match on `Some(true)` directly

src/internet_identity/internet_identity.did
frederikrothenberger: I would suggest using an interval of 1 minute to start with.
I chose the `ns` because it is the general time resolution on the IC.

src/internet_identity/internet_identity.did
frederikrothenberger: This should explain that the configuration is only (fully) applied if the archive is redeployed. I.e. changing the polling interval on the archive side requires an archive deployment.
I will add a clarification.

No, changing the II config does _not_ trigger an archive deployment, regardless of whether `archive_config` is set or not.

src/internet_identity/src/archive.rs
frederikrothenberger: I switched around the types here to make it more consistent. I.e. having an archive now implies having a configuration as well.

And to represent (no config, no archive) and (config, no archive) I had to introduce a new state. `Configured` represents the (config, no archive) state.

nmattia: Sorry, the question wasn't clear. I was asked "by _configured_, do you mean that the archive is created and configured or that we have a config configured in the II canister", and I know understand you mean the latter.

nmattia (approved):

PR #1120, opened 2023-01-06
Show version info in browser console

nmattia:
This introduces a new script, `scripts/version`, that infers the version based on the git state. This version (a string) is then baked into the build for docker builds (`scripts/docker-build` and CI build).

The frontend uses this to display some version information in the browser console.


frederikrothenberger (approved):
Very nice, thanks!

src/frontend/src/index.ts
```diff
@@ -33,6 +34,25 @@ const readCanisterId = (): string => {
 };
 
 const init = async () => {
+  // Show version information for the curious programmer
+  try {
```
peterpeterparker, line 38: nth: extract a function - e.g. `printToConsole();`

src/frontend/src/index.ts
```diff
@@ -33,6 +34,25 @@ const readCanisterId = (): string => {
...
+    console.log("Welcome to Internet Identity!");
+    console.log(
+      "The code can be found here: https://github.com/dfinity/internet-identity"
+    );
+    console.log(
+      `https://github.com/dfinity/internet-identity/commit/${version.commit}`
+    );
+    if (version.release !== "") {
```
peterpeterparker, line 46: here you check against `!== ""` but in `version.ts` it sets `undefined`. is that expected?

src/frontend/src/version.ts
```diff
@@ -0,0 +1,8 @@
+/** parse the II_VERSION, see ./scripts/version for more information */
+export const versionString = process.env.II_VERSION ?? "";
+export const versionList = versionString.split("\n");
+export const version = {
```
peterpeterparker, line 4: no type?

e.g.

```

btw. there is no type for version?

```
interface Version {
   release?: string,
   dirty: boolean;
   etc.
```

File: src/frontend/src/index.ts (hunk 2 above)
Comment (nmattia) [reply src/frontend/src/index.ts:46]: last minute change, forgot to update here. Thanks a lot!

File: src/frontend/src/index.ts (hunk 1 above)
Comment (nmattia) [reply src/frontend/src/index.ts:38]: `printVersion()`

File: src/frontend/src/index.ts (hunk 2 above)
Comment (nmattia) [reply src/frontend/src/index.ts:46]: fixed

File: src/frontend/src/version.ts (hunk 3 above)
Comment (nmattia) [reply src/frontend/src/version.ts:4]: Added!

File: src/frontend/src/index.ts [hunk 4]
```diff
@@ -32,7 +33,30 @@ const readCanisterId = (): string => {
...
+  if (version.dirty) {
+    console.warn("This version is dirty");
+  }
+};
+
 const init = async () => {
+  try {
+    printVersion();
+  } catch (e) {
+    console.warn("Error when printing version information:", e);
```
peterpeterparker, line 57: not sure, if printVersion fails it's most probably for some weird reason because console.something would fails so here it would probably fails again no? I would just silent it or remove the try catch

peterpeterparker (approved):
LGTM, minor comment

src/frontend/src/index.ts
nmattia: Yeah but if for _some_ reason e.g. `version.commit` cannot be printed or something, I don't want the whole app to fail because of this... just being extra cautious here since if anything breaks we wouldn't even show an error message

peterpeterparker: Sounds good

PR #1122, opened 2023-01-09
Store IC commit in .ic-commit

nmattia:
Similar to how we store and update the node version, this introduces a new file `.ic-commit` which is used by CI to fetch the correct state machine tests executable given the IC repo commit, as well as update the commit.

This prevents some token authorization concerns that surface when editing workflow files in a workflow.


frederikrothenberger (approved):
Thanks, could you also update the error message in `framework::env()`?

PR #1127, opened 2023-01-09
Support for the archive pulling entries

frederikrothenberger:
This PR adds support for the archive to pull the entries from II (instead of II pushing them to the archive). This makes the archive integration robust againtst rejected messages.

The pull integration has to be enabled explicitly and is only supported for layout versions 6 and higher.

## Overview of the changes
* Archive
  * Use the `cdk` `timers` feature to periodically trigger fetching entries from II.
  * Add metrics related to pulling entries: last successful fetch, how many entries are fetched
  * Provide information about failures when fetching or acknowledging entries.
* Internet Identity
  * Expand the II interface with canister calls that allow the archive to pull entries.
  * Allow configuration of how the entries should be transferred to the archive (push/pull).
  * Add metrics related to pulling entries: how many entries are buffered on the II side ready to be pulled.
  * Change the archive deployment code to allow a redeployment of the same archive wasm if the configuration is different (i.e. to change the polling interval).


src/canister_tests/src/framework.rs
```diff
@@ -420,20 +425,35 @@ frame-ancestors 'none';$"
...
+            url: "/metrics".to_string(),
+            headers: vec![],
+            body: ByteBuf::new(),
+        },
+    )
+    .expect("HTTP request to /metrics failed");
+    String::from_utf8_lossy(&response.body).to_string()
+}
+
+pub fn parse_metric(body: &str, metric: &str) -> (f64, SystemTime) {
```
frederikrothenberger, line 443: There is some refactoring of test infrastructure (with regard to metrics) in here too. I could split that out, if you'd like.

src/archive/src/main.rs
```diff
@@ -229,8 +246,11 @@ impl BoundedStorable for AnchorIndexKey {
     }
 }
 
+/// This method is kept for legacy compatibility and easier testability of the archive.
```
nmattia, line 249: But I'm guessing we'll remove it eventually, right? If we're happy with the poll we have no reason to go back to push

src/archive/src/main.rs
```diff
@@ -239,6 +259,84 @@ fn write_entry(anchor: AnchorNumber, timestamp: Timestamp, entry: ByteBuf) {
...
+    let ii_canister = with_config(|config| config.ii_canister);
+    let fetch_result: CallResult<(Vec<BufferedEntry>,)> =
+        call(ii_canister, "fetch_entries", ()).await;
+
+    let entries = match fetch_result {
+        Ok((entries,)) => entries,
+        Err((code, message)) => {
+            // failed to fetch entries --> store failure information and exit early
+            store_call_error(CallErrorInfo {
+                time: time(),
```
nmattia, line 277: nit: how about storing the time when the call was made? if the call e.g. times out this value may be misleading

src/archive/src/main.rs
```diff
@@ -239,6 +259,84 @@ fn write_entry(anchor: AnchorNumber, timestamp: Timestamp, entry: ByteBuf) {
...
+                canister: ii_canister,
+                method: "fetch_entries".to_string(),
```
nmattia, line 279: I can see this and the method used in `call` above getting out of sync very quickly

src/archive/src/main.rs
```diff
@@ -239,6 +259,84 @@ fn write_entry(anchor: AnchorNumber, timestamp: Timestamp, entry: ByteBuf) {
...
+                argument: ByteBuf::from(candid::encode_one(()).unwrap()),
```
nmattia, line 280: Is there ever not `()`?

src/archive/src/main.rs
```diff
@@ -476,18 +591,47 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
+async fn status() -> ArchiveStatus {
     let canister_id = id();
-    let (status,) = canister_status(CanisterIdRecord { canister_id })
+    let (canister_status,) = canister_status(CanisterIdRecord { canister_id })
         .await
         .expect("failed to retrieve canister status");
-    status
+    let config = with_config(|config| ArchiveInit {
+        ii_canister: config.ii_canister,
+        max_entries_per_call: config.max_entries_per_call,
```
nmattia, line 622: what does `max_entries_per_call` correspond to now?

src/archive/src/main.rs
```diff
@@ -239,6 +259,84 @@ fn write_entry(anchor: AnchorNumber, timestamp: Timestamp, entry: ByteBuf) {
             ))
         }
     });
+    write_entry_internal(anchor_number, timestamp, entry)
+}
+
+/// Fetches, archives and acknowledges a batch of entries.
+/// *Note:* Must be written in a way that nothing breaks on overlapping executions of [fetch_entries].
```
nmattia, line 266: what do you mean by this?

> in a way that nothing breaks on overlapping executions of [fetch_entries]

Can you clarify why we'd have overlapping executions? Also, why not just protect this with a mutex or lock?

I'm guessing this stems from user the Timer API? Can you shed some light on how this work, also w.r.t. the "last successful call" metric?

src/archive/src/main.rs
```diff
@@ -239,6 +259,84 @@ fn write_entry(anchor: AnchorNumber, timestamp: Timestamp, entry: ByteBuf) {
...
+            })
+        });
+        // nothing to do --> exit early
+        return;
+    }
+
+    let expected_seq_nr = with_log(|log| log.len() as u64);
+    let entries_count = entries.len();
+    let lowest_seq_nr = entries.first().unwrap().sequence_number;
+    let highest_seq_nr = entries.last().unwrap().sequence_number;
```
nmattia, line 303: note: implicit assumption that entries are sorted

src/archive/src/main.rs
```diff
@@ -239,6 +259,84 @@ fn write_entry(anchor: AnchorNumber, timestamp: Timestamp, entry: ByteBuf) {
...
+                method: "fetch_entries".to_string(),
+                argument: ByteBuf::from(candid::encode_one(()).unwrap()),
+                rejection_code: code as i32,
+                message,
+            });
+            return;
+        }
+    };
+
+    if entries.is_empty() {
```
nmattia, line 288: note: you make an assumption below about `entries` not being empty, though if someone decided to simplify the logic by not returning early and simply removed this block here (`if entries.is_empty()...`) then the assumption wouldn't hold anymore

src/archive/src/main.rs
```diff
@@ -239,6 +259,84 @@ fn write_entry(anchor: AnchorNumber, timestamp: Timestamp, entry: ByteBuf) {
...
+    // acknowledge message got rejected.
+    // In such cases just the acknowledge is sent again.
+    if lowest_seq_nr >= expected_seq_nr {
+        entries
+            .into_iter()
+            .for_each(|e| write_entry_internal(e.anchor_number, e.timestamp, e.entry));
+    }
+
+    let fetch_result: CallResult<()> =
+        call(ii_canister, "acknowledge_entries", (highest_seq_nr,)).await;
```
nmattia, line 316: What do we gain from acknowledging entries?

As far as I understand, II will always have room for a full buffer. Why not simply use it as a circular buffer? In which case do we benefit from having the archive acknowledge the entries it has read? Also, by keeping all entries around on II side (for, say, a couple of hours) we would open the door to other archiving mechanisms (for instance, off-chain) and wouldn't limit it to only this canister

src/archive/src/main.rs
frederikrothenberger: Good point.

src/archive/src/main.rs
frederikrothenberger: `max_entries_per_call` is unchanged still the number of entries _clients_ of the archive can receive per call.

The number of entries fetched is controlled on the II side.

src/archive/src/main.rs
frederikrothenberger: Added explicit sorting.

src/archive/src/main.rs
frederikrothenberger: Ok, if checked with the message routing team and apparently it is possible for responses to get reordered. This means that we could potentially benefit from having overlapping calls because the call started later could finish before the other call.

The "last successful call" metric would always show the timestamp of the last completed invocation of the poll. From this we could infer a healthy system if the timestamp changes frequently enough and is reasonably recent.

PR #1128, opened 2023-01-09
Download prebuilt ic-cdk-optimizer

nmattia:
This speeds up the base docker image build by using a prebuilt version of ic-cdk-optimizer.

It also improves the bootstrap script's UX by adding confirmation to a few steps.

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -53,7 +53,6 @@ then
             optimizer_url="$release_url/ic-cdk-optimizer-0.3.5-x86_64-apple-darwin.tar.gz"
             ;;
         Linux)
-            # At the time of writing, this is indeed 0.3.4 but packaged as 0.3.5 (sigh)
             optimizer_url="$release_url/ic-cdk-optimizer-0.3.5-x86_64-unknown-linux-gnu.tar.gz"
             ;;
         *)
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -16,15 +16,52 @@ function run() {
     fi
 }
 
-rust_version=$(cat ./rust-toolchain.toml | sed -n 's/^channel[[:space:]]*=[[:space:]]"\(.*\)"/\1/p')
-echo "using rust version '$rust_version'"
+# Ask user for confirmation for action "$1" before going further
+# (skipped on CI and non-interactive shells in general)
+function confirm() {
+    if [[ -n ${CI:-} || -z ${PS1:-} ]]; then return; fi
+    read -p "$1 [y/n]" -n 1 -r
+    if [[ ! $REPLY =~ ^[Yy]$ ]]
+    then
+        exit 1
+    fi
+}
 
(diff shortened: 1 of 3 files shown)


scripts/bootstrap
```diff
@@ -16,15 +16,51 @@ function run() {
...
+echo "Installing rust '$rust_version'"
 # here we set the toolchain to 'none' and rustup will pick up on ./rust-toolchain.toml
 run curl --fail https://sh.rustup.rs -sSf | run sh -s -- -y --default-toolchain "none" --no-modify-path
 
 echo "looking for ic-cdk-optimizer 0.3.4"
 if [[ ! "$(command -v ic-cdk-optimizer)" || "$(ic-cdk-optimizer --version)" != "ic-cdk-optimizer 0.3.4" ]]
 then
-    echo "installing ic-cdk-optimizer 0.3.4"
-    run cargo install ic-cdk-optimizer --version 0.3.4
+    bin_dir="$HOME/.local/bin"
```
frederikrothenberger, line 40: Shouldn't this be `/usr/local/bin` on macos?

nmattia: Why particularly on macOS? in general I think it's not really a good idea to install anything globally; at least `$HOME/.local` is per-user

frederikrothenberger: I thought you wanted to use a standard directory. On my Mac I don't have the `$HOME/.local/bin` directory and also not added to the path. I.e. this adds more friction for users who want to run the script.

But given it's `bootstrap`, it might not be that important.

scripts/bootstrap
```diff
@@ -16,15 +16,51 @@ function run() {
     fi
 }
 
+# Ask user for confirmation for action "$1" before going further
+# (skipped on CI and non-interactive shells in general)
+function confirm() {
+    if [[ -n ${CI:-} || -z ${PS1:-} ]]; then return; fi
```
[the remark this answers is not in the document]
nmattia: that's really odd. How did you test this? What are the values of `$CI` and `$PS1`?

scripts/bootstrap
nmattia: It's not "standard" standard, but it's the directory python and a lot of package managers use; I'm happy to go with something else if you have suggestions (although nothing global)

scripts/bootstrap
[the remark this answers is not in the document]
frederikrothenberger: I just ran `scripts/bootstrap`.

The values for `$CI` is always empty and for `PS1` it is either `%{%f%b%k%}$(build_prompt)` or `\s-\v\$` (zsh and bash).

scripts/bootstrap
frederikrothenberger: Ah ok. Yeah, I think it's ok.

scripts/bootstrap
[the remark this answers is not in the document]
nmattia: Fixed! Forgot bash would not set `$PS1` when run directly

frederikrothenberger (approved):

nmattia:
The `ic-cdk-optimizer` release is not very stable. I think we might be better off sticking with our current caching solution.

PR #1129, opened 2023-01-09
Update state machine (client) to support additional commands

frederikrothenberger:
This PR updates the state machine binary commit and extends the state machine client crate with additional methods to start, stop and delete canisters as well as executing canister timer methods.

### DIFF
--- a/src/state_machine_client/src/lib.rs
+++ b/src/state_machine_client/src/lib.rs
@@ -25,6 +25,14 @@ pub enum Request {
     AddCycles(AddCyclesArg),
     SetStableMemory(SetStableMemoryArg),
     ReadStableMemory(RawCanisterId),
+    Tick,
+    RunUntilCompletion(RunUntilCompletionArg),
+}
+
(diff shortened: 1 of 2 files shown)


src/state_machine_client/src/lib.rs
```diff
@@ -189,6 +224,16 @@ impl StateMachine {
...
 
+    pub fn tick(&self) {
+        self.call_state_machine(Request::Tick)
+    }
+
+    pub fn run_until_completion(&self, max_ticks: u64) {
+        self.call_state_machine(Request::RunUntilCompletion(RunUntilCompletionArg {
+            max_ticks,
+        }))
+    }
```
nmattia, line 235: I'm confused, is this used anywhere?

frederikrothenberger: Not yet. But it will be used by the integration test for the archive pull integration.

nmattia (approved):

PR #1133, opened 2023-01-10
Use obj as arguments for register

nmattia:
This is a very minor refactoring which changes the typescript `register()` method to take in named parameters, to avoid confusion.


peterpeterparker (approved):
Nice, thx

PR #1134, opened 2023-01-10
Make expiration periods more readable

nmattia:
This changes the way expiration periods are defined:

* Comments are made more meaningful (instead of repeating the actual value, which may get out of sync)
* Some helpers (minutes, hours, days) are introduced to make the values of expiration periods clear
* The signature expiration period is not called "default" anymore, since it's the only possible expiratin period


frederikrothenberger (approved):
Thanks!

PR #1135, opened 2023-01-10
change the order of items on the captcha screen

meodai:
To make it clear to the user where to type, I've chosen to put the text on top of the input field. I think it was not too clear before where to type. Otherwise! Nice Job @nmattia !

Screenshot 2023-01-10 at 15 27 53


meodai:
I've added an area and class for the error messages and I added an error class on the input.

I did not manage make it only show up if it fails. Would be great if you could show me how this `Chan` utility works.

Screenshot 2023-01-13 at 14 28 10

PR #1136, opened 2023-01-10
Update staging deployment

frederikrothenberger:
In #1119 the init argument of II was changed which is relevant for updating the archive wasm hash. This PR updates the deployment script to use the new init argument type.


nmattia (approved):
Good catch

PR #1137, opened 2023-01-10
Proof of concept for single WebAuthn prompt

nmattia:
This is a proof of concept for skipping the second WebAuthn prompt when registering a new anchor.

In II's current state, the second prompt is necessary because we cannot both create WebAuthn credentials _and_ sign a delegation in a single prompt. This change here introduces a new key-pair created in the browser during registration, which is used for the first authentication. The public key is kept in the canister in-memory, with an expiration period equal to the default expiration for delegations.

This is a straw man proposal and everything is up for discussion.


nmattia:
Closing in favor of #1171

PR #1138, opened 2023-01-11
Regenerate typescript defs

frederikrothenberger:
This PR adds the result of `npm run generate` to include the archive pull integration typing changes.


src/frontend/generated/internet_identity_types.d.ts
```diff
@@ -1,4 +1,6 @@
 import type { Principal } from '@dfinity/principal';
+import type { ActorMethod } from '@dfinity/agent';
+
 export type AddTentativeDeviceResponse = {
```
peterpeterparker, line 4: Randomly noticing this PR.

FYI the extension of the files seems incorrect to me here. It's either - commonly at least - a typescript definition file - `.d.ts` and no `export` within - or a typescript file - `.ts` with `export`.

frederikrothenberger: Yeah, something is wonky anyway. I need to check with @nmattia why I get _this_ particular diff... 🤔

peterpeterparker: Nevermind my comment, just noticed that's actually the standard on the IC ...

PR #1139, opened 2023-01-12
Add features to browser message

nmattia:
This updates the "version" message browser console log (now a "dev" message) to show a warning if any features were enabled (and which).

### DIFF
--- a/src/frontend/src/index.ts
+++ b/src/frontend/src/index.ts
@@ -51,11 +51,10 @@ const printDevMessage = () => {
   }
 
   if (anyFeatures()) {
-    const featuresRec: Record<string, boolean> = features;
     const message = `
 Some features are enabled:
-${Object.keys(featuresRec)
-  .map((k) => ` - ${k}: ${featuresRec[k]}`)
+${Object.entries(features)
+  .map(([k, v]) => ` - ${k}: ${v}`)
   .join("\n")}
 see more at https://github.com/dfinity/internet-identity#features
       `;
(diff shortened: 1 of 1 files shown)


src/frontend/src/index.ts
```diff
@@ -10,6 +10,7 @@ import { checkRequiredFeatures } from "./utils/featureDetection";
 import { showWarningIfNecessary } from "./banner";
 import { displayError } from "./components/displayError";
 import { Connection } from "./utils/iiConnection";
+import { anyFeatures, features } from "./features";
```
peterpeterparker, line 13: nth: following of course an opinionated feedback but, since `any` is the keyword that must absolutely be avoided, I would rather like a variable name that does not contain it 😁

src/frontend/src/index.ts
```diff
@@ -48,11 +49,23 @@ const printVersion = () => {
   if (version.dirty) {
     console.warn("This version is dirty");
   }
+
+  if (anyFeatures()) {
+    const featuresRec: Record<string, boolean> = features;
```
peterpeterparker, line 54: not sure wht's the purpose of this `featuresRec`?

src/frontend/src/index.ts
nmattia: Gotcha! Will know better next time, although the variable was introduced a few months ago (i.e. not in this PR).

src/frontend/src/index.ts
```diff
@@ -48,11 +49,23 @@ const printVersion = () => {
...
+    const message = `
+Some features are enabled:
+${Object.keys(featuresRec)
```
peterpeterparker, line 57: nth: would `Object.entries` fits better?

```
Object.entries(features).map(([key, value]) => ` - ${key}: ${value}`)
```

nmattia: it would! thanks!

src/frontend/src/index.ts
nmattia: none 🙃

peterpeterparker (approved):
LGTM

PR #1140, opened 2023-01-12
Automatically commit generated type interfaces

nmattia:
This ensures that type interfaces between rust and typescript are kept up to date by committing generated changes automatically.


frederikrothenberger (approved):
Thanks a lot! 👍

PR #1141, opened 2023-01-12
Add method to atomically replace one device with another

frederikrothenberger:
This PR adds a new method to atomically replace a device with another.
It also introduces a new archive operation variant to distinguish
it from doing remove and add separately.


src/internet_identity/internet_identity.did
```diff
@@ -242,6 +242,8 @@ service : (opt InternetIdentityInit) -> {
     register : (DeviceData, ChallengeResult) -> (RegisterResponse);
     add : (UserNumber, DeviceData) -> ();
     update : (UserNumber, DeviceKey, DeviceData) -> ();
+    // Atomically replace device matching the device key with the new device data
+    swap : (UserNumber, DeviceKey, DeviceData) -> ();
```
nmattia, line 246: I think "replace" is more accurate (as also suggested by your comment here) since "swap" suggests device 1 will replace device 2 and device 2 will replace device 1

src/internet_identity/src/anchor_management.rs
```diff
@@ -109,6 +109,36 @@ pub fn update(user_number: AnchorNumber, device_key: DeviceKey, device_data: Dev
     );
 }
 
+pub fn replace(anchor_number: AnchorNumber, device_key: DeviceKey, device_data: DeviceData) {
+    let mut anchor = state::anchor(anchor_number);
+    // must be called before the first await because it requires caller()
```
nmattia, line 114: note: function is not async

src/internet_identity/src/anchor_management.rs
```diff
@@ -109,6 +109,36 @@ pub fn update(user_number: AnchorNumber, device_key: DeviceKey, device_data: Dev
...
+    trap_if_not_authenticated(&anchor);
+
+    anchor.remove_device(&device_key).unwrap_or_else(|err| {
+        trap(&format!(
+            "failed to replace device of anchor {}: {}",
+            anchor_number, err
+        ))
+    });
+    let new_device = Device::from(device_data);
```
nmattia, line 123: I've asked this a thousand times already and apologize for that, but can you confirm that a canister call is pretty much a transaction? i.e. if this line traps then all changes to orthogonal persistence and stable memory will be rolled back and the `remove_device` call above will be rollbacked too?

src/internet_identity/src/anchor_management.rs
```diff
@@ -109,6 +109,36 @@ pub fn update(user_number: AnchorNumber, device_key: DeviceKey, device_data: Dev
...
+            "failed to replace device of anchor {}: {}",
+            anchor_number, err
+        ))
+    });
+    write_anchor(anchor_number, anchor);
+    archive_operation(
+        anchor_number,
+        caller(),
+        Operation::ReplaceDevice {
+            device: device_key,
```
nmattia, line 135: note: I didn't immediately realize this was the "old" device. Since the other field is `new_device` is expected `old_device` or something

nmattia (approved):

src/internet_identity/src/anchor_management.rs
frederikrothenberger: Yes, it is. All heap changes and stable memory changes will be rolled back.

src/internet_identity/src/anchor_management.rs
frederikrothenberger: True, removed.

PR #1143, opened 2023-01-12
Replace state machine client lib with external dependency

frederikrothenberger:
The state machine client lib has been opensourced and published on crates.io. Therefore we remove our local copy.


nmattia (approved):
🥳

PR #1145, opened 2023-01-13
Refactor register/finish

nmattia:
This modernizes the code of the registration wrap up, in the style of other new components (like captcha.ts): more structured, more idiomatic `lit` constructs, object parameters and more.

### DIFF
--- a/src/frontend/src/flows/register/finish.ts
+++ b/src/frontend/src/flows/register/finish.ts
@@ -1,7 +1,13 @@
 import { html, render } from "lit-html";
 import { warnBox } from "../../components/warnBox";
 
-const pageContent = (userNumber: bigint) => html`
+export const displayUserNumberTemplate = ({
+  onContinue,
+  userNumber,
+}: {
+  onContinue: () => void;
+  userNumber: bigint;
+}) => html`
   <div class="l-container c-card c-card--highlight">
     <hgroup>
       <h1 class="t-title t-title--main">Congratulations!</h1>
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/register/finish.ts
```diff
@@ -18,21 +24,28 @@ const pageContent = (userNumber: bigint) => html`
       ${userNumber}
     </div>
     <div class="l-stack">
-      <button id="displayUserContinue" class="c-button">Continue</button>
+      <button
+        @click=${() => onContinue()}
```
peterpeterparker, line 28: nth: maybe you can even simplify here with `@click=${onContinue}` - not sure, to be tested

src/frontend/src/flows/register/finish.ts
```diff
@@ -18,21 +24,28 @@ const pageContent = (userNumber: bigint) => html`
...
     </div>
   </div>
 `;
 
-export const displayUserNumber = async (userNumber: bigint): Promise<void> => {
-  const container = document.getElementById("pageContent") as HTMLElement;
-  render(pageContent(userNumber), container);
-  return init();
+export const displayUserNumberPage = (
+  props: Parameters<typeof displayUserNumberTemplate>[0],
```
peterpeterparker, line 39: actually never seen - I think - `[0]` used to express a type, does that work? nice to know

peterpeterparker (approved):
LGTM, minor comments

src/frontend/src/flows/register/finish.ts
nmattia: yep, in this case that takes the type of the first param!

peterpeterparker: cool

PR #1147, opened 2023-01-13
Refactor promptUserNumber

nmattia:
This updates `promptUserNumber` to be more structured like other pages (captcha, addRemoteDevice, etc) and use more idiomatic lit and ts constructs.

### DIFF
--- a/src/frontend/src/flows/promptUserNumber.ts
+++ b/src/frontend/src/flows/promptUserNumber.ts
@@ -2,20 +2,21 @@ import { html, render, TemplateResult } from "lit-html";
 import { Ref, ref, createRef } from "lit-html/directives/ref.js";
 import { mkAnchorInput } from "../components/anchorInput";
 
-const pageContent = (
-  title: string,
-  userNumber: bigint | null,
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/promptUserNumber.ts
```diff
@@ -2,20 +2,21 @@ import { html, render, TemplateResult } from "lit-html";
...
-  callbacks: { onContinue: (ret: bigint) => void; onCancel: () => void }
-): { template: TemplateResult } => {
+const promptUserNumberTemplate = ({
+  title,
+  userNumber,
+  onContinue: onSubmit,
+  onCancel,
+}: {
+  title: string;
+  userNumber?: bigint;
```
peterpeterparker, line 12: the `userNumber` used to be `biging` or `null` but becomes `bigint` or `undefined`. that's fine? no side effect?

peterpeterparker (approved):
LGTM, minor comment

src/frontend/src/flows/promptUserNumber.ts
nmattia: Side effects taken care of, see useRecovery and showcase!

PR #1148, opened 2023-01-13
Drop 'addRemoteDevice'

nmattia:
This was an intermediary screen that was only replicating the `promptUserNumber` screen. Instead, the `promptUserNumber` is moved to `components/` and adapted to be used in lieu of `welcomeView/index`'s `addRemoteDevice`.


src/frontend/src/components/authenticateBox.ts
```diff
@@ -215,3 +216,20 @@ const page = (slot: TemplateResult) => {
...
+// adding a device to that anchor.
+const addRemoteDevice = async (connection: Connection) => {
+  const userNumber = await promptUserNumber({
+    title: "New Device",
+    message:
+      "Please provide the Identity Anchor to which you want to add your device.",
+  });
+
+  if (userNumber === "canceled") {
+    // TODO L2-309: do this without reload
```
nmattia, line 230: Just copied over from the old code, will work on removing this next

peterpeterparker (approved):
LGTM

PR #1149, opened 2023-01-13
Clean up Chan

nmattia:
This improves the `Chan` structure in two ways:

* Fixes a bug where `undefined` values sent through the `Chan` would be dropped (due to bad handling of return value of `.shift()`
* The `.map()` is made not async anymore, making it clear that listeners are added synchronously (although processing of values is still async)


peterpeterparker (approved):
Sounds good.

Note: in the automatic screenshots the "Next" button has become "disabled", I assume it is the expected behavior/fix

nmattia:
>  I assume it is the expected behavior/fix

It is!

PR #1152, opened 2023-01-16
Slight Design Refresh

meodai:
After talking with Maurice, we did a few subtle design changes.
Future changes will be easier to do, because of the wrapper. 

On a technical level the main change is the introduction of `mainWindow.` wrapper.
It centralizes the content around each content.

From a design perspective a few things changes:

- We now use Inter as the main font (instead of Montserrat)
- The Inputs and buttons are not round anymore (it did not look nice as soon as there was more than one button on a page)
- Warning card can now have additional classes (mainly used to add some space on top of the warnings)
- The main card is now layered (the footer is on its own level visually)
- Lots of border-radius were set manually, we now use a variable `var(--vs-border-radius);`
- The logo is now a bit smaller and part of the `mainWindow`


meodai:
@nmattia I fixed all e2e texts I text, but I fait to understand why this fails, since the title seams to be there. Is there an ID missing?

src/frontend/src/styles/main.css
```diff
@@ -753,7 +809,7 @@ a:hover,
 }
 
 .c-input--rounded {
-  border-radius: 999999px;
+  /*border-radius: 999999px;*/
```
nmattia, line 812: remove?

src/frontend/src/components/mainWindow.ts
```diff
@@ -0,0 +1,53 @@
...
+  showFooter = true,
+  showLogo = true,
+  isWideContainer = false,
+  additionalContainerClasses = [],
+}: mainWindowProps): TemplateResult => {
+  const containerClasses = ["l-container"];
+  if (isWideContainer === true) {
+    containerClasses.push("l-container--wide");
+  }
+  if (additionalContainerClasses.length > 0) {
```
nmattia, line 38: I think you can skip the condition here, can't hurt to push an empty array

src/frontend/src/components/mainWindow.ts
```diff
@@ -0,0 +1,53 @@
+import { html, TemplateResult } from "lit-html";
+import { ifDefined } from "lit/directives/if-defined.js";
+import { icLogo } from "./icons";
+import { footer } from "./footer";
+
+export type mainWindowProps = {
+  slot: TemplateResult;
+  uid?: string;
```
nmattia, line 8:
```suggestion
  id?: string;
```
nit: `uid` is not used anywhere else in the codebase

src/frontend/src/components/mainWindow.ts
```diff
@@ -0,0 +1,53 @@
+import { html, TemplateResult } from "lit-html";
+import { ifDefined } from "lit/directives/if-defined.js";
+import { icLogo } from "./icons";
+import { footer } from "./footer";
+
+export type mainWindowProps = {
```
nmattia, line 6: Just a thought: for readability (and to avoid going back and forth between this type and the function definition) you could just write the type in the function definition; I don't think it needs to be exported. If you do decide to keep the type here please capitalize the first letter for consistency with other types!

src/frontend/src/flows/about/index.ts
```diff
@@ -1,10 +1,12 @@
 import { html, render } from "lit-html";
 import { hydrate } from "lit-html/experimental-hydrate.js";
 import { compatibilityData } from "../../components/compatibilityChart";
+import { mainWindow } from "../../components/mainWindow";
 
 // The About page
-export const pageContent = html`
-  <div class="l-container l-container--wide c-card c-card--highlight">
+export const pageContent = mainWindow({
```
nmattia, line 7: Wrapping all the functions (here and elsewhere) with `mainWindow` breaks git and makes the diff hard to read and look huge. How about something like this?

```
export const pageContent = mainWindow({ isWideContainer: true, slot: innerPageContent });

const innerPageContent = /* same as before */;
```

It would also clarify what is specific to this page and what isn't

frederikrothenberger:
> @nmattia I fixed all e2e texts I text, but I fait to understand why this fails, since the title seams to be there. Is there an ID missing?
> 
> ```js
>   async waitForDisplay(): Promise<void> {
>     await this.browser
>       .$("//h1[string()='Anchor Management']")
>       .waitForDisplayed({ timeout: 10_000 });
>   }
> ```
> 
> @frederikrothenberger I mention you since you showed up on the test's log :D

So if you look at the screenshots it took when it failed, my best guess is that it fails because the title is not actually visible due to the page being scrolled down:

Register_new_identity_and_add_additional_remote_device

Register_new_identity_and_add_additional_remote_device_starting_on_new_device

meodai:
@frederikrothenberger thanks so much, I had no idea it actually had to be in view! This explain quite a lot.

src/frontend/src/components/authenticateBox.ts
```diff
@@ -203,16 +202,11 @@ export const authenticate = async (
 
 // Wrap the template with header & footer and render the page
 const page = (slot: TemplateResult) => {
-  const template = html`
-    <div class="l-container c-card c-card--highlight">
-      <!-- The title is hidden but used for accessibility -->
+  const template = mainWindow({
+    slot: html` <!-- The title is hidden but used for accessibility -->
```
nmattia, line 206: is this still relevant? alternatively, shouldn't this be applied to all pages or no pages?

src/frontend/src/components/mainWindow.ts
```diff
@@ -0,0 +1,53 @@
...
+ * main window template
+ *
+ * To avoid having to repeat the same structure in every page,
+ * we use this component to wrap the content of each page.
+ * This way, we can change the structure of the main window
+ * in one place.
+ *
+ * Before it was just a DIV with a class, but now its a more complex
+ * component that includes the logo, the footer, and the container.
+ *
```
nmattia, line 16:
```suggestion
 * It is a component that includes the logo, the footer, and the container.
 *
```

After this is merged, no one will care what it was before

src/frontend/src/components/mainWindow.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export const mainWindow = ({
+  slot,
+  id,
+  showFooter = true,
+  showLogo = true,
```
nmattia, line 22: Should these two really be `true` by default? it looks like more often than not they're explicitly set to `false`

nmattia (commented):
couple comments; also, looks like the chasm took a hit:

authorizePickAltOpen

src/frontend/src/components/authenticateBox.ts
meodai: I don't think so only the ones that don't have title in the content

meodai:
> couple comments; also, looks like the chasm took a hit:
> 
> authorizePickAltOpen

I had seen that, but I thought it looked cool

nmattia (approved):
LGTM!

PR #1154, opened 2023-01-17
Disable showcase workflow on non-main

nmattia:
This disables the entire showcase deployment when the branch is not `main`. Previously only the last step was disabled which failed the job on PRs; the assumption here is that since the last step was disabled, the `environment` could not be set from its `.outputs`.


peterpeterparker (approved):
I ❤️ devops

PR #1156, opened 2023-01-19
Shortens / Changes lots of the texts

meodai:
Requested by Mary: https://docs.google.com/document/d/1nxeo_PU1eKmfDzrZyGL_36_7yeArYQirrW8WJXMT3LU/edit?usp=sharing


meodai:
@nmattia there is a list of things was unable to do:

## FAQ

Mary want us to get the contents from [here](https://support.dfinity.org/hc/en-us/sections/8730568843412-Internet-Identity). Would we have a way to pull them from there? Because it looks like they change from time to time. Or should we just link to that page?

## Delete Device / Recovery key

We need to change some text on the confirmation modal:
Screenshot 2023-01-19 at 15 33 50
But it looks like there is quite a lot of logic in that component. Can you help me with that one? 

## Unknown  Domain Warning
Screenshot 2023-01-19 at 15 35 02

I am not sure under what circumstances this occurs, can you tell me if its ok to change it to what is requested?

meodai:
@nmattia the e2e test fails that looks for "Manage your Anchor" I had adapted the title and the text. But it sill fails, it said "Anchor Management" before

meodai:
Regarding the FAQ: Mary thinks we can link to the external FAQ anyway. Is this an idea you support?

meodai:
@nmattia I made the change in: ad70376c917569b429d08249f8663b2cb80a101b if its ok for you, you can remove our local FAQ page. (I was not sure how to actually do that :D )

meodai:
the "Manage your Anchor" test is still failing I have no idea why

src/frontend/src/components/compatibilityChart.ts
```diff
@@ -1,7 +1,16 @@
 import { html } from "lit-html";
 
 export const compatibilityData = {
-  note: "Browser support for WebAuthentication is constantly evolving. Your preferred browser may not support WebAuthentication, or may only support it using a security key. If you run into issues, please try again with one of our recommended browsers.",
+  note: `Internet Identity leverages the widely used secure web2 authentication
+      framework known as WebAuthentication (WebAuthn). This way, you can use
```
marydwyer, line 5:
```suggestion
      framework known as Web Authentication (WebAuthn). This way, you can use
```

nmattia (approved):

PR #1158, opened 2023-01-19
Make showVerificationCode more lit-idiomatic

nmattia:
This updates the screen showing the verification code to be more idiomatic. This also updates the logic and countdown to be a bit more resilient. In particular:

* The `poll` function wraps IC calls in try/catch to avoid erroring out and dying and losing the timeout handle
* The page template is rewritten to use idiomatic lit
* A new countdown is introduced which generates async values for the remaining time, separating the concerns between counting down and HTML handling (also meaning we can now show the timer value in the showcase)
* The `poll` function is now implemented as a loop to avoid nested calls to itself (avoiding non-optimized tail-recursion calls)
* The `poll` function now has basic debouncing to avoid hot looping


src/frontend/src/utils/countdown.ts
```diff
@@ -100,3 +101,87 @@ const calculateTimeRemaining = (
...
+    .toString()
+    .padStart(2, "0");
+  return `${mins}:${secs}`;
+}
+
+/// A one shot countdown timer that emits the remaining time as a formatted string ("mm:ss")
+// through an AsyncIterable.
+//
+// The countdown cannot be restarted.
+export class AsyncCountdown {
```
frederikrothenberger, line 120: There seems to be an issue here somewhere, because I got this when I tried it on chrome (and no timer showing):
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'expirationSeconds')
    at remainingSeconds (index.js:654:2560)
    at [Symbol.asyncIterator] (index.js:654:2720)
    at AsyncGenerator.next (<anonymous>)
    at index.js:654:966
    at Ci.update (index.js:654:1000)
    at Ci._$AS (index.js:610:2083)
    at k (index.js:2:385378)
    at S._$AI (index.js:2:386609)
    at G.p (index.js:2:386094)
    at S.$ (index.js:2:387284)
```

src/frontend/src/utils/countdown.ts
```diff
@@ -100,3 +101,87 @@ const calculateTimeRemaining = (
...
+  }
+
+  // The number of seconds elapsed since the epoch
+  static seconds() {
+    return new Date().getTime() / 1000;
+  }
+
+  // Whether the countdown was stopped explicitely.
+  // XXX: Some assumptions are made in the code that the countdown is never restarted.
+  private stopped;
```
frederikrothenberger, line 133: Isn't this a blind boolean as well? 😉

src/frontend/src/utils/countdown.ts
nmattia: Ah, eslint complained about me referencing `this` outside the object and I changed the code a bit just before pushing; I think that's the issue, thanks for catching it!

frederikrothenberger (approved):
Works now 👍

src/frontend/src/utils/countdown.ts
nmattia: Reverted the eslint-pleasing change and adapted the code to make it happy nonethless!

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
```diff
@@ -65,79 +85,114 @@ export const showVerificationCode = async (
...
+  );
+
+  showVerificationCodePage({
+    userNumber,
+    alias,
+    tentativeRegistrationInfo,
+    remaining: countdown.remainingFormattedAsync(),
+    cancel: () => {
+      countdown.stop();
+      window.location.reload();
```
peterpeterparker, line 99: is this window reload new and if yes, can we spare it? 

both for performance reason and ux, I think reload should just be used in last ressort but if not new, shall be it of course

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
```diff
@@ -65,79 +85,114 @@ export const showVerificationCode = async (
...
     userNumber,
     connection,
-    tentativeRegistrationInfo.device_registration_timeout,
-    credentialToBeVerified
-  );
+    credentialToBeVerified,
+    shouldStop: () => countdown.hasStopped(),
+  });
+
+  return await handlePollResult(userNumber, pollResult);
```
peterpeterparker, line 110: nth: there is a particular reason why this function does not take an object as argument while other functions of this PR do?

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
```diff
@@ -65,79 +85,114 @@ export const showVerificationCode = async (
...
-
-        if (credentialIdEqual(credentialId, credentialToBeVerified)) {
-          return true;
-        }
-      }
+
+    return res ?? false;
+  };
+
+  while (!shouldStop()) {
```
peterpeterparker, line 144: side note: "while something infinite" tends to make me "anxious" they might never stops, I rather like readin recursive function 🤷‍♂️

that said, should we add a max timeout to this loop? like shoult it stops in error after e.g. 1min?

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
```diff
@@ -65,79 +85,114 @@ export const showVerificationCode = async (
...
+  userNumber,
+  credential,
+  connection,
+}: {
+  userNumber: bigint;
+  credential: Array<number>;
+  connection: Connection;
+}) => {
+  const devices = await connection.lookupAuthenticators(userNumber);
+  for (const device of devices) {
```
peterpeterparker, line 186: nth: `devices.find((....) => device.credential_id.length === 1 && ...) !== undefined`

src/frontend/src/utils/countdown.ts
```diff
@@ -100,3 +101,80 @@ const calculateTimeRemaining = (
       .padStart(2, "0"),
   ];
 };
+
+// Prettify the number into a string: "mm:ss"
+function prettifySeconds(a: number): string {
```
peterpeterparker, line 106: - can it become more than 99 minutes and would the function and display handle this?

- side note: is there an API like Intl.NumberFormat would format this out of the box?

src/frontend/src/utils/countdown.ts
```diff
@@ -100,3 +101,80 @@ const calculateTimeRemaining = (
...
+// The countdown cannot be restarted.
+export class AsyncCountdown {
+  // Create a countdown from nanoseconds
+  static fromNanos(nanos: bigint) {
+    return new AsyncCountdown(Number(nanos / BigInt(1e9)));
+  }
+
+  // The number of seconds elapsed since the epoch
+  static seconds() {
+    return new Date().getTime() / 1000;
```
peterpeterparker, line 128: should it be rounded?

src/frontend/src/utils/countdown.ts
```diff
@@ -100,3 +101,80 @@ const calculateTimeRemaining = (
...
+  }
+
+  // Whether the countdown was stopped explicitely.
+  // XXX: Some assumptions are made in the code that the countdown is never restarted.
+  private stopped;
```
peterpeterparker, line 133: no type declaration?

src/frontend/src/utils/countdown.ts
```diff
@@ -100,3 +101,80 @@ const calculateTimeRemaining = (
...
+    this.stopped = this.pastExpiration();
+  }
+
+  // Stop the countdown explicitly
+  stop() {
+    this.stopped = true;
+  }
+
+  // Number of seconds remaining
+  remainingSeconds() {
```
peterpeterparker, line 146: return type declaration is missing

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
peterpeterparker: speaking of maybe no while or recursive function shall be use, would not this be a use case for using `setinterval`?

nmattia: the loop may need to keep running for 15mn. The previous version was recursive, which just hid the problem. An issue with`setInterval` is that we'd then need to make sure we don't lose the handle in case we abort... all in all, this is the safest I could think off. I'm also a bit hesitant to add a max timeout (like one hour) because anyway 15mn is quite long!

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
nmattia: I'm actually trying to get rid of this `reload()` 😓 But first needed to move a few things around in this PR, although I can't quite remove it just yet! But good catch

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
nmattia: Nope! Will change, good point

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
nmattia: Thanks!

src/frontend/src/utils/countdown.ts
nmattia: Currently it cannot be more than 99minutes; I just copied the existing code but I'll have a look at how we can format it better

src/frontend/src/utils/countdown.ts
nmattia: Definitely not! we use the fractional part to schedule a delay

src/frontend/src/utils/countdown.ts
nmattia: 🙃 will add

src/frontend/src/utils/countdown.ts
nmattia: 🙃 will add

src/frontend/src/flows/addDevice/welcomeView/showVerificationCode.ts
peterpeterparker: thanks for the explanation, no perfect solution at a time. maybe something that websocket on the IC will ultimately solve (don't know the exact use case here). all good anyway then

peterpeterparker (approved):
Already approving it too, thx

PR #1163, opened 2023-01-23
Change the order of dependencies in HACKING.md

dskloetd:
Change the order of dependencies in HACKING.md such that fulfilling a dependency doesn't require dependency listed below it.


nmattia (approved):
Thanks!

PR #1164, opened 2023-01-23
Remove  `--argument '(null)'` from HACKING.md

dskloetd:
Remove  `--argument '(null)'` from HACKING.md.
It was there to specify an empty optional parameter to the service but dfx no longer requires it.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -57,7 +57,7 @@ The fastest workflow to get the development environment running is to deploy onc
 ```bash
 npm ci
 dfx start [--clean] [--background]
-II_FETCH_ROOT_KEY=1 dfx deploy --no-wallet --argument '(null)'
+II_FETCH_ROOT_KEY=1 dfx deploy --no-wallet
 ```
 
 To serve the frontend locally via webpack (recommended during development), run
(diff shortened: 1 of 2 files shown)


nmattia (commented):
Thanks! Can you bump the required dfx version shown in this file as well? I think this would break for dfx <0.10.0: https://github.com/dfinity/sdk/blob/master/CHANGELOG.md#feat-use-null-as-default-value-for-opt-arguments

dskloetd:
I changed it. But how can I verify that it actually works with that version (and not an earlier version)?

nmattia:
The SDK has releases for each version, here's the one for 0.10.0: https://github.com/dfinity/sdk/releases/tag/0.10.0

You should be able to do something like this:

``` shell
$  mkdir -p bin && curl -SL https://github.com/dfinity/sdk/releases/download/0.10.0/dfx-0.10.0-x86_64-darwin.tar.gz | tar xzf - -C bin
$ ./bin/dfx ...
```

dskloetd:
I verified that `dfx deploy` without `--argument '(null)'` works with version 0.10.0 and not with version 0.9.3.

nmattia:
Thanks! I see there's at least one other occurence of `--argument '(null)'` in that file, and maybe more in the rest of the repo. Do you want to update those as well?

dskloetd:
Good point. Done

PR #1166, opened 2023-01-23
Redesign actionable list

meodai:
Changes the design of the actionable list in order to look good:

- Changes the look and feel of the device / recover list
- Adds a conditional warning to the device list if there is less than two devices.
- Adds the actionable list (devices, recovery methods) to the styleguide

Screenshot 2023-01-23 at 14 59 09

Screenshot 2023-01-23 at 15 20 57

Screenshot 2023-01-23 at 15 20 11


src/frontend/src/flows/manage/index.ts
```diff
@@ -115,36 +115,65 @@ const anchorSection = (userNumber: bigint): TemplateResult => html`
...
+  const isWarning = devices.length < 2;
+
+  if (isWarning === true) {
+    wrapClasses.push("c-card", "c-card--narrow", "c-card--warning");
+  }
+
+  return html`
+    <aside class="${wrapClasses.join(" ")}">
+      ${
+        isWarning === true
```
nmattia, line 129: here and elsewhere you can just have `isWarning` without `=== true`, our typescript config will make sure you never check for anything truthy instead of a boolean

src/frontend/src/flows/manage/index.ts
```diff
@@ -115,36 +115,65 @@ const anchorSection = (userNumber: bigint): TemplateResult => html`
...
-        >
-          <span class="c-tooltip__message c-tooltip__message--right c-card c-card--narrow"
-            >You can register up to ${MAX_AUTHENTICATORS} authenticator devices.
-            Remove a device before you can add a new one.</span
-          >
-          <span class="t-link t-link--discreet">
-            <i class="t-link__icon" aria-hidden=true>+</i>
-            Add new device
+): TemplateResult => {
+  const wrapClasses = ["l-stack"];
```
nmattia, line 119: out of curiosity, why "wrap"?

src/frontend/src/flows/manage/index.ts
```diff
@@ -115,36 +115,65 @@ const anchorSection = (userNumber: bigint): TemplateResult => html`
...
+  const isWarning = devices.length < 2;
```
nmattia, line 120: nit:

```suggestion
  const shouldWarn = devices.length < 2;
```

I think it's clearer that "this" is not a warning (yet), but instead that we _should_ be warning the user

nmattia (commented):
Nice! couple minor comments

src/frontend/src/flows/manage/index.ts
meodai: I thought of it as the component has a warning state that is true. So `isWarning` seamed like a good name. Should sounds a bit a like suggestion less then a state the component is in no?

src/frontend/src/flows/manage/index.ts
meodai: oh OK i guess I am used to do it, do you want me to remove it?

nmattia: All good!

nmattia (approved):

PR #1167, opened 2023-01-23
Update copy and spacing

meodai:
- Make the writing sound a bit nicer
- More constant use of titles and lead
- Center a few titles


meodai:
@nmattia any idea why this fails?

src/frontend/src/flows/authorize/index.ts
```diff
@@ -32,27 +32,26 @@ export const authnTemplateAuthorize = ({
...
 
-  const wrap = (slot: string) => html`
-    <div class="l-stack t-centered">
+  const wrap = (title: string) => html`
+    <div class="t-centered">
+      <h1 class="t-title t-title--main">${title}</h1>
       <p class="t-lead">
-        ${slot}<br />
-        <span class="t-strong">${origin}</span><br />
+        to continue to <a href="${origin}" target="_blank">${origin}</a><br />
```
nmattia, line 39: Is `noopener noreferrer` still relevant nowadays? CC @peterpeterparker

src/frontend/src/flows/manage/index.ts
```diff
@@ -24,20 +24,25 @@ import { mainWindow } from "../../components/mainWindow";
 
 /* Template for the authbox when authenticating to II */
 export const authnTemplateManage = (): AuthnTemplates => {
-  const wrap = (slot: string): TemplateResult => html`
-    <div class="t-centered l-stack">${slot}</div>
+  const wrap = (slot: string, title: string): TemplateResult => html`
```
nmattia, line 27:
```suggestion
  const wrap = ({slot, title}: {slot: string, title: string}): TemplateResult => html`
```
We've started using named args whenever possible

src/frontend/src/flows/recovery/pickRecoveryDevice.ts
```diff
@@ -1,17 +1,17 @@
...
-      <h2 class="t-title">Recovery devices</h2>
-      <ol class="c-list l-stack" id="deviceList"></ol>
-    </div>
+    <hgroup>
+      <h1 class="t-title t-title--main">Choose Recovery Method</h1>
+      <p class="t-paragraph t-lead">
+        How do you want to recover your Internet Identity?
+      </p>
+    </hgroup>
+    <div id="deviceList"></div>
```
nmattia, line 14: This changed from an `ol` to a `div`, just double checking that it's on purpose (since the id suggest it's a list)!

nmattia (commented):
Nice! just a couple of questions

src/frontend/src/flows/authorize/index.ts
meodai: I think it is, good point.

src/frontend/src/flows/recovery/pickRecoveryDevice.ts
meodai: But before it resulted in a list wrapped by an other list in the dom. the "DeviceList" is used as an ID to append the actual list, I think its fine.

meodai:
@nmattia The part I could is done, discuss the "copy" to clipboard feature with mary, feel free to remove it, if you don't feel good about having it in, in this release.

PR #1168, opened 2023-01-24
Rename 'Cargo fmt' workflow

nmattia:
The workflow wasn't only running `cargo fmt`, but also `cargo clippy`. This meant that the checks reported on PR pages were confusing.


.github/workflows/rust-checks.yml
```diff
@@ -1,4 +1,5 @@
-name: Cargo fmt
+# Rust specific tests like formatting, clippy, etc
+name: Rust checks
```
dskloetd, line 2: Does cargo fmt part just check whether the rust files are formatted correctly, or does it actually change them if they are not?
I'm asking because I'm assuming that cargo fmt can actually change files but "Rust checks" makes it sound like it only checks and doesn't actually change anything.
(I don't know if a workflow could actually change what's committed or if that's not possible anyway.)

nmattia: That's correct, the cargo fmt bit will actually push the formatted changes; however the clippy bit won't. How would you call this?

dskloetd: I would call it "Cargo fmt and clippy" and remove the "etc.".
If/when we decide to add more, maybe then it's clear what it should be called instead.

nmattia: Right; that was my first thought but I assumed whomever would add a job would forget to update the name. I'm happy with "Cargo fmt and clippy" though. How would you name the file though...?

dskloetd: Can it be just `rust.yml`?

.github/workflows/rust.yml
```diff
@@ -1,5 +1,5 @@
-# Rust specific tests like formatting, clippy, etc
-name: Rust checks
+# Rust specific tests
+name: Rust
```
dskloetd, line 2: If you worry that this could be confused with actual tests written in rust, we could also go with something like "Rust source code".

dskloetd (approved):

.github/workflows/rust.yml
```diff
@@ -1,4 +1,5 @@
-name: Cargo fmt
+# Rust specific tests
```
dskloetd, line 1: I thought we would keep "# Rust specific tests like formatting, clippy" and just remove the "etc.".

.github/workflows/rust.yml
nmattia: I've stopped worrying a long time ago :)

.github/workflows/rust.yml
nmattia: Let's go with this, no risk of outdated comments this way...

PR #1171, opened 2023-01-24
Introduce temporary keys

nmattia:
A temporary key can be used in lieu
of the device but has a short expiration time. These keys are used as a workaround for WebAuthn needing two users interactions: one for "create" and one for "sign". So instead we only "create" and instead authenticate the user with a temporary key for their first visit.

Note: we link the temporary keys to a device (as opposed to an anchor) so that we can make sure the temporary key is dropped if the device itself is removed. This ensures the temporary key is no more powerful than the device, i.e. if the user decides to remove the device right after registration, then the the temporary key cannot be used to authenticate (similary to how the browser session key pair cannot be used to authenticate if the actual delegated WebAuthn device is removed)


src/internet_identity/src/main.rs
```diff
@@ -459,13 +468,22 @@ fn authenticated_anchor_operation<R>(
 
 /// Checks if the caller is authenticated against the anchor provided and returns a reference to the device used.
 /// Traps if the caller is not authenticated.
-fn trap_if_not_authenticated(anchor: &Anchor) -> &Device {
+fn trap_if_not_authenticated(anchor_number: AnchorNumber, anchor: &Anchor) -> &Device {
```
nmattia, line 471: this opens the door to errors (i.e. mismatched `anchor_number` and `anchor`), how about fetching the `Anchor` in here instead of passing it as an argument?

src/internet_identity/src/main.rs
```diff
@@ -459,13 +468,22 @@ fn authenticated_anchor_operation<R>(
...
+    state::with_temp_keys_mut(|temp_keys| temp_keys.prune_expired_keys());
```
nmattia, line 472: I missed this line originally, since it's quite far away from the `if` below. Would it make sense to have `check_temp_key` do the pruning?

src/internet_identity/src/state/temp_keys.rs
```diff
@@ -0,0 +1,150 @@
...
+    ///
+    /// In addition, the temporary key is also linked to an anchor so that even if the same device is
+    /// added to multiple anchors, it can only be used in the context of the anchor it was created for.
+    ///
+    /// Since temp keys can only be added during registration, the the max number of temp keys is
+    /// bounded by the registration rate limit.
+    temp_keys: HashMap<DeviceKey, TempKey>,
+
+    /// Heap to efficiently prune expired temp keys
+    expirations: BinaryHeap<TempKeyExpiration>,
```
nmattia, line 32: Have you had performance issues with iterating the HashMap? How many temp keys do we expect at any given time?

Out of curiosity: why use a binary heap? Wouldn't a double ended queue be more efficient with O(1) push and pop? Or do you not expect to insert `TempKeyExpiration` "monotonously"?

src/internet_identity/src/state/temp_keys.rs
```diff
@@ -0,0 +1,150 @@
...
+        if let Some(temp_key) = self.temp_keys.get(device_key) {
+            if temp_key.anchor == anchor {
+                // Only remove temp key if the anchor matches
+                self.temp_keys.remove(device_key);
+            }
+        }
+    }
+
+    pub fn prune_expired_keys(&mut self) {
+        const MAX_TO_PRUNE: usize = 100;
```
nmattia, line 68: Do you expect the number of temp keys to often be beyond 100? Could we instead limit the number of concurrent sign ups -- or don't we actually rate limit more aggressively anyway?

src/internet_identity/src/state/temp_keys.rs
```diff
@@ -0,0 +1,150 @@
...
+        };
+        if temp_key.expiration < time() {
+            return Err(TempKeyError::Expired);
+        }
+        if temp_key.anchor != anchor {
+            return Err(TempKeyError::AnchorMismatch);
+        }
+        if &temp_key.principal != caller {
+            return Err(TempKeyError::Invalid);
+        }
```
nmattia, line 101: Are these various error types read/used somewhere?

Can you always guarantee that an expired temp key will return `TempKeyError::Expired`? Won't it most likely return `NotFound`?

nmattia (commented):
Nice, thanks! Couple questions

src/internet_identity/src/main.rs
frederikrothenberger: Makes, sense -> refactored.

src/internet_identity/src/state/temp_keys.rs
frederikrothenberger: Yeah, I thought the details would be more useful. But if needed, they can be added back later.

frederikrothenberger (approved):
Approved with review from @nmattia on the changes I made.

src/internet_identity/src/state/temp_keys.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        let now = time();
+        for _ in 0..MAX_TO_PRUNE {
+            let Some(expiration) = self.expirations.front() else {
+                break;
+            };
+
+            if expiration.expiration > now {
+                break;
+            }
```
nmattia, line 103: Could add a note here that we know that the expiration objects are sorted because we control the expiration period (and because `time()` is monotonous) (implicit assumption used here)

PR #1175, opened 2023-01-27
Prune old Chan listeners

nmattia:
This updates the `Chan` to only use weak references to its listeners, so that it can detect gone/GCed listeners and stop sending to them.


peterpeterparker (approved):
LGTM, thanks

PR #1176, opened 2023-01-27
Use support FAQ page

nmattia:
This drops our FAQ page in favor of the official II support FAQ, where the questions were ported already. Whenever necessary we link to that external FAQ, and we redirect (301) `/faq` to it.


src/frontend/src/components/footer.ts
```diff
@@ -34,7 +34,7 @@ export const footer = html`<footer class="l-footer">
   ${navigationLink({
     labelText: "FAQ",
     id: "faqLink",
-    url: "/faq",
+    url: "https://support.dfinity.org/hc/en-us/sections/8730568843412-Internet-Identity",
```
peterpeterparker, line 37: do you need to extend `navigationLink` with `target=_blank` and `rel=noopener noreferrer`?

src/frontend/src/index.ts
```diff
@@ -75,9 +74,13 @@ const init = async () => {
   // https://github.com/dfinity/internet-identity#build-features
   showWarningIfNecessary();
 
-  // Custom routing to the FAQ page
+  // Redirect to the FAQ
+  // The canister should already be handling this, this is just a safety
```
peterpeterparker, line 78: can you improve the comment by saying that the canister retuns a redirect 301 for /faq?

I was confuse at first, added few review comments until I read the rust code

peterpeterparker (approved):
LGTM, minor questions

src/frontend/src/index.ts
```diff
@@ -75,9 +74,13 @@ const init = async () => {
...
+  // measure
   if (window.location.pathname === "/faq") {
-    return faqView();
+    const faqUrl =
+      "https://support.dfinity.org/hc/en-us/sections/8730568843412-Internet-Identity";
```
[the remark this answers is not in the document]
nmattia: I did originally, but then it turned out to be a an extra file for nothing (only 2 uses; the 3 one in the frontend to an actual FAQ entry has a different URL). I've added to my todolist to actually add the constant to e.g. the .did file or something so that at least the canister can benefit from it as well.

src/frontend/src/components/footer.ts
nmattia: Good point, spotted another link that was missing it

src/frontend/src/index.ts
nmattia: Done, thanks! Added a comment in the canister as well

PR #1184, opened 2023-01-31
Introduce support for i18n and extract copy in about and authz pages

nmattia:
The first commit introduces a helper for loading string copy objects and turning them into dynamic lit templates. The second commit applies this to JSON-defined copies for the about and authorize pages.

The helper is labeled I18n as it supports dynamically updating the language. For now we only support english, with more languages to come soon.

These changes will also allow product people to update copy without creating conflicts with engineering changes.


src/frontend/src/flows/about/copy.json
```diff
@@ -0,0 +1,24 @@
...
+    "easy_secure_login": "Easy and Secure Login",
+    "privacy_and_convenience": " Get the privacy of many separate accounts, with the convenience of a single sign-on. Internet Identity generates a new private key for every dapp that you use, and stores the keys on your device. With just your Identity Anchor and device, you can manage hundreds of private keys and accounts.",
+    "explore_dapps": "Explore dapps on the Internet Computer",
+    "ic_dapps": "With an Identity Anchor, you can create accounts with dapps on the Internet Computer, like:",
+
+    "distrikt_name": "Distrikt",
+    "distrikt_link": "https://distrikt.app/",
+    "distrikt_desc": "A decentralized, professional social media network that empowers users to own and control their identity and data.",
+
+    "discover_name": "Discovr",
```
peterpeterparker, line 16: `Dscvr` no?

src/frontend/src/flows/about/copy.json
```diff
@@ -0,0 +1,24 @@
...
+    "discover_link": "https://dscvr.ic0.app/",
```
peterpeterparker, line 17: `dscvr.one`

src/frontend/src/flows/about/index.ts
```diff
@@ -2,115 +2,102 @@ import { html, render } from "lit-html";
...
-        </ul>
-      </div>
-      <div class="about__compatibility-flexchild">
-        <h3 class="t-title">Recommended Mobile</h3>
-        <ul class="c-list c-list--bulleted l-stack">
-          ${compatibilityData.mobile.map(
-            (browser) => html`<li>${browser}</li>`
-          )}
-        </ul>
+const aboutContentSlot = (i18n: I18n) => {
```
peterpeterparker, line 9: nth: because I like to descruct everything, would following work and be cool?

```
const aboutContentSlot = ({i18n}: I18n) => {
    const copy = i18n(copyJson);
```

src/frontend/src/utils/i18n.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+// The dynamic templates, i.e. { some_key_title: <language-dependent template> }
+type DynamicCopy<Keys extends string> = { [key in Keys]: TemplateResult };
+
+/// A class that converts a string copy definition into dynamic lit templates dependent on the
+// currently selected language.
+export class I18n<Lang extends string> {
+  // The 'Chan' used to propagate language updates
+  private chan: Chan<Lang>;
```
peterpeterparker, line 22: not related to the PR question: JavaScript newly I think supports `#` to identify `private` member. Do you know if TypeScript converts them?

`private chan`-> bundle `#chan`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

src/frontend/src/utils/i18n.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.chan = new Chan(def);
+    this.lang = def;
+  }
+
+  i18n<Keys extends string>(copy: StringCopy<Keys, Lang>): DynamicCopy<Keys> {
+    // Use the default language as the definition for expected keys.
+    // Missing keys in other languages are a type error.
+    type DefaultCopy = StringCopy<Keys, Lang>[Lang];
+    const defCopy: DefaultCopy = copy[this.lang];
+    const keys: [Keys] = Object.keys(defCopy) as [Keys];
```
peterpeterparker, line 37: really not sure at all, can `defCopy` be `undefined` here? 

if the `lang` is initialized with a unknown language, couldn't it be the case? if yes, `Ojbect.keys` will throw an error

src/frontend/src/utils/i18n.ts
```diff
@@ -0,0 +1,62 @@
...
+    const internationalized = keys.reduce((acc, k) => {
+      const value: AsyncIterable<string> = this.chan.map((lang) => {
+        return copy[lang][k];
+      });
+
+      acc[k] = html`${asyncReplace(value)}`;
+      return acc;
+    }, {} as DynamicCopy<Keys>);
+
+    return internationalized;
```
peterpeterparker, line 49: lit takes care of freeing the memory when all these dynamic templates aren't use anymore right?

src/frontend/src/utils/i18n.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  // Set the language, causing all templates to update
+  setLanguage(lang: Lang) {
+    this.chan.send(lang);
+    this.lang = lang;
+  }
+
+  // Retrieve the currently selected language
+  getLanguageAsync(): AsyncIterable<Lang> {
+    return this.chan.map((x) => x);
```
peterpeterparker, line 60: not sure, why the `map` here?

peterpeterparker (commented):
Looks super duper cool 👍

Just few questions for my understanding if you don't mind.

src/frontend/src/flows/about/copy.json
nmattia: nope, there's a typo in the original copy, which I don't want to change without approval

src/frontend/src/flows/about/copy.json
nmattia: Same as above, there's a typo; will need to work with the relevant people to figure out why it was written like this

src/frontend/src/utils/i18n.ts
nmattia: My bad, I should add a comment or fix it. It's a work around because currently `.map` is the only `Chan` function that copies the output (you can't call `.recv()` several times). Good catch!

src/frontend/src/utils/i18n.ts
nmattia: Yes that's right! Although in practice the user may change language a couple of times so "old" values will probably be kept in memory.

src/frontend/src/utils/i18n.ts
nmattia: So Typescript would prevent this, because (from the types) `copy` _must_ have a key `this.lang`. Now, that's the theory....

src/frontend/src/flows/about/copy.json
peterpeterparker: Makes sense

src/frontend/src/utils/i18n.ts
peterpeterparker: Nice then

peterpeterparker (approved):
Looks great to me, thanks

src/frontend/src/utils/i18n.ts
nmattia: Didn't know that was a thing, thanks!

And yes, looks like it's supported: https://www.typescriptlang.org/play?ts=3.8.3#example/private-class-fields

src/frontend/src/flows/about/index.ts
nmattia: For consistency I'll keep "1 arg => inline, 2+ => destruct"

PR #1186, opened 2023-01-31
Fix ic-commit update

nmattia:
The workflow was looking for the wrong files to upload. It also contained trailing whitespaces which are now removed.


frederikrothenberger (approved):
Good catch!

PR #1189, opened 2023-01-31
Fix scrolling on mobile

nmattia:
Some card effects were causing extra scrollbars to appear and break scrolling on small devices. We just disable those on mobile.


peterpeterparker (approved):
Thanks

PR #1191, opened 2023-02-01
Hide alias in public lookup method

frederikrothenberger:
This PR removes the alias from the `lookup` method. In order to not break existing clients, the field is kept but blanked out.

### DIFF
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -270,40 +270,50 @@ export class Connection {
     authenticatedConnection: AuthenticatedConnection
   ) {
     // Find the device based on the credential_id
-    const device = devices.find((device) => {
-      const credentialId = device.credential_id[0];
-      if (credentialId === undefined) {
-        return false;
-      }
-      return bufferEqual(
-        Buffer.from(credentialId),
-        attachmentInfo.credentialId
-      );
-    });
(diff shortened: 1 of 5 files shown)


src/frontend/src/utils/iiConnection.ts
```diff
@@ -270,40 +270,50 @@ export class Connection {
...
       // if it fails, no harm done
-      authenticatedConnection.update(device).catch((error) => {
-        console.warn(
-          `unable to update device ${JSON.stringify(device)}, error: ${error}`
-        );
-      });
+
+      authenticatedConnection
+        // We need to refetch the device using `getAnchorInfo` because the devices obtained by `lookup` no longer have an alias
+        .getAnchorInfo()
```
frederikrothenberger, line 285: I'm really not happy that I had to do this here. So I think we should do the following:
* Properly set the `key_type` on register, if available
* Remove this code once the domain migration is live. This will force _all_ devices, where this property is meaningful to go through re-registration, making this code here (that updates it for old devices) irrelevant due to point 1. Firefox and Edge will still create credentials with unknown key_type, however the subset of people that will use the same credential on multiple browsers (where one of which supports the `key_type` attribute) is probably small enough to simply ignore.

WDYT?

nmattia (approved):
rubber stamp

PR #1194, opened 2023-02-02
Allow changing anchor range on upgrade

frederikrothenberger:
This PR introduces the ability to change the configured anchor range on upgrade. It introduces additional safe guards so that it cannot be changed in a way that would affect existing anchors.


src/internet_identity/src/main.rs
```diff
@@ -233,14 +219,15 @@ fn post_upgrade(maybe_arg: Option<InternetIdentityInit>) {
     // load the persistent state after initializing storage, otherwise the memory address to load it from cannot be calculated
     state::load_persistent_state();
 
+    apply_init_arg(maybe_arg);
+}
+
+fn apply_init_arg(maybe_arg: Option<InternetIdentityInit>) {
```
nmattia, line 225: small nit about the name, it's not really "init" arg now is it?

src/internet_identity/src/storage.rs
```diff
@@ -340,6 +343,23 @@ impl<M: Memory> Storage<M> {
...
+
+        // restrict further if II has users to protect existing anchors
+        if self.header.num_anchors > 0 {
+            if self.header.id_range_lo != lo {
+                trap(&format!(
+                    "set_anchor_number_range: specified range [{lo}, {hi}) does not start from the same number ({}) \
+                     as the existing range thus would make existing anchors invalid"
+                    , {self.header.id_range_lo}));
+            }
+            if (hi - lo) < self.header.num_anchors as u64 {
```
nmattia, line 355: it would be enough to check `hi >= self.header.id_range_hi` here, right? or am I missing something?

src/internet_identity/tests/tests.rs
```diff
@@ -53,11 +53,35 @@ mod upgrade_tests {
...
+        )?;
+
+        let stats = api::stats(&env, canister_id)?;
+        assert_eq!(stats.assigned_user_number_range, (2000, 4000));
+        Ok(())
+    }
+
+    /// Test to verify that the user range cannot be changed to modify anchor numbers of existing anchors.
+    #[test]
+    fn should_not_affect_existing_anchors_on_user_range_change() {
```
nmattia, line 80: ah, I misunderstood this at first. I assumed this was checking that anchors would still work after changing the range. Is there a test for that? (just asking)

nmattia (approved):

src/internet_identity/src/storage.rs
frederikrothenberger: Hm, I don't think it would check for the same thing?
* `(hi - lo) < self.header.num_anchors` checks that all existing anchors would still fit (but allows shrinking the range).
* checking for `hi >= self.header.id_range_hi` would only allow growing the range.

I'm happy to change it though, if you feel it makes things clearer.

src/internet_identity/src/main.rs
frederikrothenberger: Renamed to `apply_install_arg` 👍

src/internet_identity/src/storage.rs
nmattia: Sorry, I meant in the context of the earlier check! I assumed you were checking that (1) existing anchors are not shifted (by making sure `lo_before` = `lo_after`) and that (2) all anchors still fit (and since `lo_after == lo_before`, this would mean checking that `hi_after >= hi_before`).

But maybe you're checking something else here?

src/internet_identity/tests/tests.rs
frederikrothenberger: No, actually not. I will add one 👍

src/internet_identity/src/storage.rs
frederikrothenberger: But your proposed check number (2) does not actually do that: I.e. `hi_before` does not correspond to the highest _existing_ anchor number, but to the highest _allowed_ anchor number. I.e. if we have the range `(0, 10]` and 5 anchors it is ok to change the range down to `(0, 6]` since the anchors `6, 7, 8, 9` are unused. But maybe this is unnecessary complexity...

nmattia: Oh my, you're right! Can you add a comment about this?

PR #1196, opened 2023-02-03
Make authentication the first action of authenticated canister calls

frederikrothenberger:
This PR refactors call authentication to happen as the very first thing of all the calls that require user authentication. Moving the the authentication to main makes it very easy to quickly determine which calls are authenticated and which are not.


src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -109,8 +105,6 @@ fn get_verified_device(
     anchor_number: AnchorNumber,
     user_verification_code: DeviceVerificationCode,
 ) -> Result<DeviceData, VerifyTentativeDeviceResponse> {
-    trap_if_not_authenticated(&state::anchor(anchor_number));
```
nmattia, line 112: just to be clear, this was hoisted to `verify_tentative_device`, correct?

nmattia (approved):
Just one question to double check something, lgtm otherwise

src/internet_identity/src/anchor_management/tentative_device_registration.rs
frederikrothenberger: Yes, exactly. This was the most ill-placed authentication check that I moved.

PR #1199, opened 2023-02-03
Deconstruct manage

nmattia:
The convention is to deconstruct function arguments when the function has more than one argument.


peterpeterparker (approved):
Thanks

PR #1203, opened 2023-02-06
Fix: typos

omahs:
Fix: typos

### DIFF
--- a/docs/ii-spec.md
+++ b/docs/ii-spec.md
@@ -409,7 +409,7 @@ But upon closer inspection (heh), this is not actually useful.
 
 -   Another is to protect against a malicious actor. But that is only useful if the malicious actor doesn't have an equally effective attack vector anyways, and in our case they do: If they want to flood the NNS with calls, they can use calls that do authenticate (e.g. keeping removing and adding devices, or preparing delegations); these calls would pass message inspection.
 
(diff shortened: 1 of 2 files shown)


docs/ii-spec.md
```diff
@@ -409,13 +409,13 @@ But upon closer inspection (heh), this is not actually useful.
 
 -   Another is to protect against a malicious actor. But that is only useful if the malicious actor doesn't have an equally effective attack vector anyways, and in our case they do: If they want to flood the NNS with calls, they can use calls that do authenticate (e.g. keeping removing and adding devices, or preparing delegations); these calls would pass message inspection.
 
-On the flip side, implementing `canister_inspect_message` adds code, and thus a risk for bugs. In particular it increases the risk that some engineer might wrongly assume that the authentication check in `canister_inspect_message` is sufficient and will not do it again in the actual method, which could lead to a serious bug.
+On the flip side, implementing `canister_inspect_message` adds code, and thus a risk for bugs. In particular it increases the risk that some engineers might wrongly assume that the authentication check in `canister_inspect_message` is sufficient and will not do it again in the actual method, which could lead to a serious bug.
```
nmattia, line 412: I think this one is actually correct! 
```suggestion
On the flip side, implementing `canister_inspect_message` adds code, and thus a risk for bugs. In particular it increases the risk that some engineer might wrongly assume that the authentication check in `canister_inspect_message` is sufficient and will not do it again in the actual method, which could lead to a serious bug.
```

I.e. "it increases the risk that some engineer _who's working on feature XYZ_ might wrongly assume..."

nmattia (commented):
Thanks a lot!

nmattia (approved):

PR #1204, opened 2023-02-06
Fix max expiration in spec

nmattia:
The spec mentioned the max delegation expiration was 8 days, which was changed a few months ago to 30 days.


peterpeterparker (approved):
LGTM, thx

PR #1205, opened 2023-02-06
Extract more copy in authz

nmattia:
Some of the text in the authorization flow was left behind in the code.


peterpeterparker (approved):
LGTM

PR #1207, opened 2023-02-06
Add aliases for utils and generated modules

nmattia:
This introduces webpack & ts aliases for the `generated` and `utils` modules, which are used from many places and used to include lots of directory changes (`../../../utils/`, etc)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,7 +15,12 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "types": ["node", "jest", "webdriverio/async"]
+    "types": ["node", "jest", "webdriverio/async"],
+    "baseUrl": "./",
+    "paths": {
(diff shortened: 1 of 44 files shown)


tsconfig.json
```diff
@@ -15,7 +15,12 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "types": ["node", "jest", "webdriverio/async"]
+    "types": ["node", "jest", "webdriverio/async"],
+    "baseUrl": "./",
+    "paths": {
+      "@generated/*": ["src/frontend/generated/*"],
```
peterpeterparker, line 21: can you use `$` instead of `@` for prefix?

- org scoped library published to npm are prefixed with `@`
- consistency with NNS-dapp

nmattia:
Closing until I can figure out how to work around some static page build issues

PR #1208, opened 2023-02-07
Introduce new method for public anchor information

frederikrothenberger:
This PR introduces a new method to retrieve public anchor info to replace
the existing `lookup` method. The `lookup` method is replaced for the
following reasons:
* The vec DeviceData type no longer matches the information give (since the alias is stripped).
* The vec type offers no flexibility with regards to additional information unrelated to devices.
* Introducing a new type for the public data gives us more flexibility going forward to add information to the device that should only be available for authenticated callers.

This PR is the first step in replacing the `lookup` method. The next step is to change the front-end to no longer use the `lookup` method.


frederikrothenberger:
Will be replaced by a new `get_credentials` endpoint.

PR #1209, opened 2023-02-07
Add notify call to report asset loads

frederikrothenberger:
This PR introduces a new method to report asset loads. It will be used to collect basic page load metrics for the II front-end.

### DIFF
--- a/src/internet_identity/src/assets.rs
+++ b/src/internet_identity/src/assets.rs
@@ -2,10 +2,12 @@
 //
 // This file describes which assets are used and how (content, content type and content encoding).
 
+use crate::http::FAQ_PATH;
+use crate::state::{usage_metrics_mut, AssetPath, StatusCode, UsageMetrics};
 use crate::{http, state};
 use base64::engine::general_purpose::STANDARD as BASE64;
 use base64::Engine;
-use ic_cdk::api;
+use ic_cdk::{api, trap};
 use lazy_static::lazy_static;
 use sha2::Digest;
 
@@ -87,6 +89,12 @@ pub fn init_assets() {
     });
 }
 
+pub fn trap_if_not_asset(path: &str) {
+    if !get_assets().iter().any(|(name, _, _, _)| *name == path) {
+        trap(&format!("Asset '{path}' does not exist"))
+    }
+}
+
 // Get all the assets. Duplicated assets like index.html are shared and generally all assets are
 // prepared only once (like injecting the canister ID).
 fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType); 8] {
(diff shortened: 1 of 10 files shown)


src/internet_identity/src/assets.rs
```diff
@@ -134,3 +142,57 @@ fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType);
         ),
     ]
 }
+
+pub fn increase_asset_counter(asset: String, status_code: u16) {
+    const MAX_TRACKED_ASSETS: u8 = 30;
+
+    if (200..400).contains(&status_code) {
```
nmattia, line 149: this looks like it checks for the following:

``` rust
if status_code >= 200 && status_code < 400 {
```

although the intent is not clear, and without looking at the doc I can't tell whether `400` is included in the range. Although, this seems orders of magnitude slower... so surely I'm missing something?

src/internet_identity/src/assets.rs
```diff
@@ -134,3 +142,57 @@ fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType);
...
+        // prevent clients from recording successful status codes for non-existing assets
+        trap_if_not_asset(&asset);
+
+        // prevent clients from recording impossible status codes for existing assets
```
nmattia, line 153: why not record "impossible" status codes? wouldn't those be the most interesting ones? also with this code in place I'm 100% sure we'll forget to update it if we change the return code for the faq

src/internet_identity/src/assets.rs
```diff
@@ -134,3 +142,57 @@ fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType);
...
+    }
+
+    usage_metrics_mut(|metrics| {
+        match metrics
+            .asset_loading_counters
+            .get_mut(&(asset.clone(), status_code))
+        {
+            None => {
+                if metrics.asset_loading_counters.len() >= MAX_TRACKED_ASSETS as usize {
+                    // This simple logic will cause churn on the least used asset.
```
nmattia, line 166: I'm confused. Doesn't the logic above skip tracking on unknown assets? how could we have more than `get_assets().len()` assets?

src/internet_identity/src/assets.rs
```diff
@@ -134,3 +142,57 @@ fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType);
...
+    for (asset, counter) in metrics.asset_loading_counters.iter() {
+        if *counter < lowest_asset_counter {
+            least_used_asset = Some(asset.clone());
+            lowest_asset_counter = *counter
+        }
+    }
+    if let Some(asset) = least_used_asset {
+        metrics.asset_loading_counters.remove(&asset);
+    }
+}
```
nmattia, line 198: Couldn't this be done with something simpler like sorting and dropping the last one? If you do need the "speed" then a psqueue or some other datastructure might be more appropriate

src/internet_identity/src/assets.rs
frederikrothenberger: I made #1228 to disable the lint.

src/internet_identity/src/assets.rs
frederikrothenberger: No, it does track all status codes indicating failure, regardless if the path corresponds to an asset or not.

src/internet_identity/src/assets.rs
frederikrothenberger: I switched it to `clone` and `sort` which should be more readable.

src/internet_identity/src/assets.rs
```diff
@@ -134,3 +142,57 @@ fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType);
...
+
+pub fn increase_asset_counter(asset: String, status_code: u16) {
+    const MAX_TRACKED_ASSETS: u8 = 30;
+
+    if (200..400).contains(&status_code) {
+        // prevent clients from recording successful status codes for non-existing assets
+        trap_if_not_asset(&asset);
+
+        // prevent clients from recording impossible status codes for existing assets
+        if asset == FAQ_PATH && status_code != 301 || asset != FAQ_PATH && status_code != 200 {
```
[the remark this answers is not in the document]
frederikrothenberger: I thought extracting `trap_if_not_asset` would make the intent of that block more clear, whereas here the intent was clearer. I extracted this one now as well.

src/internet_identity/src/assets.rs
frederikrothenberger: No intermediary network component should modify success status codes, i.e. reporting status code 202 on an existing asset is a malicious input, so I discard it here.

nmattia (approved):
Still feel a bit uneasy about this, but rubber stamping to unblock!

PR #1210, opened 2023-02-07
Change recovery picker depending on context

nmattia:
This updates the component we use to pick a new recovery method to show different text depending on whether we're adding a recovery during sign in/sign up or from the "manage" page.

The implementation shoehorns the changes in the current project structure which should be revised a little. In particular, the `chooseRecoveryMethod` and `setupRecovery` modules should be moved to `components/`. This will be done separately.

There's a couple of things that could be improved in those modules (using e.g. `"cancel"` instead of `null` for cancellation, deconstructing args, etc) so I'll do another pass later focused on cleaning up.


meodai:
@nmattia since there are no CSS changes, this looks good to me, I think its well implemented

meodai (approved):

PR #1212, opened 2023-02-07
Add hover effect on list-actions

meodai:
As requested I added a hover effect to the settings icon. 

Hovers don’t make things look clickable, on touch devices there is no hover. 
The visual affordance is not bad imho. But I can try to make it look more clickable at some point. 

https://user-images.githubusercontent.com/608386/217272549-f61e8852-d2d3-4604-9c18-ab39a64c055e.mov


nmattia (approved):
LGTM, thanks!

PR #1213, opened 2023-02-07
Text Change on Enroll screen

meodai:
The text was out of date (referencing copy that was updated recently)


nmattia (approved):
LGTM, thanks!

PR #1217, opened 2023-02-10
Add daily and monthly active anchor statistics 

frederikrothenberger:
This PR adds stats about the number of unique active anchors
* per 24h window
* per 30 day window (updated every 24h)

In addition to the metrics exposed on the /metrics endpoint for completed
collection periods, the `stats` endpoint is extended to allow inspecting
ongoing data collection periods.


src/internet_identity/src/active_anchor_stats.rs
```diff
@@ -0,0 +1,101 @@
...
+            Some(ref mut stats) => {
+                rotate_active_anchor_stats_if_necessary(stats);
+                increment_stats_on_activity(stats, previous_activity_timestamp);
+            }
+        },
+    )
+}
+
+#[allow(clippy::identity_op)]
+fn rotate_active_anchor_stats_if_necessary(stats: &mut ActiveAnchorStatistics) {
```
nmattia, line 25: I think a line or two about what this does and how would be nice here

src/internet_identity_interface/src/lib.rs
```diff
@@ -227,6 +228,37 @@ pub struct ArchiveConfig {
...
+    pub daily_active_anchors: Option<ActiveAnchorCounter>,
+    pub monthly_active_anchors: Option<ActiveAnchorCounter>,
+}
+
+#[derive(Clone, CandidType, Deserialize, Eq, PartialEq, Debug)]
+pub struct OngoingActiveAnchorStats {
+    // Ongoing active anchor counter for
+    pub daily_active_anchors: ActiveAnchorCounter,
+    // Monthly active users are collected using 30 day sliding windows.
+    // This vec contains up to 30 30-day active windows each offset by one day.
```
przydatek, line 250: the comment in .did-file talks about up to 29 entries, not 30 -- is this intentional?

src/internet_identity/internet_identity.did
```diff
@@ -180,6 +181,33 @@ type ArchiveInfo = record {
...
+    daily_active_anchors: opt ActiveAnchorCounter;
+    monthly_active_anchors: opt ActiveAnchorCounter;
+};
+
+type OngoingActiveAnchorStats = record {
+    // Ongoing active anchor counter for
+    daily_active_anchors: ActiveAnchorCounter;
+    // Monthly active users are collected using 30 day sliding windows.
+    // This vec contains up to 29 30-day active windows each offset by one day.
+    // The vec is sorted with the first element being the next window to reach the end of the
```
przydatek, line 201: nit: this description reads a bit convoluted, maybe explain in terms of where are the newer buckets added: at the end of the vector, right?

src/internet_identity/src/active_anchor_stats.rs
```diff
@@ -0,0 +1,112 @@
...
+            let new_start_timestamp = counter.start_timestamp + 30 * DAY;
+            stats.completed.monthly_active_anchors = Some(counter);
+            stats
+                .ongoing
+                .monthly_active_anchors
+                .push(new_active_anchor_counter(new_start_timestamp));
+        }
+    }
+    if let Some(monthly_stats) = stats.ongoing.monthly_active_anchors.last() {
+        // start a new 30 day time window if the last one starts more than 24h in the past.
```
przydatek, line 56: I don't quite see why this cannot result in a montly-vector of more than 29 (or more than 30) entries -- maybe add a comment with an explanation?

src/internet_identity/src/active_anchor_stats.rs
```diff
@@ -0,0 +1,112 @@
...
+///
+/// For the monthly (30 day) active user counters:
+/// if the oldest ongoing 30 day bucket was started 30 days ago (or earlier), it replaces the current
+/// completed monthly active user counter and a new ongoing 30 day counter is created
+/// (starting from the end of the now completed counter).
+#[allow(clippy::identity_op)]
+fn rotate_active_anchor_stats_if_necessary(stats: &mut ActiveAnchorStatistics) {
+    let now = time();
+    if stats.ongoing.daily_active_anchors.start_timestamp + 1 * DAY <= now {
+        let new_start_timestamp = stats.ongoing.daily_active_anchors.start_timestamp + 1 * DAY;
```
przydatek, line 39: Strictly speaking, setting `new_start_time` like this (rather than setting it to `now`) is not quite accurate, as the new counter will miss the hits between `daily_active_anchors.start_timestamp` and `now` (i.e. they will be counted in the previous counter), and the same remark applies to the monthly counters below. 
The alternative of using `now` has the danger of "drifting", i.e. each window will end up being 24h + some milliseconds or seconds, so over time it can accumulate.  I guess then it is ok as is, but maybe it is worth mentioning in a comment?

src/internet_identity_interface/src/lib.rs
```diff
@@ -227,6 +228,37 @@ pub struct ArchiveConfig {
...
+pub struct CompletedActiveAnchorStats {
+    pub daily_active_anchors: Option<ActiveAnchorCounter>,
+    pub monthly_active_anchors: Option<ActiveAnchorCounter>,
+}
+
+#[derive(Clone, CandidType, Deserialize, Eq, PartialEq, Debug)]
+pub struct OngoingActiveAnchorStats {
+    // Ongoing active anchor counter for
+    pub daily_active_anchors: ActiveAnchorCounter,
+    // Monthly active users are collected using 30 day sliding windows.
```
przydatek, line 249: nit: "30 day" -> "30-day"

src/internet_identity_interface/src/lib.rs
```diff
@@ -227,6 +228,37 @@ pub struct ArchiveConfig {
...
+    // This vec contains up to 30 30-day active windows each offset by one day.
+    // The vec is sorted with the first element being the next window to reach the end of the
```
przydatek, line 251: not sure whether this file or .did-file is the source-of-truth, but also here explaining in terms of newer buckets could make things easier to understand.

przydatek (commented):
Sorry for the delay @frederikrothenberger, please see some questions/suggestions below.

src/internet_identity/src/active_anchor_stats.rs
```diff
@@ -0,0 +1,167 @@
...
+    previous_activity_timestamp: Option<Timestamp>,
+) {
+    // increase all counters and exit early if there is no previous activity
+    let Some(timestamp) = previous_activity_timestamp else {
+        stats.ongoing.daily_active_anchors.counter += 1;
+        stats.ongoing.monthly_active_anchors.iter_mut().for_each(|monthly_stats| monthly_stats.counter +=1 );
+        return;
+    };
+
+    if stats.ongoing.daily_active_anchors.start_timestamp > timestamp {
```
przydatek, line 134: nit: maybe add a comment explaining how this results in counting each anchor only once per period, as this seems not quite obvious.

przydatek (approved):
Thanks a lot for the improvements and additional explanations.  LGTM.

PR #1218, opened 2023-02-10
Remove legacy archive push integration

frederikrothenberger:
This PR removes the old archive integration mode where II pushed
entries to the archive using notify calls.


nmattia (approved):
Rubber stamp

PR #1219, opened 2023-02-10
Update ic-cdk and ic dependencies

frederikrothenberger:
This PR updates `ic-cdk` to v0.7 and `ic` to the latest master (as it is unpinnend). The most notable change in ic-cdk is that the timers feature is now its own crate ic-cdk-timers.

In addition, all other dependencies were upgraded within their minor version.


src/internet_identity/src/assets.rs
```diff
@@ -49,7 +51,7 @@ lazy_static! {
     // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
     pub static ref JS_SETUP_SCRIPT_SRI_HASH: String = {
         let hash = &sha2::Sha256::digest(JS_SETUP_SCRIPT.as_bytes());
-        let hash = base64::encode(hash);
+        let hash = BASE64.encode(hash);
```
nmattia, line 54: this is surprising; did `base64` introduce a breaking change on a minor version change?

EDIT: after checking, this was bumped from 0.20 to 0.21 (technically, 0.20 was removed). I'm still a bit confused why we were hitting 0.20 before since the manifest mentions 0.21

nmattia (approved):

src/internet_identity/src/assets.rs
frederikrothenberger: > I'm still a bit confused why we were hitting 0.20 before since the manifest mentions 0.21

We had inconsistencies between the test app and the main II / archive code. The test app started using 0.21 before the rest of the code.

PR #1222, opened 2023-02-13
Fix recovery device type checks based on alias

frederikrothenberger:
This PR fixes checks for recovery devices and recovery phrases that
were based on the device name. This change is required because the
public endpoints no longer give information about the alias.

This PR also includes some typing improvements around the alias being
stripped on the public 'lookup'  endpoint.

### DIFF
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -400,6 +400,7 @@ export class Connection {
     userNumber: UserNumber
   ): Promise<RecoveryDevice[]> => {
     const actor = await this.createActor();
+    // lookup blanks out the alias for privacy reasons -> omit alias from DeviceData
     const allDevices: Omit<DeviceData, "alias">[] = await actor.lookup(
       userNumber
     );


frederikrothenberger:
FYI: Screenshot changes happen because the `showcase` had capitalized recovery device names (whereas the actual recovery setup code did not capitalize the second word).

src/frontend/src/utils/recoveryDeviceLabel.ts
```diff
@@ -0,0 +1,14 @@
+import { DeviceData } from "../../generated/internet_identity_types";
+
+export const recoveryDeviceToLabel = (
+  device: Omit<DeviceData, "alias">
```
nmattia, line 4: I think you can enforce that `device` has the correct `purpose` and `key_type` here and mostly avoid runtime errors

frederikrothenberger: I implemented that in commit `517e9e4f0c51d68828fa4029e087770018a83674` but it seems quite involved. Please have a look, maybe there is a way to simplify.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -393,10 +396,16 @@ export class Connection {
...
+  lookupRecovery = async (
+    userNumber: UserNumber
+  ): Promise<RecoveryDevice[]> => {
     const actor = await this.createActor();
-    const allDevices = await actor.lookup(userNumber);
-    return allDevices.filter((device) => "recovery" in device.purpose);
+    const allDevices: Omit<DeviceData, "alias">[] = await actor.lookup(
+      userNumber
+    );
+    return allDevices.filter((device): device is RecoveryDevice =>
```
nmattia, line 406: doesn't `isRecoveryDevice` already have `is RecoveryDevice`?

src/frontend/src/utils/recoveryDevice.ts
```diff
@@ -0,0 +1,17 @@
+import { DeviceData, Purpose } from "../../generated/internet_identity_types";
+
+export type RecoveryDevice = Omit<DeviceData, "alias"> & {
+  purpose: Extract<Purpose, { recovery: null }>;
```
nmattia, line 4: `Extract<Purpose, { recovery: null }>`

isn't that just `{ recovery: null }`?

src/frontend/src/utils/recoveryDevice.ts
```diff
@@ -0,0 +1,17 @@
+import { DeviceData, Purpose } from "../../generated/internet_identity_types";
+
+export type RecoveryDevice = Omit<DeviceData, "alias"> & {
```
nmattia, line 3: Do `RecoveryDevice`s never have an alias? Also, there are a few occurrences of `RecoveryDevice & Omit<DeviceData, "alias">` and `RecoveryDevice & DeviceData` here and in other modules. Shouldn't these all just be `RecoveryDevice`?

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: Yes, but TypeScript is unfortunately not able to use that for type narrowing: https://github.com/microsoft/TypeScript/issues/16069

src/frontend/src/utils/recoveryDevice.ts
frederikrothenberger: Technically they do have an alias, in particular when interacting with authenticated canister calls (i.e. `add` and `update`) the alias is _required_.

Agreed, `RecoveryDevice & Omit<DeviceData, "alias">` is redundant. Thanks for pointing it out. However, `RecoveryDevice & DeviceData` is not, as this combines the existence of `alias` with the restriction on `purpose`.

src/frontend/src/utils/iiConnection.ts
nmattia:
```suggestion
    return allDevices.filter(isRecoveryDevice);
```

wouldn't this work though?

frederikrothenberger: Ah right. I thought I tried this, but apparently I had another typing thing mixed in... 🤔 
Thanks!

src/frontend/src/flows/manage/deviceSettings.ts
```diff
@@ -118,8 +128,14 @@ export const deviceSettings = async (
       device,
       isOnlyDevice,
       back: resolve,
-      protectDevice: () =>
-        protectDevice(userNumber, connection, device, isOnlyDevice, resolve),
+      protectDevice: (recoveryDevice: DeviceData & RecoveryDevice) =>
```
nmattia, line 131: I find it unfortunate that `protectDevice` now needs to take a device but I get it that it needs a `RecoveryDevice`. We'll be remodelling the device settings page soon so not a big problem.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -393,10 +396,14 @@ export class Connection {
     });
   };
 
-  lookupRecovery = async (userNumber: UserNumber): Promise<DeviceData[]> => {
+  lookupRecovery = async (
```
nmattia, line 399: would be nice to explain why you're removing the `alias`, even though it's temporary

nmattia (approved):

PR #1223, opened 2023-02-14
Remove pruning signatures from anchor operations

frederikrothenberger:
This removes the delegation pruning from anchor operations to keep the code focused. Instead the pruning in prepare_delegation is a bit more aggressive now.

One call to prepare_delegation is still far below the cycles limit with about 300M cycles to create the new delegation and at most 300M cycles to prune 50 expired delegations.


nmattia (approved):
LGTM, thanks!

PR #1225, opened 2023-02-14
Refactor new device alias

nmattia:
This refactors the alias picker in the add device flow, using more idiomatic lit constructs and using function naming more consistent with the rest of the project.

This is preparatory work for extracting copy, although this should probably be merged with the [other alias picker](https://github.com/dfinity/internet-identity/blob/adf3aab1d311bef4197035cea4b7e270c8b346ae/src/frontend/src/flows/register/alias.ts) (tracked).


peterpeterparker (approved):
LGTM

peterpeterparker:
Merging as requested.

PR #1227, opened 2023-02-14
Extract copy for alias picker

nmattia:
This extracts the alias picker's copy into a JSON file.


nmattia:
Lit doesn't like the attribute being set to a TemplateResult, will need to look into this

meodai (approved):
LGTM, but I think we can be a bit more descriptive in the keys:

input_placehodler. cancel_button, next__button?

nmattia:
>  I think we can be a bit more descriptive in the keys:

Yeah I agree. I've been pretty succinct with the other pages, but we can always change if/when copywriters & translators are confused

PR #1228, opened 2023-02-15
Disable clippy warning on range contains

frederikrothenberger:
This PR disables the clippy warning to avoid writing conditions like this: `status_code >= 200 && status_code < 400`


nmattia (approved):
Thanks! might be worth leaving a note why

PR #1229, opened 2023-02-15
Add last usage timestamp to devices

frederikrothenberger:
This PR introduces the last_usage_timestamp on devices when reading
from get_anchor_info.
To make clear that this field cannot be written, get_anchor_info
now returns a different type from the type used to modify devices.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -75,6 +75,16 @@ type DeviceData = record {
     origin: opt text;
 };
 
+type ReadOnlyDeviceData = record {
+    pubkey : DeviceKey;
+    alias : text;
+    credential_id : opt CredentialId;
+    purpose: Purpose;
+    key_type: KeyType;
+    protection: DeviceProtection;
+    last_usage_timestamp: opt Timestamp;
+};
+
 type RegisterResponse = variant {
     // A new user was successfully registered.
     registered: record {
@@ -198,7 +208,7 @@ type DeviceRegistrationInfo = record {
 };
 
 type IdentityAnchorInfo = record {
-    devices : vec DeviceData;
+    devices : vec ReadOnlyDeviceData;
     device_registration: opt DeviceRegistrationInfo;
 };
 
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -75,14 +75,16 @@ type DeviceData = record {
     origin: opt text;
 };
 
-type ReadOnlyDeviceData = record {
+// The same as `DeviceData` but with the `last_usage` field.
+// This field cannot be written, hence the separate type.
+type DeviceWithUsage = record {
     pubkey : DeviceKey;
     alias : text;
     credential_id : opt CredentialId;
     purpose: Purpose;
     key_type: KeyType;
     protection: DeviceProtection;
-    last_usage_timestamp: opt Timestamp;
+    last_usage: opt Timestamp;
 };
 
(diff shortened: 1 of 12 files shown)


src/internet_identity/internet_identity.did
```diff
@@ -74,6 +74,16 @@ type DeviceData = record {
...
 };
 
+type ReadOnlyDeviceData = record {
+    pubkey : DeviceKey;
+    alias : text;
+    credential_id : opt CredentialId;
+    purpose: Purpose;
+    key_type: KeyType;
+    protection: DeviceProtection;
+    last_usage_timestamp: opt Timestamp;
```
nmattia, line 84: that's quite a long field name! the longest, even!

src/internet_identity/internet_identity.did
```diff
@@ -74,6 +74,16 @@ type DeviceData = record {
     protection: DeviceProtection;
 };
 
+type ReadOnlyDeviceData = record {
```
nmattia, line 77: that's quite a long name! why is it RO btw? I guess there's no "RW" devices, but some are "read" and some are "write"

Also, can you add a note about why this differs from the other device type? That's not immediately clear to me.

Finally, I've often seen apis wrap a type with e.g. `type Timestamped<T> = { dat: T, timestamp: Timestamp }`. That might help deduplicate things!

src/internet_identity/src/anchor_management.rs
```diff
@@ -107,6 +108,7 @@ pub fn replace(anchor_number: AnchorNumber, old_device: DeviceKey, new_device: D
             "failed to replace device of anchor {anchor_number}: {err}"
         ))
     });
+
```
nmattia, line 111:
```suggestion
```

nmattia (approved):
I'm not 100% sure about the motivation for duplicating the type, and I think the naming could be improved slightly, but rubber stamping to unblock!

src/internet_identity/internet_identity.did
frederikrothenberger: Making `IdentityAnchorInfo.devices` a vector of `TimestampedDevice = record { dat: DeviceData, timestamp: Timestamp }` would be a breaking change. I'd rather avoid that.

src/internet_identity/src/storage/anchor.rs
```diff
@@ -15,6 +15,18 @@ pub struct Anchor {
     devices: Vec<Device>,
 }
 
+impl Device {
+    /// Applies the values of `device_data` to self while leaving the other fields intact.
+    pub fn apply_device_data(&mut self, device_data: DeviceData) {
```
nmattia, line 20: btw just my 2cts but I can easily see us forgetting to update this function if a field is added!

src/internet_identity/src/storage/anchor.rs
```diff
@@ -15,6 +15,18 @@ pub struct Anchor {
     devices: Vec<Device>,
 }
 
+impl Device {
+    pub fn apply_device_data(&mut self, device_data: DeviceData) {
```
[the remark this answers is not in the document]
frederikrothenberger: Thanks for spotting this, it was actually a bug. It should not touch fields that are not available on `DeviceData`.

src/internet_identity/src/storage/anchor.rs
frederikrothenberger: Agreed, I'll add a test that verifies it by checking that `DeviceData::from` a `Device` that had data applied to it is equal to the data that was applied.

PR #1230, opened 2023-02-16
Add /.well-known/ic-domains asset

frederikrothenberger:
This adds the asset required for the custom domains setup for II.

### DIFF
--- a/src/internet_identity/src/assets.rs
+++ b/src/internet_identity/src/assets.rs
@@ -21,6 +21,7 @@ pub enum ContentType {
     ICO,
     WEBP,
     CSS,
+    OCTETSTREAM,
 }
 
 // The <script> tag that loads the 'index.js'
@@ -86,7 +87,7 @@ pub fn init_assets() {
 
 // Get all the assets. Duplicated assets like index.html are shared and generally all assets are
 // prepared only once (like injecting the canister ID).
-fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType); 7] {
+fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType); 8] {
     let index_html: &[u8] = INDEX_HTML_STR.as_bytes();
     let about_html: &[u8] = ABOUT_HTML_STR.as_bytes();
     [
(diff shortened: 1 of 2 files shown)


src/internet_identity/src/assets.rs
```diff
@@ -21,6 +21,7 @@ pub enum ContentType {
     ICO,
     WEBP,
     CSS,
+    OCTETSTREAM,
```
frederikrothenberger, line 24: This is the content-type that the asset canister currently uses if you follow the instructions [here](https://github.com/r-birkner/portal/blob/rjb/custom-domains-docs-v2/docs/developer-docs/production/custom-domain/custom-domain.md#custom-domains-on-the-boundary-nodes). It _should_ not make a difference, if we used a different one, like `text/plain`. However, this would afaik be untested.

src/internet_identity/src/assets.rs
```diff
@@ -132,5 +133,11 @@ fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType);
             ContentEncoding::Identity,
             ContentType::ICO,
         ),
+        (
```
nmattia, line 136: Can you add a line or two or a link to https://github.com/r-birkner/portal/blob/rjb/custom-domains-docs-v2/docs/developer-docs/production/custom-domain/custom-domain.md#custom-domains-on-the-boundary-nodes? thanks!

src/internet_identity/src/assets.rs
```diff
@@ -132,5 +133,11 @@ fn get_assets() -> [(&'static str, &'static [u8], ContentEncoding, ContentType);
...
+            "/.well-known/ic-domains",
+            include_bytes!("../../../dist/ic-domains"),
```
nmattia, line 138: how about inlining the content here? I think that might make things easier

src/internet_identity/src/assets.rs
nmattia: gotcha, thanks for checking though!

src/internet_identity/src/assets.rs
frederikrothenberger: Ah, of course. I didn't even think of that. 👍

nmattia (approved):
Thanks!

Technically, can this go in even if `identity.internetcomputer.org` is not live yet?

frederikrothenberger:
> Technically, can this go in even if `identity.internetcomputer.org` is not live yet?

Yes, but everything else required is apparently already in place.

PR #1231, opened 2023-02-16
Infer agent-js host

nmattia:
This pulls the host inference logic out of agent-js and into our codebase so that we can swap `ic0.app` with anything else if necessary without updating agent-js.


frederikrothenberger (approved):
Thanks!

PR #1232, opened 2023-02-16
Extract the official II url in frontend

nmattia:
This extracts `https://identity.ic0.app` into a constant so that we can change it easily in the frontend code.


src/frontend/src/config.ts
```diff
@@ -0,0 +1,7 @@
+/* Global configuration for II and constants */
+
+// The URL where the official, production II is served
+export const OFFICIAL_II_URL = "https://identity.ic0.app";
```
frederikrothenberger, line 4: Shouldn't this include `OFFICIAL_II_URL_NO_PROTOCOL`?

frederikrothenberger (approved):

src/frontend/src/config.ts
nmattia: Yeah first did that but it seems like a lot of noise for not much benefit

nmattia: actually no, I did the oppose: stripped `https://` from `OFFICIAL_II_URL` to make `OFFICIAL_II_URL_NO_PROTOCOL`. What you suggest is actually much nicer

PR #1233, opened 2023-02-16
Check alternative origins asset against new IC domain

frederikrothenberger:
This removes the dependency to ic0.app from the alternative origins feature. Additionally it somewhat loosens the restriction on the subdomain to be a canister ID (can now also be `"nns"`).


frederikrothenberger:
Currently blocked, dependent on canisters being available on icp0.io.

frederikrothenberger (commented):
Manually tested that `ic0.app` indeed still works using `https://kvusz-kaaaa-aaaad-aabwa-cai.ic0.app`.

👍  (cannot formally approve since it's my own PR)

PR #1236, opened 2023-02-16
Ignore compose setup in docker builds

nmattia:
The docker compose setup may depend on the docker builds, but not the other way around. By ignoring the compose setup we avoid unnecessary docker rebuilds on docker compose changes.


frederikrothenberger (approved):
👍

PR #1237, opened 2023-02-16
Add link-list for external links

meodai:
Composition of existing components to create a list of external links:

Screenshot 2023-02-16 at 15 28 43

I think it will look better with no c-card once placed.


meodai:
Will be replaced by #1410

PR #1238, opened 2023-02-16
add QR Code Component

meodai:
Comes with a "pulse" animation. Can be placed pretty much in every context and will always be square. 
Screenshot 2023-02-16 at 16 07 11


meodai:
I think this has become redundant because of #1409 . We could still choose to use a SVG implementation at some point though.

PR #1239, opened 2023-02-16
Enable II on identity.internetcomputer.org

nmattia:
This updates the CSP rules to allow II to be served on `https://identity.internetcomputer.org` in addition to `https://identity.ic0.app`.

Note that the code now refers to
`https://identity.internetcomputer.org`. The test infrastructure is also updated.


src/frontend/src/banner.ts
```diff
@@ -14,7 +14,10 @@ export const showWarningIfNecessary = (): void => {
         href="https://github.com/dfinity/internet-identity#build-features"
         >more</a
       >`);
-  } else if (window.location.origin !== OFFICIAL_II_URL) {
+  } else if (
+    window.location.origin !== OFFICIAL_II_URL &&
```
frederikrothenberger, line 18: We should remove this banner for the new _and_ the old domain. We are actively going to publish documentation that points people to the new II domain as soon as it is live.

Let's treat them both as official here.

PR #1243, opened 2023-02-17
Collect origin information on device registration

frederikrothenberger:
This PR adds origin information to newly registered devices, which
can be used to facilitate the upcoming domain migration.


src/internet_identity/src/storage/anchor.rs
```diff
@@ -300,6 +306,19 @@ fn check_device_limits(device: &Device) -> Result<(), AnchorError> {
             limit: CREDENTIAL_ID_LEN_LIMIT,
         });
     }
+
+    let n = device
+        .origin
+        .as_ref()
+        .map(|bytes| bytes.len())
+        .unwrap_or_default();
```
nmattia, line 314: nit: I find this a little confusing, I know `Option` is a functor but I always expect an array first and then do a double take. Also I'm guessing this is implicitly using `0` as the default value for naturals, so that if there's no `origin` then `n` is effectively `0`?

nmattia (approved):

src/internet_identity/src/storage/anchor.rs
frederikrothenberger: Yes. This is the same approach as is in use for the `credentialId`.

I would suggest that I refactor that later in a separate PR in both places. WDYT?

PR #1244, opened 2023-02-17
Use icp-api as the new IC api

nmattia:
This swaps the legacy `ic0.app` IC HTTP api endpoint with the new official `icp-api.io`.


frederikrothenberger (approved):
Thanks!

PR #1247, opened 2023-02-17
Always use legacy domain for principal derivation

frederikrothenberger:
In order to provide a stable principal across both, legacy and new domains without mandating every developer to use alternative origins, we always map back the derivation origin to the legacy domain.

### DIFF
--- a/src/frontend/src/flows/authorize/fetchDelegation.ts
+++ b/src/frontend/src/flows/authorize/fetchDelegation.ts
@@ -21,10 +21,20 @@ export const fetchDelegation = async (
   const sessionKey = Array.from(authContext.authRequest.sessionPublicKey);
 
   // at this point, derivationOrigin is either validated or undefined
-  const derivationOrigin =
+  let derivationOrigin =
     authContext.authRequest.derivationOrigin !== undefined
       ? authContext.authRequest.derivationOrigin
       : authContext.requestOrigin;
+
+  // In order to give dapps a stable principal regardless whether they use the legacy (ic0.app) or the new domain (icp0.io)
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/authorize/fetchDelegation.ts
```diff
@@ -5,6 +5,8 @@ import {
   SignedDelegation,
 } from "../../../generated/internet_identity_types";
 
+const ORIGIN_MAPPING_REGEX = /^https:\/\/([\w-]+)(?:\.raw)?\.icp0\.io$/;
```
nmattia, line 8: let's bring this one closer to where it's used, or even inline it

src/frontend/src/flows/authorize/fetchDelegation.ts
```diff
@@ -21,11 +23,18 @@ export const fetchDelegation = async (
...
+  let derivationOrigin =
     authContext.authRequest.derivationOrigin !== undefined
       ? authContext.authRequest.derivationOrigin
       : authContext.requestOrigin;
 
+  // In order to give dapps a stable principal regardless whether they use the legacy (ic0.app) or the new domain (icp0.io)
+  // we map back the derivation origin to the ic0.app domain.
+  const matches = ORIGIN_MAPPING_REGEX.exec(derivationOrigin);
+  if (matches !== null) {
+    derivationOrigin = `https://${matches[1]}.ic0.app`;
```
nmattia, line 35: how about using a named group, or at the very least move the `matches[1]` out of the string interpolation (just to clarify any potential errors)

nmattia (approved):

PR #1248, opened 2023-02-17
Extra-nudge users to add recovery on legacy

nmattia:
This adds another nudge on the anchor management screen to make sure users of ic0.app _do_ add a recovery phrase.


frederikrothenberger (approved):
Thanks!

PR #1250, opened 2023-02-20
Clean up manage page

nmattia:
This refactors the anchor management page by using more idiomatic lit constructs, and clarifying the distinction (in the code) between regular authentication devices and recovery devices.


meodai (approved):
I think its much cleaner

PR #1251, opened 2023-02-21
Don't expect alias in lookup in tests

nmattia:
The latest release includes the change where aliases are set to `""`. This updates the tests to use `get_anchor_info` instead of `lookup` for rollback tests.


peterpeterparker (approved):
Rubber stamp

PR #1253, opened 2023-02-21
Refactor displayManage

nmattia:
Previously `displayManage` would mix logic for figuring out devices and actual rendering. This extracts the rendering to `diplayManagePage`, making it more consistent with other pages.


peterpeterparker (approved):
LGTM, thanks

PR #1255, opened 2023-02-21
Clarify device origin on management page

nmattia:
This updates the `displayManage` page (i.e. anchor management page) to show information about device origin/validity.

Since the same device may have been added on two different domain, we now show a hint if two devices have the same name (see e.g. `(2)` in screenshots).

Also, if a device is found on a domain where it most likely would _not_ work (i.e. another `origin`) then we show a warning.

Tooltips in general are also fixed to work on mobile.


peterpeterparker (approved):
Rubber stamp

PR #1256, opened 2023-02-23
Fix tooltips

meodai:
Tooltips can now have a direction. 

https://user-images.githubusercontent.com/608386/220936355-2a29bc98-b914-42df-bf92-cb1feca7837b.mov

- Tooltips are not an ideal UI solution, we should think of other ways to present that information: https://inclusive-components.design/tooltips-toggletips/
- The direction is static or now. Meaning we need to set this class manually. If we keep using this pattern, we might need to add some JS at some point to make sure it opens somewhere it actually can be read.

### DIFF
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -1301,6 +1301,10 @@ a:hover,
   max-width: 20rem;
   transition: opacity 0.2s ease-in;
   background-color: var(--rc-background);
+
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+  hyphens: auto;
 }
 
 .c-tooltip--left .c-tooltip__message {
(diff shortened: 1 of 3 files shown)


src/frontend/src/styles/main.css
```diff
@@ -575,6 +575,11 @@ a:hover,
   padding: var(--rs-card-bezel-narrow);
 }
 
+.c-card--tight {
```
nmattia, line 578: `tight` is like `narrow` but takes up more horizontal space, is that it?

nmattia (approved):
Nice, thanks!

src/frontend/src/styles/main.css
meodai: tight is even tighter than narrow but when the space is too small it looks unequal in horizontally and vertically because of the lineheight. So we correct that with some uneven spacing

PR #1263, opened 2023-02-27
Update spec to reflect icp0 & migration changes

nmattia:
Some changes were made to the way we handle frontend hosts and pseudonym generation for the domain migration. This updates the spec to reflect those changes.


frederikrothenberger (approved):
Thanks!

PR #1265, opened 2023-02-27
Add domain warning tests

nmattia:
This adds some tests for the domain-specification warnings (based on device origins). The warning changes had to be merged quickly, and this is part of the clean up.

### DIFF
--- a/src/frontend/src/flows/manage/migration.test.ts
+++ b/src/frontend/src/flows/manage/migration.test.ts
@@ -1,0 +1,88 @@
+import { domainWarning } from "../manage";
+import { DeviceData } from "../../../generated/internet_identity_types";
+
+function onOrigin(origin: string, fn: () => void) {
+  const oldOrigin = window.origin;
+  Object.defineProperty(window, "origin", {
+    writable: true,
+    value: origin,
+  });
+
+  fn();
+  Object.defineProperty(window, "origin", {
+    writable: true,
+    value: oldOrigin,
+  });
+}
+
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/manage/migration.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  });
+});
+
+test("no origin on legacy domain is not warning", () => {
+  onOrigin("https://identity.ic0.app", () => {
+    expect(domainWarning({ ...authenticator, origin: [] })).toBe(undefined);
+  });
+});
+
+test("bad origin on legacy domain is warning", () => {
```
frederikrothenberger, line 60: I noticed that on `LEGACY_II_URL` devices added on `OFFICIAL_II_URL` get treated the same way as any other 'bad' URL.
For me this feels counter-intuitive. I.e. we specifically told people to do that in order to be prepared if `ic0.app` gets taken down. To then warn with a red exclamation mark is ...unexpected?
I think it is fine to inform the user, but I would suggest something that does not imply an action is necessary. I.e. a blue ℹ️ with "This device is bound to the new II domain and cannot be used on `identity.ic0.app`."

nmattia: Sure! Happy to discuss this offline, I don't have strong feelings about this. Resolving this since it's not directly relevant to the tests, but thanks for bringing it up!

frederikrothenberger (approved):

PR #1268, opened 2023-02-27
Enable e2e-tests on identity.ic0.app

nmattia:
This makes CI run all e2e tests on https://identity.ic0.app, in addition to those already running on https://identity.internetcomputer.org. We do this to ensure both domains are supported.


frederikrothenberger (approved):
Thanks!

PR #1271, opened 2023-02-28
Deduplicate alias prompt

nmattia:
There were two pages prompting for a device alias. This merges them into a single component. The `promptDeviceAlias` page was favored, as it was tested and had received extra CSS love.

The main difference between the two pages was the title, which is no passed as an argument. Some orphan comment about alias validation was moved to `validateAlias`.


src/frontend/src/components/alias/test.ts
```diff
@@ -1,10 +1,13 @@
-import { promptDeviceAliasPage } from "./alias";
+import { promptDeviceAliasPage } from "../alias";
 import { I18n } from "../../i18n";
 
 test("can be canceled", async () => {
   const cancel = jest.fn();
-  const ctn = jest.fn((str) => {});
+  const ctn = jest.fn((_str) => {
```
nmattia, line 6: the filename changed and eslint is now picking up this file

frederikrothenberger: Does this mean we should change the eslint config to be less picky about the filename?

PR #1273, opened 2023-02-28
Clean up device HTML

nmattia:
This modifies the device list HTML to be more useful from tests:
* An unused `div` wrapper is removed (both from authenticators and recoveries)
* The `data-device` attribute is hoisted to the top-level `li` element, so that sub-element can be selected more easily
* The "protected" `div` has a new `data-role="protected"` attribute so that tests can check that a device is indeed shown as protected
* Tests are updated to use `data-role="protected"`


meodai (approved):
Much nicer to read!

PR #1274, opened 2023-02-28
Move deviceListItem and recoverMethods to a component

meodai:
To make it easier to do future tasks, I wanted to create components out of those two. To make it more manageable to review my code in the future, I thought I'd create a PR with just this.

### DIFF
--- a/src/frontend/src/components/deviceListItem.ts
+++ b/src/frontend/src/components/deviceListItem.ts
@@ -1,0 +1,84 @@
+import { TemplateResult, html } from "lit-html";
+import { Setting, deviceSettings } from "../flows/manage/deviceSettings";
+import { warningIcon, dropdownIcon, lockIcon } from "./icons";
+
+// A simple representation of "device"s used on the manage page.
+export type Device = {
+  // All the settings allowed for a particular device
+  settings: Setting[];
+  // The displayed name of a device (not exactly the "alias") because
+  // recovery devices handle aliases differently.
+  label: string;
+  isRecovery: boolean;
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/recoveryMethods.ts
```diff
@@ -0,0 +1,49 @@
+import { TemplateResult, html } from "lit-html";
+import { deviceListItem, Device } from "./deviceListItem";
+
+// The list of recovery devices
+export const recoveryMethods = ({
```
nmattia, line 5: Why export `recoverySection` but not `devicesSection`?

src/frontend/src/components/deviceListItem.ts
```diff
@@ -0,0 +1,84 @@
...
+  // recovery devices handle aliases differently.
+  label: string;
+  isRecovery: boolean;
+  isProtected: boolean;
+  warn?: TemplateResult;
+};
+
+// A device with extra information about whether another device (earlier in the list)
+// has the same name.
+export type DedupDevice = Device & { dupCount?: number };
```
nmattia, line 19: there's something odd here, shouldn't this live closer to `dedupLabels`? Feels weird to have the type in one module and the function in another one

nmattia (commented):
I'm not sure what you mean by "component". Do you mean modules? Are we planning on showing devices in other modules than `manage`?

meodai:
> 

No, but to attach some logic to it (nag user when not all recovery methods are used)

nmattia (approved):
Nice thanks!

PR #1275, opened 2023-02-28
Remove unused argument & clean up

nmattia:
There was a bit of copypasta between `removeDevice` and `protectDevice` and `protectDevice` had an unused argument. This also deconstructs args for consistency with other modules.


meodai (approved):
LGTM, but why don't we need "isOnlyDevice" anymore?

nmattia:
It was unused! It's being used in the calling function though, but the logic is embedded in one of the callbacks we pass

PR #1276, opened 2023-02-28
Allow unprotecting recovery phrases

nmattia:
This adds an option on the management page to unprotect recovery phrases. Upon unprotection, the user is prompted to enter their recovery phrase. After that, the device is unprotected and can be deleted at will.


meodai (approved):
LGTM

PR #1277, opened 2023-02-28
Enable require-await lint

nmattia:
This ensures that only functions making used of `await` as marked as `async`.


peterpeterparker (approved):
Thanks

PR #1279, opened 2023-02-28
Checkbox-Type Recovery-Methods

meodai:


meodai:
Screenshot 2023-02-28 at 20 47 53
Screenshot 2023-02-28 at 20 47 10

meodai:
@nmattia I am done with the UX part of this I think! Can you take over for the "we don't have a single call to action that says: add recovery method, but separate actions for each of them"? (the tests need to be adapted as well) The tricky part will be to have to "open" the collapsed state when there are at least two methods added

meodai:
I will get a new icon from design for the lock and unlocked state soon. In the mean time I need help with the following:

The test fails because `setupRecovery` can't be called anymore (since instead of having one call to action to setup a generic recoveryMethod we now know if a person wants to set up a device or phrase right away), the `setupRecovery` is not documented. I struggle to adapt separate the flow.

The component I build is "ready" to take separate functions for each method.

As discussed with @nmattia it would be nice to separate devices from recoveries. It made sense in the past because we handled it in the same way. 

It the mean time it would be super nice if @peterpeterparker could have at the code I wrote?

src/frontend/src/components/deviceListItem.ts
```diff
@@ -1,17 +1,32 @@
...
+  successIcon,
+  dropdownIcon,
+  lockIcon,
+} from "./icons";
 
 // A simple representation of "device"s used on the manage page.
 export type Device = {
   // All the settings allowed for a particular device
-  settings: Setting[];
+  settings?: Setting[];
```
peterpeterparker, line 14: is this needed? looks like empty `[]` is used in the code when there is no settings

src/frontend/src/components/deviceListItem.ts
```diff
@@ -1,17 +1,32 @@
...
   // The displayed name of a device (not exactly the "alias") because
   // recovery devices handle aliases differently.
-  label: string;
-  isRecovery: boolean;
-  isProtected: boolean;
-  warn?: TemplateResult;
+  label: string | TemplateResult;
+  recovery?: "phrase" | "device";
+  isProtected?: boolean;
```
peterpeterparker, line 19: what's `isProtected` undefined? does `undefined` has a different meaning that `false`?

src/frontend/src/components/deviceListItem.ts
```diff
@@ -1,17 +1,32 @@
...
+  status?: {
```
peterpeterparker, line 20: instead of `status?` wouldn't be safer to have always a `status` but with an initial `statusType: "init"`?

src/frontend/src/components/deviceListItem.ts
```diff
@@ -1,17 +1,32 @@
...
+    statusType: "error" | "warning" | "ok";
+    statusText?: TemplateResult;
+  };
+};
+
+const iconByStatusType = {
```
peterpeterparker, line 26: nth: can you provide its type?

src/frontend/src/components/deviceListItem.ts
```diff
@@ -21,42 +36,45 @@ export type DedupDevice = Device & { dupCount?: number };
...
   device: DedupDevice;
   index: string;
+  block?: TemplateResult;
 }) => {
   return html`
     <li class="c-action-list__item" data-device=${device.label}>
+      ${device.status !== undefined
+        ? html`<div class="c-action-list__action">
+            <span
+              class="c-tooltip c-icon c-icon--${device.status.statusType}"
```
peterpeterparker, line 50: would it be worth to add a comment somewhere to specify that the type matches a CSS style? (to avoid changing one and not the other in the future)

src/frontend/src/components/deviceListItem.ts
```diff
@@ -21,42 +36,45 @@ export type DedupDevice = Device & { dupCount?: number };
...
+          </div>`
+        : undefined}
       <div class="c-action-list__label">
         ${device.label}
         ${device.dupCount !== undefined && device.dupCount > 0
           ? html`<i class="t-muted">&nbsp;(${device.dupCount})</i>`
           : undefined}
       </div>
-      ${device.isProtected
+      ${device.isProtected !== undefined && device.isProtected
```
peterpeterparker, line 67: nth: `isProtected === true` does both

src/frontend/src/components/deviceListItem.ts
```diff
@@ -21,42 +36,45 @@ export type DedupDevice = Device & { dupCount?: number };
...
-              tabindex="0"
-              >${warningIcon}<span
-                class="c-tooltip__message c-card c-card--tight"
-                >${device.warn}</span
-              ></span
-            >
-          </div>`
-        : undefined}
-      ${device.settings.length > 0
+      ${block !== undefined ? block : undefined}
```
peterpeterparker, line 76: does it throw and error if you just do `${block}`? look the same codewise

src/frontend/src/components/icons.ts
```diff
@@ -100,6 +100,23 @@ export const closeIcon = html`
...
+      fill="#F5B400"
```
peterpeterparker, line 114: you don't want to use css variables for the colors?

src/frontend/src/components/recoveryMethodsSection.ts
```diff
@@ -1,6 +1,85 @@
...
+
+const sectionList = ({
+  recoveries,
+  onAddRecovery,
+}: {
+  recoveries: Device[];
+  onAddRecovery?: () => void;
+}): TemplateResult => {
+  console.log(recoveries);
+  const recoveryPhrases = recoveryByType({ type: "phrase", recoveries });
```
peterpeterparker, line 20: I guess per doesn't matter much but if it would, you can probably do this in one shot with a reducer. Something like:

```
const {phrases, devices} = recovries.reduce((acc, device) => {
  if (device.recovery === "phrase") {
     return {phrases: [...acc.phrases], devices: [...acc.devices, device]}
  }

     return {phrases: [...acc.phrases, device], devices: [...acc.devices]}  
}, {phases: [], devices []})
```

Just sharing in case you like it the idea.

src/frontend/src/components/recoveryMethodsSection.ts
```diff
@@ -1,6 +1,85 @@
...
+                index: `recovery-phrase-${index}`,
+              })
+            )
+          : deviceListItem({
+              device: {
+                label: "Recovery Phrase",
+                recovery: "phrase",
+                status: {
+                  statusType: "error",
+                  statusText: html`We strongly recommend that you use a recovery
```
peterpeterparker, line 38: i18n labels?

src/frontend/src/components/recoveryMethodsSection.ts
```diff
@@ -1,6 +1,85 @@
...
+                @click="${onAddRecovery}"
+                class="c-button c-button--primary c-button--minimal"
+                id="addRecoveryPhrase"
+                aria-label="Add recovery phrase"
+              >
+                Enable
+              </button>`,
+            })}
+        ${recoveryDevices.length > 0
+          ? recoveryDevices.map((device, index) =>
```
peterpeterparker, line 53: this piece of code looks similar to previous, worth extracting a function?

src/frontend/src/components/deviceListItem.ts
meodai: @peterpeterparker yeah only phrases can be protected, ok I will separate devices form recovery methods

src/frontend/src/components/deviceListItem.ts
meodai: these are exceptions, we use this to show some kind of notification or error on each device or recovery method.

src/frontend/src/components/deviceListItem.ts
meodai: ah yes definitely!

src/frontend/src/components/deviceListItem.ts
meodai: yes. when we want you to add a new device, there are no settings passed. Do you think its nicer to alway pass empty settings everywhere?

src/frontend/src/components/deviceListItem.ts
meodai: I added a comment on the type definition

src/frontend/src/components/recoveryMethodsSection.ts
```diff
@@ -11,32 +90,19 @@ export const recoveryMethodsSection = ({
...
-                  @click="${onAddRecovery}"
-                  class="c-button c-button--primary"
-                  id="addRecovery"
-                >
-                  Add recovery method
-                </button>
-              </div>
-            </div>
-          `}
+      <details class="c-details" .open="${recoveries.length < 2}">
```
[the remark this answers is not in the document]
meodai: I just did! thanks!

src/frontend/src/styles/main.css
```diff
@@ -1227,13 +1274,29 @@ a:hover,
   left: calc(-1 * var(--vs-inline) / 2);
   text-decoration: none;
   transform: translateX(-100%);
-  transition: 200ms cubic-bezier(0.3, 0.7, 0, 1);
+  transition: 300ms cubic-bezier(0.3, 0.7, 0, 1);
```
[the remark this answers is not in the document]
meodai: it is, its the recovery methods that can be collapsed, but i just noticed that the transition property is missing! Thanks for pointing it out

nmattia:
@meodai is this still a draft? or should it be reviewed?

> In the mean time I need help with the following:
> 
> The test fails because setupRecovery can't be called anymore (since instead of having one call to action to setup a generic recoveryMethod we now know if a person wants to set up a device or phrase right away), the setupRecovery is not documented. I struggle to adapt separate the flow.


there's a lot of questions about the code here (should `settings` really be optional, etc). As discussed last week @meodai please focus on the HTML & CSS and I'll take care of the rest

meodai:
@nmattia I would say it is from a CSS & HTML perspective now. I have added some comments to the CSS and introduces the concept of a "c-action-list__status" that is district from the actions. I am not a 100% happy with the names "status" 

I just checked with our Design. And they like the idea of always showing the status in front (also for the domain migration warning)

Screenshot 2023-03-20 at 15 20 30

They also told me that the whole section does not need to be collapsible anymore. 

So from and HTML & CSS perspective we are good. Let me remove the  "collapsible" part and ping you when I'm done.

meodai:
@nmattia I just removed the collapsing part, so I think the HTML & CSS are good.

nmattia:
Superseded by #1356

PR #1280, opened 2023-03-01
Introduce get_anchor_credentials method as a replacement for lookup

frederikrothenberger:
`lookup` provides more information than necessary for a public endpoint.
This is why it is being deprecated. As a replacement `get_device_credentials` is introduced
which is more tailored to just provide the information required for authentication.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -216,6 +216,17 @@ type IdentityAnchorInfo = record {
     device_registration: opt DeviceRegistrationInfo;
 };
 
+type GetCredentialsResponse = record {
+    credentials : vec Credential;
+    recovery_credentials : vec Credential;
+    recovery_phrase: bool;
+};
+
+type Credential = record {
+    credentialId : CredentialId;
+    pubkey: PublicKey;
+};
+
 type DeployArchiveResult = variant {
     // The archive was deployed successfully and the supplied wasm module has been installed. The principal of the archive
     // canister is returned.
@@ -245,8 +256,9 @@ service : (opt InternetIdentityInit) -> {
     remove : (UserNumber, DeviceKey) -> ();
     // Returns all devices of the user (authentication and recovery) but no information about device registrations.
     // Note: Clears out the 'alias' fields on the devices. Use 'get_anchor_info' to obtain the full information.
-    // Note: Will be changed in the future to be more consistent with get_anchor_info.
+    // Deprecated: Use 'get_credentials' instead.
     lookup : (UserNumber) -> (vec DeviceData) query;
+    get_credentials : (UserNumber) -> (GetCredentialsResponse) query;
     get_anchor_info : (UserNumber) -> (IdentityAnchorInfo);
     get_principal : (UserNumber, FrontendHostname) -> (principal) query;
     stats : () -> (InternetIdentityStats) query;
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -216,13 +216,13 @@ type IdentityAnchorInfo = record {
     device_registration: opt DeviceRegistrationInfo;
 };
 
-type GetDeviceCredentialsResponse = record {
-    credentials : vec DeviceCredential;
-    recovery_credentials : vec DeviceCredential;
+type AnchorCredentials = record {
+    credentials : vec WebauthnCredential;
+    recovery_credentials : vec WebauthnCredential;
     recovery_phrase: bool;
 };
 
-type DeviceCredential = record {
+type WebauthnCredential = record {
     credentialId : CredentialId;
     pubkey: PublicKey;
 };
@@ -256,9 +256,9 @@ service : (opt InternetIdentityInit) -> {
     remove : (UserNumber, DeviceKey) -> ();
     // Returns all devices of the user (authentication and recovery) but no information about device registrations.
     // Note: Clears out the 'alias' fields on the devices. Use 'get_anchor_info' to obtain the full information.
-    // Deprecated: Use 'get_device_credentials' instead.
+    // Deprecated: Use 'get_anchor_credentials' instead.
     lookup : (UserNumber) -> (vec DeviceData) query;
-    get_device_credentials : (UserNumber) -> (GetDeviceCredentialsResponse) query;
+    get_anchor_credentials : (UserNumber) -> (AnchorCredentials) query;
     get_anchor_info : (UserNumber) -> (IdentityAnchorInfo);
     get_principal : (UserNumber, FrontendHostname) -> (principal) query;
     stats : () -> (InternetIdentityStats) query;
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -219,7 +219,7 @@ type IdentityAnchorInfo = record {
 type AnchorCredentials = record {
     credentials : vec WebauthnCredential;
     recovery_credentials : vec WebauthnCredential;
-    recovery_phrase: bool;
+    recovery_phrase: vec PublicKey;
 };
 
 type WebauthnCredential = record {
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -216,13 +216,13 @@ type IdentityAnchorInfo = record {
     device_registration: opt DeviceRegistrationInfo;
 };
 
-type GetCredentialsResponse = record {
-    credentials : vec Credential;
-    recovery_credentials : vec Credential;
+type GetDeviceCredentialsResponse = record {
+    credentials : vec DeviceCredential;
+    recovery_credentials : vec DeviceCredential;
     recovery_phrase: bool;
 };
 
-type Credential = record {
+type DeviceCredential = record {
     credentialId : CredentialId;
     pubkey: PublicKey;
 };
@@ -258,7 +258,7 @@ service : (opt InternetIdentityInit) -> {
     // Note: Clears out the 'alias' fields on the devices. Use 'get_anchor_info' to obtain the full information.
     // Deprecated: Use 'get_credentials' instead.
     lookup : (UserNumber) -> (vec DeviceData) query;
-    get_credentials : (UserNumber) -> (GetCredentialsResponse) query;
+    get_device_credentials : (UserNumber) -> (GetDeviceCredentialsResponse) query;
     get_anchor_info : (UserNumber) -> (IdentityAnchorInfo);
     get_principal : (UserNumber, FrontendHostname) -> (principal) query;
     stats : () -> (InternetIdentityStats) query;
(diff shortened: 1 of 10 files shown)


src/internet_identity_interface/src/lib.rs
```diff
@@ -188,6 +188,19 @@ impl IdentityAnchorInfo {
     }
 }
 
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub struct Credential {
```
przydatek, line 192: I wonder whether a more specific/narrow name would be more appropriate here, as `Credential` is quite generic, and in a not-so-distant future we may add support for **verifiable credentials** (in the context of attribute sharing), so maybe we could avoid a potential future name collision.   IIUC, this is specific to a device, so maybe `DeviceCredential` would make sense?  (Note: if we change the struct name, then probably we'd also rename the related structs and functions, like `GetCredentialsResponse` or `get_credentials`). Wdyt?

src/internet_identity/src/storage/anchor.rs
```diff
@@ -164,6 +167,20 @@ impl Anchor {
         &self.devices
     }
 
+    /// Returns matching [Credential]s.
```
przydatek, line 170: nit: maybe add a description of what happens when `filter` is `None`?
Alternatively, maybe drop `Option`, and leave it up to the caller to provide the dummy "catch-all" filter?  Maybe then also `Box` would be not needed?

src/internet_identity/src/storage/anchor.rs
```diff
@@ -85,6 +85,9 @@ impl From<Device> for DeviceDataWithoutAlias {
     }
 }
 
+/// Shorthand type for a predicate when iterating over the devices.
+pub type DeviceFilter = Box<dyn FnMut(&&Device) -> bool>;
```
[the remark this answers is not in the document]
przydatek: Why is it a `FnMut` rather than eg. `Fn(&Device) -> bool`?

przydatek (commented):
Hi @frederikrothenberger, please see some questions comments below.
As this is my first review for II, please take the comments/suggestions with a grain of salt ;-)

src/internet_identity/src/storage/anchor.rs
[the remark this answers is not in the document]
frederikrothenberger: `filter` uses `FnMut` so I opted for the same type. But happy to change.

src/internet_identity/src/main.rs
```diff
@@ -98,6 +98,7 @@ fn remove(anchor_number: AnchorNumber, device_key: DeviceKey) {
 /// Returns all devices of the anchor (authentication and recovery) but no information about device registrations.
 /// Note: Will be changed in the future to be more consistent with get_anchor_info.
 #[query]
+#[deprecated(note = "please use `get_credentials` instead")]
```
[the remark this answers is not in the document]
frederikrothenberger: I removed it for not being useful.

src/internet_identity/src/storage/anchor.rs
frederikrothenberger: Yeah, I think the whole interface gets a lot nicer if the `filter` is mandatory.

src/internet_identity_interface/src/lib.rs
frederikrothenberger: I have renamed the interface. I don't mind the longer names. @nmattia wdyt?

przydatek (approved):
LGTM, thanks!
(Of course, let's wait for @nmattia's feedback)

src/internet_identity_interface/src/lib.rs
frederikrothenberger: Having thought about this for a bit, I think `get_webauthn_credentials` would be most accurate, if not for the fact that we also have a bit of information in there about the existence recovery phrases. But at least it would be more specific than just `device`.

src/internet_identity/src/storage/anchor.rs
```diff
@@ -164,6 +164,25 @@ impl Anchor {
         &self.devices
     }
 
+    /// Returns matching [DeviceCredential]s.
+    pub fn get_device_credentials<P>(&self, filter: P) -> Vec<DeviceCredential>
+    where
+        P: Fn(&&Device) -> bool,
```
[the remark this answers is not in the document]
frederikrothenberger: If I only do `&Device` then I cannot do `.filter(filter)` but would have to first dereference `device`, i.e. something like `.filter(|&d| filter(d))`.

src/internet_identity/internet_identity.did
```diff
@@ -244,8 +255,9 @@ service : (opt InternetIdentityInit) -> {
     remove : (UserNumber, DeviceKey) -> ();
     // Returns all devices of the user (authentication and recovery) but no information about device registrations.
     // Note: Clears out the 'alias' fields on the devices. Use 'get_anchor_info' to obtain the full information.
-    // Note: Will be changed in the future to be more consistent with get_anchor_info.
+    // Deprecated: Use 'get_device_credentials' instead.
     lookup : (UserNumber) -> (vec DeviceData) query;
+    get_device_credentials : (UserNumber) -> (GetDeviceCredentialsResponse) query;
```
nmattia, line 260: I would suggest naming this `get_credentials` (or if really necessary `get_anchor_credentials`), I started looking at this assuming it would somehow only list the credentials for a device

src/internet_identity/internet_identity.did
```diff
@@ -215,6 +215,17 @@ type IdentityAnchorInfo = record {
     device_registration: opt DeviceRegistrationInfo;
 };
 
+type GetDeviceCredentialsResponse = record {
```
nmattia, line 218: This is very descriptive, have you considered something like `Credentials` or `CredentialsInfo` or `AnchorCredentials`?

src/internet_identity/internet_identity.did
```diff
@@ -215,6 +215,17 @@ type IdentityAnchorInfo = record {
...
+    credentials : vec DeviceCredential;
+    recovery_credentials : vec DeviceCredential;
+    recovery_phrase: bool;
```
nmattia, line 221: nit: thinking about making things a bit more correct by construction. What happens if `{ recovery_credentials = [], recovery_phrase = true }`?

src/internet_identity/src/main.rs
```diff
@@ -115,6 +115,24 @@ fn lookup(anchor_number: AnchorNumber) -> Vec<DeviceData> {
     })
 }
 
+#[query]
+fn get_device_credentials(anchor_number: AnchorNumber) -> GetDeviceCredentialsResponse {
+    let anchor = state::anchor(anchor_number);
+
+    let credentials = anchor.get_device_credentials(|d| d.purpose == Purpose::Authentication);
+    let recovery_credentials = anchor.get_device_credentials(|d| d.purpose == Purpose::Recovery);
```
nmattia, line 123: Why not fetch all devices and partition them? If a new `Purpose` is introduced, we may forget to fetch it here

src/internet_identity/src/storage/anchor.rs
```diff
@@ -164,6 +164,25 @@ impl Anchor {
         &self.devices
     }
 
+    /// Returns matching [DeviceCredential]s.
+    pub fn get_device_credentials<P>(&self, filter: P) -> Vec<DeviceCredential>
```
nmattia, line 168: I think it may be simpler to do everything in `get_query_credentials`, I find it a little odd that this module ("storage") needs to know about extracting credentials

src/internet_identity/tests/archive_integration_tests_pull.rs
```diff
@@ -262,7 +262,7 @@ mod pull_entries_tests {
             operation: Operation::RegisterAnchor {
                 device: DeviceDataWithoutAlias {
                     pubkey: device_data_1().pubkey,
-                    credential_id: None,
+                    credential_id: device_data_1().credential_id,
```
nmattia, line 265: out of curiosity, why did this change?

src/internet_identity/internet_identity.did
nmattia: Err, I just realized there are no credentials for recovery _phrases_. So I'm guessing the recovery credentials are only for "recovery keys"?

frederikrothenberger: Yes, exactly. So your example would indicate no recovery devices but the existence of a recovery phrase.

src/internet_identity/internet_identity.did
frederikrothenberger: Makes sense. I have renamed it `get_anchor_credentials`.

src/internet_identity/src/main.rs
frederikrothenberger: TIL about `partition`. Thanks for pointing it out.

src/internet_identity/src/storage/anchor.rs
frederikrothenberger: I originally had it that way, but without `partition` the code was very cumbersome. But since I now know it exists, I have removed this method again.

src/internet_identity/tests/archive_integration_tests_pull.rs
frederikrothenberger: I introduced a credential id on the `device_data_1()` test data (to make it a proper webauthn device), so the archive test that uses it now also archives the credential id.

src/internet_identity/internet_identity.did
```diff
@@ -215,6 +215,17 @@ type IdentityAnchorInfo = record {
     device_registration: opt DeviceRegistrationInfo;
 };
 
+type AnchorCredentials = record {
+    credentials : vec WebauthnCredential;
+    recovery_credentials : vec WebauthnCredential;
+    recovery_phrase: bool;
```
nmattia, line 221: really nit: since we send the pubkeys everywhere, does it make sense to send it for the recovery phrase as well?

nmattia (approved):
Thanks a lot!

I didn't think that we may not be able to build a WebAuthn device since phrases don't have credentials. This is the ultimate nit so feel free to ignore, but couldn't we get rid of the `TryFrom` by folding over the devices once?

```rust
#[query]
fn get_foo(anchor_number: AnchorNumber) -> AnchorCredentials {
    let anchor = state::anchor(anchor_number);

    struct Foo { auth: Vec<WebauthnCredential>, recover: Vec<WebauthnCredential>, phrase: bool }

    let foo = anchor.into_devices().into_iter().fold(Foo { auth: vec![], recover: vec![], phrase: false}, |mut acc, d| {
        if d.key_type == KeyType::SeedPhrase {
            acc.phrase = true;
        } else if let Some(credential_id) = d.credential_id {
            if d.purpose == Purpose::Recovery {
                acc.recover.push(WebauthnCredential { pubkey: d.pubkey, credential_id })
            } else {
                acc.auth.push(WebauthnCredential { pubkey: d.pubkey, credential_id })
            }
        }
        acc
    });

    AnchorCredentials {
        credentials: foo.auth,
        recovery_credentials: foo.recover,
        recovery_phrase: foo.phrase,
    }
}
```

src/internet_identity/internet_identity.did
frederikrothenberger: I made it a `Vec<PubKey>` because there might still be anchors with multiple recovery phrases (from before we enforced the limit).

nmattia (approved):
Looking really nice, thanks!

PR #1281, opened 2023-03-01
Refactor HTTP gateway interface types

frederikrothenberger:
This PR separates the types used to implement the `http_request`
and `http_request_update` canister methods into their own module.

Additionally the `upgrade` flag is introduce in preparation for the fix
the /faq redirect (which will need to set that flag).

### DIFF
--- a/src/internet_identity_interface/src/http_gateway.rs
+++ b/src/internet_identity_interface/src/http_gateway.rs
@@ -1,3 +1,6 @@
+//! Types as defined by the HTTP gateway spec.
+//! See https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-gateway-interface
+
 use candid::{CandidType, Deserialize, Func};
 use serde_bytes::{ByteBuf, Bytes};
 use std::borrow::Cow;
--- a/src/internet_identity_interface/src/http_gateway.rs
+++ b/src/internet_identity_interface/src/http_gateway.rs
@@ -1,0 +1,36 @@
+use candid::{CandidType, Deserialize, Func};
+use serde_bytes::{ByteBuf, Bytes};
+use std::borrow::Cow;
+
+pub type HeaderField = (String, String);
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub struct Token {}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub enum StreamingStrategy {
+    Callback { callback: Func, token: Token },
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub struct StreamingCallbackHttpResponse {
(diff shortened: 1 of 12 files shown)


src/internet_identity_interface/src/lib.rs
```diff
@@ -1,9 +1,11 @@
-use candid::{CandidType, Deserialize, Func, Principal};
-use serde_bytes::{ByteBuf, Bytes};
-use std::borrow::Cow;
+use candid::{CandidType, Deserialize, Principal};
+use serde_bytes::ByteBuf;
 
 /// types related to the archive
 pub mod archive;
+/// types as specified by the HTTP gateway protocol.
+/// See https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-gateway-interface
```
przydatek, line 7: nit: maybe use https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-gateway instead?  (it points to the main section on the HTTP gateway protocol, rather than to a subsection)

src/internet_identity_interface/src/http_gateway.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub headers: Vec<(String, String)>,
+    pub body: ByteBuf,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize)]
+pub struct HttpResponse {
+    pub status_code: u16,
+    pub headers: Vec<HeaderField>,
+    pub body: Cow<'static, Bytes>,
+    pub upgrade: Option<bool>,
```
przydatek, line 34: What is the meaning of the flag, i.e. what does it mean when it is true vs. when it is false? Will  in the future some value be assumed as a default when the value is `None`?  Is it worth documenting the semantics in a comment?  (maybe the meaning is clear for a knowledgable reader, so sorry if the question is stupid)

przydatek:
LGTM, but please see some questions below.

src/internet_identity_interface/src/lib.rs
frederikrothenberger: I specifically linked to the candid section, because this is where the types are defined that are represented here.

src/internet_identity_interface/src/http_gateway.rs
frederikrothenberger: Given the semantics of these types are quite complex (also with regards to streaming) I'd rather just delegate to the spec that explains all those fields.
So I suggest to also add a link to the spec in this file.

src/internet_identity_interface/src/lib.rs
frederikrothenberger: I have now changed the link to make it consistent with the doc in the other file. 👍

przydatek (approved):
LGTM, thanks!

PR #1283, opened 2023-03-02
Fix /faq redirect

frederikrothenberger:
This PR fixes the /faq link redirect by setting the upgrade to update flag on the redirect and allowing http_requests to be sent as update calls.

The redirects can be certified for query calls with the V2 certification at which point the support to receive http assets via update call can be dropped again.

### DIFF
--- a/src/internet_identity/src/http.rs
+++ b/src/internet_identity/src/http.rs
@@ -39,6 +39,9 @@ pub fn http_request(req: HttpRequest) -> HttpResponse {
                     .to_string(),
             )],
             body: Cow::Owned(ByteBuf::new()),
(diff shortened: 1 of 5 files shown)


src/internet_identity/src/http.rs
```diff
@@ -39,7 +39,7 @@ pub fn http_request(req: HttpRequest) -> HttpResponse {
                     .to_string(),
             )],
             body: Cow::Owned(ByteBuf::new()),
-            upgrade: None,
+            upgrade: Some(true),
```
nmattia, line 42: nit: Can you add a quick note here?

nmattia (approved):
Minor comment

PR #1284, opened 2023-03-02
Only allow resetting recovery phrases

nmattia:
In order to avoid users forgetting to add a new recovery phrase after removing an old one, we only allow _resetting_, meaning a new phrase is created upon removal. The `remove` setting is replaced with a `reset` setting.

Note:
* this is only a webapp matter and is not enforced by the canister.
* the phrase is replaced _before_ the user has confirmed they have copied the new phrase (but _after_ they have confirmed they want to reset it). This is not ideal and will be changed subsequently since the behavior also needs to be changed upon initial phrase creation.
* the phrase is not actually removed and a new one created, but `replace`d atomically.

### DIFF
--- a/src/frontend/src/flows/manage/deviceSettings.ts
+++ b/src/frontend/src/flows/manage/deviceSettings.ts
@@ -13,8 +13,12 @@ import {
   isRecoveryDevice,
   isRecoveryPhrase,
   isProtected,
-  RecoveryDevice,
+  RecoveryPhrase,
 } from "../../utils/recoveryDevice";
+import { generate } from "../../crypto/mnemonic";
+import { fromMnemonicWithoutValidation } from "../../crypto/ed25519";
+import { IC_DERIVATION_PATH } from "../../utils/iiConnection";
+import { displaySeedPhrase } from "../recovery/displaySeedPhrase";
 
 // A particular device setting, e.g. remove, protect, etc
 export type Setting = { label: string; fn: () => void };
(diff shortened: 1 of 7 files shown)


src/frontend/src/flows/manage/deviceSettings.ts
```diff
@@ -36,53 +40,47 @@ export const deviceSettings = ({
...
-
-  // Whether the device can be unprotected or not
-  if (shouldOfferToUnprotect(device)) {
-    settings.push({
-      label: "unprotect",
-      fn: () =>
-        unprotectDevice(userNumber, connection, device, isOnlyDevice, reload),
-    });
+  // Whether the device can be (un)protected or not (only recovery phrases can be protected)
+  if (isRecoveryPhrase(device)) {
```
nmattia, line 44: I've inlined the "shouldOfferToProtect" checks because it wasn't clear what was happening

src/frontend/src/flows/manage/deviceSettings.ts
```diff
@@ -116,12 +114,10 @@ const deviceConnection = async (
 
 /* Remove the device and return */
 const deleteDevice = async ({
-  userNumber,
```
nmattia, line 119: Some logic was there only for protected phrases; by not allowing phrases here anymore things can be simplified a bit

src/frontend/src/flows/manage/deviceSettings.ts
```diff
@@ -178,6 +157,69 @@ const deleteDevice = async ({
...
+  reload,
+}: {
+  userNumber: bigint;
+  connection: AuthenticatedConnection;
+  device: DeviceData & RecoveryPhrase;
+  reload: () => void;
+}) => {
+  const confirmed = confirm(
+    "Reset your recovery phrase\n\nWas your recovery phrase compromised? Delete your recovery phrase and generate a new one by confirming."
+  );
```
nmattia, line 174: We make do with `confirm` for now but nicer modals by @meodai are on the way

src/frontend/src/flows/manage/deviceSettings.ts
```diff
@@ -178,6 +157,69 @@ const deleteDevice = async ({
...
+  }
+
+  // Save the old pubkey (used as index for replace)
+  const oldKey = device.pubkey;
+  device.pubkey = Array.from(
+    new Uint8Array(recoverIdentity.getPublicKey().toDer())
+  );
+
+  try {
+    await withLoader(() => newConnection.replace(oldKey, device));
```
frederikrothenberger, line 209: This removes the recovery phrase. Now, if you did recover using this recovery phrase and then reset it, it will cause an error _after_ the reset because the user is no longer authenticated to display their devices. I.e. this:
Screenshot 2023-03-02 at 13 11 16

It would be nice to have also the "you will be logged out warning" here and then show the home screen instead.

frederikrothenberger:
> the phrase is replaced before the user has confirmed they have copied the new phrase (but after they have confirmed they want to reset it). This is not ideal and will be changed subsequently since the behavior also needs to be changed upon initial phrase creation.

Does this mean that the replace will be moved to after the confirmation in the future?
I.e.
1. confirm reset
2. display new phrase
3. input new phrase
4. only now make a change to the anchor

src/frontend/src/flows/manage/deviceSettings.ts
nmattia: Good catch! Since we have the key locally we may simply want to use that as the connection. Will give this a think.

nmattia:
> Does this mean that the replace will be moved to after the confirmation in the future?

Yes, something like that! Though nothing is set in stone yet. Do you think this is not a good idea?

frederikrothenberger:
> > Does this mean that the replace will be moved to after the confirmation in the future?
> 
> Yes, something like that! Though nothing is set in stone yet. Do you think this is not a good idea?

I think it would be better than the current flow. 👍 
It is probably also the user expectations that changes are only made after confirmation.

src/frontend/src/flows/manage/deviceSettings.ts
frederikrothenberger: Ah, that's a possibility too of course. 👍

nmattia: @frederikrothenberger I've changed it to authenticate the user with the _new_ phrase iff the user was authenticated with the recovery phrase. PTAL!

frederikrothenberger (approved):
Tested and worked a charm 👍

PR #1285, opened 2023-03-02
Mention protected flag in spec

nmattia:
The spec was incorrect re. removal of protected devices. This adds a note mentioning that the user must be authenticated with the protected device they want to remove in order to remove it.


frederikrothenberger (approved):
Thanks!

PR #1286, opened 2023-03-02
Offer to enroll device upon recovery

nmattia:
This adds a prompt right after the user has recovered their anchor, asking them if they want to enroll a new authenticator. The prompt offers to "skip" if the user does not want to enroll a new device.

_If_ the user decides to enroll a new device, then a new WebAuthn identity is constructed and `add`ed to the anchor.

### DIFF
--- a/src/frontend/src/flows/recovery/useRecovery.ts
+++ b/src/frontend/src/flows/recovery/useRecovery.ts
@@ -1,11 +1,15 @@
 import { displayError } from "../../components/displayError";
-import { Connection } from "../../utils/iiConnection";
+import { Connection, AuthenticatedConnection } from "../../utils/iiConnection";
 import { renderManage } from "../manage";
 import { promptUserNumber } from "../../components/promptUserNumber";
+import { promptDeviceAlias } from "../../components/alias";
 import { phraseRecoveryPage } from "./recoverWith/phrase";
 import { deviceRecoveryPage } from "./recoverWith/device";
 import { pickRecoveryDevice } from "./pickRecoveryDevice";
 import { isRecoveryPhrase } from "../../utils/recoveryDevice";
+import { setAnchorUsed } from "../../utils/userNumber";
+import { unknownToString } from "../../utils/utils";
+import { constructIdentity } from "../register/construct";
 
 export const useRecovery = async (
   connection: Connection,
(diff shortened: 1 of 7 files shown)


src/frontend/src/flows/recovery/useRecovery.ts
```diff
@@ -53,5 +57,75 @@ const runRecovery = async (
...
+  userNumber,
+  deviceAlias,
+}: {
+  connection: AuthenticatedConnection;
+  userNumber: bigint;
+  deviceAlias: string;
+}): Promise<void> => {
+  let newDevice;
+  try {
+    newDevice = await constructIdentity({
```
frederikrothenberger, line 89: I would expect to get to try again on error here instead of going to the management screen.

frederikrothenberger: Another thing I noticed: (again Firefox) the WebAuthn prompt appears noticably slower then when going through the registration flow (i.e. after pressing "create new"). Do you know why that is?

nmattia: > I would expect to get to try again on error here instead of going to the management screen.

Ah yes, that makes sense 👍 

> Another thing I noticed: on Firefox the WebAuthn prompt appears slower than when going through the registration flow (i.e. after pressing "create new"). Do you know why that is?

I can't think of any reason, but I remember Safari sometimes being very slow for WebAuthn, although a restart did help. Does this happen consistently?

nmattia: @frederikrothenberger actually will need to figure out what exactly we want here. Are you ok with this being done subsequently?

frederikrothenberger: > Does this happen consistently?

It happens with all browsers I tested FF, Chrome and Safari, consistently. Maybe it's also just my machine 🤷 

Yeah, this can also be improved later. 👍

frederikrothenberger (approved):

src/frontend/src/flows/recovery/useRecovery.ts
nmattia: Ah I see! ok, that's actually normal because we're first fetching the devices to exclude the credentials

PR #1293, opened 2023-03-03
Remove lookup from tests

frederikrothenberger:
`lookup` was recently deprecated. As a consequence, integration tests should not rely on the usage of `lookup`. This PR refactors the tests to not use it anymore, except for one test that ensures that `lookup` keeps working as intended.


src/internet_identity_interface/src/lib.rs
```diff
@@ -196,6 +196,18 @@ pub struct WebauthnCredential {
...
+
+    fn try_from(device: DeviceData) -> Result<Self, Self::Error> {
+        let credential_id = device.credential_id.ok_or(())?;
+        Ok(Self {
+            pubkey: device.pubkey,
+            credential_id,
+        })
+    }
+}
+
```
nmattia, line 210: Is this intended?

src/internet_identity_interface/src/lib.rs
```diff
@@ -196,6 +196,18 @@ pub struct WebauthnCredential {
     pub credential_id: CredentialId,
 }
 
+impl TryFrom<DeviceData> for WebauthnCredential {
```
przydatek, line 199: nit: maybe add a test for the fail-path of this new functionality?

przydatek (approved):
LGTM, modulo a minor suggestion.

PR #1294, opened 2023-03-03
Make manage flows explicit

nmattia:
The `renderManage` page is the starting point for many flows, but all these flows also lead back to it. This was somewhat implicit.

These changes make the "infinite loop" nature of the management page explicit and updates all flows to _not_ decide what to do next; instead this is left to the caller (often `renderManage`). This makes the flows more modular and clarifies the return values. Another benefit is that calls to `renderManage` aren't nested anymore (js doesn't generally have tail recursion optimization).

This made an issue apparent in `pollForTentativeDevice`, where the loader was never exited; now fixed.


frederikrothenberger (approved):
Thanks 👍

PR #1296, opened 2023-03-03
Toast Component

meodai:
Introduces Irregularities and toasts.

- For now there are only errors
- I've added a basic demo to the styleguide. Maybe it would be good to create a "toaster" component that handles adding the `c-toasts` to the dom? I was not sure what was the smartest way to do so.


meodai:
https://user-images.githubusercontent.com/608386/222785919-e85304c7-4e1a-46f6-8313-3e865b50c86d.mov

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,48 @@
...
+  message: string | TemplateResult;
+  additionalClasses?: string[];
+  type?: "warning" | "info" | "error";
+  closeFn?: () => void;
+}
+
+export const irregularity = ({
+  message,
+  additionalClasses = [],
+  type = "error",
```
nmattia, line 14: I think type is a reserved keyword for typescript, this might cause issues

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,48 @@
+import { html, TemplateResult } from "lit-html";
+import { warningIcon, closeIcon } from "./icons";
+
+interface irregularityProps {
+  message: string | TemplateResult;
+  additionalClasses?: string[];
+  type?: "warning" | "info" | "error";
+  closeFn?: () => void;
```
nmattia, line 8: why is this optional? what happens if no `closeFn` is specified?

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const irregularity = ({
```
nmattia, line 11: Can you add a line or two about what this does? Also where does the term `irregularity` come from? Is "irregularity" applicable to `type = "info"` too?

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,48 @@
+import { html, TemplateResult } from "lit-html";
+import { warningIcon, closeIcon } from "./icons";
+
+interface irregularityProps {
+  message: string | TemplateResult;
```
nmattia, line 5: it looks like you only use `TemplateResult`, why allow both?

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,48 @@
...
+  additionalClasses?: string[];
```
nmattia, line 6: Looks like this isn't used, can we remove it until necessary?

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,48 @@
...
+  type?: "warning" | "info" | "error";
```
nmattia, line 7: Looks like this isn't used, can we remove it until necessary?

src/frontend/src/styles/main.css
```diff
@@ -794,6 +797,133 @@ a:hover,
...
 
+/**
+ *  irregularity component
+ */
+.c-irregularity {
+  display: flex;
+  align-items: stretch;
+  gap: calc(var(--rs-card-bezel-x) * 0.25);
+
+  --state-color: var(--rc-warningLight);
```
nmattia, line 808: what is `--state-color`? "state" suggests that it might change but it seems to be set once and for all

src/frontend/src/styles/main.css
```diff
@@ -794,6 +797,133 @@ a:hover,
...
+}
+
+.c-toast--closing {
+  z-index: 1;
+  animation: disappearToBottom 200ms cubic-bezier(0.7, 0.3, 0, 1);
+}
+
+@keyframes appearFromBottom {
+  from {
+    transform: translateY(100%);
```
nmattia, line 907: is this 100% of the element height or of the window height? If it's the former, then this assumes the element is at the bottom of the window, right?

src/frontend/src/components/irregularity.ts
meodai: @nmattia because I can see some cases where there should not be a close button. Sometimes they might be timed notification. Or an error that we want to show until the issue is resolved.

src/frontend/src/components/irregularity.ts
meodai: I will rename it to error type then, (The linter did not complain about it though)

src/frontend/src/components/irregularity.ts
meodai: I did add a comment about it, let be know if this is ok for you or if you need a better explanation.

src/frontend/src/components/irregularity.ts
meodai: I think we should leave this one in, it will definitely give guidance to whoever wants to extend this to use it for notifications or positive feeback.

src/frontend/src/styles/main.css
meodai: translate percents always are in relation to the targeted element, so the height of the element. Yes it assumes the element is at the bottom, thats why I called it `appearFromBottom`.

nmattia (commented):
Looks like there's a lot of configurability added in and I'm not sure it's all warranted, especially since there's no practical use of it yet. Here's a suggestion: would you try to use this in a couple places were errors are currently silenced, and remove everything that's not necessary? I don't think we have a use case for this beyond errors for now.

Examples of currently silenced errors:
* https://github.com/dfinity/internet-identity/blob/166cff030263a0ebe7cca3c093dbd75103dc59b0/src/frontend/src/utils/lit-html.ts#L12
* https://github.com/dfinity/internet-identity/blob/166cff030263a0ebe7cca3c093dbd75103dc59b0/src/frontend/src/flows/register/finish.ts#L42

meodai:
@nmattia sure I would like to discuss how to make the "toast" drawer omnipresent though. Thats why I only added it to the style-guide for now.

meodai:
@nmattia I did what you asked for. But placing them where we used to use `console.error` alone will not cut in IMHO.
If we show an error, the user needs to know what she/he can do about it. 

Possible followup tasks to this:
- Create Meaningful texts for common errors
- Maybe link to FAQ?

src/frontend/src/components/toast.ts
```diff
@@ -0,0 +1,55 @@
...
+  const $toaster = document.querySelector("[data-toaster]");
+  if (!$toaster) {
+    const $toaster = document.createElement("div");
+    $toaster.classList.add("c-toasts");
+    const $container = document.createElement("div");
+    $container.classList.add("l-container");
+    $container.setAttribute("data-toasts", "");
+    $toaster.setAttribute("data-toaster", "");
+    $toaster.appendChild($container);
+    document.body.appendChild($toaster);
```
nmattia, line 39: What happens if other children are appended after the toaster? will the toasts still be displayed correctly?

src/frontend/src/components/toast.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @returns the container reference for the toasts
+ */
+const getOrCreateToaster = (): Element | null => {
+  const $toaster = document.querySelector("[data-toaster]");
+  if (!$toaster) {
+    const $toaster = document.createElement("div");
+    $toaster.classList.add("c-toasts");
+    const $container = document.createElement("div");
+    $container.classList.add("l-container");
+    $container.setAttribute("data-toasts", "");
```
nmattia, line 36:
```suggestion
    // Attribute used to retrieve the element when creating toasts
    $container.setAttribute("data-toasts", "");
```

nmattia (commented):
I love it! Left a few comments

src/frontend/src/components/toast.ts
```diff
@@ -0,0 +1,55 @@
...
+export const removeToast = (toast: HTMLElement): void => {
+  toast.addEventListener("animationend", () => {
+    toast.remove();
+  });
+  toast.classList.add("c-toast--closing");
+};
+
+const createToast = (messageStr: string): HTMLDivElement => {
+  const message = html`${messageStr}`;
+  const $toast = document.createElement("div");
```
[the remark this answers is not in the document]
meodai: I removed it.

src/frontend/src/components/toast.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Get the container for the toasts, or create it if it doesn't exist.
+ * @returns the container reference for the toasts
+ */
+const getOrCreateToaster = (): Element | null => {
+  const $toaster = document.querySelector("[data-toaster]");
+  if (!$toaster) {
+    const $toaster = document.createElement("div");
+    $toaster.classList.add("c-toasts");
+    const $container = document.createElement("div");
```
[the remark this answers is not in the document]
meodai: I added a comment to try to explain why its needed

meodai:
@nmattia thanks for your feedback, I think I addressed everything you pointed out.

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,51 @@
...
+  message: string | TemplateResult;
+  closeFn: () => void;
+}
+
+/**
+ * Irregularity component
+ *
+ * To use when something is out of the ordinary. For example, when a user has
+ * not yet completed a task, or when a user has not yet verified their email
+ * address. Or if there is an error with a form. It can also be used as a tost.
```
nmattia, line 14:
```suggestion
 * address. Or if there is an error with a form. It can also be used as a toast.
```

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,51 @@
...
+export const irregularity = ({
+  message,
+  closeFn,
+}: irregularityProps): TemplateResult => {
+  const cssClasses = ["c-irregularity", "c-irregularity--error"];
+
+  if (typeof closeFn === "function") {
+    cssClasses.push(`c-irregularity--closable`);
+  }
+
```
nmattia, line 33:
```suggestion
```

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param {function} closeFn - When provided, the irregularity will be closable.
+ *
+ * For now we assume that all irregularities are errors. In the future, we may
+ * want to add a type parameter to this component.
+ **/
+export const irregularity = ({
+  message,
+  closeFn,
+}: irregularityProps): TemplateResult => {
+  const cssClasses = ["c-irregularity", "c-irregularity--error"];
```
nmattia, line 28:
```suggestion
  const cssClasses = ["c-irregularity", "c-irregularity--error", "c-irregularity--closable"];
```

src/frontend/src/components/irregularity.ts
```diff
@@ -0,0 +1,51 @@
...
+      <p class="c-irregularity__message">${message}</p>
+      ${typeof closeFn === "function"
+        ? html`<button
+            class="c-irregularity__close"
+            aria-label="Close"
+            @click="${() => closeFn()}"
+          >
+            ${closeIcon}
+          </button>`
+        : null}
```
nmattia, line 46:
```suggestion
      <button
            class="c-irregularity__close"
            aria-label="Close"
            @click="${() => closeFn()}"
          >
            ${closeIcon}
          </button>
```

src/frontend/src/components/toast.ts
```diff
@@ -0,0 +1,60 @@
+import { html, render } from "lit-html";
+import { irregularity } from "./irregularity";
+
+export const removeToast = (toast: HTMLElement): void => {
+  toast.addEventListener("animationend", () => {
```
nmattia, line 5:
```suggestion
  // The animation is triggered by the class added below
  toast.addEventListener("animationend", () => {
```

src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -102,6 +103,7 @@ const init = (): Promise<void> =>
           seedCopy.classList.add("is-copied");
         })
         .catch((e) => {
+          toast.error("Unable to copy seed phrase, check console for details");
```
nmattia, line 106:
```suggestion
          toast.error("Unable to copy seed phrase");
```

I think anyone with some dev knowledge will know to check the console, but for everyone else this will just be confusing

src/frontend/src/flows/register/finish.ts
```diff
@@ -39,6 +40,9 @@ export const displayUserNumberTemplate = ({
               elem.classList.add("is-copied");
             });
           } catch (e: unknown) {
+            toast.error(
+              "Unable to copy Identity Anchor, check the console for details"
```
nmattia, line 44:
```suggestion
              "Unable to copy Identity Anchor"
```

I think anyone with some dev knowledge will know to check the console, but for everyone else this will just be confusing

nmattia (commented):
Looking great, thanks! Left a couple suggestions

PR #1298, opened 2023-03-06
Rename recovery_phrase to recovery_phrases

frederikrothenberger:
This fixes an error introduced in a late change on #1280 that introduced
the `get_credentials` endpoint.


nmattia (approved):
Thanks!

PR #1299, opened 2023-03-06
Normalize ambiguous characters in captcha

frederikrothenberger:
This PR makes the captcha easier to solve for humans by normalizing confusing characters (such as `l` and `1`).

E.g. this means that no captcha containing l will be generated and all inputs into captcha solutions will map `l` to `1`.

### DIFF
--- a/src/internet_identity/src/anchor_management/registration.rs
+++ b/src/internet_identity/src/anchor_management/registration.rs
@@ -13,7 +13,7 @@ use rand_core::{RngCore, SeedableRng};
 use std::collections::HashMap;
 
 #[cfg(not(feature = "dummy_captcha"))]
-use captcha::filters::{Grid, Wave};
+use captcha::filters::Wave;
 use lazy_static::lazy_static;
 
 // 5 mins
--- a/src/internet_identity/src/anchor_management/registration.rs
+++ b/src/internet_identity/src/anchor_management/registration.rs
@@ -166,7 +166,7 @@ fn check_challenge(res: ChallengeAttempt) -> Result<(), ()> {
     }
     // Normalize challenge attempts by replacing characters that are not in the captcha character set
     // with the respective replacement from CHAR_REPLACEMENTS.
-    let normalized_challenge: String = res
+    let normalized_challenge_res: String = res
         .chars
         .chars()
         .map(|c| *CHAR_REPLACEMENTS.get(&c).unwrap_or(&c))
@@ -175,7 +175,7 @@ fn check_challenge(res: ChallengeAttempt) -> Result<(), ()> {
     state::inflight_challenges_mut(|inflight_challenges| {
         match inflight_challenges.remove(&res.key) {
             Some(challenge) => {
-                if normalized_challenge != challenge.chars {
+                if normalized_challenge_res != challenge.chars {
                     return Err(());
                 }
                 Ok(())
--- a/src/internet_identity/src/anchor_management/registration.rs
+++ b/src/internet_identity/src/anchor_management/registration.rs
@@ -10,9 +10,11 @@ use ic_cdk::{call, caller, trap};
 use internet_identity_interface::archive::{DeviceDataWithoutAlias, Operation};
 use internet_identity_interface::*;
 use rand_core::{RngCore, SeedableRng};
+use std::collections::HashMap;
 
 #[cfg(not(feature = "dummy_captcha"))]
-use captcha::filters::Wave;
+use captcha::filters::{Grid, Wave};
+use lazy_static::lazy_static;
 
 // 5 mins
 const CAPTCHA_CHALLENGE_LIFETIME: u64 = secs_to_nanos(300);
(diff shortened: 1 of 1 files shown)


src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -10,9 +10,11 @@ use ic_cdk::{call, caller, trap};
 use internet_identity_interface::archive::{DeviceDataWithoutAlias, Operation};
 use internet_identity_interface::*;
 use rand_core::{RngCore, SeedableRng};
+use std::collections::HashMap;
 
 #[cfg(not(feature = "dummy_captcha"))]
-use captcha::filters::Wave;
+use captcha::filters::{Grid, Wave};
```
nmattia, line 16: is `Grid` needed?

src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -139,10 +160,22 @@ fn create_captcha<T: RngCore>(rng: T) -> (Base64, String) {
 
 // Check whether the CAPTCHA challenge was solved
 fn check_challenge(res: ChallengeAttempt) -> Result<(), ()> {
+    // avoid processing too many characters
+    if res.chars.len() > CAPTCHA_LENGTH {
+        return Err(());
+    }
```
nmattia, line 166: This is unrelated, right?

src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -118,11 +120,30 @@ fn create_captcha<T: RngCore>(rng: T) -> (Base64, String) {
...
 #[cfg(not(feature = "dummy_captcha"))]
 fn create_captcha<T: RngCore>(rng: T) -> (Base64, String) {
     let mut captcha = captcha::RngCaptcha::from_rng(rng);
+    let mut chars = captcha.supported_chars();
+    chars.retain(|c| !CHAR_REPLACEMENTS.contains_key(c));
+
     let captcha = captcha
-        .add_chars(5)
+        .set_chars(&chars)
+        .add_chars(CAPTCHA_LENGTH as u32)
```
nmattia, line 146: Would it be simpler to normalize the challenge result instead?

src/internet_identity/src/anchor_management/registration.rs
frederikrothenberger: I added it since we now do processing on the input where previously we did only an equality check.

src/internet_identity/src/anchor_management/registration.rs
frederikrothenberger: The challenge result is normalized. But I can also have the captcha generated _with_ ambiguous characters and normalize both, the challenge and the input by the user when checking the result. Is this what you would prefer?

src/internet_identity/src/anchor_management/registration.rs
frederikrothenberger: Ah, no. Removed.

src/internet_identity/src/anchor_management/registration.rs
nmattia: Ah my bad, I was confused by `CAPTCHA_LENGTH`. I thought you were adding as many chars to the captcha as there were in `CHAR_REPLACEMENTS` and then only got more confused from there.

If the challenge result is normalized, why also replace `chars`? Is that just for consistency?

src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -139,10 +160,22 @@ fn create_captcha<T: RngCore>(rng: T) -> (Base64, String) {
 
 // Check whether the CAPTCHA challenge was solved
 fn check_challenge(res: ChallengeAttempt) -> Result<(), ()> {
+    // avoid processing too many characters
+    if res.chars.len() > CAPTCHA_LENGTH {
+        return Err(());
+    }
+    // Normalize challenge attempts by replacing characters that are not in the captcha character set
+    // with the respective replacement from CHAR_REPLACEMENTS.
+    let normalized_challenge: String = res
```
nmattia, line 169:
```suggestion
    let normalized_challenge_res: String = res
```

got confused and thought you were re-normalizing the challenge

src/internet_identity/src/anchor_management/registration.rs
frederikrothenberger: I think we have 2 options:
1. Normalize (reduce) input chars, such that the captcha does not contain ambiguous chars (i.e. `1lXcx` will never be generated) and also normalize the input by the user.
2. Generate Captcha as before (i.e. `1lXcx`) and only when checking the solution normalize _both_ the actual challenge and the user input. So in this example the user input would be normalize and checked against `11xcx`.

This PR so far implements approach 1.

nmattia (approved):
LGTM modulo "challenge" vs "challenge result" (see suggestion)

src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -118,11 +120,30 @@ fn create_captcha<T: RngCore>(rng: T) -> (Base64, String) {
     return (resp, captcha.chars_as_string());
 }
 
+lazy_static! {
+    /// Map of problematic characters that are easily mixed up by humans to the "normalized" replacement.
+    /// I.e. the captcha will never contain any of the characters in the key set and any input provided
+    /// will be mapped to the matching value.
+    /// Note: the captcha library already excludes the characters o, O and 0.
```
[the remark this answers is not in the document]
frederikrothenberger: No, I suspect it is because for those characters it is less controversial that they are ambiguous. Contributing more to the library could also help us make use of more filters.

Happy to discuss offline.

PR #1300, opened 2023-03-06
Refactor pollForTentativeDevice

nmattia:
This updates the `pollForTentativeDevice` screen to be more consistent with other pages, and includes the following fixes:
* The `poll` function loop is made explicit (also meaning the function won't be nested indefinitely anymore)
* The user flow logic (which screens to show, error handling, etc) is moved out of `pollForTentativeDevice` and is instead handled in a new function, `addRemoteDevice` mirroring `addLocalDevice` (meaning the only concern of `pollForTentativeDevice` is to poll for tentative device)
* All promise results are changed to async/await instead of `.then()`
* The `AsyncCountdown` is used, meaning less state is shared


src/frontend/src/flows/addDevice/manage/addRemoteDevice.ts
```diff
@@ -0,0 +1,55 @@
...
+// The flow for adding a remote (i.e. other browser, non-hardware) device
+export const addRemoteDevice = async ({
+  userNumber,
+  connection,
+}: {
+  userNumber: bigint;
+  connection: AuthenticatedConnection;
+}): Promise<void> => {
+  // Enter registration mode and get info about the anchor
+  const [timestamp, anchorInfo] = await withLoader(() =>
```
peterpeterparker, line 16: 1. the function use to be `async`, not needed anymore?

```
await withLoader(withLoader(async () => 
```

2. nth: even if deferred I would not be against specifying the return type for readability

```
const [timestamp, anchorInfo] = await withLoader((): Promise<[something, something]> => ...
```

src/frontend/src/flows/addDevice/manage/addRemoteDevice.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    if (result === "timeout") {
+      // On timeout, notify the user
+      await displayError({
+        title: "Timeout Reached",
+        message:
+          'The timeout has been reached. For security reasons the "add device" process has been aborted.',
+        primaryButton: "Ok",
+      });
+      return;
```
peterpeterparker, line 40: can result be here something else that timeout or that's an edge case? can you maybe add a comment next to this return to explains what happens if we land here?

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -55,110 +67,64 @@ const pageContent = (userNumber: bigint) => {
...
-    await init(userNumber, connection, timestamp);
-  }
-};
+  // Poll repeatedly
+  const polling = poll(userNumber, connection, () => countdown.hasStopped());
 
-export const renderPollForTentativeDevicePage = (userNumber: bigint): void => {
-  const container = document.getElementById("pageContent") as HTMLElement;
-  render(pageContent(userNumber), container);
+  return await Promise.race([page, polling]);
```
peterpeterparker, line 106: the `await` is useful here?

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -55,110 +67,64 @@ const pageContent = (userNumber: bigint) => {
...
     if (shouldStop()) {
-      return null;
+      return "timeout";
     }
-    const tentativeDevice = getTentativeDevice(response);
-    if (tentativeDevice) {
+
+    const anchorInfo = await connection.getAnchorInfo();
+    const tentativeDevice =
+      anchorInfo.device_registration[0]?.tentative_device[0];
```
peterpeterparker, line 121: this - `anchorInfo.device_registration[0]?.tentative_device[0]` - appears twice in this PR, maybe worth a utils?

peterpeterparker (approved):
LGTM, minor comments. Thanks

src/frontend/src/flows/addDevice/manage/addRemoteDevice.ts
nmattia: > the function use to be async, not needed anymore?

Not needed; wasn't needed before either as far as I could tell

> nth: even if deferred I would not be against specifying the return type for readability

Sure, I'll actually add it to the LHS instead if ok for you (i.e.` const [ ... ] : [Timestamp, ...] =`)

src/frontend/src/flows/addDevice/manage/addRemoteDevice.ts
nmattia: The result can be a timeout, a cancellation or device information. For the two first cases (timeout, cancel) there's not much we can do. I'm adding a comment!

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
nmattia: It's not! good catch

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
nmattia: There was a helper around this but I found it more confusing than just inlining the definition; if you feel strongly I'm happy to re-add one!

src/frontend/src/flows/addDevice/manage/addRemoteDevice.ts
nmattia: Done!

frederikrothenberger (approved):
This is a nice refactoring 👍

PR #1301, opened 2023-03-06
Offer to retry enrolling on recovery

nmattia:
This fixes an issue in the recovery flow.

Before this, an error when enrolling the device would exit the flow and take the user to the anchor management page. Now we offer to retry enrolling a new device.


frederikrothenberger (approved):
Thanks 👍 
It would be nice if we had tests for error handling like this as well. WDYT?

nmattia:
> It would be nice if we had tests for error handling like this as well. 

Yeah 100% agree. This needs to go out somewhat soon, so let me merge this and track the tests somewhere. Thanks!

PR #1302, opened 2023-03-06
Add configurable rate limit to register

frederikrothenberger:
This PR adds a configurable rate limit to register. This ensures that the stable memory reserve cannot be used up in an arbitrary short amount of time.

The exported metrics allow the configuration of alerts on the rate limit being hit.


src/internet_identity/tests/tests.rs
```diff
@@ -64,11 +64,7 @@ mod upgrade_tests {
             &env,
             canister_id,
             II_WASM.clone(),
-            Some(InternetIdentityInit {
```
frederikrothenberger, line 67: I extracted the helper function, because I had to touch all these locations anyways because of the new field in the init argument.

src/internet_identity/tests/tests.rs
```diff
@@ -497,6 +466,108 @@ mod registration_tests {
...
+
+        env.advance_time(Duration::from_secs(100));
+
+        // some activity required to process the rate limit (which will use one token)
+        flows::register_anchor(&env, canister_id);
+
+        assert_metric(
+            &get_metrics(&env, canister_id),
+            "internet_identity_register_rate_limit_current_tokens",
+            1f64, // 1 and not 2 because of register call required to update rate limit
```
przydatek, line 546: nit: maybe use a more realistic value for `max_tokens`?

frederikrothenberger: Nvm, as discussed added higher limit.

PR #1304, opened 2023-03-06
Fix hash equality

nmattia:
This uses the recommended `===` instead of `==` for (string) hash comparison. Typo the first time around.


peterpeterparker (approved):
LGTM

PR #1305, opened 2023-03-07
Update device settings copy

nmattia:
This updates or adds prompts explaining to the user what they're about to do when confirming.

### DIFF
--- a/src/frontend/src/flows/manage/deviceSettings.ts
+++ b/src/frontend/src/flows/manage/deviceSettings.ts
@@ -136,15 +136,24 @@ const deleteDevice = async ({
     pubKey
   );
 
-  const shouldProceed = sameDevice
-    ? confirm(
-        "This will remove your current device and you will be logged out."
-      )
-    : confirm(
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/manage/deviceSettings.ts
```diff
@@ -136,15 +136,24 @@ const deleteDevice = async ({
...
+    confirmationPrompt.push("Remove your Recovery Device");
+    confirmationPrompt.push(
+      "Are you sure you want to remove your recovery device? You will no longer be able to use it to recover your account."
+    );
+  } else {
+    confirmationPrompt.push(
+      `Do you really want to remove the device "${device.alias}"?`
+    );
+  }
+  if (sameDevice) {
```
peterpeterparker, line 151: it's correct that the message can be one of the above + this message - i.e. not `else if (sameDevice)`?

nmattia: Yep!

PR #1306, opened 2023-03-07
Basic Modal

meodai:
Introduces native modals using the browsers "dialog" feature that is widely supported now.
the `modal` function takes a `block`.


src/frontend/src/components/modal.ts
```diff
@@ -0,0 +1,59 @@
+import { html, TemplateResult, render } from "lit-html";
+
+export const createModal = ({
+  slot,
+  id,
+  confirmText = "Ok",
+}: {
+  slot: TemplateResult;
+  id?: string;
+  confirmText?: string;
```
nmattia, line 10: I think we can leave those two out until we have a good use case for them

src/frontend/src/components/modal.ts
```diff
@@ -0,0 +1,59 @@
...
+          </button>
+        </div>
+      </div>
+    </form>
+  `;
+
+  const $modal = document.createElement("dialog");
+  $modal.id = id != null ? id : "modal";
+  $modal.classList.add("c-modal");
+  $modal.setAttribute("aria-modal", "true");
```
nmattia, line 29: Why mix `render` and `createElement`? Why not return a `TemplateResult` instead?

src/frontend/src/components/modal.ts
```diff
@@ -0,0 +1,59 @@
...
+  confirmText = "Ok",
+}: {
+  slot: TemplateResult;
+  id?: string;
+  confirmText?: string;
+}): HTMLDialogElement => {
+  const modalHtml = html`
+    <form method="dialog" class="c-modal__content c-card c-card--modal">
+      <div class="c-modal__inner">
+        <span class="c-modal__close" data-closemodal>&times;</span>
```
nmattia, line 15: How about adding the `close` functions here via `@click=${ () => ... }` instead of looking up the elements later on with `querySelector`? If you need a placeholder for the modal before it is created you can use `createRef()` and `Ref`

nmattia (commented):
I love it!

src/frontend/src/components/modal.ts
```diff
@@ -0,0 +1,59 @@
...
+          <button type="submit" class="c-button c-button--primary">
+            ${confirmText}
+          </button>
+        </div>
+      </div>
+    </form>
+  `;
+
+  const $modal = document.createElement("dialog");
+  $modal.id = id != null ? id : "modal";
```
peterpeterparker, line 27: is there a risk to have two or more ids next to each other on the dom? if yes, maybe worth to use a custom attribute instead?

src/frontend/src/components/modal.ts
```diff
@@ -0,0 +1,59 @@
...
+  render(modalHtml, $modal);
+  return $modal;
+};
+
+export const modal = (slot: TemplateResult): Promise<void> => {
+  // create modal and append it to the body
+  const $modal = createModal({ slot });
+  document.body.appendChild($modal);
+
+  return new Promise((resolve) => {
```
peterpeterparker, line 40: just an idea: instead of keeping a promise open, why not return a callback instead?

src/frontend/src/components/modal.ts
```diff
@@ -0,0 +1,59 @@
...
+    // open modal
+    $modal.showModal();
+    const $modalClose = $modal.querySelector(
```
peterpeterparker, line 43: would it be useful to use a mutationobserver to be sure those buttons are added to the dom? you are casting to HTMLElement but they can be null and throw an exception when the onclick is attached

src/frontend/src/components/modal.ts
meodai: the ID is there to be used for accessibility (target)

src/frontend/src/components/modal.ts
meodai: not sure how a mutation observer would solve that? can you make an example?

src/frontend/src/components/modal.ts
meodai: do all the buttons say OK ?

src/frontend/src/components/modal.ts
meodai: ah good to know

src/frontend/src/components/modal.ts
peterpeterparker: should we prevent the potential error then?

src/frontend/src/components/modal.ts
peterpeterparker: Just noticing this comment of @nmattia https://github.com/dfinity/internet-identity/pull/1306/files#r1128004572 which per extension makes my comment obsolte since queries will disappear

src/frontend/src/components/modal.ts
nmattia: that's because of me, I've asked @meodai to do that so that we can await the modal (i.e. wait for the user to ack)

src/frontend/src/components/modal.ts
meodai: thanks for caring though!

src/frontend/src/components/modal.ts
meodai: we could check for existing ID's but this would happen on runtime.

src/frontend/src/components/modal.ts
meodai: Because it needs to return a native DOM element that I can append to the body. How wo that work with TempleteResult?

src/frontend/src/components/modal.ts
meodai: addressed this in 83e0a6f9ad571c77710b51ae7e9889a9cab866eb I passed the reference in from the promise since it needs to resolve it. Can you think of a nicer way to do that?

PR #1309, opened 2023-03-08
Fix typo in candid file

frederikrothenberger:
Detected while working on auto-generating candid files.


przydatek (approved):
LGTM, but I wonder if it affects NNS dapp (cc @lmuntaner)

frederikrothenberger:
> LGTM, but I wonder if it affects NNS dapp (cc @lmuntaner)

It does not, the NNS dapp is not a client for the II canister.

lmuntaner:
Only if `InternetIdentityInit` changes. We use it to show the payload of the proposal to upgrade II.

PR #1310, opened 2023-03-08
Add tests for candid file compatibility

frederikrothenberger:
Unfortunately, the candid schema generation requires an additional annotation in order to work.

The tests added in this PR also uncovered out of sync type definitions in the archive canister. This is fixed as well.

### DIFF
--- a/src/archive/src/main.rs
+++ b/src/archive/src/main.rs
@@ -685,7 +685,6 @@ async fn status() -> ArchiveStatus {
 /// tools, can seamlessly integrate with it. The concrete interface (method name etc.) is
 /// provisional, but works.
 #[query]
-#[candid_method(query)]
 fn __get_candid_interface_tmp_hack() -> String {
     include_str!("../archive.did").to_string()
 }
@@ -709,5 +708,11 @@ mod test {
             CandidSource::File(Path::new("archive.did")),
         )
         .unwrap_or_else(|e| panic!("the canister code is incompatible to the did file: {:?}", e));
+
+        service_compatible(
+            CandidSource::File(Path::new("archive.did")),
+            CandidSource::Text(&canister_interface),
+        )
+        .unwrap_or_else(|e| panic!("the did file is incompatible to the canister code: {:?}", e));
     }
 }
--- a/src/archive/src/main.rs
+++ b/src/archive/src/main.rs
@@ -36,7 +36,7 @@
 //! - prefix scan with anchor to retrieve entries by anchor
 //! - prefix scan with (anchor, timestamp) to narrow down on the time period for a specific anchor
 //! - prefix scan with (anchor, timestamp, log index) to do pagination (with the key of the first entry not included in the previous set)
-use candid::{CandidType, Deserialize, Principal};
+use candid::{candid_method, CandidType, Deserialize, Principal};
 use ic_cdk::api::call::CallResult;
 use ic_cdk::api::management_canister::main::{canister_status, CanisterIdRecord};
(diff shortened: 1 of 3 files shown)


src/archive/src/main.rs
```diff
@@ -680,8 +685,29 @@ async fn status() -> ArchiveStatus {
...
 /// provisional, but works.
 #[query]
+#[candid_method(query)]
 fn __get_candid_interface_tmp_hack() -> String {
     include_str!("../archive.did").to_string()
 }
 
 fn main() {}
+
+// Order dependent: do not move above any function annotated with #[candid_method]!
```
nmattia, line 695: If the order is moved the `archive.did` changes and needs to be updated, or does it break more things?

src/archive/src/main.rs
```diff
@@ -680,8 +685,29 @@ async fn status() -> ArchiveStatus {
...
+#[cfg(test)]
+mod test {
+    use crate::__export_service;
+    use candid::utils::{service_compatible, CandidSource};
+    use std::path::Path;
+
+    #[test]
+    fn check_candid_interface_compatibility() {
+        let canister_interface = __export_service();
+        service_compatible(
```
nmattia, line 707: does "compatible" mean "equal"?

nmattia (commented):
I'm a bit confused, I assumed this would either:
* Ensure that the `.did` files were up to date, i.e. matching what's in this commit, or
* Ensure that the `.did` files were compatible with the latest released `.did` files (i.e. what `didc check` would do)

But this seems to do something in between?

src/archive/src/main.rs
frederikrothenberger: If the order is changed, the test will fail because the canister interface generated from the `.did` will be incomplete.

src/archive/src/main.rs
frederikrothenberger: No. It is actually a subtype check i.e. the canister interface is a subtype of the `.did` service.

frederikrothenberger:
> Ensure that the .did files were up to date, i.e. matching what's in this commit, or

This is what the test does. I.e. make sure that the canister actually implements what's in the `.did`. What do you mean it does something in between?

I added this test because it is a nice check until we have either gotten rid of the `.did` and have it auto-generated or we generate rust types from the `.did`. The latter is actually preferred, because we then have the `.did` as the source of truth for both the TS/JS and the rust types. However, rust code generation is experimental. @przydatek and I have a sync with the candid team on that later today.

src/archive/src/main.rs
nmattia: Ah, why incomplete? Won't it simply be reordered?

src/archive/src/main.rs
nmattia: Thanks. Might be worth leaving a note, from the code it's unclear if this is checking `:>` or `<:`.

src/archive/src/main.rs
frederikrothenberger: No, there is some weird state handling going on in the rust macros. There is some global state that both the `candid::candid_method` annotations and the `candid::export_service!` access. And since macros are executed in order of appearance, anything after `candid::export_service!` will not be included in the generated `.did`. This then means that it is not a proper subtype (since it's missing methods) and causes the test to fail.

frederikrothenberger:
> This does however mean that -- while compatible -- the canister interface and the candid interface might mismatch.

That is true. Let me change the check to do it in both direction, which should then force equality.

nmattia (approved):
LGTM, thanks!

PR #1313, opened 2023-03-09
Add helper for rendering page

nmattia:
This adds a helper function that renders a page to `#pageContent`, which is the element to which we attach content.

The logic for attaching to `#pageContent` can be removed from pretty much all pages.


peterpeterparker (approved):
Cool, thanks! 👍

PR #1315, opened 2023-03-09
Clean up displaySeedPhrase

nmattia:
This updates the recovery phrase display page to use more idiomatic lit and constructs consistent with the rest of the codebase. It also fixes a minor bug where the logic on checkbox click tried to remove a non-existing class on the button (`is-hidden`).


src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -1,9 +1,52 @@
...
+  const phraseElement: Ref<HTMLElement> = createRef();
+  const phraseCopyElement: Ref<HTMLElement> = createRef();
+
+  const continueButton: Ref<HTMLButtonElement> = createRef();
+  const checkbox: Ref<HTMLInputElement> = createRef();
+
+  // Selects the phrase content to give visual feedback of what's been copied
+  const selectPhrase = () => {
+    withRef(phraseElement, (phraseElement) => {
+      const selection = window.getSelection();
```
peterpeterparker, line 26: just in case, a get selection that supports old browser I used in ddg and stylo:

```
export const getSelection = (): Selection | null => {
  if (window && window.getSelection) {
    return window.getSelection();
  } else if (document && document.getSelection) {
    return document.getSelection();
  } else if (document && (document as any).selection) {
    return (document as any).selection.createRange().text;
  }

  return null;
};
```

src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -1,9 +1,52 @@
...
+      if (selection !== null) {
+        const range = document.createRange();
+        range.selectNodeContents(phraseElement);
+        selection.removeAllRanges();
+        selection.addRange(range);
```
peterpeterparker, line 31: I generally detach the range afterwards to free memory

```
...
selection.addRange(range);

range.detach();
```

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/recovery/displaySeedPhrase.ts
nmattia: Nice, thanks! I've just reused whatever we were using but I'm making note of this; will add soon but want to test it a bit first.

src/frontend/src/flows/recovery/displaySeedPhrase.ts
nmattia: Thanks! I've only moved the code for now; will do this subsequently with the other fix you suggested above.

PR #1317, opened 2023-03-10
Don't prefix used variables with underscore

nmattia:
This replaces a few (used) variables like `_foo` with `foo`.

Some local variables were prefixed with an underscore to differentiate them from a similarly named argument, e.g.:

```js
function foo(bar) {
  const _bar = bar +1;
}
```

Now instead the argument has an underscore appended:

```js
function foo(bar_) {
  const bar = bar_ +1;
}
```

this ensures that by default, `bar` is used.

### DIFF
--- a/src/frontend/src/components/authenticateBox.ts
+++ b/src/frontend/src/components/authenticateBox.ts
@@ -255,25 +255,23 @@ const page = (slot: TemplateResult) => {
 
 // Start the "remote device" add flow by prompting for an anchor, and then
 // adding a device to that anchor.
-const addRemoteDevice = async (
-  connection: Connection,
-  userNumber_?: bigint
-) => {
+const addRemoteDevice = async (connection: Connection, userNumber?: bigint) => {
   // Prompt the user for an anchor (only used if we don't know the anchor already)
   const askUserNumber = async () => {
-    const userNumber = await promptUserNumber({
+    const result = await promptUserNumber({
       title: "Add a Trusted Device",
       message: "What’s your Identity Anchor?",
     });
-    if (userNumber === "canceled") {
+    if (result === "canceled") {
       // TODO L2-309: do this without reload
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/authenticateBox.ts
```diff
@@ -255,22 +255,25 @@ const page = (slot: TemplateResult) => {
...
       // TODO L2-309: do this without reload
       return window.location.reload() as never;
     }
 
-    return _userNumber;
+    return userNumber;
   };
 
-  const _userNumber = userNumber ?? (await askUserNumber());
+  const userNumber = userNumber_ ?? (await askUserNumber());
```
peterpeterparker, line 276: technically if you inline the check I think you would maybe not even need `userNumber_` here right?

```
const addRemoteDevice = async (
  connection: Connection,
  userNumber?: bigint
) => {
  // Prompt the user for an anchor (only used if we don't know the anchor already)
  const askUserNumber = async () => {
    const userNumber = await promptUserNumber({
      title: "Add a Trusted Device",
      message: "What’s your Identity Anchor?",
    });
    if (userNumber === "canceled") {
      // TODO L2-309: do this without reload
      return window.location.reload() as never;
    }

    return userNumber;
  };

  return registerTentativeDevice(userNumber ?? (await askUserNumber()), connection);
};
```

(not a call to action or comment, meant as question only)

peterpeterparker (approved):
LGTM, thx

src/frontend/src/components/authenticateBox.ts
nmattia: yes that's correct! let me give this a think, might be less error prone

nmattia: @peterpeterparker I've changed it to do like you suggest, thanks! 👍

PR #1319, opened 2023-03-10
Show recovery words table

nmattia:
This updates the page showing the recovery phrase to be more user friendly.

The new component (designed by @meodai, adapted here from the styleguide) displays each word separately, similarly to what most wallets do.

https://user-images.githubusercontent.com/6930756/224312745-ac06a738-6d32-44c8-8a9f-d62854773d08.mov

The tests are updated to use the preferred storage method, which is the "copy" button; previously tests copied the value from the `output` element, which should but may not have matched the clipboard content.

Some workarounds are needed to make the clipboard work in the browser tests; see code for details.

NOTE: `prettier` is going a little crazy on this diff, so I've not run the formatter yet on purpose, for readability. Will run it once PR is approved.


src/frontend/src/test-e2e/views.ts
```diff
@@ -126,7 +126,37 @@ export class RecoveryMethodSelectorView extends View {
   }
 
   async getSeedPhrase(): Promise<string> {
-    return await this.browser.$("#seedPhrase").getText();
+
+    // This tries to read the recovery phrase by first copying it to the clipboard.
+
+    await this.copySeedPhrase();
+
+    // Our CSP policy prevents us from directly reading the clipboard.
```
peterpeterparker, line 134: For what is worth I don't see this as a mock but sponteneously finds it better than reading directly the clipboard - i.e. in addition to the copy you test also that the OS will paste it correctly

src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -14,30 +14,18 @@ const displaySeedPhraseTemplate = ({
...
-      if (selection !== null) {
-        const range = document.createRange();
-        range.selectNodeContents(phraseElement);
-        selection.removeAllRanges();
-        selection.addRange(range);
-      }
-    });
-  };
+  // Assume the phrase is a list of space-separated words
+  const recoveryWords = seedPhrase.split(" ");
```
peterpeterparker, line 23: does `" "` finds somewhere else in the code base for the seed phrase and if yes, maybe extract a constant?

should you add a test that the count of words is exactly the expected count (25)?

nmattia: > does " " finds somewhere else in the code base for the seed phrase and if yes, maybe extract a constant?

Not really in any meaningful way, but I'll keep that in mind if that's the case in the future

> should you add a test that the count of words is exactly the expected count (25)?

exact count is not important, also we'd then be testing a third party lib (we don't generate the words ourselves) so I would rather not

src/frontend/src/test-e2e/views.ts
nmattia: yep, 100% agreed that it's better to emulate what the user is doing. However there's no actual user, that's why I'm saying we "mock" the user input :) Happy to reword if necessary!

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/test-e2e/views.ts
peterpeterparker: all good

src/frontend/src/flows/recovery/displaySeedPhrase.ts
peterpeterparker: ok

meodai:
nice!

PR #1320, opened 2023-03-10
Add interface compatibility check to latest release

frederikrothenberger:
Adds a CI check to make sure the main branch did file is compatible to the latest release.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -651,3 +651,13 @@ jobs:
         with:
           # we check that ubuntu builds match the docker build
           sha256: ${{ startsWith(matrix.os, 'ubuntu') && steps.sha256.outputs.sha256 || '' }}
+
+  interface-compatibility:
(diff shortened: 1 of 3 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -651,3 +651,13 @@ jobs:
...
+
+  interface-compatibility:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-didc
+      - name: "Check canister interface compatibility"
+        run: |
+          curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity.did -o internet_identity_previous.did
+          didc check src/internet_identity/internet_identity.did internet_identity_previous.did
```
nmattia, line 663: thanks! can you confirm that `didc check foo bar` ensures that `foo` is backwards compatible with `bar` and not the other way around?

nmattia (approved):

PR #1323, opened 2023-03-14
Add recovery phrase confirmation

nmattia:
This adds a screen asking the user to confirm (re-input) part of their recovery phrase.

This only covers the logic & copy of the screen; the design is implemented separately.

The confirmation is done by presenting the user with their phrase with some words removed. The words need to be re-typed. The user may only continue once all words have be re-input correctly. This is done by storing the expected word as `data-expected` and regularly checking all words against their `data-expected` fields. Per-word feedback is also given to show which words have been input correctly and which still need attention.

https://user-images.githubusercontent.com/6930756/224961800-d41ecebc-8f1a-4603-8530-17df90c1b935.mov


src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -0,0 +1,157 @@
...
+  i18n: I18n;
+}) => {
+  const copy = i18n.i18n(copyJson);
+
+  // All words, where a `Ref` was added if the word needs checking
+  // NOTE: typescript can't follow if word is deconstructed with {...word}
+  const words: Word[] = words_.map((word) => {
+    if (word.check) {
+      const elem: Ref<HTMLInputElement> = createRef();
+      return { word: word.word, check: word.check, elem };
```
peterpeterparker, line 42: nth: destructing here would be more readable because we would avoid `word.word`

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -0,0 +1,157 @@
...
+  );
+
+  return html`<li style="background-color: ${asyncReplace(backgroundColor)};">
+    <input
+      ${ref(word.elem)}
+      data-expected=${word.word}
+      data-state=${asyncReplace(
+        state.map(
+          (x) => x
+        ) /* workaroudn because chan supports only one .recv() */
```
peterpeterparker, line 122: typo "workaroudn"

src/frontend/src/flows/recovery/setupRecovery.ts
```diff
@@ -96,6 +97,11 @@ export const setupPhrase = async (
     seedPhrase,
     IC_DERIVATION_PATH
   );
+
+  const phrase = userNumber.toString(10) + " " + seedPhrase;
```
peterpeterparker, line 101: what's 10?

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -0,0 +1,157 @@
...
+  | { check: false }
+  | { check: true; elem: Ref<HTMLInputElement> }
+);
+
+// A list of indices nicely spread over the 25 words (anchor + 24 BIP39)
+export const checkIndices = [3, 7, 12, 13, 15, 17, 18, 20, 22, 24];
+
+const checkWord = (word: Word): boolean =>
+  word.check
+    ? withRef(word.elem, (elem) => elem.value === elem.dataset.expected) ===
```
peterpeterparker, line 22: I'm not sure where is `dataset.expected` set. I guess not part of this PR and already done in the code base. We can safely assume it's correctly set right?

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -0,0 +1,157 @@
...
+// A list of words, where "check" indicates if the user needs to double check (re-input) a word
+type Word = { word: string } & (
+  | { check: false }
+  | { check: true; elem: Ref<HTMLInputElement> }
+);
+
+// A list of indices nicely spread over the 25 words (anchor + 24 BIP39)
+export const checkIndices = [3, 7, 12, 13, 15, 17, 18, 20, 22, 24];
+
+const checkWord = (word: Word): boolean =>
```
peterpeterparker, line 20: just fyi, I actually spent quite some time reviewing this particular function. I did not picked directly it was use to test if the input entry was valid. So if you want to add a small comment, review its name or make the return more obvious, not against.

peterpeterparker (approved):
LGTM, minor questions. Thanks

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
nmattia: >   // NOTE: typescript can't follow if word is deconstructed with {...word}

guess the comment is a little too far then :)

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
nmattia: Ah, this could a be a little more clear then. It's `data-expected` that's set on the element: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
nmattia: Ah my bad, that definitely deserves clarification

PR #1324, opened 2023-03-14
Extract copy for recovery creation

nmattia:
This helps clarify code vs content and will help support i18n.


peterpeterparker (approved):
Rubber stamp

PR #1325, opened 2023-03-14
Add back button to recovery confirmation

nmattia:
This allows the user to go back to the page that displays the recovery phrase, from the phrase confirmation/input page.

### DIFF
--- a/src/frontend/src/flows/recovery/setupRecovery.ts
+++ b/src/frontend/src/flows/recovery/setupRecovery.ts
@@ -8,7 +8,7 @@ import {
   IC_DERIVATION_PATH,
   AuthenticatedConnection,
 } from "../../utils/iiConnection";
-import { unknownToString } from "../../utils/utils";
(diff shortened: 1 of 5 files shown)


src/frontend/src/flows/recovery/setupRecovery.ts
```diff
@@ -99,8 +99,24 @@ export const setupPhrase = async (
...
+
+    const result = await confirmSeedPhrase({ phrase });
+    // User has confirmed, so break out of the loop
+    if (result === "confirmed") {
+      break;
+    }
+
+    // User has clicked the back button, so we retry
+    if (result === "back") {
+      continue;
```
nmattia, line 115: Again, I'm not 100% fan of the infinite loop + `continue` combo, but I haven't found a better (non-recursive) way. Will try to fix this up when adding wizard components.

peterpeterparker (approved):
LGTM, thx

PR #1330, opened 2023-03-15
gzip the production asset wasm

frederikrothenberger:
This PR introduces gzipping the production wasm and publishing it on the releases page.
It is a prerequisite to dropping ic-cdk-optimizer in favor of ic-wasm, which does not shrink the wasm as much but offers convenient editing of the wasm metadata.
This in turn will prepare II pipeline for upcoming requirements with regards to the canister metadata, such as providing the candid interface or metadata for dfx pull.

The release build check / clean build checks are not yet switched because it requires at least one release that provides gzipped assets.


scripts/build
```diff
@@ -127,6 +127,7 @@ function build_canister() {
         ic-cdk-optimizer \
             "$CARGO_TARGET_DIR/$TARGET/release/$canister.wasm" \
             -o "./$canister.wasm"
+        gzip -n -f -k "$canister.wasm"
```
nmattia, line 130: can you use the long-form options?

nmattia (approved):
> This in turn will prepare II pipeline for upcoming requirements with regards to the canister metadata, such as providing the candid interface or metadata for dfx pull.

😬

PR #1331, opened 2023-03-16
Recovery-phrase CSS & Design cleanup

meodai:
- Makes sure the editable pills look more interactive and the disabled ones more "passive" (There was a z-index issue that made the "blur" effect not look as it was intended)
- Autofocus the first editable element.
- Style the focus of every pill
- Clean up the CSS and add comments where something could be unclear

- [x] Tested in latest Firefox, Chrome and Safari.

https://user-images.githubusercontent.com/608386/225722496-186f2134-b353-486f-a368-f324fad5fb51.mov

Screenshot 2023-03-16 at 19 40 29


src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -38,14 +38,27 @@ const confirmSeedPhraseTemplate = ({
...
+
   // All words, where a `Ref` was added if the word needs checking
   const words: Word[] = words_.map((word) => {
+    let isFirstInput = false;
+
     if (word.check) {
       const elem: Ref<HTMLInputElement> = createRef();
       // NOTE: typescript can't follow if word is deconstructed with {...word}
-      return { word: word.word, check: word.check, elem };
+      if (isFistCheckedWord) {
```
peterpeterparker, line 50: can you find another way to solve this? mixing imperative and declarative code is a bit an anti-pattern and therefore prone to error

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -156,6 +169,7 @@ export const wordTemplate = ({
       type="text"
       class="c-recoveryInput"
       ${ref(word.elem)}
+      ?autofocus=${word.autofocus === true}
```
peterpeterparker, line 172: dumb question: what's the question mark `?`? is that a lit thing?

peterpeterparker (commented):
In addition to my two comments, may I ask: does the auto-focus works on mobile devices too? if yes, does it automatically open the keyboard and is it expected? Just asking in case not.

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
nmattia: I agree with @peterpeterparker; I think using [`words.find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) would be cleaner

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
nmattia: [yes!](https://lit.dev/docs/templates/expressions/#boolean-attribute-expressions)

nmattia, line 172: nice, I didn't know about `autofocus`! We could use it in other places too then (different PRs). Once thing though, it looks like it doesn't work on iOS: https://caniuse.com/?search=autofocus

nmattia (commented):
Looks really nice thanks! LGTM modulo the comment by @peterpeterparker

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
meodai: It is, but It I think its by design in the case of IOS, it would be a bit disturbing if the keyboard would show up all of the sudden. But if you think its still a good idea to do it on IOS as well, I can try doing in the JS way

meodai: @peterpeterparker yes it sets the property only if the computed value is true(isch)

nmattia: I get the virtual keyboard argument and don't care too strongly. If you do want to do it the js way, you can reuse this: https://github.com/dfinity/internet-identity/blob/d8a9371ec23d1771b6ec0ae8c3209799e3ac2fb5/src/frontend/src/components/anchorPicker.ts#L49

Up to you!

meodai: @nmattia I don't min changing it to the JS way. After @peterpeterparker question I thought its kinda nice that it does not do it on IOS. About the `words.find()` I wanted to do something like that first but then thought it was less performant, but I should probably only care about performance if it becomes a problem. I am changing it to find. Its gonna be easier to understand and maintain.

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
meodai: @nmattia I think its cleaner now. Let me know what you think.

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -12,7 +12,7 @@ import copyJson from "./confirmSeedPhrase.json";
 // A list of words, where "check" indicates if the user needs to double check (re-input) a word
 type Word = { word: string } & (
   | { check: false }
-  | { check: true; elem: Ref<HTMLInputElement> }
+  | { check: true; elem: Ref<HTMLInputElement>; shouldFocus?: boolean }
```
nmattia, line 15: What happens if I pass `shouldFocus: true` to more than one word in `confirmSeedPhraseTemplate`?

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
```diff
@@ -43,12 +43,22 @@ const confirmSeedPhraseTemplate = ({
     if (word.check) {
       const elem: Ref<HTMLInputElement> = createRef();
       // NOTE: typescript can't follow if word is deconstructed with {...word}
-      return { word: word.word, check: word.check, elem };
+      return {
+        word: word.word,
+        check: word.check,
+        elem,
+      };
```
nmattia, line 50: did something change here? are you running a different version of prettier?

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
meodai: only the first one will focus: https://cdpn.io/pen/debug/eYLLMLQ/bf640e6faa44d15f891c9e1d55b7f9d2 (so theoretically we could all set them to autofocus :D but I would find this strange plus it would go against the spec) MDN sais:

> No more than one element in the document or dialog may have the autofocus attribute. If applied to multiple elements the first one will receive focus.

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
meodai: I used `npm run format` as always. My editor thought it was a great idea to put it on new lines after I have removed the property I had added earlier. Apparently prettier likes both.

src/frontend/src/flows/recovery/confirmSeedPhrase.ts
nmattia: heh sorry that was a rhetorical question. What I meant is: why does the caller of `confirmSeedPhraseTemplate` need to care about this? Also why is `shouldFocus` optional? see what we do with `elem`, which (like `shouldFocus`) is data internal to the function, but added to `Word` for convenience.

I suspect `shouldFocus` was made optional just because, when calling `confirmSeedPhraseTemplate`, it is not set; but really it shouldn't even be "settable" at all.

nmattia: I didn't try running this, but I think you can easily mark `shouldFocus` an not optional, and avoid `"elem" in ...` bit:

```diff
diff --git a/src/frontend/src/flows/recovery/confirmSeedPhrase.ts b/src/frontend/src/flows/recovery/confirmSeedPhrase.ts
index 7339f9ee..de4288ec 100644
--- a/src/frontend/src/flows/recovery/confirmSeedPhrase.ts
+++ b/src/frontend/src/flows/recovery/confirmSeedPhrase.ts
@@ -12,7 +12,7 @@ import copyJson from "./confirmSeedPhrase.json";
 // A list of words, where "check" indicates if the user needs to double check (re-input) a word
 type Word = { word: string } & (
   | { check: false }
-  | { check: true; elem: Ref<HTMLInputElement>; shouldFocus?: boolean }
+  | { check: true; elem: Ref<HTMLInputElement>; shouldFocus: boolean }
 );
 
 // A list of indices nicely spread over the 25 words (anchor + 24 BIP39)
@@ -31,7 +31,7 @@ const confirmSeedPhraseTemplate = ({
   back,
   i18n,
 }: {
-  words: Omit<Word, "elem">[];
+  words: Omit<Word, "elem" | "shouldFocus">[];
   confirm: () => void;
   back: () => void;
   i18n: I18n;
@@ -47,16 +47,18 @@ const confirmSeedPhraseTemplate = ({
         word: word.word,
         check: word.check,
         elem,
+        shouldFocus: false,
       };
     } else {
       return { word: word.word, check: word.check };
     }
   });
 
-  // find the first editable word and focus it
-  const firstEditable = words.find((w) => w.check);
-  if (firstEditable && "elem" in firstEditable) {
-    firstEditable.shouldFocus = true;
+  for(const word of words) {
+      if(word.check) {
+          word.shouldFocus = true;
+          break;
+      }
   }
 
   // if all "check" words have been re-input correctly
```

meodai: good point, thats much nicer.

peterpeterparker (approved):
Rubber stamp

nmattia:
Screenshot 2023-03-21 at 12 37 16

@peterpeterparker @meodai FYI, probably worth an `npm run test` locally

PR #1332, opened 2023-03-17
Refactor auth to give access to device used

frederikrothenberger:
This is a refactoring in preparation of introducing domain specific anchor activity stats.


przydatek (approved):
LGTM.

PR #1333, opened 2023-03-17
Update chromedriver

frederikrothenberger:
GitHub has update chrome on their runners, which causes the demo dev-build test to fail. Update chromedriver to fix.


przydatek (approved):
LGTM

PR #1334, opened 2023-03-17
Move active anchor stats maintenance into its own submodule

frederikrothenberger:
This PR just moves code. This makes it easier to separate the actual activity counting from activity unrelated maintenance. This is done in preparation for the domain specific anchor activity stats.


przydatek (approved):
LGTM.

PR #1336, opened 2023-03-17
Separate anchor activity bookkeeping from post-operation bookkeeping

frederikrothenberger:
In addition, the `authenticated_anchor_operation` wrapper is made more flexible to allow also funneling through return values. This has two benefits:
* verifyTentativeDevice can now also make use of the wrapper
* it prepares the other anchor operations to be able to provide proper errors (which will be required in the future)

### DIFF
--- a/src/internet_identity/src/anchor_management.rs
+++ b/src/internet_identity/src/anchor_management.rs
@@ -55,7 +55,9 @@ pub fn get_anchor_info(anchor_number: AnchorNumber) -> IdentityAnchorInfo {
 /// * Records anchor activity within aggregated activity stats
 /// * Updates the last usage timestamp
 ///
-/// Note: modifies the anchor but not does not write to storage.
+/// Note: modifies the anchor but not does not write to storage. It is the responsibility of the
+/// caller to persist the changes. This allows anchor operations to write to storage only once,
+/// combining the modifications for bookkeeping reasons (made here) with other changes to the anchor.
 pub fn activity_bookkeeping(anchor: &mut Anchor, current_device_key: &DeviceKey) {
     let previous_activity = anchor.last_activity();
     active_anchor_stats::update_active_anchors_stats(previous_activity);
--- a/src/internet_identity/src/anchor_management.rs
+++ b/src/internet_identity/src/anchor_management.rs
@@ -51,20 +51,27 @@ pub fn get_anchor_info(anchor_number: AnchorNumber) -> IdentityAnchorInfo {
     })
 }
 
+/// Handles all the bookkeeping required on anchor activity:
+/// * Records anchor activity within aggregated activity stats
+/// * Updates the last usage timestamp
+///
+/// Note: modifies the anchor but not does not write to storage.
+pub fn activity_bookkeeping(anchor: &mut Anchor, current_device_key: &DeviceKey) {
+    let previous_activity = anchor.last_activity();
+    active_anchor_stats::update_active_anchors_stats(previous_activity);
+    anchor
+        .set_device_usage_timestamp(current_device_key, time())
+        .expect("unable to update last usage timestamp");
+}
+
 /// Handles all the bookkeeping required after a successful anchor operation:
 /// * Adds the operation to the archive buffer
 /// * Increments the anchor operation counter
-/// * Records anchor activity
-pub fn anchor_operation_bookkeeping(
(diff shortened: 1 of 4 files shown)


src/internet_identity/src/anchor_management.rs
```diff
@@ -51,20 +51,27 @@ pub fn get_anchor_info(anchor_number: AnchorNumber) -> IdentityAnchorInfo {
     })
 }
 
+/// Handles all the bookkeeping required on anchor activity:
+/// * Records anchor activity within aggregated activity stats
+/// * Updates the last usage timestamp
+///
+/// Note: modifies the anchor but not does not write to storage.
```
przydatek, line 58: nit: maybe add more explanation of why this is the case (as we had it for `update_last_device_usage`) -- otherwise it is not clear why this exceptional treatment.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -86,15 +89,16 @@ pub async fn add_tentative_device(
 }
 
 pub fn verify_tentative_device(
+    anchor: &mut Anchor,
     anchor_number: AnchorNumber,
     user_verification_code: DeviceVerificationCode,
-) -> VerifyTentativeDeviceResponse {
+) -> Result<(VerifyTentativeDeviceResponse, Operation), VerifyTentativeDeviceResponse> {
```
przydatek, line 95: Returning `Result<(R, ...), R>` looks like an unusual pattern -- maybe comment this function, especially what is semantics of the returned `R` in both cases?  Or maybe this pattern is not necessary?

src/internet_identity/src/main.rs
```diff
@@ -363,36 +374,43 @@ fn update_root_hash() {
...
+    anchor_management::activity_bookkeeping(&mut anchor, &device_key);
+    state::storage_mut(|storage| storage.write(anchor_number, anchor)).unwrap_or_else(|err| {
+        panic!("last_usage_timestamp update: unable to update anchor {anchor_number}: {err}")
+    })
 }
 
 /// Authenticates the caller (traps if not authenticated) calls the provided function and handles all
 /// the necessary bookkeeping for anchor operations.
-fn authenticated_anchor_operation(
+fn authenticated_anchor_operation<R>(
```
przydatek, line 391: Also here documenting the intended role of `R` and the meaning of the result of `op` would be useful.

przydatek (commented):
Sorry for the delay @frederikrothenberger.  Please see some questions below.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
frederikrothenberger: I agree, I'm not quite happy with the interface either. The reason we get `Result<R,R>` is basically that we have variants on the `candid` interface which unify success and error conditions on the same type. E.g.:
```
type RegisterResponse = variant {
    // A new user was successfully registered.
    registered: record {
        user_number: UserNumber;
    };
    // No more registrations are possible in this instance of the II service canister.
    canister_full;
    // The challenge was not successful.
    bad_challenge;
};
```
Also, the `authenticated_anchor_operation` requires information about an operation to do bookkeeping and archiving.


So, we could go with `(R, Option<Operation>)`, but this would then (from a typing perspective) indicate that the `Operation` is entirely optional (which it is not).

Or we could split the interface to `Result<R,E>` and only assemble the final response type outside of the `authenticated_anchor_operation`. I think, this would be the "nice" solution, but introduce quite some additional internal types. So essentially, the current design is a compromise between having accurate types and still being able to communicate success / failure to the caller.

przydatek: As discussed, I agree that the "nice" solution would be probably an overkill, so leaving the interface as-is SGTM.  Maybe refer in a comment to the documentation of `authenticated_anchor_operation`, so that it is clear why the `Result...` has this structure?

src/internet_identity/src/main.rs
przydatek: Maybe extend the comment, with an explanation that `R` is usually bound to a candid-type, which contains both successful and failure variants, hence it can be used in both positions of `Result<T, E>`?

przydatek (approved):
LGTM, with two additional minor suggestions.  Thanks!

PR #1337, opened 2023-03-17
Make active anchor counter counting logic pluggable

frederikrothenberger:
This PR serparates the logic responsible for iterating over counters from the logic on how to update a counter. This is a refactoring in preparation for the domain specific activity counters.


przydatek (approved):
LGTM, thanks!

PR #1339, opened 2023-03-20
Polish recovery flow

nmattia:
This makes a few edits to the recovery flow:

* Copy is updated to be clearer
* Copy is updated to be tailored to the flow the user is in (i.e. either "creating" a phrase for the for time or "resetting" a phrase)
* "Cancel" button is added
* Recovery tests are updated to not rely on specific copy (instead use a custom attribute)


src/frontend/src/flows/recovery/displaySeedPhrase.json
```diff
@@ -1,13 +1,14 @@
 {
   "en": {
-    "title": "Store Recovery Phrase",
-    "header": " If you lose access to your devices, use your recovery phrase to access your Internet Identity.",
-    "your_recovery_phrase": "Your recovery phrase",
+    "title": "Safely Store your Recovery Phrase",
```
peterpeterparker, line 3: shouldn't it be "store" (lower letter as in the verb)?

src/frontend/src/flows/recovery/displaySeedPhrase.ts
```diff
@@ -9,14 +9,19 @@ import { toast } from "../../components/toast";
...
 const displaySeedPhraseTemplate = ({
+  op,
   seedPhrase,
   onContinue,
+  cancel,
   copyPhrase: copyPhrase_,
   i18n,
 }: {
+  /* whether the phrase is created for the first time or just reset */
+  op: "create" | "reset";
```
peterpeterparker, line 20: `op` as `operation`?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/recovery/displaySeedPhrase.ts
nmattia: yeah, do you think `operation` is clearer? any other suggestions? I struggled a bit with this one

src/frontend/src/flows/recovery/displaySeedPhrase.json
peterpeterparker: so "Your" not "your" then according to https://capitalizemytitle.com/ 😉

nmattia: I'll forward this to the relevant people 😄 I don't care one bit

PR #1342, opened 2023-03-20
Statically host google fonts 

frederikrothenberger:
By including fonts as statically hosted assets we can tighten the CSP
to no longer allow requests to google.


frederikrothenberger:
For some reason changing the font changed the font-rendering on mobile a bit. I discussed with @meodai and he was happy enough with the new rendering. WDYT @nmattia?

src/frontend/src/styles/main.css
```diff
@@ -45,7 +45,19 @@
...
 */
 
-@import url("https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap");
+/* Fonts */
+
+/* latin */
+@font-face {
+  font-family: "Inter";
+  font-style: normal;
+  font-weight: 100 900;
```
nmattia, line 54: previous was `100..900`, can you confirm that this is the same as `100 900` here? I would expect the first to list all weights from 100 to 900, whereas here it looks like it'd only expect weights 100 and 900, but I haven't checked

nmattia (approved):
Yeah works for me! although it looks like something fishy's going on if everything changes whenever we pull the font from somewhere else

src/frontend/src/styles/main.css
frederikrothenberger: If you look at the resource at https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap, you'll find that it generates CSS blocks like this:
```css
/* latin */
@font-face {
  font-family: 'Inter';
  font-style: normal;
  font-weight: 100 900;
  font-display: swap;
  src: url(https://fonts.gstatic.com/s/inter/v12/UcC73FwrK3iLTeHuS_fvQtMwCp50KnMa1ZL7W0Q5nw.woff2) format('woff2');
  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
}
```

So yes, the CSS was always like this.

frederikrothenberger:
> although it looks like something fishy's going on if everything changes whenever we pull the font from somewhere else

Well, in this case only the mobile screenshots changed (desktop are all still the same). So to me this seems more like something fishy with mobile emulation...

PR #1343, opened 2023-03-21
Update recovery confirmation indices

nmattia:
Only ask to confirm anchor, first and last word (requested by product), and fix an index rendering issue (indices were shown as 0-indexed).


peterpeterparker (approved):
Rubber stamp

PR #1348, opened 2023-03-22
Wait for alert to be displayed in selenium tests

frederikrothenberger:
I noticed increased flakyness on selenium tests due to the alert not yet being open. This should fix it.


nmattia (approved):
Thanks!

PR #1349, opened 2023-03-23
Actually abort phrase setup on cancel

nmattia:
This fixes a bug where canceling the recovery phrase setup didn't abort the phrase setup/replacement procedure.

This is a three part PR:
* The first commit adds a test that ensures that, when a phrase is reset, the _new_ phrase can be used to recover
* The second commit adds a (failing) test that ensures that, when the reset procedure is canceled, the _old_ phrase can be used to recover
* The third commit fixes that actual issue.

This issue came from the return value of `displayAndConfirmPhrase` not being checked. The value is now checked, and upon confirmation, the phrase is set up/created.

@peterpeterparker I'm not super happy with the solution because it won't prevent the issue from happening again, and duplicates the "confirmation checking" logic. Do you think passing a callback (e.g. `onConfirmation: () => void`) would make more sense? Any other suggestions?

### DIFF
--- a/src/frontend/src/flows/manage/deviceSettings.ts
+++ b/src/frontend/src/flows/manage/deviceSettings.ts
@@ -138,7 +138,7 @@ export const resetPhrase = async ({
 
     const res = await displayAndConfirmPhrase({ phrase, operation: "reset" });
 
-    if (res === "ok") {
+    if (res === "confirmed") {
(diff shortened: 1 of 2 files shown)


peterpeterparker (approved):
LGTM, two minor suggestions

PR #1350, opened 2023-03-23
Add domain specific active anchor statistics 

frederikrothenberger:
This PR introduces additional statistics about active anchors
based on the domain of the device that was used for the activity.

The stats are collected only for the well-known II domains
(and the combinations of both).

Caveat: the stats will be inaccurate for identity.ic0.app because
most devices there still do not have domain information attached.
This issue will be addressed in a follow-up PR, where domain
information will be added on activity if it is missing.


src/internet_identity/src/active_anchor_stats.rs
```diff
@@ -62,22 +76,100 @@ fn update_active_anchor_counter(
...
+            IC0_APP_ORIGIN => Ok(IIDomain::Ic0AppDomain),
+            INTERNETCOMPUTER_ORG_ORIGIN => Ok(IIDomain::InternetComputerOrgDomain),
+            _ => Err(()),
+        }
+    }
+}
+
+/// Increases the counter on a window for a single II domain if the there was no activity before.
+///
+/// If there has been activity on an II domain and this is activity on the _other_ domain, then
```
przydatek, line 117: "the _other_ domain" here can be (mis)understood as referring to `DomainActivity::Other`, while it refers indeed to the other **II** domain, right?  Maybe rephrase the comment and/or rename `DomainActivity::Other`? (as the name also "collides" with `IIDomain::other_domain()`

src/internet_identity/src/storage/anchor.rs
```diff
@@ -195,6 +196,72 @@ impl Anchor {
...
+    /// * no information is most likely caused by the device having been added before we started
+    ///   collecting domain information
+    /// * combinations of an unknown domain and an II domain shows that the anchor is at least partially
+    ///   active on the II domain (but also does non-standard / unsupported things to their anchor).
+    ///   If we are interested in this user group, we might consider extending this function to give
+    ///   them their own [DomainActivity] value.
+    pub fn domain_activity_since(&self, timestamp: Timestamp) -> DomainActivity {
+        #[derive(Default)]
+        struct Accumulator {
+            other: bool,
```
przydatek, line 212: nit: maybe rename `other` to something like `non_ii`?

src/internet_identity/src/storage/anchor.rs
```diff
@@ -195,6 +196,72 @@ impl Anchor {
...
+                    INTERNETCOMPUTER_ORG_ORIGIN => acc.internet_computer_org = true,
+                    _ => acc.other = true,
+                };
+                acc
+            });
+
+        // Activity on other domains is discarded if there is also activity on an II domain.
+        // The reason is that II might not have complete information since domain information was
+        // only introduced recently.
+        match (result.ic0_app, result.internet_computer_org, result.other) {
```
przydatek, line 243: nit: the order of the fields in `match` is different than the order of the fields in the definition of `Accumulator`, which then needs a second look to make sure that the code is correct.  Maybe use the same order in both places?

przydatek (approved):
LGTM, but please see some minor questions and suggestions below.
(and sorry for the delay in reviewing)

src/internet_identity/src/active_anchor_stats.rs
frederikrothenberger: I did all of those things, thanks for pointing it out:
* renamed `DomainActivity::Other` to `DomainActivity::NonIIDomain`
* renamed `IIDomain::other_domain` to `IIDomain::other_ii_domain`
* clarified the comment to say "other II domain"

PR #1353, opened 2023-03-24
Clarify devices and recovery setup

nmattia:
This includes a couple UX fixups:

* The copy is updated when resetting a phrase to clarify the current phrase will be replaced
* The message shown when the user has few devices is updated to be clearer


nmattia:
Nevermind; sorry for the noise

PR #1354, opened 2023-03-24
Refactor II interface crate

frederikrothenberger:
This refactoring restructures the internet_identity_interface crate such that:
* the II modules and archive specific modules are now siblings
* the types and impls are separated into submodules (this will be useful once the types are generated from candid)

The better decoupling from types and impls opens up the crate to have more implementations for the types, without compromising on the ability to generate those types from candid.


przydatek (approved):
LGTM, thanks!

PR #1355, opened 2023-03-24
Regroup phrase creation functions

nmattia:
This introduces `phraseWizard`, a function that creates a phrase, asks user for confirmation, and uploads the phrase. This avoids duplicating the phrase creation and confirmation code.

A helper is also introduced, `assertType`, used to exhaust type altneratives. This simplifies handling of return values.

### DIFF
--- a/src/frontend/src/flows/manage/deviceSettings.ts
+++ b/src/frontend/src/flows/manage/deviceSettings.ts
@@ -6,18 +6,15 @@ import {
 } from "../../utils/iiConnection";
 import { displayError } from "../../components/displayError";
 import { withLoader } from "../../components/loader";
-import { unreachable, unreachableLax } from "../../utils/utils";
+import { unreachable, unknownToString, assertType } from "../../utils/utils";
 import { DeviceData } from "../../../generated/internet_identity_types";
 import { phraseRecoveryPage } from "../recovery/recoverWith/phrase";
-import { displayAndConfirmPhrase } from "../recovery/setupRecovery";
+import { phraseWizard } from "../recovery/setupRecovery";
 import {
   isRecoveryDevice,
   isProtected,
   RecoveryPhrase,
 } from "../../utils/recoveryDevice";
-import { generate } from "../../crypto/mnemonic";
-import { fromMnemonicWithoutValidation } from "../../crypto/ed25519";
-import { IC_DERIVATION_PATH } from "../../utils/iiConnection";
 
 /* Remove the device and return */
 export const deleteDevice = async ({
@@ -84,7 +81,7 @@ export const resetPhrase = async ({
   connection: AuthenticatedConnection;
   device: DeviceData & RecoveryPhrase;
   reload: (connection?: AuthenticatedConnection) => void;
-}) => {
+}): Promise<void> => {
   const confirmed = confirm(
     "Reset your Recovery Phrase\n\nWas your recovery phrase compromised? Delete your recovery phrase and generate a new one by confirming."
   );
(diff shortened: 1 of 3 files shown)


src/frontend/src/utils/utils.ts
```diff
@@ -1,3 +1,8 @@
+// Ensures the argument has a given type (from ts' perspective, not js)
+export function assertType<T>(_: T) {
```
nmattia, line 2: @peterpeterparker do you know if there's a ts builtin that could be used here?

peterpeterparker: That's the assertion function we are using:

```
export const assertNonNullish: <T>(
  value: T,
  message?: string
) => asserts value is NonNullable<T> = <T>(
  value: T,
  message?: string
): void => {
  if (value === null || value === undefined) {
    throw new NullishError(message);
  }
};
```

`asserts` is the TypeScript keyword.

https://github.com/dfinity/ic-js/blob/main/packages/utils/src/utils/asserts.utils.ts

It throws an error.

src/frontend/src/flows/manage/deviceSettings.ts
```diff
@@ -122,37 +105,43 @@ export const resetPhrase = async ({
...
+    return reload(nextConnection);
+  } else if (typeof res === "object" && "error" in res) {
     await displayError({
       title: "Could not reset recovery phrase",
       message: "An unexpected error occurred",
-      detail: e instanceof Error ? e.toString() : "unknown error",
+      detail: unknownToString(res.error, "unknown error"),
       primaryButton: "Ok",
     });
+    return reload();
```
peterpeterparker, line 140: nth: you can spare on


```
  } else if (typeof res === "object" && "error" in res) {
     ...
  } else {
    assertType<"canceled">(res);
  }

  return reload();
}

File: src/frontend/src/utils/utils.ts (hunk 1 above)
Comment (nmattia) [reply src/frontend/src/utils/utils.ts:2]: I know that one, I've read the blog post 😉 

`assertTy` does something different, in that it doesn't care about the JS value at all, but only serves to assert the TypeScript type

File: src/frontend/src/flows/recovery/setupRecovery.ts [hunk 3]
```diff
@@ -131,40 +134,62 @@ export const setupPhrase = async (
   userNumber: bigint,
   connection: AuthenticatedConnection
 ): Promise<"ok" | "error" | "canceled"> => {
-  const name = "Recovery phrase";
+  const res = await phraseWizard({
+    userNumber,
+    operation: "create",
+    uploadPhrase: (pubkey) =>
+      withLoader(() =>
+        connection.add(
```
peterpeterparker, line 142: existing code that's why no object is use for the `.add()` parameter?

src/frontend/src/flows/recovery/setupRecovery.ts
```diff
@@ -131,40 +134,62 @@ export const setupPhrase = async (
...
+// Set up a recovery phrase
+export const phraseWizard = async ({
+  userNumber,
+  operation,
+  uploadPhrase,
+}: {
+  userNumber: bigint;
+  operation: "create" | "reset";
+  uploadPhrase: (pubkey: DerEncodedPublicKey) => Promise<void>;
+}): Promise<{ ok: SignIdentity } | { error: unknown } | "canceled"> => {
```
peterpeterparker, line 170: any particular reason not using `canceled` as one of the variant?

that would spare these `typeof res === "object"` checks

```
Promise<{ ok: SignIdentity } | { error: unknown } | {canceled: never}>
```

P.S.: not sure about the `never` above

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/manage/deviceSettings.ts
nmattia: true, that's kind of what the original code was doing, although I found it less clear/explicit. I don't mind the extra line but I can move it if you feel strongly about it!

src/frontend/src/flows/recovery/setupRecovery.ts
peterpeterparker: Not strong on my side but removing `typeof res === "object"` would have ease my reading

PR #1357, opened 2023-03-24
Fix not waiting for the recovery phrase to be added

frederikrothenberger:
This PR adds an additional check in the `addRecoveryMechanismSeedPhrase` flow to ensure that the phrase is actually added before the test continues.


nmattia (approved):
Thanks!

PR #1361, opened 2023-03-24
Update test readme

frederikrothenberger:
The test readme has gone out of date.
Changes made:
* Fix flags required to build test flavour
* Include building the archive
* Include downloading the latest archive release in instructions

### DIFF
--- a/src/internet_identity/tests/README.md
+++ b/src/internet_identity/tests/README.md
@@ -4,10 +4,16 @@ This is a test suite for the Internet Identity canister. To run the tests:
 
 ``` bash
 # Make sure II is built with the "test" flavor
-II_DUMMY_CAPTHA=1 ./scripts/build
+II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=1 ./scripts/build
+
+# Build the archive canister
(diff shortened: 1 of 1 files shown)


src/internet_identity/tests/README.md
```diff
@@ -4,11 +4,17 @@ This is a test suite for the Internet Identity canister. To run the tests:
 
 ``` bash
 # Make sure II is built with the "test" flavor
-II_DUMMY_CAPTHA=1 ./scripts/build
+II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=1 ./scripts/build
```
nmattia, line 7: out of curiosity, why do these tests need II_FETCH_ROOT_KEY?

frederikrothenberger: Actually, they don't. I added the flag to be consistent with the description in the main README (and the comment here).

But I can also revert that, but then I would not call it the `test` flavor here.

nmattia: just curious :) thanks!

przydatek (approved):
LGTM, thanks!

PR #1364, opened 2023-03-25
Use 'satisfies' instead of 'assertType'

nmattia:
TypeScript 4.9 shipped with the `satisfies` keyboard which can be used instead of our custom `assertType` to ensure a value satisfies a given type.


peterpeterparker (approved):
Coolio, thx

PR #1367, opened 2023-03-27
Make release(-notes) a normal composite action

frederikrothenberger:
This removes the node scaffolding that was required to run the action.

Test run of the composite action: https://github.com/dfinity/internet-identity/actions/runs/4532626193/jobs/7984382416#step:8:47


.github/actions/release/run
```diff
@@ -166,5 +166,3 @@ cat "$section_wasm_verification" >> "$body" && echo >> "$body" && rm "$section_w
 cat "$body"
 
 echo "notes-file=$body" >> "$GITHUB_OUTPUT"
-
-[ -n "${GITHUB_ENV:-}" ] && echo "RELEASE_ACTION_BODY_FILE=$body" >>"$GITHUB_ENV"
```
nmattia, line 170: What was this for?

nmattia (commented):
Thanks! ~~For context, this action predates composite actions~~

EDIT: I think that's the reason actually: https://github.com/actions/runner/issues/1478

.github/actions/release/run
frederikrothenberger: This used to set the filename as an env variable in Github (see [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files)). But we don't actually use it, so I removed it.

nmattia: Are there no other occurrences of `RELEASE_ACTION_BODY_FILE`?

nmattia (approved):

PR #1368, opened 2023-03-28
Fix issues in release-script action

frederikrothenberger:
This PR fixes various issues in the release-script action:
* Fix links to CI runs to verify hashes for filenames that are a subset of each other
* Explicit production asset so that the link in the intro is not repeated for the wasm and wasm.gz
* Update readme with latest changes
* Move shellcheck suppression to the appropriate line

### DIFF
--- a/.github/actions/release/run.sh
+++ b/.github/actions/release/run.sh
@@ -7,6 +7,7 @@ set -euo pipefail
 
 GITHUB_TOKEN=${INPUT_TOKEN:-${GITHUB_TOKEN:?No token given}}
 
+PRODUCTION_ASSET=${INPUT_PRODUCTION_ASSET:?No production asset specified}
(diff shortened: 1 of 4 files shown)


.github/actions/release/run.sh
```diff
@@ -82,9 +83,9 @@ do
         # Additionally grab the step number of the 'sha256sum' step
         step=$(curl --silent "https://api.github.com/repos/dfinity/internet-identity/actions/runs/$GITHUB_RUN_ID/jobs" \
             | jq -cMr \
-            --arg job_id "$job_id" \
+            --argjson job_id "$job_id" \
```
frederikrothenberger, line 86: Turns out this change actually broke the `jq` command due to types. `argjson` parses the value as `number` whereas `arg` always parses to `string`

nmattia (approved):
Rubber stamp

PR #1369, opened 2023-03-28
Add all II flavours as gzipped variant to releases

frederikrothenberger:
The switch to ic-wasm makes the not-gzipped wasm file too large for deployment in a single message (i.e. what `dfx deploy` does).

In order to have access to a deployable version of the latest prebuilt II, the releases should have all flavours attached as gzipped variant too.

I have updated the [latest release](https://github.com/dfinity/internet-identity/releases/tag/release-2023-03-27) manually now.


przydatek (approved):
LGTM, thanks!

PR #1370, opened 2023-03-28
Switch from ic-cdk-optimizer to ic-wasm

frederikrothenberger:
This PR introduces `ic-wasm` in place of the deprecated `ic-cdk-optimizer`.
`ic-wasm` will allow to modify the medatada sections of the generated
wasm files.

However, `ic-wasm shrink` does not achieve the same
level of size reduction as `ic-cdk-optimizer` did. This is why
the production asset will be switched to the gzipped wasm in an upcoming
PR.

### DIFF
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -28,14 +28,14 @@ runs:
                 ;;
         esac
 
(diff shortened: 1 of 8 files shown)


przydatek (approved):
LGTM, thanks!

PR #1372, opened 2023-03-28
Switch the production asset to the gzipped version

frederikrothenberger:
This PR switches the production asset to be the gzipped wasm instead of the uncompressed wasm. This means that the build recproducibility checks are run against the gzipped wasm as well.

Note: Since the release-build-check checks the release build using the check-build action of when the release was made, we cannot yet adapt this workflow. This must be done after the next release.


przydatek (approved):
LGTM, thanks!

PR #1373, opened 2023-03-28
Replace __get_candid_interface_tmp_hack with metadata section

frederikrothenberger:
This replaces the candid interface hack with the proper public metadata section added with `ic-wasm`.


przydatek (approved):
LGTM, thanks!
(and sorry for the delay!)

PR #1375, opened 2023-03-29
Collect list of latest used delegation origins

frederikrothenberger:
This PR introduces the list of latest used delegation origins
containing origins that had at least one delegation issued in the
last 30 days. At most `max_num_latest_delegation_origins`
(default 1000) origins are tracked. If the limit is reached, the
least recently used origin is discarded.

This is part of a bigger feature to provide a list of dapps integrating
with II in the II front-end.

### DIFF
--- a/src/internet_identity/src/delegation.rs
+++ b/src/internet_identity/src/delegation.rs
@@ -1,4 +1,4 @@
-use crate::state::AssetHashes;
+use crate::state::{persistent_state_mut, AssetHashes};
 use crate::{hash, state, update_root_hash, DAY_NS, LABEL_ASSETS, LABEL_SIG, MINUTE_NS};
 use candid::Principal;
 use ic_cdk::api::{data_certificate, time};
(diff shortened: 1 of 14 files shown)


src/internet_identity/src/delegation.rs
```diff
@@ -45,15 +45,51 @@ pub async fn prepare_delegation(
...
+fn delegation_bookkeeping(frontend: FrontendHostname) {
+    state::usage_metrics_mut(|metrics| {
+        metrics.delegation_counter += 1;
+    });
+    update_latest_delegation_origins(frontend);
+}
+
+/// Add the current front-end to the list of latest used front-end origins.
+fn update_latest_delegation_origins(frontend: FrontendHostname) {
+    let now = time();
```
przydatek, line 66: nit: maybe rename to `now_ns`, so that when it is used later (e.g. line 80 below), it is clear it is in nanoseconds?  (`timestamp` is somewhat more clear by convention, although using `timestamp_ns` could be also considered)

src/internet_identity/src/delegation.rs
```diff
@@ -45,15 +45,51 @@ pub async fn prepare_delegation(
...
+        // drop entries older than 30 days
+        latest_delegation_origins.retain(|_, timestamp| now - *timestamp < 30 * DAY_NS);
+
+        // if we still have too many entries, drop the oldest
+        if latest_delegation_origins.len() as u64
+            > persistent_state.max_num_latest_delegation_origins.unwrap()
+        {
+            let mut values: Vec<_> = latest_delegation_origins.clone().into_iter().collect();
+            values.sort_by(|(_, timestamp_1), (_, timestamp_2)| timestamp_1.cmp(timestamp_2));
+            latest_delegation_origins.remove(&values[0].0);
```
przydatek, line 88: It seems a bit unfortunate that once we fill up `latest_delegation_origins` this relatively costly operation will be executed on each insertion of a new origin (i..e clone the values, sort them, remove the single oldest one).  I suppose with the current usage filling up is probably not a concern, but maybe in the future we should change the structure to something that would allow for more efficient removal?  (i.e. I believe there is no need for change right now, but maybe we should add a comment for the future?)

src/internet_identity/tests/integration/latest_delegation_origins.rs
```diff
@@ -0,0 +1,115 @@
...
+    let latest_origins = api::stats(&env, canister_id)?.latest_delegation_origins;
+    assert_eq!(latest_origins, vec!["https://some-dapp.com".to_string()]);
+    Ok(())
+}
+
+/// Verifies that the configured limit for the number of latest used origins is respected.
+#[test]
+fn should_record_max_delegation_origins() -> Result<(), CallError> {
+    let env = env();
+    let canister_id = install_ii_with_low_latest_origin_limit(&env);
```
przydatek, line 35: nit: maybe pass the limit as an argument, so that the code below (the for-loop, and the asserts) don't have to depend on the constant used in the helper, but rather one can use a constant local to the test?

przydatek (approved):
LGTM, with some minor suggestions, thanks!

src/internet_identity/src/delegation.rs
frederikrothenberger: I agree. Added a comment.

PR #1379, opened 2023-03-31
Update agent-js

frederikrothenberger:
This update `agent-js` to the latest version. The new version allows setting the platform binding of WebAuthn devices upon device registration. This allows us to get rid of the mechanism that sets it on device usage.


nmattia (approved):
Thanks!

PR #1381, opened 2023-04-03
Set platform attachment on device registration / addition

frederikrothenberger:
This PR add platform attachment information to newly registered / added devices.

By adding it right away
we are no longer dependent on the hacky update on usage. This will be changed in an upcoming PR.

### DIFF
--- a/src/frontend/src/utils/authenticatorAttachment.ts
+++ b/src/frontend/src/utils/authenticatorAttachment.ts
@@ -1,0 +1,21 @@
+import { KeyType } from "../../generated/internet_identity_types";
+
+/**
+ * Converts a WebAuthn authenticator attachment to an II key type
+ */
+export function authenticatorAttachmentToKeyType(
+  authenticatorAttachment: AuthenticatorAttachment | undefined
+): KeyType {
+  let keyType: KeyType = { unknown: null };
+  switch (authenticatorAttachment) {
+    case "cross-platform":
+      keyType = { cross_platform: null };
+      break;
+    case "platform":
+      keyType = { platform: null };
+      break;
+    default:
+      break;
+  }
+  return keyType;
+}
--- a/src/frontend/src/utils/authenticatorAttachment.ts
+++ b/src/frontend/src/utils/authenticatorAttachment.ts
@@ -6,16 +6,11 @@ import { KeyType } from "../../generated/internet_identity_types";
 export function authenticatorAttachmentToKeyType(
   authenticatorAttachment: AuthenticatorAttachment | undefined
 ): KeyType {
-  let keyType: KeyType = { unknown: null };
-  switch (authenticatorAttachment) {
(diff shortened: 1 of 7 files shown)


src/frontend/src/utils/authenticatorAttachment.ts
```diff
@@ -0,0 +1,21 @@
...
+    case "cross-platform":
+      keyType = { cross_platform: null };
+      break;
+    case "platform":
+      keyType = { platform: null };
+      break;
+    default:
+      break;
+  }
+  return keyType;
```
nmattia, line 20: do you think something like this would be simpler to understand?

```ts

if(authenticatorAttachment === "cross-platform") {
  return { cross_platform: null };
} else if (authenticatorAttachment === "platform") {
  return { platform: null };
} else {
  return undefined;
}
```

src/frontend/src/utils/iiConnection.ts
```diff
@@ -101,6 +102,8 @@ export type { ChallengeResult } from "../../generated/internet_identity_types";
 
 export interface IdentifiableIdentity extends SignIdentity {
   rawId: ArrayBuffer;
+
+  getAuthenticatorAttachment(): AuthenticatorAttachment | undefined;
```
nmattia, line 106: Do you think `IdentifiableIdentity` is still a good name if the interface also takes care of providing authenticator attachment data?

frederikrothenberger: Ideally, it would be called `WebAuthnIdentity` but then it clashes with the `agent-js` type. Maybe `AuthenticatorIdentity`? Or `WebAuthnSignIdentity`?

nmattia: I like `AuthenticatorIdentity`!

nmattia: Note that you can also `import { WebAuthnIdentity as OtherName } from ...` although it can be a little confusing

frederikrothenberger: > Note that you can also import { WebAuthnIdentity as OtherName } from ... although it can be a little confusing

Yes, but I didn't do that to avoid the confusion. I now called it `IIWebAuthnIdentity` and added an explanation on how it relates to  `WebAuthnIdentity`.

src/frontend/src/utils/authenticatorAttachment.ts
nmattia, line 1: Do you think we could get rid of this module by having `connection.add` take `"cross-platform" | "platform"` as the key type? Or even have `connection.add` take a device directly, instead of deconstructing it when calling `connection.add` just to reconstruct the device within `connection.add`?

nmattia (approved):

meodai:
hey! Looks like this merge broke `npm run showcase`.

PR #1382, opened 2023-04-03
Move loader CSS to main.css

meodai:
The CSS for the loader was in the loader component itself. 
Because of that I missed it when renaming the `z-index` variables.

- Move the loader styes from `loader.ts` to `main.css`
- Make use of design tokens for the loader css
- Make the CSS a bit nicer to read
- Make sure the loader icon is not "selectable"


nmattia (approved):
Thanks!

PR #1384, opened 2023-04-03
Update device origin on usage instead of key type

frederikrothenberger:
This PR replaces the hack to update the device key type (now obsolete due to 1381) with a hack to update the origin instead.

We need origin information for existing devices in order to accurately report user activity by II domain.

Since we already add origin information to new devices, this hack can be removed in a few weeks, when most active II users with old devices have had some activity.

### DIFF
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -282,7 +282,7 @@ export class Connection {
 
     // only update devices without origin information
     if (device !== undefined && device.origin.length === 0) {
-              device.origin satisfies string[];
+      device.origin satisfies string[];
       // we purposely do not await the promise as we just optimistically update
       // if it fails, no harm done
 


src/frontend/src/utils/iiConnection.ts
```diff
@@ -272,22 +272,16 @@ export class Connection {
...
-      devices,
-      attachmentInfo.credentialId
-    );
+    // Find the device based on the pubkey
+    const device = findDeviceByPubkey(devices, pubkey);
 
-    // only update devices with key-type unknown
-    if (device !== undefined && "unknown" in device.key_type) {
+    // only update devices without origin information
+    if (device !== undefined && device.origin.length === 0) {
```
nmattia, line 284:
```suggestion
    if (device !== undefined && device.origin.length === 0) {
              device.origin satisfies string[];
```

Would add that just in case we ever fix the generated interface, in which case `origin` would be a `string` but `length` would still work

nmattia (approved):

PR #1386, opened 2023-04-03
Fix Safari blur effect issue on some Inputs

meodai:
In the recovery phrase the "blur" effect showed some strange artefacts when the focus would leave the input. 
We avoid this by forcing the browser to paint this on its own `layer` (meaning the GPU handles this area separately).


src/frontend/src/styles/main.css
```diff
@@ -1374,6 +1374,14 @@ a:hover,
   opacity: 0.3;
 
   transition: 200ms opacity linear, 400ms filter cubic-bezier(0.3, 0.7, 0, 1);
+
+  /* 
+   * the blur does not render nicely in Safari if its not on its own paint layer,
```
nmattia, line 1379:
```suggestion
   * the blur does not render nicely in Safari if it's not on its own paint layer,
```

src/frontend/src/styles/main.css
```diff
@@ -1374,6 +1374,14 @@ a:hover,
...
+   * so we use will-change to let the browser know that we want to animate this
```
nmattia, line 1380:
```suggestion
   * so we use will-change to let the browser know that we want to animate
```

nmattia (approved):
Nice, thanks!

PR #1388, opened 2023-04-04
Restrict latest derivation origin list to II domains

frederikrothenberger:
Given the front-end will now update devices missing domain information, this information can now be used to restrict the latest derivation origins to origins integrating with II.


przydatek (approved):
LGTM, thanks!

PR #1392, opened 2023-04-05
Deferring storage init till InternetIdentityInit is available.

przydatek:
Changing the default `State` to be created with an uninitialised storage, so that it can be properly initialised once initialisation config is available.  
Also, fixing `save_persistent_state` to use a storage-access helper, rather than directly (this was accidentally omitted previously).


frederikrothenberger (approved):
LGTM! Thanks

PR #1393, opened 2023-04-05
Add showcase icons

frederikrothenberger:
This PR adds the showcase icons as certified assets to the II canister.

This is a prerequisite for the showcase dapp list that will make use of those icons.


frederikrothenberger:
Moved back to draft because it would be a nicer to just have the canister serve the contents of `dist` in general.

frederikrothenberger:
Superseded by #1410.

PR #1394, opened 2023-04-05
Add add-device link

nmattia:
This changes the "add device" flow slightly (when starting from an existing device) to show a link to be opened on the new device.

The link looks like this: `HOST/?action=add-device&anchor=1234`

This is then interpreted by the new device as a cue to be added tentatively to anchor `1234`.


src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
```diff
@@ -32,17 +32,14 @@ const pollForTentativeDeviceTemplate = ({
       <li>
         Open
         <em class="c-tooltip">
-          <strong class="t-strong">${LEGACY_II_URL}</strong>
+          <strong data-role="add-device-link" class="t-strong t-wrap"
+            >${link}</strong
```
peterpeterparker, line 36: would it be handy ux wise to provide the link has a link?

The device can be a device on the same laptop

src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
```diff
@@ -25,7 +25,7 @@ import { authenticatorAttachmentToKeyType } from "../../../utils/authenticatorAt
 export const registerTentativeDevice = async (
   userNumber: bigint,
   connection: Connection
-): Promise<void> => {
+): Promise<"ok"> => {
```
peterpeterparker, line 28: not sure, it isn't `"ok" | never`?

src/frontend/src/index.ts
```diff
@@ -94,6 +97,28 @@ const init = async () => {
...
   const connection = new Connection(readCanisterId());
 
+  // Figure out if user is trying to add a device. If so, use the anchor from the URL.
+  const addDeviceAnchor = getAddDeviceAnchor();
+  if (addDeviceAnchor !== undefined) {
+    // Register this device (tentatively)
+    await registerTentativeDevice(addDeviceAnchor, connection);
+
+    // Show a good bye message
+    await new Promise((_) => {
```
peterpeterparker, line 107: not sure I get why the `new Promise` here, it's necessary? there is also no `resolve()` within, its ok?

src/frontend/src/index.ts
```diff
@@ -94,6 +97,28 @@ const init = async () => {
...
+      const container = document.getElementById("pageContent") as HTMLElement;
+      render(
+        html`<h1
+          style="position: absolute; max-width: 100%; top: 50%; transform: translate(0, -50%);"
+          data-role="notify-auth-success"
+        >
+          You may close this page.
```
peterpeterparker, line 114: i18n?

src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
nmattia: I think `"ok" | never` is pretty much the same as `"ok"` right?

src/frontend/src/index.ts
nmattia: We have other screens like this and I've just copied the code for now. I'll extract this in a component and extract the i18n text separately if OK for you.

src/frontend/src/index.ts
nmattia: That's a very good question; copy pasta probably. Will remove

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
nmattia: Right; I'll replace this with a QR code very soon anyway, but the idea is that you anyway need to open the link on _another_ device so there's no real point of having an `<a>` here. Does that make sense? Or do you see a use case for opening the link in the same browser?

src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
peterpeterparker: guess so, just makes it explicit for the reader as well I guess.

src/frontend/src/index.ts
peterpeterparker: Sure

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
peterpeterparker: I was more thinking on copy/pasting the link easily but without link, it's also more explicit. works for me as it, anyhow, not my call 😉

peterpeterparker: curiosity question it was

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts
nmattia: A very good one too; I did wonder about this too

src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
nmattia: I think it would make it slightly less explicit actually; `never` suggests the function never returns (i.e. there's an infinite loop or something). At least that's how I interpret it, but maybe that's just the Haskell way of looking at it 😅

PR #1395, opened 2023-04-05
Add missing assets to test

frederikrothenberger:
The test for checking the certification of existing assets was incomplete. This PR adds the missing assets to check.


frederikrothenberger:
> Thanks! Though made moot by the next change to serve the entire `dist/` right?

Not entirely, `/.well-known/ic-domains` is not part of dist, since it's not a file.

PR #1402, opened 2023-04-06
Dapps Explorer component

meodai:
Add a dapps teaser and a dapps list component.

CSS additions:

- `.c-marquee` and subsequent classes. To create a marquee kind of thumbnail (see video) that gives the illusion that it scroll indefinitely.
- `.t-title--discrete` class to be able to create these "over-title" kind of titles commonly used in the new designs.
- `.t-title--list` class to use within list items.
- `.c-card__teaser` is a teaser that goes to the edge of the card. (expect to the top)

CSS changes:
- `.c-card` now has local variables called `--bezel-x` and `--bezel-y` to be able to make the teaser go to the edge of a card. It also makes it easier to overwrite in `.c-card--narrow` and `.c-card--tight`
-  `.c-action-list` classes were slightly modified to be able to accommodate multiline text and the "dapp" icons.

Screenshot 2023-04-06 at 15 50 25

https://user-images.githubusercontent.com/608386/230398327-ea2b0b95-7931-404b-8005-067003c2c5d4.mov


meodai:
I had added the `dapps.json` so we could fetch it in the demo like so:

```ts
const getDappsMeta = async (): Promise<DappDescription[]> => {
  const dappsList = await fetch("/dapps.json").then((d) => d.json());
  return dappsList;
};
```

but after a discussion with @nmattia I will drop that part.

PR #1387, opened 2023-04-04
Node LTS

peterpeterparker:

PR #1405, opened 2023-04-06
Use 127.0.0.1 in webpack configs

nmattia:
The recent move to node 18 means that 127.0.0.1 should be preferred over `localhost` for local development. Otherwise webpack cannot proxy to the canister.

c.f.: https://github.com/dfinity/internet-identity/pull/1387


peterpeterparker (approved):
Thx

PR #1406, opened 2023-04-06
Extract index message and copy

nmattia:
This extracts a `showMessage` component/function from the index and extracts the copy too.

The new `showMessage` is also reused in other places that need to display a centered message, and the `data-role` is fixed (in index.ts, the attribute had been copy/pasted but was incorrectly suggesting some authorization was successful).

### DIFF
--- a/src/frontend/src/components/message.ts
+++ b/src/frontend/src/components/message.ts
@@ -1,0 +1,33 @@
+import { html, TemplateResult } from "lit-html";
+import { ifDefined } from "lit-html/directives/if-defined.js";
+import { renderPage } from "../utils/lit-html";
+
+const showMessageTemplate = ({
+  message,
+  role,
+}: {
+  message: string;
+  role?: string;
+}): TemplateResult => {
+  return html`<h1
+    style="position: absolute; max-width: 100%; top: 50%; transform: translate(0, -50%);"
+    data-role=${ifDefined(role)}
+  >
+    ${message}
+  </h1>`;
+};
+
+export const showMessagePage = renderPage(showMessageTemplate);
+
+// Show a simple, centered message (with optional "data-role" attribute)
+export const showMessage = ({
+  message,
+  role,
+}: {
+  message: string;
+  role?: string;
+}): void =>
+  showMessagePage({
+    message,
+    role,
+  });
--- a/src/frontend/src/index.ts
+++ b/src/frontend/src/index.ts
@@ -1,17 +1,20 @@
-import { html, render } from "lit-html";
 import { showWarningIfNecessary } from "./banner";
 import { displayError } from "./components/displayError";
+import { showMessage } from "./components/message";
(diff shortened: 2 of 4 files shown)


src/frontend/src/components/message.ts
```diff
@@ -0,0 +1,33 @@
+import { html, TemplateResult } from "lit-html";
+import { ifDefined } from "lit-html/directives/if-defined.js";
```
peterpeterparker, line 2: the explicit `.js` is intended?

src/frontend/src/components/message.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const showMessageTemplate = ({
+  message,
+  role,
+}: {
+  message: string;
+  role?: string;
+}): TemplateResult => {
+  return html`<h1
+    style="position: absolute; max-width: 100%; top: 50%; transform: translate(0, -50%);"
```
peterpeterparker, line 13: not for this PR: shouldn't it be a `class`?

src/frontend/src/components/message.ts
```diff
@@ -0,0 +1,33 @@
...
+import { renderPage } from "../utils/lit-html";
+
+const showMessageTemplate = ({
+  message,
+  role,
+}: {
+  message: string;
+  role?: string;
+}): TemplateResult => {
+  return html`<h1
```
peterpeterparker, line 12: not for this PR: should it be a h1 ?

src/frontend/src/index.ts
```diff
@@ -103,17 +106,14 @@ const init = async () => {
...
-        style="position: absolute; max-width: 100%; top: 50%; transform: translate(0, -50%);"
-        data-role="notify-auth-success"
-      >
-        You may close this page.
-      </h1>`,
-      container
-    );
+    showMessage({
+      message: staticCopy.close_page_device_added,
+      role: "notify-device-added",
```
peterpeterparker, line 115: renamed to `notify-device-added` on purpose right?

peterpeterparker (approved):
LGTM, thx

Two questions related to the PR, two just because I noticed these while reading the code.

src/frontend/src/components/message.ts
nmattia: yeah I think that'd make sense. I was just moving code around though but happy to do this in this PR if you prefer

src/frontend/src/index.ts
nmattia: yep! thanks for paying attention though

src/frontend/src/components/message.ts
nmattia: Good question; probably not. Added to todo list alongside your comment about making the css attributes into a class

src/frontend/src/components/message.ts
peterpeterparker: good to know

PR #1408, opened 2023-04-11
Include all files in dist folder as assets

frederikrothenberger:
This PR changes the code to dynamically pick up all assets in the dist folder.

Some properties of the files in `dist` are assumed:
* `.` is only used in known file endings (i.e. `foo.bar.html` is not allowed)
* Only known file types are supported
* All html files require `fixup_html`
* `js.gz` files should be served as `.js`
* `**/index.html` files should be served as `**/`
* `**/<file>.html` files should be served as `**/<file>`


src/internet_identity/src/assets.rs
```diff
@@ -55,21 +58,19 @@ lazy_static! {
...
-    };
-
-    static ref ABOUT_HTML_STR: String = {
-        fixup_html(include_str!("../../../dist/about.html"))
-    };
+    /// Map of path aliases
+    /// This creates a copy of the asset at the mapped path with the same content as the asset at the
+    /// original path.
+    static ref PATH_ALIASES: HashMap<String, String> = vec![
+        ("/index.html".to_string(), "/".to_string()),
```
nmattia, line 65: idea: how about mapping `**/index.html` to `**/` and `**/<foo>.html` to `**/<foo>` in general? I think that's what most HTTP servers do (module some details like how they handle `/foo` vs `/foo/`).

src/internet_identity/src/assets.rs
```diff
@@ -86,61 +87,95 @@ pub fn init_assets() {
...
+                ContentEncoding::Identity,
+                ContentType::PNG,
+            ),
+            "webp" => (
+                file_path,
+                file_bytes,
+                ContentEncoding::Identity,
+                ContentType::WEBP,
+            ),
+            _ => panic!("Unknown asset type: {}", asset.path().display()),
```
meodai, line 124: I think we also need `webp` (most of the dapp icons) and `svg`

frederikrothenberger: `webp` is already supported. I'm happy to add support for more file types, but I would prefer to do this in a separate PR.

src/internet_identity/src/assets.rs
frederikrothenberger: Good idea 👍 
Done.

src/internet_identity/src/assets.rs
```diff
@@ -86,61 +78,104 @@ pub fn init_assets() {
...
+    let mut assets = collect_assets_from_dir(dir);
+    for subdir in dir.dirs() {
+        assets.extend(collect_assets_recursive(subdir).into_iter());
+    }
+    assets
+}
+
+fn collect_assets_from_dir(dir: &Dir) -> Vec<(String, Vec<u8>, ContentEncoding, ContentType)> {
+    let mut assets: Vec<(String, Vec<u8>, ContentEncoding, ContentType)> = vec![];
+    for asset in dir.files() {
```
nmattia, line 109: do you know how this behaves with symlinks? I guess we'll want to symlink `about.html` to `index.html`

webpack.config.ts
```diff
@@ -165,9 +166,13 @@ const defaults = {
...
           from: path.join(__dirname, "src", "frontend", "assets"),
           to: path.join(__dirname, "dist"),
-          // We want html files from HtmlWebpackPlugin, not the original ones
           filter: (resourcePath) => {
-            return !resourcePath.endsWith(".html");
+            return (
+              // we want html files from HtmlWebpackPlugin, not the original ones
+              !resourcePath.endsWith(".html") &&
+              // exclude dotfiles
+              !resourcePath.split("/").pop()?.startsWith(".")
```
nmattia, line 174: when would there be dotfiles in the dist?

nmattia (commented):
Looking nice! Just a couple questions.

@peterpeterparker do the serving rules for assets look good to you? I remember you had some issues with the way the asset canister served assets at some point.

src/internet_identity/src/assets.rs
```diff
@@ -86,61 +78,104 @@ pub fn init_assets() {
...
+        .unwrap()
+        .split_once('.')
+        .unwrap()
+        .1
+}
+
+/// Returns the asset path for a given file:
+/// * make relative path absolute
+/// * map **/index.html to **/
+/// * map **/<foo>.html to **/foo
```
peterpeterparker, line 152: @nmattia should we support trailing slash strategy as well or just don't forget that we don't?

browser url `/foo` -> asset `/foo/index.html`

nmattia: Up to you and @frederikrothenberger really!

frederikrothenberger: @peterpeterparker: Could you clarify: Does your comment mean that rather than serving `about.html` on `/about`, we should instead serve `about/index.html`?

I.e. we move the emitted `about.html`?

peterpeterparker: No no nothing to be changed. With my comment I just meant that it works for now but if we ever change the build process to a trailing slash strategy (`./dist/foo/index.html` instead of `./dist/foo.html/), it might no work. Therefore asking if we already want to support this or just remember that if we ever do so, we have to extend the mapping.

frederikrothenberger: Ah, ok. In that case, I would leave it as is. Once we change, it will break in obvious ways, so even if we forget, we should not loose too much time. And then we can change the mapping.

peterpeterparker: Works for me

webpack.config.ts
frederikrothenberger: > Don't .DS_Store files show up on git status anyway?

No, we have them explicitly ignored in our `.gitignore`. I think this behavior is consistent with that.

src/internet_identity/src/assets.rs
frederikrothenberger: Unfortunately this causes the `include_dir!` macro to panic. We would first need to add support in the library for that.

nmattia: ah, that's a shame. Maybe we can deduplicate the assets afterwards (and store them in a CAS) if necessary, but we'll worry about this only if the wasm module gets too big.

nmattia (approved):

src/internet_identity/tests/integration/http.rs
```diff
@@ -17,7 +17,6 @@ fn ii_canister_serves_http_assets() -> Result<(), CallError> {
     let assets: Vec<(&str, Option<&str>)> = vec![
         ("/", None),
         ("/about", None),
-        ("/index.html", None),
         ("/index.js", Some("gzip")),
         ("/index.css", None),
         ("/loader.webp", None),
```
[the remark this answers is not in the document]
frederikrothenberger: I think we should have some test that verifies that assets are served and also certified correctly. This is the only test that would fail if we were to screw up asset certification. So what about we don't make this a full list of assets but rather a smaller selection of assets as a sanity check.
I.e.:
* `/` because we want to have some asset there anyway
* `/index.js` because it is gzipped
* `/.well-known/ic-domains` because it is the only asset that does _not_ have a file in `dist` and it would break the new domain

WDYT?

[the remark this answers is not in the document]
nmattia: SGTM! Good thinking.

PR #1410, opened 2023-04-11
dapps explorer component 

meodai:
Add a dapps teaser and a dapps list component. (Cleaned up version of https://github.com/dfinity/internet-identity/pull/1402)

CSS additions:

- `.c-marquee` and subsequent classes. To create a marquee kind of thumbnail (see video) that gives the illusion that it scroll indefinitely.
- `.t-title--discrete` class to be able to create these "over-title" kind of titles commonly used in the new designs.
- `.t-title--list` class to use within list items.
- `.c-card__teaser` is a teaser that goes to the edge of the card. (expect to the top)

CSS changes:
- `.c-card` now has local variables called `--bezel-x` and `--bezel-y` to be able to make the teaser go to the edge of a card. It also makes it easier to overwrite in `.c-card--narrow` and `.c-card--tight`
-  `.c-action-list` classes were slightly modified to be able to accommodate multiline text and the "dapp" icons.

Screenshot 2023-04-06 at 15 50 25

https://user-images.githubusercontent.com/608386/230398327-ea2b0b95-7931-404b-8005-067003c2c5d4.mov


frederikrothenberger:
> @frederikrothenberger added you as a reviewer, because I was unsure what to do about the logos.

#1408 will automatically include the icons here under `/icons/<icon>`. If this is the path you would like to have these icons served then it's ok to add them like this.

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
+import { html, TemplateResult } from "lit-html";
+
+export type DappDescription = {
+  id: string;
+  name: string;
+  oneLiner?: string;
+  link: string;
+  derivationOrigin?: string;
```
peterpeterparker, line 8: why is the derivation origin needed here?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+  return html`<div
+    class="c-marquee__list"
+    style="--itemsPerRow: ${itemsPerRow}; --totalRows: ${totalRows}"
+  >
+    ${rows.map(
+      (row, i) => html`<div class="c-marquee__row" style="--rowEq: ${i}">
+        <div class="c-marquee__rowHalf">
+          ${row.map(
+            (dapp) => html`<div class="c-marquee__item">
+              <img
```
peterpeterparker, line 45: - lazy load the images
- should the role "presentation" be set?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+              />
+            </div>`
+          )}
+        </div>
+        <div class="c-marquee__rowHalf c-marquee__rowHalf--second">
+          ${row.map(
+            (dapp) => html`<div class="c-marquee__item">
+              <img
+                src="icons/${dapp.logo}"
+                alt="${dapp.name}"
```
peterpeterparker, line 58: should the alt be the name of the dapp or something like "The logo of the dapp name"?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+      </div>`
+    )}
+  </div>`;
+};
+
+export const dappsListElement = (
+  dappsList: DappDescription[]
+): TemplateResult => {
+  return html` <article class="c-card c-card--narrow">
+    <h2 class="t-title t-title--discrete">Dapps explorer</h2>
```
peterpeterparker, line 73: extract i18n keys?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+    <h2 class="t-title">Explore dapps</h2>
+    <section aria-label="List of dapps">
+      <div class="c-action-list">
+        ${dappsList.map(
+          (dapp) => html` <a
```
peterpeterparker, line 78: nth: the object can be decontructed for better readability

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+            (dapp) => html`<div class="c-marquee__item">
+              <img
+                src="icons/${dapp.logo}"
+                alt="${dapp.name}"
+                class="c-marquee__image"
+              />
+            </div>`
+          )}
+        </div>
+        <div class="c-marquee__rowHalf c-marquee__rowHalf--second">
```
peterpeterparker, line 53: isn't this the same code as few lines above? can a component be extracted for it?

src/frontend/src/components/dappList.ts
meodai: @peterpeterparker I worked with the data that come from @frederikrothenberger json. It sometimes has this key and sometimes not.

src/frontend/src/components/dappList.ts
meodai: I am not sure what you mean by that.

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+ * when the animation reaches the end of the list, it is reset to the beginning. (default keyframe animation behaviour)
+ */
+
+const createMarqueeList = (dapps: DappDescription[]): TemplateResult => {
+  const itemsPerRow = 5;
+  const totalRows = 4;
+
+  const rows = [];
+
+  for (let i = 0; i < totalRows; i++) {
```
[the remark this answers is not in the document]
meodai: I needed an array per row. Would you use something like `new Array(totalRows).fill([])` instead? I thought it was more readable this way.

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+  // rows are duplicated to create the infinite scrolling effect
+  return html`<div
+    class="c-marquee__list"
+    style="--itemsPerRow: ${itemsPerRow}; --totalRows: ${totalRows}"
+  >
+    ${rows.map(
+      (row, i) => html`<div class="c-marquee__row" style="--rowEq: ${i}">
+        <div class="c-marquee__rowHalf">
+          ${row.map(
+            (dapp) => html`<div class="c-marquee__item">
```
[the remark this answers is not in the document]
meodai: great Idea! Il do that

src/frontend/src/components/dappList.ts
meodai: yes! Probably, or even aria-hidden, since it does not have any semantical meaning.

src/frontend/src/components/dappList.ts
meodai: I don't think so. For a photo definitely, but this is a simple list of dapps. Since it only serves a decorative function, the alt should probably be empty if we will be using `aria-role` presentation on its parent

src/frontend/src/components/dappList.ts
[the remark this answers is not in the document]
meodai: I did that in https://github.com/dfinity/internet-identity/pull/1410/commits/03e0ed4bead2595cf6bbb13a592c73f2ed5d4b2b

src/frontend/src/components/dappList.ts
meodai: Did so in: https://github.com/dfinity/internet-identity/pull/1410/commits/665f3a189954236a755581fe85e53e21385ca8f9

src/frontend/src/components/dappList.ts
meodai: Did it in: https://github.com/dfinity/internet-identity/pull/1410/commits/838012b7af29b61b0883183aa10942125e692d50

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,107 @@
...
+      <div class="c-action-list">
+        ${dappsList.map(
+          (dapp) => html` <a
+            href="${dapp.link}"
+            class="c-action-list__item"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            <div class="c-action-list__icon" aria-hidden>
+              <img src="icons/${dapp.logo}" alt="${dapp.name} logo" />
```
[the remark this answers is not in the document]
meodai: @peterpeterparker the parent is aria-hidden

src/frontend/src/components/dappList.ts
meodai: @nmattia is taking care of that. Its only about the design

src/frontend/src/components/dappList.ts
nmattia: Actually this is the same template even, right? No need to recreate it, you can just reuse it (it's still gonna be two different HTML elements)

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,119 @@
...
+          </a>`;
+        })}
+      </div>
+    </section>
+  </article>`;
+};
+
+export const dappsTeaser = (
+  dappsList: DappDescription[],
+  clickFn: () => void
```
nmattia, line 102: what would this do? does this open any specific dapp or does it take you to another screen?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,119 @@
+import { html, TemplateResult } from "lit-html";
+
+export type DappDescription = {
+  id: string;
+  name: string;
+  oneLiner?: string;
+  link: string;
+  derivationOrigin?: string;
+  description: string;
```
nmattia, line 9: is `description` used?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,119 @@
...
+  logo: string;
```
nmattia, line 10: can we simply reuse the dapp name as path for the logo?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,119 @@
+import { html, TemplateResult } from "lit-html";
+
+export type DappDescription = {
+  id: string;
```
nmattia, line 4: is `id` used?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,119 @@
...
+        <div class="c-marquee__rowHalf c-marquee__rowHalf--second">
+          ${createMarqueeListRowItems(row)}
+        </div>
+      </div>`
+    )}
+  </div>`;
+};
+
+export const dappsListElement = (
+  dappsList: DappDescription[]
```
nmattia, line 70:
```suggestion
  dapps: DappDescription[]
```

wdyt?

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,119 @@
...
+  const rows = [];
+
+  for (let i = 0; i < totalRows; i++) {
+    const row = [];
+    for (let j = 0; j < itemsPerRow; j++) {
+      const dapp = dapps[(i * itemsPerRow + j) % dapps.length];
+      row.push(dapp);
+    }
+    rows.push([...row]);
+  }
```
nmattia, line 50: this seems a bit error prone. How about splitting the array into chunks instead? https://stackoverflow.com/a/8495740

src/frontend/src/components/dappList.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  // rows are duplicated to create the infinite scrolling effect
+  return html`<div
+    aria-hidden="true"
+    class="c-marquee__list"
+    style="--itemsPerRow: ${itemsPerRow}; --totalRows: ${totalRows}"
+  >
+    ${rows.map(
+      (row, i) => html`<div class="c-marquee__row" style="--rowEq: ${i}">
```
nmattia, line 59: what does `rowEq` mean?

meodai: @nomeata its usually used for CSS selectors. (I started to use it when jQuery did: `https://api.jquery.com/eq-selector/`) But probably index would be better. Il change it.

src/frontend/src/components/dappList.ts
meodai: it would show the list of dapps.

src/frontend/src/components/dappList.ts
meodai: @nmattia the initial idea was that @frederikrothenberger would add it separately. But that changed now, so I will include the JSON file.

src/frontend/src/components/dappList.ts
nmattia: Gotcha, do we need a screen for that too then? Or is that an external link?

src/frontend/src/components/dappList.ts
meodai: the second row needs `c-marquee__rowHalf--second` the content of the row is in its own function now.

src/frontend/src/components/dappList.ts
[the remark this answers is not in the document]
peterpeterparker: The new implementation looks better because immutable. Does it work out too if the number of dapps does not exactly feel the rows?

Like 4 apps for 2 rows

> dapp0 dapp1 dapp2
> dapp3 dapp0 dapp1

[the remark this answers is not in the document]
peterpeterparker: Would not be against a test ;)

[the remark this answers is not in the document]
meodai: @peterpeterparker yes I also added a comment that explains how it works. (thats what the modulo is doing)

src/frontend/src/components/dappList.ts
meodai: Ive updated my code, please have a look and tell me if you understand my implementation now if it works for you

src/frontend/src/components/dappList.ts
nmattia: I mean this:

```ts
const foo = marqueeListRow(row);
... 
<div class="c-marquee__rowHalf">foo</div>
<div class="c-marquee__rowHalf c-marquee__rowHalf--second">foo</div>
```

src/frontend/src/components/dappList.ts
nmattia: ok ok was just out of curiosity

src/frontend/src/components/dappList.ts
nmattia: (I'm not saying you should do this, just that you can)

meodai:
@nmattia I have implemented the changes discussed. Let me know if this is something you can take over or if you need me to change anything else

src/frontend/src/components/dappList.ts
meodai: I think its a good idea to keep it separated since the name could contain special characters.

nmattia:
Closing in favor of #1428

PR #1412, opened 2023-04-11
Show tooltip for recovery phrase

nmattia:
This adds a tooltip displayed when the green "recovery phrase" checkmark is hoverd. This makes the checkmark more consistent with the warning signs shown on missing recoveries.


meodai (approved):
LGTM!

PR #1414, opened 2023-04-11
Nuke unnecessary assets

nmattia:
Removes dummy test assets that aren't required anymore since we simply bundle the whole dist and don't test for separate assets.


frederikrothenberger (approved):
Thanks!

PR #1416, opened 2023-04-12
Extract copy in pollForTentativeDevice

nmattia:
This extracts the copy into a JSON file for use with I18n.

### DIFF
--- a/src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.json
+++ b/src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.json
@@ -1,0 +1,14 @@
+{
+  "en": {
+    "identity_anchor": "Identity Anchor",
+    "add_trusted_device": "Add Trusted Device",
+    "follow_steps_on_new_device": "Follow these steps on your new device",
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.json
```diff
@@ -0,0 +1,14 @@
+{
```
peterpeterparker, line 1: - You don't save all the JSON file in a same directory?
- Should this directory or filename be prefixed with the lang?

like `/src/i18n/en/...`?

peterpeterparker (approved):
Thx. A minor question.

src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.json
peterpeterparker: If you have all files in a same folder it is easier for contributors or translators to find which files to translate.

peterpeterparker: Sure language within the json sounds good 👍

PR #1417, opened 2023-04-12
Build with vitejs

peterpeterparker:
# Changes

- drop `webpack` for `vite`
  - remove webpack dependencies
  - add various vite dependencies
  - add configuration `vite.config.ts`
  - convert webpack plugins - e.g. for canister IDs or to remove `script` or prerendering - to `vite.plugin.ts`
  - move and adapt `html` file for `index.html`
  - move `showcase` html and ts file to separate directory
-  set `buffer`, `process` and `stream-browserify` as dependencies instead of dev dependencies because these are such
- polyfill `buffer` manually in `index.ts`
- set script to be loaded as `module` by canister

Remaining non-production blocker tasks in https://github.com/dfinity/internet-identity/issues/1458.


vite.plugins.ts
```diff
@@ -0,0 +1,94 @@
...
+      internet_identity: { local: canisterId },
+    } = JSON.parse(readFileSync(canisterIdsJsonFile, "utf-8"));
+
+    const assertNonNullish: (
+      value: string
+    ) => asserts value is NonNullable<string> = (value: string): void => {
+      if (value === null || value === undefined) {
+        throw new Error("Internet identity canister ID undefined");
+      }
+    };
```
nmattia, line 24: I think config files really benefit from being concise, even if they're not super robust (i.e. it's ok if they break on bad input). Would it make sense to use something simpler and keep the config short? Any problem with the original code?

vite.plugins.ts
```diff
@@ -0,0 +1,94 @@
...
+    );
+  },
+});
+
+/**
+ * Pre-render about.html
+ *
+ * 1. Plugin transform only /about.html
+ * 2. When such file is detected, it uses lit SSR render function to render the TypeScript file to a string
+ * 3. It finally replaces the JavaScript content in the HTML content as if it would be loaded at runtime with the static content that was evaluated
```
nmattia, line 74:
```suggestion
 * 1. Plugin transforms only /about.html
 * 2. When that file is detected, the lit SSR render function is used to render the TypeScript file to a string
 * 3. The rendered HTML is then injected into the file, as it would have been loaded at runtime
```

Tried to clarify

vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+        $generated: resolve(__dirname, "src/frontend/generated"),
+      },
+    },
+    build: {
+      outDir: "../../dist",
+      emptyOutDir: true,
+      rollupOptions: {
+        input: {
+          main: resolve(__dirname, "src/frontend/index.html"),
+          about: resolve(__dirname, "src/frontend/about.html"),
```
nmattia, line 38: should this be set based on `mode`? you seem to differentiate in the rest of the config

src/frontend/src/flows/about/index.ts
```diff
@@ -97,7 +97,7 @@ export const aboutView = (): void => {
   document.title = "About | Internet Identity";
   const container = document.getElementById("pageContent") as HTMLElement;
   const i18n = new I18n();
-  if (process.env.HYDRATE_STATIC_PAGES !== "0") {
+  if (process.env.II_HYDRATE_STATIC_PAGES !== "0") {
```
nmattia, line 100: out of curiosity: why the rename? Other options are actual II (runtime) option, hence why they're prefixed with `II_`. This is a build option that could do without the prefix (for clarity)

vite.plugins.ts
```diff
@@ -0,0 +1,94 @@
...
+    return html.replace(
+      `<script id="setupJs"></script>`,
+      `<script data-canister-id="${readCanisterId()}" id="setupJs"></script>`
+    );
+  },
+});
+
+/**
+ * Remove the <script /> that loads the index.js for production build.
+ * A script loader is injected by the backend.
```
nmattia, line 54: Ah that's a shame, is there no vite option for not injecting the js?

nmattia (commented):
Thanks, this is great! Crazy how fast the build is now. 

Couple of general questions:

* Will you also take care of the [demo](https://github.com/dfinity/internet-identity/tree/main/demos) projects? Can be done separately but I don't want to mix build tools
* The [`HACKING.md`](https://github.com/dfinity/internet-identity/blob/main/HACKING.md#building-the-frontend) document still refers to webpack, is that accurate?
* Do we still [only bundle the english bip39 words](https://github.com/dfinity/internet-identity/blob/23870005b15f5960fbb16494f8a1e0fdcc2cc945/webpack.config.ts#L161)?
* Are the [webpack chunk name hints](https://github.com/dfinity/internet-identity/blob/23870005b15f5960fbb16494f8a1e0fdcc2cc945/src/frontend/src/flows/addDevice/manage/pollForTentativeDevice.ts#L178) still relevant?
* The main chunk (`index.js.gz`) doubles in size (was 177K gzipped with webpack, is 356K gzipped with vite). Do you know why?

package.json
```diff
@@ -5,22 +5,23 @@
   "description": "",
   "keywords": [],
   "scripts": {
+    "dev": "vite",
+    "start": "npm run dev",
```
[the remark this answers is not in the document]
peterpeterparker: Vite, SvelteKit, David use `npm run dev`.

Kept for `start` for backwards compatibility. Can remove `start` if you want?

src/internet_identity/src/assets.rs
```diff
@@ -31,7 +31,7 @@ pub enum ContentType {
 }
 
 // The <script> tag that loads the 'index.js'
-const JS_SETUP_SCRIPT: &str = "let s = document.createElement('script');s.async = true;s.src = 'index.js';document.head.appendChild(s);";
+const JS_SETUP_SCRIPT: &str = "let s = document.createElement('script');s.type = 'module';s.src = 'index.js';document.head.appendChild(s);";
```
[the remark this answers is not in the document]
peterpeterparker: We target ES modules so it seems to me more accurate to load these as `modules`.

https://gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6#typemodule-vs-non-module-typetextjavascript-vs-script-nomodule

vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+        },
+        output: {
+          entryFileNames: `[name].js`,
+          // II canister only supports resources that contains a single dot in their filenames. qr-creator.js.gz = ok. qr-creator.min.js.gz not ok. qr-creator.es6.min.js.gz no ok.
```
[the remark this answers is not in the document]
peterpeterparker: Not an absolutely must I guess it would be a bit more flexible while it is not a big issue to generate only what's needed.

vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+    ...loadEnv(mode ?? "development", process.cwd(), envPrefix),
+    ...process.env,
+  };
+
+  return {
+    envDir: "../../",
+    publicDir: "assets",
+    envPrefix,
+    resolve: {
+      // TODO: use aliases for imports
```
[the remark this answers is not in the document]
peterpeterparker: Yeah for another PR.

```
import {something} from "@lib/something";
```

vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+      alias: {
+        $assets: resolve(__dirname, "src/frontend/assets"),
+        $app: resolve(__dirname, "src/frontend/src"),
+        $generated: resolve(__dirname, "src/frontend/generated"),
```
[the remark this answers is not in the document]
peterpeterparker: Not yet, that's the TODO.

vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+    },
+    build: {
+      outDir: "../../dist",
```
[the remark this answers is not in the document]
peterpeterparker: Added a comment.

vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+    },
+    server: {
+      port: 8080,
+      proxy: {
+        "/api": "http://127.0.0.1:4943",
+      },
+    },
+    define: {
+      // TODO: replace process.env with import
```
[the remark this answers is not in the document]
peterpeterparker: Yeah, in another PR I'll get rid of the `process.env` in the code and use vite `import.meta.env` instead.

vite.plugins.ts
peterpeterparker: Once `@dfinity/utils` is added I can remove this piece of code and use the utils. It will become shorter that way.

src/frontend/src/flows/about/index.ts
peterpeterparker: > Only VITE_SOME_KEY will be exposed as import.meta.env.VITE_SOME_KEY to your client source code, but DB_PASSWORD will not.
>
> console.log(import.meta.env.VITE_SOME_KEY) // 123
> console.log(import.meta.env.DB_PASSWORD) // undefined

https://vitejs.dev/guide/env-and-mode.html

We use `II_` instead of `VITE_`. So when we will replace `process.env` it might become `undefined`.

vite.plugins.ts
peterpeterparker: Don't know. Reused the work I did few months ago. I guess I had a look at that time.

package.json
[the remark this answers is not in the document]
nmattia: Then let's not add `npm run dev` now, and we'll replace `npm run start` with `npm run dev` later

vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+import {
+  injectCanisterIdPlugin,
+  preRenderAboutPlugin,
+  stripInjectJsScript,
+} from "./vite.plugins";
+
+const defaultConfig = (mode?: string): Omit<UserConfig, "root"> => {
+  const envPrefix = "II_" as const;
+
+  // Expand environment - .env files - with canister IDs
```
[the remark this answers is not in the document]
nmattia: > Not really, I find quite easy to integrate these .env file. Also I really dislike command line argument so I'm quite happy to not have to care about it. It's also good for the community because sharing simplier command is easier.

Ok, this seems like a personal preference, so maybe we can stick to what we have now? In particular, options like `II_FETCH_ROOT_KEY` really need to be set correctly/have a safe default (of being disabled). In this case it looks like there's no actual default.

The files suggest the command is set, which is not the case (e.g. dev & test builds use different options). I don't mind having the infrastructure for reading those files, but then I would suggest not committing any files (you can add yours to `.git/info/exclude`)

> Canister IDs are injected by a plugin

But why is that mentioned here?

vite.config.ts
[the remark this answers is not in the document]
nmattia: Then please track this separately, not in the code

vite.config.ts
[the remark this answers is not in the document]
nmattia: Then please track this separately, not in the code

vite.plugins.ts
nmattia: Ok, worth using the original code until then? Or did the code not work with vite anymore?

vite.config.ts
peterpeterparker: `about.html` removed so this does not exist anymore

vite.plugins.ts
peterpeterparker: `@dfinity/utils` is now in so I was able to use its assertion utils.

package.json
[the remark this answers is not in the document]
peterpeterparker: Done

vite.config.ts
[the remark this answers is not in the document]
peterpeterparker: Not just a personaly preference, I assume any frontend developer in the world nowadays use `.env`.
Anyway removed then.

src/frontend/src/flows/about/index.ts
peterpeterparker: `.env` removed.

peterpeterparker:
@nmattia 

> Will you also take care of the demo projects?

Sure. In other PRs.

>  The HACKING.md document still refers to webpack, is that accurate?

No. Repheased to avoid mentioning "webpack".

> Do we still only bundle the english bip39 words?

I have added now the regex as `external` rollup. Works for you?

> Are the webpack chunk name hints still relevant?

No. Thought about removing these comments in another PR. That way this PR has no changes in code.

>  The main chunk (index.js.gz) doubles in size (was 177K gzipped with webpack, is 356K gzipped with vite). Do you know why?

Improved. 188kb now.

package.json
```diff
@@ -5,22 +5,23 @@
   "description": "",
   "keywords": [],
   "scripts": {
+    "start": "II_FETCH_ROOT_KEY=1 vite",
+    "host": "II_FETCH_ROOT_KEY=1 vite --host",
```
nmattia, line 9: leftover?

package.json
```diff
@@ -63,8 +53,11 @@
     "@dfinity/principal": "0.15.5",
     "@dfinity/utils": "^0.0.14",
     "bip39": "^3.0.4",
-    "lit-html": "^2.7.2",
+    "buffer": "^6.0.3",
+    "lit-html": "^2.3.1",
```
nmattia, line 57: Merge issue?

screenshots/desktop/dappsExplorer.png
nmattia, line 1: there seems to be an issue here, images are gone

src/showcase/.eslintrc.json
nmattia, line 1: is this for vite? why did all the showcase files have to move btw?

vite.config.ts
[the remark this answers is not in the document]
nmattia: CC @frederikrothenberger just FYI

vite.plugins.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+const readCanisterId = (): string => {
+  const canisterIdsJsonFile = "./.dfx/local/canister_ids.json";
+
+  try {
+    const {
+      internet_identity: { local: canisterId },
+    } = JSON.parse(readFileSync(canisterIdsJsonFile, "utf-8"));
+
+    assertNonNullish(canisterId);
```
nmattia, line 15: just checking, this throws right? so we'll get the following message?

> throw Error(`Could not get canister ID from ${canisterIdsJson}: ${e}`);

HACKING.md
nmattia, line 1: do you know why all list items changed from `*` to `-`?

package.json
peterpeterparker: No, I added it because it's the shorthand to expose the host on the lan. It reflects the update of the HACKING.md fie.

package.json
```diff
@@ -63,8 +53,11 @@
...
+    "process": "^0.11.10",
```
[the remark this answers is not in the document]
peterpeterparker: on purpose as decribed in the PR description.

process, buffer and stream-browserify needs to be polyfied because used at runtime.

src/showcase/.eslintrc.json
peterpeterparker: > is this for vite?

this what?

>  why did all the showcase files have to move btw?

vite needs one single entry, an `index.html`. did not really found how to have `index.html` and `showcase.html` and having it separate is cleaner anyway.

src/showcase/src/showcase.ts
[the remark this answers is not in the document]
peterpeterparker: `npm run format` 🤷‍♂️

vite.plugins.ts
peterpeterparker: I can add the additional label if you wish, done.

vite.config.ts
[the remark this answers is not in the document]
frederikrothenberger: Let me know if it becomes an issue. We can make the the file extension parsing smarter, I just avoided it because it was not necessary previously.

HACKING.md
peterpeterparker: prettier format

frederikrothenberger:
Maybe this is a stupid question, but doesn't `vite` imply `rollup`? And isn't `rollup` causing all the reproducibility issues that the NNS dapp is having?
Won't this be an issue for II too?

screenshots/desktop/dappsExplorer.png
peterpeterparker: thanks having a look

peterpeterparker:
First reproducibility issue did not work so I'm trying  now the option discussed in https://github.com/rollup/plugins/issues/1425#issuecomment-1465626736

peterpeterparker:
Reproducibility issue solved and tested @frederikrothenberger.

There was one issue discovered by @nmattia (💪). Rollup does not generate `process.env` code the same way if the value is defined or undefined. That lead to inconsistency at comparing the sha in the CI. Providing default value solves it - i.e. having `process.env....` used in code always defined solves it.

peterpeterparker:
Let's have a clean start. I created a patch from this PR and started a new clean one 👉 https://github.com/dfinity/internet-identity/pull/1461

PR #1419, opened 2023-04-13
Simplify modal component

nmattia:
This simplifies the modal component code by handling more of the logic through lit constructs.

### DIFF
--- a/src/frontend/src/components/modal.ts
+++ b/src/frontend/src/components/modal.ts
@@ -14,14 +14,25 @@ export const createModal = ({
 }) => {
   const modalElement: Ref<HTMLDialogElement> = createRef();
 
+  // Close modal using the browser API
   const closeModal = () =>
-    withRef(modalElement, (modalElement) => {
-      modalElement.close();
-      modalElement.remove();
-    });
+    withRef(modalElement, (modalElement) => modalElement.close());
 
+  // A container into which lit renders, which is removed from the DOM as soon as the dialog closes
+  const container = document.createElement("div");
+  const removeContainer = () => container.remove();
+
+  // dialog role="dialog": closes modal on Esc
+  // form method="dialog": closes modal on submit
+  // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
   const modalHtml = html`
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/modal.ts
```diff
@@ -7,78 +7,47 @@ import { withRef } from "../utils/lit-html";
...
-        <button
-          class="c-modal__close"
-          aria-label="Close Modal"
-          ${ref(closeButtonRef)}
-        >
-          &times;
-        </button>
-        ${slot}
-        <div class="c-modal__footer">
+    <dialog class="c-modal" aria-modal ${ref(modalElement)}>
```
nmattia, line 24: @peterpeterparker `aria-modal` is equivalent to `aria-modal="true"`, right?

peterpeterparker: Good question, not sure. The default value is `false`.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal

peterpeterparker (approved):
Rubber stamp

src/frontend/src/components/modal.ts
nmattia: > Boolean attributes are considered to be true if they're present on the element at all.
> https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute

all good

peterpeterparker: Great!

src/frontend/src/components/modal.ts
```diff
@@ -7,78 +7,47 @@ import { withRef } from "../utils/lit-html";
...
-  submitButtonRef: Ref<HTMLButtonElement>;
-}): HTMLDialogElement => {
+  submit: () => void;
+}) => {
+  const modalElement: Ref<HTMLDialogElement> = createRef();
+
+  const closeModal = () =>
+    withRef(modalElement, (modalElement) => {
+      modalElement.close();
+      modalElement.remove();
```
nmattia, line 20: this means that clicking the "X" button removes the element, but submitting doesn't. Will need to figure out what we want exactly.

PR #1420, opened 2023-04-13
Add modals without form

nmattia:
Before this, modals could not be created without forms. This allows creating modals that can be closed though do not necessarily contain forms.

The previous default behavior (a modal with form) is moved to `formModal`. I'm not sure we need it but leaving it because it's used in the style guide.

### DIFF
--- a/src/frontend/src/components/modal.ts
+++ b/src/frontend/src/components/modal.ts
@@ -17,7 +17,6 @@ export const createModal = ({ slot }: { slot: TemplateResult }) => {
   const removeContainer = () => container.remove();
 
   // dialog role="dialog": closes modal on Esc
-  // form method="dialog": closes modal on submit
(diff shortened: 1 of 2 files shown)


meodai:
I had made it a form because all the dialogs I could find where some kind of "make a decision". + MDN sais

> [<form>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) elements can close a <dialog> if they have the attribute method="dialog" or if the button used to submit the form has formmethod="dialog" set. In this case, the state of the form controls are saved, not submitted, the <dialog> closes, and the [returnValue](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/returnValue) property gets set to the value of the button that was used to save the form's state.

But in case its not needed, I think a div is completely fine.

meodai (approved):

PR #1421, opened 2023-04-13
Add Dapps Explorer without design

nmattia:
This introduces a modal view from the management page, which lists (and links to) some dapps that are known to use Internet Identity.

Couple of notes:

* For simplicity, the list of dapps (`dapps.json`) as well as the icons are vendored in this repo. They are derived from [dfinity/portal](https://github.com/dfinity/portal).
* The modal view is unstyled, in particular the scrolling doesn't work. The styling will be done separately by @meodai.
* The "Explore Dapps" button from the management page is temporary and will be replaced with a nice marquee by @meodai.
* The dapps explorer is added to the showcase


src/frontend/src/flows/manage/index.ts
```diff
@@ -132,6 +133,11 @@ const displayManageTemplate = ({
       </p>
     </hgroup>
     ${anchorSection(userNumber)}
+    <p class="t-paragraph">
```
meodai, line 136: in general you can use `l-stack? (on the button) a `p` should only be used for actual paragraphs. But since this will probably be replaced its definitely not blocking

meodai (approved):

src/frontend/src/flows/manage/index.ts
nmattia: I thought I'd make it as ugly as possible so that I make sure you look at everything carefully when you add the design 😄  


but good to know, thanks!

PR #1422, opened 2023-04-13
Remove trailing whitespace

nmattia:
A couple of files had trailing whitespaces. It triggers both my editor and my OCDs so I removed it.


frederikrothenberger (approved):
Thanks!

PR #1424, opened 2023-04-13
Don't use templates in attributes

nmattia:
Unfortunately `lit-html` does not support `TemplateResult`s for attributes (which I keep forgetting and which I don't know just yet how to check for/prevent). This updates some attributes to use `string`s instead.

before:

Screenshot 2023-04-13 at 15 42 52

after:

Screenshot 2023-04-13 at 15 43 16


peterpeterparker (approved):
LGTM, thx

PR #1426, opened 2023-04-13
Improve c-card CSS

meodai:
`.c-card` is quite essential to the functioning of the current visual design. Many containers are so called cards. It felt important to remove some clutter and prepare for upcoming changes.

- introduce `--bezel-x` and `--bezel-y` as local on `.c-card`. (Makes it nicer to read, to set to different sizes and different contexts and easier to reuse especially when something is placed off canvas, like it will be in #1410)
- add  `--rc-card` and `--rc-card-background` as reference tokens instead of consuming the background variable directly. (Was not clear when working on the design refresh)
- remove `.c-card--background::after` since it was still there from a previews design and not doing anything.


nmattia (approved):
LGTM, thanks!

PR #1427, opened 2023-04-14
Bring dapps explorer close button to front

nmattia:
If the button is behind the `hgroup`, the `h2` element in the `hgroup` prevents the button from being clicked. This moves the button to _after_ the `hgroup`, which doesn't matter visually since the button is positioned absolutely.


meodai (approved):
LGTM!

PR #1429, opened 2023-04-14
Remove static copy

nmattia:
Static versions of copy were used in places where a `TemplateResult` couldn't get inlined in an another `TemplateResult` (for instance, in HTML attribute values). On the other hand, `DirectiveResult`s can be inlined nicely, so this changes the `DynamicCopy` to produce `DirectiveResult`s instead of `TemplateResult`s.

A new helper type is introduced (`TemplateElement`) for all "elements" that can be inserted in templates (strings, etc).

### DIFF
--- a/src/frontend/src/flows/authorize/index.ts
+++ b/src/frontend/src/flows/authorize/index.ts
@@ -7,9 +7,9 @@ import {
 import { displayError } from "../../components/displayError";
 import { caretDownIcon, spinner } from "../../components/icons";
 import { showMessage } from "../../components/message";
-import { I18n } from "../../i18n";
+import { DynamicKey, I18n } from "../../i18n";
 import { Connection } from "../../utils/iiConnection";
-import { withRef } from "../../utils/lit-html";
+import { TemplateElement, withRef } from "../../utils/lit-html";
 import { unreachable } from "../../utils/utils";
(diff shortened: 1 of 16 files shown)


nmattia:
> A new helper type is introduced (TemplateElement) for all "elements" that can be inserted in templates (strings, etc).

@peterpeterparker I remember you asked for this recently in a review but I can't find the comment anymore.

src/frontend/src/flows/authorize/index.ts
```diff
@@ -38,7 +38,7 @@ export const authnTemplateAuthorize = ({
         })
       : undefined;
 
-  const wrap = (title: string | TemplateResult) => html`
+  const wrap = (title: DynamicKey) => html`
```
peterpeterparker, line 41: correct? it used to be a `string | TemplateResult` so I would have expected `TemplateElement`

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/authorize/index.ts
nmattia: here it doesn't matter since it's a `DynamicKey`, which is more restrictive anyway! if we need a `string` or `TemplateResult` in the future we can change for a `TemplateElement`. Good catch though :)

PR #1430, opened 2023-04-14
Remove non-gzipped wasms from build infrastructure

frederikrothenberger:
This PR removes the dual use of non-gzipped and gzipped wasms in favour of having gzipped wasms everywhere. This change only concerns the II canister and the archive but not the test or demo app (that use orthogonal build infrastructure anyway).


nmattia (approved):
Nice, LGTM!

PR #1431, opened 2023-04-14
Simplify toasts

nmattia:
This simplifies the toast component:

* All elements are created through lit templates instead of manually calling `createElement`
* The logic for looking up & reusing elements is delegated to lit (`render` does it automatically)
* The `repeat` directive is used with a singleton `toasts` object instead of manually inserting toast elements in the DOM
* The nested `data-toaster` and `data-toasts` divs are collapsed into a single one (page background can still be interacted with AFAICT, see video below but @meodai please double check in case I did something stoopid)
* The `irregularity` module is merged into the toasts module because it is currently only used there (we can always re-extract it if necessary, but for now having 2 modules just makes refactorings a bit more annoying)

https://user-images.githubusercontent.com/6930756/232058547-eeff284d-1716-4828-8f28-dd15dee93187.mov


src/frontend/src/components/toast.ts
```diff
@@ -1,60 +1,67 @@
...
-    document.body.appendChild(toasterEl);
-    return containerEl;
-  }
-  return toasterEl.querySelector("[data-toasts]");
-};
+// A toast element containing a message, and the (static) list of all toasts. The `Toast` element
+// must be an object because we (lit) use object-references in `repeat` to figure out which DOM elements
+// to replace.
+type Toast = { message: TemplateElement };
+const toasts: Toast[] = []; // the "toasts" singleton
```
nmattia, line 12: @peterpeterparker how bad is this in your eyes?

src/frontend/src/components/toast.ts
```diff
@@ -1,60 +1,67 @@
...
+  const removeToast = () => {
+    const toastIndex = toasts.indexOf(toast);
+    if (toastIndex > -1) {
+      toasts.splice(toastIndex, 1);
+    }
+    renderToasts();
+  };
+
+  return html` <div
+    class="c-toast ${asyncReplace(
```
meodai, line 48: I think we could use

```
role="alert" aria-live="assertive" 
```

So user using screenreaders would have the error read out loud to them the moment it is triggered.

nmattia: Added to the todo list! I'll do this separately, I've tried not to change anything in the generated HTML (except for squishing the divs to simplify)

src/frontend/src/components/toast.ts
peterpeterparker: Sounds good to me

peterpeterparker: it's kind of a store, should we call it `toastStore`? Just an idea, no strong opinion

nmattia: That's a good idea, let me do that!

src/frontend/src/components/toast.ts
```diff
@@ -1,60 +1,67 @@
...
  * A lot like console.error, but with a toast.
  * use like: toast.error("Something went wrong");
  */
 export const toast = {
-  error: (message: TemplateElement): void => {
-    const toasterEl = getOrCreateToaster();
-    const toastEl = createToast(message);
-    toasterEl?.appendChild(toastEl);
+  error: (message: Toast["message"]): void => {
+    toasts.push({ message });
```
peterpeterparker: In Stencil the change detection was not triggered by doing "push" so I am used to update state in following way:

```
toasts = [...toasts, {message}];
```

ditto in Svelte (example of NNS-dapp, `update` is the Svelte store callback):

```
update((messages: ToastMsg[]) => {
        return [...messages, { ...rest, id: toastId }];
      });
```

src/frontend/src/components/toast.ts
```diff
@@ -1,60 +1,67 @@
...
+    document.body
+  );
+};
+
+const toastTemplate = (toast: Toast): TemplateResult => {
+  const closing = new Chan<boolean>(false);
+  const closeToast = () => closing.send(true);
+
+  const removeToast = () => {
+    const toastIndex = toasts.indexOf(toast);
```
peterpeterparker: does that work if user closes the toast in the middle?

```
toast 1
toast 2 <--- user close this toast
toast 3
```

wouldn't it be safer to set id (`symbol`) to the toast and closes the particular one?

src/frontend/src/components/toast.ts
nmattia: gotcha! in this case though we trigger the push explicitly (by calling `renderToasts`) because we're not cool

src/frontend/src/components/toast.ts
nmattia: > does that work if user closes the toast in the middle?

Definitely, any particular reason you think it might not work in all cases?

> wouldn't it be safer to set id (symbol) to the toast and closes the particular one?

I think the current way is fine as long as the `Toast` is an object (see comment near the type) because then this will be compared by reference, which are just as unique as `Symobl`s (if I understand the question correctly)

src/frontend/src/components/toast.ts
peterpeterparker: good

src/frontend/src/components/toast.ts
nmattia: > The indexOf() method compares searchElement to elements of the array using [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality) (the same algorithm used by the === operator).
> (about ===) If both operands are objects, return true only if they refer to the same object.


https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

peterpeterparker: I've trust issue in comparing object, too much Java probably. That's why I generally rather like to test on a unique identifier.

Good then. Thanks for the feedback.

peterpeterparker (approved):
Sounds good, thx

PR #1433, opened 2023-04-14
Add missing aria-label

nmattia:
The label is a simple string because the management page has not been i18n-ified yet.


meodai (approved):
LGTM

PR #1437, opened 2023-04-18
Simplify captcha chans

nmattia:
This changes the Chan internals a bit to allow for a nicer interface, which in turns simplifies the captcha logic:

* The `Chan` now _always_ has a `latest`/default value
* A `Chan`'s `map()` can now return `Chan.unchanged` to signal no changes
* The captcha screen makes use of this new interface to simplify state handling

### DIFF
--- a/src/frontend/src/flows/register/captcha.ts
+++ b/src/frontend/src/flows/register/captcha.ts
@@ -39,27 +39,20 @@ export const promptCaptchaTemplate = <T>({
 }) => {
   const copy = i18n.i18n(copyJson);
 
-  // We define a few Chans that are used to update the page in a
-  // reactive way; see template returned by this function
-
-  // The image shown
-  const img = new Chan<TemplateResult>();
-
-  // The text input where the chars can be typed
-  const input: Ref<HTMLInputElement> = createRef();
-
-  // The error shown on bad input
-  const errorText = new Chan<DynamicKey | undefined>();
-  const hasError = errorText.map((e) => (e !== undefined ? "has-error" : ""));
-
-  // The "next" button behavior
-  const next = new Chan<((e: SubmitEvent) => void) | undefined>();
-  const nextDisabled = next.map((f) => f === undefined);
-  const nextCaption = new Chan<DynamicKey>();
-
-  // The "retry" button behavior
-  const retry = new Chan<(() => void) | undefined>();
-  const retryDisabled = retry.map((f) => f === undefined);
+  const spinnerImg: TemplateResult = html`
+    <div class="c-captcha-placeholder c-spinner-wrapper">
+      <div class="c-spinner">${spinner}</div>
+    </div>
+  `;
+  const captchaImg = (base64: string): TemplateResult =>
+    html`<div class="c-captcha-placeholder">
+      <img
+        src="data:image/png;base64,${base64}"
+        id="captchaImg"
+        class="c-image"
+        alt="captcha image"
+      />
+    </div>`;
 
   // The various states the component can inhabit
   type State =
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/register/captcha.ts
```diff
@@ -39,108 +39,101 @@ export const promptCaptchaTemplate = <T>({
...
 
   // The image shown
-  const img = new Chan<TemplateResult>();
+  const img: Chan<TemplateResult> = state.map({
+    f: (state) =>
+      state.status === "requesting"
+        ? spinnerImg
+        : state.status === "prompting"
+        ? captchaImg(state.challenge.png_base64)
+        : Chan.unchanged,
```
nmattia, line 75: @peterpeterparker in the first review of the captcha code you pointed out that we may sometimes forget to set the value for some components (for instance in the case of the img, although in that case we didn't update the value on purpose). "Not updating the value" in now done explicitly through `Chan.unchanged`.

src/frontend/src/flows/register/captcha.ts
```diff
@@ -39,108 +39,101 @@ export const promptCaptchaTemplate = <T>({
...
+      <div class="c-spinner">${spinner}</div>
+    </div>
+  `;
+  const captchaImg = (base64: string): TemplateResult =>
+    html`<div class="c-captcha-placeholder">
+      <img
+        src="data:image/png;base64,${base64}"
+        id="captchaImg"
+        class="c-image"
+        alt="captcha image"
```
peterpeterparker, line 53: not related but, I guess this accessbility label can be improved

src/frontend/src/flows/register/captcha.ts
```diff
@@ -39,108 +39,101 @@ export const promptCaptchaTemplate = <T>({
...
+        : undefined
+  );
 
-  // The "retry" button behavior
-  const retry = new Chan<(() => void) | undefined>();
-  const retryDisabled = retry.map((f) => f === undefined);
+  const nextDisabled: Chan<boolean> = next.map((f) => f === undefined);
+  const nextCaption: Chan<DynamicKey> = state.map(({ status }) =>
+    status === "requesting"
+      ? copy.generating
```
peterpeterparker, line 105: it's correct that "requesting" map to "generating"? just in case not expected

src/frontend/src/utils/utils.ts
```diff
@@ -229,14 +229,48 @@ export class Chan<A> implements AsyncIterable<A> {
...
+  // that the element shouldn't be changed, in which case a default (initial) value
+  // also needs to be provided.
+  map<B>(
+    opts: ((a: A) => B) | { f: (a: A) => B | typeof Chan.unchanged; def: B }
+  ): Chan<B> {
+    // How the mapped chan should handle the value
+    let handleValue: (opts: { send: (b: B) => void; value: A }) => void;
+    let latest: B;
+
+    if (typeof opts === "function") {
```
peterpeterparker, line 247: nth: would be nice to have this logic, then if / else / assign value in a function. 

Like:

```
const {handledValue, latest} = initHandledValue(...);
```

src/frontend/src/utils/utils.ts
```diff
@@ -229,14 +229,48 @@ export class Chan<A> implements AsyncIterable<A> {
...
+  static readonly unchanged = Symbol("unchanged");
+
   // Return a new Chan mapped with `f`.
-  map<B>(f: (a: A) => B): Chan<B> {
-    const latest = this.latest === undefined ? undefined : f(this.latest);
+  // In the simplest case, a mapping function is provided.
+  // For advanced cases, the mapping function may return 'Chan.unchanged' signalling
+  // that the element shouldn't be changed, in which case a default (initial) value
+  // also needs to be provided.
+  map<B>(
```
peterpeterparker, line 240: nth: function is tricky, I spent some minutes reading it to understand it => it's often a signal it would deserve a test

peterpeterparker (approved):
LGTM, minor comments

src/frontend/src/flows/register/captcha.ts
nmattia: yep, we're requesting a new captcha but the copy says "generating". Could make them both the same but don't want to have to update the state name every time the copy changes. Thanks for paying attention though!

src/frontend/src/utils/utils.ts
nmattia: ah yes, that'd be much nicer!

PR #1438, opened 2023-04-18
Don't lazy load images in teaser

nmattia:
When lazy loaded, teaser images start taking up space in the marquee but aren't shown yet, leading to them suddenly popping into existence.

Before:

https://user-images.githubusercontent.com/6930756/232816754-c88f5bee-ed46-40dc-ba28-8fe8dcf6ccf8.mov

After:

https://user-images.githubusercontent.com/6930756/232816797-41c77aa9-89a9-45cd-8bfe-1172ad84d269.mov

Thanks @peterpeterparker for the hint


src/frontend/src/flows/dappsExplorer/teaser.ts
```diff
@@ -56,7 +56,6 @@ const marqueeList = (): TemplateResult => {
             src=${logo}
             alt="${name}"
             class="c-marquee__image"
-            loading="lazy"
```
peterpeterparker, line 59: can you add a comment about it?

pretty sure that if I land on it in the future I would ask my self why it is not lazy loaded since it isn't my default approach.

peterpeterparker (approved):
LGTM, minor request

PR #1439, opened 2023-04-18
Fix dapps reference in chunkArray

nmattia:
The `chunkArray` helper was erroneously refering to the global `dapps` value, instead of its parameter `arr`. By chance, until now both were the same.


peterpeterparker (approved):
Thx

PR #1440, opened 2023-04-18
Don't reload devices when closing explorer

nmattia:
This ensures the management page doesn't reload when coming back from the explorer. Reloading (e.g. fetching devices) takes time and isn't very user-frienly. Since the user couldn't have e.g. modified devices from the explorer, we just re-display the page as we had originally displayed.

NOTE: the first commit makes the actual change, and the second one is only formatting.

Before:

https://user-images.githubusercontent.com/6930756/232828821-c7bee1de-8df3-4d3c-8f6e-bbd75bfd7546.mov

After:

https://user-images.githubusercontent.com/6930756/232828904-9d27f4f9-0874-4e87-b73d-04a035a8a76d.mov

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -223,55 +223,55 @@ export const displayManage = (
       );
     }
     const display = () =>
-    displayManagePage({
-      userNumber,
-      devices,
-      onAddDevice: async () => {
-        const nextAction = await chooseDeviceAddFlow();
-        switch (nextAction) {
-          case "canceled": {
-            resolve();
-            break;
+      displayManagePage({
+        userNumber,
+        devices,
+        onAddDevice: async () => {
+          const nextAction = await chooseDeviceAddFlow();
+          switch (nextAction) {
+            case "canceled": {
+              resolve();
+              break;
+            }
+            case "local": {
+              await addLocalDevice(userNumber, connection, devices_);
+              resolve();
+              break;
+            }
+            case "remote": {
+              await addRemoteDevice({ userNumber, connection });
+              resolve();
+              break;
+            }
+            default:
+              unreachable(nextAction);
+              resolve();
+              break;
           }
-          case "local": {
-            await addLocalDevice(userNumber, connection, devices_);
-            resolve();
-            break;
+        },
+        addRecoveryPhrase: async () => {
+          await setupPhrase(userNumber, connection);
+          resolve();
+        },
+        addRecoveryKey: async () => {
+          const confirmed = confirm(
+            "Add a Recovery Device\n\nUse a FIDO Security Key, like a YubiKey, as an additional recovery method."
+          );
+          if (!confirmed) {
+            // No resolve here because we don't need to reload the screen
+            return;
           }
-          case "remote": {
-            await addRemoteDevice({ userNumber, connection });
-            resolve();
-            break;
-          }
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/manage/index.ts
```diff
@@ -222,52 +222,58 @@ export const displayManage = (
...
-            break;
+    const display = () =>
+      displayManagePage({
+        userNumber,
+        devices,
+        onAddDevice: async () => {
+          const nextAction = await chooseDeviceAddFlow();
+          switch (nextAction) {
+            case "canceled": {
+              resolve();
```
peterpeterparker, line 233: You can avoid `resolve` duplication:

```
switch (nextAction) {
           case "cancel":
                 break;
            case "local": {
              await addLocalDevice(userNumber, connection, devices_);
              break;
            }
            case "remote": {
              await addRemoteDevice({ userNumber, connection });
              break;
            }
            default:
              unreachable(nextAction);
          }
          
          resolve();
     }
```

src/frontend/src/flows/manage/index.ts
```diff
@@ -222,52 +222,58 @@ export const displayManage = (
...
-      },
-    });
+          await setupKey({ connection });
+          resolve();
+        },
+        exploreDapps: async () => {
+          await dappsExplorer();
+          // We know that the user couldn't have changed anything (the user can't delete e.g. delete
+          // a device from the explorer), so we just re-display without reloading devices etc.
+          // the page without
```
peterpeterparker, line 271: it's maybe just me but I had to read this comment few times...

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/manage/index.ts
nmattia: I think this is just formatted code, see for commit for actual changes

src/frontend/src/flows/manage/index.ts
```diff
@@ -222,52 +222,58 @@ export const displayManage = (
...
+          display();
```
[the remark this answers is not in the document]
nmattia: The `lit-html` render function is also called `render`, any other suggestions?

src/frontend/src/flows/manage/index.ts
nmattia: Is there anything in particular that wasn't clear?

src/frontend/src/flows/manage/index.ts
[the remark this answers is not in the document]
peterpeterparker: `renderFlow` or whatever thing it renders but was just a suggestion, no worries.

src/frontend/src/flows/manage/index.ts
peterpeterparker: The ends `// the page without`, without what? My brain could not connect the dots at first but no worries.

src/frontend/src/flows/manage/index.ts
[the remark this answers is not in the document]
nmattia: `displayManage` and `renderManage` are already used 😬 I'll try to think of something

src/frontend/src/flows/manage/index.ts
peterpeterparker: Oki doki, might reformat it some day...or not ;)

PR #1441, opened 2023-04-18
Lazy load dapps list

nmattia:
This exports the dapps list (json file) as a new chunk which is lazy loaded, to avoid pulling the big dapps JSON list e.g. when authenticating to a 3rd party dapp.


src/frontend/src/flows/dappsExplorer/index.ts
```diff
@@ -58,7 +61,11 @@ const dappTemplate = ({
       rel="noopener noreferrer"
     >
       <div class="c-action-list__icon" aria-hidden="true">
-        <img src=${logo} alt=${name} loading="lazy" />
+        <img
+          src=${logo.replace("/img/showcase/", "/icons/")}
```
peterpeterparker, line 65: second time I notice `logo.replace("/img/showcase/", "/icons/"`, probably worth mapping the icons or a function

src/frontend/src/flows/dappsExplorer/index.ts
```diff
@@ -58,7 +61,11 @@ const dappTemplate = ({
...
+          alt=${name}
+          loading="lazy"
```
peterpeterparker, line 67: arent' these icons not lazy anymore?

src/frontend/src/flows/manage/index.ts
```diff
@@ -200,12 +205,14 @@ export const renderManage = async (
...
-export const displayManage = (
+export const displayManage = async (
   userNumber: bigint,
   connection: AuthenticatedConnection,
   devices_: DeviceData[]
-): Promise<void | AuthenticatedConnection> =>
-  new Promise((resolve) => {
+): Promise<void | AuthenticatedConnection> => {
+  // Fetch the dapps used in the teaser & explorer
+  const dapps = await getDapps();
```
peterpeterparker, line 214: shouldn't the dapps list be loaded within `dappsExplorer` or any reason to load these that way?

src/frontend/src/flows/dappsExplorer/dapps.ts
```diff
@@ -12,8 +14,20 @@ type ElementOf<Arr> = Arr extends readonly (infer ElementOf)[]
 export type DappDescription = ElementOf<typeof dappsJson>;
 
 // The list of dapps we showcase
-export const dapps: DappDescription[] = dappsJson.map((dapp) => ({
-  ...dapp,
-  /* fix up logo path (inherited from dfinity/portal) to match our assets */
-  logo: dapp.logo.replace("/img/showcase/", "/icons/"),
-}));
+export const getDapps = async (): Promise<DappDescription[]> => {
+  const dapps: typeof dappsJson | null = (
```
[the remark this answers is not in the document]
nmattia: I actually have no idea, I decided to play it safe based on https://github.com/dfinity/gix-components/blob/14cf17fa63d59af95588b372dff4071abaf31568/src/lib/components/QRCode.svelte#L46-L63 (when I looked at the code it was `null` though now seems to be `undefined`)

src/frontend/src/flows/dappsExplorer/index.ts
nmattia: One of these needs to go, that's my bad. Will remove after this PR is merged (got another one ready)

src/frontend/src/flows/dappsExplorer/index.ts
nmattia: I thought the same, but actually not. This is the explorer (i.e. a new page) whereas the icons you're referring to are the ones in the "teaser", i.e. the marquee/button on the management page.

src/frontend/src/flows/manage/index.ts
nmattia: They're shown in the teaser (the marquee) on this page (which is a button that takes you to the explorer) but good catch

src/frontend/src/flows/dappsExplorer/dapps.ts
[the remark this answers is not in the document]
peterpeterparker: In gix-cmp I did not added any try/catch because the build fails if the path is incorrect but, that's a good point I can add a try/catch there as well for extra safety.

peterpeterparker (approved):
LGTM, thx

PR #1443, opened 2023-04-19
Add DeviceData Implementation

frederikrothenberger:
This PR adds an implementation to create a test device data and easily get the principal from DeviceData.

In addition tests now only set the fields they care about. This makes changes to DeviceData (like additional fields) much easier.


src/internet_identity_interface/src/internet_identity/device_data.rs
```diff
@@ -0,0 +1,22 @@
...
+        Principal::self_authenticating(&self.pubkey)
+    }
+
+    /// Creates a new device that can be used for test purposes.
+    pub fn test_device() -> Self {
+        Self {
+            pubkey: ByteBuf::from([0; 32]),
+            alias: "test device".to_string(),
+            credential_id: None,
+            purpose: Purpose::Authentication,
```
przydatek, line 16: nit: maybe add something about `auth` to the helper name, say `test_device_auth` or `auth_test_device`?
(The default of `Purpose::Authentication` is somewhat arbitrary, and it is in fact the second value of the `Purpose`-enum)

przydatek (approved):
LGTM, thanks!

src/internet_identity_interface/src/internet_identity/device_data.rs
frederikrothenberger: Makes sense 👍

PR #1447, opened 2023-04-19
Add device metadata

frederikrothenberger:
This PR adds a flexible a flexible meta data map to each device. It can be used to more flexibly store data within the II canister. Currently the meta_data map supports String, Blob and nested maps.

The top-level metadata key is not private (as changed keys are recorded in the archive). However, all values (and nested maps) require authentication to access.


src/internet_identity/internet_identity.did
```diff
@@ -74,6 +74,7 @@ type DeviceData = record {
     key_type: KeyType;
     protection: DeviceProtection;
     origin: opt text;
+    meta_data: opt MetaDataMap;
```
nmattia, line 77: I've actually never seen "metadata" spelled in two words. Would a single-word "metadata" work for you?

src/internet_identity/src/main.rs
```diff
@@ -496,6 +496,7 @@ mod test {
     #[test]
     fn check_candid_interface_compatibility() {
         let canister_interface = __export_service();
+        println!("{}", canister_interface);
```
nmattia, line 499: leftover?

nmattia (commented):
Thanks! Didn't look at the implementation, only the interface which LGTM (modulo comment).

The candid type is a bit awkward so I guess we'll wrap it in the frontend.

Quick question: if we move the aliases to the metadata, would now be a good time for the migration?

src/internet_identity/internet_identity.did
frederikrothenberger: Yes, I'll change it.

src/internet_identity/src/main.rs
frederikrothenberger: Thanks, removed.

frederikrothenberger:
> Didn't look at the implementation

Ok, I'll ask @przydatek to do the implementation review then.

> if we move the aliases to the metadata, would now be a good time for the migration

You are talking about the external interface, right?
This would be another breaking change. We could expose it in `metadata` too (and always prefer the value from `metadata`, if present), officially deprecate `alias` and then after some time, remove it entirely...

Or what did you have in mind?

src/frontend/generated/internet_identity_types.d.ts
```diff
@@ -146,6 +148,14 @@ export type KeyType = { 'platform' : null } |
   { 'seed_phrase' : null } |
   { 'cross_platform' : null } |
   { 'unknown' : null };
+export type MetadataMap = Array<
+  [
+    string,
+    { 'map' : MetadataMap } |
+      { 'string' : string } |
+      { 'bytes' : Uint8Array | number[] },
```
przydatek, line 156: nit: the indentation here looks odd.

przydatek (commented):
Hi @frederikrothenberger, please see some questions below.

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -24,6 +25,13 @@ pub struct DeviceData {
     pub key_type: KeyType,
     pub protection: DeviceProtection,
     pub origin: Option<String>,
+    // Metadata map for additional device information.
+    //
+    // Note: some fields above will be moved to the metadata map in the future.
+    // If the metadata map already contains a field with the same name, the explicit
+    // field will be dropped and the data will be lost!
```
przydatek, line 32: Not sure I understand this comment: what is meant by "a field with the same name", and that the explicit field will be dropped? If I insert to metadata key-value pair ("alias", some_entry), then the value of the field `DeviceData.alias` will be "dropped"?   What does "dropped" mean in this case?

src/internet_identity/tests/integration/archive_integration.rs
```diff
@@ -333,6 +336,127 @@ mod pull_entries_tests {
...
+
+        // the archive polls for entries once per second
+        env.advance_time(Duration::from_secs(2));
+        // execute the timer
+        env.tick();
+
+        let entries = archive_api::get_entries(&env, archive_canister, None, None)?;
+        assert_eq!(entries.entries.len(), 1);
+
+        let register_entry = Entry {
```
przydatek, line 373: nit: maybe rename to `expected_register_entry`

przydatek (approved):
LGTM, but I believe one comment needs clarification (cf. below).  Thanks!

src/internet_identity_interface/src/internet_identity/types.rs
frederikrothenberger: Actually, let's not just have a comment but enforce this using a canister side check. This way we'll be sure that no data is lost in an upcoming migration.

przydatek: Now the intention is clear, and I do agree that the enforcing is the right way to go. Thanks!

przydatek:
Still LGTM, thanks!

PR #1448, opened 2023-04-19
Use utils for nullish comparison

peterpeterparker:
# Motivation

Use ic-js utils for handy comparison and transformation beginning with `isNullish` that checks undefined and null.

# Changes

- `npm i @dfinity/utils`
- replace `=== undefined` with `isNullish`


package-lock.json
nmattia, line 1: We're talking about those 4 lines right? Can we simply copy them instead?

``` ts
/** Is null or undefined */
export const isNullish = <T>(
  argument: T | undefined | null
): argument is undefined | null => argument === null || argument === undefined;
```

src/frontend/src/utils/iiConnection.ts
```diff
@@ -393,7 +394,8 @@ export class Connection {
     const actor = await this.createActor();
     return await actor.add_tentative_device(userNumber, {
       ...device,
-      origin: window?.origin === undefined ? [] : [window.origin],
+      // TODO: use toNullable
```
nmattia, line 397: leftover?

nmattia (commented):
Can you explain the benefits of using `isNullish` here? 

It looks like in most cases this just replaces `=== undefined`, maybe saving some characters but at the cost of an extra dependency and extra magic, which will be surprising for anyone discovering the codebase. In general I'd rather keep things explicit if possible.

peterpeterparker:
> It looks like in most cases this just replaces === undefined, maybe saving some characters but at the cost of an extra dependency and extra magic, which will be surprising for anyone discovering the codebase. In general I'd rather keep things explicit if possible.

- it unifies the way we develop II and NNS-dapp for which and in which we rely on `@dfinity/utils`
- `isNullish` is the first of the utilities we can use but I plan to use other utility such as `toNullable` and assertion (see documentation https://github.com/dfinity/ic-js/tree/main/packages/utils)
- from my experience, these are really convenient helpers once you begin to use these

src/frontend/src/utils/iiConnection.ts
peterpeterparker: No todo for a separate PR.

package-lock.json
peterpeterparker: I rather not.

1. There are other utils that are interesting
2. These utils are tested in the library, so if we move the function we also have to move the test and maintain both in two places

src/frontend/src/utils/iiConnection.ts
nmattia: Then please track this separately, not in the code

package.json
```diff
@@ -62,10 +62,11 @@
     "@dfinity/candid": "0.15.5",
     "@dfinity/identity": "0.15.5",
     "@dfinity/principal": "0.15.5",
+    "@dfinity/utils": "^0.0.14",
     "bip39": "^3.0.4",
     "lit-html": "^2.3.1",
-    "text-encoding": "^0.7.0",
-    "qr-creator": "^1.0.0"
+    "qr-creator": "^1.0.0",
+    "text-encoding": "^0.7.0"
```
nmattia, line 69: what's happened here?

nmattia:
Out of curiosity: why call it "nullish"? Wouldn't ["nullable"](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype) make more sense?

package.json
peterpeterparker: `npm i @dfinity/utils` outcome. I guess my npm (v9.6.1) sorted the members alphabetically.

peterpeterparker:
> Out of curiosity: why call it "nullish"? Wouldn't ["nullable"](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype) make more sense?

Don't remember exactly. I think we quickly braistormed some naming idea with @mstrasinskis and we picked that one. 
You can probably found the comments in an old PR of NNS-dapp.

mstrasinskis:
> Out of curiosity: why call it "nullish"? Wouldn't ["nullable"](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype) make more sense?

@nmattia  as I remember `nullable` was already used in utils for response transformation `from/toNullable` etc and the idea came from  `??` operator - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

PR #1450, opened 2023-04-20
Switch to non-extractable session keys

frederikrothenberger:
This PR fixes a long standing security issue with regards to II front-end delegation: The private keys have been held in memory, vulnerable to XSS attacks. Now, the session private keys are being managed through the SubtleCrypto API which does not allow to directly access the private key.

### DIFF
--- a/src/frontend/src/flows/register/captcha.ts
+++ b/src/frontend/src/flows/register/captcha.ts
@@ -1,4 +1,4 @@
-import { Ed25519KeyIdentity } from "@dfinity/identity";
+import { ECDSAKeyIdentity } from "@dfinity/identity";
 import { html, TemplateResult } from "lit-html";
 import { asyncReplace } from "lit-html/directives/async-replace.js";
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/register/captcha.ts
```diff
@@ -219,7 +219,9 @@ export const promptCaptcha = ({
     promptCaptchaPage({
       cancel: () => resolve(cancel),
       verifyChallengeChars: async ({ chars, challenge }) => {
-        const tempIdentity = Ed25519KeyIdentity.generate();
+        const tempIdentity = await ECDSAKeyIdentity.generate({
+          extractable: false,
```
nmattia, line 223: is this not the default, or are you being explicit?

nmattia (approved):
Thanks a lot!

src/frontend/src/flows/register/captcha.ts
frederikrothenberger: It is the default. But since this is an important property for us, I set it explicitly.

PR #1451, opened 2023-04-20
Refactor canister test API to work with borrows

frederikrothenberger:
This is refactoring to make the canister test APIs more ergonomic: In particular, wherever possible the API no longer takes ownership of the arguments. `String` args are changed to work with `&str` instead.

This paves the way to slowly move away from the test helpers in `framework` and use more appropriate local helpers instead.


src/internet_identity/tests/integration/delegation.rs
```diff
@@ -288,26 +288,24 @@ fn should_issue_different_principals() -> Result<(), CallError> {
     let env = env();
     let canister_id = install_ii_canister(&env, II_WASM.clone());
     let user_number = flows::register_anchor(&env, canister_id);
-    let frontend_hostname_1 = "https://dapp1.com";
-    let frontend_hostname_2 = "https://dapp2.com";
```
przydatek, line 292: nit: the reason for "in-lining" these is that they were used only once, right? 
I believe there is still some advantage of having named constants (`frontend_hostname_1` is more informative than `https://dapp1.com`) so maybe we could keep them anyway?

src/internet_identity/tests/integration/delegation.rs
```diff
@@ -496,23 +494,21 @@ fn should_return_different_principals_for_different_frontends() -> Result<(), Ca
     let canister_id = install_ii_canister(&env, II_WASM.clone());
     api::init_salt(&env, canister_id)?;
     let user_number = flows::register_anchor(&env, canister_id);
-    let frontend_hostname_1 = "https://dapp-1.com";
-    let frontend_hostname_2 = "https://dapp-2.com";
```
przydatek, line 500: ditto

przydatek (approved):
LGTM, with some nitpicking (totally up to you if you want to change it). 
Thanks!

PR #1452, opened 2023-04-20
Only import from lit-html

nmattia:
This ensures that all our lit imports (directives, etc) come from the `lit-html` package and not from `lit`. The `lit` imports may work but only because `lit` is a transitive dependency (through `lit-ssr`).


peterpeterparker (approved):
Thanks

PR #1453, opened 2023-04-20
Clean up qr-creator import

nmattia:
This removes the unnecessary `| null` alternative of the import, but instead wraps the import in `try/catch` and returns `undefined` in case of an error (and logs the error).


peterpeterparker (approved):
LGTM, thx

PR #1454, opened 2023-04-20
Remove About page and static pre-rendering

peterpeterparker:
# Motivation

We can remove the "About" page and replaces it with a link to the dedicated page about II on the portal.

As we do so, we can remove pre-rendering as well as no more used.

# Changes

- update about links to point to portal
- remove `about` component
- `npm rm @lit-labs/ssr -D`
- remove static pre-rendering from `webpack.config`
- remove flag `HYDRATE_STATIC_PAGES`


webpack.config.ts
```diff
@@ -207,11 +177,9 @@ export default [
     },
     plugins: [
       ...defaults.plugins,
-      // In production, we generate static versions of our static pages (during development we
-      // prefer hot-reloading dynamic pages);
       // in development, we inject canister ID when using the dev server, so that the local
```
nmattia, line 180:
```suggestion
      // In development, we inject canister ID when using the dev server, so that the local
```

nmattia (approved):
Thanks, LGTM!

I think there are still some dead references here:
* src/frontend/src/test-e2e/constants.ts
* src/frontend/src/test-e2e/register.test.ts
(though they weren't even used)

peterpeterparker:
> I think there are still some dead references here:

Thx. Both constants removed.

PR #1455, opened 2023-04-20
Don't duplicate logo fixup

nmattia:
This removes a replace function that cleaned up a dapp's logo path. The logo is already cleaned up by the `getDapps` function.


peterpeterparker (approved):
Thx

PR #1456, opened 2023-04-20
Reject recovery phrases with credential id

frederikrothenberger:
This restriction will make future migrations easier, given no such devices already exist (needs to be checked).


nmattia (approved):
Makes sense!

PR #1459, opened 2023-04-21
Make verification screens logic more consistent

nmattia:
This makes the countdown code a bit more robust and cleans up a few issues around verification & device addition screen:

* Add missing loader on verification cancel
* Thread all results through countdown, to ensure countdown is stopped always
* Use same `while` loops everywhere


peterpeterparker (approved):
LGTM

PR #1460, opened 2023-04-21
Expose II configuration as metrics

frederikrothenberger:
This PR exposes more configuration parameters as metrics. Exposing the config has 2 advantages:
* alerts can take the configuration into account and adapt based on the configured limits
* it declutters the candid interface once we remove the info from the `stats` endpoint


src/internet_identity/src/http.rs
```diff
@@ -193,6 +208,11 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
                 register_rate_limit_config.max_tokens as f64,
                 "The maximum number of `register` calls that are allowed in any time window.",
             )?;
+            w.encode_gauge(
+                "internet_identity_register_rate_limit_time_per_tokens_seconds",
+                Duration::from_nanos(register_rate_limit_config.time_per_token_ns).as_secs() as f64,
+                "Min number of nanoseconds between two register calls to not exceed the rate limit (sustained).",
```
przydatek, line 214: Given the conversion to seconds in the previous line, shouldn't it say "Min number of seconds"?

src/internet_identity/tests/integration/archive_integration.rs
```diff
@@ -460,6 +460,35 @@ mod pull_entries_tests {
...
+            &env,
+            II_WASM.clone(),
+            arg_with_wasm_hash(ARCHIVE_WASM.clone()),
+        );
+        deploy_archive_via_ii(&env, ii_canister);
+
+        assert_metric(
+            &get_metrics(&env, ii_canister),
+            "internet_identity_archive_config_entries_buffer_limit",
+            10_000f64,
```
przydatek, line 477: nit: I suppose these values are set by tools in `canister_tests/src/framework.rs`, right?
It would be probably nicer to set & check some local constants, but I'm not sure whether this could be easily achieved -- if it is to much work, then leaving as-is is also fine.

przydatek (approved):
LGTM, modulo minor suggestions below.  Thanks!

src/internet_identity/src/http.rs
frederikrothenberger: Ah, yes. Thanks!

Issue #1458, opened 2023-04-21
Post vitejs TODOs


peterpeterparker:
maybe:

- use `.env` (`envDir: "../../",` in config will be needed)
- `npm run format --write .` (prettier everything)

done:

- `tsc` 👉 #1467
- remove `text-encoding` 👉 #1469
- remove webpack chunk name in code 👉 #1470
- use root for favicon 👉 #1476
- add `$lib` etc. alias 👉 #1475 and #1488
- replace webpack in `test-app` and `using-dev-build` 👉 #1485
- `import` assets in code instead of copying and referencing from `public` the images 👉 #1482
- ~~use `import.meta.env.BASE_URL`~~ use asset handling for `BASEL_URL/loader.webp` 👉 #1472
- ~~remove `ts-node` for `screenshots`~~ 
- `npm run dev` instead of `npm run start` 👉 #1501
- minify HTML - e.g. with https://www.npmjs.com/package/html-minifier-terser 👉 #1576
- switch to port `5173` 👉 #1577
- use `import.env....` instead of `process.env` 👉 #1514
- `vitetest` instead of `jest` 👉 https://github.com/dfinity/internet-identity/pull/1579

PR #1461, opened 2023-04-21
Replace webpack with vitejs

peterpeterparker:
# Changes

- drop `webpack` for `vite`
  - remove webpack dependencies
  - add various vite dependencies
  - add configuration `vite.config.ts`
  - convert webpack plugins - e.g. for canister IDs or to remove `script` - to `vite.plugin.ts`
  - move and adapt `html` file for `index.html`
  - move `showcase` html and ts file to separate directory
-  set `buffer`, `process` and `stream-browserify` as dependencies instead of dev dependencies because these are such
- polyfill `buffer` manually in `index.ts`
- set script to be loaded as `module` by canister

Remaining non-production blocker tasks in https://github.com/dfinity/internet-identity/issues/1458.


.github/workflows/canister-tests.yml
```diff
@@ -621,7 +621,6 @@ jobs:
 
   clean-build:
     runs-on: ${{ matrix.os }}
-    if: ${{ github.ref == 'refs/heads/main' }}
```
nmattia, line 624: leftover?

vite.config.ts
```diff
@@ -0,0 +1,94 @@
+import { extname } from "path";
+import { defineConfig, UserConfig } from "vite";
+import viteCompression from "vite-plugin-compression";
+import { injectCanisterIdPlugin, stripInjectJsScript } from "./vite.plugins";
+
+const defaultConfig = (mode?: string): Omit<UserConfig, "root"> => {
+  // Path "../../" have to be expressed relative to the "root".
+  // e.g.
+  // root = src/frontend
+  // envDiv = ../../ because .env is at the base of the project
```
nmattia, line 10:
```suggestion
  // envDir = ../../ because .env is at the base of the project
```

vite.config.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  return {
+    envDir: "../../",
```
nmattia, line 13: @peterpeterparker can you confirm that you use this locally?

nmattia (commented):
Cool! couple questions

When I do `npm run build`, the `index.html` doesn't seem to be minified. Is that on purpose?

.github/workflows/canister-tests.yml
peterpeterparker: yes. redo ✅

vite.config.ts
peterpeterparker: Removed.

vite.plugins.ts
```diff
@@ -0,0 +1,59 @@
...
+  name: "html-transform";
+  transformIndexHtml(html: string): string;
+} => ({
+  name: "html-transform",
+  transformIndexHtml(html): string {
+    return html.replace(
+      `<script type="module" crossorigin src="/index.js"></script>`,
+      ``
+    );
+  },
```
nmattia, line 58: I'm afraid this is going to break silently if vite ever changes a single character from the generated `<script...>` tag. Wdyt?

peterpeterparker: It isn't like we have much choice, don't have other idea.

We can write a test that checks there is no  script with mention of `index.js` after `npm run build`?

vite.plugins.ts
```diff
@@ -0,0 +1,59 @@
...
+export const stripInjectJsScript = (): {
+  name: "html-transform";
+  transformIndexHtml(html: string): string;
+} => ({
+  name: "html-transform",
+  transformIndexHtml(html): string {
+    return html.replace(
+      `<script type="module" crossorigin src="/index.js"></script>`,
+      ``
+    );
```
nmattia, line 57:
```suggestion
    const match = `<script type="module" crossorigin src="/index.js"></script>`;
    if (!html.includes(match)) {
      throw new Error("Expecting script tag to replace, found none");
    }
    return html.replace(match, ``);
```

vite.plugins.ts
nmattia: Yeah, added suggestion

peterpeterparker: Thanks!

nmattia (approved):

PR #1463, opened 2023-04-24
Shuffle dapps for explorer

nmattia:
This ensures that the ordering of the dapps in the teaser & explorer change every time the user reloads the management screen, encouraging discovery of new dapps.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -455,11 +455,10 @@ const unknownError = (): Error => {
   return new Error("Unknown error");
 };
 
-// Return a shuffled version of the array. Adapted from
-// https://stackoverflow.com/questions/2450954/how-to-randomize-shuffle-a-javascript-array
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/manage/index.ts
```diff
@@ -453,3 +454,15 @@ export const domainWarning = (
 const unknownError = (): Error => {
   return new Error("Unknown error");
 };
+
+// Return a shuffled version of the array. Adapted from
+// https://stackoverflow.com/questions/2450954/how-to-randomize-shuffle-a-javascript-array
```
peterpeterparker, line 459: can you update the link to point to the exact answer used?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/manage/index.ts
nmattia: thought I did! my bad

PR #1464, opened 2023-04-24
Make duplicate device error message understandable for users

frederikrothenberger:
This PR replaces the very unreadable error (technical) error message when trying to register an already registered device with something that users hopefully understand.

In particular, in replaces this message:
image

New screen:
Screenshot 2023-04-24 at 11 18 34

Tested (so far) on MacOs for FF, Chrome and Safari. Other errors I could provoke (like deny auth on the WebAuthn prompt) did retain their own error message.

### DIFF
--- a/src/frontend/src/utils/webAuthnErrorUtils.ts
+++ b/src/frontend/src/utils/webAuthnErrorUtils.ts
@@ -1,0 +1,8 @@
+/** Checks whether the error corresponds with the WebAuthnSpec for duplicate device:
+ *  * https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 20)
+ *  * https://www.w3.org/TR/webauthn-2/#sctn-op-make-cred (Step 3)
+ * @param error error to check
+ */
+export function isDuplicateDeviceError(error: unknown): boolean {
+  return error instanceof DOMException && error.name === "InvalidStateError";
+}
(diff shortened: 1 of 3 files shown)


frederikrothenberger:
Added @marydwyer to review the copy.

src/frontend/src/utils/webAuthnErrorUtils.ts
nmattia, line 1: Do you expect many more webauthn error helpers? how about dropping this in `utils.ts` for now?

src/frontend/src/utils/webAuthnErrorUtils.ts
```diff
@@ -0,0 +1,8 @@
+/** Checks whether the error corresponds with the WebAuthnSpec for duplicate device:
+ *  * https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 20)
+ *  * https://www.w3.org/TR/webauthn-2/#sctn-op-make-cred (Step 3)
```
nmattia, line 3: Looking at these, there are no guarantees that this is actually a "duplicate" error, right?

nmattia (commented):
Very helpful!

src/frontend/src/utils/webAuthnErrorUtils.ts
frederikrothenberger: I planned on adding at least a second one for "user denied access".

src/frontend/src/utils/webAuthnErrorUtils.ts
frederikrothenberger: No, unfortunately not. However, this is also something that [other libraries](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/browser/src/helpers/identifyRegistrationError.ts#L45) do (because it's the next best thing).

nmattia (approved):

PR #1466, opened 2023-04-24
Suggest adding dapp in explorer

nmattia:
This adds a small paragraph suggesting to the user to add their own dapp(s) to the explorer, with a link to create a ticket on GitHub.

Screenshot 2023-04-24 at 11 28 28


peterpeterparker (approved):
LGTM

Notes:

- i18n in another PR?
- Not sure if saving few pixels of spacing under the new link would looks better? (there is more space above that under)

nmattia:
> i18n in another PR?

woops, you're right!

> Not sure if saving few pixels of spacing under the new link would looks better? (there is more space above that under)

I'll let a designer look at it first and adapt when we have a suggestion + the final copy

PR #1467, opened 2023-04-24
Add tsc types check

peterpeterparker:
# Motivation

Vite transpile only. This PR adds `tsc` to the build and a separate `watch` task to run optionnally tsc while developping.

Documentation: https://vitejs.dev/guide/features.html#transpile-only


package.json
```diff
@@ -8,7 +8,8 @@
     "start": "II_FETCH_ROOT_KEY=1 vite",
     "host": "II_FETCH_ROOT_KEY=1 vite --host",
     "showcase": "II_FETCH_ROOT_KEY=1 vite --mode showcase",
-    "build": "vite build",
+    "build": "vite build && tsc --noEmit",
```
nmattia, line 11: I think we could swap these (i.e. start with `tsc`) since there's no point in building if typescript fails. Worse even, if for some reason the build is canceled before ts is run, the dist will exist but it may not be well typed.

Note: that's also what astro suggest: https://docs.astro.build/en/guides/typescript/#type-checking

nmattia (commented):
Very nice, thanks!

package.json
peterpeterparker: Sure

nmattia (approved):

PR #1471, opened 2023-04-24
Fix showcase deploy

nmattia:
This updates the `showcase` CI job to use the correct showcase dist directory. The showcase output dir is also renamed for clarity (showcase -> dist-showcase) and the dir added to the gitignore. Finally, the CI job is changed to run on all branches (the build is; the deploy is left to only run on main).


peterpeterparker (approved):
LGTM, thx

PR #1472, opened 2023-04-24
Import loader.webp without process.env

peterpeterparker:
# Motivation

Use Vite static asset handling to import `loader.webp`.

https://vitejs.dev/guide/assets.html

Note:

- because we explicitely set chunk name in the build config, the keep its name during build process (as currently) - i.e. no timestamp is added to the filename as displayed in the documentation

- not sure it would make contribution easy but we could transform `dapps.json` to a `dapps.ts` and use the same pattern if we want to

### DIFF
--- a/jest-transform.cjs
+++ b/jest-transform.cjs
@@ -1,0 +1,18 @@
+const path = require("path");
+
+/**
+ * Jest cannot process images imported within the code. This loads their path as modules for test purpose.
+ *
+ * e.g.
+ * import loaderUrl from "../../assets/loader.webp";
+ */
+module.exports = {
+  process(sourceText, sourcePath, options) {
+    return {
+      code: `module.exports = ${JSON.stringify(path.basename(sourcePath))};`,
+    };
+  },
+  getCacheKey() {
+    return "img-transform";
+  },
+};
--- a/src/frontend/env.d.ts
+++ b/src/frontend/env.d.ts
@@ -1,0 +1,2 @@
+/// <reference types="vite/client" />
+declare module "envConfig";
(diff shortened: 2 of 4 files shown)


src/frontend/env.d.ts
```diff
@@ -0,0 +1,2 @@
+/// <reference types="vite/client" />
+declare module "envConfig";
```
nmattia, line 2: What is this for?

jest-transform.cjs
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * e.g.
+ * import loaderUrl from "../../assets/loader.webp";
+ */
+module.exports = {
+  process(sourceText, sourcePath, options) {
+    return {
+      code: `module.exports = ${JSON.stringify(path.basename(sourcePath))};`,
+    };
+  },
```
nmattia, line 14: should this somehow be limited by file extension?

src/frontend/env.d.ts
peterpeterparker: > TypeScript, by default, does not recognize static asset imports as valid modules. To fix this, include [vite/client](https://vitejs.dev/guide/features.html#client-types).

https://vitejs.dev/guide/assets.html

jest-transform.cjs
peterpeterparker: It's done by the transformer pattern of the configuration

```
"\\.(svg|png|webp)$": "<rootDir>/jest-transform.cjs",
```

src/frontend/env.d.ts
nmattia: Can you add a note about this in the file?

peterpeterparker: Added a link to the doc

nmattia (approved):
Thanks!

PR #1473, opened 2023-04-24
replace external icon with open button

meodai:


src/frontend/src/flows/dappsExplorer/index.ts
```diff
@@ -47,12 +49,10 @@ const dappsExplorerTemplate = ({
...
 /* Template for a single dapp */
-const dappTemplate = ({
-  link,
-  logo,
-  name,
-  oneLiner,
-}: DappDescription): TemplateResult => {
+const dappTemplate = (
+  { link, logo, name, oneLiner }: DappDescription,
+  copy: any
```
meodai, line 54: @nmattia I might need help with that, I was unsure how to declare the correct type here.

nmattia: you can pass `{ open_cta: DynamicKey }` I think, might be the easiest

PR #1475, opened 2023-04-24
Use import alias for generated

peterpeterparker:
# Motivation

Shorten import syntax with an alias for `generated` resources.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,4 +1,4 @@
-import { extname } from "path";
+import { extname, resolve } from "path";
 import { defineConfig, UserConfig } from "vite";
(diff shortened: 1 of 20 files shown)


vite.config.ts
```diff
@@ -14,6 +14,8 @@ const defaultConfig = (mode?: string): Omit<UserConfig, "root"> => {
       alias: {
         // Polyfill stream for the browser. e.g. needed in "Recovery Phrase" features.
         stream: "stream-browserify",
+        // Custom alias we are using to shorten the imports
```
nmattia, line 17: shorten _and_ make absolute :)

nmattia (approved):
NICE

PR #1476, opened 2023-04-24
Use root for favicon

peterpeterparker:
# Motivation

> files in the public directory are served at the root path.
> Instead of /assets/favicon.ico, use /favicon.ico.


nmattia (approved):
LGTM, though we may need to change this to include BASE_URL (for instance for http://dfinity.github.io/internet-identity/)

PR #1479, opened 2023-04-24
Correctly set base url for showcase build

nmattia:
This ensures that the showcase can find
`/internet-identity/index.{js,css}` when deployed on GitHub pages.


frederikrothenberger (approved):
Rubber stamp

PR #1481, opened 2023-04-24
Don't compress showcase assets

nmattia:
This disables vite compression for all showcase assets, because GitHub pages doesn't serve `index.js.gz` as `index.js`.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -39,18 +39,20 @@ const defaultConfig = (mode?: string): Omit<UserConfig, "root"> => {
     plugins: [
       [...(mode === "development" ? [injectCanisterIdPlugin()] : [])],
       [...(mode === "production" ? [stripInjectJsScript()] : [])],
-      viteCompression({
-        // II canister only supports one content type per resource. That is why we remove the original file.
-        deleteOriginFile: true,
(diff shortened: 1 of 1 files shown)


vite.config.ts
```diff
@@ -42,8 +42,14 @@ const defaultConfig = (mode?: string): Omit<UserConfig, "root"> => {
       viteCompression({
         // II canister only supports one content type per resource. That is why we remove the original file.
         deleteOriginFile: true,
+        // GitHub pages doesn't serve gzipped files as .js, so we disable compression
+        // for the showcase build
         filter: (file: string): boolean =>
-          ![".html", ".css", ".webp", ".png", ".ico"].includes(extname(file)),
+          mode === "showcase"
```
peterpeterparker, line 48: for consistency reason can you rather not use the plugin if showcase?

```
[...(mode === "showcase" ? [viteCompression({
        // II canister only supports one content type per resource. That is why we remove the original file.
        deleteOriginFile: true,
        filter: (file: string): boolean =>
          ![".html", ".css", ".webp", ".png", ".ico"].includes(extname(file)),
      })] :  [])],
```

peterpeterparker (approved):
LGTM, thx

PR #1482, opened 2023-04-24
Remove loader.webp from assets

nmattia:
With the `loader.webp` still in the assets, vite was still bundling it (and was complaining about it being both bundled and imported directly).


src/frontend/src/components/loader.ts
```diff
@@ -1,5 +1,5 @@
 import { html, render } from "lit-html";
-import loaderUrl from "../../assets/loader.webp";
+import loaderUrl from "./loader.webp";
```
peterpeterparker, line 2: I did not moved it in previous PR because I thought you would not like it 😉

What about making a a subfolder `assets` under `src` and later on rename `assets` to `public` (default value of vite)?

```
src/frontend/src/assets -> images and json `import` in code
src/frontend/public -> static stuffs like favicon.ico and robots.txt
```

nmattia: I'd rather actually have related things together, i.e. "copy and assets for screen 'welcome' are near screen 'welcome'" instead of arbitrarily creating structure; what do you think?

That helps with discovery & mental model/map, plus it's much easier to nuke a feature (by simply removing a directory) if necessary (instead of hunting for tons of files scattered throughout the codebase)

peterpeterparker: Works for me sure

peterpeterparker (approved):
LGTM

PR #1484, opened 2023-04-24
Use correct BASE_URL in showcase routing

nmattia:
This ensures the correct base url (set by `--base` when running vite) is used; otherwise the showcase routing does not work.

### DIFF
--- a/src/showcase/src/showcase.ts
+++ b/src/showcase/src/showcase.ts
@@ -621,8 +621,7 @@ const init = () => {
   // This is not very user friendly (in particular we don't show anything like a
   // 404) but this is an dev page anyway.
   const route = window.location.pathname
-    .replace(import.meta.env.BASE_URL ?? "", "")
-    .substring(1);
+    .replace(import.meta.env.BASE_URL ?? "/", "");
   const page = iiPages[route] ?? defaultPage;
 
   page();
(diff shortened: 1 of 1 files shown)


src/showcase/src/showcase.ts
```diff
@@ -621,7 +621,7 @@ const init = () => {
   // This is not very user friendly (in particular we don't show anything like a
   // 404) but this is an dev page anyway.
   const route = window.location.pathname
-    .replace(process.env.BASE_URL ?? "", "")
+    .replace(import.meta.env.BASE_URL ?? "", "")
```
peterpeterparker, line 624: > TS2339: Property 'env' does not exist on type 'ImportMeta'.

Can you either:

- copy/paste `env.d.ts` in `showcase/src`
or
- add following to `showcase/tsconfig.json`

```
"compilerOptions": {
    "types": ["vite/client"]
  }
```

nmattia: It works for me; do you get this when running `tsc --noEmit`?

nmattia: Can you try with `tsc`? That's the source of truth; you may have an editor setup issue

peterpeterparker: Same with `tsc`

Capture d’écran 2023-04-24 à 18 23 29

PR #1485, opened 2023-04-24
Replace webpack with vitejs for `using-dev-build`

peterpeterparker:
# Motivation

Vite, vite everywhere

# Changes

- Replace webpack libs and config with vite
- use `ts-node` to run the test
- add plugin from Juno to load the canister ids
- adapt `index.html` and `index.js/ts` for Vite and TypeScript
- replicate prettier formatting from root


nmattia (commented):
Very nice! couple questions

demos/using-dev-build/webapp/index.ts
```diff
@@ -2,44 +2,50 @@
...
   let iiUrl;
   if (process.env.DFX_NETWORK === "local") {
-    iiUrl = `http://localhost:4943/?canisterId=${process.env.II_CANISTER_ID}`;
+    iiUrl = `http://localhost:4943/?canisterId=${process.env.INTERNET_IDENTITY_CANISTER_ID}`;
   } else if (process.env.DFX_NETWORK === "ic") {
-    iiUrl = `https://${process.env.II_CANISTER_ID}.ic0.app`;
+    iiUrl = `https://${process.env.INTERNET_IDENTITY_CANISTER_ID}.ic0.app`;
   } else {
-    iiUrl = `https://${process.env.II_CANISTER_ID}.dfinity.network`;
+    iiUrl = `https://${process.env.INTERNET_IDENTITY_CANISTER_ID}.dfinity.network`;
```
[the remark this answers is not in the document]
peterpeterparker: From its declaration in `.dfx.json`

```
"canisters": {
    "internet_identity": {
```

Before canisters' names were manually mapped and for some reason only the II canister name was shortened while the two others, webapp and whoami, weren't.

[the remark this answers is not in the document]
peterpeterparker: > Does it break if you leave II_CANISTER_ID?

Yes. I can change the way the canister IDs are mapped to make an exception but I rather like to have things loaded automtically than having to write "if this then that". Also the scripts used to map the canister IDs can be used by anyone in the community in any vite project that uses dfx.

But again, if you want to add an exception, I can.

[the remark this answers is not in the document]
peterpeterparker: See `demos/using-dev-build/vite.plugins.ts` in this PR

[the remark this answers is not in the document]
frederikrothenberger: @nmattia: The webpack config used to map these for some reason: `II_CANISTER_ID: process.env.INTERNET_IDENTITY_CANISTER_ID`, this part I did not touch.

The demo webapp works as described in the readme. But yes, the login flow does not seem to be covered by the selenium tests.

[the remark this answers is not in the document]
nmattia: Sorry, I'm a bit slow. I think I get it.

This is how the value was set before:

``` ts
new webpack.EnvironmentPlugin({
       WHOAMI_CANISTER_ID: process.env.WHOAMI_CANISTER_ID,
       II_CANISTER_ID: process.env.INTERNET_IDENTITY_CANISTER_ID,
       DFX_NETWORK: process.env.DFX_NETWORK || "local",
     }),
```

now it makes sense.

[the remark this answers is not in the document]
nmattia: About the webapp: the README here states this:

> See [`index.js`](./webapp/index.js) for the full working example.

I don't know if the webapp is actually tested/run/built, I might simply remove it

demos/using-dev-build/vite.plugins.ts
```diff
@@ -0,0 +1,38 @@
+import { readFileSync } from "fs";
+import { join } from "path";
+
+export const readCanisterIds = ({
+  prefix,
```
nmattia, line 5: is prefix used?

demos/using-dev-build/vite.config.ts
```diff
@@ -0,0 +1,41 @@
...
+      emptyOutDir: true,
+      commonjsOptions: {
+        // Source: https://github.com/rollup/plugins/issues/1425#issuecomment-1465626736
+        strictRequires: true,
+      },
+    },
+    define: {
+      "process.env": {
+        ...readCanisterIds({ network }),
+        DFX_NETWORK: network,
```
nmattia, line 24: is this used?

demos/using-dev-build/vite.plugins.ts
```diff
@@ -0,0 +1,38 @@
...
+  network,
```
nmattia, line 6: why not read the network here?

demos/using-dev-build/vite.plugins.ts
peterpeterparker: its a plugin / mapper that works in any project for any developer of the community.

since `using-dev-build` is an example of project for the community, I think it makes sense to provide a mapper that works all the time therefore the few unused options.

if you want me to remove anything that we do not use here, can do. Let me know.

demos/using-dev-build/vite.config.ts
peterpeterparker: same as https://github.com/dfinity/internet-identity/pull/1485#discussion_r1176188911

peterpeterparker: agent-js uses `DFX_NETWORK`. while we might not hits it, same as https://github.com/dfinity/internet-identity/pull/1485#discussion_r1176188911

demos/using-dev-build/vite.plugins.ts
peterpeterparker: because it is also needed elsewhere currently in `vite.config.ts` that uses also `vite.plugin.ts` - avoid code duplication

demos/using-dev-build/vite.plugins.ts
nmattia: Yeah especially if we're gonna share it with the community I'd rather make it as simple as possible; I think having a parameter that's just unused is more confusing than anything. What do you think?

demos/using-dev-build/vite.config.ts
nmattia: Ok gotcha; confusing still. Let's remove anything we don't need (also for the community's sake, otherwise other people might copy this blindly and also end up with unused values)

demos/using-dev-build/vite.plugins.ts
peterpeterparker: I removed the options we don't use.

demos/using-dev-build/vite.plugins.ts
```diff
@@ -1,22 +1,17 @@
...
-  prefix,
-  network,
-}: {
-  prefix?: string;
-  network: string | "ic" | "local";
-}): Record<string, string> => {
-  const canisterIdsJsonFile =
-    network === "ic"
-      ? join(process.cwd(), "canister_ids.json")
-      : join(process.cwd(), ".dfx", "local", "canister_ids.json");
```
nmattia: I think this is still needed for mainnet deployments of the app 😬

peterpeterparker: Discussed offline. `network` added back to `vite.plugins`

demos/using-dev-build/package.json
```diff
@@ -2,31 +2,31 @@
   "name": "@dfinity/internet-identity-e2e-tests",
   "version": "1.0.0",
   "private": true,
+  "type": "module",
   "scripts": {
     "wdio": "wdio run ./wdio.conf.ts",
-    "test": "./test.ts",
+    "opts": "NODE_OPTIONS='--loader ts-node/esm --experimental-specifier-resolution=node' \"$@\"",
```
nmattia, line 8: Is this being used by the scripts?

peterpeterparker: Ah no good catch, I tried it out during development to solve a problem I had. Removed.

nmattia (approved):
Thanks a lot! not 100% sure this'll still work on mainnet but we'll have to try

PR #1487, opened 2023-04-25
Refactor compression in plugin and only for production

peterpeterparker:
# Motivation

Simplify usage of the compression plugin.

- Refactor configuration to `vite.plugins.ts`. That way it's easier to comment it locally if required
- Use the compression plugin only for `production` build instead of if `not showcase`


nmattia (approved):
Thanks!

PR #1489, opened 2023-04-25
Replace webpack with vitejs for `test-app`

peterpeterparker:
# Changes

- replace webpack libs and config with vite
- update JS bundle path in rust test canister


nmattia (approved):
Nice, thanks!

PR #1494, opened 2023-04-25
Add role=alert to toast messages

nmattia:
The `role="alert"` attribute (equivalent to setting `aria-live="assertive"`) is used to communicate an important and usually time-sensitive message to the user, like toast errors.

Addresses https://github.com/dfinity/internet-identity/pull/1431/files#r1166919990
Related: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role


meodai (approved):
LGTM

PR #1495, opened 2023-04-25
Improve CAPTCHA HTML

nmattia:
This adds (better) a11y labels to the CAPTCHA elements and removes an unused attribute.

Unfortunately we do not provide a non-visual fallback for the CAPTCHA.


meodai (approved):
LGTM! I think we can do more in the future though: https://www.boia.org/blog/how-to-make-captcha-accessible-to-everyone

PR #1496, opened 2023-04-25
Extract WebAuthn Cancel Copy

frederikrothenberger:
This extracts the copy for the WebAuthn cancel error.

To make it work, I had to change the types in LoginFlowResult causing additional changes in `phrase.ts`. I figured the change is still small enough, but I'm also happy to split.


src/frontend/src/utils/flowResult.ts
```diff
@@ -20,9 +21,9 @@ export type LoginFlowError = {
 } & LoginError;
 
 export type LoginError = {
-  title: string;
-  message: string;
-  detail?: string;
+  title: TemplateElement;
+  message: TemplateElement;
+  detail?: TemplateElement;
```
nmattia, line 26: `TemplateElement`s can be anything (including html), do you mean to use this or would the lighter version (`DynamicKey`, specific to the copy) work?

frederikrothenberger: It needs to be at least `string | DynamicKey` (I think). I.e. still allow the previous version of the code. But I should be able restrict to that.

src/frontend/src/utils/webAuthnErrorUtils.ts
```diff
@@ -1,16 +1,18 @@
...
+import { I18n } from "../i18n";
+import copyJson from "./webAuthnErrorUtils.json";
 
-export const WEBAUTHN_CANCEL_TEMPLATE = {
-  title: "Operation canceled",
-  message:
-    "The interaction with your security device was canceled or timed out. Please try again.",
+export const webAuthnCancelTemplate = () => {
+  const i18n = new I18n();
+  return i18n.i18n(copyJson);
```
nmattia, line 7: this is not a template (at least not in the lit sense), plus as much as possible we avoid having the components create their own I18n (instead the screens/pages provide the copy). I think there are other ways to solve this but I'm happy to merge this (if we really need to extract I18n for this specific error)

nmattia (approved):

PR #1497, opened 2023-04-25
Fix up add device success

nmattia:
This only shows the "success" add device message on successful device verification and renames the `addDeviceSuccess` screen to be more consistent with other screens.


peterpeterparker (approved):
LGTM, thx

Note: should we add an e2e?

nmattia:
> Note: should we add an e2e?

I was debating internally, but I realized the issue wasn't that we introduced a bug but rather that we (at least I) didn't think of checking that; so I don't think a test would have helped. For future proper behavior of the flow I think the types should be enough but I'm happy to add a test if you want!

peterpeterparker:
> I'm happy to add a test if you want!

I would not be against because testing it requires two browser which is not the common daily dev flow but indeed, types should covers it with current routing way.

nmattia:
Clarified offline, won't add a test just now; we'll need to review the flows first

PR #1498, opened 2023-04-25
Make CI fail on rust warnings

frederikrothenberger:
This changes the clippy check to fail on all rust warnings.


nmattia (approved):
Thanks!

nmattia:
out of curiosity, aren't those cargo warnings (instead of clippy)?

frederikrothenberger:
> out of curiosity, aren't those cargo warnings (instead of clippy)?

Yes, but clippy will emit those too.

PR #1500, opened 2023-04-25
Show the success screen on the new device

peterpeterparker:
# Motivation

In addition to displaying the new success screen on register new device for the signed-in anchore, it was requested to also route to the new screen on the newly added tentative device.

# TODOs (in separate PRs)

- refactor code
  - extract code to "add device" from `src/index.ts` to a particular component
  - use object parameters for `authenticate`
- we should clean the browser url (e.g. [SO](https://stackoverflow.com/a/41061471/5404186)) on the device that used the link / on the new device otherwise, if user sign in and then out on that particular device, user lands on the device link again instead of the home screen 

# Changes

- implement sign-in flow in `src/index.ts` with known `userNumber`
  1. `addDeviceSuccess`
  2. `authenticate`
  3. `handleLoginFlowResult`
  4. go to `manage`

# Screenshots

demo2


src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
```diff
@@ -37,7 +37,7 @@ const displayAlreadyRegisteredDevice = () =>
 export const registerTentativeDevice = async (
   userNumber: bigint,
   connection: Connection
-): Promise<"ok"> => {
+): Promise<{ userNumber: bigint; alias: string } | never> => {
```
nmattia, line 40: doesn't this just collapse to `Promise<{ userNumber: bigint; alias: string }>`?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -77,27 +78,39 @@ export const authenticateBox = async (
...
-          detail: result.detail !== "" ? result.detail : undefined,
-          primaryButton: "Try again",
-        });
-        break;
-      case "canceled":
-        break;
-      default:
-        unreachable(result);
-        break;
+    const { success: loginData } = await handleLoginFlowResult(result);
```
nmattia, line 81: what's the case for having `success` be nullish? wouldn't something like this be more clearer?

```
{ success: LoginData } | undefined
```

src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
```diff
@@ -96,13 +96,15 @@ export const registerTentativeDevice = async (
 
   // If everything went well we can now ask the user to authenticate on an existing device
   // and enter a verification code
-  return await showVerificationCode(
+  await showVerificationCode(
```
nmattia, line 99: this returns `"ok"` which was previously returned; that ensured that if `showVerificationCode` was ever modified to return something else (like `"ok" | "failed") the typecheck would warn us here.

Do you think would make sense to ensure the result is `"ok"` here?

src/frontend/src/index.ts
```diff
@@ -104,17 +105,24 @@ const init = async () => {
...
+    await addDeviceSuccess({ deviceAlias });
+
+    // If user "Click" continue in success page, proceed with authentication
+    const result = await authenticate(connection, userNumber);
+    const { success } = await handleLoginFlowResult(result);
+
+    // User have successfully signed-in we can jump to manage page
+    if (nonNullish(success)) {
+      const { connection: authenticatedConnection } = success;
+      await manage({ userNumber, authenticatedConnection });
```
nmattia, line 123: what do you think about this?

```suggestion
      return await manage({ userNumber, authenticatedConnection });
```

this would ensure that if `manage` ever returns anything meaningful, the typechecker would warn us about this call here (you can change `manage` to return `never`)

src/frontend/src/index.ts
```diff
@@ -104,17 +105,24 @@ const init = async () => {
...
+
+    // Display a success page once device added (above registerTentativeDevice **never** returns if it fails)
+    await addDeviceSuccess({ deviceAlias });
+
+    // If user "Click" continue in success page, proceed with authentication
+    const result = await authenticate(connection, userNumber);
+    const { success } = await handleLoginFlowResult(result);
+
+    // User have successfully signed-in we can jump to manage page
+    if (nonNullish(success)) {
```
nmattia, line 121: what does the user see if it's not a success?

src/frontend/src/test-e2e/register.test.ts
```diff
@@ -105,51 +105,123 @@ test("Register new identity and add additional device", async () => {
   });
 }, 300_000);
 
+const initDeviceLink = async (
```
nmattia, line 108: so far we've either duplicated code or added helpers to `flows.ts`. Would it make sense to stay consistent?

src/frontend/src/test-e2e/register.test.ts
```diff
@@ -105,51 +105,123 @@ test("Register new identity and add additional device", async () => {
...
+        addDeviceLink,
+      });
+    });
+
+    await mainView.waitForDisplay();
+    await mainView.waitForDeviceDisplay(DEVICE_NAME2);
+  });
+}, 300_000);
+
+test.only("Register new identity and sign-in with new additional remote device", async () => {
```
nmattia, line 187: leftover?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -77,27 +78,39 @@ export const authenticateBox = async (
...
+
+    if (nonNullish(loginData)) {
+      return loginData;
     }
   }
 };
 
+export const handleLoginFlowResult = async (
```
nmattia, line 89: would it make sense for this to leave in `flowResult.ts`?

src/frontend/src/index.ts
nmattia, line 123: it looks like `manage` is this:

``` ts
  // Here, if the user doesn't have any recovery device, we prompt them to add
  // one. The exact flow depends on the device they use.
  await recoveryWizard(userNumber, authenticatedConnection);
  // From here on, the user is authenticated to II.
  void renderManage(userNumber, authenticatedConnection);
```

would it make sense to simply call `renderManage`? if you've just added a device, you probably don't care about adding a recovery just now and you just want to complete the flow

src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
peterpeterparker: Yes but it's more readable. That way I understand straight from the begin that this function might never returns.

src/frontend/src/components/authenticateBox.ts
peterpeterparker: On error and when user click `Cancel`.

src/frontend/src/flows/addDevice/welcomeView/registerTentativeDevice.ts
peterpeterparker: Not sure but no strong opinion. What would be the check and what should happens if not matching?

src/frontend/src/components/authenticateBox.ts
nmattia: Sure I understand, I'm wondering about the field; would it make sense to simplify and nullify everything, or even have `LoginData | undefined`?  or `LoginData | "canceled"` for clarity?

src/frontend/src/index.ts
peterpeterparker: I never do this and never had to but if you wish, no strong opinion.

src/frontend/src/index.ts
peterpeterparker: There was an error which was prompted or user clicked "Cancel" to which he/she except that nothing else happens.

src/frontend/src/test-e2e/register.test.ts
peterpeterparker: Yes thanks

src/frontend/src/test-e2e/register.test.ts
peterpeterparker: sure

src/frontend/src/components/authenticateBox.ts
peterpeterparker: Yes but, see PR description, in a separate PR.

src/frontend/src/index.ts
peterpeterparker: No idea, don't know the flow and routing enough well. Should I make this change in this PR or would you like to implement it yourself afterwards?

src/frontend/src/components/authenticateBox.ts
peterpeterparker: I like it and find it readable as it is, otherwise I would not have coded it that way. If you want something else, please let me know what and happy to obligate. No strong opinion here.

src/frontend/src/index.ts
nmattia: I think you can do it in this PR. You can simply call `renderManage` instead of the new `manage` you created. Probably less surprising/annoying for the user!

PR #1501, opened 2023-04-25
Replace npm run start with npm run dev

peterpeterparker:
# Motivation

Use default development command and the one used in NNS-dapp and gix-cmp to run local development.

# Changes

- `npm run start` 👉 `npm run dev`

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -64,7 +64,7 @@ To serve the frontend locally (recommended during development), run
 the following:
 
 ```bash
-npm run start
+npm run dev
 ```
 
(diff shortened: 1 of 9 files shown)


nmattia (commented):
Actually, the most useful command (currently) is `npm run showcase`, should this be `npm run dev`?

peterpeterparker:
> Actually, the most useful command (currently) is npm run showcase, should this be npm run dev?

I would not because in dapps of the community and NNS-dapp, `npm run dev` is meant to develop locally the app with logic and against canister. Showcase is particular to II.

PR #1503, opened 2023-04-25
Fix showcase iframe links

nmattia:
The links weren't built correctly, and the routes weren't inferred correctly.


peterpeterparker (approved):
Thx

PR #1510, opened 2023-04-25
Extract domainWarning from manage page

nmattia:
This ensures the domaim warning can be used from the migration tests without depending on the management page. Jest cannot load the management page because it uses next-gen Vite features.

We'll need to move to vitest or find a workaround.


peterpeterparker:
Just want to have a quick test on main, will merge it afterwards, give me 5 as discussed offline

nmattia:
Closing in favor of #1512

PR #1512, opened 2023-04-25
Mock import.meta.env

peterpeterparker:
# Motivation

Jest are failing with following errors:

>     src/frontend/src/environment.ts:1:25 - error TS1343: The 'import.meta' meta-property is only allowed when the '--module' option is 'es2020', 'es2022', 'esnext', 'system', 'node16', or 'nodenext'.
>    
>        1 export const BASE_URL = import.meta.env.BASE_URL;

To overcome the issue, in NNS-dapp, we regroup the env in a unique file and mock the values for jest test.

### DIFF
--- a/src/frontend/src/environment.ts
+++ b/src/frontend/src/environment.ts
@@ -1,0 +1,1 @@
+export const BASE_URL = import.meta.env.BASE_URL;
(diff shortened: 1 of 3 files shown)


nmattia (approved):
Would be nice to have all config values in `config.ts` but we can tackle that later

PR #1513, opened 2023-04-25
Replace external link unicode by icon

meodai:
We were using a unicode character for external links that looked a bit questionable on many systems.
Also makes the icon a bit softer.

Introduces a `.c-icon--circle` class for icons that should be surrounded by a circle.
Screenshot 2023-04-25 at 18 52 11


nmattia (approved):
LGTM, thanks! not a designer but it looks a little cramped in that circle

PR #1514, opened 2023-04-26
Use import.env instead of process.env

peterpeterparker:
# Motivation

Vite exposes env variables on the special import.meta.env object.

Documentation: https://vitejs.dev/guide/env-and-mode.html

# TODOs in another PR

Refactoring: Integrate `version.ts` into `environment.ts`

# Changes

- replace `process.env` with `import.meta.env`
- mock env for jest test
- set `II_` prefix in vite config
- remove vite `define` for `process.env`


src/frontend/test-setup.ts
```diff
@@ -6,6 +6,20 @@ jest.mock("./src/environment.ts", () => ({
...
+  DUMMY_CAPTCHA: false,
+}));
+
+jest.mock("./src/version.ts", () => ({
+  version: {
+    commit: "",
+    release: undefined,
+    dirty: false,
+  },
+}));
```
nmattia, line 21: shouldn't all env variables be read in `environment.ts`?

nmattia (commented):
In general LGTM, just one question

src/frontend/test-setup.ts
peterpeterparker: See PR description.

nmattia: Do you mean this?

> Refactoring: Integrate version.ts into environment.ts

I wasn't quite sure what it was referring to. Will this include `feature.ts` too?

peterpeterparker: I was not planing to move `features.ts`, thought you like it that way. No strong opinion, we can move these as well, your call.

nmattia: Ah sorry, I think we can keep `features.ts` but would make sense to import the environment values from `environment.ts`. Otherwise we'd be pulling in env vars from two different places (and it would break if any test imports `features.ts` (transitively))

nmattia (approved):

PR #1515, opened 2023-04-26
Don't record dev origins for stats

frederikrothenberger:
II keeps a list of latest front-end origins that are used with II devices. We are not interested in dev setups, so these should be excluded.


nmattia (approved):
In general LGTM, but can you clarify the motivation a bit? Is this to address irrelevant data from people using the production II for local testing?

frederikrothenberger:
> In general LGTM, but can you clarify the motivation a bit? Is this to address irrelevant data from people using the production II for local testing?

Yes, exactly. Local dev environments clutter our data that we collect on mainnet.

PR #1519, opened 2023-04-27
Autoupdate dapps file

nmattia:
This adds a daily job that pulls in the latest dapps list from the portal: [dfinity/portal/showcase.json](https://github.com/dfinity/portal/blob/master/showcase.json)

The first commit only runs the update. The resulting file is smaller than the current one because unnecessary fields are dropped (although some dapps are added). A field name is also changed (link -> website)

The second commit introduces the autoupdate workflow, based on our other autoupdate workflows.

### DIFF
--- a/.github/workflows/update-dapps.yml
+++ b/.github/workflows/update-dapps.yml
@@ -1,0 +1,50 @@
+# A GitHub Actions workflow that regularly creates a pull request to update dapps list
+name: Dapps Update
+
+on:
+  schedule:
+    # create a new pull request every day
+    - cron:  '0 0 * * *'
+
+jobs:
+  dapps-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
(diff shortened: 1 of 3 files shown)


frederikrothenberger:
> The first commit only runs the update. The resulting file is smaller than the current one because unnecessary fields are dropped (although some dapps are added). A field name is also changed (link -> website)

So dapps are dropped or added due to inaccurate information on the `usesInternetIdentity` flag. (i.e. `kinic` gets dropped but shouldn't, `Sand Bless` gets added but shouldn't).

But if the portal file gets updated (and maintained better) then this is definitely the way to go.

peterpeterparker:
> So dapps are dropped or added due to inaccurate information on the `usesInternetIdentity` flag. 

I like the idea! Portal will need a PR though, most dapps (Dscvr, Taggr, Distrikt, OC etc.) have not the flag set currently.

nmattia:
Superseded by #1634

PR #1520, opened 2023-04-27
Revert "Use import.env instead of process.env (#1514)"

nmattia:
This reverts commit 313615f4fb38d7d1ef98199f08b9f859ac839d60.

This commit introduced reproducibility issues in the build.


src/frontend/src/features.ts
```diff
@@ -2,9 +2,9 @@
 // from the `features` object below. This file also contains helper functions
 // for displaying a banner if features are enabled.
 export const features = {
-  FETCH_ROOT_KEY: import.meta.env.II_FETCH_ROOT_KEY === "1",
-  DUMMY_AUTH: import.meta.env.II_DUMMY_AUTH === "1",
-  DUMMY_CAPTCHA: import.meta.env.II_DUMMY_CAPTCHA === "1",
+  FETCH_ROOT_KEY: process.env.II_FETCH_ROOT_KEY === "1",
```
peterpeterparker, line 5: before rushing can you try first to set default values?

That was my plan but since the PR build was OK I did had to try out.

```
FETCH_ROOT_KEY: `${process.env.II_FETCH_ROOT_KEY ?? "0"}`=== "1",
etc.
```

frederikrothenberger (approved):
@nmattia asked me to rubber stamp

src/frontend/src/features.ts
nmattia: https://github.com/dfinity/internet-identity/pull/1521

PR #1524, opened 2023-04-28
Fix instructions for adding device

nmattia:
We've recently defaulted "Add device" to "browser", meaning the user does not have to select it explicitly anymore. This updates the instructions to reflect that.


frederikrothenberger (approved):
Thanks!

PR #1525, opened 2023-04-28
Eagerly load images before screenshots

nmattia:
This ensures images are loaded in the showcase by changing all `loading=lazy` to `loading=eager` and then iterating the images until all are `complete`.


frederikrothenberger (approved):
Thanks!

PR #1528, opened 2023-04-28
Decouple captcha solving from WebAuthn register

frederikrothenberger:
This PR adds a new canister call to submit captcha solutions. The caller principal will automatically be entitled to call register _without_ having to submit another captcha solution for 10 minutes (which is the default front-end session duration).

This mechanism replaces the previous temp_key mechanism where the temp_key could be submitted alongside the captcha solution on register.

### DIFF
--- a/src/internet_identity/src/state/temp_keys.rs
+++ b/src/internet_identity/src/state/temp_keys.rs
@@ -1,4 +1,4 @@
-use crate::MINUTE_NS;
+use crate::{state, MINUTE_NS};
 use candid::Principal;
 use ic_cdk::api::time;
 use internet_identity_interface::internet_identity::types::{AnchorNumber, DeviceKey, Timestamp};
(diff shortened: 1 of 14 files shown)


src/internet_identity/src/state/temp_keys.rs
```diff
@@ -25,56 +39,98 @@ pub struct TempKeys {
...
                 // Only remove temp key if the anchor matches
-                self.temp_keys.remove(device_key);
+                self.device_bound_keys.remove(device_key);
             }
         }
     }
 
+    /// Checks that the temporary key is a valid onboarding key (i.e. has solved the captcha).
+    ///
+    /// Requires a mutable reference because it does amortized clean-up of expired temp keys.
```
nmattia, line 109: Would it make sense to lift the pruning out then? The function would be less surprising and you wouldn't need mutability

src/internet_identity/src/state/temp_keys.rs
```diff
@@ -25,56 +39,98 @@ pub struct TempKeys {
...
+
+    pub fn add_device_bound_temp_key(
         &mut self,
         device_key: &DeviceKey,
         anchor: AnchorNumber,
         temp_key: Principal,
     ) {
-        let tmp_key = TempKey {
+        // Remove the temp key from onboarding keys if it exists
+        self.onboarding_keys.remove(&temp_key);
```
nmattia, line 80: Is it a bug if the key exists? Or maybe will be a bug if the key does not exist once we fully deprecate the API?

frederikrothenberger: No. Later on, it will become a bug. Each temp key will need to have submitted a captcha solution (i.e. be in the list of onboarding keys) to be transferred to a device bound key.

frederikrothenberger:
Made obsolete by changing requirements.

PR #1529, opened 2023-04-28
Improve WebAuthn error messages when enrolling devices on recovery

frederikrothenberger:
This PR extracts the copy for the duplicate device message and introduces the nicer error messages also when recovering.


src/frontend/src/utils/webAuthnErrorUtils.ts
```diff
@@ -1,15 +1,26 @@
...
   return displayError({
-    ...webAuthnCancelTemplate(),
+    ...webAuthnErrorCancelCopy(),
+    primaryButton,
+  });
+};
+
+export const displayDuplicateDeviceError = (
+  primaryButton: string
+): Promise<void> => {
```
nmattia, line 20: How about accepting an object here? i.e. `{ primaryButton: string }`

Do you think the intent would be clearer?

```ts
       await displayDuplicateDeviceError("Back to manage");
       await displayDuplicateDeviceError({ primaryButton: "Back to manage"});
```

src/frontend/src/utils/webAuthnErrorUtils.ts
```diff
@@ -1,15 +1,26 @@
 import { displayError } from "$src/components/displayError";
 import { I18n } from "../i18n";
-import copyJson from "./webAuthnErrorUtils.json";
+import cancelCopy from "./webAuthnErrorCancel.json";
```
nmattia, line 3: So far we've called the json object `copyJson` and had the copy file have the same name as the file importing it (modulo the extension). Do you think it makes sense to do the same here to stay consistent?

If the file becomes too big we can always split it later

src/frontend/src/utils/flowResult.ts
```diff
@@ -94,7 +94,7 @@ export const apiResultToLoginFlowResult = (
     case "cancelOrTimeout": {
       return {
         tag: "err",
-        ...webAuthnCancelTemplate(),
+        ...webAuthnErrorCancelCopy(),
```
nmattia, line 97: something is a little odd, since this is probably end up being display somewhere where `displayCancelError` could be used instead; happy to fix this up later though

nmattia (approved):
Thanks!

src/frontend/src/utils/webAuthnErrorUtils.ts
frederikrothenberger: Done. But now we can no longer use the spread operator. At some point it would be nice to have nested copy. I.e.
```json
{
  "en": {
    "foo": {
      "title": "..."
    },
    "bar": {
      "title": "..."
    }
  }
}

PR #1533, opened 2023-04-29
Use nonNullish for not undefined comparison

peterpeterparker:
# Motivation

As for `isNullish`, refactor not undefined comparison with `nonNullish` utils.


src/frontend/src/utils/userNumber.test.ts
```diff
@@ -90,11 +91,11 @@ function testLocalStorage(
 ) {
   test(name, () => {
     localStorage.clear();
-    if (opts?.before !== undefined) {
+    if (nonNullish(opts) && nonNullish(opts?.before)) {
```
nmattia, line 94: why is the first check necessary?

peterpeterparker: Capture d’écran 2023-04-29 à 11 30 29

src/frontend/src/utils/userNumber.test.ts
```diff
@@ -90,11 +91,11 @@ function testLocalStorage(
...
       setLocalStorage(opts.before);
```
nmattia, line 95:
```suggestion
    const before = opts?.before;
    if (nonNullish(before)) {
      setLocalStorage(before);
```


what do you think about this? I think it makes more sense

nmattia (approved):

src/frontend/src/utils/userNumber.test.ts
```diff
@@ -90,11 +91,12 @@ function testLocalStorage(
...
     localStorage.clear();
-    if (opts?.before !== undefined) {
-      setLocalStorage(opts.before);
+    const before = opts?.before;
+    if (nonNullish(before)) {
+      setLocalStorage(before);
     }
     fn();
-    if (opts?.after !== undefined) {
+    if (nonNullish(opts) && nonNullish(opts?.after)) {
```
nmattia, line 99: would it make sense to do the same for `after` as we do with `before`?

peterpeterparker: sure go for it

PR #1537, opened 2023-05-02
Add 'Save Passkey' page

nmattia:
This introduces a new page, "Save Passkey", that prompts the user for creating a WebAuthn identity, while giving some information on what a "passkey" is.

This also extracts the "rendering" bits of `constructIdentity`, such that `constructIdentity` now only focuses on the WebAuthn interaction, and the caller is responsible for e.g. showing a loading screen.


src/frontend/src/flows/register/passkey.ts
```diff
@@ -0,0 +1,94 @@
...
+      <p class="t-paragraph">
+        ${copy.select} <strong class="t-strong">${copy.save_passkey}</strong> ${
+    copy.and_complete_prompts
+  }</p>
+    </hgroup>
+    <button @click=${() =>
+      construct()} data-action="construct-identity" class="c-button">${
+    copy.save_passkey
+  }</button>
+    <p class="t-paragraph">
```
meodai, line 36: a paragraph should only contain text. I would wrap the whole "copy" part of this screen in a `section.c-marketing-block` and make `aside.l-stack` with  the different section not paragraphs.

meodai: instead of `<strong class="t-strong">` you can just use a h3 or something with the `t-title` class

src/frontend/src/flows/register/passkey.ts
```diff
@@ -0,0 +1,94 @@
...
+import { unknownToString } from "$src/utils/utils";
+import { constructIdentity } from "$src/utils/webAuthn";
+import { isCancel, webAuthnErrorCopy } from "$src/utils/webAuthnErrorUtils";
+import { html, TemplateResult } from "lit-html";
+
+import copyJson from "./passkey.json";
+
+/* Anchor construction component (for creating WebAuthn credentials) */
+
+const savePasskeyTemplate = ({
```
frederikrothenberger, line 16: Shouldn't this screen have a `cancel` button of some sort?

src/frontend/src/flows/register/passkey.ts
nmattia: Thanks a lot!

PR #1538, opened 2023-05-02
Use module aliases in styleguide

nmattia:
This updates the styleguide to use module aliases like the showcase & the app itself.


meodai (approved):
LGTM!

PR #1540, opened 2023-05-02
Don't show recovery warning on anchor creation

nmattia:
This changes the anchor creation/registration flow to _not_ show a warning in case of single device, no recovery, etc; but instead only shows those warning for returning users.


frederikrothenberger (approved):
LGTM, thanks

PR #1541, opened 2023-05-03
Simplify chasm implementation

nmattia:
This simplifes the chasm in a few ways:

* The logic for toggling the chasm open/closed is rewritten using `Chan`s instead of raw lit `Ref`s
* The CSS is changed to use `grid` to avoid having to arbitrary set the `max-height`


meodai (approved):
LGTM

PR #1544, opened 2023-05-04
Make footer sticky

nmattia:
This redesigns the foot by only having links to GitHub & the support page, and by making the footer sticky.


meodai:
@nmattia I cleaned the CSS up a bit, mainly by using tokens. In the description you say something about it being "sticky", what do you mean by that? (I ask because its just on the bottom of the page in the branch I checked out)

meodai:
@nmattia Never mind I understood what was meant by "sticky" as soon as I saw it on sorter pages. Added some space to the bottom of the container to make sure it does not stick to the footer too much + I have separated the footer layout block from the footer component

meodai (approved):
LGTM

PR #1545, opened 2023-05-04
Show dapps marquee on landing screens

nmattia:
This shows the marquee component on the landing (manage & authz) screens.


peterpeterparker:
Randomly had a look to the screenshots of this PR. 
In that formThis will be an issue for developers using II sign-in through agent-js in modals.

e.g.

Capture d’écran 2023-05-04 à 11 18 27

nmattia:
@peterpeterparker probably, though pop-ups are not officially supported; in particular the experience is terrible on Safari. Good to know though, I'll relay the feedback.

peterpeterparker:
@nmattia if the feature is officially supported in agent-js, I can assume the function is official no? The experience in Safari is fine also btw. give it a try.

I was also thinking, in addition to the modal, which to some extension is not that important, displaying the dapps explorer in the sign-in process from a dapp feels also like a threat for the user acquisition rate of the dapps on the IC in my opinion. 

e.g.. if a client (dscvr, etc.) uses the II sign-in, the client most probably want to the user to sign-in not click on another app in II instead of proceeding with sign-in.

For both reasons, I think it's better to not display the dapps explorer within the sign-in flow coming from a client.

meodai (approved):
@nmattia LGTM did a few CSS tweaks

PR #1547, opened 2023-05-04
Allow loading font from canister

nmattia:
This changes the asset handler to add support for `woff2` files and generalizes gzip support for all file formats, not just `.js`.


frederikrothenberger (approved):
LGTM, thanks!

PR #1551, opened 2023-05-05
Ensure Create Passkey CTA is in view

nmattia:
This adds an automatic scroll when the "Create Passkey" page is shown to ensure the CTA is in view. On some devices it may not always be visible.

### DIFF
--- a/src/frontend/src/flows/register/passkey.ts
+++ b/src/frontend/src/flows/register/passkey.ts
@@ -3,7 +3,7 @@ import { mainWindow } from "$src/components/mainWindow";
 import { toast } from "$src/components/toast";
 import { I18n } from "$src/i18n";
 import { IIWebAuthnIdentity } from "$src/utils/iiConnection";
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/register/passkey.ts
```diff
@@ -17,14 +17,17 @@ const savePasskeyTemplate = ({
   construct,
   i18n,
   cancel,
+  scrollTo = false,
 }: {
   construct: () => void;
   i18n: I18n;
   cancel: () => void;
+  /* put the CTA into view */
```
frederikrothenberger, line 25: What is "CTA"?

frederikrothenberger (approved):

src/frontend/src/flows/register/passkey.ts
nmattia: [Call-to-Action](https://en.wikipedia.org/wiki/Call_to_action_(marketing))

PR #1553, opened 2023-05-05
Align copy with new wording

nmattia:
This updates the copy to use the new official wording, in particular Anchor -> Internet Identity & Device -> Passkey


frederikrothenberger (approved):
Things I noticed:
* Sometimes the layout is now more narrow (i.e. in the screnshot screenshots/desktop/authorizePick.png). Is this intended?
* the test banner still uses the word anchor
* the error message on recover when no recovery device exists still uses the word anchor
* the `register` method in iiConnection logs messages that contain the word anchor.
* the duplicate error message in webAuthnErrorUtils.json should probably also be reworded to better fit the new passkey terminology.

nmattia:
Thanks for the thorough review!

> Sometimes the layout is now more narrow (i.e. in the screnshot screenshots/desktop/authorizePick.png). Is this intended?

Jein, we've decided to not fiddle with this too much for now
 
> the test banner still uses the word anchor
> the error message on recover when no recovery device exists still uses the word anchor
> the register method in iiConnection logs messages that contain the word anchor.
> the duplicate error message in webAuthnErrorUtils.json should probably also be reworded to better fit the new passkey terminology.

Thanks, will adapt!

PR #1554, opened 2023-05-05
Allow renaming devices

nmattia:
This adds an option in the management page for renaming authenticator devices.


frederikrothenberger (approved):
Nice!
But we should really introduce nicer prompts instead of using alerts...

PR #1555, opened 2023-05-05
Fix URL autofill for local deployments

frederikrothenberger:
In addition, it removes a no longer used URL fallback to `https://<canister_id>.dfinity.network`.


nmattia (approved):
This makes sense, thanks! I would still leave the `"local"` branch just to be explicit, even if we default to the same value. It makes it clearer that the value is used for local development, and that any value other than `local` or `ic` for `DFX_NETWORK` is not actively supported but will just use a fallback. Wdyt?

PR #1560, opened 2023-05-08
Catch prepareDelegation errors

nmattia:
This ensures canister errors when preparing the delegation are caught and an error message is shown to the user.


src/frontend/src/flows/authorize/fetchDelegation.ts
```diff
@@ -36,12 +36,18 @@ export const fetchDelegation = async (
...
     authContext.authRequest.maxTimeToLive
   );
 
+  if ("error" in result) {
+    return result;
+  }
+
+  const [userKey, timestamp] = result;
+
   const signed_delegation = await retryGetDelegation(
```
frederikrothenberger, line 51: Shouldn't there be some try catch around `get_delegation` too?

frederikrothenberger (approved):

PR #1561, opened 2023-05-08
Align Identity card with dapps explorer

nmattia:
The "Internet Identity" image card uses negative borders to make up for its container's padding; however the wrong padding values were used. This makes sure the correct margins are used to ensure the cards are aligned on the management page.

Before:

Screenshot 2023-05-08 at 10 28 15

After:

Screenshot 2023-05-08 at 10 28 25


frederikrothenberger (approved):
Unrelated, but the mobile screenshots are off. Is this to do with min-width exceeding the mobile screen width? We should probably switch to a more realistic mobile device resolution...

nmattia:
> Unrelated, but the mobile screenshots are off. Is this to do with min-width exceeding the mobile screen width? We should probably switch to a more realistic mobile device resolution...

Yeah I know, spotted this here too. I'll have a look into this but I suspect it's related to last-minute changes we made last week to fix the container width on desktop. I think the resolution is somewhat realistic (iPhone SE IIRC)

frederikrothenberger:
> > Unrelated, but the mobile screenshots are off. Is this to do with min-width exceeding the mobile screen width? We should probably switch to a more realistic mobile device resolution...
> 
> Yeah I know, spotted this here too. I'll have a look into this but I suspect it's related to last-minute changes we made last week to fix the container width on desktop. I think the resolution is somewhat realistic (iPhone SE IIRC)

But is it actually broken on an iPhone SE now?

nmattia:
> But is it actually broken on an iPhone SE now?

According to Chrome's mobile emulator, it is not; might be because of different pixel densities. Will need to look into it to get a better idea of what's happening.

PR #1562, opened 2023-05-08
Show Passkey alias when renaming

nmattia:
This ensures the current Passkey's alias is shown and set as a default when the user renames their Passkey. This gives the user context about which Passkey they're renaming, and improves the flow (since renaming is often just slightly tweaking the current alias).

https://user-images.githubusercontent.com/6930756/236777144-3a96ea2d-7f3a-40d4-898a-b93d8b161cf8.mov


frederikrothenberger (approved):
Thanks!

PR #1565, opened 2023-05-08
Infer more correct alias for Chrome

nmattia:
This updates the Passkey alias inferrence to be less specific but more correct on Chrome:

> Chrome has a concept of shared "Passkeys". If the user is signed in (to Chrome), then Chrome will create a
> Passkey shared across all the users' Chromes; if the user is _not_ signed in, then it will be local. We haven't
> found a way to figure out if the generated Passkey is shared or not, so to be safe we just say "Chrome" as a tradeoff
> between "Chrome on [OS]" (local) and "Chrome Passkey" (shared).

### DIFF
--- a/src/frontend/src/flows/register/index.ts
+++ b/src/frontend/src/flows/register/index.ts
@@ -126,10 +126,17 @@ export const inferAlias = async ({
     return FIDO;
   }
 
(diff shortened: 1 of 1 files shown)


frederikrothenberger (approved):
LGTM!

PR #1566, opened 2023-05-08
Add storage migration to MemoryManager (v6 to v7)

przydatek:
Add functionality to migrate storage from v6 to v7, where v7 uses `ic_stable_structures::MemoryManager`. The migration does not copy the entire data, but rather uses 2nd page of the memory (which was unused/reserved so far) for `MemoryManager`'s metadata.  The anchor-memory starts in both versions from the 3rd page.
The migration is not enabled yet, and will be controlled by a flag added in a subsequent MR.
(GIX-1423)


src/internet_identity/src/storage.rs
```diff
@@ -396,6 +425,79 @@ impl<M: Memory + Clone> Storage<M> {
         }
     }
 
+    #[allow(dead_code)]
+    pub fn from_memory_v6_to_v7(memory: M) -> Option<Self> {
+        let maybe_storage_v6 = Self::from_memory(memory.clone());
+        maybe_storage_v6.as_ref()?;
```
frederikrothenberger, line 431: What is the purpose of this line?

src/internet_identity/src/storage.rs
```diff
@@ -396,6 +425,79 @@ impl<M: Memory + Clone> Storage<M> {
...
+        let storage_v6 = maybe_storage_v6.unwrap();
```
frederikrothenberger, line 432: It would be nicer to return `None` on empty rather than panicking, no?

src/internet_identity/src/storage.rs
```diff
@@ -396,6 +425,79 @@ impl<M: Memory + Clone> Storage<M> {
...
+                std::mem::size_of::<mm::Header>(),
+            )
+        };
+        let mut mm_header_memory = RestrictedMemory::new(memory.clone(), 1..2);
+        let mut writer = Writer::new(&mut mm_header_memory, 0);
+        writer
+            .write_all(mm_header_bytes)
+            .expect("bug: failed to grow memory");
+        // Update bucket-to-memory assignments.
+        let buckets_offset: u64 = 8 + 32 + 255 * 8;
```
frederikrothenberger: Didn't you want to add some comments here?

src/internet_identity/src/storage.rs
```diff
@@ -396,6 +425,100 @@ impl<M: Memory + Clone> Storage<M> {
...
+        // --------------------------------------------------
+        // ...
+        // --------------------------------------------------
+        // Size of memory 254 (in pages)         ↕ 8 bytes
+        // -------------------------------------------------- <- Bucket allocations
+        // ...
+        let buckets_offset: u64 = (3 + 1 + 2 + 2) + 32 + (255 * 8);
+        let mut writer = Writer::new(&mut mm_header_memory, buckets_offset);
+        let mut bucket_to_memory = [mm::UNALLOCATED_BUCKET_MARKER; mm::MAX_NUM_BUCKETS as usize];
+        for i in 0..num_allocated_buckets {
```
frederikrothenberger, line 512: You could avoid the loop by using `splice`:
```
let mut bucket_to_memory = vec![mm::UNALLOCATED_BUCKET_MARKER; mm::MAX_NUM_BUCKETS as usize];
bucket_to_memory.splice(
    0..num_allocated_buckets as usize,
    vec![0; num_allocated_buckets as usize],
);
```

Requires using vec though, so up to you.

src/internet_identity/src/storage.rs
```diff
@@ -311,7 +338,9 @@ impl<M: Memory + Clone> Storage<M> {
             header_memory,
             anchor_memory,
             maybe_memory_manager,
-        }
+        };
+        storage.flush();
```
frederikrothenberger, line 342: Why is this flush now necessary?

src/internet_identity/src/storage/tests.rs
```diff
@@ -80,6 +82,112 @@ fn should_recover_header_from_memory_v7() {
     assert_eq!(storage.version(), 7);
 }
 
+fn add_test_anchor_data<M: Memory + Clone>(storage: &mut Storage<M>, number_of_anchors: usize) {
+    for _ in 0..number_of_anchors {
+        let (anchor_number, mut anchor) = storage
+            .allocate_anchor()
+            .expect("Failure allocating an anchor.");
+        anchor
+            .add_device(sample_device())
```
frederikrothenberger, line 91: It would be nice to make the devices different (i.e. depend on the index) because this would detect mix-ups in the data.

src/internet_identity/src/storage/tests.rs
```diff
@@ -80,6 +82,112 @@ fn should_recover_header_from_memory_v7() {
...
+        storage.write(anchor_number, anchor.clone()).unwrap();
+    }
+}
+
+fn range_as_hex(memory: &RefCell<Vec<u8>>, offset: u64, length: usize) -> String {
+    let mut buf = vec![0u8; length];
+    memory.read(offset, &mut buf);
+    let s = hex::encode(buf);
+    let mut answer = String::new();
+    for (i, c) in s.chars().enumerate() {
```
frederikrothenberger, line 102: Hm, I stumbled on this loop because it was not immediately obvious to me that you also do formatting here.
Either add a comment or drop the formatting entirely.

src/internet_identity/src/storage.rs
przydatek: So that a new, empty storage (without any anchors) is properly initialised.

src/internet_identity/src/storage.rs
przydatek: It checks whether `maybe_storage_v6` is `Some` yet without unwrapping (and returns `None` immediately if the value is `None`)

src/internet_identity/src/storage.rs
przydatek: It will not panic, due to the check on the previous line.
But I noticed now that both lines can be combined, so I did that, and hopefully it is more clear now. Also, I added a test for the case when `maybe_storage_v6` is `None`.

src/internet_identity/src/storage.rs
przydatek: Thanks for pointing it out, I'll leave as-is though, ok?

src/internet_identity/src/storage/tests.rs
przydatek: Added a comment (with the formatting it is easier to debug).

src/internet_identity/src/storage/tests.rs
przydatek: Good idea, done.

przydatek (commented):
Thanks for the review @frederikrothenberger, PTAL.

src/internet_identity/src/storage.rs
frederikrothenberger: TIL! Thanks, I thought it only worked on `Result`.

src/internet_identity/src/storage.rs
frederikrothenberger: So up to now, the `storage` was initialized only half-way and then `flush`ed later. Definitely a good change. 👍

frederikrothenberger (approved):
LGTM, thanks!

PR #1567, opened 2023-05-08
Preload Identity card background

nmattia:
The "Internet Identity" card background is an image that must be loaded from the canister, which can take a few seconds. This ensures that when used in the identity creation flow, the browser can already start fetching the image as the user is dealing with the CAPTCHA (optimistically assume the user will succeed, at very little cost if the user does not actually succeed).

Before, with network throttling (2s/request):

https://user-images.githubusercontent.com/6930756/236805453-4826804f-4101-4dd5-8011-2bf60104794f.mov

After, with network throttling (2s/request):

https://user-images.githubusercontent.com/6930756/236805523-b54f7b55-e444-47b7-8000-39898684c4ab.mov

### DIFF
--- a/src/frontend/src/flows/register/finish.ts
+++ b/src/frontend/src/flows/register/finish.ts
@@ -95,7 +95,7 @@ export const displayUserNumberTemplate = ({
 };
 
 // A non-exported wrapper around HTMLImageElement, to ensure this exact image is loaded
-// (a no other image is passed as an argument to this page)
+// (and no other image is passed as an argument to this page)
 class IdentityBackground {
   public img: HTMLImageElement;
 
--- a/src/frontend/src/flows/register/finish.ts
+++ b/src/frontend/src/flows/register/finish.ts
@@ -9,12 +9,14 @@ import { createRef, ref, Ref } from "lit-html/directives/ref.js";
 export const displayUserNumberTemplate = ({
   onContinue,
   userNumber,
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/register/finish.ts
```diff
@@ -92,10 +94,56 @@ export const displayUserNumberTemplate = ({
   });
 };
 
+// A non-exported wrapper around HTMLImageElement, to ensure this exact image is loaded
+// (a no other image is passed as an argument to this page)
```
nmattia, line 98:
```suggestion
// (and no other image is passed as an argument to this page)
```

src/frontend/src/flows/register/finish.ts
```diff
@@ -92,10 +94,56 @@ export const displayUserNumberTemplate = ({
...
+    img.classList.add("c-input--id__art");
+    img.alt = "";
+    this.img = img;
+  }
+}
+
 export const displayUserNumberPage = renderPage(displayUserNumberTemplate);
 
-export const displayUserNumber = (userNumber: bigint): Promise<void> => {
+// Omit specified functions parameters, for instance OmitParams<..., "foo" | "bar">
```
frederikrothenberger, line 113: Technically, it omits properties from the first argument rather than parameters, right?

frederikrothenberger (approved):
Thanks!

PR #1575, opened 2023-05-09
Fix empty recovery clipboard on Firefox

peterpeterparker:
# Motivation

Parsing the recovery phrase does not work in Firefox, only the first input gets populated.

# Changes

- workaround: set again the same data to the clipboard one the event is created


src/frontend/src/flows/recovery/recoverWith/phrase.ts
```diff
@@ -227,12 +227,18 @@ export const wordTemplate = ({
           if (rest.length <= 0) {
             return;
           }
+
```
nmattia, line 230:
```suggestion
```

src/frontend/src/flows/recovery/recoverWith/phrase.ts
```diff
@@ -227,12 +227,18 @@ export const wordTemplate = ({
...
           const newData = new DataTransfer();
           newData.setData("text", rest.join(" "));
+
```
nmattia, line 233:
```suggestion
```

nmattia (approved):
LGTM, thanks!

PR #1578, opened 2023-05-09
Add a flag to control storage migration v6 to v7.

przydatek:
Add a flag to `InternetIdentityInit` to control the storage layout migration from v6 to v7 (with `MemoryManager`).
GIX-1424

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -232,7 +232,6 @@ impl<M: Memory> AnchorMemory<M> {
 
 pub enum StableMemory<M: Memory> {
     Single(M),
-    #[allow(dead_code)]
     Managed(M),
 }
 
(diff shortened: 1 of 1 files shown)


src/internet_identity/src/main.rs
```diff
@@ -328,10 +328,18 @@ fn acknowledge_entries(sequence_number: u64) {
...
+    }
+    let arg = maybe_arg.as_ref().unwrap();
+    arg.migrate_storage_to_memory_manager.unwrap_or(false)
+}
+
 #[init]
 fn init(maybe_arg: Option<InternetIdentityInit>) {
     init_assets();
-    state::init_new();
+    state::init_new(migrate_to_memory_manager(&maybe_arg));
```
frederikrothenberger, line 342: Does `init_new` need the option to initialize as `v6`? I think it's fine to initialize as v7 now.

frederikrothenberger (approved):
You should be able to remove the `#[allow(dead_code)]` attributes, right?

src/internet_identity/src/main.rs
przydatek: As discussed offline, keeping as-is, for consistency.  The flag and this code will be removed after migration is done.

przydatek (commented):
Thanks for the review!
I've removed all `dead_code`-annotations, except for one which is still needed.
Will proceed with merging once CI is green.

PR #1581, opened 2023-05-10
II interface v2

frederikrothenberger:
Strawman PR to discuss a possible v2 II interface. The interface introduces the following changes:
* remove any mention of anchor. The new API is based around "identity". (I decided against "user", as a user could have many identities.)
* remove the concept of "device" (the new API is based around keys)
* by doing the above two renames, all methods now have a new name that does not clash with the v1 api (except for `create_challenge` which has been renamed to `create_captcha`). This avoids having to introduce a `_v2` suffix (or similar).
* all changed methods now return `opt variant`, which allows to properly return errors (and also extend the types later without breaking clients)
* `update` is now clearly scoped to only allow changing metadata
* the `last_usage` timestamp can now be provided on write. The write will be rejected if the timestamp does not match the canister side value. This can be used to prevent writes of stale data.
* Some redundant type aliases have been removed.
* No changes have been made to the endpoints related to delegations as well as the non user-facing the technical ones (i.e. stats and archive related endpoints).

The Plan is to gradually introduce the new API alongside the old one in an experimental state. Once we are happy with the new interface, we can stabilize the API and deprecate (and later remove) the current one.

This work is motivated by repeatedly bumping into limitations with the current API, the latest two examples being the inability to introduce proper errors and the need to introduce additional WebAuthnSignatures (which causes problems on the current `update` call as it mixes sensitive and non-sensitive data changes).

This PR is _not_ intended to be merged, but rather to form agreement on the new API, so that we can then introduce it gradually as needed in separate PRs.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -1,8 +1,6 @@
-type UserNumber = nat64;
+type IdentityNumber = nat64;
 type PublicKey = blob;
 type CredentialId = blob;
-type DeviceKey = PublicKey;
-type UserKey = PublicKey;
 type SessionKey = PublicKey;
 type FrontendHostname = text;
 type Timestamp = nat64;
@@ -41,22 +39,10 @@ type StreamingStrategy = variant {
     };
 };
 
-type Purpose = variant {
-    recovery;
-    authentication;
-};
-
-type KeyType = variant {
-    unknown;
-    platform;
-    cross_platform;
-    seed_phrase;
-};
-
-// This describes whether a device is "protected" or not.
-// When protected, a device can only be updated or removed if the
-// user is authenticated with that very device.
-type DeviceProtection = variant {
+// This describes whether a key is "protected" or not.
+// When protected, a key can only be updated or removed if the
+// user is authenticated with that very key.
+type KeyProtection = variant {
     protected;
     unprotected;
 };
@@ -66,35 +52,6 @@ type Challenge = record {
     challenge_key: ChallengeKey;
 };
 
-type DeviceData = record {
-    pubkey : DeviceKey;
-    alias : text;
-    credential_id : opt CredentialId;
-    purpose: Purpose;
-    key_type: KeyType;
-    protection: DeviceProtection;
-    origin: opt text;
-    // Metadata map for additional device information.
-    //
-    // Note: some fields above will be moved to the metadata map in the future.
-    // All field names of `DeviceData` (such as 'alias', 'origin, etc.) are
-    // reserved and cannot be written.
-    metadata: opt MetadataMap;
-};
-
-// The same as `DeviceData` but with the `last_usage` field.
-// This field cannot be written, hence the separate type.
-type DeviceWithUsage = record {
-    pubkey : DeviceKey;
-    alias : text;
-    credential_id : opt CredentialId;
-    purpose: Purpose;
-    key_type: KeyType;
-    protection: DeviceProtection;
-    origin: opt text;
-    last_usage: opt Timestamp;
-    metadata: opt MetadataMap;
-};
 
 // Map with some variants for the value type.
 // Note, due to the Candid mapping this must be a tuple type thus we cannot name the fields `key` and `value`.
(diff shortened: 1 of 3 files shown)


frederikrothenberger:
> the last_usage timestamp can now be provided on write. The write will be rejected if the timestamp does not match the canister side value. This can be used to prevent writes of stale data.

Open point about this: While it is true that the last usage can be used to detect stale writes, it can give false positives. I.e. any authenticated action, such as reading data or modifying _other_ keys will update the `last_usage` timestamp, thus leading to rejects of non-stale data. To avoid this, we would need to introduce a second timestamp that updates on writes _only_ (e.g. `last_update`).

Any opinion on this @nmattia?

src/internet_identity/internet_identity.did
```diff
@@ -333,42 +272,112 @@ type DeployArchiveResult = variant {
...
+type result = variant {
+    ok;
+};
+
+type WebAuthnDirectSigMode = variant {
+    mandatory;
+    optional;
+};
+
+type key = variant {
```
przydatek, line 293: I would suggest using some other name than `key`, as "key" is a very generic and often overloaded term. For example in this specification it collides with `PublicKey` and `SessionKey`, which are not instances of `key` as defined in this line. 

Maybe something more specific to the actual functionality/abstraction would be more suitable, like `AuthnMethod` or `AuthnForm`?  I realise, that these are significantly longer than just "key" but they make it (more) clear what is represented by the structures.  As we move away from "anchor" to "identity", speaking of "identity's authentication method" is more clear than about "identity's key", no?

Another alternative could be a `pass`: an identity can have many "passes" (which allow the identity to authenticate itself).

src/internet_identity/internet_identity.did
```diff
@@ -333,42 +272,112 @@ type DeployArchiveResult = variant {
...
+    };
+};
+
+type KeyRecord = record {
+    key: key;
+    // contains now the following fields
+    // - alias
+    // - key_type: platform,
+    // - origin
+    // - key_type: reduced to "platform", "cross_platform" on migration
```
przydatek, line 311: nit: `key_type` comes up twice on this list.

src/internet_identity/internet_identity.did
przydatek: I find `auth` better than `key`, but my current preference would be `pass`:  the short for authentication is usually `authn`, like in `WebAuthn` (also to distinguish it from authorization, `authz`), so `auth` is a bit ambiguous, and even if we assume it refers to authentication, the connection to "method" is not obvious.  We could introduce e.g. `authm` for "authentication method" to disambiguate, but that's not perfect either...  
I like `pass` because it is short and describes "something that can be used to authenticate oneself", which I believe is what we are dealing with here, so we get a nice abstraction.
Maybe let's talk briefly together with @nmattia and settle on some naming (say this afternoon?), to avoid repeated renamings?

frederikrothenberger: I discussed with @nmattia yesterday:
* he suggested that we could also go for `passkey`, but since `passkeys` are strongly associated with WebAuthn (at least for now) I would like to go with something that conveys a more general meaning.
* he suggested also to change the method names by using prefixes in lieu of REST URL path segments: i.e. do `pass_update_settings`, `pass_add`, ... instead of of `add_pass`, `update_pass_settings`
* I'd be fine with using `pass` although I've have not seen it used so far as a synonym for "generic authentication method". I have slight preference for `authn` but I'm fine with `pass` too.

frederikrothenberger:
Quick summary of the discussion @przydatek, @nmattia and I had today:

* Renaming `anchor` -> `identity`
* Renaming `device_data` -> `authn_data`
  * The actual authentication mechanism (WebAuthn / generic pubkey) will be called `authn_method`
* Methods will be renamed to have matching prefixes (i.e `authn_data_add`, `authn_data_remove`, ...)
* The term `tentative` will be dropped in favour of `authn_data_register` and `authn_data_confirm`

frederikrothenberger:
@nmattia: Update after a discussion I had with @przydatek:
We prefer to _not_ remove the `purpose` and split the authn_methods in two different list. The reasons are:
* an `AuthnMethodData` would no longer be self-describing, but the semantics become dependent on the context
* it would duplicate the `add` method
* it would require a new method to switch a device from authn to recovery (or vice versa)

frederikrothenberger:
#2162 marks the last PR of an early implementation of the new API that has feature parity with the legacy API. More development is necessary, but at this point I don't think keeping around this draft is useful anymore.

PR #1583, opened 2023-05-11
Update comment for dropdown trigger in e2e test

peterpeterparker:
# Motivation

Update comment for dropdown trigger in e2e test as it requires now a click (no more hover).


nmattia (approved):
Thanks!

PR #1584, opened 2023-05-11
Wait for alert being open before accepting it

peterpeterparker:
# Motivation

Test are currently flaky. They work locally but often fails in the CI. By comparing the test I noticed that the two that fails are not awaiting for the alert to be open before clicking it on the contrary of the other tests that do await.


nmattia (approved):
Thanks!

PR #1585, opened 2023-05-11
Import dapps list in main bundle

nmattia:
This bundles the dapps in the script loaded on page load to avoid extra fetching (and extra latency if the first page to load needs information about the dapps).

The dapps list is also cleaned up to only include what we actually need.

Before:
dist/dapps.js.gz                            15.19kb / gzip: 5.30kb
dist/index.js.gz                            574.37kb / gzip: 189.43kb

After:
dist/index.js.gz                            578.50kb / gzip: 190.90kb


peterpeterparker (approved):
lgtm

PR #1588, opened 2023-05-12
Don't show single device warning

nmattia:
This removes the "single device warning" that we used to show (also incorrectly after the recovery phrase warning). The warning is less relevant now that most browsers offer Passkeys.


src/frontend/src/flows/recovery/recoveryWizard.ts
```diff
@@ -1,46 +1,17 @@
...
-  }
-};
-
-const recoveryWizardDefault = async (
-  userNumber: bigint,
-  connection: AuthenticatedConnection
 ): Promise<void> => {
   // Here, if the user doesn't have any recovery device, we prompt them to add
   // one. If after returning from the prompt they still haven't added one, then
   // we display a big warning with full explanation about the risks of having a
```
frederikrothenberger, line 11: nit: The comment is now outdated, since the second warning has been removed.

frederikrothenberger (approved):

src/frontend/src/flows/recovery/recoveryWizard.ts
nmattia: good catch, thanks!

PR #1591, opened 2023-05-12
Scroll to top of onboarding steps

nmattia:
This instructs the steps of the onboarding flow to always reset the scroll to the top, to ensure the title, explanation or CTA are always in view.

We've tried a couple of HTML & CSS fixes but nothing seems to work as reliably as this.


frederikrothenberger (approved):
Thanks! Seems to work great on FF.

PR #1592, opened 2023-05-12
Center CAPTCHA on page resize

nmattia:
On mobile devices, opening the virtual keyboard tends to push the CAPTCHA out of view. This ensures the CAPTCHA is recentered when the page is resized (e.g. when the virtual keyboard opens).

This is implemented via a window resize handler, which deregisters itself as soon as it can.

This also works on non-mobile although should never really kick-in.

### DIFF
--- a/src/frontend/src/flows/register/captcha.ts
+++ b/src/frontend/src/flows/register/captcha.ts
@@ -88,6 +88,9 @@ export const promptCaptchaTemplate = <T>({
   // The text input where the chars can be typed
   const input: Ref<HTMLInputElement> = createRef();
 
+  // The CAPTCHA container
+  const captchaContainer: Ref<HTMLElement> = createRef();
+
   // The error shown on bad input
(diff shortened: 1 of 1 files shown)


frederikrothenberger (approved):
Thanks! Seems to work fine. 👍 
FYI: Leads to slightly different outcomes on Chrome vs FF. But approved since the captcha is visible either way.

Chrome

Screenshot_2023-05-12-16-06-25-36_40deb401b9ffe8e1df2f1cc5ba480b12


FF



Screenshot_2023-05-12-16-06-35-02_3aea4af51f236e4932235fdada7d1643

nmattia:
@frederikrothenberger thanks for checking thoroughly! I think it's Firefox that does things a bit differently (we had the same issue on the "Create Passkey" page where Firefox would simply align the element with the top of the page whereas Chrome&Safari would first check whether any scrolling was necessary). Although here it seems to be the opposite 😂

PR #1596, opened 2023-05-15
Fix bug in archive fetch interval metric

frederikrothenberger:
This PR fixes a bug that caused II to expose the wrong data as the `internet_identity_archive_config_polling_interval` metric.

Also, the metric is adjusted to give the interval in seconds as is recommended for time durations for prometheus. Additionally, the test is changed to use different values at are more easily distinguished to avoid similar bugs in the future.


przydatek (approved):
LGTM, thanks!
nit: maybe extend the PR-description with a summary of what was the actual bug being fixed.

PR #1597, opened 2023-05-15
Fix centering on passkey & explorer pages

nmattia:
This removes a workaround to have the CTA in view in the passkey page, and adds the other, better workaround for scrolling to the top of pages to the dapps explorer.


meodai (approved):
LGTM

PR #1598, opened 2023-05-15
Catch error in get_delegation

nmattia:
This ensures errors (e.g. network failure, 500s, etc) are caught within `getDelegation` and returned to the caller.

In this case, we show a toast error and retry.


frederikrothenberger (approved):
LGTM, thanks!

PR #1599, opened 2023-05-15
Split e2e tests into smaller files

nmattia:
This ensures there is a better test distribution between shards, meaning the total test duration difference will be smaller between the longest and the shortest shard, meaning the test duration overall (longest shard duration) will be shorter.

No test was modified, code was just moved (and some tests renamed).


frederikrothenberger (approved):
Thanks! I faintly remember a file where the different test files were listed in ascending order of execution duration, or something...

Did that thing go away with the transition to `vitest`?

nmattia:
> I faintly remember a file where the different test files were listed in ascending order of execution duration

It did indeed go away, `vitest` does not support setting up a custom sequencer

nmattia:
@frederikrothenberger FYI maybe I was a bit quick to create a PR, I'm still experimenting a bit

frederikrothenberger:
> @frederikrothenberger FYI maybe I was a bit quick to create a PR, I'm still experimenting a bit

No worries, let me know when / if I should take another look.

nmattia:
@frederikrothenberger PTAL, thank you!

PR #1600, opened 2023-05-15
Don't duplicate BASE_URL in dapp logo

nmattia:
This fixes an issue where the BASE_URL is prepended twice to the dapp logo URL. The BASE_URL is already included in the `dapps.ts` module, which loads the dapps.

This issue turns `/foo.png` into `//foo.png`, which browsers then interpret as `https://foo.png` instead of `https://<this website>/foo.png`.


frederikrothenberger (approved):
LGTM, but this also requires now to change the `DUMMY_CAPTCHA` dummy logo, right?

nmattia:
> LGTM, but this also requires now to change the DUMMY_CAPTCHA dummy logo, right?

Good catch, thanks! I really need to clean that bit up, it's not maintainable

PR #1604, opened 2023-05-15
Fix nns name on known dapps

nmattia:
This ports changes made to dfinity/portal. The name of the NNS dapp is clearer.


src/showcase/src/showcase.ts
```diff
@@ -134,7 +134,7 @@ const authzTemplatesKnownAlt = authnTemplateAuthorize({
   i18n,
 
   knownDapp: {
-    name: "NNS Front-End Dapp",
+    name: "NNS Dapp",
```
peterpeterparker, line 137: Should we update NNS-dapp meta and the portal as well?

peterpeterparker: Same for the logo?

nmattia: It was updated in the portal, this is porting that change; same for the logo (see PR description)

PR #1608, opened 2023-05-15
Don't cache dfx cache

nmattia:
This stops caching `~/.cache/dfx`, which contains a lot of runtime data for `dfx`, but not at all what we actually care about: the executable, which is stored in `/usr/local/bin/dfx`.

The dropped path was pretty big (1GB) so better to not store it if we don't need it.


frederikrothenberger (approved):
Thanks! I assume the screenshots are a left-over from yesterdays release?

PR #1609, opened 2023-05-15
Add stepper header to onboarding screens

nmattia:
This adds a stepper "header" on the onboarding screen to give the users context. Design bits by @meodai .


meodai (approved):
LGTM!

PR #1612, opened 2023-05-16
Fix toaster position

meodai:
Makes sure the error messages are always above the footer.


nmattia (commented):
@meodai didn't we decide on making the toasts appear at the bottom of the screen, but on top of the footer? otherwise when there's no footer they look like they're floating

PR #1616, opened 2023-05-17
Introduce v2 api with identity_info

frederikrothenberger:
This PR adds the first method (`identity_info`) of the v2 API as discussed in #1581.

This PR adds:
* the new method `identity_info`
* the new types required
* infrastructure & tests to convert between some of the new and old types
* integration tests for the new method


src/internet_identity/tests/integration/v2_api/identity_info.rs
```diff
@@ -0,0 +1,198 @@
+//! Tests that `get_identity_info` returns the correct information.
+
+use candid::Principal;
+use canister_tests::api::internet_identity as api;
```
frederikrothenberger, line 4: The goal is to keep refactoring this test until the v2 api is complete enough so that this test does not need a dependency on `api` but only `api_v2`.

src/internet_identity_interface/src/internet_identity/types/v2.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub struct AuthnMethodRegistration {
+    pub expiration: Timestamp,
+    pub authn_method: Option<AuthnMethodData>,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub struct IdentityInfo {
+    pub authn_methods: Vec<AuthnMethodData>,
+    pub recovery_authn_methods: Vec<AuthnMethodData>,
```
przydatek, line 57: Thanks a lot for the zoom-discussion.  As agreed, having the `AuthMethods` in a single list (with a `Purpose`-flag, as it is in api-v1) does have some advantages, so let's keep it this way.

przydatek (approved):
LGTM, thanks!

PR #1617, opened 2023-05-17
Fix flakyness in e2e tests

frederikrothenberger:
This PR fixes flakyness with regards to accepting
alerts related to recovery by explicitly waiting for the
alert to be open.
The alert is now part of the reset action in order to prevent
new tests from reintroducing the flakyness again.

In addition, this PR adds a missing `waitForSeedPhrase` call when
creating a new recovery phrase.


nmattia (approved):
Thanks!

PR #1620, opened 2023-05-19
Add an integration test for storage v7 with extra bucket.

przydatek:
Add an integration test for storage v7.  
The test starts with a full first bucket, hence upon the addition of another anchor allocates a new bucket.


src/internet_identity/tests/integration/stable_memory.rs
```diff
@@ -379,9 +380,68 @@ fn should_read_persistent_state_v7() -> Result<(), CallError> {
...
+    let random_anchor_offset = rand::thread_rng().gen_range(0..anchor_count);
+    let random_anchor = first_anchor_number + random_anchor_offset;
+    let principal = test_setup_helpers::principal(random_anchor_offset as usize);
+    let devices =
+        api::get_anchor_info(&env, canister_id, principal, random_anchor)?.into_device_data();
+    assert_eq!(devices.len(), 2);
+    let device = &devices[0];
+    assert_eq!(format!("device #{}", random_anchor_offset), device.alias);
+
+    // Add a new anchor (this triggers an allocation of a new managed memory bucket).
```
frederikrothenberger, line 416: It would be nice if you would also verify that the metrics reflect that. See here: https://github.com/dfinity/internet-identity/blob/631975942cbc5fcd25013ec82e15124bc1f15a23/src/internet_identity/tests/integration/http.rs#L219

frederikrothenberger (approved):
Thanks!

src/internet_identity/tests/integration/stable_memory.rs
przydatek: Good idea, and thanks for the pointer -- indeed, this has uncovered that the test file had already 2 buckets.
Fixed now.

przydatek (commented):
Thanks, PTAL.

PR #1621, opened 2023-05-19
Add a test for bucket allocation (storage v7).

przydatek:
Add a test that verifies that the managed memory bucket is allocated as expected, i.e. after anchor memory exceeds 2048 anchors.


frederikrothenberger (approved):
LGTM, thanks!

PR #1623, opened 2023-05-22
Refactor authentication check to no longer trap

frederikrothenberger:
This PR changes the authentication check to return a `Result` instead of trapping. This refactoring is a prerequisite to introduce authentication failure as a proper error in the v2 API.


przydatek (approved):
LGTM, thanks!

PR #1625, opened 2023-05-22
Show loader when fetching recoveries

nmattia:
This ensures the `recoveryWizard` displays a loader when fetching the list of recovery methods. Otherwise, it's not clear that the app is waiting for canister data.

Before:

https://github.com/dfinity/internet-identity/assets/6930756/75386a62-de7d-4b8d-ae47-073d12cd15e0

After:

https://github.com/dfinity/internet-identity/assets/6930756/875d0013-385a-4302-bbe1-50f1a75494ba


frederikrothenberger (approved):
In theory, the app has already loaded the recoveries (with the initial lookup for the credential ids) and the state could be funneled through. But it would be a bigger refactoring...

nmattia:
> But it would be a bigger refactoring...

Which I plan on taking on soon 😬 stay tuned...

PR #1626, opened 2023-05-22
Clean up authz authentication flow

nmattia:
This refactors the authentication flow for authz, by making it more consistent with the other (non-authz, i.e. manage) authentication flow and by making the code a bit simpler and more robust. ~~This also fixes an issue where known dapps were not randomized.~~

### DIFF
--- a/src/frontend/src/flows/authorize/index.ts
+++ b/src/frontend/src/flows/authorize/index.ts
@@ -11,7 +11,7 @@ import { recoveryWizard } from "$src/flows/recovery/recoveryWizard";
 import { I18n } from "$src/i18n";
 import { Connection } from "$src/utils/iiConnection";
 import { TemplateElement } from "$src/utils/lit-html";
-import { Chan, shuffleArray } from "$src/utils/utils";
+import { Chan } from "$src/utils/utils";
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/authorize/index.ts
```diff
@@ -138,14 +138,12 @@ export const authnTemplateAuthorize = ({
  */
 export const authFlowAuthorize = async (
   connection: Connection
-): Promise<void> => {
+): Promise<never> => {
   const i18n = new I18n();
-  const container = document.getElementById("pageContent") as HTMLElement;
+  const dapps = shuffleArray(getDapps());
```
frederikrothenberger, line 143: You don't need to shuffle here, right? Only a single logo is needed.

frederikrothenberger (approved):
LGTM, thanks

src/frontend/src/flows/authorize/index.ts
nmattia: ah, that's why it wasn't being shuffled. Thanks!

nmattia: Clarified (inlined). Thanks!

PR #1629, opened 2023-05-22
Shard rust canister tests

frederikrothenberger:
This speeds um CI significantly by making the following changes:
* tests are compiled from the very beginning of the CI pipeline (not waiting for the docker build to complete)
* the compiled tests are reused in each running shard
* the test execution is sharded

Canister test CI times are now dominated by the MacOs compile times. On cache hits, rust tests should be a lot faster now.


.github/workflows/canister-tests.yml
```diff
@@ -237,6 +230,15 @@ jobs:
           echo "now: $commit_now"
 
       - uses: ./.github/actions/bootstrap
+      - name: Install nextest (macos)
+        if: ${{ matrix.os == 'macos-latest' }}
+        run: |
+          curl -LsSf https://get.nexte.st/latest/mac | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
```
nmattia, line 236: nit: I think you could use a single step by either using `fromJSON` and index with the `os` or use something like [this](https://github.com/actions/runner/issues/409) to produce `mac` or `linux`, depending

.github/workflows/canister-tests.yml
```diff
@@ -326,11 +328,51 @@ jobs:
...
+          # downloaded file
+          path: ./canister-tests-${{ matrix.os }}.tar.zst
+
+  canister-tests-run:
+    runs-on: ${{ matrix.os }}
+    needs: [canister-tests-build, docker-build-ii, docker-build-archive]
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
+        partition: ['1/3', '2/3', '3/3']
```
nmattia, line 351: I remember having issues with GHA not liking the `/`s, good to know it works!

.github/workflows/canister-tests.yml
```diff
@@ -362,7 +404,7 @@ jobs:
           # a relatively small price to pay to make sure PRs are always tested against the latest release.
           curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity_test.wasm.gz -o internet_identity_previous.wasm.gz
           curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/archive.wasm.gz -o archive_previous.wasm.gz
-          cargo test --release
+          ./cargo-nextest nextest run --archive-file canister-tests-${{ matrix.os }}.tar.zst --partition hash:${{ matrix.partition }}
```
nmattia, line 407: I didn't dig too deep in the docs, but shouldn't this be `--partition count:1/3`?

frederikrothenberger: They have `count` and `hash`. Using `hash` makes sure, that the partition is stable, even if tests are added or removed (but might lead to slightly worse balancing on few tests).
But I think having stable partitioning is beneficial and we have enough tests that the shards are balanced enough.

nmattia (approved):

PR #1631, opened 2023-05-23
Fix false positives of cargo mtime check

frederikrothenberger:
If the canister code cache layer (which is small) is restored too fast (i.e. within the same second) the cargo mtime check might indicate changes because touch does not set fractional seconds. This PR fixes that by shifting the mtime.


nmattia (approved):
Thanks!

PR #1633, opened 2023-05-23
Skip rust toolchain on test archive cache hit

frederikrothenberger:
This PR optimizes the CI pipeline to not compile
any tests at all, if the rust code has not changed. Instead, the compiled tests from the last commit that changed rust code will be reused.

This paves the way to simplify the rust caching and hopefully get rid of mtime fiddling again.


nmattia (approved):
LGTM, thanks! Would it make sense to extract the bootstrapping + building into another job so that you wouldn't need to add `if:`s everywhere?

frederikrothenberger:
> Would it make sense to extract the bootstrapping + building into another job so that you wouldn't need to add `if:`s everywhere?

Do you mean extracting into a composite action? Because bootstrapping in a different job doesn't work as each job gets a fresh VM, right?

nmattia:
@frederikrothenberger sorry for unqueuing! Might have spotted an issue, just wanna double check. If it's nothing then I'll re-schedule for merge.

.github/workflows/canister-tests.yml
```diff
@@ -202,136 +202,54 @@ jobs:
...
-      # update the mtime of newer files)
-      - name: Retrieve commit information
-        id: git_info
-        run: |
-          # Last time dependencies or toolchain were updated; used for ~/.cargo and dependencies
-          commit_deps=$(git log -1 --pretty=format:%H -- Cargo.* src/*/Cargo.toml rust-toolchain.toml)
-          echo "commit_deps=$commit_deps" >> "$GITHUB_OUTPUT"
-          echo "deps: $commit_deps"
+          path: /tmp/test-archive
+          key: ${{ runner.os }}-rust-test-archive-${{ hashFiles('src/**/*.rs', 'Cargo.*', 'src/*/*.toml', 'rust-toolchain.toml') }}
```
nmattia, line 210: Is there a better way to check for all rust (file) dependencies? I think you might get incorrect cache hits if e.g. the bundled tarball changes; for instance if it stops including an `index.js` (or rather if you don't `touch` the index.js below) you'd still get a cache hit even though in theory the tests should have been rebuilt.

Alternatively, create the dist early and add it to the hash?

Maybe I'm just worried about nothing

nmattia:
> Do you mean extracting into a composite action? 

No sorry, I wasn't clear. I meant separating the building/uploading of the archive into another (or two other) job(s). That way you wouldn't have to call `if:` everywhere. I think the current solution is good enough for now!

.github/workflows/canister-tests.yml
frederikrothenberger: The JS files are not part of the test binaries. Rather the test binaries load the `.wasm` (that has been correctly built) later. So if the JS file changes, it is ok to reuse the same test binary, because the test will see the new file once it loads the `.wasm`.

nmattia: Ah, yes! And as long as the tests are only made of pure rust files we're good. Thanks!

PR #1637, opened 2023-05-25
Ensure logo files exist

nmattia:
This adds a frontend check to ensure that all logo files (images) referenced in the dapp list do exist on disk.


frederikrothenberger (approved):
Thanks!

PR #1638, opened 2023-05-25
Fix dapps.json tmp update

nmattia:
When pruning big files, the dapps.json was not replaced with the updated `dapps.json`. This ensures the original file is replaced with the temp file where the output of `jq` is saved.

See also: https://github.com/dfinity/internet-identity/pull/1637


frederikrothenberger (approved):
Thanks!

PR #1651, opened 2023-05-30
Refactor asset state

frederikrothenberger:
This PR refactors how assets are kept in the `state` module. It creates a clearer separation between assets and the rest of the state and it allows to extend the asset state with v2 certification without breaking clients that rely on the certification (i.e. signed delegations).

### DIFF
--- a/src/internet_identity/src/assets.rs
+++ b/src/internet_identity/src/assets.rs
@@ -6,9 +6,33 @@ use crate::{http, state};
 use base64::engine::general_purpose::STANDARD as BASE64;
 use base64::Engine;
 use ic_cdk::api;
+use ic_certified_map::{AsHashTree, Hash, HashTree, RbTree};
 use include_dir::{include_dir, Dir, File};
+use internet_identity_interface::http_gateway::HeaderField;
 use lazy_static::lazy_static;
 use sha2::Digest;
+use std::collections::HashMap;
+
+const LABEL_ASSETS: &[u8] = b"http_assets";
+
+#[derive(Debug, Default, Clone)]
+pub struct CertifiedAssets {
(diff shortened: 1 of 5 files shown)


src/internet_identity/src/assets.rs
```diff
@@ -6,9 +6,33 @@ use crate::{http, state};
...
+
+#[derive(Debug, Default, Clone)]
+pub struct CertifiedAssets {
+    pub assets: HashMap<String, (Vec<HeaderField>, Vec<u8>)>,
+    pub certification_v1: RbTree<String, Hash>,
+}
+
+impl CertifiedAssets {
+    /// Returns the root_hash of the asset certification tree.
+    /// In the future, this will include both certification v1 and v2.
```
przydatek, line 26: nit: not sure we need this remark about the future -- it doesn't contribute to explaining the current state of the code.  If you prefer to keep it, maybe turn into a TODO?

src/internet_identity/src/main.rs
```diff
@@ -414,10 +412,10 @@ fn save_persistent_state() {
 
 fn update_root_hash() {
     use ic_certified_map::{fork_hash, labeled_hash};
-    state::asset_hashes_and_sigs(|asset_hashes, sigs| {
+    state::assets_and_signatures(|assets, sigs| {
         let prefixed_root_hash = fork_hash(
-            // NB: Labels added in lexicographic order
-            &labeled_hash(LABEL_ASSETS, &asset_hashes.root_hash()),
+            &assets.root_hash(),
+            // NB: sigs have to be added last due to lexical ordering of labels
```
przydatek, line 418: nit: I'm not a native speaker, but I'd keep "lexicographic ordering" here.
(cf. https://wikidiff.com/lexical/lexicographic)

przydatek (approved):
LGTM, thanks!

PR #1652, opened 2023-05-30
Add Astro build for showcase

nmattia:
This introduces [astro](https://astro.build) build files for the showcase.

Before this, the showcase had some hacks in place to provide routing between the "pages", which also required all pages to be served as 404s (since there were no actual pages). The showcase build was also hacked together from the main II vite config, and had virtually no support for testing deploy on other paths (base url) than `/`.

Astro solves all these issues by providing build support (now 1 HTML for each showcase page) and consistency through conventions.


src/showcase/src/pages/index.astro
```diff
@@ -0,0 +1,10 @@
+---
+import "$src/styles/main.css";
+---
+
+<div id="pageContent"></div>
+<script>
+  import { defaultPage } from "../../../showcase/src/showcase";
```
peterpeterparker, line 7: can you use paths `$src/showcase`?

src/showcase/src/pages/[page].astro
```diff
@@ -0,0 +1,21 @@
...
+
+const { page } = Astro.params;
+---
+
+<div id="pageContent" data-page-name={page}></div>
+<script>
+  import { iiPages } from "../../../showcase/src/showcase";
+
+  const page = (document.querySelector("[data-page-name]") as HTMLDivElement)
+    .dataset?.pageName as string;
```
Comment (peterpeterparker) [inline src/showcase/src/pages/[page].astro:17]: Any particular reason not using `define:vars` to access the `page` variable?

src/showcase/src/pages/404.astro
```diff
@@ -0,0 +1,10 @@
+---
+import "$src/styles/main.css";
+---
+
+<div id="pageContent"></div>
```
peterpeterparker, line 5: you can create a component for `<div id="pageContent"></div>` and `import "$src/styles/main.css";` (it's repeated three times in this PR)?

tsconfig.json
```diff
@@ -16,6 +16,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "types": ["node", "vite/client", "vitest/globals"],
+    "baseUrl": ".",
```
peterpeterparker, line 19: is this use only for the showcase or for II as well?
if only showcase, then it would be cleaner to have add it in another  `tsconfig` (that inherits this `tsconfig`)

peterpeterparker:
Not for this PR as it was incomplete before but would be nice to define meta tags and favicons.

If it should be crawlable or not, likewise a robots.txt would be useful.

And if crawlable other required information such as sitemap.xml etc.

https://6zvwc-sqaaa-aaaal-aalma-cai.raw.icp0.io/d/prepare-your-dapps-for-social-platform-and-seo

src/showcase/src/pages/[page].astro
Comment (nmattia) [reply src/showcase/src/pages/[page].astro:17]: Nice! Looked for something like this but couldn't find it

src/showcase/src/pages/404.astro
nmattia: I really hate this `pageContent` thingy so I wanted to avoid hiding it in a component/layout. I'll leave it like this for now and see what the future brings...

tsconfig.json
nmattia: yep good point! tbh I wasn't sure how the `compilerOptions` would merge but I'm guessing it doesn't do anything stupid

src/showcase/astro.config.ts
```diff
@@ -0,0 +1,21 @@
+import { defineConfig } from "astro/config";
+
+const ROOT = "../../";
+
+// https://astro.build/config
+export default defineConfig({
+  publicDir: ROOT + "src/frontend/assets",
```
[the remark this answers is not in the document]
meodai: `path.join()`?

src/showcase/src/pages/[page].astro
Comment (nmattia) [reply src/showcase/src/pages/[page].astro:17]: Actually, this then inlines the script which creates other issues. I'll have a look at this separately but will leave as is for now.

nmattia:
@peterpeterparker PTAL

src/showcase/astro.config.ts
```diff
@@ -1,11 +1,12 @@
 import { defineConfig } from "astro/config";
+import path from "path";
```
peterpeterparker, line 2: Just in case you want to apply the same pattern as the existing config:

```
import { join } from "path";
```

peterpeterparker (approved):
LGTM, thx

src/showcase/astro.config.ts
nmattia: almost did, but then realized it was confusing since there a lots of `join` functions doing different things. WDYT?

peterpeterparker: No opinion, like I said, "just in case".

PR #1655, opened 2023-06-01
Import identity card image from vite

nmattia:
This removes the identity card image from the assets directory and instead creates a direct import, meaning we let `vite` deal with setting the correct BASE_URL. The image is now also closer to where it's used in the code.


frederikrothenberger (approved):
Thanks!

PR #1656, opened 2023-06-01
Always scroll to actions in tests

nmattia:
For tests to work, the dropdown trigger on the management page neds to be in view. Until now we've scrolled to it separately every time an action is called, except in the `reset` case.

Now the scrolling is added to the "open device actions" helper, which simplifies the test code, and ensures the scroll is always performed.


frederikrothenberger (approved):
Thanks!

PR #1657, opened 2023-06-01
Add caching for fonts

frederikrothenberger:
The fonts will be cached for 1 week. If we are happy with the results, we can increase the `max-age` and expand to more file types.

Caching only works for the service worker 1.8.0 and up (scheduled to be rolled out to mainnet today, currently still in canary).

This was tested using the deployment https://y2aaj-miaaa-aaaad-aacxq-cai.ic0.app/ and the canary service worker that you can get by adding the following lines to the hosts file:

```
147.75.108.42 y2aaj-miaaa-aaaad-aacxq-cai.ic0.app
[2604:1380:45e1:a603:5000:cfff:feaf:ee86] y2aaj-miaaa-aaaad-aacxq-cai.ic0.app
```


src/internet_identity/tests/integration/http.rs
```diff
@@ -451,3 +465,30 @@ fn metrics_anchor_operations() -> Result<(), CallError> {
 
     Ok(())
 }
+
+fn verify_response_certification(
```
przydatek, line 469: That's a nice helper! 👍

przydatek (approved):
LGTM, thanks!

PR #1660, opened 2023-06-02
Update rust dependencies

frederikrothenberger:
Most notable dependency updated is the ic-cdk, which now supports the `is_controller` system API call.


Cargo.lock
```diff
@@ -28,6 +28,12 @@ dependencies = [
  "memchr",
 ]
 
+[[package]]
+name = "android-tzdata"
```
nmattia, line 32: I'm really surprised, why do we depend on this? Is there a feature we can disable?

frederikrothenberger: Thanks for pointing it out. This came from `serde_with` which we apparently don't need at all.

nmattia (approved):
LGTM, thanks!

PR #1661, opened 2023-06-02
Migrate stable memory tests to v7

frederikrothenberger:
This PR migrates the stable memory tests unrelated to the memory manager layout migration to v7.

This is a preparation for removing all v6 related code / tests.

In addition, it adds two helpers to more easily migrate/recreate the stable memory backups in the future.


przydatek (approved):
LGTM, thanks!

PR #1665, opened 2023-06-05
Auto reload CAPTCHA on incorrect input

nmattia:
This changes the CAPTCHA flow to fetch a new challenge automatically if the user provides incorrect input.

The copy is slightly modified to be shorter and more understandable. The text input is also changed to stay black on error, to show that the input content is still interactable/modifiable.


peterpeterparker (approved):
LGTM

PR #1666, opened 2023-06-05
Support multiple char replacements

nmattia:
This adds support for mapping multiple chars to one char in the CAPTCHA (e.g. `1`, `I`, `l` etc can all map to `i`).

This is done by reworking the `CHAR_REPLACEMENTS` map to read the "replacement" chars from its keys and "replaced" chars from its values (now an array).

Additionally, this extracts the font creation/parsing to a (lazy) static value, so that the font doesn't have to be recreated every time.

---

NOTE: had to perform some changes [upstream](https://github.com/nmattia/captcha/tree/nm-fonts), will try to get those merged

### DIFF
--- a/src/internet_identity/src/anchor_management/registration.rs
+++ b/src/internet_identity/src/anchor_management/registration.rs
@@ -9,10 +9,12 @@ use ic_cdk::{call, caller, trap};
 use internet_identity_interface::archive::types::{DeviceDataWithoutAlias, Operation};
 use internet_identity_interface::internet_identity::types::*;
 use rand_core::{RngCore, SeedableRng};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 #[cfg(not(feature = "dummy_captcha"))]
 use captcha::filters::Wave;
+use captcha::fonts::Default as DefaultFont;
+use captcha::fonts::Font;
 use lazy_static::lazy_static;
 
 mod rate_limit;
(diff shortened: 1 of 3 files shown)


src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -9,10 +9,12 @@ use ic_cdk::{call, caller, trap};
 use internet_identity_interface::archive::types::{DeviceDataWithoutAlias, Operation};
 use internet_identity_interface::internet_identity::types::*;
 use rand_core::{RngCore, SeedableRng};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 #[cfg(not(feature = "dummy_captcha"))]
 use captcha::filters::Wave;
+use captcha::fonts::Default;
```
nmattia, line 16: import `as` something else

src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -122,32 +124,55 @@ fn create_captcha<T: RngCore>(rng: T) -> (Base64, String) {
 }
 
 lazy_static! {
-    /// Map of problematic characters that are easily mixed up by humans to the "normalized" replacement.
-    /// I.e. the captcha will never contain any of the characters in the key set and any input provided
-    /// will be mapped to the matching value.
+    /// Problematic characters that are easily mixed up by humans to the "normalized" replacement.
```
przydatek, line 127: nit: this line doesn't seem to be grammatically correct anymore (it was "map of ... chars ... to the ...", now it is only "chars  ... to the ...".

przydatek (approved):
LGTM, thanks!

src/internet_identity/src/anchor_management/registration.rs
nmattia, line 127:
```suggestion
    /// Problematic characters that are easily mixed up by humans to "normalized" replacement.
```

PR #1671, opened 2023-06-06
Update support links

nmattia:
Support has some new links. We now link (or rather, redirect) the old `/faq` endpoint to the identity support landing, and update the dapps explorer submission link to point to the new dapps submission support page: https://identitysupport.dfinity.org/hc/en-us/articles/16195954040724


peterpeterparker (approved):
LGTM

PR #1672, opened 2023-06-06
Focus CAPTCHA input on retry

nmattia:
The CAPTCHA input loses focus when the user clicks the "Verify"/"Next" button.

This re-focuses the input in case the user has to retry, to simplify the flow & draw attention back to the input.


peterpeterparker (approved):
LGTM

PR #1673, opened 2023-06-06
Fix card label z-index

nmattia:
This removes the z-index that was set on the card-label. The z-index was there to push the content above warning & modal `::after`s, but then caused issues by covering e.g. the close button in the dapps explorer.

By setting the z-index once and for all on the `page-content` and using said `page-content` to wrap all the inner content of cards we ensure that elements inside cards are consistently shown.


meodai (approved):
LGTM! Thanks

PR #1675, opened 2023-06-07
Remove c-card__content

nmattia:
The `c-card__content` class was a workaround for pushing forward content inside cards where the z-index had to be adapted to display `::after` behind the content. The workaround itself then led to more marging workarounds.

Instead the wrapping cards now use `isolation: isolate` to create a new stacking context and allow the `::before` and `::after` to simply use arbitrarily large neg. z-indexs.

### DIFF
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -937,17 +937,6 @@ by all browsers (FF is missing) */
   fill: var(--rc-onButton);
 }
 
-.c-card__content {
-  position: relative;
-  z-index: var(--z-foreground);
-  /* The card content is insed the c-card, so is padded. But since it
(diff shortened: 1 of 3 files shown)


src/frontend/src/styles/main.css
```diff
@@ -964,14 +953,17 @@ by all browsers (FF is missing) */
...
+  isolation: isolate;
   border-radius: var(--rs-card-border-radius);
 }
 
 .c-card--warning::before,
 .c-card--modal::before {
   position: absolute;
   content: "";
-  z-index: 0;
+  z-index: -10;
```
nmattia, line 966: I think they could both simply be `-1` as well

meodai:
nice, I have to read more into this feature then!

meodai (approved):

PR #1676, opened 2023-06-08
Don't use index as 404 for showcase

nmattia:
The previous showcase build (pre-Astro) needed the index served on 404, since there was actually only a single page built which took care of the routing.

Now that we have one page built per actual page, we don't need this workaround, and instead can serve the actual 404 on missing pages.


peterpeterparker (approved):
LGTM

PR #1677, opened 2023-06-08
Clean up card warning CSS

nmattia:
This cleans up the card warning CSS class furth by clarifying what is used to actually style the card vs. set up for being able to style the pseudo elements, and simplifies the `z-index` values.


meodai (approved):
much nicer! LGTM

PR #1678, opened 2023-06-08
Wrap dapp descriptions in KnownDapp

nmattia:
This adds a type wrapper (class), `KnownDapp`, that ensures that the logo field of `DappDescription` is used correctly.

This ensures the path to actual logo files (`"<BASE_URL>/icons/<logo>"`) has to be set only once.


peterpeterparker (approved):
LGTM

PR #1681, opened 2023-06-12
Remove unnecessary nullish check

nmattia:
At some point, `latest` was potentially `undefined` so we checked before `yield`ing it. Some time after that, the value was made mandatory, but the check was never removed.


meodai (approved):
LGTM!

PR #1682, opened 2023-06-12
Update copy on identity & phrase creation

nmattia:
This clarifies the copy & removes the bullet points, since the paragraphs were not actually lists.


src/frontend/src/styles/main.css
```diff
@@ -1631,6 +1631,7 @@ by all browsers (FF is missing) */
 
 .c-list--bulleted {
   padding-left: var(--rs-list-bullet-space);
+  margin-top: 1em; /* match paragraph margin */
```
meodai, line 1634: I would add something like: `--rs-list-stack` to the tokens in`:root`

and then reuse it here: `margin-top: var(--rs-list-item-stack)`

PR #1683, opened 2023-06-12
Add stepper for phrase creation

nmattia:
This adds a stepper component at the top of the phrase creation & confirmation pages.

The stepper is adapted from the other wizard (identity creation).


meodai (approved):
LGTM, very clean and simple!

PR #1686, opened 2023-06-13
Drop WeakRef handling of Chan listeners

nmattia:
We used `WeakRef`s to know if a particular `Chan` listener was still alive or not; if not, we'd re-register it automatically.

This caused extra complexity in the code, and the implementation was not correct in all cases. In addition to that, the codebase as it is today does not really benefit from this. For these reasons, the extra logic is now removed.


peterpeterparker (approved):
LGTM

PR #1691, opened 2023-06-13
Show recovery phrase info on management

nmattia:
This re-uses the "phrase info page" that is shown when we nudge the user to create a phrase, and shows it also when the user proactively creates a new phrase from the management page (though slightly modified to fit the flow).


peterpeterparker (commented):
LGTM

PR #1692, opened 2023-06-13
Make recovery info page less scary

nmattia:
This updates the recovery info page (shown before recovery creation):
* The "warning" icon is now an "info" icon (red -> purple)
* The text ("SECURITY WARNING") is a bit softer ("Learn about recovery")


screenshots/desktop/addPhrase.png
nmattia, line 1: I feel like the icon _could_ be lighter, i.e. maybe use the outline instead of a filled icon

src/frontend/src/styles/main.css
```diff
@@ -698,6 +701,12 @@ a:hover,
 .c-icon--ok {
   --onBgc: var(--rc-positive);
 }
+.c-icon--info__flipped {
```
meodai, line 704: `__` would be a child in BEM. I would create a separate class called `.c-icon--inverted` that combines with any icon class. But since it looks like info is always flipped, it might not need a modifier? So just `c-icon--info`, makes sense?

src/frontend/src/components/icons.ts
```diff
@@ -256,6 +256,21 @@ export const infoIcon = html`
   </svg>
 `;
 
+export const infoIconNaked = html`
```
meodai, line 259: not sure what is meant by Naked would be super nice to have a comment or something here.

meodai (approved):

meodai (approved):

PR #1694, opened 2023-06-15
Differentiate phrase info vs phrase reminder

nmattia:
Originally, we only showed phrase information in a "warning" page that reminded the user to add a phrase. We then replaced this with a purely information (i.e. no "warning") page, also shown when the user themselves start the phrase creation flow.

This takes this a bit further by changing the page we show, depending on the intent: as an info if user initiated, or as a warning if it is a reminder.


meodai (approved):
LGTM

PR #1698, opened 2023-06-15
Send Slack notification on RC deploy fail

nmattia:
This updates the RC deployment to notify the team if the deployment failed. This is similar to what we do for other scheduled jobs (node updated, rust update, etc).

### DIFF
--- a/.github/workflows/deploy-rc.yml
+++ b/.github/workflows/deploy-rc.yml
@@ -84,3 +84,12 @@ jobs:
             Internet Identity release candidate
             RC link: https://${{ env.ii_canister_id }}.ic0.app
             test app: https://${{ env.testnet_app_canister_id }}.ic0.app
+
+        # Since the this is a scheduled job, a failure won't be shown on any
+        # PR status. To notify the team, we send a message to our Slack channel on failure.
+      - name: Notify Slack on failure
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-rc.yml
```diff
@@ -84,3 +84,12 @@ jobs:
             Internet Identity release candidate
             RC link: https://${{ env.ii_canister_id }}.ic0.app
             test app: https://${{ env.testnet_app_canister_id }}.ic0.app
+
+        # Since the this is a scheduled job, a failure won't be shown on any
+        # PR status. To notify the team, we send a message to our Slack channel on failure.
+      - name: Notify Slack on failure
```
przydatek, line 90: nit: not sure whether this is relevant, but maybe: most of the other `- name:` entries above are in quotation marks (except for the one in line 79), so maybe use them also here?

przydatek (approved):
LGTM, thanks!

.github/workflows/deploy-rc.yml
nmattia: That's right! We use both somewhat inconsistently, as YAML allows for both. Other "Notify Slack on failure" entries use this style (without quotes), I just copy/pasted one.

PR #1699, opened 2023-06-15
Disable spellcheck when not necessary

nmattia:
This disables browser spellchecking on captcha + recovery words.

The CAPTCHA definitely doesn't spellchecking, and the recovery words anyway get validated against the BIP39 word checklist so don't need the red browser squiggly/dashed underline.


meodai (approved):
awesome!

PR #1701, opened 2023-06-16
Remove a flag remaining from migration to MemoryManager.

przydatek:
The migration to `MemoryManager` (storage v7) is completed, so the flag `InternetIdentityInit.migrate_storage_to_memory_manager` is now obsolete hence can be removed.


nmattia (approved):
Thanks, LGTM! No candid changes?

przydatek:
> Thanks, LGTM! No candid changes?
Thanks for the review. 
Indeed, we didn't have the field in the `.did` file, that was probably an oversight... 🤔 😓

PR #1707, opened 2023-06-19
Use HTML autofocus attribute

nmattia:
This replaces our custom `autofocus` lit directive with the simple, more standard
[`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) HTML attribute. When set to true, this attribute will ensure the element it is attached to is focused when inserted in the DOM. If several elements have the attribute set, only the first element will be focused.


peterpeterparker (approved):
LGTM, thanks

PR #1710, opened 2023-06-20
Remove mentions of Haskell

nmattia:
We don't have haskell tests anymore, this updates the workflow comment to reflect that.


przydatek (approved):
LGTM, thanks!

PR #1713, opened 2023-06-21
Fix screenshot workflow on empty description

nmattia:
The screenshot workflow made the assumption that all PRs had a body; in practice this may not be the case. PRs created without a description will have a `null` body.


przydatek (approved):
LGTM, thanks!

PR #1715, opened 2023-06-21
Use get_anchor_credentials for checking recovery

nmattia:
This changes `fromSeedPhrase` to use the "new" `get_anchor_credentials` API endpoint to retrieve a list of known recovery phrase public keys.

Previously we required the caller to provide a `device` which had to be looked up before.

This is preparation work for simplifying the recovery flow, where we won't even be asking the user for their identity number prior to prompting them with the recovery phrase input.


przydatek (approved):
LGTM, thanks!

PR #1716, opened 2023-06-21
Remove updateOriginIfNecessary hack

nmattia:
This removes a hack that was introduced during when we first started recording `origin`s for passkeys.

The intent was to record a meaningful `origin` for a passkey, when that passkey was used for authentication. While newly created passkeys do record an `origin`, older passkeys did not.

In practice however it is not clear how this worked, as the passkey seems to be looked up using an incorrect public key (that of the delegation instead of the passkey's itself).

Also, this was only intended as a temporary hack:

>  this hack can be removed in a few weeks [#1384](https://github.com/dfinity/internet-identity/pull/1384)

### DIFF
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -270,60 +270,12 @@ export class Connection {
       actor
     );
 
-    try {
-      this.updateOriginIfNecessary(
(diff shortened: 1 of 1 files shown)


przydatek (approved):
LGTM, thanks!

PR #1717, opened 2023-06-21
Use credentials instead of devices to authenticate

nmattia:
This replaces `fromWebauthnDevices` with `fromWebauthnCredentials`, which is more correct. The reason is that all devices may not have credentials. Previously, trying to build an (sign) identity with a device without credential was an error implicitly ignored.

This is part of the work for simplifying the recovery flow, and is also necessary in order to fully deprecate the `lookup` endpoint.

The error handling in `recoverWith/device.ts` will be replaced very soon to reflect the new planned recovery flow.


przydatek (approved):
LGTM, thanks!

PR #1720, opened 2023-06-21
Read user number from recovery phrase input

nmattia:
This changes the recovery phrase screens to not require the userNumber to be known before the user types their phrase. Instead, the number is read from the input.

This is part of the ongoing work for simplifying the recovery flow. By not requiring the user number to be known, we'll be able to skip the user number prompt step.


przydatek (approved):
LGTM, thanks!

PR #1727, opened 2023-06-26
Remove styling on #pageContent

meodai:
`#pageContent` is currently styled alongside `.l-wrap`. this sometimes has unexpected behaviours since `#pageContent` sometimes wraps `.l-wrap`.

The goal of this PR is to not apply any styles to `#pageContent` and add `.l-wrap` wherever needed.

### DIFF
--- a/.ic-commit
+++ b/.ic-commit
@@ -1,3 +1,3 @@
 # the commit used to pull the state machine executable
 # see rust canister tests for more info
-387f39130bc2f2e5b081ad65b16b61eefb1e908b
+5317b1a6089cca20a51c078e738b4eb3f6eec68f
--- a/src/frontend/src/flows/compatibilityNotice.ts
+++ b/src/frontend/src/flows/compatibilityNotice.ts
@@ -1,23 +1,28 @@
 import { compatibilityChart } from "$src/components/compatibilityChart";
+import { mainWindow } from "$src/components/mainWindow";
 import { warnBox } from "$src/components/warnBox";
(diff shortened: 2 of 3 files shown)


src/frontend/src/components/mainWindow.ts
```diff
@@ -30,6 +30,16 @@ export const mainWindow = ({
   additionalContainerClasses?: string[];
   HTMLwrapperTag?: string;
 }): TemplateResult => {
+  const pageContent = document.querySelector("#pageContent");
+  const wrap =
+    pageContent !== null ? pageContent : document.querySelector("body");
```
nmattia, line 35: I think we use `nonNullish` now instead of explicitly checking for null/undefined

src/frontend/src/components/mainWindow.ts
```diff
@@ -30,6 +30,16 @@ export const mainWindow = ({
...
+
+  // make sure the wrapper always has the l-wrap class, since it is used to
+  // position the content and makes sure it has appropriate spacings
+  if (wrap !== null && wrap.classList.contains("l-wrap") === false) {
```
nmattia, line 39: this is a bit hacky, why not wrap the window in another element that contains `l-wrap`?

meodai: I did that because, we are doing this to be able to accommodate the sidebar later (with l-wrap--sidebar). But I guess I will be solving this problem in the PR of the sidebar then

PR #1732, opened 2023-06-27
Add main window with sidebar component and css

meodai:
Screenshot 2023-06-27 at 20 45 34

### DIFF
--- a/src/frontend/src/components/mainWindowWithSidebar.ts
+++ b/src/frontend/src/components/mainWindowWithSidebar.ts
@@ -50,6 +50,8 @@ export const mainWindowWithSidebar = ({
   if (additionalContainerClasses.length > 0) {
     containerClasses.push(...additionalContainerClasses);
   }
+  // we do not use the logo from the icons file because it duplicates the the ID's
+  // (also hides those corresponding IDs in the other instances of the logos when the sidebar is not visible)
   return html`
     <div class="l-wrap l-wrap--sidebar">
       <div class="l-sidebar is-hidden--mobile">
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/mainWindowWithSidebar.ts
nmattia, line 1: Can I put this in the `authenticateBox.ts` module or nearby? There's not need to make this a general "component" if we don't use this anywhere else. Plus the comments & name are misleading, this isn't really a "main window" "[used] to wrap the content of each page" since we only use it in one page

src/frontend/src/components/mainWindowWithSidebar.ts
```diff
@@ -0,0 +1,127 @@
...
+    <div class="l-wrap l-wrap--sidebar">
+      <div class="l-sidebar is-hidden--mobile">
+        <div class="l-sidebar__main">
+          <div class="c-logo c-logo--sidebar">
+          <svg
+            xmlns="http://www.w3.org/2000/svg"
+            fill="none"
+            viewBox="0 0 233 111"
+          >
+            <defs>
```
nmattia, line 63: IIRC you're inlining the logo because the IDs cannot be duplicated, can you leave a comment about this?

src/frontend/src/components/mainWindowWithSidebar.ts
meodai: yes of course. Let me know if I should look into that

src/frontend/src/components/mainWindowWithSidebar.ts
meodai: exactly. Im adding a comment

sea-snake:
A sidebar has been implemented in another PR

PR #1744, opened 2023-07-12
Add a narrow variant for the stepper

meodai:
Screenshot 2023-07-12 at 15 21 31

Adds a space to the right and the left of the stepper, that can be used when there are only a few steps.

### DIFF
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -2295,6 +2295,11 @@ a.c-action-list__item {
   text-align: center;
 }
 
+.c-progress-stepper--narrow {
+  padding-left: 10%;
+  padding-right: 10%;
+}
+
 .c-progress-stepper__step {
   counter-increment: progress-step;
   display: flex;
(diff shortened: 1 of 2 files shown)


src/frontend/src/styles/main.css
```diff
@@ -2295,6 +2295,11 @@ a.c-action-list__item {
   text-align: center;
 }
 
+.c-progress-stepper--narrow {
+  padding-left: 10%;
```
peterpeterparker, line 2299: Just to be sure, you on purpose used percent right?

all other paddings of the dapp seems to be expressed in relative size `em`.

peterpeterparker (approved):
LGTM, just added one question in case interesting.

src/frontend/src/styles/main.css
meodai: @peterpeterparker yes, especially on mobile I want the space to not be a fixed size, in order to alway look appropriate! Thanks for cheking!

PR #1756, opened 2023-07-26
Upgrade IC related dependencies

frederikrothenberger:
In the candid update in  particular finally allows more flexibility on extending candid variants and brings a performance improvement.

### DIFF
--- a/src/archive/Cargo.toml
+++ b/src/archive/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2021"
 # local dependencies
 internet_identity_interface = { path = "../internet_identity_interface" }
 # ic dependencies
-candid = { version = "0.9", features = ["parser"] }
+candid = "0.9"
 ic-cdk = "0.10"
 ic-cdk-timers = "0.4"
 ic-cdk-macros = "0.7"
@@ -21,6 +21,7 @@ serde = "1"
 serde_bytes = "0.11"
 
 [dev-dependencies]
+candid = { version = "0.9", features = ["parser"] }
 canister_tests = { path = "../canister_tests" }
 hex = "0.4"
 ic-test-state-machine-client = "3"
--- a/src/archive/Cargo.toml
+++ b/src/archive/Cargo.toml
@@ -10,10 +10,10 @@ edition = "2021"
 # local dependencies
 internet_identity_interface = { path = "../internet_identity_interface" }
 # ic dependencies
-candid = "0.8"
-ic-cdk = "0.8"
-ic-cdk-timers = "0.2"
-ic-cdk-macros = "0.6"
+candid = { version = "0.9", features = ["parser"] }
+ic-cdk = "0.10"
+ic-cdk-timers = "0.4"
+ic-cdk-macros = "0.7"
 ic-metrics-encoder = "1"
 ic-stable-structures = "0.5"
 # other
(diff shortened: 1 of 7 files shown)


src/archive/Cargo.toml
```diff
@@ -10,10 +10,10 @@ edition = "2021"
 # local dependencies
 internet_identity_interface = { path = "../internet_identity_interface" }
 # ic dependencies
-candid = "0.8"
-ic-cdk = "0.8"
-ic-cdk-timers = "0.2"
-ic-cdk-macros = "0.6"
+candid = { version = "0.9", features = ["parser"] }
```
dskloetd, line 13: What does `features = ["parser"]` do and why was it added?

src/canister_tests/src/api/internet_identity.rs
```diff
@@ -16,7 +16,7 @@ pub fn health_check(env: &StateMachine, canister_id: CanisterId) {
     let user_number: types::AnchorNumber = 0;
     // XXX: we use "IDLValue" because we're just checking that the canister is sending
     // valid data, but we don't care about the actual data.
-    let _: (candid::parser::value::IDLValue,) =
+    let _: (candid::types::value::IDLValue,) =
```
dskloetd, line 19: Was the package moved or are you using something else now?

src/internet_identity_interface/src/http_gateway.rs
```diff
@@ -1,17 +1,22 @@
...
-use candid::{CandidType, Deserialize, Func};
+use candid::{define_function, CandidType, Deserialize};
 use serde_bytes::ByteBuf;
 
 pub type HeaderField = (String, String);
 
 #[derive(Clone, Debug, CandidType, Deserialize)]
 pub struct Token {}
 
+define_function!(pub StreamingCallbackFunction : (Token) -> (StreamingCallbackHttpResponse) query);
```
dskloetd, line 12: Can you explain this change?

src/archive/Cargo.toml
frederikrothenberger: The languages team decided to split the candid library (see [here](https://github.com/dfinity/candid/blob/master/Changelog.md#breaking-changes)), or rather hide not as often used functionality behind this feature flag. In particular we require the `parser` feature to test whether the code actually implements the did file.

But we don't need the parser feature for the canister. I have now only enabled in in `dev-dependencies`. Thanks.

src/canister_tests/src/api/internet_identity.rs
frederikrothenberger: It was moved, see [here](https://github.com/dfinity/candid/blob/master/Changelog.md#breaking-changes).

src/internet_identity_interface/src/http_gateway.rs
frederikrothenberger: Previously, the typings for the streaming callback function were inaccurate (because we just had the raw `Func` type because it was super cumbersome to actually define it properly). But it didn't matter, since we don't use the full set of all HTTP gateway interactions.

Now, the languages team made this (common) shortcut error out at runtime and instead provides better functionality (namely the `define_function!` macro) to conveniently type a function including the arguments, return values and the mode (i.e. query).

So essentially cleans up an old piece of tech debt.

PR #1758, opened 2023-07-28
Continue with another device copy changes

frederikrothenberger:
Copy changes for the happy path when pressing the
"continue with another device" button on the landing page.

Includes a change to always prompt the anchor pickor with additional text.

The error message has not been changed, since it is used in multiple places and would first need to be split.


mstrasinskis (approved):
LGTM.

PR #1759, opened 2023-07-28
Minor refactor add new device flow

frederikrothenberger:
A recent flow change made some names misleading.
This PR fixes that.

### DIFF
--- a/src/frontend/src/components/authenticateBox.ts
+++ b/src/frontend/src/components/authenticateBox.ts
@@ -331,14 +331,18 @@ const page = (slot: TemplateResult) => {
 };
 
 // Register this device as a new device with the anchor
-const asNewDevice = async (connection: Connection, userNumber?: bigint) => {
-  // Prompt the user for an anchor (only used if we don't know the anchor already)
-  const askUserNumber = async (userNumber?: bigint) => {
+const asNewDevice = async (
+  connection: Connection,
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/authenticateBox.ts
```diff
@@ -331,14 +331,18 @@ const page = (slot: TemplateResult) => {
...
+  connection: Connection,
+  prefilledUserNumber?: bigint
+) => {
+  // Prompt the user for an anchor and provide additional information about the flow.
+  // If the user number is already known, it is prefilled in the screen.
+  const promptUserNumberWithInfo = async (prefilledUserNumber?: bigint) => {
     const result = await promptUserNumber({
       title: "Continue with another device",
       message:
         "Is this your first time connecting to Internet Identity on this device? In the next steps, you will add this device as an Internet Identity passkey. Do you wish to continue?",
```
dskloetd, line 344: Given that the user number might be absent, in which case the user would have to enter it, I still find it strange that the only question here is whether the user wants to continue.

dskloetd (approved):

src/frontend/src/components/authenticateBox.ts
frederikrothenberger: This might be implied enough, given you can only continue with a number entered.

I'll ask @marydwyer about it. Thanks.

PR #1760, opened 2023-07-28
Remove outdated compatibility test API

frederikrothenberger:
It was previously required for a change to the stats endpoint. Given that the change is now also contained in the previous release, these types are no longer needed.

### DIFF
--- a/src/internet_identity/tests/integration/upgrade.rs
+++ b/src/internet_identity/tests/integration/upgrade.rs
@@ -152,7 +152,7 @@ fn ii_upgrade_should_allow_same_user_range() -> Result<(), CallError> {
     let env = env();
     let canister_id = install_ii_canister(&env, II_WASM_PREVIOUS.clone());
 
-    let stats = api::compat::stats(&env, canister_id)?;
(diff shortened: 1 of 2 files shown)


src/internet_identity/tests/integration/upgrade.rs
```diff
@@ -152,7 +152,7 @@ fn ii_upgrade_should_allow_same_user_range() -> Result<(), CallError> {
     let env = env();
     let canister_id = install_ii_canister(&env, II_WASM_PREVIOUS.clone());
 
-    let stats = api::compat::stats(&env, canister_id)?;
+    let stats = api::stats(&env, canister_id)?;
```
dskloetd, line 155: Does this test still test anything or do you only keep it around for the next time there is a change in the type?

dskloetd (approved):

src/internet_identity/tests/integration/upgrade.rs
frederikrothenberger: Yeah, good point. This seems like a very low signal test. Let me just delete it. 👍

PR #1762, opened 2023-07-28
Move test helper to make it useable in other tests

frederikrothenberger:
Creating new identities based off authn_methods will be requiered in almost all v2 API integration tests. This PR moves it into a helper file where it can be used from any other test.

### DIFF
--- a/src/internet_identity/tests/integration/v2_api/authn_method_test_helpers.rs
+++ b/src/internet_identity/tests/integration/v2_api/authn_method_test_helpers.rs
@@ -1,5 +1,9 @@
+use canister_tests::api::internet_identity as api;
+use ic_cdk::api::management_canister::main::CanisterId;
+use ic_test_state_machine_client::StateMachine;
 use internet_identity_interface::internet_identity::types::{
-    AuthnMethod, AuthnMethodData, AuthnMethodProtection, PublicKeyAuthn, Purpose,
+    AuthnMethod, AuthnMethodData, AuthnMethodProtection, ChallengeAttempt, DeviceData,
+    DeviceWithUsage, IdentityNumber, PublicKeyAuthn, Purpose, RegisterResponse,
 };
 use serde_bytes::ByteBuf;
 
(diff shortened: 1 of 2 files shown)


src/internet_identity/tests/integration/v2_api/authn_method_test_helpers.rs
```diff
@@ -26,3 +30,20 @@ pub fn test_authn_method() -> AuthnMethodData {
         last_authentication: None,
     }
 }
+
+pub fn create_identity_with_authn_method(
+    env: &StateMachine,
+    canister_id: CanisterId,
+    authn_method: &AuthnMethodData,
```
dskloetd, line 37: Why is this now passed by reference while before it wasn't?

dskloetd (commented):
Why does the PR description say some screens were changed?

src/internet_identity/tests/integration/v2_api/authn_method_test_helpers.rs
frederikrothenberger: Having the function only borrow the argument allows tests to use it multiple times. This moves the call to `clone` into the function making the tests less cluttered.

frederikrothenberger:
> Why does the PR description say some screens were changed?

There seems to be some flakyness regarding the screenshots. I will talk to @nmattia about that, since there have been a lot of changes to that infrastructure since I went on PTO.

src/internet_identity/tests/integration/v2_api/authn_method_test_helpers.rs
dskloetd: None of the tests seem to be using the argument multiple times, nor calling clone. Am I missing something?

Is this change related to the moving the helper or orthogonal?

Is it needed/used by your follow-up/main PR?

frederikrothenberger: > None of the tests seem to be using the argument multiple times, nor calling clone. Am I missing something?

I think so? There is a line diff in `src/internet_identity/tests/integration/v2_api/authn_method_add.rs` that changes
```
create_identity_with_authn_method(&env, canister_id, authn_method.clone());
```

to 

```
let identity_number = create_identity_with_authn_method(&env, canister_id, &authn_method);
```
Similarly in the upcoming changeset.

> Is this change related to the moving the helper or orthogonal?

It does not directly have anything to do with moving, but goes in the same vein as it also makes the helper more useful (imho).

dskloetd: Ah, yes, I missed it.

It does mean that clone is now called more often than before, right? Does that go against the spirit or Rust? Anyway it's fine.

dskloetd (approved):

frederikrothenberger:
> It does mean that clone is now called more often than before, right? Does that go against the spirit or Rust?

Potentially yes. However, in most cases `clone` was called anyway (because the argument needed to be still available later). Also the integration tests are definitely not constrained by `clone`. 😉 
In this case, for better readability, I prefer simplifying the test code as much as possible even if it results in a few additional clones.

PR #1766, opened 2023-07-31
Remove magic numbers in storage test

frederikrothenberger:
There are currently magic numbers in the storage tests that change whenever a field is added to the Anchor type. The magic number can be replaced, by calculating
the length in the test itself.


dskloetd (approved):
Thanks!

PR #1772, opened 2023-08-06
Add metadata section "icp:public git_commit_id" to canister wasms

aterga:
This PR is a suggestion for adding a metadata section that marks the git_commit_id of this II canister build.

Intended usage:

On locally built canisters:
```
dfx canister metadata $IIC_CANISTER_ID git_commit_id
0000000000000000000000000000000000000000
```

On canisters built via CI:
```
dfx canister metadata $IIC_CANISTER_ID git_commit_id
# Outputs an actual git commit SHA that allows distinguishing canister versions via CLI
```

Inspired by https://sourcegraph.com/github.com/dfinity/ic/-/blob/bazel/canisters.bzl?L120

### DIFF
--- a/.github/actions/check-build/action.yml
+++ b/.github/actions/check-build/action.yml
@@ -12,6 +12,8 @@ runs:
     # run the build
     - run: npm ci
       shell: bash
+    - run: echo "${GITHUB_REF}" > git_commit_id.txt
+      shell: bash
     - run: ./scripts/build
       shell: bash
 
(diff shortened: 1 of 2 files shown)


frederikrothenberger:
Hi Arshavir

Could you give a little bit of context, what this metadata section is used for (other than just being consistent with canisters from the monorepo)?

Is it a requirement, to set `0000000000000000000000000000000000000000` for locally built canisters? Because the build script already has the II_VERSION variable (that is provided to give release information in the front-end), which is "$commit,$release,$dirty". The metadata field should consistent with that.

Is it also an option to just leave the field off completely instead of filling it with bogus content for locally / dirty built canisters?

nmattia:
Closing because this seems stale, feel free to reopen

PR #1776, opened 2023-08-08
Package json license and private

peterpeterparker:
# Motivation

Add flags to package.json for better definition.

## Private

`"private": true`

source: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#private

> If you set "private": true in your package.json, then npm will refuse to publish it.

## License

`"license": "SEE LICENSE IN LICENSE.md",`

source: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#license

> If you are using a license that hasn't been assigned an SPDX identifier, or if you are using a custom license, use a string value like this one...

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "type": "module",
   "description": "",
   "keywords": [],
+  "private": true,
   "license": "SEE LICENSE IN LICENSE.md",
   "scripts": {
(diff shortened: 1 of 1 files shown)


frederikrothenberger (approved):
Thanks! Sorry for the delay.

PR #1777, opened 2023-08-08
Introduce new key type variant for browser storage keys

frederikrothenberger:
The new key type variant is introduced as a preparation for the browser storage keys authentication mechanism.

The variant cannot be written, it is just introduced so that the upcoming release can be used as a rollback version to the release that allows writing the new variant.


przydatek (approved):
LGTM, thanks!

PR #1778, opened 2023-08-09
Retain key_type information in AuthnMethodData

frederikrothenberger:
This PR changes the AuthnMethodData usedin the API v2 to retain all information that is present in the key_type field of DeviceData.

The reason is that this is useful for stats purposes as well as for the front-end to handle certain authn methods differently.

The tests for the type conversion are expanded to
cover more cases.


przydatek (approved):
LGTM, thanks!

PR #1781, opened 2023-08-09
Improve candid interface tests

frederikrothenberger:
This PR replaces the existing subtype based candid interface tests with the simpler and more accurate `candid_equal` check.

The previuos method did not pick up type inequalities hidden by `opt` (which allows mismatched types, they will just be coerced to `null`). The `candid_equal` check was introduced with candid 0.9.

Using the new way of testing, uncovered two bugs in the candid files which this PR also fixes.


przydatek (approved):
LGTM, thanks!

PR #1786, opened 2023-08-10
Remove activity counter internal state from public interface

frederikrothenberger:
This PR removes the activity counter internal state from the public interface and moves the code into the internet identity crate.

This PR only _moves_ code. It will be followed by additional refactorings to separate the counting infrastructure more from the things being counted.


src/internet_identity/src/active_anchor_stats/activity_counter.rs
```diff
@@ -0,0 +1,9 @@
+use internet_identity_interface::internet_identity::types::Timestamp;
+
+pub mod active_anchor_counter;
+pub mod domain_active_anchor_counter;
+
+pub trait ActivityCounter: Clone {
```
przydatek, line 6: Maybe add some documentation, to explain the intended functionality of the trait?

przydatek (approved):
LGTM, modulo one suggestion. Thanks!

PR #1787, opened 2023-08-11
Refactor activity stats infrastructure to be independent of the activity being tracked

frederikrothenberger:
This PR refactors the activity counter infrastructure to no longer bother with the content of the counters. This is important because we want to introduce counters that do no longer count information per anchor but per authentication method.

Changes:
- Move the counting logic into the ActivityCounter trait
- Rename types, fields and comments mentioning anchors or users
- Provide an implementation for ActivityStats to do counting across all counters
  on the object rather than calling an external function


przydatek (approved):
Nice!
LGTM, thanks!

PR #1788, opened 2023-08-11
Improve upgrade tests for activity stats

frederikrothenberger:
Previously, there was no test that would catch if activity data was lost on upgrade. The reason for that being, that the field in persistent state is marked `opt` for compatibility reasons when it was newly introduced.

This PR improves the stats upgrade tests to catch such issues.


przydatek (approved):
LGTM, thanks!

PR #1789, opened 2023-08-11
Rename `active_anchor_stats` module

frederikrothenberger:
This PR is part of a bigger refactoring. This part is just renaming files so that the next PR does not rename _and_ modify.

The module is renamed because the code therein will no longer be used just for anchors but for other stats as well.


przydatek (approved):
LGTM, thanks!

PR #1791, opened 2023-08-14
Fix downloading nextest not failing on error

frederikrothenberger:
This PR sets the standard options to make a bash script fail more reliably to fix the pipe operator masking curl errors.


nmattia (approved):
Thanks!

PR #1792, opened 2023-08-14
Make the browser storage key key_type writeable

frederikrothenberger:
A new key type has been introduced internally (for rollback reasons)
in the last release. Now that it has been rolled out, the external
API is extended to allow reading and writing that key type.
It is intended to be used for the new browser storage based
authentication mechanism.

This PR introduces a breaking change on the candid interface:
* Old clients reading the new key type variant will fail when
  decoding the II responses. However, since the variant has to
  be written before it can be read, this should not cause issues.
* The test detecting breaking changes has been disabled until the
  next II release.
* The new API (v2) is designed in a way that interface extensions,
  such as this one, will no longer cause breaking changes.


src/internet_identity/tests/integration/anchor_management/device_management.rs
```diff
@@ -802,3 +802,24 @@ fn should_not_allow_reserved_metadata_keys() -> Result<(), CallError> {
     }
     Ok(())
 }
+
+/// Verifies that a device with key type BrowserStorageKey can be added.
+#[test]
+fn should_add_device_with_browser_storage_key_key_type() -> Result<(), CallError> {
```
przydatek, line 808: nit: maybe `should_add_device_with_key_type_browser_storage_key` (to avoid `...key_key_type`)

src/internet_identity/tests/integration/anchor_management/device_management.rs
```diff
@@ -802,3 +802,24 @@ fn should_not_allow_reserved_metadata_keys() -> Result<(), CallError> {
...
+    let device = DeviceData {
+        key_type: KeyType::BrowserStorageKey,
+        ..DeviceData::auth_test_device()
+    };
+
+    api::add(&env, canister_id, principal_1(), user_number, &device)?;
+
+    let mut devices =
+        api::get_anchor_info(&env, canister_id, principal_1(), user_number)?.into_device_data();
+    devices.sort_by(|a, b| a.pubkey.cmp(&b.pubkey));
```
przydatek, line 822: nit: maybe add a comment that `auth_test_device()` has pubkey with all 0, hence the sorting will put it first?

przydatek (approved):
LGTM, thanks!

PR #1793, opened 2023-08-14
Split encoding metrics into its own submodule

frederikrothenberger:
The metrics encoding code is getting more complex
as more metrics are added. With the new authentication methods stats being added, it makes sense to first clean up the code a bit.

This is the first step of more refactoring that moves the `encode_metrics` function unchanged and introduces a wrapper around it so that the caller does not need to know about the `MetricsEncoder`.


przydatek (approved):
LGTM, thanks!

PR #1796, opened 2023-08-14
Refactor metrics encoding derived from persistent state

frederikrothenberger:
This PR moves persistent state related metrics to their own function and introduces a helper for labelled activity stats.


src/internet_identity/src/http/metrics.rs
```diff
@@ -189,3 +123,92 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
+    Ok(())
+}
+
+fn labelled_activity_metrics<T: ActivityCounter>(
+    w: &mut MetricsEncoder<Vec<u8>>,
+    stats: &ActivityStats<T>,
+    daily_stats_label: &str,
+    daily_stats_doc: &str,
+    monthly_stats_label: &str,
+    monthly_stats_doc: &str,
```
przydatek, line 202: maybe package these arguments into a struct, so that one doesn't have to remember the order?

przydatek (approved):
LGTM, thanks!

PR #1799, opened 2023-08-16
Increase the number of jobs fetched when generating release notes

frederikrothenberger:
The release script currently only works due to fortunate ordering. Reordering the jobs might push some to the second page causing the release script to no longer find the `job_id`.

Increasing the number of jobs fetched to 100 gives us enough leeway to not run into the issue again immediately (we are currently at 36 jobs). We should anyway try to stay well below 100 jobs per pipeline.

### DIFF
--- a/.github/actions/release/run.sh
+++ b/.github/actions/release/run.sh
@@ -67,7 +67,10 @@ do
         html_url="https://example.com"
         step="step"
     else
-        job_id=$(curl --silent "https://api.github.com/repos/dfinity/internet-identity/actions/runs/$GITHUB_RUN_ID/jobs" \
+        # We have so many jobs in our CI run that we need to explicitly increase the number of jobs per page in order to
+        # not run into errors due to pagination (default is 30 jobs per page).
+        # See https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#list-jobs-for-a-workflow-run
+        job_id=$(curl --silent "https://api.github.com/repos/dfinity/internet-identity/actions/runs/$GITHUB_RUN_ID/jobs?per_page=100" \
             | jq -cMr \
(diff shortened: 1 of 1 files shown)


.github/actions/release/run.sh
```diff
@@ -67,7 +67,10 @@ do
         html_url="https://example.com"
         step="step"
     else
-        job_id=$(curl --silent "https://api.github.com/repos/dfinity/internet-identity/actions/runs/$GITHUB_RUN_ID/jobs" \
+        # We have so many jobs in our CI run that we need to explicitly increase the number of jobs per page in order to
+        # not run into errors due to pagination (default is 30 jobs per page).
+        # See https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#list-jobs-for-a-workflow-run
+        job_id=$(curl --silent "https://api.github.com/repos/dfinity/internet-identity/actions/runs/$GITHUB_RUN_ID/jobs?per_page=100" \
```
nmattia, line 73: probably best to paginate then wdyt?

nmattia (approved):
I think this will only push the can down the road but approving to unblock

.github/actions/release/run.sh
frederikrothenberger: I feel like more than 100 jobs is excessive for a single CI pipeline. Also, I'd rather not add more complexity to that script.

So, alternative suggestion: stop linking to the GitHub actions run in the release notes.
It does not seem that useful anyway because:
* people outside of the DFINITY organization cannot see the workflow logs, thus cannot verify that GitHub actions actually produced that hash
* trusted neurons verify the hash on their local machines anyway

WDYT?

nmattia: Don't care too strongly! I think using `github-script` would probably be the right thing to do

PR #1800, opened 2023-08-17
Collect and expose authentication method stats

frederikrothenberger:
For keys bound to an II domain, II will now collect the number of unique keys that were active daily and monthly, aggregated by key type.

The following key types are distinguished:
- WebAuthn authentication
- WebAuthn recovery
- Recovery phrase
- Browser storage key
- other (catch-all for anything not matching the above types)


src/internet_identity/src/activity_stats.rs
```diff
@@ -103,5 +102,11 @@ pub fn update_active_anchors_stats(anchor: &Anchor, current_domain: &Option<IIDo
                 .get_or_insert_with(|| ActivityStats::new(time()));
             stats.update_counters(&context);
         }
+
+        // Active authn methods stats, II domains only
+        let stats = persistent_state
+            .active_authn_method_stats
+            .get_or_insert_with(|| ActivityStats::new(time()));
+        stats.update_counters(&current_device);
```
przydatek, line 110: I see that in `AuthnMethodCounter::update_counters()` below we do an internal check whether on II domain (so the code here is correct), but the difference with the impl. of `DomainActiveAnchorCounter::update_counters()` (where we have a comment    `/// Only called if `current_domain` corresponds to an II domain.`) is confusing, especially that both cases are commented with "II domains only", but for the active anchor stats above we have the `if II domain..` and here not.  Maybe change `AuthnMethodCounter::update_counters()` so that `if ..` is not needed?

src/internet_identity/tests/integration/activity_stats/authn_methods.rs
```diff
@@ -0,0 +1,276 @@
...
+    assert!(
+        !get_metrics(&env, canister_id).contains("internet_identity_daily_active_authn_methods")
+    );
+
+    for (authn_type, authn_method) in authn_methods_all_types() {
+        // advance time to ensure a new collection period
+        env.advance_time(Duration::from_secs(DAY_SECONDS));
+
+        create_identity_with_authn_method(&env, canister_id, &authn_method);
+        let identity_nr = create_identity_with_authn_method(&env, canister_id, &authn_method);
```
przydatek, line 44: why calling `create_identity_with_authn_method` twice and ignoring the first identity?  (maybe add a comment)

src/internet_identity/tests/integration/activity_stats/authn_methods.rs
```diff
@@ -0,0 +1,276 @@
...
+    assert!(
+        !get_metrics(&env, canister_id).contains("internet_identity_daily_active_authn_methods")
+    );
+
+    for (authn_type, authn_method) in authn_methods_all_types() {
+        // advance time to ensure a new collection period
+        env.advance_time(Duration::from_secs(MONTH_SECONDS));
+
+        create_identity_with_authn_method(&env, canister_id, &authn_method);
+        let identity_nr = create_identity_with_authn_method(&env, canister_id, &authn_method);
```
przydatek, line 98: same here (double call to `create_identity_with_authn_method`)

przydatek (approved):
LGTM, modulo minor questions/suggestions.
Thanks!

src/canister_tests/src/framework.rs
```diff
@@ -470,7 +470,23 @@ pub fn parse_metric(body: &str, metric: &str) -> (f64, SystemTime) {
...
+}
+
+pub fn assert_labelled_metric(
+    metrics: &str,
+    metric_name: &str,
+    expected_value: f64,
+    label_name: &str,
+    label_values: &[&str],
+) {
+    for other_type in label_values {
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, this was a left-over. Thanks!

src/internet_identity/src/activity_stats.rs
frederikrothenberger: Ah, yes. Good point, changing that now. 👍

src/internet_identity/tests/integration/activity_stats/authn_methods.rs
frederikrothenberger: Ah, yes this is to make the counter have a value of `2` to make it distinct from an off-by-one error from `0`. 😉 
Added a comment.

PR #1803, opened 2023-08-18
Don't force captcha focus

nmattia:
The captcha was autofocused in the showcase, which made the page automatically scroll to it. This is now disabled, and the captcha is only focused when shown on an actually page in the webapp.


peterpeterparker (approved):
LGTM, thx

PR #1806, opened 2023-08-21
Introduce helper macro for handling variant results

frederikrothenberger:
This PR introduces a macro `match_value!` which allows casting an enum to a specific value (checked at runtime). This is useful especially for handling API v2 call results as they are all `opt varinat`.

The due to the way our formatter behaves, this macro does not actually lead to less lines, but much shorter ones.

### DIFF
--- a/src/canister_tests/src/framework.rs
+++ b/src/canister_tests/src/framework.rs
@@ -664,3 +664,17 @@ pub fn test_principal(n: u64) -> Principal {
     bytes.push(0xfe); // internal marker for user test ids
     Principal::from_slice(&bytes[..])
 }
+
+/// Helper macro to make API v2 return types easier to handle.
+/// In particular, API v2 calls all return variants, requiring a match on the result.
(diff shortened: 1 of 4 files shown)


src/canister_tests/src/framework.rs
```diff
@@ -664,3 +664,17 @@ pub fn test_principal(n: u64) -> Principal {
     bytes.push(0xfe); // internal marker for user test ids
     Principal::from_slice(&bytes[..])
 }
+
+/// Helper macro to make API v2 return types easier to handle.
+/// In particular, API v2 calls all return variants, requiring a match on the result.
+/// This macro allows to write the match in terms of the expected variant, with a fallback
+/// on unexpected variants.
+#[macro_export]
```
przydatek, line 672: nit: maybe add also a brief description of what the macro actually does, with an example?

przydatek (approved):
LGTM, thanks!

PR #1809, opened 2023-08-21
Add API v2 authn_method_remove method

frederikrothenberger:
Add the authn_method_remove method to slowly build the API v2.

### DIFF
--- a/src/internet_identity_interface/src/internet_identity/types/api_v2.rs
+++ b/src/internet_identity_interface/src/internet_identity/types/api_v2.rs
@@ -74,6 +74,12 @@ pub enum AuthnMethodAddResponse {
 }
 
 #[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub enum AuthnMethodRemoveResponse {
+    #[serde(rename = "ok")]
+    Ok,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
 pub enum IdentityMetadataReplaceResponse {
     #[serde(rename = "ok")]
     Ok,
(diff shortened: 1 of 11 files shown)


src/internet_identity_interface/src/internet_identity/types/api_v2.rs
```diff
@@ -73,6 +73,12 @@ pub enum AuthnMethodAddResponse {
     InvalidMetadata(String),
 }
 
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub enum AuthnMethodRemoveResponse {
```
przydatek, line 77: Would it make sense to return an error if the specified authn_method does not exist?

frederikrothenberger: It currently `traps` in that case. My idea was to first get a similar coverage on the API so that the front-ends could switch to v2 without losing functionality.
Then, we can start introducing additional error types. I think it makes more sense to do this holistically, because it affects a lot of other methods as well (i.e. the future `authn_method_replace`).

I'd like to split the work into smaller pieces and deal with errors later. Thankfully the APIv2 is designed to allow extending the API later in this way. 😉

przydatek: That makes sense, but maybe put a note about these plans in the code (e.g. a TODO) or in a JIRA ticket?

przydatek (approved):
LGTM, thanks!

src/internet_identity_interface/src/internet_identity/types/api_v2.rs
frederikrothenberger: I added it to the documentation in the `main.rs` file, because it applies to all the V2 API methods. 👍

PR #1810, opened 2023-08-22
Improve test names

frederikrothenberger:
This PR renames 2 tests that had bad or misleading names.


przydatek (approved):
Oops, the names were misleading indeed!
LGTM, thanks!

PR #1816, opened 2023-08-24
Fix PIN pasting

nmattia:
This prevents the browser from actually pasting anything. Without this, the browser tries to re-paste the entire input on the last element (which ended up being focused).

### DIFF
--- a/src/frontend/src/components/pinInput.ts
+++ b/src/frontend/src/components/pinInput.ts
@@ -207,15 +207,16 @@ const onPaste = ({
     return;
   }
 
-  // Actually set the value (for reasons unclear, preventDefault() is not necessary here as tested on
-  // Chrome, Safari)
+  // Set the values manually
   for (const [input, char] of toBePasted) {
     input.value = char;
   }
+  // Prevent actually pasting the value in any of the fields since we just did this manually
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
LGTM, thx

src/frontend/src/components/pinInput.ts
```diff
@@ -207,15 +207,16 @@ const onPaste = ({
...
   }
 
-  // Actually set the value (for reasons unclear, preventDefault() is not necessary here as tested on
-  // Chrome, Safari)
+  // Set the values manually
   for (const [input, char] of toBePasted) {
     input.value = char;
   }
+  // Prevent actually pasting the value in any of the fields since we just did this manually
+  event.preventDefault();
```
peterpeterparker, line 215: Note, `event` seems to be a reserved key if I get https://www.w3schools.com/js/js_reserved.asp correctly.
I generally use `$event` to avoid any unexpected issue even if transpilled but not sure it's relevant here.

nmattia: Thanks, I think you've mentioned this before (might have been about `error`?). I'll do a batch replace soon.

PR #1817, opened 2023-08-24
Fix selenium tests for paste

nmattia:
This fixes a paste test for the PIN input component, and introduces a aggregating job for all selenium tests that can be used to gate for branch protection.


frederikrothenberger (approved):
Thanks!

PR #1819, opened 2023-08-25
Fix outdated outputs on release build check step

frederikrothenberger:
I forgot to update the `outputs` section of the `latest-release` release build check CI workflow. This is the fix for that.

Additionally, there was a mixup between the dev and prod wasms, so I updated the naming to make it more clear as well.

However, the workflow should not rely on the dev build at all. This issue will be resolved in a separate PR.

Link to test-run, to be sure that I actually fixed it: https://github.com/dfinity/internet-identity/actions/runs/5973919511


nmattia (approved):
Not much context, rubber stamp

PR #1821, opened 2023-08-25
Delete previous char on backspace in PIN input

nmattia:
This adds a handler that captures the backspace keydown if the input is empty, moving to the previous input, and deleting the digit there.

https://github.com/dfinity/internet-identity/assets/6930756/eaa78d66-704c-4e95-a070-303eede76732

### DIFF
--- a/src/frontend/src/components/pinInput.ts
+++ b/src/frontend/src/components/pinInput.ts
@@ -242,7 +242,7 @@ const onKeydown = ({
 }) => {
   // If the input is empty, then move to the previous input
   if (event.code === "Backspace" && element.value === "") {
-    const didMove = focusPrevious(element);
+    const didMove = clearPrevious(element);
 
     if (didMove) {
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/pinInput.ts
```diff
@@ -258,3 +284,19 @@ const focusNextOrBlur = (element: HTMLInputElement) => {
     element.blur();
   }
 };
+
+// Focus the previous input in the list, if any. Returns true if the focus was moved
+// to a previous input.
+const focusPrevious = (element: HTMLInputElement): boolean => {
```
peterpeterparker, line 290: it does not just "focus" it also empty the value if I get it so it would be worth reflecting this in the method name for readability

peterpeterparker (approved):
LGTM, thx

src/frontend/src/components/pinInput.ts
nmattia: You're completely right; I added that at the last minute and forgot to update

PR #1823, opened 2023-08-25
Remove interface check exception

frederikrothenberger:
This PR removes a previously introduced exception to the interface check (which was necessary due to bad API choices in the past).


przydatek (approved):
LGTM, thanks!

PR #1825, opened 2023-08-25
Clean up close icon

nmattia:
We use a "close" (X) icon in toasts and in the dapps explorer.

The previous cross was pretty much an upper case X, which looked a bit odd. The scaling & the animating was inconsistent.

This replaces the base icon with a proper cross and makes the proportions & animations a bit nicer.

before/after/before/after:

https://github.com/dfinity/internet-identity/assets/6930756/46997230-3d95-43d9-afba-dfb35c654097

### DIFF
--- a/src/frontend/src/components/icons.ts
+++ b/src/frontend/src/components/icons.ts
@@ -97,21 +97,32 @@ export const icLogo = html`
 
 `;
 
-export const closeIcon = html`
-  <svg
-    class="closeIcon"
-    width="16"
-    height="16"
-    viewBox="0 0 16 16"
-    fill="none"
-    xmlns="http://www.w3.org/2000/svg"
-  >
-    <path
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/icons.ts
```diff
@@ -97,21 +97,32 @@ export const icLogo = html`
...
+      width="16"
+      height="16"
+      viewBox="0 0 16 16"
+      stroke="currentColor"
+      stroke-width="2"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        d="M${small} ${small} L${big} ${big} M${small} ${big} L${big} ${small}"
```
peterpeterparker, line 121: `fill` is gone, it's expected?

src/frontend/src/styles/main.css
```diff
@@ -933,6 +933,7 @@ by all browsers (FF is missing) */
 }
 
 .c-card__close {
+  --scale: 1;
```
peterpeterparker, line 936: don't know well the css of two but is `--scale` name a bit generic and therefore a higher chance to be overwritten? should it be something like `--c-scale`?

src/frontend/src/components/icons.ts
nmattia: Yep, wasn't actually used (was being overwritten by the CSS)

src/frontend/src/styles/main.css
nmattia: We don't have that many variables, I think `--scale` is fine for now

peterpeterparker (approved):
LGTM, thanks

PR #1827, opened 2023-08-28
Make max number of inflight captchas configurable

frederikrothenberger:
This PR fixes a long standing issue with a very long running integration test that maxes out a non configurable limit.
It is fixed by making the limit configurabel. This is done in preparation for the API v2 registration endpoint that would also have duplicated the long running test aggravating the issue further. Fixing this issue cuts down rust integration test time by about 90 seconds.

In addition, the storage tests for the persistent state are made more robust so that they no longer break on persistent state changes.

### DIFF
--- a/src/internet_identity/src/storage/tests.rs
+++ b/src/internet_identity/src/storage/tests.rs
@@ -349,7 +349,7 @@ fn should_overwrite_persistent_state_with_next_anchor() {
 fn should_read_previously_stored_persistent_state() {
     const NUM_ANCHORS: u64 = 3;
     const EXPECTED_ADDRESS: u64 = RESERVED_HEADER_BYTES + NUM_ANCHORS * 4096;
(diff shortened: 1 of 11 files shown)


src/internet_identity/src/storage/tests.rs
```diff
@@ -349,7 +349,7 @@ fn should_overwrite_persistent_state_with_next_anchor() {
 fn should_read_previously_stored_persistent_state() {
     const NUM_ANCHORS: u64 = 3;
     const EXPECTED_ADDRESS: u64 = RESERVED_HEADER_BYTES + NUM_ANCHORS * 4096;
-    const PERSISTENT_STATE_BYTES: &str = "4949505368010000000000004449444c116c03949d879d0701f7f5cbfb0778eed5f3af090a6b04dee7beb6080291a5fcf10a7fd1d3dab70b05c8bbeff50d066c01c2adc9be0c036c04c3f9fca002788beea8c5047881cfaef40a0487eb979d0d7a6d7b6c02d6a9bbae0a78c2adc9be0c036c02aaac8d930407c2adc9be0c036c03c7e8ccee037884fbf0820968cfd6ffea0f086d096c04c7e8ccee0378f2f099840704938da78c0a78d6a9bbae0a786e0b6c028bc3e2f9040cbbd492d8090f6c0297beb4cb080d8b858cea090d6e0e6c02fcdde6ea0178f9a6ebf703786c0297beb4cb08108b858cea090e6d0e0100032700000000000000010a00000000006000b0010100005847f80d0000001027000000000000206363636363636363636363636363636363636363636363636363636363636363e8038002e1df0200000001000163000000000000006dbb0e000000000001420000000000000030f1c520000000002c00000000000000d133b45001000000";
+    let persistent_state_bytes = candid::encode_one(&sample_persistent_state()).unwrap();
```
przydatek, line 352: Having a hard-coded persistent state example serves also as a stability test for the serialised persistent state, and changing to the on-the-fly generation of the state bytes somewhat looses the stability testing aspect, no?  
I'm aware that we have upgrade/downgrade integration tests that cover stability, but wonder whether it wouldn't be prudent to keep the hard-coded bytes (despite the extra overhead upon changes of the state).  Wdyt?

przydatek (approved):
LGTM, thanks!

src/internet_identity/src/storage/tests.rs
frederikrothenberger: We also have the hardcoded stable memory backups that test for stability. I don't think that we also need it here as well...

PR #1830, opened 2023-08-28
Rename 'event' to 'evnt'

nmattia:
This renames all occurrences of `event` to `evnt`. The word `event` shouldn't be used in JS code because it is a reserved keyword.

https://www.w3schools.com/js/js_reserved.asp
https://github.com/dfinity/internet-identity/pull/1816#discussion_r1304208748

### DIFF
--- a/src/frontend/src/components/pinInput.test.ts
+++ b/src/frontend/src/components/pinInput.test.ts
@@ -16,13 +16,13 @@ function dispatchInput(elem: HTMLInputElement, text: string) {
 
 // Dispatch an "ClipboardEvent"-ish event that works in jsdom
 function dispatchPaste(elem: HTMLInputElement, text: string) {
-  const event: any = new Event("paste", { bubbles: true, cancelable: true });
-  event.clipboardData = {
+  const evnt: any = new Event("paste", { bubbles: true, cancelable: true });
+  evnt.clipboardData = {
     getData() {
       return text;
     },
   };
-  elem.dispatchEvent(event);
+  elem.dispatchEvent(evnt);
 }
 
 // Get all inputs on the page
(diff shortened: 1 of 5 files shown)


src/frontend/src/components/pinInput.test.ts
```diff
@@ -16,13 +16,13 @@ function dispatchInput(elem: HTMLInputElement, text: string) {
 
 // Dispatch an "ClipboardEvent"-ish event that works in jsdom
 function dispatchPaste(elem: HTMLInputElement, text: string) {
-  const event: any = new Event("paste", { bubbles: true, cancelable: true });
-  event.clipboardData = {
+  const evnt: any = new Event("paste", { bubbles: true, cancelable: true });
```
peterpeterparker, line 19: not related: `any`?

peterpeterparker (approved):
LGTM

`$event` is my standard approach just in case interesting.

src/frontend/src/components/pinInput.test.ts
nmattia: had to force it to accept `.clipboardData`, which TypeScript doesn't know about

nmattia: @peterpeterparker the following is a tad safer, do you prefer that?

```
// Dispatch an "ClipboardEvent"-ish event that works in jsdom
function dispatchPaste(elem: HTMLInputElement, text: string) {
  const event = new Event("paste", { bubbles: true, cancelable: true });
  (event as any).clipboardData = {
    getData() {
      return text;
    },
  };
  elem.dispatchEvent(event);
}
```

peterpeterparker: `as ClipboardEvent` ?

peterpeterparker: or `as unknown as ClipboardEvent`?

peterpeterparker: ok

PR #1834, opened 2023-08-28
Don't mandate color support in the verify script

frederikrothenberger:
Github actions does not support colors / tput so this PR gates it behind a check.

### DIFF
--- a/scripts/verify-hash
+++ b/scripts/verify-hash
@@ -9,9 +9,15 @@ cd "$SCRIPTS_DIR/.."
 
 declare expected_ii_hash=
 declare expected_archive_hash=
-GREEN=$(tput setaf 2)
-RED=$(tput setaf 1)
-NC=$(tput sgr0)
+
+# Check that colors are supported (i.e. that we are running in a terminal)
+# I.e. github actions does not support colors
+if [ -n "$TERM" ]
+then
+    GREEN=$(tput setaf 2)
+    RED=$(tput setaf 1)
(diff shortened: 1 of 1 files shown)


nmattia (commented):
my 2cts: are the colors worth the extra complexity?

frederikrothenberger:
> my 2cts: are the colors worth the extra complexity?

In this case I think yes. This script is mainly made for humans and would just be run in CI to make sure we don't accidentally break it. Given the rest of the build generates a lot of output it is nice to separate the end-result of the verification using color.

przydatek (approved):
LGTM, thanks!

scripts/verify-hash
```diff
@@ -9,9 +9,15 @@ cd "$SCRIPTS_DIR/.."
...
-RED=$(tput setaf 1)
-NC=$(tput sgr0)
+
+# Check that colors are supported (i.e. that we are running in a terminal)
+# I.e. github actions does not support colors
+if [ -n "$TERM" ]
+then
+    GREEN=$(tput setaf 2)
+    RED=$(tput setaf 1)
+    NC=$(tput sgr0)
```
nmattia, line 19: `-u` is used, won't this trigger undefined variable errors?

frederikrothenberger: Thanks for making me recheck the script. Yes it does. I thought I tested it on CI,  but apparently I didn't.

Will submit the proper fix, with actual test-runs shortly.

PR #1836, opened 2023-08-29
Showcase registration flow

nmattia:
This refactors the `register` function to be able to showcase the registration flow without needing to spin up a canister.

A type parameter had to be added to `RegisterResult` to allow "mocking" the result without creating an actual canister connection.

The flow part is extracted from `register`, which is now a wrapper around this new `registerFlow`; `register` now only deals with canister/concrete implementation details whereas `registerFlow` only focuses on the user interaction.

### DIFF
--- a/src/frontend/src/flows/register/index.ts
+++ b/src/frontend/src/flows/register/index.ts
@@ -1,60 +1,124 @@
+import { Challenge } from "$generated/internet_identity_types";
 import {
+  LoginFlowCanceled,
+  LoginFlowResult,
   apiResultToLoginFlowResult,
   cancel,
-  LoginFlowResult,
 } from "$src/utils/flowResult";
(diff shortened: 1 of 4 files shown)


src/frontend/src/flows/register/index.ts
```diff
@@ -1,60 +1,124 @@
...
+import { ECDSAKeyIdentity } from "@dfinity/identity";
 import { nonNullish } from "@dfinity/utils";
 import type { UAParser } from "ua-parser-js";
-import { promptCaptcha } from "./captcha";
+import { badChallenge, precomputeFirst, promptCaptcha } from "./captcha";
 import { displayUserNumberWarmup } from "./finish";
 import { savePasskey } from "./passkey";
 
 /** Registration (anchor creation) flow for new users */
+export const registerFlow = async <T>({
```
nmattia, line 24: most of this is was just extracted from `register`

PR #1841, opened 2023-09-01
Showcase login flow

nmattia:
This extracts the flow essence of the authenticateBox so that the flow can be showcased on `/flows/loginManage`.

In practice, most canister calls were abstracted away so that the flow can be used without an actual canister. Similar to the registration flow showcased in https://github.com/dfinity/internet-identity/pull/1836.


frederikrothenberger (approved):
Rubber stamp 👍

PR #1844, opened 2023-09-05
Add crypto operations for browser-storage key

nmattia:
This introduces the cryptographic operations for storing an II keypair in browser storage.

The keypair is encrypted twice (once with a generated key, once with a key derived from a user PIN, see module) before being stored in IndexedDB.

The test-setup's crypto module is updated. The previous crypto module did not support `CryptoKey`.

### DIFF
--- a/src/frontend/src/crypto/pinIdentity.ts
+++ b/src/frontend/src/crypto/pinIdentity.ts
@@ -2,7 +2,7 @@ import { SignIdentity } from "@dfinity/agent";
 import { ECDSAKeyIdentity } from "@dfinity/identity";
 import { z } from "zod";
 
-/** This modules defines the crypto operations for the browser storage Identity.
+/** This module defines the crypto operations for the browser storage Identity.
  *
  * The identity is an ECDSA keypair. The keypair is meant to be indirectly stored in browser storage (IndexedDB).
  * Before being stored, the keypair in encrypted twice (symmetrically).
@@ -51,7 +51,7 @@ export const constructPinIdentity = async ({
   const browserIv = window.crypto.getRandomValues(new Uint8Array(96));
   const pinIv = window.crypto.getRandomValues(new Uint8Array(96));
   const pinSalt = window.crypto.getRandomValues(new Uint8Array(96));
-  const pinPbkdfIters: number = 10000;
+  const pinPbkdfIters: number = 100000;
   const keypairNamedCurve: NistEc = "P-256";
 
   const keypair = await generateKeyPair({ namedCurve: keypairNamedCurve });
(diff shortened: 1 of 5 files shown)


src/frontend/src/crypto/pinIdentity.ts
```diff
@@ -0,0 +1,215 @@
...
+}: {
+  pin: string;
+}): Promise<{
+  identity: SignIdentity;
+  pinIdentityMaterial: PinIdentityMaterial;
+}> => {
+  const browserIv = window.crypto.getRandomValues(new Uint8Array(96));
+  const pinIv = window.crypto.getRandomValues(new Uint8Array(96));
+  const pinSalt = window.crypto.getRandomValues(new Uint8Array(96));
+  const pinPbkdfIters: number = 10000;
```
nmattia, line 54: Please double check these values

src/frontend/src/crypto/pinIdentity.ts
```diff
@@ -0,0 +1,215 @@
...
+  const keypairNamedCurve: NistEc = "P-256";
+
+  const keypair = await generateKeyPair({ namedCurve: keypairNamedCurve });
+  const secretKey = keypair.getKeyPair().privateKey;
+
+  const browserKey = await generateBrowserKey();
+  const encryptedOnce = await crypto.subtle.wrapKey(
+    "pkcs8" /* the export format */,
```
nmattia, line 62: Please double check this, not sure if this or jwk is better

src/frontend/src/crypto/pinIdentity.ts
```diff
@@ -0,0 +1,215 @@
+import { SignIdentity } from "@dfinity/agent";
+import { ECDSAKeyIdentity } from "@dfinity/identity";
+import { z } from "zod";
```
frederikrothenberger, line 3: If we start using the library in our code, shouldn't we have a direct dependency on it too?

src/frontend/src/crypto/pinIdentity.ts
frederikrothenberger: I don't think this matters here, as long as we are consistent when importing and exporting.

frederikrothenberger (approved):
This looks good to me, thanks! Left a few minor comments. 

I would recommend trying to bump the number of PBKDF2 iterations by ~10x (maybe more, but requires testing on older phones to see how much delay it causes) to make brute-forcing more expensive.

This definitely requires a second approval by prod-sec and / or the crypto team.

src/frontend/src/crypto/pinIdentity.ts
nmattia: bumped to 10000, we'll see what happens when we do some manual testing before releasing

nmattia:
@tmu0 addressed here: https://github.com/dfinity/internet-identity/pull/1861

PR #1846, opened 2023-09-05
Remove CAPTCHA flow from components showcase

nmattia:
This removes the "complete CAPTCHA" flow from the components showcase; the flow is already showcased in the "flows" section (`/flows/captcha`).


peterpeterparker (approved):
Rubber stamp

PR #1853, opened 2023-09-07
Clarify filetype error

nmattia:
This clarifies the error message in case on unknown asset type. The previous error showed the asset basename, suggesting this had been used as the extension.

The new error message clarifies what was assumed to be the filetype/extension, and also clarifies word (string) boundaries to avoid subtle space/invisible char issues.


frederikrothenberger (approved):
LGTM, thanks!

PR #1869, opened 2023-09-11
Add workaround for node errors in show-case

frederikrothenberger:
Adding the PIN identity functionality to the showcase leads to breakage of the astro build due to node not supporting subtle crypto and indexeddb.
This PR creates an ugly temporary workaround to unblock feature development, but needs to be cleaned up after.


src/showcase/src/pages/flows/[flow].astro
```diff
@@ -1,10 +1,10 @@
 ---
 import Layout from "../../layouts/Layout.astro";
 import "$src/styles/main.css";
-import { iiFlows } from "$showcase/flows";
 
+export const iiFlowNames = ["loginManage", "register"];
 export function getStaticPaths() {
-  return Object.keys(iiFlows).map((key) => ({ params: { flow: key } }));
+  return iiFlowNames.map((name) => ({ params: { flow: name } }));
```
Comment (peterpeterparker) [inline src/showcase/src/pages/flows/[flow].astro:7]: Maybe you can use another variable then `name`. It's commonly advised to not use such a name.

https://www.w3schools.com/js/js_reserved.asp

> JavaScript Objects, Properties, and Methods
> You should also avoid using the name of JavaScript built-in objects, properties, and methods: 
> ...
> name

Maybe `key`

peterpeterparker (approved):
LGTM, thanks.

Minor comment if you can have a look before merge.

PR #1872, opened 2023-09-13
Add support for dynamic error keys to pin input component

frederikrothenberger:
Support for dynamic keys is required to make the pin input compatible with the error texts provided by `LoginFlowError`.

### DIFF
--- a/src/frontend/src/components/pinInput.ts
+++ b/src/frontend/src/components/pinInput.ts
@@ -1,4 +1,5 @@
 import { toast } from "$src/components/toast";
+import { DynamicKey } from "$src/utils/i18n";
 import { mount, withRef } from "$src/utils/lit-html";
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/pinInput.ts
```diff
@@ -1,4 +1,5 @@
 import { toast } from "$src/components/toast";
+import { DynamicKey } from "$src/utils/i18n";
```
peterpeterparker, line 2: don't know if you use this in II but, `import type { DynamicKey } from "$src/utils/i18n";` would work here

peterpeterparker (approved):
LGTM

src/frontend/src/components/pinInput.ts
frederikrothenberger: It is, thanks!

PR #1873, opened 2023-09-13
Add fake indexeddb to support testing pin input

frederikrothenberger:
This is a PR in preparation for adding the PIN auth feature, which will require support for indexeddb in tests.


peterpeterparker (approved):
LGTM, thanks

PR #1874, opened 2023-09-13
Provide verify function externally to usePin flow

frederikrothenberger:
This PR expands the usePin function to take a PIN verification function as an argument. This will later be used to immediately check whether the identity can be decrypted using the provided PIN.

### DIFF
--- a/src/frontend/src/flows/pin/usePin.ts
+++ b/src/frontend/src/flows/pin/usePin.ts
@@ -1,23 +1,25 @@
 import { mainWindow } from "$src/components/mainWindow";
-import { pinInput } from "$src/components/pinInput";
+import { pinInput, PinResult } from "$src/components/pinInput";
 import { I18n } from "$src/i18n";
 import { mount, renderPage } from "$src/utils/lit-html";
-import { TemplateResult, html } from "lit-html";
+import { html, TemplateResult } from "lit-html";
 
 import copyJson from "./usePin.json";
 
 /* Prompt the user to input their PIN for use */
 
-const usePinTemplate = ({
+const usePinTemplate = <T>({
   i18n,
   cancel,
+  verify,
   onContinue,
   onUsePasskey,
   scrollToTop = false,
 }: {
   i18n: I18n;
   cancel: () => void;
-  onContinue: (pin: string) => void;
+  verify: (pin: string) => Promise<PinResult<T>>;
+  onContinue: (result: T) => void;
   onUsePasskey: () => void;
   /* put the page into view */
   scrollToTop?: boolean;
@@ -25,9 +27,7 @@ const usePinTemplate = ({
   const copy = i18n.i18n(copyJson);
   const pinInput_ = pinInput({
     onSubmit: onContinue,
-    verify: (pin) =>
-      // XXX: here we don't do any verification, but ideally the identity reconstruction should happen here
-      ({ ok: true, value: pin }),
+    verify: verify,
     secret: true,
   });
   const slot = html`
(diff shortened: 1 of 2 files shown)


peterpeterparker (commented):
Any particular reason to use a weak generic type? Can it be narrowed?

frederikrothenberger:
> Any particular reason to use a weak generic type? Can it be narrowed?

The pinInput is a generic component that does not care _what_ you do with the PIN. So having the verify function be able to construct any type (or fail) and `onContinue` consume any type gives the most flexiblity.

In particular, the showcase currently uses `string` whereas the production code will probably use `LoginFlowSuccess`. These types do not have any intersection.

src/frontend/src/flows/pin/usePin.ts
```diff
@@ -1,33 +1,33 @@
...
   /* put the page into view */
   scrollToTop?: boolean;
 }): TemplateResult => {
   const copy = i18n.i18n(copyJson);
   const pinInput_ = pinInput({
     onSubmit: onContinue,
-    verify: (pin) =>
-      // XXX: here we don't do any verification, but ideally the identity reconstruction should happen here
-      ({ ok: true, value: pin }),
+    verify: verify,
```
peterpeterparker, line 30: can be shortened to `verify`

src/frontend/src/flows/pin/usePin.ts
```diff
@@ -1,33 +1,33 @@
 import { mainWindow } from "$src/components/mainWindow";
-import { pinInput } from "$src/components/pinInput";
+import { pinInput, PinResult } from "$src/components/pinInput";
 import { I18n } from "$src/i18n";
 import { mount, renderPage } from "$src/utils/lit-html";
-import { TemplateResult, html } from "lit-html";
+import { html, TemplateResult } from "lit-html";
```
peterpeterparker, line 5: `type TemplateResult`?

src/frontend/src/flows/pin/usePin.ts
```diff
@@ -1,33 +1,33 @@
...
   cancel,
+  verify,
   onContinue,
   onUsePasskey,
   scrollToTop = false,
 }: {
   i18n: I18n;
   cancel: () => void;
-  onContinue: (pin: string) => void;
+  verify: (pin: string) => Promise<PinResult<T>>;
```
peterpeterparker, line 21: not sure, should actually the generic be repeated on the function level?

```
verify: <T>(pin: string) => Promise<PinResult<T>>;
onContinue: <T>(result: T) => void;
```

src/frontend/src/flows/pin/usePin.ts
```diff
@@ -57,14 +57,19 @@ const usePinTemplate = ({
...
 
 export const usePinPage = renderPage(usePinTemplate);
-export const usePin = (): Promise<
-  { kind: "pin"; pin: string } | { kind: "canceled" } | { kind: "passkey" }
+export const usePin = <T>({
+  verifyPin,
+}: {
+  verifyPin: (pin: string) => Promise<PinResult<T>>;
+}): Promise<
+  { kind: "pin"; result: T } | { kind: "canceled" } | { kind: "passkey" }
```
peterpeterparker, line 65: it used to returns `{ kind: "pin"; pin: string }` but now it returns ` { kind: "pin"; result: T }`. Changing the variable name from `pin` to `result` is expected?

src/frontend/src/flows/pin/usePin.ts
```diff
@@ -57,14 +57,19 @@ const usePinTemplate = ({
 };
 
 export const usePinPage = renderPage(usePinTemplate);
```
peterpeterparker, line 59: Is `usePinPage` still used somewhere else?

src/frontend/src/flows/pin/usePin.ts
frederikrothenberger: Yes, because `verify` can transform the value now. I.e. the result indicates that the `pin` was successfully used to get some `result`.

I'm open to rename stuff, if it is confusing now.

src/frontend/src/flows/pin/usePin.ts
peterpeterparker: Gotcha

peterpeterparker (approved):
LGTm, thx

PR #1875, opened 2023-09-13
Require exactly one issuer in relying party's request.

przydatek:
Modify the VC-spec to require exactly one issuer in the request from relying party.


frederikrothenberger (approved):
Thanks!

PR #1878, opened 2023-09-14
Refactor register to allow setting key type

frederikrothenberger:
This PR is in preparation for pin auth registration. For this feature, we need to have more control over the key type submitted to II.

### DIFF
--- a/src/frontend/src/flows/register/index.ts
+++ b/src/frontend/src/flows/register/index.ts
@@ -1,6 +1,11 @@
-import { Challenge } from "$generated/internet_identity_types";
+import {
+  Challenge,
+  CredentialId,
+  KeyType,
+} from "$generated/internet_identity_types";
 import { registerStepper } from "$src/flows/register/stepper";
 import { registerDisabled } from "$src/flows/registerDisabled";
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/register/index.ts
```diff
@@ -62,6 +70,10 @@ export const registerFlow = async <T>({
       const result = await register({
         identity,
         alias,
+        keyType: authenticatorAttachmentToKeyType(
```
peterpeterparker, line 73: The `identity` is already passed to the `register` function, shouldn't the `keytype` be computed within that function rather than extending its declaration?

frederikrothenberger: Ah, this also includes a change to have less constraints on the `identity`. Previously, we had an `IIWebAuthnIdentity` which had information attached about the `KeyType`. Now, we only pass a `SignIdentity` (in preparation of the PIN key _not_ being an `IIWebAuthnIdentity`).

peterpeterparker: Ummm ok. Not sure it makes the code that readable from having a look at this PR but you know better.

peterpeterparker (approved):
LGTM, thx

PR #1879, opened 2023-09-14
Allow pin registration for Apple devices

frederikrothenberger:
This PR enables registration of PIN protected browser keys when signing up for II. The feature is enabled on Apple devices only, because it is meant as a workaround for the forced Apple iCloud integration.

When adding a device later, the PIN protected browser key is not allowed, even on Apple devices.

The feature is still very basic. In particular, still missing are:
- the temporary key info screen
- updates to the management page to correctly list temporary keys separate from passkeys
- warnings on the management screen

Almost all of the code is taken from the `nm-set-pin` branch by @nmattia.


peterpeterparker (approved):
Rubber stamp

PR #1880, opened 2023-09-14
Show pin protected keys in a separate list on the management page

frederikrothenberger:
This allows users to distinguish between the temporary keys and passkeys when visiting the management page.

### DIFF
--- a/src/frontend/src/flows/manage/authenticatorsSection.ts
+++ b/src/frontend/src/flows/manage/authenticatorsSection.ts
@@ -109,6 +109,32 @@ export const authenticatorsSection = ({
     </aside>`;
 };
 
+export const tempKeysSection = ({
+  authenticators: authenticators_,
+}: {
+  authenticators: Authenticator[];
+}): TemplateResult => {
+  const authenticators = dedupLabels(authenticators_);
+
+  return html` <aside class="l-stack c-card c-card--narrow">
+    <div class="t-title t-title--complications">
+      <h2 class="t-title">Temporary Keys</h2>
+    </div>
+
+    <p style="max-width: 30rem;" class="t-paragraph t-lead">
+      These keys are stored in browser cache, clearing your browser storage will
+      erase them.
+    </p>
+    <div class="c-action-list">
+      <ul>
+        ${authenticators.map((authenticator, index) =>
+          authenticatorItem({ authenticator, index })
+        )}
+      </ul>
+    </div>
+  </aside>`;
+};
+
 export const authenticatorItem = ({
   authenticator: { alias, dupCount, warn, remove, rename },
   index,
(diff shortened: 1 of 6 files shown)


src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -109,6 +109,32 @@ export const authenticatorsSection = ({
...
+  authenticators: Authenticator[];
+}): TemplateResult => {
+  const authenticators = dedupLabels(authenticators_);
+
+  return html` <aside class="l-stack c-card c-card--narrow">
+    <div class="t-title t-title--complications">
+      <h2 class="t-title">Temporary Keys</h2>
+    </div>
+
+    <p style="max-width: 30rem;" class="t-paragraph t-lead">
```
peterpeterparker, line 124: why using a `style` and not a class? is that common in the code base?

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -109,6 +109,32 @@ export const authenticatorsSection = ({
...
+      These keys are stored in browser cache, clearing your browser storage will
```
peterpeterparker, line 125: i18n (generally speaking for any text provideed or modified in this PR)?

or this will be extracted later on?

src/frontend/src/flows/manage/index.ts
```diff
@@ -446,17 +452,29 @@ export const devicesFromDeviceDatas = ({
...
         alias: device.alias,
         warn: domainWarning(device),
         rename: () => renameDevice({ connection, device, reload }),
         remove: hasSingleDevice
           ? undefined
           : () => deleteDevice({ connection, device, reload }),
-      });
+      };
+
+      if ("browser_storage_key" in device.key_type) {
```
peterpeterparker, line 464: I am not a great fan of these kind of imperative code, it's prone to issues because we push within an existing object.

Is there a way to instantiate the object with the appropriate values? e.g. using deconstruction and inner condition or else

frederikrothenberger: Are you talking about the usage of `reduce` and pushing into the accumulator? If so, I'd rather not change that (as it was like that before).

src/frontend/src/flows/manage/authenticatorsSection.ts
frederikrothenberger: This is analogous to the `authenticatorsSection` above (see line 79).

Also, I'm definitely not the guy to do CSS. 🙃

src/frontend/src/flows/manage/authenticatorsSection.ts
frederikrothenberger: I don't want to change more than I need to, so I kept to the way that the `authenticatorsSection` works (which also does not have copy extracted).

This should definitely be extracted later.

src/frontend/src/flows/manage/authenticatorsSection.ts
peterpeterparker: Ok

src/frontend/src/flows/manage/index.ts
peterpeterparker: Ok

peterpeterparker (approved):
Rubber stamp

PR #1882, opened 2023-09-14
Enable pin info page

frederikrothenberger:
This PR adds the pin info page to the pin registration flow.

### DIFF
--- a/src/frontend/src/flows/register/index.ts
+++ b/src/frontend/src/flows/register/index.ts
@@ -10,6 +10,7 @@ import {
 } from "$src/crypto/pinIdentity";
 import { confirmPin } from "$src/flows/pin/confirmPin";
 import { idbStorePinIdentityMaterial } from "$src/flows/pin/idb";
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/register/index.ts
```diff
@@ -76,6 +77,11 @@ export const registerFlow = async <T>({
   }
   const result_ = await (async () => {
```
peterpeterparker, line 78: it's me or the function as no return type defined?

frederikrothenberger: Yes, although the result is later destructured on lines 133 to 147 (after the `canceled` case has been handled).

peterpeterparker (approved):
Rubber stamp

src/frontend/src/flows/register/index.ts
peterpeterparker: Ok

PR #1883, opened 2023-09-15
Extract temporary key template and copy

frederikrothenberger:
Additional warnings, etc. need to be added.
Extracting copy and separating everything related to temp keys into its own file makes it easier to modify later.


peterpeterparker (approved):
Rubber stamp

PR #1885, opened 2023-09-15
Add warning to PIN info screen

frederikrothenberger:
This adds the warning as shown in the figmas to the PIN info screen.

### DIFF
--- a/src/frontend/src/flows/pin/pinInfo.json
+++ b/src/frontend/src/flows/pin/pinInfo.json
@@ -5,6 +5,10 @@
     "create_temporary_key_form_pin": "Create a temporary key from a PIN",
     "set_memorable_pin": "Set a memorable PIN and securely store a temporary key in your browser cache",
 
+    "security_warning": "Security Warning",
+    "are_you_sure": "Are you sure you want to set up a temporary key?",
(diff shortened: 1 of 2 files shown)


src/frontend/src/flows/pin/pinInfo.json
```diff
@@ -5,6 +5,10 @@
     "create_temporary_key_form_pin": "Create a temporary key from a PIN",
     "set_memorable_pin": "Set a memorable PIN and securely store a temporary key in your browser cache",
 
+    "security_warning": "Security Warning",
+    "are_you_sure": "Are you sure you want to set up a temporary key?",
+    "clear_browser_storage_add_passkey": "Clearing your browser storage will erase this key. Add a passkey if you want to hold assets with this identity.",
```
peterpeterparker, line 10: Should this even be possible? i.e. shouldn't the UX be "You cannot delete your key, add a passkey first?"

peterpeterparker (approved):
Rubber stamp

src/frontend/src/flows/pin/pinInfo.json
frederikrothenberger: Well, we cannot prevent users from clearing the browser cache in the browser settings.

Of course you cannot remove the temp key on the management page of II until you have added something else. But that's not what's meant here.

PR #1886, opened 2023-09-15
Change PIN registration stepper to 3 steps

frederikrothenberger:
It also renames the first step from "set" to "set_pin" because `set` is a reserved keyword in js.


peterpeterparker (approved):
Thx

PR #1887, opened 2023-09-15
Make stepper on registration finish step dependent on the auth flow

frederikrothenberger:
This PR extracts the stepper on the finish page and adjusts it to show the appropriate steps depending on whether a passkey or a pin protected storage key was registered.


peterpeterparker (approved):
LGTM, thx

PR #1889, opened 2023-09-15
Move temp keys section above passkeys section in manage view

frederikrothenberger:
According to the figmans, the temp keys need to be shown first.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -192,14 +192,14 @@ const displayManageTemplate = ({
         },
       })}
     </p>
+    ${pinAuthenticators.length > 0
+      ? tempKeysSection({ authenticators: pinAuthenticators, i18n })
+      : ""}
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/manage/index.ts
```diff
@@ -192,14 +192,14 @@ const displayManageTemplate = ({
         },
       })}
     </p>
+    ${pinAuthenticators.length > 0
```
peterpeterparker, line 195: btw. would following work?

```
${pinAuthenticators.length > 0 && tempKeysSection({ authenticators: pinAuthenticators, i18n })}
```

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/flows/manage/index.ts
frederikrothenberger: No, it will then print "false" into the dom, if there are no temp keys.

PR #1888, opened 2023-09-15
Highlight recovery box on no recoveries

frederikrothenberger:
If there are no recovery methods, the recovery section is now highlighted with a warning.

The code for the warning highlight box has been copied from the `authenticatorsSection`.
The warning box wrapper should be extracted as a common helper function in the future.

<hr/><details><summary>🟡 Some screens were changed</summary></details>

PR #1890, opened 2023-09-15
Highlight passkey warning only on 0 passkeys

frederikrothenberger:
This changes the warning highlight around passkeys to only appear if the number of passkeys is zero. Also, there is now a text shown for passkeys in the non-warning state.

Together with #1888 this shifts the warning box highlight on an identity with just a single passkey from the passkey box to the recovery box.

### DIFF
--- a/src/frontend/src/flows/manage/authenticatorsSection.ts
+++ b/src/frontend/src/flows/manage/authenticatorsSection.ts
@@ -33,16 +33,16 @@ export const dedupLabels = (
 export const authenticatorsSection = ({
   authenticators: authenticators_,
   onAddDevice,
-  warnFewDevices,
+  warnNoPasskeys,
 }: {
   authenticators: Authenticator[];
   onAddDevice: () => void;
-  warnFewDevices: boolean;
+  warnNoPasskeys: boolean;
 }): TemplateResult => {
-  const wrapClasses = ["l-stack"];
+  const wrapClasses = ["l-stack", "c-card", "c-card--narrow"];
 
-  if (warnFewDevices) {
-    wrapClasses.push("c-card", "c-card--narrow", "c-card--warning");
+  if (warnNoPasskeys) {
+    wrapClasses.push("c-card--warning");
   }
 
   const authenticators = dedupLabels(authenticators_);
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -76,15 +76,18 @@ export const authenticatorsSection = ({
           </span>
         </div>
         ${
-          warnFewDevices
+          warnNoPasskeys
```
peterpeterparker, line 79: ummm would it be cleaner to have a single `html` parser and the condition within the string?

```
html`<p
                style="max-width: 30rem;"
                ${warnNoPasskeys ? class="warning-message t-paragraph t-lead" : ''}
              >
              ...
```
              
what do you think?

peterpeterparker (commented):
Is this a good idea? I mean feels like we are lowering the UX security for all existing users here.

frederikrothenberger:
> Is this a good idea? I mean feels like we are lowering the UX security for all existing users here.

No, we are just shifting the warning to a different box. See [here](https://www.figma.com/file/a8jl32LDcYODRwwzrZd5UU/Internet-Identity-Design?type=design&node-id=2081%3A21903&mode=design&t=VhqFhLoXlAUzW7u3-1). PR #1888 adds the highlight to the recovery box.

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -33,24 +33,24 @@ export const dedupLabels = (
...
   onAddDevice: () => void;
-  warnFewDevices: boolean;
+  warnNoPasskeys: boolean;
 }): TemplateResult => {
-  const wrapClasses = ["l-stack"];
+  const wrapClasses = ["l-stack", "c-card", "c-card--narrow"];
 
-  if (warnFewDevices) {
-    wrapClasses.push("c-card", "c-card--narrow", "c-card--warning");
+  if (warnNoPasskeys) {
```
peterpeterparker, line 44: I would not be against sparing this "push"

peterpeterparker (approved):
Rubber stamp

PR #1893, opened 2023-09-18
Fix dapps update job not creating a PR

frederikrothenberger:
The PR action of the dapps update workflow skipped on a variable that was never set, thus never actually updating the dapps file.


przydatek (approved):
LGTM, thanks!

PR #1896, opened 2023-09-18
Remove openssl install step in canister tests CI job

frederikrothenberger:
The macos runners no longer require installation of `openssl`, because the component is now preinstalled. This shaves of some CI time and makes it more consistent, as the installation process was dependent on network speed (sometimes it would hang for up to 1 minute).


przydatek (approved):
LGTM, nice!

PR #1895, opened 2023-09-18
Fix formatting of updated dapps.json file

frederikrothenberger:
The dapps.json update could trigger the formatter on badly formatted json. This PR fixes the job to create the PR with acceptable formatting in the first place.

<hr/><details><summary>🟡 Some screens were changed</summary></details>

PR #1897, opened 2023-09-18
Install dependencies in dapps update

frederikrothenberger:
The formatter needs to be installed first before it can be run in the dapps update workflow. Fixes oversight in #1895.


przydatek (approved):
LGTM, thanks!

PR #1898, opened 2023-09-18
Support JPG assets

frederikrothenberger:
The dapps list now contains logos in jpg format.
This PR adds support for jpg images so that they can be included.


peterpeterparker:
Probably opiniated but, I think we should rather exclude JPG and dapp that uses such format.
These are weak in terms of performance.
Portal should enforce SVG or Webp and community should provide such types of assets.

PR #1904, opened 2023-09-19
Add PIN login e2e test

frederikrothenberger:
This PR expands the previous test to also
include a login scenario. In addtion, the asserts are improved to now check that the temp key / passkey is listed in the correct section.


peterpeterparker (approved):
Rubber stamp

PR #1907, opened 2023-09-19
Fix being prompted for PIN input after temporary key is removed

frederikrothenberger:
When deleting the temporary key from the identity, the key remains in browser storage. This leads to the awkward situation of the browser still prompting for the PIN even if subsequent authentication will fail.

This PR introduces a check, that the browser will only use the PIN protected key if the public key is still present on the identity.


peterpeterparker (approved):
Rubber stamp

PR #1908, opened 2023-09-19
Improve confirm pin flow on pin mismatch

frederikrothenberger:
Previously, if the pin on the confirmation page did not match the previous input the only option was to cancel and start over. This is a very bad user experience.

The behaviour is replaced with a mechanism that replaces the cancel button with a retry button that sends the user to the previous page.


peterpeterparker (approved):
Rubber stamp

PR #1909, opened 2023-09-20
Add more e2e test for non-passkey auth

frederikrothenberger:
This PR adds more e2e tests for the PIN auth feature:
* Login attempt with wrong PIN first
* PIN registration during auth flow
* auth into client application after PIN registration


src/frontend/src/test-e2e/views.ts
```diff
@@ -168,13 +168,17 @@ export class PinAuthView extends View {
...
+      .waitForDisplayed({ timeout: 5_000 });
+  }
 }
 
 export class RecoveryMethodSelectorView extends View {
   async waitForDisplay(): Promise<void> {
-    await this.browser
-      .$('[data-action="cancel"]')
-      .waitForDisplayed({ timeout: 10_000 });
+    await this.browser.$('[data-page="add-recovery-phrase"]').waitForExist();
```
frederikrothenberger, line 181: This needed to change, because the new test exposed that quite a few pages do have a cancel button. 😉

src/frontend/src/test-e2e/pinAuth.test.ts
```diff
@@ -61,3 +110,63 @@ test("Should not prompt for PIN after deleting temp key", async () => {
     await FLOWS.login(userNumber, DEVICE_NAME1, browser);
   }, APPLE_USER_AGENT);
 }, 300_000);
+
+test("Log into client application, after PIN registration", async () => {
```
lmuntaner, line 114: I find this description confusing. Maybe `"Log into client application using PIN registration flow"`?

src/frontend/src/test-e2e/pinAuth.test.ts
```diff
@@ -61,3 +110,63 @@ test("Should not prompt for PIN after deleting temp key", async () => {
...
+    await demoAppView.open(TEST_APP_NICE_URL, II_URL);
+    await demoAppView.waitForDisplay();
+    expect(await demoAppView.getPrincipal()).toBe("2vxsx-fae");
+    await demoAppView.signin();
+    await switchToPopup(browser);
+    await FLOWS.registerPinNewIdentityAuthenticateView(pin, browser);
+    await waitToClose(browser);
+    await demoAppView.waitForDisplay();
+    const principal = await demoAppView.getPrincipal();
+    expect(principal).not.toBe("2vxsx-fae");
```
lmuntaner, line 128: Where does `"2vxsx-fae"` come from?

src/frontend/src/test-e2e/pinAuth.test.ts
```diff
@@ -39,6 +58,36 @@ test("Register and Log in with PIN identity", async () => {
...
+    await mainView.logout();
+
+    const welcomeView = new WelcomeView(browser);
+    await welcomeView.waitForDisplay();
+    await welcomeView.login();
+    await welcomeView.typeUserNumber(userNumber);
+    await browser.$("button[data-action='continue']").click();
+    const pinAuthView = new PinAuthView(browser);
+    await pinAuthView.waitForDisplay();
+    await pinAuthView.enterPin("456321"); // wrong PIN
```
lmuntaner, line 78: I would create a `wrongPin` variable next to the `pin` variable. This way you don't need the comment.

src/frontend/src/test-e2e/pinAuth.test.ts
```diff
@@ -39,6 +58,36 @@ test("Register and Log in with PIN identity", async () => {
...
+    await pinAuthView.waitForError();
+    await pinAuthView.enterPin(pin); // enter correct PIN
```
lmuntaner, line 80: Ditto, comment not needed then

src/frontend/src/test-e2e/pinAuth.test.ts
```diff
@@ -61,3 +110,63 @@ test("Should not prompt for PIN after deleting temp key", async () => {
...
+  await runInBrowser(async (browser: WebdriverIO.Browser) => {
+    const pin = "123456";
+
+    await browser.url(II_URL);
+    const userNumber = await FLOWS.registerPinWelcomeView(browser, pin);
+
+    const demoAppView = new DemoAppView(browser);
+    await demoAppView.open(TEST_APP_NICE_URL, II_URL);
+    await demoAppView.waitForDisplay();
+    expect(await demoAppView.getPrincipal()).toBe("2vxsx-fae");
```
lmuntaner, line 150: Ditto, where does this principal come from?

src/frontend/src/test-e2e/views.ts
```diff
@@ -168,13 +168,17 @@ export class PinAuthView extends View {
       await input.setValue(digit);
     }
   }
+
+  async waitForError(): Promise<void> {
+    await this.browser
+      .$('[data-haserror="true"]')
```
lmuntaner, line 174: Is this `[data-haserror="true"]` something global in II?

Could you make create a class variable `WelcomeView.ErrorSelector` and use it here?

src/frontend/src/test-e2e/views.ts
lmuntaner: Ditto to use a class variable `RecoveryMethodSelectorView.selector`?

lmuntaner (approved):
A few comments, but already approving.

We have the rule that PRs can't be merged without resolving all comments. Therefore, we can approve while asking for changes.

src/frontend/src/test-e2e/pinAuth.test.ts
frederikrothenberger: It's the anonymous principal. I replaced it with a call to `Principal.anonymous()`.

src/frontend/src/test-e2e/pinAuth.test.ts
frederikrothenberger: See above, fixed.

src/frontend/src/test-e2e/views.ts
frederikrothenberger: No, it is only used in this instance here. But I can extract a variable nonetheless. 👍

PR #1912, opened 2023-09-20
Add header slot to warn box

frederikrothenberger:
Extends the `warnBox` helper with a slot to add some template next to the exclamation mark icon.

The PR will make the helper usable in more contexts relate to the non-passkey auth feature.

### DIFF
--- a/src/frontend/src/components/warnBox.ts
+++ b/src/frontend/src/components/warnBox.ts
@@ -9,6 +9,7 @@ import { warningIcon } from "./icons";
 
 interface warnBoxProps {
   title: TemplateElement;
+  headerSlot?: TemplateElement;
   message: TemplateElement;
   slot?: TemplateResult;
   htmlElement?: "div" | "aside";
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/warnBox.ts
```diff
@@ -28,7 +30,10 @@ export const warnBox = ({
   }
   const contents: TemplateResult = html`
     <span class="c-card__label c-card__label--hasIcon" aria-hidden="true">
-      <i class="c-card__icon c-icon c-icon--error__flipped">${warningIcon}</i>
+      <i class="c-card__icon c-icon c-icon--error__flipped c-icon--inline"
+        >${warningIcon}</i
+      >
+      ${headerSlot}
```
lmuntaner, line 36: I assume that `html` function renders `undefined` well and it doesn't show anything weird in the html?

lmuntaner (approved):
LGTM

src/frontend/src/components/warnBox.ts
frederikrothenberger: Yes, then it is just skipped.

PR #1916, opened 2023-09-20
Add temp key warning to registration success screen

frederikrothenberger:
This adds the temp key warning also to the registration success page.

### DIFF
--- a/src/frontend/src/flows/register/finish.ts
+++ b/src/frontend/src/flows/register/finish.ts
@@ -21,12 +21,14 @@ export const displayUserNumberTemplate = ({
   userNumber,
   identityBackground,
   stepper,
+  slot,
   scrollToTop = false,
 }: {
   onContinue: () => void;
   userNumber: bigint;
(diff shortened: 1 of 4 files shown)


src/frontend/src/flows/register/finish.ts
```diff
@@ -21,12 +21,14 @@ export const displayUserNumberTemplate = ({
...
   identityBackground,
   stepper,
+  slot,
   scrollToTop = false,
 }: {
   onContinue: () => void;
   userNumber: bigint;
   identityBackground: IdentityBackground;
   stepper: TemplateResult;
+  slot?: TemplateResult;
```
lmuntaner, line 31: Can you name it something else?

`slot` seems very generic. The name doesn't need to be about what's inside, maybe about the function that it does inside this template.

I see it's inside a marketing bloc, how about `marketingIntroSlot`?

lmuntaner (approved):
Done! One minor comment. Already approving.

PR #1918, opened 2023-09-21
Migrate Integration Tests to PocketIC

fxgst:
Migrating the integration tests from the StateMachine to PocketIC.

* Download PocketIC binary instead of StateMachine
* Run integration tests on PocketIC instead of StateMachine


Cargo.lock
nmattia, line 1: It looks like this is pulling a lot of new dependencies. Can you explain why PocketIC tests are so much heavier than the state machine ones? Can you also briefly explain why it's better to use PocketIC?

.github/workflows/canister-tests.yml
```diff
@@ -382,9 +382,21 @@ jobs:
...
+      #     POCKET_IC_LOG_DIR: /tmp/test-archive/pic
+      
+      # - name: PocketIC logs
+      #   if: ${{ always() }}
+      #   uses: actions/upload-artifact@v3
+      #   with:
+      #     name: pocket-ic-logs
+      #     path: /tmp/test-archive/pic/*_pocket_ic_*
+
+
```
nmattia, line 399: Is this relevant?

Cargo.lock
fxgst: Sure. PocketIC does not communicate via stdin/out, but communicates over HTTP/JSON. I suppose the vast majority of new dependencies are coming from that change. We made this change to create a more language agnostic interface, such that we can add more testing libraries in different languages (e.g., [PocketIC Python](https://github.com/dfinity/pocketic-py)); for Rust users, at the moment, the feature set is identical with the StateMachine. We are exploring adding new features to PocketIC in the future.

.github/workflows/canister-tests.yml
fxgst: We added this such that we can gather PocketIC logs when we need them; If this bothers you, the comments can be removed.

nmattia: It bothers me, please remove it :) Or if you really think it's useful leave a note about why this code is commented out.

Cargo.lock
nmattia: Thank! I don't think II needs the extra agnostic interface so we could stick with the state machine test infra, unless of course it's being deprecated in favor of PocketIC

.github/workflows/canister-tests.yml
fxgst: Gotcha. I removed the comments

nmattia:
This looks a bit stale so closing for now; feel free to reopen when necessary!

PR #1922, opened 2023-09-22
Do not clear localstorage on logout

frederikrothenberger:
Deleting identity numbers on logout is unexpected and destructive. A change was requested to no longer do that.

Instead, we should give users a different way of managing the numbers shown on the landing page (to be solved as a separate PR).


frederikrothenberger:
Superseded by #1929.

PR #1925, opened 2023-09-25
Introduce release check for verify script

frederikrothenberger:
This makes CI check the `verify-hash` script so that we get notified if we break it.
It is not added to the canister tests workflow because it would increase runtime.

CI test run: https://github.com/dfinity/internet-identity/actions/runs/6296464305


przydatek (approved):
LGTM, thanks!

PR #1929, opened 2023-09-25
Do not clear localstorage on logout

frederikrothenberger:
Deleting identity numbers on logout is unexpected and destructive. A change was requested to no longer do that.

Instead, we should give users a different way of managing the numbers shown on the landing page (to be solved as a separate PR).

Note: 3 lines were changed in the production code. The rest is adapting the e2e tests to use the pick identity flow instead of entering it new.

### DIFF
--- a/src/frontend/src/test-e2e/flows.ts
+++ b/src/frontend/src/test-e2e/flows.ts
@@ -88,11 +88,9 @@ export const FLOWS = {
     deviceName: string,
     browser: WebdriverIO.Browser
   ): Promise<void> => {
-    const welcomeView = new WelcomeView(browser);
-    await welcomeView.waitForDisplay();
-    await welcomeView.login();
-    await welcomeView.typeUserNumber(userNumber);
-    await browser.$("button[data-action='continue']").click();
+    const authenticateView = new AuthenticateView(browser);
+    await authenticateView.waitForDisplay();
+    await authenticateView.pickAnchor(userNumber);
     // This flow assumes no recovery phrase, so we explicitly skip the recovery nag here
     await FLOWS.skipRecoveryNag(browser);
     const mainView = new MainView(browser);
@@ -101,21 +99,16 @@ export const FLOWS = {
   loginPin: async (
     userNumber: string,
     pin: string,
-    deviceName: string,
     browser: WebdriverIO.Browser
   ): Promise<void> => {
-    const welcomeView = new WelcomeView(browser);
-    await welcomeView.waitForDisplay();
-    await welcomeView.login();
-    await welcomeView.typeUserNumber(userNumber);
-    await browser.$("button[data-action='continue']").click();
+    const authenticateView = new AuthenticateView(browser);
+    await authenticateView.waitForDisplay();
+    await authenticateView.pickAnchor(userNumber);
     const pinAuthView = new PinAuthView(browser);
     await pinAuthView.waitForDisplay();
     await pinAuthView.enterPin(pin);
     // This flow assumes no recovery phrase, so we explicitly skip the recovery nag here
     await FLOWS.skipRecoveryNag(browser);
-    const mainView = new MainView(browser);
-    await mainView.waitForTempKeyDisplay(deviceName);
   },
   addRecoveryMechanismSeedPhrase: async (
     browser: WebdriverIO.Browser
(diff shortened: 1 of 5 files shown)


src/frontend/src/test-e2e/flows.ts
```diff
@@ -88,11 +88,9 @@ export const FLOWS = {
...
     browser: WebdriverIO.Browser
   ): Promise<void> => {
-    const welcomeView = new WelcomeView(browser);
-    await welcomeView.waitForDisplay();
-    await welcomeView.login();
-    await welcomeView.typeUserNumber(userNumber);
-    await browser.$("button[data-action='continue']").click();
+    const authenticateView = new AuthenticateView(browser);
+    await authenticateView.waitForDisplay();
+    await authenticateView.pickAnchor(userNumber);
```
lmuntaner, line 93: Does this e2e depend on logging in at some point before? Or at least on creating the anchor in the same machine without clearing the local storage?

src/frontend/src/test-e2e/flows.ts
```diff
@@ -101,21 +99,16 @@ export const FLOWS = {
...
     pin: string,
-    deviceName: string,
     browser: WebdriverIO.Browser
   ): Promise<void> => {
-    const welcomeView = new WelcomeView(browser);
-    await welcomeView.waitForDisplay();
-    await welcomeView.login();
-    await welcomeView.typeUserNumber(userNumber);
-    await browser.$("button[data-action='continue']").click();
+    const authenticateView = new AuthenticateView(browser);
```
lmuntaner, line 104: Ditto

src/frontend/src/test-e2e/views.ts
```diff
@@ -32,16 +32,6 @@ export class WelcomeView extends View {
     await this.browser.$("#loginButton").click();
     await this.browser.$("#addNewDeviceButton").click();
   }
-
-  async recover(): Promise<void> {
```
lmuntaner, line 36: Why this move? Does it have something to do with clearing the local storage on logging out?

src/frontend/src/test-e2e/flows.ts
```diff
@@ -172,8 +165,8 @@ export const FLOWS = {
     browser: WebdriverIO.Browser,
     recoveryPhrase: string
   ): Promise<void> => {
-    const welcomeView = new WelcomeView(browser);
-    await welcomeView.recover();
+    const authenticateView = new AuthenticateView(browser);
```
lmuntaner, line 168: Why this change?

The user arrives at a different page if the storage is not cleared?

lmuntaner (commented):
Why don't you clear the local storate between e2e tests?

I would differentiate e2e tests that require local storage filled and those that don't.

frederikrothenberger:
> Why don't you clear the local storate between e2e tests?
> 
> I would differentiate e2e tests that require local storage filled and those that don't.

The local storage _is_ cleared between each test. The reason why this has such a big impact on the e2e tests is because we have many variations of the initial landing screen:
* sign into known dapp (unknown identity): https://dfinity.github.io/internet-identity/authorizeNewKnown/
* sign into II management (unknown identity): https://dfinity.github.io/internet-identity/manageNew/
* sign in known identity: https://dfinity.github.io/internet-identity/authorizePick/

There are more differences, but you get the idea. And now, having logout no longer clear the local storage, makes the e2e tests that use logout in the middle (i.e. create identity, logout, do some other action with sign-in / recovery) use a different screen.

src/frontend/src/test-e2e/flows.ts
frederikrothenberger: Yes, see explanation [here](https://github.com/dfinity/internet-identity/pull/1929#issuecomment-1734928421).

src/frontend/src/test-e2e/views.ts
frederikrothenberger: Yes, see explanation [here](https://github.com/dfinity/internet-identity/pull/1929#issuecomment-1734928421).

src/frontend/src/test-e2e/flows.ts
frederikrothenberger: Yes, all tests that use the `login` flow star from a different screen now.

src/frontend/src/test-e2e/flows.ts
frederikrothenberger: See above.

lmuntaner:
In the Flows, you have different names and methods depending on where the flow starts: `registerNewIdentityWelcomeView`, `registerNewIdentityAuthenticateView`, `registerPinWelcomeView`, ...

Wouldn't it make sense to rename also the login parts as well to be clear where the flow starts?

Does it make sense to add a login e2e test that checkts you can still login from the welcomeView after logging out?

I have the feeling that the e2e now stop testing some functionality and test a different one. The flows that you are changings, could still be tested, right?

lmuntaner (approved):
Discussed on slack. PR with more e2e will come later.

frederikrothenberger:
> Discussed on slack. PR with more e2e will come later.

See https://github.com/dfinity/internet-identity/pull/1933.

PR #1933, opened 2023-09-26
Add e2e tests for login without prefilled user number

frederikrothenberger:
Due to the change in #1929 the login flow with unknown user number was no longer tested. This  reintroduces these tests for both webAuthn and PIN login.


lmuntaner (approved):
Thanks!

PR #1937, opened 2023-09-29
Add API spec for VC-issuer

przydatek:
Add specification of API that should be implemented by a VC-issuer.
This is an initial high-level cut, to be extended in the future, to include details on
 - Credential specification
 - DID methods for IC principals
 - ...

### DIFF
--- a/docs/vc-spec.md
+++ b/docs/vc-spec.md
@@ -1,6 +1,163 @@
-# II Verifiable Credential Spec
+# II Verifiable Credential Spec (MVP)
 
-## Window Post Message Interface
+
+## Issuer API
+
+This section describes the (Candid) interface to be implemented by an issuer of verifiable credentials on the IC.
+This interface is used by the II-canister during attribute sharing flow (cf. [flow description](https://github.com/dfinity/wg-identity-authentication/blob/main/topics/attribute-sharing.md))
+An example implementation of the interface is given in [demos/vc_issuer](../demos/vc_issuer).
+
+The Candid interface is as follows, and the subsequent sections describe the
+services and the corresponding messages in more detail.
+
+```candid
+type CredentialSpec = record { info : text };
+type GetCredentialRequest = record {
+    signed_id_alias : SignedIdAlias;
+    vc_jwt : opt text;
+    credential_spec : CredentialSpec;
+};
+type GetCredentialResponse = variant {
+    Ok : IssuedCredentialData;
+    Err : IssueCredentialError
+};
+type Icrc21ConsentInfo = record { consent_message : text; language : text };
+type Icrc21ConsentMessageRequest = record {
+    arg : vec nat8;
+    method : text;
+    preferences : Icrc21ConsentPreferences;
+};
+type Icrc21ConsentMessageResponse = variant {
+    Ok : Icrc21ConsentInfo;
+    Err : Icrc21Error;
+};
+type Icrc21ConsentPreferences = record { language : text };
+type Icrc21Error = variant {
+    GenericError : Icrc21ErrorInfo;
+    MalformedCall : Icrc21ErrorInfo;
+    NotSupported : Icrc21ErrorInfo;
+    Forbidden : Icrc21ErrorInfo;
+};
+type Icrc21ErrorInfo = record { description : text; error_code : nat64 };
+type IssueCredentialError = variant {
+    Internal : text;
+    SignatureNotFound;
+    InvalidIdAlias;
+    UnauthorizedSubject;
+    UnknownSubject;
+};
+type IssuedCredentialData = record { vc_jws : text };
+type PrepareCredentialRequest = record {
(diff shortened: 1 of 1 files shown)


docs/vc-spec.md
```diff
@@ -1,6 +1,163 @@
...
+};
+type Icrc21ConsentPreferences = record { language : text };
+type Icrc21Error = variant {
+    GenericError : Icrc21ErrorInfo;
+    MalformedCall : Icrc21ErrorInfo;
+    NotSupported : Icrc21ErrorInfo;
+    Forbidden : Icrc21ErrorInfo;
+};
+type Icrc21ErrorInfo = record { description : text; error_code : nat64 };
+type IssueCredentialError = variant {
```
frederikrothenberger, line 42: Do we have a way for the II front-end to handle the UX if the user does not fulfill the prerequisites for the credential.

I assume this is what the `UnauthorizedSubject` error is for? If so, it would be nice if the issuer could attach a human readable message to be shown to the user (with some explanation).

Example: "You are not eligible for this credential because your 8 year neuron has insufficient stake".

Although, in that case we would need to consider user-preferences (similar to the consent message). 🙃

docs/vc-spec.md
```diff
@@ -1,6 +1,163 @@
...
+
+type PreparedCredentialData = record { vc_jwt : opt text };
+```
+
+Specifically, the issuer checks via `prepared_id_alias.credential_jws` that user identified via `id_dapp` on the issuer
+side can be identified using `id_alias` for the purpose of attribute sharing, and that the credential
+described by `credential_spec` does apply to the user.  When these checks are successful, the issuer
+prepares and returns the credential in JWT format.
+
+**NOTE**: `prepare_credential`-service is needed only for some issuers, e.g. those that use [canister signatures](https://internetcomputer.org/docs/current/references/ic-interface-spec/#canister-signatures)
```
frederikrothenberger, line 123: This is backwards, no? `get_credential` is only needed if you do canister signatures. If you do e.g. threshold ECDSA then you can return the signed credential already in `prepare_credential` because you don't need the data certificate. However, it is _not_ possible to get a threshold ECDSA signature in `get_credential` because it is a query call.

In general, to keep the API consistent, I would make all the issuers implement both methods and just have them split the work in different ways.

docs/vc-spec.md
```diff
@@ -1,6 +1,163 @@
...
+The issuer performs the same checks as during the `prepare_credential`-call,
+plus verify that the JWT-credential is consistent with the other parameters.
+
+Upon successful checks, issuer returns the signed credential in JWS-format.
+
+
+
 
-## Window Post Message Interface
+##  Relying Party API (Window Post Message Interface)
```
frederikrothenberger, line 160: IDP API? (to be consistent with the issuer API being implemented _by_ the issuer)

docs/vc-spec.md
```diff
@@ -1,6 +1,163 @@
...
+An example implementation of the interface is given in [demos/vc_issuer](../demos/vc_issuer).
+
+The Candid interface is as follows, and the subsequent sections describe the
+services and the corresponding messages in more detail.
+
+```candid
+type CredentialSpec = record { info : text };
+type GetCredentialRequest = record {
+    signed_id_alias : SignedIdAlias;
+    vc_jwt : opt text;
```
frederikrothenberger, line 17: I'm not sure that we should mention the VC being a `JWT` in the API. If we leave it off (just call it `vc` or `prepared_vc`) then we could reuse the same API later for VCs with JSON-LD proofs.

frederikrothenberger (commented):
Left a few comments 🙂

docs/vc-spec.md
przydatek: I believe we had the same intentions in mind, but the description was somewhat misleading.  Changed now, PTAL.

docs/vc-spec.md
przydatek: As discussed, I added a text message to all the variants.

przydatek (commented):
Thanks for the review and discussion @frederikrothenberger , PTAL.

docs/vc-spec.md
```diff
@@ -1,6 +1,163 @@
...
+An example implementation of the interface is given in [demos/vc_issuer](../demos/vc_issuer).
+
+The Candid interface is as follows, and the subsequent sections describe the
+services and the corresponding messages in more detail.
+
+```candid
+type CredentialSpec = record { info : text };
+type GetCredentialRequest = record {
+    signed_id_alias : SignedIdAlias;
+    prepared_context : opt blob;
```
frederikrothenberger, line 17: Is this field `blob` or `text`? Further down, it is defined as `text`.

frederikrothenberger (approved):

docs/vc-spec.md
przydatek: `blob`, fixed now.

PR #1940, opened 2023-10-02
Clean up PIN identity validation

nmattia:
This cleans up a few things related to verifying whether some PIN identity material is still valid (i.e. if the corresponding authenticator/passkey still exists):

* The validation is untangled from the IDB lookup
* A loader is shown during the API call retrieving the existing authenticators/passkeys
* A comment is added about why we don't actually remove the IDB material, even if no corresponding authenticator seems to exist.


frederikrothenberger (approved):
LGTM, thanks!

PR #1947, opened 2023-10-04
Fix captcha expiry not being checked in register

frederikrothenberger:
Fixes a long standing bug related to captcha expiry. This bug is fixed now, because the API v2 call `identity_register` would be affected as well.


src/internet_identity/src/anchor_management/registration.rs
```diff
@@ -75,6 +68,14 @@ pub async fn create_challenge() -> Challenge {
     })
 }
 
+/// Remove challenges older than CAPTCHA_CHALLENGE_LIFETIME from the inflight challenges map
+fn prune_expired_challenges(inflight_challenges: &mut HashMap<ChallengeKey, ChallengeInfo>) {
+    // 5 mins
+    const CAPTCHA_CHALLENGE_LIFETIME: u64 = secs_to_nanos(300);
```
przydatek, line 74: nit: maybe `CAPTCHA_CHALLENGE_LIFETIME_NS`?

przydatek (approved):
LGTM, thanks!

PR #1948, opened 2023-10-04
Add API v2 identity_register method

frederikrothenberger:
This PR adds the API v2 `captcha_create` and `identity_register` methods to II. It also removes the calls to the legacy API from the API v2 integration tests.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -357,6 +357,21 @@ type IdentityInfo = record {
     metadata: MetadataMap;
 };
 
+type CaptchaCreateResponse = variant {
+    ok: Challenge;
+};
+
+type IdentityRegisterResponse = variant {
+    // Registration successful.
+    ok: IdentityNumber;
+    // No more registrations are possible in this instance of the II service canister.
+    canister_full;
+    // The challenge was not successful.
+    bad_challenge;
+    // The metadata of the provided authentication method contains invalid entries.
+    invalid_metadata: text;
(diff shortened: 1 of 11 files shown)


src/internet_identity/internet_identity.did
```diff
@@ -357,6 +357,21 @@ type IdentityInfo = record {
...
+    ok: Challenge;
+};
+
+type IdentityRegisterResponse = variant {
+    // Registration successful.
+    ok: IdentityNumber;
+    // No more registrations are possible in this instance of the II service canister.
+    canister_full;
+    // The challenge was not successful.
+    bad_challenge;
```
przydatek, line 370: nit: this is slightly ambiguous, maybe using "captcha" in the name/comment would be better, e.g. 
```
// The provided CAPTCHA solution was incorrect
bad_captcha_solution;
```
I'm aware that "challenge" is used throughout the code, and e.g.  we have `ChallengeResult`-type rather than something like `CaptchaSolution` but maybe that's something we could "fix" in the future, and use more descriptive types/labels in the new API?  Or maybe it is not worth the effort?

przydatek (approved):
LGTM, thanks!
(and sorry for the long delay!)

src/internet_identity/internet_identity.did
frederikrothenberger: Ah, good point. 👍  I'll change it in the API v2 error so that we are consistently using `captcha` in that version.

PR #1956, opened 2023-10-09
Add an example failure to RP-flow spec.

przydatek:
Add an example failure to the specification of interaction between relying party and identity provider.

### DIFF
--- a/docs/vc-spec.md
+++ b/docs/vc-spec.md
@@ -183,7 +183,7 @@ After receiving the notification that II is ready, the relying party can request
     * `credentialId`: The ID of the credential that the relying party wants to request from the issuer.
   * `credentialSubject`: The subject of the credential as known to the relying party. Internet Identity will use this principal to ensure that the flow is completed using the matching identity.
 
-#### Example
+_Example_
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


docs/vc-spec.md
```diff
@@ -183,7 +183,7 @@ After receiving the notification that II is ready, the relying party can request
     * `credentialId`: The ID of the credential that the relying party wants to request from the issuer.
   * `credentialSubject`: The subject of the credential as known to the relying party. Internet Identity will use this principal to ensure that the flow is completed using the matching identity.
 
-#### Example
+_Example_
```
frederikrothenberger, line 186: I'm not a fan of not making it a header. If it's a header, you can link to it, if it's just in italics, you cant.

docs/vc-spec.md
```diff
@@ -223,6 +225,27 @@ After the user has successfully completed the flow, Internet Identity will respo
...
+(This may change in the future, as some failures can be reported to the relying
+party without impacting user's privacy.)
+
+_Example_
+
+``` json
+{
+    "id": 1,
+    "jsonrpc": "2.0",
+    "error": {
```
frederikrothenberger, line 241: Not your mistake, but I realized that I copied the error definition from a broken example. We should fix that here too, and have a `code` instead of an `errorType`. See here: https://www.jsonrpc.org/specification#error_object

frederikrothenberger (approved):
Approved, but also left 2 comments.

docs/vc-spec.md
przydatek: I did that as `##### Example` renders (sometimes) in the same font as the section title.  But I'm not attached to it, so I reverted this change.

docs/vc-spec.md
przydatek: Thanks for caching it, changed to `code`

przydatek (commented):
Thanks!

PR #1961, opened 2023-10-10
Factor out signature_map, for re-use in other contexts.

przydatek:
`signature_map` is put into a new crate `canister_sig_util`, to which more utils related to canister signatures will be added in subsequent PRs.


src/internet_identity/src/lib.rs
```diff
@@ -1,5 +1,4 @@
 //! Various APIs for managing internet identities.
```
frederikrothenberger, line 1: This comment is no longer accurate, maybe remove it?

It would be nice if we could get rid of the whole `lib.rs` file, but I think the `get_random` fix below only works if placed in the `lib.rs` file for some reason...

frederikrothenberger (approved):
LGTM, thanks!

src/internet_identity/src/lib.rs
przydatek: Removed the comment. 
But indeed, it seems we still need `lib.rs`.

PR #1966, opened 2023-10-16
Allow toast message overflow

nmattia:
This adds a scrollbar to toast messages in case they are too long. Before this, content too long would push the close button out of the toast.

---

Before:

Screenshot 2023-10-16 at 15 54 13

After:

Screenshot 2023-10-16 at 15 54 21

### DIFF
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -1131,6 +1131,9 @@ by all browsers (FF is missing) */
 }
 
 .c-toast-body__message {
+  /* Ensures the content doesn't grow out of the toast or push other items */
+  overflow-x: auto;
+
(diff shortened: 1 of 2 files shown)


src/frontend/src/styles/main.css
```diff
@@ -1131,6 +1131,9 @@ by all browsers (FF is missing) */
 }
 
 .c-toast-body__message {
+  /* Ensures the content doesn't grow out of the toast or push other items */
+  overflow-x: auto;
```
peterpeterparker, line 1135: Have you double checked that it would not display an unexpected scrollbar (since Macos per default hide scrollbars and if not expected)?

peterpeterparker (approved):
LGTM, thx

PR #1969, opened 2023-10-17
Preload loader

nmattia:
This ensures the loader (astronaut) is loaded as soon as the app itself loads.

The asset is fairly heavy and can take a couple seconds to load from the IC. By loading it eagerly we ensures that when it is needed, the astronaut is ready.


peterpeterparker (approved):
LGTM, thanks

PR #1976, opened 2023-10-23
Fix temp keys clashing on equal public keys

frederikrothenberger:
This PR fixes a bug that surfaced in the context of the dev build where the dummy auth public key is static for all anchors: The temp keys were kept by device public key only, which lead to subsequent registrations evicting the temp key of the previous dummy auth device.
Additionally, expirations are only cleaned up lazily and also were only checking the device public key. This lead to the behaviour where a newly registered temp key was not considered valid, because it matched the expiration of a previous registration.

None of the issues above are observed in production, because WebAuthn generates new public keys for each registration, even if the same hardware is used.

The fix moves the anchor into the temp key identifier, making it different for different anchors. This addresses both of the issues outlined above.

### DIFF
--- a/src/internet_identity/src/state/temp_keys.rs
+++ b/src/internet_identity/src/state/temp_keys.rs
@@ -9,9 +9,10 @@ const TEMP_KEY_EXPIRATION_NS: u64 = 10 * MINUTE_NS;
 
 #[derive(Default, Debug)]
 pub struct TempKeys {
-    /// A map of "temporary keys" attached to devices (and a specific anchor). A temporary key can be used in lieu
-    /// of the device but has a short expiration time. These keys are used as a workaround for WebAuthn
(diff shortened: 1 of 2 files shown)


src/internet_identity/src/state/temp_keys.rs
```diff
@@ -130,4 +122,6 @@ pub struct TempKeyExpiration {
     pub device_key: DeviceKey,
     /// The expiration timestamp of the temp key
     pub expiration: Timestamp,
+    /// The anchor the temporary key is linked to
+    anchor: AnchorNumber,
```
nmattia, line 126: nit: I'd put the anchor with the `device_key` or rename `device_key` to reflect that it's also the key used in the map

nmattia (approved):
Nice, thanks! Definitely not a use case I originally considered.

Note: I'm surprised by some of these `.clone()`s but I'll trust clippy on this

PR #1972, opened 2023-10-20
Apply pending changes to Cargo.lock

frederikrothenberger:
The PR #1961 updated the `Cargo.toml` but not the `Cargo.lock`. This PR updates the `Cargo.lock` file accordingly.

PR #1978, opened 2023-10-23
Change rust build to use the --locked flag

frederikrothenberger:
This change makes builds fail if the `Cargo.lock` file is outdated, which prevents future PRs such as #1972.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -117,6 +117,7 @@ function build_canister() {
         --manifest-path "$SRC_DIR/Cargo.toml"
         --target "$TARGET"
         --release
+        --locked
         -j1
         )
     # XXX: for bash > 4.4, empty arrays are considered unset, so do some substitution


nmattia (commented):
Thanks that makes a lot of sense! should it be part of the build script though? we may only want it enabled on CI

frederikrothenberger:
> Thanks that makes a lot of sense! should it be part of the build script though? we may only want it enabled on CI

CI does not call `cargo build` directly. Should I make it an optional flag on the build script?

nmattia:
@frederikrothenberger could may even be able to get away with an extra script like so:

```
$ cargo generate-lockfile --locked
```

frederikrothenberger:
> @frederikrothenberger may even be able to get away with an extra step like so:
> 
> ```
> $ cargo generate-lockfile --locked
> ```

I don't think this does what we want? It fails if dependencies _could_ be updated. What we want is failure if the lock file _must_ be updated, right?

frederikrothenberger:
Superseded by #1979.

PR #1980, opened 2023-10-24
Remove CSP meta tag

nmattia:
This removes the `<meta>` tag used for CSP. We originally included the CSP in the HTML because the HTTP headers could not be certified.

HTTP headers are now certified so the `Content-Security-Policy` header _should_ be enough. Additionally, the `<meta>` tag hasn't been replaced correctly for some time leading to an irrelevant HTML tag.

### DIFF
--- a/src/internet_identity/src/http.rs
+++ b/src/internet_identity/src/http.rs
@@ -180,7 +180,7 @@ pub fn security_headers() -> Vec<HeaderField> {
 /// Full content security policy delivered via HTTP response header.
 ///
 /// This policy also includes the `frame-ancestors` directive in addition to the policies included in the HTML `meta` tag.
-/// We deliver the CSP by header _and_ meta tag because the headers are not yet certified.
+/// We deliver the CSP by header
 fn content_security_policy_header() -> String {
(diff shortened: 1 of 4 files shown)


src/internet_identity/src/http.rs
```diff
@@ -180,7 +180,7 @@ pub fn security_headers() -> Vec<HeaderField> {
 /// Full content security policy delivered via HTTP response header.
 ///
 /// This policy also includes the `frame-ancestors` directive in addition to the policies included in the HTML `meta` tag.
-/// We deliver the CSP by header _and_ meta tag because the headers are not yet certified.
+/// We deliver the CSP by header
 fn content_security_policy_header() -> String {
     let meta_policy = content_security_policy_meta();
```
frederikrothenberger, line 185: nit: we could now inline the `content_security_policy_meta` function, because we no longer need the CSP without the `frame-ancestors` directive.

frederikrothenberger (approved):
Thanks, LGTM!

frederikrothenberger (approved):
Yes, this is what I meant with inline `content_security_policy_meta`. Thanks! 👍

PR #1984, opened 2023-10-26
Add several utils for handling canister sigs.

przydatek:
Add to `canister_sig_util`-crate several functions for handling canister signatures.


src/canister_sig_util/src/lib.rs
```diff
@@ -1 +1,398 @@
...
+
+lazy_static! {
+    /// The root public key used when verifying canister signatures.
+    static ref IC_ROOT_PUBLIC_KEY: RwLock<Vec<u8>> = RwLock::new(
+        extract_ic_root_key_from_der(IC_ROOT_PK_DER).expect("Failed decoding IC root key.")
+    );
+}
+
+/// Resets the root public key to the given value, should be used for testing only.
+pub fn set_ic_root_public_key_for_testing(pk_der: Vec<u8>) {
```
frederikrothenberger, line 33: I don't like this:
* it makes the tests stateful, hence the `#[serial]` annotation
* so far we have been careful to only have state in a single place (so that we can easily find all actions that mutate state). Scattering global mutable state in a library makes this much harder to track.

I think it would be nicer to have the following:
* private fn that takes the root key as an argument (the enclosed `test` module has access to that)
* a public wrapper that calls the internal function above, but supplies the static IC root key.

frederikrothenberger (commented):
A lot of this code in the library is already implemented in other places, no? I feel like we might be better of by pulling in the `ic-standalone-sig-verifier` instead, especially since this PR is also pulling in a lot of additional dependencies...

Or is there a reason, why we can't use that?

src/canister_sig_util/src/lib.rs
przydatek: That's a good idea, although comes with a drawback that the tests do test the private function, rather than the public one. Still, I agree the public wrapper is simple enough, so done.

przydatek:
> A lot of this code in the library is already implemented in other places, no? I feel like we might be better of by pulling in the `ic-standalone-sig-verifier` instead, especially since this PR is also pulling in a lot of additional dependencies...
> 
> Or is there a reason, why we can't use that?
The reason is the size of added code.

Most of the added dependencies added in this PR are dev-dependencies.  Moreover, `ic-standalone-sig-verifier` comes itself with substantially more dependencies, as it contains verification of multiple schemes (canister sigs, RSA, ECDSA for two curves, EdDSA), and APIs/structs of many intermediate layers of IC-replica code.  

But as discussed offline, it is good to look at the actual numbers: I've built II-canister (with shrinking/removal of dead code) for three scenarios: 
- `plain`: the current master, without any code for verifying canister signatures
- `standalone_verifier`: here no new code is added to `canister_sig_util`, only an re-export of `ic_crypto_standalone_sig_verifier::verify_canister_sig`
- `canister_sig_util`: with the code added in this PR

Moreover for the latter two scenarios I've built the canister twice, once without making any call to verify canister signatures (hence the corresponding code can be removed by an optimizer), and once with a call to verify a canister signature (`+verify_call`).  Here are resulting sizes (in bytes):

```
1856405 internet_identity-plain.wasm.gz

1856480 internet_identity+canister_sig_util.wasm.gz
1880829 internet_identity+standalone_verifier.wasm.gz

1923703 internet_identity+canister_sig_util+verify_call.wasm.gz
2008872 internet_identity+standalone_verifier+verify_call.wasm.gz
```

So basically the actual verification code from this PR adds around 66kB to the (gzipped) wasm, while the code from the standalone library adds over 150kB.  Moreover, when the code is not used in the canister, the increase from `crypto_sig_util` crate is negligible, while the standalone library adds ~24kB.  So I'd say it is worthwhile to keep the code for now.  Wdyt?

src/canister_sig_util/Cargo.toml
```diff
@@ -5,8 +5,25 @@ version = "0.1.0"
...
+serde_bytes = "0.11"
+serde_cbor = "0.11"
+sha2 = "^0.10" # set bound to match ic-certified-map bound
 
 [dev-dependencies]
+assert_matches = "1.5.0"
 rand = { version ="0.8.5" }
-sha2 = "^0.10" # set bound to match ic-certified-map bound
+ic-cbor = "1.2.0"
+ic-certification-testing = { git = "https://github.com/dfinity/response-verification.git" }
```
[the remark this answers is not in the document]
przydatek: Done.

przydatek:
Thanks @frederikrothenberger, PTAL.

frederikrothenberger:
> > A lot of this code in the library is already implemented in other places, no? I feel like we might be better of by pulling in the `ic-standalone-sig-verifier` instead, especially since this PR is also pulling in a lot of additional dependencies...
> > Or is there a reason, why we can't use that?
> > The reason is the size of added code.
> 
> Most of the added dependencies added in this PR are dev-dependencies. Moreover, `ic-standalone-sig-verifier` comes itself with substantially more dependencies, as it contains verification of multiple schemes (canister sigs, RSA, ECDSA for two curves, EdDSA), and APIs/structs of many intermediate layers of IC-replica code.
> 
> But as discussed offline, it is good to look at the actual numbers: I've built II-canister (with shrinking/removal of dead code) for three scenarios:
> 
>     * `plain`: the current master, without any code for verifying canister signatures
> 
>     * `standalone_verifier`: here no new code is added to `canister_sig_util`, only an re-export of `ic_crypto_standalone_sig_verifier::verify_canister_sig`
> 
>     * `canister_sig_util`: with the code added in this PR
> 
> 
> Moreover for the latter two scenarios I've built the canister twice, once without making any call to verify canister signatures (hence the corresponding code can be removed by an optimizer), and once with a call to verify a canister signature (`+verify_call`). Here are resulting sizes (in bytes):
> 
> ```
> 1856405 internet_identity-plain.wasm.gz
> 
> 1856480 internet_identity+canister_sig_util.wasm.gz
> 1880829 internet_identity+standalone_verifier.wasm.gz
> 
> 1923703 internet_identity+canister_sig_util+verify_call.wasm.gz
> 2008872 internet_identity+standalone_verifier+verify_call.wasm.gz
> ```
> 
> So basically the actual verification code from this PR adds around 66kB to the (gzipped) wasm, while the code from the standalone library adds over 150kB. Moreover, when the code is not used in the canister, the increase from `crypto_sig_util` crate is negligible, while the standalone library adds ~24kB. So I'd say it is worthwhile to keep the code for now. Wdyt?

Thanks for the in depth analysis!

frederikrothenberger (commented):
As discussed: despite the bigger size, for low-level crypto operations we should pull in libraries from the crypto team (or similar) rather than have a copy of the low-level crypto code. This way we are sure to receive well maintained and up-to-date crypto libraries.

src/canister_sig_util/src/lib.rs
```diff
@@ -1 +1,379 @@
...
+    der.push(1 + bitstring.len() as u8);
+    der.push(0x00);
+    der.extend(bitstring);
+    der
+}
+
+/// Verifies the validity of `ic_certificate`.
+/// If `ic_certificate` contains a delegation, verifies the delegation as well, up to
+/// `IC_ROOT_PUBLIC_KEY`, which is required to contain the correct root public key.
+pub fn verify_signature_wrt_ic_root(
```
nathanosdev, line 128: Since I was tagged here about the publishing of `ic-certification-testing` I took a look at the PR and I noticed this code. We also have a standalone library for the verification of certificate signatures [here](https://github.com/dfinity/response-verification/blob/main/packages/ic-certificate-verification/src/certificate_verification.rs#L41). This one is already published on [crates.io](https://crates.io/crates/ic-certificate-verification).

nathanosdev: P.S. we will very likely refactor `agent-rs` to use our library too to reduce duplication there, pending a conversation with the SDK.

przydatek: Thanks @nathanosdev, I had no idea that `ic_certificate_verification` exists.  Did you talk with crypto team, who recently put together `ic-crypto-standalone-sig-verifier` (yet only in ic-repo), to sync to avoid duplication of work?

(btw, I glanced over the source, and see that you're using `miracl_core_bls12381`.  IIRC, crypto team has stopped using this crate due to security concerns, so maybe you want to double-check.  To verify BLS signatures you can use `ic-verify-bls-signature`)

nathanosdev: I wasn't aware of `ic-crypto-standalone-sig-verifier`, thank you! That looks like it can be a direct replacement for `ic-certificate-verification`.

The security concern with `miracl_core_bls12381` was related to side channel attacks on signature generation, so it was still seen as okay for signature verification. At the time that we wrote the library, file size was important since it was being used by the Service Worker and the alternative libraries were exploding the file size.

With the Service Worker gone soon, we could review those libraries again if the crypto team does the necessary refactoring to publish those crates.

src/canister_sig_util/src/lib.rs
```diff
@@ -1 +1,379 @@
...
+fn validate_delegation(
+    delegation: &Option<Delegation>,
+    signing_canister_id: Principal,
+    root_pk: &[u8],
+) -> Result<Vec<u8>, CanisterSigVerificationError> {
+    match delegation {
+        None => Ok(root_pk.to_owned()),
+        Some(delegation) => {
+            let cert: Certificate = serde_cbor::from_slice(&delegation.certificate).unwrap();
+            let _ = verify_signature_wrt_root_pk(&cert, signing_canister_id, root_pk);
```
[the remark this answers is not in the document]
przydatek: A good catch (and indeed, this was a copy-paste error, will notify the source).
The code is now removed, so resolving the conversation.

przydatek:
> As discussed: despite the bigger size, for low-level crypto operations we should pull in libraries from the crypto team (or similar) rather than have a copy of the low-level crypto code. This way we are sure to receive well maintained and up-to-date crypto libraries.

I removed now the custom code for verifying canister sigs.  If the wasm-size becomes an issue, we may ask crypto team to improve `ic-crypto-standalone-sig-verifier` (or revisit addition of custom code).

przydatek:
Thanks @frederikrothenberger for the comments, PTAL.

Dockerfile
```diff
@@ -81,6 +81,7 @@ ARG DFX_METADATA=
 
 RUN touch src/internet_identity/src/lib.rs
 RUN touch src/internet_identity_interface/src/lib.rs
+RUN touch src/canister_sig_util/src/lib.rs
```
frederikrothenberger, line 84: I think you need to rebase / merge main. This clashes with the fix I made in #1989.

src/canister_sig_util/src/lib.rs
```diff
@@ -1 +1,146 @@
...
+    let mut bitstring: Vec<u8> = vec![];
+    bitstring.push(canister_id.as_ref().len() as u8);
+    bitstring.extend(canister_id.as_ref());
+    bitstring.extend(seed);
+
+    let mut der: Vec<u8> = vec![];
+    // sequence of length 17 + the bit string length
+    der.push(0x30);
+    der.push(17 + bitstring.len() as u8);
+    der.extend(vec![
```
frederikrothenberger, line 34: You could use the `CANISTER_SIG_PK_DER_OID` here, right?

frederikrothenberger (approved):
LGTM, modulo the merge conflict and one nit. 😉 
Thanks!

Dockerfile
przydatek: Done.

przydatek (commented):
Thanks, will merge once green.

PR #1986, opened 2023-10-26
Add option to run dev server using TLS

frederikrothenberger:
This PR introduces the env variable `TLS_DEV_SERVER` that can be set to `1` to enable the dev server serving content using tls.


nmattia (approved):
Nice thanks!

suggestion: what about introducing this only when we actually use it? then you could also add a comment explaining why it's used/what it's used for

frederikrothenberger:
Superseded by #1995.

PR #1987, opened 2023-10-27
Fix asset URL paths

nmattia:
This updates the logic for serving assets from the canister.

The logic previously did not cover all paths where an asset may be found. For instance, an asset `/foo/index.html` may have returned 200 on `/foo/` but 404 on `/foo`.

Moreover the `/faq` endpoint is fixed to actually return the expected redirect to the FAQ. In practice the canister has extra logic for handling `/faq`, but this fixes the HTML-redirect fallback. This ensures that `/faq/` & `/faq/index.html` also redirect (which the canister does not currently check for).

### DIFF
--- a/src/frontend/faq.html
+++ b/src/frontend/faq.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta
+      http-equiv="Refresh"
+      content="0; url='https://identitysupport.dfinity.org/hc/en-us'"
+    />
+    <title>Internet Identity</title>
+    <link rel="shortcut icon" href="/favicon.ico" />
+  </head>
+  <body></body>
+</html>
--- a/src/internet_identity/src/assets.rs
+++ b/src/internet_identity/src/assets.rs
@@ -359,8 +359,7 @@ fn filepath_to_urlpaths(file_path: String) -> Vec<String> {
 
 #[test]
 fn test_filepath_urlpaths() {
-    fn assert_gen_paths(inp: String, exp: Vec<String>) {
-        let mut exp = exp.clone();
+    fn assert_gen_paths(inp: String, mut exp: Vec<String>) {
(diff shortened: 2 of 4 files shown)


src/internet_identity/src/assets.rs
```diff
@@ -283,33 +293,127 @@ fn file_extension<'a>(asset: &'a File) -> &'a str {
...
+    };
+
+    // Prefix everything with "/"
+    paths.into_iter().map(|path| format!("/{path}")).collect()
+}
+
+#[test]
+fn test_filepath_urlpaths() {
+    fn assert_gen_paths(inp: String, exp: Vec<String>) {
+        let mut exp = exp.clone();
```
nmattia, line 363: clone should not be necessary here

src/frontend/faq.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta
+      http-equiv="Refresh"
```
peterpeterparker, line 8: With certification v2 you can respond with a permanently redirect 301. I literally build a proof of concept today, funny. Just saying out at loud though.

peterpeterparker (commented):
LGTM to me but, I happily defer the responsability of the review to Frederik.

I added a comment about the FAQ redirect in case interesting.

src/frontend/faq.html
nmattia: yep we [do that already](https://github.com/dfinity/internet-identity/blob/bcf5bfada2e3f7370fee5dffb67ca8f3e1807d23/src/internet_identity/src/http.rs#L41-L53)! though it's slightly broken, see PR description. This is just a fallback

peterpeterparker: I meant a valid certified 301 response. I guess your current implementation is slightly broken because not certified.
Not sure, just a guess though. Anyway, was just saying in case interesting, all good.

nmattia: oh yes you're right! time to migrate this, thanks for catching

frederikrothenberger (approved):
LGTM, thanks!

PR #1989, opened 2023-10-30
Fix docker-build script

frederikrothenberger:
The docker build is layered with a dependencies layer to cache already compiled dependencies. To not cache the non-dependency code, the last modified timestamp of the `lib.rs` files has to be updated.

If a new `lib.rs` file is introduced, its timestamp needs to be updated as well. This was overlooked when introducing the `canister_sig_utils` crate.

This PR makes the dockerfile more robust to touch _all_ the relevant `lib.rs` files rather than specific ones.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,9 +79,7 @@ ARG II_INSECURE_REQUESTS=
 # DFX specific metadata for dfx deps
 ARG DFX_METADATA=
 
-RUN touch src/internet_identity/src/lib.rs
-RUN touch src/internet_identity_interface/src/lib.rs
-RUN touch src/canister_tests/src/lib.rs
+RUN touch src/*/src/lib.rs
 RUN npm ci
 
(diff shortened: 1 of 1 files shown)


przydatek (approved):
LGTM, thanks!

Dockerfile
```diff
@@ -79,9 +79,7 @@ ARG II_INSECURE_REQUESTS=
 # DFX specific metadata for dfx deps
 ARG DFX_METADATA=
 
-RUN touch src/internet_identity/src/lib.rs
-RUN touch src/internet_identity_interface/src/lib.rs
-RUN touch src/canister_tests/src/lib.rs
+RUN touch src/*/src/lib.rs
```
nmattia, line 82: I think being explicit here would be better to avoid surprises in the future. If necessary, we can introduce other checks to ensure all libs are included; at the very least would be good to be consistent between here and the `mkdir` & `touch` above

frederikrothenberger: https://dfinity.atlassian.net/browse/GIX-2025

PR #1991, opened 2023-10-30
Don't update dapps without logo

nmattia:
This updates the dapps update script to skip dapps that do not have a logo. If the dapp doesn't have a logo, then we don't have much to display.

This also adds some comments to the jq filter used to parse the dapps list for clarity.


frederikrothenberger (approved):
Thanks!

PR #1995, opened 2023-10-31
Run selenium tests without docker

frederikrothenberger:
This PR removes the docker compose based setup of running selenium test and replaces it with infrastructure based on
 native chromium and chromedriver.

The new setup uses chromium with options to override DNS resolution so that all request are hitting the vite dev server. This dev server is reconfigured to terminate TLS and forward request to the replica (or, if hot reloading is enabled, serve assets directly).

Two new config env variables are introduced for the `npm run dev` command:
* TLS_DEV_SERVER: when set to 1, the dev server will handle https instead of http.
* NO_HOT_RELOAD: when set to 1, all requests are forwarded to the replica. Assets are not hot reloaded.

There are some changes to the test:
* the copy / paste of the seed phrase has been replaced by reading the phrase from the DOM directly, because chromium now runs in headless mode (as non-headless is not supported on github action runners natively), which makes chromium behave differently with regards to the clipboard.
* The check that the raw URL is inaccessible now returns a slightly different error message, so the assert on the message was changed.

Still to do (in a separate PR):
* Documentation


vite.plugins.ts
```diff
@@ -79,3 +81,60 @@ export const minifyHTML = (): {
...
+ * Forwards requests to the local replica.
+ * Denies access to raw URLs.
+ *
+ * @param replicaOrigin Replica URL to forward requests to
+ * @param forwardRules List of rules (i.e. hostname to canisterId mappings)
+ *                     to forward requests to a specific canister
+ */
+export const replicaForwardPlugin = (
+  replicaOrigin: string,
+  forwardRules: { hosts: string[]; canisterId: string }[]
```
nmattia, line 95: _nitpick/subjectivity alert_

```suggestion
  forwardRules: Array<{ canisterId: string; hosts: string[]; }>
```

I missed the `[]` at the end, using `Array<>` makes it clearer; plus starting with `canisterId` makes it more consistent with other maps (like `Record<Foo,Bar>`) since it's basically used as a lookup

vite.plugins.ts
```diff
@@ -79,3 +81,60 @@ export const minifyHTML = (): {
...
+        res.end("Raw IC URLs are not supported");
+        return;
+      }
+
+      let matchingRule = forwardRules.find((rule) => {
+        let host = req.headers["host"];
+        return isNullish(host) || rule.hosts.includes(host);
+      });
+
+      if (matchingRule !== undefined) {
```
nmattia, line 121: for consistency you could use


```suggestion
      if (nonNullish(matchingRule)) {
```

and for even more consistency you could do something like this:

```ts
if(isNullish(matchingRule)) { 
  return next()
}
 
// happy/expected case
```

vitest.config.ts
```diff
@@ -9,6 +9,9 @@ export default defineConfig(
     },
     test: {
       environment: "jsdom",
+      env: {
+        npm_package_json: process.env.npm_package_json ?? "./package.json",
+      },
```
nmattia, line 14: Do you know why this isn't set by `npm` itself? would it make sense to make it absolute with `process.cwd() + "package.json"`?

nmattia (commented):
Amazing work, thanks!

Left some comments and will probably need to take another look to properly wrap my head around the changes

vite.plugins.ts
```diff
@@ -79,3 +81,60 @@ export const minifyHTML = (): {
     return minify(html, { collapseWhitespace: true });
   },
 });
+
+/**
+ * Forwards requests to the local replica.
+ * Denies access to raw URLs.
```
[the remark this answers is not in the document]
frederikrothenberger: It is what the tests expect. It simulates what the service worker / icx-proxy would do on mainnet.

src/frontend/src/test-e2e/views.ts
```diff
@@ -183,37 +183,22 @@ export class RecoveryMethodSelectorView extends View {
...
-    });
+    // Ideally, we could press the copy button on the page and the read the
+    // clipboard. However, this is not possible due to the content security
+    // policy.
+    // The alternative solution of simulating key presses does not work either,
+    // since chromium does not allow to interact with the clipboard via keyboard
+    // shortcuts when run in headless mode.
+    // For the lack of a better solution, we read the seed phrase from the DOM.
+
+    const seedPhrase = (await this.browser.execute(`
```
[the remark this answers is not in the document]
frederikrothenberger: The front-end checks fail on that:
```
 Error: src/frontend/src/test-e2e/views.ts(197,10): error TS2339: Property 'map' does not exist on type 'NodeListOf<Element>'.
Error: src/frontend/src/test-e2e/views.ts(197,15): error TS7006: Parameter 'e' implicitly has an 'any' type.
Error: Process completed with exit code 2.
```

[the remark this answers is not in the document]
frederikrothenberger: If suppressed it now with` @ts-ignore`. Other suggestions welcome. 😉

vitest.config.ts
frederikrothenberger: Since we don't download chrome anymore, we don't need that variable either. Removed.

vite.config.ts
```diff
@@ -59,6 +73,32 @@ export default defineConfig(({ mode }: UserConfig): UserConfig => {
     plugins: [
       [...(mode === "development" ? [injectCanisterIdPlugin()] : [])],
       [...(mode === "production" ? [minifyHTML(), compression()] : [])],
+      [...(process.env.TLS_DEV_SERVER === "1" ? [basicSsl()] : [])],
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, we could. But then the vite dev server for local development would only run on HTTPS. Do we want that?

src/frontend/src/test-e2e/views.ts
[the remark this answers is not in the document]
nmattia: Yeah a `NodeList` is not exactly an array so failing here does probably make sense. `Array.from(document.querySelectorAll(...)).map` should work though

> Although NodeList is not an Array, it is possible to iterate over it with forEach(). It can also be converted to a real Array using [Array.from()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from).

https://developer.mozilla.org/en-US/docs/Web/API/NodeList

vite.config.ts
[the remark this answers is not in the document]
nmattia: Oh I see. By enabling `basicSsl()` , non-ssl calls are rejected? If so then I guess no, we don't want that

vite.config.ts
```diff
@@ -25,6 +30,15 @@ export default defineConfig(({ mode }: UserConfig): UserConfig => {
     II_VERSION: `${process.env.II_VERSION ?? ""}`,
   };
 
+  const testAppCanisterId = existsSync(
```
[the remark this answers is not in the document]
frederikrothenberger: I don't think we want that tough... Otherwise the vite dev server can only be started when the test app is also deployed. And I would assume that for most development activities it is not needed.

We could introduce some sort of e2e configuration for starting the dev server though. Should I look into that?

vite.config.ts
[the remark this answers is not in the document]
frederikrothenberger: Yes, unfortunately that is the case.

src/frontend/src/test-e2e/util.ts
```diff
@@ -62,17 +68,9 @@ export async function runInBrowser(
...
     },
-    automationProtocol: "webdriver",
-    // explicitly set host and port, because otherwise webdriverio will start their own chromedriver
-    hostname: "127.0.0.1",
-    port: 4444,
-    path: "/wd/hub",
-    logLevel: "info",
-    // outputDir pipes all webdriver log output into ./wdio.log
-    // stdout only contains errors on test failures
-    outputDir: "./",
```
frederikrothenberger, line 75: I opted to remove the file output logging for 2 reasons:
* Most of the logs _still_ go to stdout, even with the `outputDir` set. The log files only contain a few lines (which are also okay to appear in stdout):
  ```
  Starting ChromeDriver 119.0.6045.105 (38c72552c5e15ba9b3117c0967a0fd105072d7c6-refs/branch-heads/6045@{#1103}) on port 55890
  Remote connections are allowed by an allowlist (0.0.0.0).
  Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
  ChromeDriver was started successfully.
  ```
* wdio now creates a logfile per test, because chrome driver is started new for each test. This results in an awful lot of tiny files that are hard to match to the actual tests, because the filename only contains an opaque session id.

I think having all the e2e test infrastructure output in stdout makes analyzing test failures easier. Note that the `test-failures` folder is still created as previously, that contains the HTML file, screenshot and browser console logs of the test at the point of failure.

This change can be split into it's own PR if you'd like.

nmattia: Makes sense!

> This change can be split into it's own PR if you'd like.

yes please!

src/frontend/src/test-e2e/views.ts
```diff
@@ -183,37 +183,22 @@ export class RecoveryMethodSelectorView extends View {
...
+    // clipboard. However, this is not possible due to the content security
+    // policy.
+    // The alternative solution of simulating key presses does not work either,
+    // since chromium does not allow to interact with the clipboard via keyboard
+    // shortcuts when run in headless mode (which is the only mode accepted by CI).
+    // For the lack of a better solution, we read the seed phrase from the DOM.
+
+    const seedPhrase = (await this.browser.execute(() =>
+      Array.from(document.querySelectorAll(".c-list--recovery-word"))
+        // @ts-ignore
```
nmattia, line 196: I think you can replace with `Array.from(document.querySelectorAll(...)).map`

vite.plugins.ts
```diff
@@ -79,3 +81,63 @@ export const minifyHTML = (): {
...
+          `Denying access to raw URL ${req.method} https://${req.headers.host}${req.url}`
+        );
+        res.statusCode = 400;
+        res.end("Raw IC URLs are not supported");
+        return;
+      }
+
+      const matchingRule = forwardRules.find((rule) => {
+        let host = req.headers["host"];
+        return isNullish(host) || rule.hosts.includes(host);
```
nmattia, line 122: if I understand this correctly, when the `host` header is not set, you return the first rule in the list. Is this correct?

How about moving the host reading out of the callback? Also would it make sense to return `next()` if the `host` is nullish?

frederikrothenberger: Ah yes, this is a bug. Fixed, thanks!

src/frontend/src/test-e2e/views.ts
frederikrothenberger: This is already done?

nmattia: I thought so to. Maybe the tab was out of sync...

src/frontend/src/test-e2e/util.ts
frederikrothenberger: Done in #2015.

vite.plugins.ts
```diff
@@ -79,3 +81,70 @@ export const minifyHTML = (): {
...
+      ) {
+        console.log(
+          `Denying access to raw URL ${req.method} https://${req.headers.host}${req.url}`
+        );
+        res.statusCode = 400;
+        res.end("Raw IC URLs are not supported");
+        return;
+      }
+
+      let host = req.headers["host"];
```
nmattia, line 120:
```suggestion
      const host = req.headers["host"];
```

vite.plugins.ts
```diff
@@ -79,3 +81,69 @@ export const minifyHTML = (): {
...
+
+      const host = req.headers["host"];
+      if (isNullish(host)) {
+        // default handling
+        return next();
+      }
+
+      const matchingRule = forwardRules.find(
+        // the TS compiler doesn't get that due to the isNullish check above, host can't be undefined
+        // (which is why the explicit cast is there)
```
nmattia, line 127:
```suggestion
```

PR #1996, opened 2023-10-31
Upgrade webdriverio to latest version

frederikrothenberger:
This upgrades webdriverio to the latest version in preparation
for the change to non-dockerized selenium tests.

The new version now is also able to download chrome and chromedriver on
its own, which is why the `download-chrome.ts` script can be removed.

The testing chrome needs to run with `/dev/shm` disabled because
otherwise it will randomly crash on the GitHub action runners.


nmattia (approved):
This is absolutely amazing! thank you very much!

PR #1997, opened 2023-10-31
Port components to Astro

nmattia:
This ports all components shown in the showcase to Astro components.

This means we now have proper routing for displaying those components.

The components are wrapped in native/web components so that Astro can import them properly; the alternative of creating an arbitrary div with `id="pageContent"` like we do elsewhere is problematic because Astro bundles all javascript scripts together when e.g. using `Astro.glob`, meaning all scripts would trigger on any page importing the components (like the index) and all would fight for the same `#pageContent`.

The approach of creating new elements is a bit more robust, though also a bit more verbose.

Another benefit of this approach is that we can e.g. embed/showcase all components in the index page.


src/showcase/src/pages/components/ChoosePin.astro
```diff
@@ -0,0 +1,63 @@
...
+<script>
+  import { LitComponent } from "./lit";
+
+  import { pinInput } from "$src/components/pinInput";
+  import { Chan } from "$src/utils/utils";
+  import { html } from "lit-html";
+  import { asyncReplace } from "lit-html/directives/async-replace.js";
+
+  class ChoosePin extends LitComponent {
+    render() {
```
nmattia, line 18: this function's body was copied

src/showcase/src/pages/components/PickAnchor.astro
```diff
@@ -0,0 +1,62 @@
...
+  import { asyncReplace } from "lit-html/directives/async-replace.js";
+  import { Ref, createRef, ref } from "lit-html/directives/ref.js";
+  import { Chan, NonEmptyArray, asNonEmptyArray } from "$src/utils/utils";
+  import { LitComponent } from "./lit";
+  import { withRef } from "$src/utils/lit-html";
+  import { TemplateResult, html } from "lit-html";
+  import { mkAnchorPicker } from "$src/components/anchorPicker";
+
+  class PickAnchor extends LitComponent {
+    render() {
```
nmattia, line 18: this function's body was copied

src/showcase/src/pages/components/ToastExample.astro
```diff
@@ -0,0 +1,48 @@
...
+  <toast-example></toast-example>
+</div>
+
+<script>
+  import { LitComponent } from "./lit";
+  import { TemplateResult, html } from "lit-html";
+  import { toast } from "$src/components/toast";
+
+  class ToastExample extends LitComponent {
+    render(): TemplateResult {
```
nmattia, line 15: this function's body was copied

src/showcase/src/pages/components/PickAnchor.astro
```diff
@@ -0,0 +1,63 @@
+---
+import "$src/styles/main.css";
```
peterpeterparker, line 2: You should import the css in a component like a header that way your do not have to add an import in each component

src/showcase/src/pages/components/PickAnchor.astro
```diff
@@ -0,0 +1,63 @@
...
+---
+
+<div>
+  <pick-anchor></pick-anchor>
+</div>
+
+<script>
+  import { asyncReplace } from "lit-html/directives/async-replace.js";
```
peterpeterparker, line 10: the explicit `.js` in the import is mandatory?

src/showcase/src/pages/components/ChoosePin.astro
```diff
@@ -0,0 +1,63 @@
...
+
+<script>
+  import { LitComponent } from "./lit";
+
+  import { pinInput } from "$src/components/pinInput";
+  import { Chan } from "$src/utils/utils";
+  import { html } from "lit-html";
+  import { asyncReplace } from "lit-html/directives/async-replace.js";
+
+  class ChoosePin extends LitComponent {
```
peterpeterparker, line 17: This LitComponent is closed per default - i.e. is does not generate a true WebComponents right? Just asking in case that would prevent the propagation of CSS and therefore lead to unexpected styling issue

src/showcase/src/pages/components/index.astro
```diff
@@ -0,0 +1,25 @@
...
+    <title>II - Components</title>
+  </head>
+  <ul>
+    {
+      allComponents.map((C) => (
+        <li>
+          <div>
+            <C.default />
+          </div>
+          <a style="display: block; text-align: center;" href={C.url}>
```
peterpeterparker, line 16: you can use a style tag to avoid repetition of the style in the page

peterpeterparker (approved):
LGTM, few minor questions

src/showcase/src/pages/components/PickAnchor.astro
nmattia: Can you clarify what you mean? Then I would have to import that new component on every page here, right?

src/showcase/src/pages/components/PickAnchor.astro
nmattia: yep: https://lit.dev/docs/templates/directives/#asyncreplace

src/showcase/src/pages/components/ChoosePin.astro
nmattia: Can you clarify what you mean by "closed per default"? I don't have much experience with webcomponents.

In this case there's no shadow DOM and we really do expect the out CSS to be applied

src/showcase/src/pages/components/index.astro
nmattia: Yeah, but for now this is good enough I think :)

src/showcase/src/pages/components/PickAnchor.astro
peterpeterparker: Yes but I assumed that there are other things that are common and that can fits either a `<Layout />` or `<Head />` component.

src/showcase/src/pages/components/ChoosePin.astro
peterpeterparker: > The {mode: "open"} argument gives the page a way to break the encapsulation of your shadow DOM. If you don't want to give the page this ability, pass {mode: "closed"} instead, and then shadowRoot returns null.

https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM

If "closed", then the component is a WebComponent and only the CSS within the component style the component, defining global CSS rules have no effect on inside the component. "open" the contrary.


// <my-component>
//    #shadow-root (closed)
//    <div style="bg: green">I'm green 
//
// <my-other-component>
//    (open)
//    <div style="bg: green">I'm red
//
// div.css 
//    bg: red

nmattia: Gotcha thanks! though as mentioned, they don't even use a Shadow DOM and the goal is for the outer CSS to be applied. Everything ok then right?

Screenshot 2023-11-02 at 15 16 27

src/showcase/src/pages/components/PickAnchor.astro
nmattia: We do this [elsewhere](https://github.com/dfinity/internet-identity/blob/f3abe415ed8208ee8c0cad9ff923cafade48a89d/src/showcase/src/layouts/Layout.astro) but here it's a bit more complicated because the "components" (like `PickAnchor.astro`) are used as both components (imported in `components/index.astro`) and full blown pages. Having e.g. `PickAnchor.astro` wrapped in a Layout would mean the index would also include the layout, possibly several times

nmattia: Ok I had a long discussion with myself and I'll actually move the components out of `pages/` and will import them with the existing `Layout`; additional to the CSS they're other things (like `<meta name="viewport" content="width=device-width, initial-scale=1.0" />`) that we'd have to repeat everywhere

src/showcase/src/pages/components/ChoosePin.astro
peterpeterparker: Looks good! So no WebComponents according the book but, custom elements. Maybe was just my misunderstanding, thanks for the feedbacks.

nmattia: Ah no my bad; custom elements is definitely the right (or at least more correct) term!

PR #1998, opened 2023-10-31
Update internet_identity.did

eatahine:
1. Error: In the `HeaderField` type definition, the field names are missing.
   Fix: Added field names `name` and `value` to the `HeaderField` type.

2. Error: In the `StreamingCallbackHttpResponse` type definition, `Token` is used, but the `Token` type is not defined.
   Fix: Added the `Token` type definition.

3. Error: In the `StreamingStrategy` type definition, the `variant` was not followed by the possible variants.
   Fix: Defined the `Callback` variant with the required fields and used `Token` within the `StreamingStrategy` type.

4. Error: In the `Purpose` and `KeyType` type definitions, labels were missing for the variants.
   Fix: Added labels `recovery` and `authentication` to the `Purpose` variant and labels to the `KeyType` variants.

5. Error: In the `ChallengeResult` type definition, `chars` was of the incorrect type, and it should be a `vec text`.
   Fix: Changed the type of `chars` to `vec text` to represent a list of characters.

6. Error: The `metadata` field in the `DeviceData` and `DeviceWithUsage` types referred to a `MetadataMap`, but the `MetadataMap` type was not defined.
   Fix: Added the `MetadataMap` type definition and used it in the `DeviceData` and `DeviceWithUsage` types.

7. Error: In the `InternetIdentityInit` type definition, the field types inside the `assigned_user_number_range` record were not specified.
   Fix: Added the field types within the `assigned_user_number_range` record.

8. Error: In the `DeployArchiveResult` type definition, the field names were missing.
   Fix: Added field names to the `success` variant in the `DeployArchiveResult` type.

The provided code fixes all of these errors to make the code valid and readable.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -8,8 +8,8 @@ type FrontendHostname = text;
 type Timestamp = nat64;
 
 type HeaderField = record {
-    text;
-    text;
+    name: text;
+    value: text;
 };
 
 type HttpRequest = record {
@@ -24,7 +24,7 @@ type HttpResponse = record {
     status_code: nat16;
     headers: vec HeaderField;
     body: blob;
-    upgrade : opt bool;
+    upgrade: opt bool;
     streaming_strategy: opt StreamingStrategy;
 };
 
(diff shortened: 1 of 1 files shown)


frederikrothenberger:
Hi @eatahine 

The things you are listing here are not actually errors.

For example, the HTTP gateway protocol types are specified here: https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec#canister-http-interface 
Introducing field names would break the protocol. 

Also, the short-hands that we use have well defined behavior: https://github.com/dfinity/candid/blob/master/spec/Candid.md#syntactic-shorthands

Additionally, I don't think that removing all comments is a sensible action as they are quite useful to readers.

Please explain what you want to achieve with these changes.

PR #2006, opened 2023-11-02
Move LABEL_SIG constant  to a crate, so that it can be re-used.

przydatek:
Move `LABEL_SIG`-constant to `canister_sig_util`, so that it can be used elsewhere.
Also, make `canister_sig_util::extract_raw_root_pk_from_der` public (it was overlooked previously).


frederikrothenberger (approved):
LGTM, thanks!

PR #2016, opened 2023-11-07
Wait for principals to not be anonymous after sign-in

frederikrothenberger:
The switch to non-dockerized selenium tests exposed some flakyness in the way principals are collected from the test app after sign-in.

This PR introduces a new method on the `DemoAppView` to properly wait for authentication to finish, including a check that the principal has been updated.

### DIFF
--- a/src/frontend/src/test-e2e/views.ts
+++ b/src/frontend/src/test-e2e/views.ts
@@ -1,5 +1,7 @@
 import { zip } from "$src/utils/utils";
+import { Principal } from "@dfinity/principal";
 import { nonNullish } from "@dfinity/utils";
+import { waitToClose } from "./util";
 
 class View {
   constructor(protected browser: WebdriverIO.Browser) {}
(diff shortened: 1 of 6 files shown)


nmattia (commented):
How would you feel about instead replacing `waitToClose` with a wait that ensures the app was authenticated? this would be checking that the principal is not the anonymous principal or something like this, but it seems counterintuitive to have this check in `getPrincipal`

frederikrothenberger:
> How would you feel about instead replacing `waitToClose` with a wait that ensures the app was authenticated? this would be checking that the principal is not the anonymous principal or something like this, but it seems counterintuitive to have this check in `getPrincipal`

I did that now, thanks. This led to the removal of quite a bit of duplication. 👍

src/frontend/src/test-e2e/views.ts
```diff
@@ -606,6 +608,19 @@ export class DemoAppView extends View {
     await this.browser.$("#signinBtn").click();
   }
 
+  /** Waits for the authentication to finish, the window to close and the principal to update.
+   * Returns the principal after the user has been authenticated.
+   */
+  async waitForAuthenticated(): Promise<string> {
+    // wait for the demo app to close the II window
+    await waitToClose(this.browser);
```
nmattia, line 616: this will break if we ever have more than one window open (could picture this happening with VCs); not saying this should be changed but good to keep in mind

nmattia (approved):
Nice, thanks!

PR #2019, opened 2023-11-08
Add utils for handling verifiable credentials

przydatek:
Add crate `vc_util` that contains various structures and helpers for handling verifiable credentials in general, and specifically for working with id_alias-credentials.
Also, add `CanisterSigPublicKey`-struct to `canister_sig_util`, to simplify handling of canister sig public keys.

### DIFF
--- a/src/vc_util/Cargo.toml
+++ b/src/vc_util/Cargo.toml
@@ -16,9 +16,6 @@ canister_sig_util = { path = "../canister_sig_util" }
 identity_core = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false }
 identity_credential = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false , features = ["validator"] }
 identity_jose = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false, features = ["iccs"]}
-# identity_core = { path = "../../../identity.rs/identity_core" }
-# identity_credential = { path = "../../../identity.rs/identity_credential", default-features = false, features = ["credential", "validator"]}
-# identity_jose = { path = "../../../identity.rs/identity_jose", default-features = false, features = ["iccs"]}
 
 # other dependencies
 serde = { version = "1", features = ["rc"] }
--- a/src/vc_util/Cargo.toml
+++ b/src/vc_util/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "vc_util"
+description = "Utils for verifiable credentials on the IC"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+# ic dependencies
+candid = "0.9"
+ic-certified-map = "0.4"
+ic-crypto-standalone-sig-verifier = { git = "https://github.com/dfinity/ic", rev = "bf9bc00b032a18d2dbcfcfcfb7a76a562f350c9e" }
+ic-types = { git = "https://github.com/dfinity/ic", rev = "bf9bc00b032a18d2dbcfcfcfb7a76a562f350c9e" }
+canister_sig_util = { path = "../canister_sig_util" }
+
+# vc dependencies
+identity_core = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false }
+identity_credential = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false , features = ["validator"] }
+identity_jose = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false, features = ["iccs"]}
+# identity_core = { path = "../../../identity.rs/identity_core" }
+# identity_credential = { path = "../../../identity.rs/identity_credential", default-features = false, features = ["credential", "validator"]}
+# identity_jose = { path = "../../../identity.rs/identity_jose", default-features = false, features = ["iccs"]}
+
+# other dependencies
+serde = { version = "1", features = ["rc"] }
+serde_bytes = "0.11"
+serde_cbor = "0.11"
+serde_json = "1"
+sha2 = "^0.10" # set bound to match ic-certified-map bound
+
+[dev-dependencies]
+assert_matches = "1.5.0"
--- a/src/vc_util/src/lib.rs
+++ b/src/vc_util/src/lib.rs
@@ -49,9 +49,9 @@ pub fn vc_signing_input(
 }
 
 /// Computes and returns SHA-256 hash of the given `signing_input` prefixed with
-///      length(VC_SIGNING_INPUT_DOMAIN) || VC_SIGNING_INPUT_DOMAIN
+///      `length(VC_SIGNING_INPUT_DOMAIN) · VC_SIGNING_INPUT_DOMAIN`
 /// (for domain separation), where `length(a)` is the length of byte-array `a`,
-/// and || denotes concatenation of bytes.
+/// and `·` denotes concatenation of bytes.
 pub fn vc_signing_input_hash(signing_input: &[u8]) -> Hash {
     let mut hasher = Sha256::new();
     let buf = [VC_SIGNING_INPUT_DOMAIN.len() as u8];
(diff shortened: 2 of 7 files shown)


src/vc_util/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+ic-types = { git = "https://github.com/dfinity/ic", rev = "bf9bc00b032a18d2dbcfcfcfb7a76a562f350c9e" }
+canister_sig_util = { path = "../canister_sig_util" }
+
+# vc dependencies
+identity_core = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false }
+identity_credential = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false , features = ["validator"] }
+identity_jose = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false, features = ["iccs"]}
+# identity_core = { path = "../../../identity.rs/identity_core" }
+# identity_credential = { path = "../../../identity.rs/identity_credential", default-features = false, features = ["credential", "validator"]}
+# identity_jose = { path = "../../../identity.rs/identity_jose", default-features = false, features = ["iccs"]}
```
nmattia, line 21: is this still relevant?

src/vc_util/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+candid = "0.9"
+ic-certified-map = "0.4"
+ic-crypto-standalone-sig-verifier = { git = "https://github.com/dfinity/ic", rev = "bf9bc00b032a18d2dbcfcfcfb7a76a562f350c9e" }
+ic-types = { git = "https://github.com/dfinity/ic", rev = "bf9bc00b032a18d2dbcfcfcfb7a76a562f350c9e" }
+canister_sig_util = { path = "../canister_sig_util" }
+
+# vc dependencies
+identity_core = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false }
+identity_credential = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false , features = ["validator"] }
+identity_jose = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false, features = ["iccs"]}
```
nmattia, line 18: Are there official repos we can link to? also I think that particular dep creates a circular dependency to an older commit of II

src/vc_util/src/lib.rs
```diff
@@ -0,0 +1,421 @@
...
+pub fn vc_signing_input(
+    credential_jwt: &str,
+    canister_sig_pk: &CanisterSigPublicKey,
+) -> Result<Vec<u8>, String> {
+    let encoder = jws_encoder(credential_jwt, canister_sig_pk)?;
+    Ok(encoder.signing_input().to_vec())
+}
+
+/// Computes and returns SHA-256 hash of the given `signing_input` prefixed with
+///      length(VC_SIGNING_INPUT_DOMAIN) || VC_SIGNING_INPUT_DOMAIN
```
nmattia, line 52: out of curiosity, what does `||` mean here?

src/vc_util/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+name = "vc_util"
+description = "Utils for verifiable credentials on the IC"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+# ic dependencies
+candid = "0.9"
+ic-certified-map = "0.4"
+ic-crypto-standalone-sig-verifier = { git = "https://github.com/dfinity/ic", rev = "bf9bc00b032a18d2dbcfcfcfb7a76a562f350c9e" }
```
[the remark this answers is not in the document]
przydatek: Yep, that's unfortunate...
I've left a comment as suggested.

src/vc_util/Cargo.toml
przydatek: No, we don't have an official repo for that yet.
There is no circular dependency, as I've removed recently the dependency of `identity.rs`-crates on II-repo.

src/vc_util/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+ic-types = { git = "https://github.com/dfinity/ic", rev = "bf9bc00b032a18d2dbcfcfcfb7a76a562f350c9e" }
+canister_sig_util = { path = "../canister_sig_util" }
+
+# vc dependencies
+identity_core = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false }
+identity_credential = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false , features = ["validator"] }
+identity_jose = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false, features = ["iccs"]}
+# identity_core = { path = "../../../identity.rs/identity_core" }
```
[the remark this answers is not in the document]
przydatek: right, done.

src/vc_util/Cargo.toml
przydatek: it was useful for local development of `identity.rs`-extensions, but not needed in the main repo, so removed.

src/vc_util/src/lib.rs
przydatek: concatenation, I've added a note

src/vc_util/src/lib.rs
```diff
@@ -0,0 +1,421 @@
...
+    hasher.update(signing_input);
+    hasher.finalize().into()
+}
+
+/// Constructs and returns a JWS (a signed JWT) from the given components.
+pub fn vc_jwt_to_jws(
+    credential_jwt: &str,
+    canister_sig_pk: &CanisterSigPublicKey,
+    sig: &[u8],
+) -> Result<String, String> {
```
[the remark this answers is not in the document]
przydatek: pushed to later -- these errors should never happen (as they would indicate some internal corruption), and I had panics on them before, but switched to `Result` to avoid panics.

src/vc_util/src/lib.rs
```diff
@@ -0,0 +1,421 @@
...
+
+/// Validates that the given claims are consistent with id_alias-credential
+/// for the given alias tuple.
+fn validate_id_alias_claims(
+    claims: JwtClaims<Value>,
+    alias_tuple: &AliasTuple,
+) -> Result<(), JwtValidationError> {
+    validate_claim("sub", did_for_principal(alias_tuple.id_dapp), claims.sub())?;
+    validate_claim("iss", II_ISSUER_URL, claims.iss())?;
+    let jti = claims.jti().ok_or(inconsistent_jwt_claims(
```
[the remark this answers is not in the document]
przydatek: I've added `jti` to our "Id alias" VCs, as we might want to use them in the future.  But you're right, we are not really using them currently, so I'll remove them (and I removed the check).

przydatek (commented):
Thanks for the comments, PTAL.

src/vc_util/src/lib.rs
nmattia: We should then probably use "·" for concatenation instead to be consistent with the [other spec](https://github.com/dfinity/internet-identity/blob/main/docs/ii-spec.md#identity-design-and-data-model)

src/vc_util/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+) -> Result<(), CredentialVerificationError> {
+    let claims = verify_credential_jws(credential_jws, root_pk_raw)
+        .map_err(CredentialVerificationError::InvalidJws)?;
+    validate_id_alias_claims(claims, alias_tuple)
+        .map_err(CredentialVerificationError::InvalidClaims)
+}
+
+/// Verifies the specified JWS credential cryptographically.
+/// DOES NOT perform semantic validation of the claims in the credential.
+pub fn verify_credential_jws(
```
frederikrothenberger, line 96: One important thing about canister signature verification is that you validate it with respect to a canister id. What about adding to this method another parameter (the issuer canister id) and also validate that the credential indeed was signed by the expected issuer?

przydatek: That's a good idea, done.
(I thought we'll have another helper for checking that, but I agree that this function is a good place for the check)

przydatek (commented):
Thanks for the comments and suggestions, PTAL.

frederikrothenberger (approved):
LGTM, thanks!

PR #2021, opened 2023-11-09
Run tests with clean IndexedDB

nmattia:
This updates our test setup to use a fresh IndexedDB before each test. This ensures that data from one test doesn't pollute other tests.

The `createStore` calls a turned into function to ensure `idb-keyval` doesn't try to access the global `indexeddb` object upon module import, but only once the stores are actually needed.

### DIFF
--- a/src/frontend/src/flows/pin/idb.ts
+++ b/src/frontend/src/flows/pin/idb.ts
@@ -11,10 +11,8 @@ export {
 /* IndexedDB-specific storage for browser identities
  * (indexed by user number) */
 
-export const idbIdentitiesStore = createStore(
-  "browser-identities",
-  "identities"
-);
+export const idbIdentitiesStore = () =>
+  createStore("browser-identities", "identities");
 
 export const idbStorePinIdentityMaterial = async ({
   userNumber,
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/pin/idb.ts
```diff
@@ -11,10 +11,8 @@ export {
 /* IndexedDB-specific storage for browser identities
  * (indexed by user number) */
 
-export const idbIdentitiesStore = createStore(
-  "browser-identities",
-  "identities"
-);
+export const idbIdentitiesStore = () =>
```
peterpeterparker, line 14: that has an impact at runtime, if there would be a way to avoid it, I would not be against

also did you check it still works fine in Safari? No other reason of asking than my lack of trust in that browser given my experience using idb in safari

peterpeterparker (approved):
LGTM, thx

src/frontend/src/flows/pin/idb.ts
nmattia: >  if there would be a way to avoid it, I would not be against

I couldn't find one. The issue is that this file is imported before `test-setup` is run, meaning the `createStore` call is made before `global.indexeddb` is set (using `fake-indexeddb`). We could e.g. create a singleton or something, but I wonder if this is really worth the very tiny performance penalty.

Suggestions welcome!

peterpeterparker: Gotcha.

PR #2022, opened 2023-11-09
Don't deploy Release Candidate

nmattia:
Internet Identity used to have a weekly release schedule. This is not the case anymore, so this removes release candidates, automated RC deploys and RC notifications.


frederikrothenberger (commented):
What about just removing the schedule? I think manually dispatching release candidate deployment is useful.

nmattia:
I think this workflow will just rot then. We could still run the workflow e.g. daily to ensure it works, and just remove the slack message. WDYT?

frederikrothenberger:
> I think this workflow will just rot then. We could still run the workflow e.g. daily to ensure it works, and just remove the slack message. WDYT?

Yeah, that's good too. 👍

nmattia:
@frederikrothenberger actually, we still deploy to fgte on every commit on the `main` branch, I think we can just remove this workflow here

frederikrothenberger (approved):

nmattia:
Ok not 100% sure what to do here; I'll revisit this in a bit

PR #2023, opened 2023-11-09
Split storage ops from serialization

nmattia:
This rearranges the storage (was: `userNumber`) module to improve maintainability.

In particular, the localStorage-specific operations are extracted into dedicated localStorage functions, and the actual
pruning/migrating/updating is also split out. This gives three layers: localStorage, pruning/migrating/updating, and top-level functions (like `getAnchors`) used by the rest of the codebase.

The storage (was: `userNumber`) test suite is also updated to clarify that the test data is related to localStorage.


src/frontend/src/storage/index.ts
nmattia, line 1: This module was renamed & modified, but git cannot keep track

peterpeterparker (approved):
LGTM, thx

nmattia:
Added an extra test just in case

PR #2028, opened 2023-11-10
Typecheck vite config

nmattia:
This ensures the vite (vitest, plugins, etc) config is typechecked, and installs the http-proxy package type definitions.


frederikrothenberger (approved):
Thanks!

PR #2031, opened 2023-11-10
Refactor storage tests

nmattia:
This improves the storage tests in several ways:

* The local storage helper does _not_ wrap the whole test anymore; this caused issues because e.g. individual tests would not be singled out or skipped with test.only() or test.skip().
* The `after` parameter of the local storage helper can now also be a validating function on the storage content, making it possible to assert specific properties (for instead assert that an anchor is set without need to know the exact timestamp it was last used).
* Some comment about local storage are made a bit more general.


peterpeterparker (approved):
LGTM, thx

PR #2034, opened 2023-11-13
Refactor canister sig verification to only check canister id

frederikrothenberger:
The seed is only relevant to separate different public keys of the same canister. The vc credential check does not rely on a specific seed. In practice, clients will not care about the seed and will accept any.

### DIFF
--- a/src/vc_util/src/lib.rs
+++ b/src/vc_util/src/lib.rs
@@ -83,20 +83,21 @@ pub fn did_for_principal(principal: Principal) -> String {
 pub fn verify_id_alias_credential_jws(
     credential_jws: &str,
     alias_tuple: &AliasTuple,
-    canister_sig_pk: &CanisterSigPublicKey,
+    canister_id: &Principal,
     root_pk_raw: &[u8],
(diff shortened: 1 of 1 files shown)


src/vc_util/src/lib.rs
```diff
@@ -83,20 +83,21 @@ pub fn did_for_principal(principal: Principal) -> String {
...
 
-/// Verifies the specified JWS credential cryptographically.
+/// Verifies the specified JWS credential cryptographically and checks that the signature was
+/// created by the provided canister.
 /// DOES NOT perform semantic validation of the claims in the credential.
-pub fn verify_credential_jws(
+pub fn verify_credential_jws_with_canister_id(
     credential_jws: &str,
-    canister_sig_pk: &CanisterSigPublicKey,
+    canister_id: &Principal,
```
przydatek, line 100: nit: maybe rename to `signing_canister_id`, so that it is clear which canister id is meant?

przydatek (approved):
LGTM, thanks!

PR #2035, opened 2023-11-13
Use cdk println! macro in vc_utils

frederikrothenberger:
This PR changes the `println!` macro usages
to `ic_cdk::println!` macros so that they also
work when the library is used inside a canister.

Includes a drive-by typo fix.


przydatek (approved):
LGTM, thanks!

PR #2036, opened 2023-11-13
Add specs and APIs for attribute sharing

przydatek:
Add specs and APIs for attribute sharing (verifiable credentials) feature: 
- specs for issuers and for identity providers (II)
- API types for issuers
- API types for II

### DIFF
--- a/docs/vc-spec.md
+++ b/docs/vc-spec.md
@@ -116,13 +116,13 @@ type PreparedCredentialData = record { prepared_context : opt blob };
 Specifically, the issuer checks via `prepared_id_alias.credential_jws` that user identified via `id_dapp` on the issuer
 side can be identified using `id_alias` for the purpose of attribute sharing, and that the credential
 described by `credential_spec` does apply to the user.  When these checks are successful, the issuer
-prepares and returns a context in `PreparedCredentialData.prepared_context` (if any).  The returned prepared context is then 
+prepares and returns a context in `PreparedCredentialData.prepared_context` (if any).  The returned prepared context is then
 passed back to the issuer in a subsequent `get_credential`-call (see below).
 This call must be authenticated (i.e. it is not available for the anonymous sender).
 
-**NOTE:** 
+**NOTE:**
 The value of `prepared_context` is basically used to transfer information between `prepare_credential`
-and `get_credential` steps, and it is totally up to the issuer to decide on the content of 
+and `get_credential` steps, and it is totally up to the issuer to decide on the content of
 that field.  That is, the issuer creates `prepared_context`, and is the only entity that
 consumes it.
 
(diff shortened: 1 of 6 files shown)


frederikrothenberger (approved):
LGTM, thanks!

docs/vc-spec.md
```diff
@@ -0,0 +1,313 @@
...
+
+### 2: Request a VC
+
+After receiving the notification that II is ready, the relying party can request a VC by sending the following JSON-RPC request:
+ 
+* Method: `request_credential`
+* Params:
+  * `issuer`: An issuer that the relying party trusts. It has the following properties:
+    * `origin`: The origin of the issuer.
+    * `canisterId`: (optional) The canister id of the issuer, if applicable/known.
```
nmattia, line 188: what are the semantics or `cansiterId`? what if both `origin` and `canisterId` are provided?

docs/vc-spec.md
```diff
@@ -0,0 +1,313 @@
...
+  }
+}
+```
+
+### 3: Get a Response
+
+
+#### Receive a Verifiable Presentation
+
+After the user has successfully completed the flow, Internet Identity will respond with the following JSON-RPC response:
```
nmattia, line 264: Excerpt of offline discussion:

> Q: does the order of credentials (alias & actual credentials) matter? assuming it doesn’t since the spec doesn’t mention it

> A: The spec doesn't mandate the ordering explicitly, but the bullet points for the returned presentation list the id alias credential first, so let's stick to this ordering (I'll add to the spec an explicit note for this ordering, thanks for pointing this out!)

nmattia (changes requested):
Couple typos and suggestions

przydatek (commented):
Thanks @nmattia, PTAL.

docs/vc-spec.md
nmattia: ok, might be worth adding something along the lines of 

> (optional) The canister id of the issuer, if applicable/known. **if specified and not the same as the one reported by the BN for `origin`, this is an error**


otherwise you could understand this as "either use the canister ID of `origin` or fallback to `canisterId` if `origin` doesn't have a canister ID"

nmattia (approved):
Left one more comment on the origin/canisterId, otherwise LGTM

docs/vc-spec.md
```diff
@@ -0,0 +1,313 @@
...
+
+type PreparedCredentialData = record { prepared_context : opt blob };
+```
+
+Specifically, the issuer checks via `prepared_id_alias.credential_jws` that user identified via `id_dapp` on the issuer
+side can be identified using `id_alias` for the purpose of attribute sharing, and that the credential
+described by `credential_spec` does apply to the user.  When these checks are successful, the issuer
+prepares and returns a context in `PreparedCredentialData.prepared_context` (if any).  The returned prepared context is then 
+passed back to the issuer in a subsequent `get_credential`-call (see below).
+This call must be authenticated (i.e. it is not available for the anonymous sender).
```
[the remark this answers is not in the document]
przydatek: Explained in the description.

docs/vc-spec.md
```diff
@@ -0,0 +1,313 @@
...
+
+**NOTE:** 
+The value of `prepared_context` is basically used to transfer information between `prepare_credential`
```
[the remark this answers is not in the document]
przydatek: Done.

docs/vc-spec.md
przydatek: Added.

przydatek (commented):
Thanks!
(and sorry for missing two comments earlier)

PR #2038, opened 2023-11-14
Fix download links in readme

frederikrothenberger:
The readme download links were still pointing to the non-gzipped assets.
This PR also fixes a typo.


nmattia (approved):
Thanks!

PR #2040, opened 2023-11-14
Migrate localStorage data to IndexedDB

nmattia:
This ports our localStorage data (identity numbers used & timestamps) to IndexedDB. By doing this we follow the team's convention of storing only UI related, short-lived data in localStorage, and all meaningful data in IndexedDB. This also simplifies the storage handling by skipping the JSON serialization.

For the time being, the data is still duplicated to localStorage in case a rollback is needed. That way, data written after the upgrade but before the rollback will still be available to users after the rollback.

### DIFF
--- a/src/frontend/src/storage/index.test.ts
+++ b/src/frontend/src/storage/index.test.ts
@@ -225,7 +225,7 @@ function withStorage(
       setLocalStorage(lsBefore);
     }
 
-    await idb.delMany(await idb.keys());
+    await idb.clear();
     const idbBefore = opts?.indexeddb?.before;
     if (nonNullish(idbBefore)) {
       await setIndexedDB(idbBefore);
@@ -249,7 +249,7 @@ function withStorage(
 
     // Remove all entries
     // (cannot just reset global.indexeddb because idb-keyval stores a pointer to the DB)
-    await idb.delMany(await idb.keys());
+    await idb.clear();
 
     // Check the localStorage "after"
 
--- a/src/frontend/src/storage/index.test.ts
+++ b/src/frontend/src/storage/index.test.ts
@@ -1,6 +1,11 @@
 import { nonNullish } from "@dfinity/utils";
 import { IDBFactory } from "fake-indexeddb";
-import * as idb from "idb-keyval";
+import {
+  clear as idbClear,
+  get as idbGet,
+  keys as idbKeys,
+  set as idbSet,
+} from "idb-keyval";
 import { MAX_SAVED_ANCHORS, getAnchors, setAnchorUsed } from ".";
 
 beforeAll(() => {
@@ -225,7 +230,7 @@ function withStorage(
       setLocalStorage(lsBefore);
     }
 
-    await idb.clear();
+    await idbClear();
     const idbBefore = opts?.indexeddb?.before;
     if (nonNullish(idbBefore)) {
       await setIndexedDB(idbBefore);
@@ -249,7 +254,7 @@ function withStorage(
 
     // Remove all entries
     // (cannot just reset global.indexeddb because idb-keyval stores a pointer to the DB)
-    await idb.clear();
+    await idbClear();
 
     // Check the localStorage "after"
 
@@ -296,18 +301,18 @@ type IndexedDB = Record<string, unknown>;
 
 const setIndexedDB = async (db: IndexedDB) => {
   for (const key in db) {
-    await idb.set(key, db[key]);
+    await idbSet(key, db[key]);
   }
 };
 
 const readIndexedDB = async (): Promise<IndexedDB> => {
   const db: IndexedDB = {};
 
-  for (const k of await idb.keys()) {
+  for (const k of await idbKeys()) {
     if (typeof k !== "string") {
       throw new Error("Bad type");
     }
-    db[k] = await idb.get(k);
+    db[k] = await idbGet(k);
   }
   return db;
 };
--- a/src/frontend/src/storage/index.test.ts
+++ b/src/frontend/src/storage/index.test.ts
@@ -1,15 +1,22 @@
 import { nonNullish } from "@dfinity/utils";
+import { IDBFactory } from "fake-indexeddb";
+import * as idb from "idb-keyval";
 import { MAX_SAVED_ANCHORS, getAnchors, setAnchorUsed } from ".";
 
-test("anchors default to nothing", () => {
-  expect(getAnchors()).toStrictEqual([]);
+beforeAll(() => {
+  // Initialize the IndexedDB global
+  global.indexedDB = new IDBFactory();
+});
+
+test("anchors default to nothing", async () => {
+  expect(await getAnchors()).toStrictEqual([]);
 });
 
 test(
   "old userNumber is recovered",
   withStorage(
-    () => {
-      expect(getAnchors()).toStrictEqual([BigInt(123456)]);
+    async () => {
+      expect(await getAnchors()).toStrictEqual([BigInt(123456)]);
     },
     { localStorage: { before: { userNumber: "123456" } } }
   )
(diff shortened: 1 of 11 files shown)


src/frontend/src/storage/index.test.ts
```diff
@@ -144,33 +205,64 @@ test(
...
+      setLocalStorage(lsBefore);
     }
-    fn();
-    const after = opts?.localStorage?.after;
-    if (nonNullish(after)) {
-      const actual: LocalStorage = readLocalStorage();
 
-      if (typeof after === "function") {
-        after(actual);
+    await idb.delMany(await idb.keys());
```
peterpeterparker, line 228: - `clear();` would be more handy?
- you don't use a custom store (general question)?

src/frontend/src/storage/index.test.ts
```diff
@@ -196,3 +288,25 @@ function readLocalStorage(): LocalStorage {
 
   return ls;
 }
+
+/** Indexed DB */
+
+type IndexedDB = Record<string, unknown>;
+
+const setIndexedDB = async (db: IndexedDB) => {
```
peterpeterparker, line 297: why not `setMany` and `getMany` for set|getIndexedDB, wouldn't it be more accurate given that it's atomic?

src/frontend/src/storage/index.ts
```diff
@@ -2,6 +2,7 @@
 
 import { parseUserNumber } from "$src/utils/userNumber";
 import { nonNullish } from "@dfinity/utils";
+import * as idb from "idb-keyval";
```
peterpeterparker, line 5: can you be specific in the import to avoid bundling what's not used?

src/frontend/src/storage/index.ts
```diff
@@ -135,6 +138,27 @@ const mostUnused = (anchors: Anchors): string | undefined => {
 /** Best effort to migrate potentially old localStorage data to
  * the latest format */
 const migrated = (): Anchors | undefined => {
+  // Try to read the "v1" (localStorage.anchors) storage and return that if found
+  const v1 = migratedV1();
```
peterpeterparker, line 142: out of curiosity, why calling it v1 and v0 instead of using the names anchors and userNumber?

src/frontend/src/storage/index.ts
```diff
@@ -156,6 +180,32 @@ const migrated = (): Anchors | undefined => {
   return anchors;
 };
 
+/** "Low-level" functions to read anchors from and write anchors to IndexedDB */
+
+const readIndexedDB = async (): Promise<Anchors | undefined> => {
+  const item: unknown = await idb.get("anchors");
+
+  if (item === undefined) {
```
peterpeterparker, line 188: isNullish or can never be null or null is a value?

src/frontend/src/storage/index.ts
```diff
@@ -156,6 +180,32 @@ const migrated = (): Anchors | undefined => {
...
+      parsed.error;
+    console.warn(message);
+    return {};
+  }
+
+  return parsed.data;
+};
+
+const writeIndexedDB = async (anchors: Anchors) => {
+  await idb.set("anchors", anchors);
```
peterpeterparker, line 206: nth: you can spare the "await" here I think

peterpeterparker (commented):
Is `withAnchors` called each time a user access II?

Just not clear to me where the migration of the potential existing keys are migrated from local storage to idb

src/frontend/src/storage/index.ts
nmattia: I've avoided `isNullish` because if the value is `null` then (AFAICT) it was actually set to `null` in the DB (as opposed to simply not being set). I think it makes sense to check for `undefined` explicitly but don't feel too strongly about it.

src/frontend/src/storage/index.ts
nmattia: Yeah, just wanted to be consistent with the other functions. Don't feel strongly about it.

src/frontend/src/storage/index.ts
```diff
@@ -91,6 +92,8 @@ const updateStorage = (
   doWrite ||= pruned;
 
   if (doWrite) {
+    await writeIndexedDB(prunedAnchors);
```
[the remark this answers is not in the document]
nmattia: The localStorage calls are synchronous anyway so nothing to be resolved there

src/frontend/src/storage/index.ts
nmattia: Because I'm expecting more migrations in the future and don't want to give them all names :)

src/frontend/src/storage/index.ts
nmattia: Ah good point!

src/frontend/src/storage/index.test.ts
nmattia: > would be more handy?

Just missed it, thanks!

> you don't use a custom store (general question)?

no, just using the default `keyval-store` DB and `keyval` store

src/frontend/src/storage/index.test.ts
nmattia: This felt like the easier to write & read but I don't feel strongly about it

nmattia:
Yes, the migration is attempted/run every time -- whenever the anchors are read (`getAnchors()`) or updated (`setAnchorUsed()`)

src/frontend/src/storage/index.ts
nmattia: Just to be clear, the `await` would be replaced with a `return` otherwise the promise will be lost

src/frontend/src/storage/index.ts
nmattia: Actually I'll do this separately, everywhere in the codebase.

src/frontend/src/storage/index.test.ts
peterpeterparker: No strong opinion neither given it's a test

src/frontend/src/storage/index.ts
peterpeterparker: So you'll do in this PR right?

src/frontend/src/storage/index.ts
peterpeterparker: Not knowing the code base I wasn't unsure if it was expected or not therefore the question but sure makes sense.

src/frontend/src/storage/index.ts
peterpeterparker:
```
const writeIndexedDB = (anchors: Anchors): Promise<void> => set("anchors", anchors);
```

But doesn't matter that much

src/frontend/src/storage/index.ts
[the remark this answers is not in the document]
peterpeterparker: If `writeIndexedDB` throws an exception, `writeLocalStorage` won't be called (if I get it right). So, ss that an issue?

For example, both must succeed to be consider the process as ok:

```
await Promise.all([writeIndexedDB(), async () => writeLocalStorage()]);
```

Both must be executed regardless if one fails:

```
await Promise.allSettled([writeIndexedDB(), async () => writeLocalStorage()]);
```

It does not matter if idb fails:

```
try {
   await writeIndexedDB();
} finally {
   writeLocalStorage();
}
```

etc.

Not sure it's more clear though, maybe it is no concern or I don't get it though.

src/frontend/src/storage/index.ts
nmattia: Done!

src/frontend/src/storage/index.ts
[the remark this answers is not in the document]
nmattia: Mh that's a good question. I think it's fine _not_ to write to localStorage if indexeddb fails; better even maybe. Otherwise the storages will get out of sync. And if local storage fails I'm not sure what we'd do; we wouldn't undo the idb write anyway.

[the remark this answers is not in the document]
peterpeterparker: > will get out of sync.

Makes sense

peterpeterparker (approved):
LGTM, thx

PR #2041, opened 2023-11-14
Merge main onto vc-main-clean

przydatek:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,12 +20,5 @@
     "require-await": ["error"],
     "@typescript-eslint/no-floating-promises": ["error"]
   },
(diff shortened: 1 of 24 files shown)


przydatek:
@nmattia : there was a merge conflict in `tsconfig.all.json`, and I used main-version to resolve it, but apparently that's not right, as "Frontend checks and lints" fail -- should I keep the original version of `tsconfig.all.json`? or some mix of both versions?

nmattia:
@przydatek thanks! I had to fix this in my branch as well, you can find the diff here: https://github.com/dfinity/internet-identity/commit/7d13ffae102c65e3d104df44d809d5785b788b79

PR #2043, opened 2023-11-14
Check that caller matches id alias dapp principal in issuer API

frederikrothenberger:
This PR adds the authorization check that `prepare_credential` and `get_credential` can only be called by the id alias dapp principal.


przydatek (approved):
LGTM, thanks!

PR #2044, opened 2023-11-15
Add support for issuing id_alias credentials

przydatek:
Add to II canister two new services, `prepare_id_alias` and `get_id_alias`, to enable issuance of id_alias credentials, needed for attribute sharing flow.


src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+
+pub async fn prepare_id_alias(
+    identity_number: IdentityNumber,
+    dapps: InvolvedDapps,
+) -> PreparedIdAlias {
+    state::ensure_salt_set().await;
+    check_frontend_length(&dapps.relying_party);
+    check_frontend_length(&dapps.issuer);
+
+    let id_alias_principal = get_id_alias_principal(identity_number, &dapps);
```
frederikrothenberger, line 44: This feels a little backwards. I would have expected:
1. create `let pk = canister_sig_pk_der(...)`
2. `let id_alias_principal = Principal::self_authenticating(pk)`

Right now you are creating the `seed` twice, no?

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+    identity_number: IdentityNumber,
+    dapps: InvolvedDapps,
+    rp_id_alias_jwt: &str,
+    issuer_id_alias_jwt: &str,
+) -> GetIdAliasResponse {
+    check_frontend_length(&dapps.relying_party);
+    check_frontend_length(&dapps.issuer);
+
+    state::assets_and_signatures(|cert_assets, sigs| {
+        let id_alias_principal = get_id_alias_principal(identity_number, &dapps);
```
frederikrothenberger, line 95: Dito with regards to calculating the seed twice

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+        let seed = calculate_id_alias_seed(identity_number, &dapps);
+        let id_rp = delegation::get_principal(identity_number, dapps.relying_party.clone());
+        let id_issuer = delegation::get_principal(identity_number, dapps.issuer.clone());
+        let canister_id = ic_cdk::id();
+        let canister_sig_pk = CanisterSigPublicKey::new(canister_id, seed.to_vec());
+
+        let signing_input = vc_signing_input(rp_id_alias_jwt, &canister_sig_pk)
+            .expect("failed getting signing_input");
+        let msg_hash = vc_signing_input_hash(&signing_input);
+        let maybe_sig = get_signature(cert_assets, sigs, seed, msg_hash);
```
frederikrothenberger, line 105: This can be simplified to:
```
let Some(rp_sig) = get_signature(cert_assets, sigs, seed, msg_hash) else {
    return GetIdAliasResponse::NoSuchCredentials("rp_sig not found".to_string());
};
```

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+            return GetIdAliasResponse::NoSuchCredentials("rp_sig not found".to_string());
+        };
+        let rp_jws = vc_jwt_to_jws(rp_id_alias_jwt, &canister_sig_pk, &rp_sig)
+            .expect("failed constructing JWS");
+
+        let signing_input = vc_signing_input(issuer_id_alias_jwt, &canister_sig_pk)
+            .expect("failed getting signing_input");
+        let msg_hash = vc_signing_input_hash(&signing_input);
+        let maybe_sig = get_signature(cert_assets, sigs, seed, msg_hash);
+        let issuer_sig = if let Some(sig) = maybe_sig {
```
frederikrothenberger, line 118: Dito simplification

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+        issuer_id_alias_jwt,
+    }
+}
+
+fn id_alias_jwt_and_msg_hash(
+    alias_tuple: &AliasTuple,
+    canister_sig_pk: &CanisterSigPublicKey,
+) -> (String, Hash) {
+    let credential_jwt = prepare_id_alias_jwt(alias_tuple);
+    let signing_input =
```
frederikrothenberger, line 79: Is there ever a case where you don't do
`vc_signing_input_hash(vc_signing_input(jwt, pk))`?

Might be worth having a convenience function in the library for that?

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+            issuer_id_alias_credential: SignedIdAlias {
+                id_alias: id_alias_principal,
+                id_dapp: id_issuer,
+                credential_jws: issuer_jws,
+            },
+        })
+    })
+}
+
+fn get_signature(
```
frederikrothenberger, line 141: This is a lot of duplication from the same method in `delegation.rs`.
What about only having it once and make the one from `delegation.rs` also take a `msg_hash` instead of the `(pk, expiration)` inputs to the delegation?

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+        tree,
+    };
+
+    let mut cbor = serde_cbor::ser::Serializer::new(Vec::new());
+    cbor.self_describe().unwrap();
+    sig.serialize(&mut cbor).unwrap();
+    Some(cbor.into_inner())
+}
+
+fn add_signature(sigs: &mut SignatureMap, msg_hash: Hash, seed: Hash) {
```
frederikrothenberger, line 184: dito

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+fn id_alias_credential(alias_tuple: &AliasTuple) -> Credential {
+    let subject: Subject = Subject::from_json_value(json!({
+        "id": did_for_principal(alias_tuple.id_dapp),
+        "has_id_alias": did_for_principal(alias_tuple.id_alias),
+    }))
+    .expect("internal: failed building id_alias subject");
+
+    let credential: Credential = CredentialBuilder::default()
+        .id(prepare_credential_id())
+        .issuer(Url::parse(II_ISSUER_URL).expect("internal: bad issuer url"))
```
frederikrothenberger, line 228: Is this a meaningful URL? How did you chose it? Shouldn't it point to II instead?

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,247 @@
...
+        .type_("InternetIdentityIdAlias")
+        .subject(subject)
+        .build()
+        .expect("internal: failed building id_alias credential");
+    credential
+}
+
+fn prepare_credential_id() -> Url {
+    let url = Url::parse(II_CREDENTIAL_URL_PREFIX).expect("internal: bad credential id base url");
```
frederikrothenberger, line 237: Dito

src/internet_identity/tests/integration/vc_mvp.rs
```diff
@@ -0,0 +1,805 @@
...
+    let prepare_response = api::vc_mvp::prepare_id_alias(
+        &env,
+        canister_id,
+        principal_1(),
+        prepare_id_alias_req.clone(),
+    )
+    .expect("Result of prepare_id_alias is not Ok")
+    .expect("Got 'None' from prepare_id_alias");
+
+    let prepared_id_alias = if let PrepareIdAliasResponse::Ok(prepared_id_alias) = prepare_response
```
frederikrothenberger, line 746: You could use the `match_value!` macro here to simplify things.

src/internet_identity/tests/integration/vc_mvp.rs
```diff
@@ -0,0 +1,805 @@
...
+        identity_number,
+        relying_party,
+        issuer,
+        rp_id_alias_jwt: prepared_id_alias.rp_id_alias_jwt,
+        issuer_id_alias_jwt: prepared_id_alias.issuer_id_alias_jwt,
+    };
+    let response = api::vc_mvp::get_id_alias(&env, canister_id, principal_1(), get_id_alias_req)?
+        .expect("Got 'None' from get_id_alias");
+
+    if let GetIdAliasResponse::NoSuchCredentials(_err) = response {
```
frederikrothenberger, line 716: Use `assert!(matches!(...))`?

frederikrothenberger (commented):
Thanks! Left a few comments.

src/internet_identity/tests/integration/vc_mvp.rs
przydatek: Done.

src/internet_identity/tests/integration/vc_mvp.rs
przydatek: Hmm, but `match_value!` does not give me the matched value back, i.e. I'd still need to match manually to get `prepared_id_alias`, no?

przydatek (commented):
Thanks @frederikrothenberger, PTAL.

src/internet_identity/tests/integration/vc_mvp.rs
frederikrothenberger: `match_value!` supports bindings (see [here](https://github.com/dfinity/internet-identity/blob/49f0d0b721815c244bb0dc6c0e4c9514824ba3b2/src/internet_identity/tests/integration/v2_api/authn_method_add.rs#L38) for an example).

src/internet_identity/src/vc_mvp.rs
frederikrothenberger: Yes, but the expiration could easily be a parameter. But I don't see why we would need different signature expirations? The expiration is only to give the client time to query the certificate after the update call. Whether you do that for delegations or id aliases does not really matter.

TLDR: this expiration is a property of the canister signature process not of the signed payload.

src/internet_identity/src/vc_mvp.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+fn id_alias_credential(alias_tuple: &AliasTuple) -> Credential {
+    let subject: Subject = Subject::from_json_value(json!({
+        "id": did_for_principal(alias_tuple.id_dapp),
+        "has_id_alias": did_for_principal(alias_tuple.id_alias),
+    }))
+    .expect("internal: failed building id_alias subject");
+
+    let credential: Credential = CredentialBuilder::default()
```
frederikrothenberger, line 211: Maybe it would be prudent to put an `exp` claim inside as well? Otherwise we will break the semantics of older tokens if we ever change how the seed is calculated.
And it is also useful to emphasize the ephemeral nature of these credentials.

frederikrothenberger (approved):
LGTM, only minor comments. Thanks!

src/internet_identity/src/vc_mvp.rs
przydatek: Ok, I added `add_signature` to the ticket.

src/internet_identity/tests/integration/vc_mvp.rs
przydatek: TIL ^^, thanks for the example, done.

src/internet_identity/src/vc_mvp.rs
przydatek: Good idea, done.

przydatek (commented):
Thanks!

PR #2057, opened 2023-11-16
Add an implementation of a VC issuer

przydatek:
Add an example implementation of a VC-issuer, according to the [spec](https://github.com/dfinity/internet-identity/blob/main/docs/vc-spec.md#issuer-api). The issuer demonstrates the functionality of an issuer, but is not a "real  life" issuer that holds information about real users -- instead, it offers an additional "backdoor" API to add identities eligible for obtaining verifiable credentials: 
```
    add_employee : (principal) -> (text);
    add_graduate : (principal) -> (text);
```


demos/vc_issuer/Cargo.toml
nmattia, line 1: Can you add a README to `demos/vc_issuer/`? See other directories in `demos/` for inspiration. Should explain what this is, how to build it, deploy it, etc.

demos/vc_issuer/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Try to run from the script location
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPT_DIR"
```
nmattia, line 6:
```suggestion

# Make sure we always run from the issuer root
VC_ISSUER_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
cd "$VC_ISSUER_DIR/.."
```

Rewords to clarify (we're not just trying, we're effectively running from the script location) and rename (guessing `SCRIPT_DIR` was copy pasted)

demos/vc_issuer/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+cargo build --release --target wasm32-unknown-unknown --manifest-path ./Cargo.toml -j1
+ic-wasm "target/wasm32-unknown-unknown/release/vc_issuer.wasm" -o "./vc_issuer.wasm" shrink
+ic-wasm vc_issuer.wasm -o vc_issuer.wasm metadata candid:service -f vc_issuer.did -v public
+gzip --no-name --force --keep "vc_issuer.wasm"
```
nmattia, line 11: Why `--keep`?

demos/vc_issuer/dfx.json
nmattia, line 1: what's the intended use of this? I find myself often adding an entry in the top-level `dfx.json` pointing to this directory instead (for `build:`, `wasm:` and `candid:`)

demos/vc_issuer/dfx.json
```diff
@@ -0,0 +1,28 @@
...
+
+      "remote": {
+        "id": {
+          "ic": "rdmx6-jaaaa-aaaaa-aaadq-cai"
+        }
+      }
+    },
+
+    "issuer": {
+      "type": "custom",
```
nmattia, line 16: do we actually need a custom build here?

demos/vc_issuer/vc_issuer.did
nmattia, line 1: Can you add a general comment about this? IIUC this includes both the Issuer spec (meant to be generally implemented by issuers) and some specific endpoints (add_employee, configure, etc)

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+    IssueCredentialError, IssuedCredentialData, PrepareCredentialRequest,
+    PrepareCredentialResponse, PreparedCredentialData, SignedIdAlias,
+};
+use vc_util::{
+    did_for_principal, vc_jwt_to_jws, vc_signing_input, vc_signing_input_hash,
+    verify_id_alias_credential_jws, AliasTuple,
+};
+
+/// We use restricted memory in order to ensure the separation between non-managed config memory (first page)
+/// and the managed memory for the archived data & indices.
```
nmattia, line 32: Is this accurate?

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+    const BOUND: Bound = Bound::Unbounded;
+}
+
+impl Default for IssuerConfig {
+    fn default() -> Self {
+        Self {
+            ic_root_key_raw: extract_raw_root_pk_from_der(IC_ROOT_PK_DER)
+                .expect("failed to extract raw root pk from der"),
+            idp_canister_ids: vec![Principal::from_text(PROD_II_CANISTER_ID).unwrap()],
+        }
```
nmattia, line 80: Does it make sense to default to the official II? what is the use case?

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+    }
+}
+
+#[derive(CandidType, Deserialize)]
+struct IssuerInit {
+    /// Root of trust for checking canister signatures.
+    ic_root_key_der: Vec<u8>,
+    /// List of canister ids that are allowed to provide id alias credentials.
+    idp_canister_ids: Vec<Principal>,
+}
```
nmattia, line 100: Out of curiosity, why split the `Init` and `Config` types?

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+
+#[init]
+#[candid_method(init)]
+fn init(init_arg: Option<IssuerInit>) {
+    let Some(init) = init_arg else {
+        // nothing to do
+        return;
+    };
+    apply_config(init);
```
nmattia, line 109: I think something like this might be more concise & will convey the intent better:

```rust
if let Some(init) = init_arg {
    apply_config(init)
}
```

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+    let credential = match prepare_credential_payload(&req) {
+        Ok(credential) => credential,
+        Err(err) => return PrepareCredentialResponse::Err(err),
+    };
+    let seed = calculate_seed(&req.signed_id_alias.id_alias);
+    let canister_id = ic_cdk::id();
+    let canister_sig_pk = CanisterSigPublicKey::new(canister_id, seed.to_vec());
+    let credential_jwt = credential
+        .serialize_jwt()
+        .expect("internal: JWT serialization failure");
```
nmattia, line 185: out of curiosity, do you know in what cases serialization might fail?

frederikrothenberger (approved):
LGTM, thanks for the improvements!
Maybe add a newline at the end of the did file though.

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+
+#[query]
+#[candid_method(query)]
+fn get_credential(req: GetCredentialRequest) -> GetCredentialResponse {
+    if let Err(e) = authorize_vc_request(&req.signed_id_alias) {
+        return GetCredentialResponse::Err(e);
+    };
+    if let Err(err) = verify_credential_spec(&req.credential_spec) {
+        return GetCredentialResponse::Err(IssueCredentialError::UnsupportedCredentialSpec(err));
+    }
```
nmattia, line 217: Here and elsewhere where applicable something like this might be more readable:

```rust

fn get_credential_inner(req: GetCredentialRequest) -> Result<IssuedCredentialData, IssueCredentialError> {
    authorize_vc_request(&req.signed_id_alias)?;
    verify_credential_spec(&req.credential_spec)
        .map_err(|err| { IssueCredentialError::UnsupportedCredentialSpec(err)} )?;

    ....
```

and call `get_credential_inner` from `get_credential` (just an example)

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+        )));
+    };
+    let vc_jws =
+        vc_jwt_to_jws(&credential_jwt, &canister_sig_pk, &sig).expect("failed constructing JWS");
+    GetCredentialResponse::Ok(IssuedCredentialData { vc_jws })
+}
+
+#[update]
+#[candid_method]
+async fn vc_consent_message(req: Icrc21VcConsentMessageRequest) -> Icrc21ConsentMessageResponse {
```
nmattia, line 251: Would it make sense to add at least one more language, as a showcase?

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+            error_code: 0,
+        }));
+    }
+    Icrc21ConsentMessageResponse::Ok(Icrc21ConsentInfo {
+        consent_message: get_vc_consent_message(&req),
+        language: "en-US".to_string(),
+    })
+}
+
+fn verify_credential_spec(spec: &CredentialSpec) -> Result<(), String> {
```
nmattia, line 264: Thinking of how this might be helpful for testing:

How about having `add_employee` take both a principal _and_ an `employerName`? Then no employers have to be hard coded

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+        "UniversityDegreeCredential" => verify_single_argument(
+            spec,
+            "institutionName",
+            ArgumentValue::String("DFINITY College of Engineering".to_string()),
+        ),
+        other => Err(format!("Credential {} is not supported", other)),
+    }
+}
+
+fn verify_single_argument(
```
nmattia, line 280: Can you add a note about what this does?

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+    };
+
+    let mut cbor = serde_cbor::ser::Serializer::new(Vec::new());
+    cbor.self_describe().unwrap();
+    sig.serialize(&mut cbor).unwrap();
+    Some(cbor.into_inner())
+}
+
+fn calculate_seed(principal: &Principal) -> Hash {
+    let dummy_salt = [5u8; 32];
```
nmattia, line 395: In a real implementation, should this be random data? If that's the case, and since this is meant as an example to copy, worth leaving a note?

demos/vc_issuer/src/main.rs
```diff
@@ -0,0 +1,521 @@
...
+}
+
+fn hash_bytes(value: impl AsRef<[u8]>) -> Hash {
+    let mut hasher = Sha256::new();
+    hasher.update(value.as_ref());
+    hasher.finalize().into()
+}
+
+// Order dependent: do not move above any function annotated with #[candid_method]!
+candid::export_service!();
```
nmattia, line 497: does this export the candid interface somehow? is this & the .did file kept in sync?

nmattia (changes requested):
@przydatek can you clarify the purpose of `vc_issuer`? is it meant as a reference implementation/showcase for how to implement an issuer?

If so it'd be nice to explain how it works a bit more, both in the code (how/why it's verifying signatures, requests, etc) and in a README (how to build it, deploy it, how to add employees, etc), and also explain what it is and its limitation (implement the official issuer API _but_ extends it with add_employee etc, doesn't support multiple languages, etc)

demos/vc_issuer/src/main.rs
przydatek: Agreed in general, but in this example issuer the goal is to get to the point (i.e. show the flow of the API methods) asap, so I'd prefer not to add another indirection layer.

demos/vc_issuer/src/main.rs
przydatek: Agreed, this could be useful in the future.  Let's add it once we need it.

demos/vc_issuer/src/main.rs
przydatek: Done.

demos/vc_issuer/src/main.rs
przydatek: Correct, added a note.

demos/vc_issuer/src/main.rs
przydatek: Yes. The test below (`check_candid_interface_compatibility()`) verifies that the checked in `vc_issuer.did` is compatible with the did-file generated automatically.

demos/vc_issuer/vc_issuer.did
przydatek: Done.

przydatek (commented):
Thanks @nmattia, PTAL.

demos/vc_issuer/src/main.rs
nmattia: If there's no other reason than "maybe useful someday" then let's remove it

demos/vc_issuer/src/main.rs
nmattia: don't feel strongly about it!

demos/vc_issuer/README.md
```diff
@@ -0,0 +1,15 @@
...
+## Supported Behavior
+
+The app implements the interface of the issuer from [the spec](../../docs/vc-spec.md), plus some additional
+APIs for configuring and testing, cf. [vc_issuer.did](./vc_issuer.did).
+Please note that this issuer is for demonstrating the use of issuer APIS, and is **not meant 
+for real-world deployment**, as it does not have proper management of user data.
+
+## Development
+
+Run `./build.sh` script to build the issuer canister.
```
nmattia: worth adding notes about how to initialize it? I have this in my notes, some of it courtesy of @frederikrothenberger :


```
$ ii_canister_id=$(cat .dfx/local/canister_ids.json | jq -cMr '.internet_identity.local')
$ rootkey=$(curl -s http://localhost/:$(dfx info replica-port)/api/v2/status | npx --package=cbor-cli -- cbor2json | jq -cM '.value.root_key.data' | sed 's/\[/{/g;s/,/;/g;s/\]/}/g')
~/internet-identity$ dfx canister call issuer configure '(record { idp_canister_ids = vec{ principal "'$ii_canister_id'" }; ic_root_key_der = vec '"$rootkey"' })'

$ dfx canister call issuer add_employee '(principal "3d7z4-fytna-ykqnl-dzl5q-dgv5v-rxuhm-j4cf7-2mmbi-6wewg-k7syq-nae")'
```

demos/vc_issuer/build.sh
nmattia, line 1: Would it make sense to run this script somewhere on CI?

demos/vc_issuer/README.md
przydatek: We'll add initialisation instructions in the coming days, once we get something nice-looking -- thanks for sharing the notes: I believe we can iterate on them to get something to commit.

nmattia (approved):
Unblocking

demos/vc_issuer/build.sh
przydatek: Definitely -- @frederikrothenberger enabled tests from vc_issuer on the branch, and we'll do the same for `main`.

demos/vc_issuer/src/main.rs
frederikrothenberger: They're only identical from the perspective of the field _types_ not field _names_. I created two structs because I wanted to represent the format in the field name: `ic_root_key_der` vs `ic_root_key_raw`

For external parties it makes sense to configure the DER encoded key (because this is usually how the key is shared). 
However, internally we only use the `raw` key. I opted to persist the raw key in the stable memory because:
* I wanted the config operation to fail rather than the operation using the key if the DER encoding is broken.
* it seems unnecessary to unpack the DER encoding on every usage.

Of course, one could also validate on config and unpack on each usage, but I didn't think that having 2 similar structs was such a big deal. 🤷

przydatek: Thanks @frederikrothenberger for providing more context, leaving as-is then.

PR #2060, opened 2023-11-17
Extract app wrapper

nmattia:
This extracts everything related to setting up a single-page II app (preloading some assets, setting up polyfills, compatibility checks etc) into a helper function.

This is preparation work for adding new endpoints, so that they won't need to import/use the main app code (more readable + smaller bundlers).

An apparently unnecessary endpoint (`#compatibilityNotice`) was also removed.

NOTE: with the exception of the removed `#compatibilityNotice` and a couple of `s/void/return/`, the code was just moved and not modified


frederikrothenberger (approved):
LGTM, thanks!

PR #2069, opened 2023-11-22
Refactor credential spec verification in issuer

frederikrothenberger:
Introduces an enum to distill down the information in the credential spec to actually supported credentials. This gets rid of unnecessary match arms and is also done in preparation for the consent message rework.


przydatek (approved):
LGTM, thanks!

PR #2070, opened 2023-11-22
Add exp claim to VCs issued by the issuer

frederikrothenberger:
This PR adds the `exp` claim to the VCs issued by the issuer. This will also be relevant for the consent message rework, because the validity will be contained in that message.


przydatek (approved):
LGTM, thanks!

PR #2073, opened 2023-11-23
Add second consent message language and make message more descriptive

frederikrothenberger:
This PR improves the consent messages to be available in 2 languages. The message is improved to be more human readable and is now formatted as markdown.

Additionally, the ICRC21 types were updated to reflect the recent spec changes. Adjusting to the actual types with came case will be done in a separate PR because the issuer API needs adjustment too.

### DIFF
--- a/demos/vc_issuer/src/consent_message.rs
+++ b/demos/vc_issuer/src/consent_message.rs
@@ -1,6 +1,7 @@
 //! This module contains the various consent messages that is displayed to the user when they are asked to consent to the issuance of a credential.
 
 use crate::SupportedCredentialType;
+use lazy_static::lazy_static;
 use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use strfmt::strfmt;
@@ -8,7 +9,6 @@ use vc_util::issuer_api::{
     Icrc21ConsentInfo, Icrc21ConsentMessageResponse, Icrc21ConsentPreferences, Icrc21Error,
     Icrc21ErrorInfo,
 };
-use SupportedCredentialType::{UniversityDegreeCredential, VerifiedEmployee};
 use SupportedLanguage::{English, German};
 
 const EMPLOYMENT_VC_DESCRIPTION_EN: &str = r###"# {employer} Employment Credential
@@ -24,9 +24,65 @@ const DEGREE_VC_DESCRIPTION_DE: &str = r###"# Bachelor of Engineering, {institut
 
 Ausweis, der bestätigt, dass der Besitzer oder die Besitzerin einen Bachelorabschluss in einer Ingenieurwissenschaft des {institute} besitzt."###;
 
+lazy_static! {
+    static ref CONSENT_MESSAGE_TEMPLATES: HashMap<(CredentialTemplateType, SupportedLanguage), &'static str> =
+        HashMap::from([
+            (
+                (CredentialTemplateType::VerifiedEmployee, English),
+                EMPLOYMENT_VC_DESCRIPTION_EN
+            ),
+            (
+                (CredentialTemplateType::VerifiedEmployee, German),
+                EMPLOYMENT_VC_DESCRIPTION_DE
+            ),
+            (
+                (CredentialTemplateType::UniversityDegree, English),
+                DEGREE_VC_DESCRIPTION_EN
+            ),
+            (
+                (CredentialTemplateType::UniversityDegree, German),
+                DEGREE_VC_DESCRIPTION_DE
+            )
+        ]);
+}
+
+#[derive(Clone, Eq, PartialEq, Hash)]
+pub enum CredentialTemplateType {
+    VerifiedEmployee,
+    UniversityDegree,
+}
+
+#[derive(Clone, Eq, PartialEq, Hash)]
 pub enum SupportedLanguage {
     English,
     German,
+}
+
+impl From<&SupportedCredentialType> for CredentialTemplateType {
+    fn from(value: &SupportedCredentialType) -> Self {
+        match value {
+            SupportedCredentialType::VerifiedEmployee(_) => {
+                CredentialTemplateType::VerifiedEmployee
+            }
+            SupportedCredentialType::UniversityDegree(_) => {
+                CredentialTemplateType::UniversityDegree
+            }
+        }
+    }
+}
+
+impl SupportedCredentialType {
+    /// Re-expands a known credential type back into its parameters, which can be used for consent message templating.
+    fn to_param_tuple(&self) -> (String, String) {
+        match self {
+            SupportedCredentialType::VerifiedEmployee(employer) => {
+                ("employer".to_string(), employer.to_string())
+            }
+            SupportedCredentialType::UniversityDegree(institute) => {
+                ("institute".to_string(), institute.to_string())
+            }
+        }
+    }
 }
 
 impl From<Icrc21ConsentPreferences> for SupportedLanguage {
@@ -51,14 +107,7 @@ pub fn get_vc_consent_message(
     credential_type: &SupportedCredentialType,
     language: &SupportedLanguage,
 ) -> Icrc21ConsentMessageResponse {
-    let template_result = match (credential_type, language) {
-        (VerifiedEmployee(employer), English) => employment_consent_msg_eng(employer),
-        (VerifiedEmployee(employer), German) => employment_consent_msg_de(employer),
-        (UniversityDegreeCredential(institute), English) => degree_consent_msg_eng(institute),
-        (UniversityDegreeCredential(institute), German) => degree_consent_msg_de(institute),
-    };
-
-    match template_result {
+    match render_consent_message(credential_type, language) {
         Ok(message) => Icrc21ConsentMessageResponse::Ok(Icrc21ConsentInfo {
             consent_message: message,
             language: format!("{}", language),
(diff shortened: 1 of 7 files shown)


demos/vc_issuer/src/consent_message.rs
```diff
@@ -0,0 +1,107 @@
...
+
+use crate::SupportedCredentialType;
+use std::fmt::{Display, Formatter};
+use vc_util::issuer_api::{
+    Icrc21ConsentInfo, Icrc21ConsentMessageResponse, Icrc21ConsentPreferences,
+};
+use SupportedCredentialType::{UniversityDegreeCredential, VerifiedEmployee};
+use SupportedLanguage::{English, German};
+
+const EMPLOYMENT_VC_DESCRIPTION_ENG: &str = r###"# DFINITY Foundation Employment Credential
```
nmattia, line 11: nit:

```suggestion
const EMPLOYMENT_VC_DESCRIPTION_ENG: &str = r###"# DFINITY Foundation Employment Credential
```

to avoid confusion with ENGineer and for consistency with DE below

nmattia:
> The message is improved to be more human readable and **is now formatted as markdown**.

Does that mean the frontend should render the message as markdown formatted?

demos/vc_issuer/tests/issue_credential.rs
```diff
@@ -163,36 +163,41 @@ mod api {
 /// Verifies that the consent message can be requested.
 #[test]
 fn should_return_vc_consent_message() {
+    let test_cases = [
+        ("en-US", "en", "# DFINITY Foundation Employment Credential"),
+        ("de-DE", "de", "# Beschäftigungsausweis DFINITY Stiftung"),
+        ("ja_JP", "en", "# DFINITY Foundation Employment Credential"), // test fallback language
```
nmattia, line 169: not sure if meant to be so but you have a mix of xx-XX and xx_XX

demos/vc_issuer/src/consent_message.rs
frederikrothenberger: Ah, yes. Much better 👍

frederikrothenberger:
> > The message is improved to be more human readable and **is now formatted as markdown**.
> 
> Does that mean the frontend should render the message as markdown formatted?

Yes, ICRC-21 specifies markdown formatting.

demos/vc_issuer/tests/issue_credential.rs
frederikrothenberger: I didn't, thanks for pointing it out!

demos/vc_issuer/src/consent_message.rs
```diff
@@ -0,0 +1,107 @@
...
+        ANONYMITY_DISCLAIMER_DE,
+    )
+}
+
+fn format_message(description: &str, validity_info: &str, anonymity_disclaimer: &str) -> String {
+    format!(
+        "{} {}\n\n{}",
+        description, validity_info, anonymity_disclaimer
+    )
+}
```
nmattia, line 107: I wonder if it would be simpler & easier to read if the message was in a single part, wdyt?

demos/vc_issuer/src/main.rs
```diff
@@ -482,13 +465,19 @@ fn prepare_credential_payload(
             EMPLOYEES.with_borrow(|employees| {
                 verify_authorized_principal(credential_type, alias_tuple, employees)
             })?;
-            Ok(dfinity_employment_credential(alias_tuple.id_alias, employer_name.as_str()))
+            Ok(dfinity_employment_credential(
+                alias_tuple.id_alias,
+                employer_name.as_str(),
+            ))
```
nmattia, line 471: rustfmt having a bad day? or am I missing something?

demos/vc_issuer/tests/issue_credential.rs
```diff
@@ -163,36 +163,41 @@ mod api {
...
+            api::vc_consent_message(&env, canister_id, principal_1(), &consent_message_request)
+                .expect("API call failed")
+                .expect("Got 'None' from vc_consent_message");
+        assert_matches!(response, Icrc21ConsentMessageResponse::Ok(_));
+        if let Icrc21ConsentMessageResponse::Ok(info) = response {
+            assert_eq!(info.language, actual_language);
+            assert!(info
+                .consent_message
+                .starts_with(consent_message_snippet));
+        }
```
nmattia, line 200: not related to this PR, but would it make sense to panic if the response is not what we expect? `else { panic!(...) }`

demos/vc_issuer/tests/issue_credential.rs
```diff
@@ -331,8 +330,7 @@ fn should_fail_prepare_credential_for_wrong_sender() {
         },
     )
     .expect("API call failed");
-    assert_matches!(response, 
-        PrepareCredentialResponse::Err(IssueCredentialError::UnauthorizedSubject(e)) if e.contains("Caller 6epth-hmqup-wz4mv-svl2m-mhcbb-24skq-tbdhq-2txct-2qugv-xuzva-eqe does not match id alias dapp principal nugva-s7c6v-4yszt-koycv-5b623-an7q6-ha2nz-kz6rs-hawgl-nznbe-rqe."));
+    assert_matches!(response, PrepareCredentialResponse::Err(IssueCredentialError::UnauthorizedSubject(e)) if e.contains("Caller 6epth-hmqup-wz4mv-svl2m-mhcbb-24skq-tbdhq-2txct-2qugv-xuzva-eqe does not match id alias dapp principal nugva-s7c6v-4yszt-koycv-5b623-an7q6-ha2nz-kz6rs-hawgl-nznbe-rqe."));
```
nmattia, line 333: I'm assuming this is just formatting, but I haven't gone through every single char in the caller...

nmattia (approved):
Couple of suggestions but LGTM nonetheless!

przydatek (commented):
Thanks @frederikrothenberger!  
Please see some suggestions/remarks as discussed.

demos/vc_issuer/src/consent_message.rs
```diff
@@ -0,0 +1,107 @@
...
+use vc_util::issuer_api::{
+    Icrc21ConsentInfo, Icrc21ConsentMessageResponse, Icrc21ConsentPreferences,
+};
+use SupportedCredentialType::{UniversityDegreeCredential, VerifiedEmployee};
+use SupportedLanguage::{English, German};
+
+const EMPLOYMENT_VC_DESCRIPTION_EN: &str = r###"# DFINITY Foundation Employment Credential
+
+Credential that states that the holder is employed by the DFINITY Foundation at the time of issuance."###;
+const DEGREE_VC_DESCRIPTION_EN: &str = r###"# Bachelor of Engineering, DFINITY College of Engineering
```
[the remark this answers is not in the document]
frederikrothenberger: Yes, however for that it would make sense to use areal templating engine. `format!` unfortunately _requires_ a string literal.

Should I pull in a templating engine?

demos/vc_issuer/src/consent_message.rs
```diff
@@ -0,0 +1,107 @@
...
+
+Credential that states that the holder has a degree in engineering from the DFINITY College of Engineering."###;
+const VALIDITY_INFO_EN: &str = "The credential is valid for 15 minutes.";
+const ANONYMITY_DISCLAIMER_EN: &str = "This credential does **not** contain any additional personal information. It is issued to an ephemeral identity that is created for the sole purpose of issuing this credential.";
```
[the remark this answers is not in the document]
frederikrothenberger: I'm not sure if I agree here. It really depends on how narrow the definition of the ICRC-21 VC extension is.
Is this extension also suitable for non-privacy preserving IDPs? Can it also be used by issuers that issue long-term credentials to be stored in a wallet?

Given we do not show the actual credential anywhere in the flow, users might rightfully wonder what else is contained there. Especially for non-IC native users the anonymity is not a property that can be taken for granted. The currently available PoCs are still often employing the model where you store a _static_ VC in your wallet and show the same thing to anyone that asks.

demos/vc_issuer/src/consent_message.rs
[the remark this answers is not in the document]
przydatek: Sure, I don't mind adding a (not-too-complex) dependency if that enables the templating.

demos/vc_issuer/src/consent_message.rs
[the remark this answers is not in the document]
przydatek: Agreed, there might be some scenarios where more information is needed.  But the main goal of this issuer is to participate in a typical (for now) VC-flow, and for this purpose we don't have to cover all potential use cases of ICRC-21 or its extensions (and btw. the ICRC-21 VC extension will be defined after we see what works and what doesn't 😉 ).  
If you prefer to keep these extra lines in the consent message, I'm fine with that (and will LGTM), but am afraid that once we start demoing it with real UX, we'll have to remove these anyway.  
I believe one way of adding such additional info in the future would be to add extra fields to the consent message-response, which then can be handled by the FE appropriately (e.g. display only the main consent message, and add [more info]-button with the information from the additional field(s)).

demos/vc_issuer/src/consent_message.rs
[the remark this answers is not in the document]
przydatek: maybe https://crates.io/crates/tinytemplate would be suitable?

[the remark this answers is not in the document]
frederikrothenberger: I opted for https://crates.io/crates/strfmt because it seems even tinier.

src/vc_util/src/issuer_api.rs
```diff
@@ -121,13 +121,8 @@ pub struct Icrc21ErrorInfo {
 
 #[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
 pub enum Icrc21Error {
-    #[serde(rename = "forbidden")]
```
przydatek, line 124: So far for the issuer we were following the convention from II's API, with lower_case_names_and_labels.  Is the use of CamelCase here mandated by ICRC-21?  I believe it would be good if we had only once convention throughout VC-issuer API.

demos/vc_issuer/src/consent_message.rs
```diff
@@ -0,0 +1,112 @@
...
+    match template_result {
+        Ok(message) => Icrc21ConsentMessageResponse::Ok(Icrc21ConsentInfo {
+            consent_message: message,
+            language: format!("{}", language),
+        }),
+        Err(err) => Icrc21ConsentMessageResponse::Err(Icrc21Error::GenericError(err)),
+    }
+}
+
+fn employment_consent_msg_eng(employer: &str) -> Result<String, Icrc21ErrorInfo> {
```
przydatek, line 70: nit: if we store the templates in a map (language, VC_type) -> template, we would have only half of these helpers, one per VC_type, e.g.: 
```
fn employment_consent_msg(employer: &str, lang: SupportedLanguage) ... {
       strfmt(
          consent_msg_template((lang, "VerifiedEmployee")),
          &HashMap::from([("employer".to_string(), employer)]),
    )...
} 
```

przydatek (commented):
Thanks a lot for the updates @frederikrothenberger!
I have one more question, and one minor suggestion, PTAL.

src/vc_util/src/issuer_api.rs
frederikrothenberger: Yes, ICRC-21 mandates camel case. It does so to be consistent with ICRC-1, 2 & 3, which all use camel case variants. From the other officially specified interfaces, the [HTTP gateway protocol](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec#response-body-streaming-interface) also uses camel case, whereas the [management canister](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-candid) does not.

In general it seems to me, that newer interfaces are specified to use camel case. So I would argue, that we should actually use camel case in the issuer interface as well.

przydatek: Ok, we can change to CamelCase, but maybe let's do that in a separate PR for all of the issuer API, in coordination with API users. I.e. maybe let's keep lower_case for this PR?

przydatek (approved):
LGTM, thanks!

PR #2074, opened 2023-11-24
Bump vitest

nmattia:
This bumps vitest to work around a bug in recent Node: https://github.com/vitest-dev/vitest/issues/3826

The bump caused some failures in the crypto ed25519 tests on node where subtle did not like an ArrayBuffer as an argument. It turns out that it does not need a buffer at all and the original array can be supplied instead.

Finally the order of test commands in the workflow is changed to fail early if the fast tests fail.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -487,6 +487,10 @@ jobs:
           echo node --version
           node --version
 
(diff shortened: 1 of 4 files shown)


frederikrothenberger (approved):
LGTM, thanks!

PR #2075, opened 2023-11-24
Simplify canister ID lookup

nmattia:
This simplifies the way canister IDs are looked up, both for static config & dynamically in the devserver.

In general, canister IDs are not looked up from the canister_ids.json anymore, but instead `dfx canister id <CANISTER_NAME>` is used (so that we don't rely on dfx internals).

The dynamic replica-forwarding is changed to avoid having to specify canister IDs; instead domains can be specified which are treated as `<canister ID>.<domain>` and forwarded to the appropriate canister.

### DIFF
--- a/src/frontend/src/test-e2e/constants.ts
+++ b/src/frontend/src/test-e2e/constants.ts
@@ -1,15 +1,12 @@
-// Read canister ids from the corresponding dfx files.
-// This assumes that they have been successfully dfx-deployed
-import { readFileSync } from "fs";
-export const test_app_canister_ids = JSON.parse(
-  readFileSync(".dfx/local/canister_ids.json", "utf-8")
-);
+import { readCanisterId } from "../../../../utils";
 
-const TEST_APP_CANISTER_ID = test_app_canister_ids.test_app.local;
+// XXX: this is not exactly a constant (since it might change on every node eval) but in
+// practice is very stable, and is much easier to use as "constants" then as a lookup function.
+const testAppCanisterId = readCanisterId({ canisterName: "test_app" });
 
-export const TEST_APP_CANONICAL_URL = `https://${TEST_APP_CANISTER_ID}.icp0.io`;
-export const TEST_APP_CANONICAL_URL_RAW = `https://${TEST_APP_CANISTER_ID}.raw.icp0.io`;
-export const TEST_APP_CANONICAL_URL_LEGACY = `https://${TEST_APP_CANISTER_ID}.ic0.app`;
+export const TEST_APP_CANONICAL_URL = `https://${testAppCanisterId}.icp0.io`;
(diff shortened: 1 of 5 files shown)


vite.plugins.ts
```diff
@@ -123,26 +96,49 @@ export const replicaForwardPlugin = ({
...
-      req.headers["host"] = `${matchingRule.canisterId}.localhost`;
-      proxy.web(req, res, {
-        target: `http://${replicaOrigin}`,
-      });
+      // split the subdomain & domain by splitting on the first dot
+      const [subdomain, ...domain_] = host.split(".");
+      const domain = domain_.join(".");
+
+      if (
+        forwardDomains !== undefined &&
```
nmattia, line 131:
```suggestion
        !isNullish(forwardDomains) &&
```

vite.plugins.ts
```diff
@@ -92,10 +63,12 @@ export const minifyHTML = (): {
  */
 export const replicaForwardPlugin = ({
   replicaOrigin,
+  forwardDomains /* note: only exact match, i.e. not 'raw'*/,
```
nmattia, line 66:
```suggestion
  forwardDomains /* note: will match exactly on <canister>.<domain> */,
```

utils.ts
nmattia, line 1: I'll move this soon (top level is not ideal) but I don't know where exactly

src/frontend/src/test-e2e/constants.ts
```diff
@@ -1,15 +1,12 @@
...
-// This assumes that they have been successfully dfx-deployed
-import { readFileSync } from "fs";
-export const test_app_canister_ids = JSON.parse(
-  readFileSync(".dfx/local/canister_ids.json", "utf-8")
-);
+import { readCanisterId } from "../../../../utils";
 
-const TEST_APP_CANISTER_ID = test_app_canister_ids.test_app.local;
+// XXX: this is not exactly a constant (since it might change on every node eval) but in
+// practice is very stable, and is much easier to use as "constants" then as a lookup function.
```
nmattia, line 4:
```suggestion
// practice is very stable, and is much easier to use as "constants" than as a lookup function.
```

frederikrothenberger (approved):
Nice, just one question 👍

src/frontend/src/test-e2e/constants.ts
```diff
@@ -1,15 +1,12 @@
...
+const testAppCanisterId = readCanisterId({ canisterName: "test_app" });
```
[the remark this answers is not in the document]
nmattia: It should throw an exception which gets bubbled up to the user, I'm changing the message a bit to clarify:

```diff
-    throw Error(`Could not get canister ID with command '${command}': ${e}`);
+    throw Error(`Could not get canister ID for '${canisterName}' with command '${command}', was the canister deployed? ${e}`);
```

wdyt?

PR #2076, opened 2023-11-24
Fix CI not formatting the vc issuer correctly

frederikrothenberger:
This removes the filter to only care about changes in `src` and istead picks up all `.rs` files.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,7 +31,7 @@ jobs:
         # We don't want to commit formatting changes to main
         if: ${{ github.ref != 'refs/heads/main' }}
         with:
-          add: src
+          add: "**/*.rs"
           default_author: github_actions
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -31,7 +31,6 @@ jobs:
         # We don't want to commit formatting changes to main
         if: ${{ github.ref != 'refs/heads/main' }}
         with:
-          add: src
```
nmattia, line 34: I'd keep this in and whitelist the dirs we expect to change; I've seen dot files & more committed by mistake too often. Alternatively, does this support something like `**/*.rs`?

frederikrothenberger: Apparently, yes 👍

nmattia (approved):
Nice catch, thanks!

PR #2078, opened 2023-11-24
Provision issuer canister

nmattia:
The issuer canister needs to know the local replica's rootkey and the canister ID of the IDP (II) canister.

This adds a script to configure the canister, and a `post_install` entry in the `dfx.json` to ensure the canister is provisioned after having been installed.

### DIFF
--- a/demos/vc_issuer/provision
+++ b/demos/vc_issuer/provision
@@ -44,6 +44,9 @@ rest="${BASH_REMATCH[2]}"
 # trim to size (double the size because each byte is written as two chars)
 rootkey_hex="${rest:0:$((2*size))}"
 
+# NOTE: while theoretically candid supports `hex` blobs, `dfx` fails to parse them
+# so we have to convert this into a byte array.
+
 # new regex for matching pairs of chars (i.e. bytes)
 regex="([0-9a-z][0-9a-z])"
 arr="" # where we'll hold the resulting candid array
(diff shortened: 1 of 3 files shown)


demos/vc_issuer/provision
```diff
@@ -0,0 +1,69 @@
...
+size_hex="${BASH_REMATCH[1]}"
+size="$((16#$size_hex))"
+
+# Grab the untrimmed rest bytes
+rest="${BASH_REMATCH[2]}"
+
+# trim to size (double the size because each byte is written as two chars)
+rootkey_hex="${rest:0:$((2*size))}"
+
+# NOTE: while theoretically candid supports `hex` blobs, `dfx` fails to parse them
```
przydatek, line 47: nit: while dfx apparently doesn't support "pure" hex strings, it does work with `\xx` (i.e. with `\`-prefix) hex encoding (cf. https://forum.dfinity.org/t/how-to-use-dfx-to-query-the-balance-of-an-account-identifier/15929/2).  Using such encoding also doesn't yield a nice solution, but could be a slight improvement.

przydatek (approved):
Thanks a lot @nmattia! The solution is a bit complex, but I guess there is not much we ca n do about it.

LGTM, with a minor suggestion.

nmattia:
> The solution is a bit complex, but I guess there is not much we ca n do about it.

@frederikrothenberger pointed out to me that `dfx ping` shows the status in (broken) JSON, so I switched to that!

PR #2080, opened 2023-11-27
Introduce rust workspace dependencies

frederikrothenberger:
Using workspace dependencies makes it easier to manage common dependencies. This change is made in preparation for rust dependency upgrades of the ic dependencies.


nmattia (approved):
LGTM, thanks!

PR #2081, opened 2023-11-27
Change issuer API to camel case

frederikrothenberger:
Newer standards, such as ICRC-1,2,3 as well as the HTTP gateway spec all use came case for candid variants. To be consistent with this new development, we will do the same for the issuer API.

The rename from `credential_name` -> `credential_type` is bundled in as well, to reduce the number of breaking changes.


przydatek (approved):
LGTM, thanks!

PR #2083, opened 2023-11-28
Add issuer app

nmattia:
This adds an app for the issuer canister, that allows authenticating & adding the authenticated user as an employee.

This includes a couple changes:

* The issuer canister is updated to allow serving assets. Most of the code for asset serving is copied from `test-app` and II itself. This code will ideally be extracted and replaced with an asset serving library at some point. Includes a fix for certification by @frederikrothenberger. Thanks!
* An `app/` directory is added to the issuer. This is a simple React app with a button to authenticate and a button to add the authenticated user as an employee. The vite plugins are copied from the top-level app, and will need to be deduplicated at some point. However for now we keep the app independent from the rest of the II codebase.
* An e2e test is added to show how to use the issuer app.


demos/vc_issuer/app/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+          ) : (
+            <output data-role="principal" data-unset>
+              not authenticated
+            </output>
+          )}
+        </label>
+      </section>
+      <section>
+        <label>
+          Canister Logs:
```
frederikrothenberger, line 94: Nit: Maybe rename to canister interaction log? Otherwise, this might suggest that you are somehow getting to the log output of the canister itself.

frederikrothenberger (approved):
Already approved to unblock, only minor things. Thanks 👍

demos/vc_issuer/src/main.rs
```diff
@@ -354,6 +375,33 @@ fn add_graduate(graduate_id: Principal) -> String {
     format!("Added graduate {}", graduate_id)
 }
 
+#[query]
```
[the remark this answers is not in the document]
nmattia: Wops, yes! Good catch

demos/vc_issuer/app/index.tsx
nmattia: Makes sense! I've renamed it to "Canister Replies" although I'm not sure it's much clearer; I find "Canister Interaction Logs" a bit long. I'll see if I get a better idea over the next few days.

PR #2084, opened 2023-11-28
Don't bundle fonts twice

nmattia:
This moves the fonts to not be served as part of the assets directory.

Vite already bundles the fonts because they are imported from the CSS so there's no need to serve them twice.


peterpeterparker (approved):
Rubber stamp. I never ever had an app in which I had to move the fonts to the `src` folder. 🤷‍♂️

PR #2090, opened 2023-11-30
Clean up test-app

nmattia:
This cleans up the test app, porting it to typescript, fixing some bugs and unifying with the issuer app.

In particular:

* The test-app is now typechecked (`.js` -> `.ts`)
* A couple of bugs (missing variables, missing elements) are fixed
* The vite config is fixed: it used to proxy all asset requests (even index.html) to the canister (making local development useless)
* A small CSS file is copied from the issuer dapp


frederikrothenberger (approved):
Thanks!

Nit: It would be nice if the included CSS would allow a wider column and or would do word-wrap. Compared to the old version the delegation debug view is less usable IMHO.
New: 
Screenshot 2023-11-30 at 11 41 48
Old:
Screenshot 2023-11-30 at 11 44 34

nmattia:
Thanks! Here's the updated version. Not very ergonomic still but I might have another go at it if I end up using the field more.
Screenshot 2023-11-30 at 11 56 18

PR #2091, opened 2023-11-30
Move asset certification code into a new separate crate

frederikrothenberger:
This PR moves the asset certification code to a new crate, later to be reused by all asset serving canisters in this repository.

This PR intentionally only _moves_ code (with the only exception of switching out the implementation for the representation independent hash). There will be a subsequent PR to make the library interface nicer.

The code was only moved from `assets.rs` with the exception of the  `ContentType` impl. The functionality for certification headers that
 currently resides in `http.rs` will be added to the crate in later PR as well.


przydatek (approved):
LGTM, thanks!

src/asset_util/src/lib.rs
```diff
@@ -0,0 +1,420 @@
...
+            ContentType::OCTETSTREAM => "application/octet-stream".to_string(),
+            ContentType::PNG => "image/png".to_string(),
+            ContentType::SVG => "image/svg+xml".to_string(),
+            ContentType::WOFF2 => "application/font-woff2".to_string(),
+        }
+    }
+}
+
+// The <script> tag that loads the 'index.js'
+const JS_SETUP_SCRIPT: &str = "let s = document.createElement('script');s.type = 'module';s.src = '/index.js';document.head.appendChild(s);";
```
nmattia, line 113: does it really make sense to make this part of the lib? I feel like this is very specific

PR #2093, opened 2023-12-01
Make origin remapping consistent

nmattia:
When the "new" domains (icp0.io & internetcomputer.org) were introduced, a workaround was implemented in II to ensure that dapps would see the same principals regardless on whether they are using CANISTER.ic0.app or CANISTER.icp0.io.

This workaround was implemented for fetching the derivation, but actually needs to be applied to all operations deriving a principal from an identity number/origin pair.

This change makes sure that all calls perform the remapping, including `get_principal` (which will be used in Verifiable Credentials).


frederikrothenberger (approved):
LGTM, thanks!

PR #2097, opened 2023-12-04
Add verifiable credentials flow in frontend

nmattia:
This adds a "verifiable credentials" flow to the frontend. In four commits:

1. The canister code is updated to accept 2 index.js files; without this, the browser will refuse to fetch the new index file for the second flow due to bad CSP.
2. Adds the actual VC postMessage interface, flow & screens to the II webapp.
3. Updates the test-app by introducing a few react component to send & receive VC related requests to an IDP (II).
4. A couple of tests are added to go through a full VC flow.

This is more of a "clean proof-of-concept" than an actual feature, as many things are still missing:

- The VC flow does not support PIN identities
- The SRI hash fix is a bit ugly and should probably be fixed more generally (for 2+ index files)
- The issuer types are duplicated between the issuer app and II; moreover in both cases they include both the actual Issuer API (specced) and the helper API like `add_employee`. This should be cleaned up somehow.
- The tests merely checks that the flow is successful; they should also check the validity of the presentation and be extended to test for non-happy cases. 
- There is no support for changing the consent message language
- The test-app requests are not typechecked (which might cause frustration if the types ever change)
- The consent message is not shown as rendered markdown

All of these will come in other PRs.


src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    credentialSubject: givenP_RP,
+    issuer: { origin: issuerOrigin, canisterId: issuerCanisterId_ },
+    credentialSpec,
+  },
+  rpOrigin,
+}: { connection: Connection } & VerifyCredentialsArgs) => {
+  // Use the issuer canister ID provided by the RP, or look it up
+  // from the issuer origin if no canister ID was provided.
+  // NOTE: we do _not_ check the origin's canister ID if a canister ID was provided explicitly.
+  let issuerCanisterId = issuerCanisterId_?.toText();
```
frederikrothenberger, line 78: We _must_ always look up the canister id. @przydatek only included the option for the relying party to provide the canister id as an extra validation to make sure that if the flow succeeds, that II also resolved the same canister id.

src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    );
+    if (lookedUp === "not_found") {
+      return abortedCredentials({ reason: "no_canister_id" });
+    }
+
+    issuerCanisterId = lookedUp.ok;
+  }
+
+  const vcIssuer = new VcIssuer(issuerCanisterId);
+  // XXX: We don't check that the language matches the user's language. We need
```
frederikrothenberger, line 92: Probably, it does not make sense to support more languages, if the rest of the UI doesn't support it either...

src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    console.error(
+      [
+        "bad principals for user number/origin",
+        userNumber,
+        rpOrigin,
+        computedP_RP.toString(),
+        givenP_RP.toString(),
+      ].join(", ")
+    );
+    return abortedCredentials({ reason: "bad_principal_rp" });
```
frederikrothenberger, line 143: Instead of cancelling and letting the RP know, shouldn't there be a retry option on II?

src/internet_identity/src/assets.rs
```diff
@@ -19,23 +19,35 @@ pub fn init_assets() {
 }
 
 // The <script> tag that loads the 'index.js'
-const JS_SETUP_SCRIPT: &str = "let s = document.createElement('script');s.type = 'module';s.src = '/index.js';document.head.appendChild(s);";
+pub const JS_SETUP_SCRIPT1: &str = "let s = document.createElement('script');s.type = 'module';s.src = '/index.js';document.head.appendChild(s);";
```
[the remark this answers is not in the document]
nmattia: Yeah this is a quick hack. We're changing a couple of things in the way the assets are served and this will go away very soon. Will basically read the `indexX.js` bit from the HTML and then inject that. Thanks for checking though!

src/frontend/src/flows/verifiableCredentials/index.ts
nmattia: @frederikrothenberger PTAL

src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -0,0 +1,415 @@
...
+      origin: rpOrigin,
+    })
+  );
+  if (computedP_RP.compareTo(givenP_RP) !== "eq") {
+    console.error(
+      [
+        "bad principals for user number/origin",
+        userNumber,
+        rpOrigin,
+        computedP_RP.toString(),
```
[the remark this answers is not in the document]
nmattia: Done, I've also removed the RP & anchor since it's data known to the user anyway

PR #2098, opened 2023-12-04
Add issuer lockfile to up-to-date CI check

frederikrothenberger:
The issuer Cargo.lock was so far not checked by CI. This allowed checking in out-of-sync Cargo.lock / .toml files.


nmattia (approved):
Thanks!

PR #2099, opened 2023-12-04
Clean up asset_util interface

frederikrothenberger:
This PR cleans up the asset_util crate interface:
* hide internals of `CertifiedAssets`
* provide unified interface for both asset certification versions
* provide support for canisters that do not certify the `sigs` subtree
* add documentation to the public functions

This PR is step **3** of 4:
1. move code from `assets.rs` to the crate
2. move certification header code from `http.rs` to the crate
3. refactor the crate a bit to make the public interface nicer
4. use the crate in the other canisters


src/asset_util/src/lib.rs
```diff
@@ -28,9 +28,17 @@ pub const IC_CERTIFICATE_EXPRESSION: &str =
 /// certification to be valid.
 #[derive(Debug, Default, Clone)]
 pub struct CertifiedAssets {
```
przydatek, line 30: nit: maybe rename to `CertifiedAssetsStore` or something similar, as otherwise it is slightly confusing that we have two structs `CertifiedAssets` and `CertifiedAsset` and the former is not a container of the latter (in the definition of the structs).

src/asset_util/src/lib.rs
```diff
@@ -49,8 +57,39 @@ impl CertifiedAssets {
         )
     }
 
-    fn witness_v1(&self, path: &str) -> HashTree {
-        let witness = self.certification_v1.witness(path.as_bytes());
+    /// Returns the [CertifiedAsset] for the given URL path and certificate version, if it exists.
+    /// If the canister also uses the `certified_data` to issue canister signatures, the caller
```
przydatek, line 61: The remark about `certified_data` here is slightly confusing, as it seems that it applies to an argument of the function, while there is no such argument.
Maybe add a link to spec of `certified_data`? (here and/or to the comment in line 27, where `certified_data` is first mentioned)

src/asset_util/src/lib.rs
```diff
@@ -282,22 +335,26 @@ fn response_hash(headers: &[HeaderField], body_hash: &Hash) -> Hash {
     response_hash
 }
 
-pub fn collect_assets_recursive(
-    dir: &Dir,
-    html_transformer: fn(&str) -> String,
-) -> Vec<(String, Vec<u8>, ContentEncoding, ContentType)> {
+#[derive(Debug, Clone)]
+pub struct Asset {
```
przydatek, line 339: nit: maybe move closer to the top (near `CertifiedAssets` and `CertifiedAsset`), and comment, so that it is clear what are the functions that these structs play?

src/asset_util/src/lib.rs
```diff
@@ -191,16 +235,21 @@ lazy_static! {
 
 /// Takes a list of assets and returns a [CertifiedAssets] struct containing the assets and their
```
przydatek, line 236: nit: this line could be also improved, as "Takes a list of assets" is not really informative (as one sees it immediately).

src/asset_util/src/lib.rs
```diff
@@ -191,16 +235,21 @@ lazy_static! {
...
 /// certification.
-pub fn certify_assets(
-    assets: Vec<(String, Vec<u8>, ContentEncoding, ContentType)>,
-    shared_headers: &[HeaderField],
-) -> CertifiedAssets {
+/// The [CertifiedAssets::root_hash] must be included in the canisters `certified_data` for the
+/// certification to be valid.
+pub fn certify_assets(assets: Vec<Asset>, shared_headers: &[HeaderField]) -> CertifiedAssets {
```
przydatek, line 240: shouldn't it be a constructor of `CertifiedAssets`?

przydatek (commented):
Please see some minor comments/suggestions below.

src/asset_util/src/lib.rs
```diff
@@ -49,8 +57,39 @@ impl CertifiedAssets {
...
+            .map(|(headers, content)| CertifiedAsset {
+                headers: headers.clone(),
+                content: content.clone(),
+            });
+        certified_asset.map(|mut certified_asset| {
+            match certificate_version {
+                Some(2) => certified_asset
+                    .headers
+                    .extend(self.certificate_headers_v2(url_path, sigs_tree)),
+                // Certification v1 is also the fallback for unknown certificate versions
```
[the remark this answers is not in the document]
frederikrothenberger: The problem is that originally clients did not specify a version at all. So in order to not break old clients, we have to assume that they only understand v1.
But I'll make the logic more robust, so that anything `>=2` uses version 2 and nothing or 1 use version 1. 👍

src/asset_util/src/lib.rs
frederikrothenberger: Well semantically it does. The crate does not give visibility into the internals of `CertifiedAssets` (and a consumer shouldn't care). But you can retrieve a `CertifiedAsset` from `CertifiedAssets` using the `certified_asset` function. So all is as expected?

src/asset_util/src/lib.rs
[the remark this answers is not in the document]
przydatek: as discussed, let's rename `certificate_version` to `max_certificate_version`, and specify in a comment that it means something like "max. certificate version expected by the caller", and that the actually returned version can be lower.

przydatek (approved):
LGTM, thanks!

PR #2100, opened 2023-12-04
Add markdown support for consent message

nmattia:
This adds support for rendering consent messages as markdown.

We use [`marked`](https://www.npmjs.com/package/marked) for rendering the markdown to HTML, and then sanitize the HTML with [`DOMPurify`](https://www.npmjs.com/package/dompurify). The consent message is rendered async and the original message (raw markdown) is shown until the markdown was rendered (in practice, instantaneous).

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "format-check": "prettier --check src/showcase src/frontend tsconfig.json .eslintrc.json vite.config.ts vite.plugins.ts vitest.config.ts demos"
   },
   "devDependencies": {
+    "@types/dompurify": "^3.0.5",
     "@types/html-minifier-terser": "^7.0.0",
     "@types/http-proxy": "^1.17.14",
     "@types/selenium-standalone": "^7.0.1",
@@ -63,9 +64,11 @@
     "@dfinity/utils": "^0.0.20",
     "bip39": "^3.0.4",
     "buffer": "^6.0.3",
+    "dompurify": "^3.0.6",
     "idb-keyval": "^6.2.1",
     "jose": "^5.1.3",
     "lit-html": "^2.7.2",
+    "marked": "^11.0.0",
     "process": "^0.11.10",
     "qr-creator": "^1.0.0",
     "stream-browserify": "^3.0.0",
--- a/src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
+++ b/src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
@@ -1,6 +1,7 @@
 import { mkAnchorInput } from "$src/components/anchorInput";
 import { mainWindow } from "$src/components/mainWindow";
 import { I18n } from "$src/i18n";
+import { markdownToHTML } from "$src/utils/html";
 import { mount, renderPage, TemplateElement } from "$src/utils/lit-html";
 import { Chan } from "$src/utils/utils";
 import { html, TemplateResult } from "lit-html";
@@ -8,7 +9,6 @@ import { asyncReplace } from "lit-html/directives/async-replace.js";
 import { unsafeHTML } from "lit-html/directives/unsafe-html.js";
 
 import DOMPurify from "dompurify";
-import { parse as parseMarked } from "marked";
 
 import copyJson from "./allowCredentials.json";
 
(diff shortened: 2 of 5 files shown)


src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
```diff
@@ -1,8 +1,14 @@
...
-import { mount, renderPage } from "$src/utils/lit-html";
-import { TemplateResult, html } from "lit-html";
+import { mount, renderPage, TemplateElement } from "$src/utils/lit-html";
+import { Chan } from "$src/utils/utils";
+import { html, TemplateResult } from "lit-html";
+import { asyncReplace } from "lit-html/directives/async-replace.js";
+import { unsafeHTML } from "lit-html/directives/unsafe-html.js";
+
+import DOMPurify from "dompurify";
+import { parse as parseMarked } from "marked";
```
nmattia, line 11: @peterpeterparker want me to lazy load this? in practice, in this flow this will always be imported

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
```diff
@@ -35,6 +41,15 @@ const allowCredentialsTemplate = ({
     onSubmit: (userNumber) => onAllow(userNumber),
   });
 
+  const consentMessage = new Chan<TemplateElement>(html`${consentMessage_}`);
+
+  // Kickstart markdown parsing & sanitizing; once done, replace the consent message
+  void (async () => {
```
peterpeterparker, line 47: can this piece of code throw an error and if yes, should we catch it?

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
```diff
@@ -35,6 +41,15 @@ const allowCredentialsTemplate = ({
...
+    const parsed = await parseMarked(consentMessage_);
```
peterpeterparker, line 48: in NNS dapp, @mstrasinskis has implemented various functions before and at the top of marked to either provide more security (e.g. pre-escape svg or render images as links) or tweak rendering (e.g. render link with target blank). Would those be useful here?

source: https://github.com/dfinity/nns-dapp/blob/4619857a316be5b47a950c6b8461ffcaa1a3dde3/frontend/src/lib/utils/html.utils.ts#L94

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
peterpeterparker: Thanks for thinking at it. Agree it's fine given the use case and the fact that we ship on the IC ("often faster a big chunk of js than chunks").

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
nmattia: if anything is thrown there's not much we can do about it so I'm happy to let it escape; wdyt?

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
nmattia: nice, let me have a look!

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
peterpeterparker: It there is an error and we do nothing, does the user ends (UX wise) in an unknown state?

If not, all good
If yes, should we inform it something went wrong?
If not, all good
If yes, then we might want to catch the error in some ways.

nmattia: > does the user ends (UX wise) in an unknown state?

nope, they just get the non-rendered markdown!

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
nmattia: Included, thanks 👍

package.json
```diff
@@ -63,9 +64,11 @@
     "@dfinity/utils": "^0.0.20",
     "bip39": "^3.0.4",
     "buffer": "^6.0.3",
+    "dompurify": "^3.0.6",
     "idb-keyval": "^6.2.1",
     "jose": "^5.1.3",
     "lit-html": "^2.7.2",
+    "marked": "^11.0.0",
```
peterpeterparker, line 71: out of curiosity, you did not needed `@types/marked`?
I added those in my pet weekly project, so gonna remove it if useless

peterpeterparker (approved):
LGTM, thx

package.json
nmattia: apparently not; I saw a comment about type issues in the html utils from NNS dapp but here it seems like everything worked out of the box

nmattia: Indeed, I have this locally: `node_modules/marked/lib/marked.d.ts`

peterpeterparker: Coolio, thanks

PR #2101, opened 2023-12-04
Deploy issuer & test app canisters on release

nmattia:
This adds automated deploys of the issuer & of the test app upon new II releases. The issuer and test app can then be used to test the latest II changes.

NOTE: the provisioning script for the issuer was updated to allow running against a mainnet deploy.
NOTE: the test app & issuer canisters don't (yet?) have logic to point to e.g. the test II on mainnet (the URL has to be set manually).

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -633,11 +633,12 @@ jobs:
         run: |
           dfx stop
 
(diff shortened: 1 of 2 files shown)


frederikrothenberger (approved):
LGTM, thanks!

PR #2102, opened 2023-12-04
Proxy CANISTER.localhost

nmattia:
This changes the local setup -- in particular, the `replicaForwardPlugin` -- to allow redirecting to canister by names.

Until now, `NAME.localhost` had to be added manually as a host, alongside with a canister name.

With these changes, the replica forward plugin will try find a local canister `foo` to proxy to for any `foo.localhost:<vite port>` request.


frederikrothenberger (approved):
LGTM, thanks!

PR #2103, opened 2023-12-04
Ignore unknown test app messages

nmattia:
This updates the test app to ignore unknown window events. Some extensions, like metamask, send seemingly arbitrary messages to the window which confuses the test app.

With these changes we try to figure out if a message was intended for us by

1. Ensuring a request was started
2. Checking the jsonrpc response ID


frederikrothenberger (approved):
LGTM, works with Metamask active in FF now. Thanks!

PR #2105, opened 2023-12-05
Retry chrome start

nmattia:
webdriverio sometimes fails to download/find chrome, which causes spurious test failures. This happens rarely, but since every (100 or so) test starts chrome, and since we run all tests 4 times on CI (mobile/desktop, ic0.app/internetcomputer.org) the issues come up on almost every CI run.

This adds a retry to the chrome fetching.


frederikrothenberger (approved):
Thanks, LGTM!

PR #2107, opened 2023-12-05
Rename vc_issuer.wasm.gz and candid file

frederikrothenberger:
This PR renames the issuer wasm module and candid file to better convey that this is a demo / reference implementation and not a production asset.


nmattia (approved):
Thanks!

PR #2108, opened 2023-12-05
Limit local canister lookup to localhost

nmattia:
This limits the vite plugin canister name lookup to be performed only on `CANISTER.localhost`. Not doing this causes slowdowns in two cases:

* `identity.internetcomputer.org` used in tests
* `localhost:5173` used in local npm run dev tests

### DIFF
--- a/vite.plugins.ts
+++ b/vite.plugins.ts
@@ -137,11 +137,15 @@ export const replicaForwardPlugin = ({
       }
 
       // split the subdomain & domain by splitting on the first dot
-      const [subdomain, ...domain_] = host.split(".");
-      const domain = nonNullish(subdomain) ? domain_.join(".") : subdomain;
+      const [subdomain_, ...domain_] = host.split(".");
+      const [subdomain, domain] =
+        domain_.length > 0
+          ? [subdomain_, domain_.join(".")]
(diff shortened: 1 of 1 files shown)


vite.plugins.ts
```diff
@@ -151,13 +153,15 @@ export const replicaForwardPlugin = ({
...
       // Try to read the canister ID of a potential canister called <subdomain>
       // and if found forward to that
-      try {
-        const canisterId = execSync(`dfx canister id ${subdomain}`)
-          .toString()
-          .trim();
-        console.log("Forwarding to", canisterId);
-        return forwardToReplica({ canisterId });
-      } catch {}
+      if (domain === "localhost") {
```
frederikrothenberger, line 156: I found this logic confusing to follow. I.e. it was not immediately clear to me that when accessing `localhost:5173` the `domain` is the empty string and thus this `if` is not triggered (and the _subdomain_ is `localhost`).

Maybe add a comment for other lost souls like me. 🙃

frederikrothenberger (approved):
Thanks for fixing that!

vite.plugins.ts
frederikrothenberger: Much clearer, thanks! 👍

PR #2109, opened 2023-12-05
Pre-fill anchors in VC flow

nmattia:
This updates the storage to store (a digest of) principals of visited dapps. This allows the VC flow to know (in the general, best case scenario) which anchor (identity number) the user is intending to use.

The storage migration includes three things:

* Anchors are not written directly as `{ 10000:  <data> }` anymore but are nested as `{ anchors: { 10000: <data> } }` to allow for a new top-level field `hasher` which is global to all anchors.
* Said new field `hasher` is introduced
* Anchors themselves now have a new field, `knownPrincipalDigests`.

See storage module documentation for more information on the digests.

### DIFF
--- a/src/frontend/src/storage/index.test.ts
+++ b/src/frontend/src/storage/index.test.ts
@@ -18,7 +18,7 @@ test("anchors default to nothing", async () => {
 });
 
 test(
-  "old userNumber is recovered",
+  "old userNumber V0 is recovered",
   withStorage(
     async () => {
       expect(await getAnchors()).toStrictEqual([BigInt(123456)]);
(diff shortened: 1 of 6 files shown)


src/frontend/src/storage/index.test.ts
```diff
@@ -117,23 +115,46 @@ test(
 );
 
 test(
-  "anchors are also written to localstorage",
+  "V2 anchors are migrated",
```
frederikrothenberger, line 118: I feel like for the complexity and size of this PR, the storage V3 is really light on tests.
I.e. there are no test for:
- principal storage / lookup
- making sure many principals are stored
- making sure the principals are actually hashed

src/frontend/src/storage/index.ts
```diff
@@ -40,69 +27,151 @@ export const getAnchors = async (): Promise<bigint[]> => {
...
+    const digest = await computePrincipalDigest({
+      origin,
+      principal,
+      hasher: storage.hasher,
+    });
+
+    const oldAnchor = anchors[ix] ?? { knownPrincipalDigests: [] };
+
+    const knownPrincipalDigests = [
+      ...new Set(oldAnchor.knownPrincipalDigests.concat([digest])),
```
frederikrothenberger, line 98: I think it might be beneficial to limit the size of this array somehow.
In particular, developers that might use many deployments of II on some host (i.e. `localhost`) might get a disproportionate number of "knownPrincipals" for the anchor 10000. I.e. every redeployment of II will add a new entry for the same dapp as the salt in the canister will have changed.

src/frontend/src/storage/index.test.ts
nmattia: Yeah, that's not incorrect. Some of this is tested by the e2e tests but I'm happy to add some tests.

src/frontend/src/storage/index.ts
nmattia: That's true! and we do that for known anchors. I'll prune this too, somehow. Will need to add a timestamp too.

PR #2112, opened 2023-12-06
Add VerifiedAdult VC to the demo issuer

przydatek:
Add to the demo issuer support for issuing `VerifiedAdult` credential.

### DIFF
--- a/demos/vc_issuer/src/main.rs
+++ b/demos/vc_issuer/src/main.rs
@@ -30,7 +30,7 @@ use vc_util::{
     did_for_principal, get_verified_id_alias_from_jws, vc_jwt_to_jws, vc_signing_input,
     vc_signing_input_hash, AliasTuple,
 };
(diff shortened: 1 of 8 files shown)


demos/vc_issuer/src/main.rs
```diff
@@ -67,6 +68,7 @@ thread_local! {
     static SIGNATURES : RefCell<SignatureMap> = RefCell::new(SignatureMap::default());
     static EMPLOYEES : RefCell<HashSet<Principal>> = RefCell::new(HashSet::new());
     static GRADUATES : RefCell<HashSet<Principal>> = RefCell::new(HashSet::new());
+    static ADULTS : RefCell<HashSet<Principal>> = RefCell::new(HashSet::new());
```
frederikrothenberger, line 71: What about making this a bit more dynamic and adding the age specifically.
So that then the credential would be `VerifiedAge` and the parameter would just be checked against the stored age. I think that would be quite neat. 😉

PR #2113, opened 2023-12-06
Update vc_issuer and test-app to use asset_util

frederikrothenberger:
This PR refactors both the vc_issuer as well as the test-app to make
use of the asset_util crate.

The asset_util crate is extended to allow updating assets after the
initial certification.

Includes a minor refactoring of asset_util which makes the
`add_certification_vX` part of the `CertifiedAssets` impl.

This PR is step **4** of 4:
1. move code from `assets.rs` to the crate
2. move certification header code from `http.rs` to the crate
3. refactor the crate a bit to make the public interface nicer
4. use the crate in the other canisters


demos/test-app/lib.rs
```diff
@@ -106,160 +74,108 @@ pub enum AlternativeOriginsMode {
...
-                    // needs to be certified content for the service worker
-                    // (which the browser will then ignore and redirect anyway)
-                    headers.push(certificate_header);
-                    headers.push(("Location".to_string(), location));
-                    status_code = 302;
+                UncertifiedContent => {
+                    // drop the IC-Certificate and IC-Certificate-Expr headers
+                    certified_response
+                        .headers
+                        .retain(|(header_name, _)| !header_name.to_lowercase().starts_with("ic-"));
```
przydatek, line 90: nit: maybe this filter is a bit too generous?  How about changing to `.starts_with("ic-certificate")`?

demos/test-app/lib.rs
```diff
@@ -106,160 +74,108 @@ pub enum AlternativeOriginsMode {
...
-        // the string we are replacing here is inserted by vite during the front-end build
-        let index_html = index_html.replace(
-            r#"<script type="module" crossorigin src="/index.js"></script>"#,
-            &format!(r#"<script data-canister-id="{canister_id}" type="module" crossorigin src="/index.js"></script>"#).to_string()
-        );
-        index_html
-    };
+    // convenience asset to have an url to point to when testing with the redirect alternative origins behaviour
+    assets.push(Asset {
+        url_path: "/.well-known/evil-alternative-origins".to_string(),
```
przydatek, line 166: nit: maybe use a constant, say `EVIL_ALTERNATIVE_ORIGINS_PATH` -- currently it is not obvious that this value is different from  `ALTERNATIVE_ORIGINS_PATH` used in line 158.

PR #2116, opened 2023-12-07
Add slack message on bad release ref

nmattia:
This adds a slack notification to the team in case an error when fetching the release ref. Some failures occurred already which were silently ignored:

https://github.com/dfinity/internet-identity/actions/runs/7068859771/job/19243912571


frederikrothenberger (approved):
LGTM, thanks! Do you know what causes a release ref to be bad? I.e. what the issue with the failed job was?

nmattia:
> Do you know what causes a release ref to be bad?

As far as I can tell it's just an issue on GitHub's API side

PR #2118, opened 2023-12-07
Add a CSS comment to the c-list__item--pin class

meodai:
Make sure the CSS in that line is clear.


src/frontend/src/styles/main.css
```diff
@@ -1772,6 +1772,12 @@ by all browsers (FF is missing) */
   gap: var(--rs-input-inline);
 }
 
+/*
+* The width of the digits is the result of the gap we want between them and the
+* number of digits we want to display. 
+* 100% is the full width of the container, so we devide it by the number of digits
```
nmattia, line 1778:
```suggestion
* 100% is the full width of the container, so we divide it by the number of digits
```

nmattia (approved):
Thanks!

PR #2119, opened 2023-12-07
Make the min-width a bit smaller for mobile

meodai:
To avoid overflow issues on smaller phones 


meodai:
@nmattia Doing this will make the screens tighter on all smaller devices.

nmattia:
@meodai can't we keep the current size but limit the text width for that one problematic screen? sacrificing real estate on all screens just for one loader seems a bit odd

meodai:
@nmattia I don't see how we are sacrificing real-estate. Min-width is only the minimum size the container should have, if it has more content it gets wider. The text-width is not the issue on that particular screen.

nmattia (approved):
Ok I think I misunderstood the issue! Found it odd that all screens had to be impacted. Rubber stamp

lmuntaner:
Outdated

PR #2121, opened 2023-12-08
Add support for PIN identities in VC

nmattia:
This abstracts the passkey use flow so that it can be used both in the authenticateBox (where it was already used) and in the VC flow.

The tests are updated so that their number doesn't grow exponentially:

* One test with legacy RP
* One test with legacy ISS
* One test with PIN

By default, the tests run with non-legacy domains and run with webauthn passkey.


src/frontend/src/components/authenticateBox.ts
```diff
@@ -649,3 +588,120 @@ const pinIdentityToDerPubkey = async (
     pinIdentity.publicKey
   )) as DerEncodedPublicKey;
 };
+
+// Find and use a passkey, whether PIN or webauthn
+const usePasskeyFlow = async <T, I>({
```
nmattia, line 593: This was just moved

frederikrothenberger, line 593: I'm confused by the  naming and the comment. Passkey refers to WebAuthn only, so having this method cover PIN as well is unintuitive for me.

Maybe `authnFlow`?

nmattia: > Passkey refers to WebAuthn only

No, we call our "devices" passkeys too...

> Maybe authnFlow?

We have a few of these already (authenticate, authenticateBox, etc), I wanted to be more specific. `useIdentity`? `useDevice`?

frederikrothenberger: >> Passkey refers to WebAuthn only
>
> No, we call our "devices" passkeys too...

We don't call the PIN auth passkey, we call that temp key 🙃 

Either way `useIdentity` is fine for me 👍

nmattia: Done!

PR #2123, opened 2023-12-08
Downgrade rust to 1.73.0

nmattia:
We're hitting reproducibility issues on 1.74.0, either due to rust itself or the way it is provisioned (GHA, rustup, etc).

Downgrading to 1.73.0 makes the issue disappear. This unblocks us while a better solution is found.

The vc_issuer is updated to compile on 1.73.0.


frederikrothenberger (approved):
Thanks a lot!

PR #2126, opened 2023-12-08
Use fixed length for principal digest

nmattia:
This updates the way we compute principals to encode the length of the origins & principals in a fixed-length manner. This avoids a whole class of collisions that could be crafted intentionally when using the string representation of the numbers.

### DIFF
--- a/src/frontend/src/storage/index.ts
+++ b/src/frontend/src/storage/index.ts
@@ -292,7 +292,7 @@ const computePrincipalDigest = async ({
   const principal = enc.encode(principal_);
   const principalLen = Uint8Array.from([principal_.length]);
 
-  // Create a buffer with all for elements
+  // Create a buffer with all four elements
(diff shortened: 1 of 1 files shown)


src/frontend/src/storage/index.ts
```diff
@@ -274,27 +274,49 @@ const nowMillis = (): number => {
...
-  );
+  const principal_ = principalObj.toText();
+
+  const origin = enc.encode(origin_);
+  const originLen = Uint8Array.from([origin_.length]);
+
+  const principal = enc.encode(principal_);
+  const principalLen = Uint8Array.from([principal_.length]);
+
+  // Create a buffer with all for elements
```
nmattia, line 295:
```suggestion
  // Create a buffer with all four elements
```

frederikrothenberger (approved):

PR #2132, opened 2023-12-12
Randomize id alias for attribute exchange

frederikrothenberger:
This PR introduces randomized id aliases. In order to achieve this, the context handed out by `prepare_id_alias` has to be changed to already include the JWS header so that the `get_id_alias` call knows about the random seed.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -6,6 +6,7 @@ use crate::storage::anchor::Anchor;
 use candid::{candid_method, Principal};
 use canister_sig_util::signature_map::LABEL_SIG;
 use ic_cdk::api::{caller, set_certified_data, trap};
+use ic_cdk::call;
 use ic_cdk_macros::{init, post_upgrade, pre_upgrade, query, update};
 use internet_identity_interface::archive::types::{BufferedEntry, Operation};
 use internet_identity_interface::http_gateway::{HttpRequest, HttpResponse};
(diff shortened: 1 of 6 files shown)


src/internet_identity/src/main.rs
```diff
@@ -418,6 +419,21 @@ fn update_root_hash() {
     })
 }
 
+/// Calls raw rand to retrieve a random salt (32 bytes).
+async fn random_salt() -> Salt {
+    let res: Vec<u8> = match call(Principal::management_canister(), "raw_rand", ()).await {
```
frederikrothenberger, line 424: This was just extracted otherwise we would have it in 3 places.

src/internet_identity/src/vc_mvp.rs
```diff
@@ -135,10 +127,42 @@ pub fn get_id_alias(
...
+        .expect("internal: failed getting protected header");
+    let canister_sig_pk_raw = get_canister_sig_pk_raw(header)
+        .map_err(|e| format!("internal: failed getting canister_sig_pk_raw: {:?}", e))?;
+    CanisterSigPublicKey::try_from_raw(&canister_sig_pk_raw)
+        .map_err(|e| format!("internal: failed parsing canister_sig_pk: {}", e))
+}
+
+/// Constructs and returns a JWS (a signed JWT) from the given components.
+fn add_sig_to_signing_input(signing_input: &str, sig: &[u8]) -> Result<String, String> {
+    let decoder = Decoder::new();
```
frederikrothenberger, line 148: I can still change it to just a '.' concatenated with `sig` if you want. 😉

PR #2133, opened 2023-12-12
Extend test app and demo issuer front-ends with "is adult" credential

frederikrothenberger:
This PR adds front-end support for the demo applications to go through the attribute sharing flow with an "Is adult person" credential.


przydatek (approved):
LGTM, thanks!

PR #2134, opened 2023-12-12
Factor out get_signature and add it to canister_sig_util

przydatek:
Factor out `get_signature`, which is used in canisters that create canister signatures, and add the corresponding functionality to `canister_sig_util`-crate (as function `get_signature_as_cbor`).

### DIFF
--- a/src/canister_sig_util/src/lib.rs
+++ b/src/canister_sig_util/src/lib.rs
@@ -1,5 +1,11 @@
 use candid::Principal;
+use ic_cdk::api::data_certificate;
+use ic_certification::{fork, labeled, pruned, Hash, HashTree};
 use lazy_static::lazy_static;
+use serde::Serialize;
+use serde_bytes::ByteBuf;
+use sha2::{Digest, Sha256};
+use signature_map::{SignatureMap, LABEL_SIG};
 
 pub mod signature_map;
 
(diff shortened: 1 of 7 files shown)


src/canister_sig_util/src/lib.rs
```diff
@@ -125,6 +131,63 @@ pub fn extract_raw_canister_sig_pk_from_der(pk_der: &[u8]) -> Result<Vec<u8>, St
...
+) -> Result<Vec<u8>, String> {
+    let certificate = data_certificate()
+        .ok_or("data certificate is only available in query calls".to_string())?;
+    let witness = sig_map
+        .witness(hash_bytes(seed), message_hash)
+        .ok_or("missing witness".to_string())?;
+
+    let witness_hash = witness.digest();
+    let root_hash = sig_map.root_hash();
+    if witness_hash != root_hash {
```
frederikrothenberger, line 166: Does this check (still) make sense? We rehash the pruned witness tree every time, but if did not match the (unpruned) signature map root hash we would simply have a bug in our implementation. I wonder if this is really best addressed by wasting resources in production, or if we should rather have better tests (unless we're already sufficiently covered).
Honestly, I think we should remove it here.

src/canister_sig_util/src/lib.rs
```diff
@@ -125,6 +131,63 @@ pub fn extract_raw_canister_sig_pk_from_der(pk_der: &[u8]) -> Result<Vec<u8>, St
...
+        return Err(format!(
+            "internal error: signature map computed an invalid hash tree, witness hash is {}, root hash is {}",
+            hex::encode(witness_hash),
```
frederikrothenberger, line 169: You could also get rid of the `hex` dependency if you didn't have this check 😉

frederikrothenberger (approved):
LGTM, thanks!

src/canister_sig_util/src/lib.rs
przydatek: Filed https://dfinity.atlassian.net/browse/GIX-2178

przydatek (commented):
Thanks, I've filed a ticket for the potential code simplification.

PR #2138, opened 2023-12-14
Approve information Sharing CSS

meodai:
Screenshot 2023-12-14 at 18 22 29

- Replicate the VC screen as well as possible using existing components 
- Add the concept of ".t-formatted" for reformatted HTML text. In this case Markdown
- Slightly modify the dapps Explorer component to be useful for origins 
- Add a `.c-card--consent` class


meodai:
@frederikrothenberger I don't understand why the usePin workflow is failing :(

frederikrothenberger:
> @frederikrothenberger I don't understand why the usePin workflow is failing :(

It seems to have been a viewport issue on mobile sized chrome. I have pushed a fix to this PR.

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
```diff
@@ -8,12 +8,33 @@ import { html, TemplateResult } from "lit-html";
...
+import { getDapps, KnownDapp } from "../dappsExplorer/dapps";
+import { dappTemplate } from "../dappsExplorer/index";
+
 import DOMPurify from "dompurify";
 
 import copyJson from "./allowCredentials.json";
 
 /* A screen prompting the user to allow (or cancel) issuing verified
  * credentials */
+const getOrigin = (
```
frederikrothenberger, line 20: The comment above should be move to the template below, no?

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
```diff
@@ -8,12 +8,33 @@ import { html, TemplateResult } from "lit-html";
...
+
+  if (!foundDapp) {
+    foundDapp = new KnownDapp({
+      name: "Unknown Dapp",
+      website: origin,
+      logo: "unwnowndapp.png",
+      oneLiner: oneLiner,
+    });
+  }
+  foundDapp.oneLiner = oneLiner;
```
frederikrothenberger, line 35: Nit: I think it is slightly hacky to override the `oneLiner` with the `copy` string on the `KnownDapp`.
What about overriding the value in the call to `dappsTemplate`?

frederikrothenberger (approved):
Two minor comments, otherwise looks good to me.

Disclaimer, I did not look at the CSS.

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
meodai: yes of course! Let me fix that

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
meodai: @frederikrothenberger you are right, I think its better now. let me know if this works for you

PR #2143, opened 2023-12-15
Strip device metadata from lookup

frederikrothenberger:
This PR removes the `metadata` field from the `lookup` endpoint.


przydatek (approved):
LGTM, thanks!

PR #2147, opened 2023-12-18
Update readme with local URLs

frederikrothenberger:
The recent refactoring with multiple JS bundles has lead to II breaking in Chrome, when the canister id is supplied as an URL parameter.
There is no easy or nice solution, so for now we just advise developers to use a different URL scheme when using anything but Safari.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Use the Internet Identity canister in your local dfx project by adding the follo
 ```
 To deploy, run `dfx deploy`.
 
-To access Internet Identity use one of the following URLs:
+To access Internet Identity or configure it for your dapp, use one of the following URLs:
 * Chrome, Firefox: `http://<canister_id>.localhost:4943`
 * Safari: `http://localhost:4943?canisterId=<canister_id>`
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,11 @@ Use the Internet Identity canister in your local dfx project by adding the follo
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -62,6 +62,11 @@ Use the Internet Identity canister in your local dfx project by adding the follo
   }
 }
 ```
+To deploy, run `dfx deploy`.
+
+To access Internet Identity use one of the following URLs:
```
peterpeterparker, line 67: Maybe following to express the fact that it's also the URL to configure for auth-client?

To access Internet Identity or configure it for your dapp, you can use one of the following URLs:

demos/using-dev-build/README.md
```diff
@@ -128,7 +128,9 @@ You might get different canister IDs (and that's totally fine). If the `webapp`
 
 ![](./webapp.png)
 
-_If you actually use the webapp, make sure that the "Internet Identity URL" field points to `http://localhost:4943/?canisterId=<canister ID of the internet_identity canister>`._
+_If you actually use the webapp, make sure that the "Internet Identity URL" field points to one of these URLS_:
+* Chrome and FirefoxSafari `http://<canister ID of the internet_identity canister>.localhost:4943/`.
```
peterpeterparker, line 132: Typo "Chrome and FirefoxSafari"

peterpeterparker (approved):
LGTM, minor comments. Thanks!

PR #2148, opened 2023-12-18
Remove useless Result aliases

frederikrothenberger:
This PR removes a few types that could be replaced by `std::Result`.


przydatek (approved):
Please update also `vc_spec.md`, otherwise LGTM.  Thanks!

PR #2150, opened 2023-12-19
Move unknown app icon to other assets

meodai:


src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
```diff
@@ -26,9 +26,10 @@ const getOrigin = (origin: string, dapplist: KnownDapp[]): KnownDapp => {
     foundDapp = new KnownDapp({
       name: origin,
       website: origin,
-      logo: "unwnowndapp.png",
+      logo: "../unwnowndapp.png",
```
frederikrothenberger, line 29: Could we take this opportunity to rename the file to `unknowndapp.png`?

peterpeterparker, line 29: Randomly noticed this PR, can't you import the image given it finds place in src folder?

e.g. with vite / SvelteKit

```
import eth from '$lib/assets/eth.svg';

<img src={eth} />
```

meodai: @peterpeterparker I got a bitmap, but I will get an SVG soon from Artem and change it them. Hope that works.

frederikrothenberger (approved):
Thanks!

src/frontend/src/flows/verifiableCredentials/allowCredentials.ts
peterpeterparker: I don't mean the extension, I mean using an `import`. If you use an import, your editor will be able to resolve it for you if you ever moved again the file in the future. Furthermore, it's then easier to find where the image is used.

But not sure it is supported in II.

PR #2151, opened 2023-12-19
API V2 Refactor

frederikrothenberger:
This refactors the existing API v2 to be consistent with new patterns introduced in the issuer API:
* Use camel case variants
* Use `Result` return values
* Remove `opt`

The `Err` variants will be extended in the future with more types. These will be breaking changes. However, given we mark the API v2 as experimental this should be ok.

In order to have CI pass, we have an exception in the form of a patch file (`allowed_breaking_change.patch`) that undoes the breaking changes that we intentionally made. If we introduce additional breaking changes, the patch will no longer apply and thus make the check fail.

### DIFF
--- a/src/internet_identity/tests/integration/v2_api/identity_register.rs
+++ b/src/internet_identity/tests/integration/v2_api/identity_register.rs
@@ -47,7 +47,9 @@ fn should_not_exceed_configured_identity_range() {
     create_identity_with_authn_method(&env, canister_id, &authn_method);
     create_identity_with_authn_method(&env, canister_id, &authn_method);
 
-    let challenge = api_v2::captcha_create(&env, canister_id).unwrap().unwrap();
+    let challenge = api_v2::captcha_create(&env, canister_id)
+        .expect("API call failed")
+        .expect("captcha_create failed");
 
     let result = api_v2::identity_register(
         &env,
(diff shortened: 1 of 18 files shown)


src/internet_identity/tests/integration/v2_api/identity_register.rs
```diff
@@ -48,36 +47,29 @@ fn should_not_exceed_configured_identity_range() {
...
+        canister_id,
+        authn_method.principal(),
+        &authn_method,
+        &ChallengeAttempt {
+            chars: "a".to_string(),
+            key: challenge.challenge_key,
+        },
+        None,
+    )
+    .unwrap();
```
przydatek, line 63: nit: there is a bunch of new `unwrap()`s in the tests. I think it is generally better to use `expect()`s and provide some context what has failed, as otherwise one has no idea from the log, especially if we have a chain `unwrap().unwrap()`.  Maybe add some context in some cases, where it would make sense?

przydatek (approved):
LGTM, thanks!

src/internet_identity/tests/integration/v2_api/identity_register.rs
frederikrothenberger: I replaced the `unwrap` calls now.

PR #2152, opened 2023-12-20
Add API v2 test helpers

frederikrothenberger:
This PR adds more useful API v2 test helpers in preparation for more APIv2 methods.

In addition, the `authn_method_add` test is refactored to make use of those new helpers.


przydatek (approved):
LGTM, thanks!

PR #2153, opened 2023-12-20
Remove dependency of `vc_util` on `ic_cdk`

przydatek:
The dependency on `ic_cdk` implies that `vc_util` cannot be compiled into WASM that would work in a browser context (where ic0-API is not available).  As `vc_util` was using `ic_cdk` only for printing debug info, there is no reason to really keep this dependency.

### DIFF
--- a/src/vc_util/src/lib.rs
+++ b/src/vc_util/src/lib.rs
@@ -329,9 +329,7 @@ fn validate_claim<T: PartialEq<S> + std::fmt::Display, S: std::fmt::Display>(
         } else {
             println!(
                 "inconsistent claim [{}] in VC::  expected: {}, actual: {}",
-                label,
-                expected,
-                actual
+                label, expected, actual
             );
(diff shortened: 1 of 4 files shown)


src/vc_util/src/lib.rs
```diff
@@ -327,16 +327,14 @@ fn validate_claim<T: PartialEq<S> + std::fmt::Display, S: std::fmt::Display>(
         if expected == actual {
             Ok(())
         } else {
-            ic_cdk::println!(
+            println!(
```
frederikrothenberger, line 330: Do you want to keep this in? Usually, it's not expected for a library to print to stdout...

przydatek: I'll remove these once I can add custom error messages to returned errors (which currently require `static` strings).
Added https://dfinity.atlassian.net/browse/GIX-2190 to address this issue.

PR #2154, opened 2023-12-20
Introduce API v2 identity_authn_info method 

frederikrothenberger:
This PR introduces the `identity_authn_info` query call to retrieve
the necessary information for authentication.

It also adds test to check that last device can be removed (was not possible before due to the lack of options to query the number of authn_methods without authentication).


przydatek (approved):
LGTM, thanks!

PR #2155, opened 2023-12-20
Refactor authenticated OP helper to allow returning proper results

frederikrothenberger:
This is a refactoring in preparation for API v2 not panicking anymore. It should apply to both, errors in the `authenticated_anchor_operation` wrapper related to authentication and bookkeeping as well as errors in the actual operation.


przydatek (approved):
LGTM, thanks!

PR #2156, opened 2023-12-21
Consistently use camel case variants in APIv2

frederikrothenberger:
This refactors the API v2 to use camel case in all remaining variants:
* metadata map
* purpose

### DIFF
--- a/src/internet_identity_interface/src/internet_identity/conversions.rs
+++ b/src/internet_identity_interface/src/internet_identity/conversions.rs
@@ -71,6 +71,52 @@ impl From<AuthnMethodProtection> for DeviceProtection {
     }
 }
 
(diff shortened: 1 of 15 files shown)


src/internet_identity_interface/src/internet_identity/conversions.rs
```diff
@@ -71,6 +71,52 @@ impl From<AuthnMethodProtection> for DeviceProtection {
     }
 }
 
+impl From<AuthnMethodPurpose> for Purpose {
```
przydatek, line 74: nit: maybe add simple tests for these conversions?  They are trivial, but it would be good to have them executed regularly in some tests.

przydatek (approved):
LGTM, thanks!

PR #2157, opened 2023-12-21
Introduce v2 API for tentative authn method registration

frederikrothenberger:
This PR introduces the equivalent of the tentative device registration
in the existing API.


przydatek (approved):
LGTM, thanks!
Nice tests, btw!

PR #2158, opened 2023-12-21
Change AuthenticationFailed to Unauthorized

frederikrothenberger:
The `AuthenticationFailed` was misleading, since the issue is not with authentication (i.e. the `caller` is known), but with authorization (i.e. the `caller` is not allowed to do the specific action).


przydatek (approved):
LGTM, thanks!

PR #2159, opened 2023-12-21
Introduce API v2 authn_method_metadata_replace

frederikrothenberger:
This PR introduces the API v2 `authn_method_metadata_replace` method.
It replaces the metadata map of an existing `authn_method` of an identity.
It will be used in place of the current `update` method. The scope reduction
to just `metadata` is there to have a clear separation between security
relevant changes to an authn_method (such as modifying the `protection`) and
changes not relevant to security (modifying the `alias` entry in the metadata).

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -524,6 +524,40 @@ fn check_authentication(anchor_number: AnchorNumber) -> Result<(Anchor, DeviceKe
 mod v2_api {
     use super::*;
 
+    #[query]
+    #[candid_method(query)]
+    fn identity_authn_info(identity_number: IdentityNumber) -> Result<IdentityAuthnInfo, ()> {
+        let anchor =
+            state::storage_borrow(|storage| storage.read(identity_number).unwrap_or_default());
+
+        let authn_info = anchor.into_devices().into_iter().fold(
+            IdentityAuthnInfo {
+                authn_methods: vec![],
+                recovery_authn_methods: vec![],
+            },
+            |mut authn_info, device| {
+                let purpose = device.purpose;
+
+                let authn_method = if let Some(credential_id) = device.credential_id {
+                    AuthnMethod::WebAuthn(WebAuthn {
+                        credential_id,
+                        pubkey: device.pubkey,
+                    })
+                } else {
+                    AuthnMethod::PubKey(PublicKeyAuthn {
+                        pubkey: device.pubkey,
+                    })
+                };
+                match purpose {
+                    Purpose::Authentication => authn_info.authn_methods.push(authn_method),
+                    Purpose::Recovery => authn_info.recovery_authn_methods.push(authn_method),
+                }
+                authn_info
+            },
+        );
+        Ok(authn_info)
+    }
+
     #[update]
     #[candid_method]
     async fn captcha_create() -> Result<Challenge, ()> {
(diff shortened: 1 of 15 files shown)


src/internet_identity/src/main.rs
```diff
@@ -631,6 +631,33 @@ mod v2_api {
         Ok(())
     }
 
+    #[update]
+    #[candid_method]
+    fn authn_method_metadata_replace(
+        identity_number: IdentityNumber,
+        public_key: PublicKey,
```
przydatek, line 638: maybe rename `public_key` to `authn_method_pk` or something similar, so that is clear what is the meaning of the provided public key?

src/internet_identity/src/main.rs
```diff
@@ -631,6 +631,33 @@ mod v2_api {
...
+        metadata: HashMap<String, MetadataEntryV2>,
```
przydatek, line 639: nit: `metadata` -> `new_metadata`?

src/internet_identity/tests/integration/v2_api/authn_method_metadata.rs
```diff
@@ -0,0 +1,201 @@
...
+        identity_number,
+        &authn_method.public_key(),
+        &metadata,
+    );
+    expect_user_error_with_message(
+        result,
+        CanisterCalledTrap,
+        Regex::new("failed to modify device: Cumulative size of variable sized fields exceeds limit: length \\d+, limit \\d+\\.").unwrap(),
+    );
+    Ok(())
```
przydatek, line 125: maybe add a check that the metadata was not changed/set during the failed call?

src/internet_identity/tests/integration/v2_api/authn_method_metadata.rs
```diff
@@ -0,0 +1,201 @@
...
+use ic_test_state_machine_client::ErrorCode::CanisterCalledTrap;
+use internet_identity_interface::internet_identity::types::{
+    AuthnMethodMetadataReplaceError, MetadataEntryV2,
+};
+use regex::Regex;
+use serde_bytes::ByteBuf;
+use std::collections::HashMap;
+
+#[test]
+fn should_write_authn_method_metadata() -> Result<(), CallError> {
```
przydatek, line 19: maybe add a variant when there is some metadata initially, and the call to `authn_method_metadata_replace` really replaces it?

src/internet_identity/tests/integration/v2_api/authn_method_metadata.rs
```diff
@@ -0,0 +1,201 @@
...
+    let identity_number = create_identity_with_authn_method(&env, canister_id, &authn_method);
+
+    let identity_info =
+        api_v2::identity_info(&env, canister_id, authn_method.principal(), identity_number)?
+            .expect("identity info failed");
+    assert!(identity_info.metadata.is_empty());
+
+    let metadata = HashMap::from([(
+        METADATA_KEY.to_string(),
+        MetadataEntryV2::String("a".repeat(3000)),
```
przydatek, line 109: maybe use a value closer to `VARIABLE_FIELDS_LIMIT` (and mention the constant name in a comment)? or export the constant and use it also in the test?

przydatek (approved):
LGTM, modulo minor suggestions.
Thanks!

src/internet_identity/tests/integration/v2_api/authn_method_metadata.rs
frederikrothenberger: Given that we want to move away from `panic!` this is very sensible, yes.

PR #2160, opened 2023-12-22
Introduce API v2 authn_method_replace

frederikrothenberger:
This PR adds the `authn_method_replace` method as a replacement for the
legacy API `replace` method.


przydatek (approved):
LGTM, thanks!.

PR #2161, opened 2023-12-22
Refactor AuthnMethodData to group security relevant, modifiable fields

frederikrothenberger:
This PR refactors `AuthnMethodData` to group together security relevant,
modifiable fields. It allows to later introduce a method to update
these fields specifically which will require additional security
measures compared to modifying the `metadata`.


przydatek (approved):
LGTM, thanks!

PR #2168, opened 2024-01-04
Use correct node version in clean build

nmattia:
The clean build was using whatever node version was in the environment. This updates the `check-build` action to use the `setup-node` action, so that it uses the expected node version.

A version check is added to the build script. On CI, this aborts the build. Not on CI, this issues a warning.

The clean build reproducibility checks are also re-enabled. The reproducibility issues were caused by differences in the compressed assets. It looks like node statically links zlib, and it looks like the zlib version from the GHA environment's node differs from the one in our docker setup (set through `.node-version`).

### DIFF
--- a/.github/actions/check-build/action.yml
+++ b/.github/actions/check-build/action.yml
@@ -12,6 +12,7 @@ runs:
   using: "composite"
   steps:
     - uses: ./.github/actions/bootstrap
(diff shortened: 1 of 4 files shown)


frederikrothenberger (approved):
LGTM, thanks!

PR #2172, opened 2024-01-05
Update node version

nmattia:
This bumps the node version to 18.19.0, which slightly breaks ts-node.

In order to make the using-dev-build example work again with ts-node, we port the node loader options we use in the main II `package.json`.


peterpeterparker:
If we already bump node, can we bump straight to LTS (20.x)? That would be cool.

nmattia:
> can we bump straight to LTS (20.x)?

Unfortunately right now other stuff breaks there, but I'll get on it asap.

peterpeterparker (approved):
Schade ;)

Thanks, LGTM.

PR #2173, opened 2024-01-05
Replace jose with custom Base64UrlEncode

nmattia:
This drops the `jose` dependency in the main II codebase to instead use the browser's `btoa` base64 support with a few tweaks to URL encode in a JWT-friendly way.

### DIFF
--- a/src/frontend/src/flows/verifiableCredentials/index.ts
+++ b/src/frontend/src/flows/verifiableCredentials/index.ts
@@ -17,7 +17,6 @@ import {
   ECDSAKeyIdentity,
 } from "@dfinity/identity";
 import { isNullish, nonNullish } from "@dfinity/utils";
-import { base64url } from "jose";
 import { abortedCredentials } from "./abortedCredentials";
 import { allowCredentials } from "./allowCredentials";
 import { VcVerifiablePresentation, vcProtocol } from "./postMessageInterface";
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -374,6 +373,18 @@ const authenticateForIssuer = async ({
...
 
+// Perform a "base64url" encoding, i.e. a URL-friendly variation of base64 encoding
+const base64UrlEncode = (x: unknown): string => {
+  const json = JSON.stringify(x);
+  // Pretend the json is binary and use btoa (binary-to-ascii as base64) to base64 encode
+  const b64 = btoa(json);
+  // make it URL friendly:
+  // '=': used as padding, just remove
+  // '/': Base64Url as per jwt.io's playgrond replaces it with '_'
+  // '+': Base64Url as per jwt.io's playgrond replaces it with '-'
```
nmattia, line 384:
```suggestion
  // '/': Base64Url as per jwt.io's playground replaces it with '_'
  // '+': Base64Url as per jwt.io's playground replaces it with '-'
```

src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -374,6 +373,18 @@ const authenticateForIssuer = async ({
...
+  return b64.replace(/=+$/, "").replace("/", "_").replace("+", "-");
```
peterpeterparker, line 385: out of curiosity, why those replace?

I use a similar encoding In Juno, e.g. to generates token to execute the Cli in a GitHub action, and I don't do those replace. So wondering why and if I do it as well.

I just encode indentity in objects.

```
const key = Ed25519KeyIdentity.generate();
identity = btoa(JSON.stringify({ token: key.toJSON() }));
```

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/flows/verifiableCredentials/index.ts
nmattia: It's just the jwt spec that that asks that they are base64**URL** encoded, not just base64. It means they are URL-friendly, i.e. don't include `=`, `/` or `+`:

```
> btoa(JSON.stringify({hello: "world????", foo:"bara"}))
'eyJoZWxsbyI6IndvcmxkPz8/PyIsImZvbyI6ImJhcmEifQ=='
```

depending on what those tokens you generate you may or may not need it

peterpeterparker: Gotcha, thanks!

PR #2177, opened 2024-01-08
Extend VC RP spec to support alternative frontend origins.

przydatek:
Add to VC's RP-request an optional argument to support [alternative frontend origins](https://internetcomputer.org/docs/current/references/ii-spec#alternative-frontend-origins)-feature.


frederikrothenberger (approved):
LGTM

PR #2178, opened 2024-01-08
Bump node to v20 LTS

nmattia:
This bumps the node version we use to the latest LTS.

A couple of workarounds necessary for node <20 are removed.


peterpeterparker (approved):
LGTM, thanks

PR #2180, opened 2024-01-08
Add lit-analyzer check

nmattia:
This adds a lit-analyzer lint check. This ensures we make correct use of lit templates.

We disable three rules which currently appear to be overrestrictive:
* https://github.com/runem/lit-analyzer/issues/341
* https://github.com/runem/lit-analyzer/issues/342
* https://github.com/runem/lit-analyzer/issues/343

### DIFF
--- a/src/frontend/src/flows/dappsExplorer/dapps.ts
+++ b/src/frontend/src/flows/dappsExplorer/dapps.ts
@@ -43,7 +43,7 @@ export class KnownDapp {
     return this.descr.oneLiner;
   }
 
-  public get website(): string | undefined {
+  public get website(): string {
     return this.descr.website;
   }
 }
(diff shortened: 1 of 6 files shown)


src/frontend/src/flows/dappsExplorer/dapps.ts
```diff
@@ -43,7 +43,7 @@ export class KnownDapp {
     return this.descr.oneLiner;
   }
 
-  public get website(): string | undefined {
+  public get website(): string {
```
nmattia, line 46: lit-analyzer noticed that an `href` was being set to a potentially undefined string which came from here. Turns out the website cannot be undefined.

peterpeterparker (approved):
LGTM, thx

PR #2181, opened 2024-01-08
Remove unnecessary process dependency

nmattia:
The webapp depends on the `process` lib which is not actually used.


peterpeterparker (approved):
YEAH!

PR #2190, opened 2024-01-10
Fix test_app asset update

frederikrothenberger:
This PR updates the asset_util crate to also take shared headers into account and remove outdated subtrees from
the certification on asset updates.
In addition, a missing certified data update was added in the test app code.

These changes are made in preparation for the upgrade to dfx 0.15.3 that now checks asset certification (and thus surfaced this bug in the test_app).


przydatek (approved):
LGTM, thanks!

PR #2192, opened 2024-01-11
Add support for different status codes to asset lib

frederikrothenberger:
This PR adds support for per-asset status codes in the asset lib, which is necessary to support certified redirects. We will make use of that in the test_app in an upcoming PR.

### DIFF
--- a/src/asset_util/src/lib.rs
+++ b/src/asset_util/src/lib.rs
@@ -29,13 +29,16 @@ pub const IC_CERTIFICATE_EXPRESSION: &str =
 /// for the certification to be valid.
 #[derive(Debug, Default, Clone)]
 pub struct CertifiedAssets {
-    assets: HashMap<String, (Vec<HeaderField>, Vec<u8>)>,
+    /// Map from URL path to tuple of response (status_code, headers, body)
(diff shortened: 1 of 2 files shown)


src/asset_util/src/lib.rs
```diff
@@ -29,13 +29,16 @@ pub const IC_CERTIFICATE_EXPRESSION: &str =
 /// for the certification to be valid.
 #[derive(Debug, Default, Clone)]
 pub struct CertifiedAssets {
-    assets: HashMap<String, (Vec<HeaderField>, Vec<u8>)>,
+    /// Map from URL path to tuple of response (status_code, headers, body)
+    assets: HashMap<String, (u16, Vec<HeaderField>, Vec<u8>)>,
```
przydatek, line 33: nit: Why don't we use `HashMap<string, CertifiedAssed` here?

przydatek (approved):
LGTM, thanks!
(nit: the second sentence of the PR description seems broken, should be "We will make use..."?)

src/asset_util/src/lib.rs
frederikrothenberger: Ah yes, of course. 👍 
This now also finally resolves the issue of `CertifiedAssets` not containing `CertifiedAssed`. 😉

PR #2195, opened 2024-01-11
Split vc-api and test-app into workspaces

nmattia:
This extracts the vc issuer's API into a new workspace and changes the test-app from a standalone package to a workspace of the main package.

By extracting the issuer API, the types (like credential request) can be reused elsewhere.

By turning the test-app into a workspace, we deduplicate a lot of the dependency information (by removing one package lock) and we can use local dependencies (like the issuer types).


src/vc-api/src/index.ts
nmattia, line 1: This file was just moved, not changed (beyond the extra export)

src/vc-api/README.md
```diff
@@ -0,0 +1,11 @@
+# VC Issuer API
+
+A small TypeScript library with type & [zod](https://zod.dev) definitions for the [issuer API](../../demos/vc_issuer/README.md).
+
+## Usage
+
+Build from the root directory:
+
+```bash
+$ npm run --workspace ./src/vc-api build
```
peterpeterparker, line 10: it's not `npm run build ...` instead of `npm run .... build`?

btw. you can shorten `--workspace` with `-w` if interesting

src/vc-api/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@dfinity/internet-identity-vc-api",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc --build",
+    "check": "tsc --noEmit",
+    "postinstall": "npm run build"
+  },
+  "license": "SEE LICENSE IN LICENSE.md",
+  "devDependencies": {
```
peterpeterparker, line 10: you need to redeclare those here given that you build from root?

btw. zod and dfinity libs are runtime libs no?

peterpeterparker (approved):
Already approving, if you can have a look at my two comments though before merging and if interesting.

src/vc-api/README.md
nmattia: with `npm run build FOO`, it can be interpreted as "pass `FOO` to the build command", whereas with `npm run FOO build` it's clear that it's the build command that's being run inside the workspace

 > btw. you can shorten --workspace with -w if interesting
 
 I know :) but in scripts and READMEs I like the long form version for clarity

src/vc-api/package.json
nmattia: oh yes that's right! good catch

src/vc-api/README.md
peterpeterparker: > with npm run build FOO, it can be interpreted as "pass FOO to the build command", whereas with npm run FOO build it's clear that it's the build command that's being run inside the workspace

Clear is relative, I never seen `npm run -w ... build` before I have to say so it wasn't clear 😉

peterpeterparker: Thanks for the explanation

nmattia:
@peterpeterparker PTAL!

peterpeterparker:
Already approved, go on!

PR #2199, opened 2024-01-12
Add to get_verified_id_alias_from_jws a check for the expected vc subject

przydatek:
Add to `get_verified_id_alias_from_jws()` an extra argument to specify the expected VC subject, and verifies in the function that the subject from the JWS matches the expected VC subject. The reason for this change is to ensure that this check is made -- otherwise the caller, who should know the expected VC subject, could forget to check the consistency of the JWS with the expectation.

### DIFF
--- a/demos/vc_issuer/src/main.rs
+++ b/demos/vc_issuer/src/main.rs
@@ -148,13 +148,6 @@ fn apply_config(init: IssuerInit) {
 
 fn authorize_vc_request(
     alias: &SignedIdAlias,
-    current_time_ns: u128,
-) -> Result<AliasTuple, IssueCredentialError> {
-    extract_id_alias(alias, &caller(), current_time_ns)
-}
-
-fn extract_id_alias(
-    alias: &SignedIdAlias,
     expected_vc_subject: &Principal,
     current_time_ns: u128,
 ) -> Result<AliasTuple, IssueCredentialError> {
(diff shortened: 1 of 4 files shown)


demos/vc_issuer/src/main.rs
```diff
@@ -150,20 +150,12 @@ fn authorize_vc_request(
...
-
-    if caller() != alias_tuple.id_dapp {
-        return Err(IssueCredentialError::UnauthorizedSubject(format!(
-            "Caller {} does not match id alias dapp principal {}.",
-            caller(),
-            alias_tuple.id_dapp
-        )));
-    }
-    Ok(alias_tuple)
+    extract_id_alias(alias, &caller(), current_time_ns)
```
frederikrothenberger, line 153: Should we also remove this indirection and maybe inline `extract_id_alias`?
It would also be nice to handle `caller()` and `time()` the same way (i.e. either pass both as an argument or call the system API in this function).

src/vc_util/src/lib.rs
```diff
@@ -217,25 +226,12 @@ pub fn verify_ii_presentation_jwt_with_canister_ids(
...
         id_alias_vc_jws.as_str(),
+        &effective_vc_subject,
         &vc_flow_signers.ii_canister_id,
         root_pk_raw,
         current_time_ns,
     )
     .map_err(PresentationVerificationError::InvalidIdAliasCredential)?;
-    // TODO: change `get_verified_id_alias_from_jws` to additionally take `effective_vc_subject`
-    //       as an argument and to perform the check below internally.
-    //       Also, consider adding `ii_origin`-argument to enable custom values for II-origin,
```
frederikrothenberger, line 227: > Also, consider adding `ii_origin`-argument

Are we no longer considering that?

frederikrothenberger (approved):
LGTM, minor comments.

src/vc_util/src/lib.rs
przydatek: I did consider this 😉 , and decided not to do it, the reason being that it is not needed right now, and it is not sure it will be needed. Also, the function has already 5 arguments, so it is getting "crowded".

przydatek (commented):
Thanks for the review!

PR #2202, opened 2024-01-12
Factor out add_signature to canister_sig_util

przydatek:
Factor out `add_signature` to `canister_sig_util`-crate, so that it can be re-used. 
Additionally, hide and clean-up some methods from `SignatureMap`, to have a focused, and more consistent API.


src/canister_sig_util/src/signature_map/test.rs
```diff
@@ -81,23 +81,23 @@ fn test_random_modifications() {
     for round in 1..100 {
         let n_seeds = rng.gen_range(0..5);
         for _i in 0..n_seeds {
-            let mut seed_hash = Hash::default();
-            rng.fill_bytes(&mut seed_hash);
+            let mut seed = Hash::default();
```
frederikrothenberger, line 84: Why did you rename this? It is still a (fake) hash, right?

frederikrothenberger (approved):
LGTM, thanks!

src/canister_sig_util/src/signature_map/test.rs
przydatek: The test still uses a (fake) hash as a seed (and not as a hash of a seed), so the renaming just makes the things consistent.  I've updated the apis, so that it is clear that what we're passing is an (arbitrary) byte array to be used as a seed, and then internally we hash the seed for signature insertion (according to the spec).  Otherwise it was confusing whether the seed has to be a hash value (it doesn't, right?).

przydatek (commented):
Thanks!

PR #2204, opened 2024-01-15
Prune expired signatures in VC code

frederikrothenberger:
So far, signatures are not pruned in the vc issuer nor when adding new signatures for the vc flow. This will eventually lead to memory exhaustion (and might be expensive on cycles).
This PR fixes that by refactoring the library such that prune is called when adding new signatures.


przydatek (approved):
LGTM, thanks!

przydatek (approved):
LGTM, thanks!

PR #2205, opened 2024-01-15
Clean up VC e2e tests

nmattia:
This extracts a couple of helper functions (issuer auth, rp auth, vc flow) so they can be reused (currently both in the issuer employee setup tests & the actual VC tests)


peterpeterparker (approved):
Thx

PR #2211, opened 2024-01-18
Add VC support for alternative origins

nmattia:
This updates the frontend verifiable credentials flow to access a new `derivationOrigin` field in the VC request.

The frontend then queries the specified derivation origin and checks the RP is listed as an alternative origin, as in the authentication flow.

If successful, the specified derivation origin is used as the origin for the VC flow.

The Verifiable Credentials e2e tests are moved to their own directory and split between base flow tests, utils, and new alternative origin tests.

### DIFF
--- a/src/frontend/src/flows/verifiableCredentials/index.ts
+++ b/src/frontend/src/flows/verifiableCredentials/index.ts
@@ -4,6 +4,7 @@ import { withLoader } from "$src/components/loader";
 import { showMessage } from "$src/components/message";
 import { showSpinner } from "$src/components/spinner";
 import { fetchDelegation } from "$src/flows/authorize/fetchDelegation";
+import { validateDerivationOrigin } from "$src/flows/authorize/validateDerivationOrigin";
 import { getAnchorByPrincipal } from "$src/storage";
 import { AuthenticatedConnection, Connection } from "$src/utils/iiConnection";
 import {
@@ -69,8 +70,9 @@ const verifyCredentials = async ({
     credentialSubject: givenP_RP,
     issuer: { origin: issuerOrigin, canisterId: expectedIssuerCanisterId_ },
     credentialSpec,
+    derivationOrigin: rpDerivationOrigin,
   },
-  rpOrigin,
+  rpOrigin: rpOrigin_,
 }: { connection: Connection } & VerifyCredentialsArgs) => {
   // Look up the canister ID from the origin
   const lookedUp = await withLoader(() =>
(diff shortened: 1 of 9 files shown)


src/frontend/src/test-e2e/verifiableCredentials/alternativeOrigins.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const result = await getVCPresentation_({
+      vcTestApp: vcTestAppAlt,
+      browser,
+      authConfig,
+    });
+
+    expect(result.result).toBe("aborted");
+    if (!("reason" in result)) {
+      throw new Error("brwa");
```
nmattia, line 105: let me fix this up

src/frontend/src/test-e2e/verifiableCredentials/utils.ts
```diff
@@ -77,20 +57,33 @@ const registerWithIssuer = async ({
...
 }): Promise<VcTestAppView> => {
   const vcTestApp = new VcTestAppView(browser);
   await vcTestApp.open(relyingParty, II_URL, issuer);
 
+  if (nonNullish(derivationOrigin)) {
+    const demoView = new DemoAppView(browser);
+    await demoView.setDerivationOrigin(derivationOrigin);
+
+    await demoView.updateAlternativeOrigins(
+      `{"alternativeOrigins":["${TEST_APP_NICE_URL}"]}`,
```
nmattia, line 81: This should probably be `relyingParty`

src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -4,6 +4,7 @@ import { withLoader } from "$src/components/loader";
 import { showMessage } from "$src/components/message";
 import { showSpinner } from "$src/components/spinner";
 import { fetchDelegation } from "$src/flows/authorize/fetchDelegation";
+import { validateDerivationOrigin } from "$src/flows/authorize/validateDerivationOrigin";
```
frederikrothenberger, line 7: For a later PR: Given it is now used in multiple flows, this could be moved to some flow independent infrastructure place? Also the documentation in this file does not convey that it is used in multiple flows.

frederikrothenberger (approved):
LGTM, thanks!

src/frontend/src/flows/verifiableCredentials/index.ts
nmattia: Yep definitely! I also want to change the interface:
* the args are unnamed making it easy to swap the origins by mistake
* the validation should not be allowed to be called with an undefined derivationOrigin; handling that case should be the caller's responsibility

it's somewhere on my TODO list, but since we already pull some code from the `authorize` flow elsewhere I thought this wouldn't hurt

PR #2212, opened 2024-01-18
Add dev server logs

nmattia:
This saves the dev server logs as artifacts and cleans up the dev server canister look up & logs a bit.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -482,6 +482,10 @@ jobs:
       # Make sure that one failing test does not cancel all other matrix jobs
       fail-fast: false
 
+    env:
+      # Suffix used for tagging artifacts
+      artifact_suffix: ${{ contains(matrix.domain, 'internetcomputer') && 'current' || 'legacy' }}-${{ matrix.device }}-${{ matrix.shard }}
+
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -482,6 +482,10 @@ jobs:
       # Make sure that one failing test does not cancel all other matrix jobs
       fail-fast: false
 
+    env:
+      # Suffix used for tagging artifacts
+      artifact_suffix: ${{ contains(matrix.domain, 'internetcomputer') && 'current' || 'legacy' }}-${{ matrix.device }}-${{ matrix.shard }}
```
frederikrothenberger, line 487: Given the domain migration has been abandoned I'm not sure `current` and `legacy` are really that clear. Why not simply have the domain be part of the `artifact_suffix`?

frederikrothenberger (approved):
LGTM, thanks!

.github/workflows/canister-tests.yml
nmattia: because it's used as an artifact name, which is quite picky about allowed characters IIRC; `current` and `legacy` were also not too long, plus we use this pretty much everywhere else so I thought it was appropriate

PR #2213, opened 2024-01-19
Fix race condition in devserver

nmattia:
The devserver handles proxying requests to canisters. The previous implementation used a single `http-proxy` Proxy object, where the handlers were rewritten for each new async request.

The caused issues with concurrent request, where a new request might set up new handlers for an in-flight request.

This drops the dependency on `http-proxy` and instead proxies the request using `undici` (existing transivite dep through Vite).

The forwarding code is also moved to the utils module.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1450,14 +1450,6 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/@types/http-proxy": {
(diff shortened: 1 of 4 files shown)


frederikrothenberger (approved):
LGTM, thanks a lot for fixing it!

PR #2215, opened 2024-01-19
Clean up compatibility notice

nmattia:
This updates the compatibility notice to make it more correct and more informative.

In particular, it now only ensures that `window.PublicKeyCredential` exists, which is enough for ensuring webauthn support in most cases. For the other cases, we simply show an error later on during usage.

The compatibility page is also cleaned up a bit to make sure all the text ("Your browser doesn't support II...") is inside the error box, and clarifies the cause of the error (no webauthn support).

### DIFF
--- a/src/frontend/src/flows/compatibilityNotice.ts
+++ b/src/frontend/src/flows/compatibilityNotice.ts
@@ -1,26 +1,31 @@
-import { compatibilityChart } from "$src/components/compatibilityChart";
 import { mainWindow } from "$src/components/mainWindow";
 import { warnBox } from "$src/components/warnBox";
 import { html, render } from "lit-html";
 
 // Taken from: https://caniuse.com/?search=PublicKeyCredential
-const pageContent = (reason: string) => html`
-  <h1 class="t-title t-title--main" id="compatibilityNotice">
-    Internet Identity does not support your browser
-  </h1>
-  ${warnBox({
-    title: "Reason",
-    message: reason,
+const pageContent = () =>
+  warnBox({
+    title: "Browser Not Supported",
+    message: html`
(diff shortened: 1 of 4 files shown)


src/frontend/src/flows/compatibilityNotice.ts
```diff
@@ -1,26 +1,31 @@
...
+const pageContent = () =>
+  warnBox({
+    title: "Browser Not Supported",
+    message: html`
+      <p class="t-paragraph">
+        Internet Identity is not supported by your browser.
+      </p>
+      <p class="t-paragraph">
+        This device does not offer
+        <a href="https://webauthn.io" rel="noopener noreferrer" class="t-link"
```
frederikrothenberger, line 15: I'm not sure we want to redirect users to webauthn.io. We could also use https://fidoalliance.org/passkeys/ since we're referring to passkeys everywhere on the front-end. This is also the official site of the industry association behind the WebAuthn standard.

frederikrothenberger (approved):
It would be interesting to now use an unsupported device that was correctly caught by the compatibility check (i.e. some very old or cheap mobile device without biometrics) to see what the actual error looks like.

But this is definitely an improvement. We can also simply wait if there are any complaints with regard to errors.

Thanks!

src/frontend/src/flows/compatibilityNotice.ts
nmattia: Sure thing! should I rename it to `Passkeys` then?

frederikrothenberger: Yes, I would recommend that. We can also ask @marydwyer what wording she prefers.

PR #2221, opened 2024-01-22
Refactor tentative device registration to use Result type internally

frederikrothenberger:
The tentative device registration uses the flattened success / error enum internally. This causes issues with further changes that we want to make, such as getting rid of internal `trap` calls for the API v2.

This is the first part of a bigger refactoring with the goal to remove the `trap` calls from `authenticated_anchor_operation`.


przydatek (approved):
LGTM, thanks!

PR #2222, opened 2024-01-22
Remove traps from authenticated_anchor_operation

frederikrothenberger:
This removes the traps from the authenticated_anchor_operation helper, which allows to add proper errors to `identity_metadata_replace`.

I will do a follow-up PR to move the authentication helpers and error mappers out of the `main.rs` file to keep it focused on the endpoints.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -92,7 +92,7 @@ async fn add_tentative_device(
             TentativeDeviceRegistrationError::AnotherDeviceTentativelyAdded => {
                 AddTentativeDeviceResponse::AnotherDeviceTentativelyAdded
             }
-            TentativeDeviceRegistrationError::AnchorOpError(err) => {
+            TentativeDeviceRegistrationError::IdentityUpdateError(err) => {
                 // Legacy API traps instead of returning an error.
                 trap(String::from(err).as_str())
             }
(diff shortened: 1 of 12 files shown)


frederikrothenberger:
> Can you briefly explain why the API breakage is fine?

The API breakage is fine because it is in the API v2 only, which is labeled as experimental.

src/internet_identity/src/main.rs
```diff
@@ -492,6 +501,31 @@ fn authenticate_and_record_activity(anchor_number: AnchorNumber) -> Option<IIDom
...
+    StorageError(IdentityNumber, StorageError),
+}
+
+impl From<IdentityUpdateError> for String {
+    fn from(err: IdentityUpdateError) -> Self {
+        match err {
+            IdentityUpdateError::Unauthorized => {
+                // This error message is used by the legacy API and should not be changed, even though
+                // it is confusing authentication with authorization.
+                format!("{} could not be authenticated.", caller())
```
przydatek, line 516: nit: calling `caller()` in a conversion-to-string function seems weird.  Maybe we should have `IdenityUpdateError::Unauthorized(Principal)` instead? That would also provide more info for debugging/tests.

src/internet_identity/src/main.rs
```diff
@@ -728,20 +765,50 @@ mod v2_api {
...
+            let storage_err = match value {
+                IdentityUpdateError::Unauthorized => {
+                    return IdentityMetadataReplaceError::Unauthorized
+                }
+                IdentityUpdateError::StorageError(_, storage_err) => storage_err,
+            };
+
+            match storage_err {
+                StorageError::EntrySizeLimitExceeded(_) => {
+                    IdentityMetadataReplaceError::StorageSpaceExceeded
```
przydatek, line 779: nit: also here, maybe we should keep the extra information in `StorageSpaceExceeded` (effective size, limit) available for debugging?

przydatek (approved):
LGTM, modulo two questions.
Thanks!

frederikrothenberger:
@przydatek: I added the context to the errors. Could you please take another look?

I will then also do another PR to also adapt the `PrepareIdAliasError::Unauthorized` variant to hold information about the `principal` that is unauthorized.

przydatek (approved):
LGTM, thanks!

PR #2223, opened 2024-01-22
Speed up and improve canister tests

nmattia:
This includes a couple of fixes that clarify the canister tests (in particular end-to-end tests) and that shorten CI runs in most cases (both for frontend & backend development). A typical CI run should now be ~5mn shorter (7mn now vs 12mn before):

* The 'selenium' tests are renamed to 'e2e' (since selenium isn't being used anymore)
* A new native build job, `cached-build:`, is added and used by the canister & e2e tests in lieu of the slower docker builds. The docker builds are still used to check the output of the native cached build.
* Some npm packages are removed/deduplicated
* Some `npm ci` steps are run in the background
* The `npm test` step is removed, since unit tests are currently run as part of the sharded e2e tests
* A new shard is added to the e2e tests to balance the recently added slow verifiable credential tests
* Rustup executables are eagerly loaded (mostly to clarify timings)

### DIFF
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -34,8 +34,10 @@ runs:
     # NOTE: we include the output of uname to ensure binary compatibility (e.g. same glibc)
     - uses: actions/cache@v3
       with:
-        path: ~/.cargo/bin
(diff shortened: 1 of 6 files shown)


frederikrothenberger (approved):
Nice, thanks a lot! 😄

PR #2227, opened 2024-01-23
Fix README release update sed command

nmattia:
This changes the `sed` command to use `|` separators instead of `/`, since `/` often occurs and branch (and hence: ref) names.


frederikrothenberger (approved):
Thanks for fixing it!
The separator is `|` not `:` now, right?

PR #2230, opened 2024-01-23
Bump npm ci timeout

nmattia:
This bumps the npm ci timeout from 1mn to 5mn. 1mn might have been a bit optimistic.


frederikrothenberger (approved):
Thanks, I just ran into that timeout on my PR. 😉

PR #2231, opened 2024-01-23
Make Unauthorized and InternalCanisterError errors consistent

frederikrothenberger:
This PR changes the interface for the VC MVP canister calls to propagate the principal of authorization errors and internal errors back to the
 client.


przydatek (approved):
LGTM, thanks!

PR #2234, opened 2024-01-25
Update ic stable structures dependency

frederikrothenberger:
Updating the stable structures allows making use of new features such as unbounded values in BTreeMaps.

Note: we will still retain a dependency on v0.5.6 because it is a dependency of `ic-types` as well. However, given the crate is pretty small, we can live with the duplication.


przydatek (approved):
LGTM, thanks!

PR #2236, opened 2024-01-25
Clean up test-app authentication

nmattia:
This removes the dependency on auth-client for greater flexibility in tests, removes an unused button (signout) from the test app, and makes makes the message-display function a bit more robuest/clear.

This will allow us to be more flexible when testing out new APIs, since we now fully control the authentication flow in the test.

### DIFF
--- a/demos/test-app/src/auth.ts
+++ b/demos/test-app/src/auth.ts
@@ -1,0 +1,123 @@
+import type { SignIdentity, Signature } from "@dfinity/agent";
+import {
+  Delegation,
+  DelegationChain,
+  DelegationIdentity,
+} from "@dfinity/identity";
+import { Principal } from "@dfinity/principal";
+
+// The type of response from II as per the spec
+interface AuthResponseSuccess {
+  kind: "authorize-client-success";
+  delegations: {
+    delegation: {
+      pubkey: Uint8Array;
+      expiration: bigint;
+      targets?: Principal[];
+    };
+    signature: Uint8Array;
+  }[];
+  userPublicKey: Uint8Array;
+}
+
+// Perform a sign in to II using parameters set in this app
+export const authWithII = async ({
+  url: url_,
+  maxTimeToLive,
+  derivationOrigin,
+  sessionIdentity,
+}: {
+  url: string;
+  maxTimeToLive?: bigint;
+  derivationOrigin?: string;
+  sessionIdentity: SignIdentity;
+}): Promise<DelegationIdentity> => {
+  // Figure out the II URL to use
+  const iiUrl = new URL(url_);
+  iiUrl.hash = "#authorize";
+
+  // Open an II window and kickstart the flow
+  const win = window.open(iiUrl, "ii-window");
+  if (win === null) {
+    throw new Error(`Could not open window for '${iiUrl}'`);
+  }
+
+  // Wait for II to say it's ready
+  const evnt = await new Promise<MessageEvent>((resolve) => {
+    const readyHandler = (e: MessageEvent) => {
+      window.removeEventListener("message", readyHandler);
+      resolve(e);
+    };
+    window.addEventListener("message", readyHandler);
+  });
+
+  if (evnt.data.kind !== "authorize-ready") {
+    throw new Error("Bad message from II window: " + JSON.stringify(evnt));
+  }
+
+  // Send the request to II
+  const sessionPublicKey: Uint8Array = new Uint8Array(
+    sessionIdentity.getPublicKey().toDer()
+  );
+
+  const request = {
+    kind: "authorize-client",
+    sessionPublicKey,
+    maxTimeToLive,
+    derivationOrigin,
+  };
+
+  win.postMessage(request, iiUrl.origin);
+
+  // Wait for the II response and update the local state
+  const response = await new Promise<MessageEvent>((resolve) => {
+    const responseHandler = (e: MessageEvent) => {
+      window.removeEventListener("message", responseHandler);
+      win.close();
+      resolve(e);
+    };
+    window.addEventListener("message", responseHandler);
+  });
+
+  const message = response.data;
+  if (message.kind !== "authorize-client-success") {
+    throw new Error("Bad reply: " + JSON.stringify(message));
+  }
+
+  return identityFromResponse({
+    response: message as AuthResponseSuccess,
+    sessionIdentity,
+  });
+};
+
+// Read delegations the delegations from the response
+const identityFromResponse = ({
+  sessionIdentity,
+  response,
+}: {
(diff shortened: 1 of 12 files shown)


demos/test-app/src/auth.ts
```diff
@@ -0,0 +1,123 @@
+import type { SignIdentity, Signature } from "@dfinity/agent";
+import {
+  Delegation,
+  DelegationChain,
+  DelegationIdentity,
+} from "@dfinity/identity";
+import { Principal } from "@dfinity/principal";
+
+// The type of response from II as per the spec
+interface AuthResponseSuccess {
```
peterpeterparker, line 10: You don't have that in a did files somewhere to avoid duplication and for to improve maintability?

demos/test-app/src/auth.ts
```diff
@@ -0,0 +1,123 @@
...
+  iiUrl.hash = "#authorize";
+
+  // Open an II window and kickstart the flow
+  const win = window.open(iiUrl, "ii-window");
+  if (win === null) {
+    throw new Error(`Could not open window for '${iiUrl}'`);
+  }
+
+  // Wait for II to say it's ready
+  const evnt = await new Promise<MessageEvent>((resolve) => {
```
peterpeterparker, line 46: typo on purpose to avoid to conflict with "event"?

demos/test-app/src/auth.ts
```diff
@@ -0,0 +1,123 @@
...
+    const readyHandler = (e: MessageEvent) => {
```
peterpeterparker, line 47: 1. if "message" is never emitted, the listener is never removed

2. should all these listener be provided as return value to the caller that way if the caller destroy itself before this functions ends, the caller can destroy those remaining listeners

demos/test-app/src/auth.ts
```diff
@@ -0,0 +1,123 @@
...
+  const response = await new Promise<MessageEvent>((resolve) => {
+    const responseHandler = (e: MessageEvent) => {
+      window.removeEventListener("message", responseHandler);
+      win.close();
+      resolve(e);
+    };
+    window.addEventListener("message", responseHandler);
+  });
+
+  const message = response.data;
```
peterpeterparker, line 82: can data be null or undefined? for example if "message" is emitted on the window but do not originate from II?

demos/test-app/src/index.tsx
```diff
@@ -151,30 +167,33 @@ function addMessageElement(message: unknown, received: boolean) {
...
+    return;
+  }
+
+  addMessageElement({ message: event.data, ty: "received" });
+
+  if (event?.data?.kind !== "authorize-client-success") {
+    return;
   }
+
+  const delegations = event.data.delegations.map((signedDelegation: any) => {
```
peterpeterparker, line 180: that's the same code as previously right?

peterpeterparker, line 180: any?

src/frontend/src/test-e2e/views.ts
```diff
@@ -803,14 +802,11 @@ export class DemoAppView extends View {
     await waitToClose(this.browser);
     // wait for the demo app to update the principal
     await this.browser.waitUntil(
-      async () => (await this.getPrincipal()) !== Principal.anonymous().toText()
+      async () => (await this.getPrincipal()) !== ""
```
peterpeterparker, line 805: that's correct?

peterpeterparker (commented):
You don't need to save the delegation in idb so that next login reads the existing session? It always start with a sign-in?

demos/test-app/src/auth.ts
nmattia: not in a did file but I have the type defined in II somewhere. Next I'm gonna extract it like I did for the VC types

demos/test-app/src/auth.ts
nmattia: yep! used in a couple other places. Let me know if you have other suggestions; happy to change but I'll change separately, everywhere at once

src/frontend/src/test-e2e/views.ts
nmattia: Yep! this used to default to the anonymous principal; now we just don't set one

demos/test-app/src/auth.ts
nmattia: > if "message" is never emitted, the listener is never removed

right, though if message is never emitted there's nothing else we can do but wait. Would you prefer to have a timeout or something that just errors out after a minute?

> should all these listener be provided as return value to the caller that way if the caller destroy itself before this functions ends, the caller can destroy those remaining listeners

I think that's overkill for a test helper, but I'll keep it in mind if I extract this into a lib. That being said, if the caller decide to destroy itself early it wouldn't have access to the return value yet; IIUC I'd need to return an object that contains both a clean up function and the promise

demos/test-app/src/auth.ts
nmattia: yep definitely; I'm thinking of simply re-using the zod objects to parse this once I have extracted the types. Thoughts?

demos/test-app/src/index.tsx
nmattia: Yeah, copy-pasted from agent-js

nmattia: good catch!

demos/test-app/src/auth.ts
peterpeterparker: > Would you prefer to have a timeout or something that just errors out after a minute?

Comparing to what I did for Papyrs, looks lik I did something like that to observe if the window/popup got closed.

```
private observeCloseTab() {
    this.closeTabInterval = setInterval(() => {
      if (!this.tab?.closed) {
        return;
      }

      this.clearCloseTabInterval();

      this.throwError('User interrupted sign in.');
    }, 500);
  }
```

I did not removed the listeners because my solution was implemented in a Stencil web components, so destroying the components destroyed the listeners. That's why I would implement the "overkill" so that if the consumer destroy itself, it can destroy the listener.

Generally speaking, keeping listener opens if often a root cause of issues. Don't have statistics but, happens more than once I would guess.

That said, I understand this is just a demo. So feel free to ignore my comment.

demos/test-app/src/auth.ts
peterpeterparker: I'm not a big fan of zod but agree it would do the job in that case.

demos/test-app/src/index.tsx
peterpeterparker: I know you wait for three before extracting a function but, I would personally extract one here. Enough important to have one single map of the delegations. Of course your call.

peterpeterparker (approved):
LGTM, thx

PR #2249, opened 2024-02-01
Update parsing of filenames, to handle filenames with dots

przydatek:
The parsing of file names of assets was assuming that the dot characters (`.`) are only in the extension part, which lead to misinterpretation of filenames like `some.file.with.dots.in.the.name.js.gz`.  This PR makes the parsing more robust, by looking at the suffix of the filename.

### DIFF
--- a/src/asset_util/src/lib.rs
+++ b/src/asset_util/src/lib.rs
@@ -422,19 +422,18 @@ pub fn collect_assets(dir: &Dir, html_transformer: Option<fn(&str) -> String>) -
 /// NOTE: behavior is undefined with symlinks (and esp. symlink loops)!
 fn collect_assets_rec(dir: &Dir, assets: &mut Vec<Asset>) {
     for asset in dir.files() {
-        let file_bytes = asset.contents().to_vec();
-        let (content, encoding, content_type) = match file_extension(asset) {
-            "css" => (file_bytes, ContentEncoding::Identity, ContentType::CSS),
-            "html" => (file_bytes, ContentEncoding::Identity, ContentType::HTML),
-            "ico" => (file_bytes, ContentEncoding::Identity, ContentType::ICO),
-            "json" => (file_bytes, ContentEncoding::Identity, ContentType::JSON),
-            "js" => (file_bytes, ContentEncoding::Identity, ContentType::JS),
-            "js.gz" => (file_bytes, ContentEncoding::GZip, ContentType::JS),
-            "png" => (file_bytes, ContentEncoding::Identity, ContentType::PNG),
-            "svg" => (file_bytes, ContentEncoding::Identity, ContentType::SVG),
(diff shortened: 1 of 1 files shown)


src/asset_util/src/lib.rs
```diff
@@ -465,21 +464,29 @@ fn collect_assets_rec(dir: &Dir, assets: &mut Vec<Asset>) {
...
-        .unwrap()
-        .1
+/// Returns the extension of the given file, and the encoding type.
+/// If the text after the last dot is "gz" (i.e. this is a gzipped file), then the extension
+/// is the text after the second to last dot and the last dot in the file name
+/// (e.g. "js" for "some.gzipped.file.js.gz"), and the encoding is `ContentEncoding::GZip`.
+/// Otherwise the extension is the text after the last dot in the file name,
+/// and the encoding is `ContentEncoding::Identity`.
+/// This corresponds to the file extension for the assets handled by this crate.
+fn file_extension<'a>(asset: &'a File) -> (&'a str, ContentEncoding) {
```
nmattia: this is mixing concerns (grabbing the extension + figuring out the encoding); how do you feel about splitting the filename from extension, and if the extension is `.gz`, set the encoding to Gzip and check the extension on the rest of the filename?

nmattia: or alternatively return a truple of filename, content type & content encoding

przydatek: Thanks for the early feedback, indeed I've changed the helper to do the mapping to enums internally.

src/asset_util/src/lib.rs
```diff
@@ -465,21 +448,45 @@ fn collect_assets_rec(dir: &Dir, assets: &mut Vec<Asset>) {
...
-        .split_once('.')
-        .unwrap()
-        .1
+/// Returns the content type and the encoding type of the given file, based on the extension(s).
+/// If the text after the last dot is "gz" (i.e. this is a gzipped file), then content type
+/// is determined by the text after the second to last dot and the last dot in the file name,
+/// e.g. `ContentType::JS` for "some.gzipped.file.js.gz", and the encoding is `ContentEncoding::GZip`.
+/// Otherwise the content type is determined by the text after the last dot in the file name,
+/// and the encoding is `ContentEncoding::Identity`.
+fn content_type_and_encoding(asset: &File) -> (ContentType, ContentEncoding) {
```
nmattia, line 457: nit: you could directly pass the path here, don't think you need the actual `File` anywhere? that'll give a better idea of what's happening (inferring from the path) and will simplify your tests a tiny bit

nmattia (approved):
LGTM, thanks! one small suggestion

src/asset_util/src/lib.rs
przydatek: Done.

PR #2253, opened 2024-02-05
Add a convention connecting credential specs and the issued VCs

przydatek:
Add a convention for connecting credential specifications with the issued credentials. 
The convention enables automated validation of VC-claims in issued credentials wrt. credential spec used by a relying party.  Also, adapt the existing demo issuer to follow the convention.

A helper for the automated validation will be added to `vc_util` in a follow-up PR.


demos/vc_issuer/tests/issue_credential.rs
```diff
@@ -617,6 +617,11 @@ fn should_issue_credential_e2e() -> Result<(), CallError> {
                 prepared_context: prepared_credential.prepared_context,
             },
         )?;
+
+        println!(
```
frederikrothenberger, line 621: Did you intend to leave this here?

src/vc_util/src/lib.rs
```diff
@@ -372,8 +372,16 @@ fn validate_claims_match_spec(
     let subject = Subject::from_json_value(credential_subject.clone()).map_err(|_| {
         inconsistent_jwt_claims("missing credentialSubject in VerifiedAdult JWT vc")
     })?;
+    let credential_type = &spec.credential_type;
```
frederikrothenberger, line 375: Shouldn't this now enable _generic_ validation? I.e. this function should not specifically validate anything related to "adult", right?

frederikrothenberger (commented):
Using this convention, the vc_util custom module could be removed, right? And be replaced by a generic validation function that takes a `CredentialSpec` as input?

Is this planned in a separate PR?

przydatek:
> Using this convention, the vc_util custom module could be removed, right? And be replaced by a generic validation function that takes a `CredentialSpec` as input?
Yes and yes.

> Is this planned in a separate PR?
Yes.

demos/vc_issuer/tests/issue_credential.rs
przydatek: Oops, not intended. Thanks for catching, removed now.

src/vc_util/src/lib.rs
przydatek: Yes, this does enable generic validation, which will come in a follow-up PR.

przydatek (commented):
Thanks, PTAL.

nmattia (approved):
LGTM!

PR #2254, opened 2024-02-06
Decouple Anchor from serialized bytes

frederikrothenberger:
This PR introduces an abstraction layer between the business logic type `Anchor` and the type used to serialize the data to stable memory.
This enables further refactorings and helpful abstractions without affecting stable memory.

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -63,7 +63,7 @@
 //! The [PersistentState] is serialized at the end of stable memory to allow for variable sized data
 //! without the risk of running out of space (which might easily happen if the RESERVED_HEADER_BYTES
 //! were used instead).
-use std::convert::TryInto;
+use std::borrow::Cow;
 use std::fmt;
 use std::io::{Read, Write};
 use std::ops::RangeInclusive;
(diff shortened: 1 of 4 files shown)


src/internet_identity/src/storage.rs
```diff
@@ -279,27 +282,19 @@ impl<M: Memory + Clone> Storage<M> {
...
+        let record_number = self.anchor_number_to_record(anchor_number)?;
         let address = self.record_address(record_number);
-        let mut writer = BufferedWriter::new(
-            self.header.entry_size as usize,
-            Writer::new(&mut self.anchor_memory, address),
-        );
-        writer
-            .write_all(&(buf.len() as u16).to_le_bytes())
-            .expect("memory write failed");
+        let mut writer = Writer::new(&mut self.anchor_memory, address);
```
frederikrothenberger, line 296: We no longer need buffered reader / writers to interact with stable memory, as we write all of the parts (length and data) in one go.

src/internet_identity/src/storage/storable_anchor.rs
```diff
@@ -0,0 +1,32 @@
+use crate::storage::anchor::Device;
+use candid::{CandidType, Deserialize};
+use ic_stable_structures::storable::Bound;
+use ic_stable_structures::Storable;
+use internet_identity_interface::internet_identity::types::MetadataEntry;
+use std::borrow::Cow;
+use std::collections::HashMap;
+
+#[derive(Clone, Debug, Default, CandidType, Deserialize, Eq, PartialEq)]
+pub struct StorableAnchor {
```
frederikrothenberger, line 10: I decided to postpone the renaming to the "identity" based naming. This will be done once we actually change the terminology in the business logic / migrate stable memory.

przydatek (approved):
LGTM, thanks!

PR #2256, opened 2024-02-06
Add anchor number to Anchor type

frederikrothenberger:
This refactoring includes the anchor number in the `Anchor` business logic struct, making it impossible to mix up different anchors in storage.

### DIFF
--- a/src/internet_identity/src/storage/tests.rs
+++ b/src/internet_identity/src/storage/tests.rs
@@ -60,13 +60,13 @@ fn should_recover_header_from_memory_v7() {
 
 fn add_test_anchor_data<M: Memory + Clone>(storage: &mut Storage<M>, number_of_anchors: usize) {
     for _ in 0..number_of_anchors {
-        let (anchor_number, mut anchor) = storage
+        let mut anchor = storage
             .allocate_anchor()
             .expect("Failure allocating an anchor.");
         anchor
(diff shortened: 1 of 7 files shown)


src/internet_identity/src/storage/tests.rs
```diff
@@ -130,47 +131,37 @@ fn should_serialize_subsequent_record_to_expected_memory_location() {
...
     memory.read(
         RESERVED_HEADER_BYTES + EXPECTED_RECORD_OFFSET + LENGTH_OFFSET,
         &mut buf,
     );
     let decoded_from_memory: StorableAnchor = candid::decode_one(&buf).unwrap();
-    assert_eq!(Anchor::from(decoded_from_memory), anchor);
-}
-
-#[test]
-fn should_not_write_using_anchor_number_outside_allocated_range() {
```
przydatek, line 153: Shouldn't this test remain (after adjusting)?  One can still pass an anchor with anchor_number outside of allocated range, no?

przydatek (approved):
LGTM, modulo a question. Thanks!

src/internet_identity/src/storage/tests.rs
frederikrothenberger: Yeah, true. Although it is harder to do that now, as only the `storage` module itself can construct `Anchor` structs with arbitrary numbers.

PR #2257, opened 2024-02-08
Don't panic in vc_util

nmattia:
This ensures that the vc_util lib does not panic because of missing keys (keys specified in the cred spec argument but not in the subject properties).

This also updates the behavior when no arguments are supplied; before this the iteration was implicitly skipped by creating an empty hash map and iterating over it. Now we explicitly skip the check.


przydatek (approved):
LGTM, thanks!
(modulo clippy)

PR #2258, opened 2024-02-08
Package vc_util_js

nmattia:
This adds a new npm workspace, `src/vc_util_js`, which can be used to verify credentials.

See README.md for caveats.

NOTE: we do not pin the version of wasm-pack for simplicity (the wasm is not currently used, and should not be used in the production build anyway "ever").


przydatek (approved):
LGTM, thanks!

PR #2260, opened 2024-02-09
Don't include dev build checksum in metadata

nmattia:
This changes the dfx metadata to only include a link to the latest release asset.

Before this, the metadata included the checksum of the dev-flavored build, which incurred a big complexity cost because the implementation involved building the dev build before the production build could start.

Since `dfx` now supports not providing a checksum at all, and since the checksum is mostly useful with big canisters (bigger than II) to prevent re-downloading a cached canister, we just skip providing a checksum.

The verify-hash script now just performs the build and compare the hashes with those provided in the arguments.

Additionally, the generation of metadata now uses `jq` to avoid bash-multiline-string artifacts like newlines & inconsistent spacing in the generated JSON.


frederikrothenberger (approved):
Thanks, that's great!
`dfx pull` prints a warning if we include neither, the hash nor the hash url. How is this warning worded? Does it imply something being wrong? Because depending on that we might want to additionally nudge the SDK team to reword it to better support our usage of the feature.

PR #2261, opened 2024-02-09
Add a generic helper for validating II presentations with compliant VCs

przydatek:
Add a helper `validate_ii_presentation_and_claims` that validates presentations issued by II during the VC-flow.
The helper works for presentations that contain VCs compliant with the [convention for consistency with credential specs](https://github.com/dfinity/internet-identity/blob/main/docs/vc-spec.md#recommended-convention-connecting-credential-specification-with-the-returned-credentials).

A follow-up PR will remove `custom`-submodule of `vc_util`, and adapt the tests from the submodule to the new helper.


src/vc_util/src/lib.rs
```diff
@@ -1007,4 +1056,165 @@ mod tests {
...
+            ArgumentValue::String("string arg value".to_string()),
+        );
+        args.insert("secondArg".to_string(), ArgumentValue::Int(42));
+        CredentialSpec {
+            credential_type: "vcWithOneArg".to_string(),
+            arguments: Some(args),
+        }
+    }
+
+    fn credential_specs_for_test() -> Vec<CredentialSpec> {
```
frederikrothenberger, line 1092: If you were to go one step further I would suggest using `proptest` 😉

src/vc_util/src/lib.rs
```diff
@@ -1007,4 +1056,165 @@ mod tests {
...
+                .expect("wrong credentialSubject")
+                .remove(&spec.credential_type)
+                .expect("missing credential_type claim");
+            let result = validate_claims_match_spec(&claims, &spec);
+            assert_matches!(result, Err(e) if format!("{:?}", e).contains("missing credential_type claim"));
+        }
+    }
+
+    #[test]
+    fn should_fail_validate_claims_match_spec_with_extra_args_in_credential_type_claim() {
```
frederikrothenberger, line 1160: Maybe for better readability it would be ok to not repeate the function name verbatim and just use `should_fail_validation_...`?

src/vc_util/src/lib.rs
```diff
@@ -350,52 +384,65 @@ fn validate_expiration(
...
-                ));
-            }
-        } else {
-            return Err(inconsistent_jwt_claims(
-                "missing VerifiedAdult-claim in VerifiedAdult JWT vc",
-            ));
-        }
+    let spec_arguments_count = if let Some(spec_arguments) = spec.arguments.as_ref() {
+        spec_arguments.len()
+    } else {
```
frederikrothenberger, line 426: This could maybe be solved a bit more concisely using `Option::map_or`?

frederikrothenberger (approved):
LGTM, only minor issues.

src/vc_util/src/lib.rs
przydatek: With repeated function name it is clear what is being tested, and also I find it useful for filtering the tests.  
I don't feel strongly about it though, so if it is only useful for me, I can rename (leaving for now, can do another PR)

src/vc_util/src/lib.rs
przydatek: A good idea... for another step  😉 
(I believe we don't use `proptest` in II yet, do we?)

src/vc_util/src/custom.rs
```diff
@@ -23,28 +18,13 @@ pub fn validate_verified_adult_presentation(
     root_pk_raw: &[u8],
     current_time_ns: u128,
 ) -> Result<(), PresentationVerificationError> {
-    let (_alias_tuple, claims) = verify_ii_presentation_jwt_with_canister_ids(
+    validate_ii_presentation_and_claims(
         vp_jwt,
         effective_vc_subject,
         vc_flow_signers,
+        &verified_adult_vc_spec(),
```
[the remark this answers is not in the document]
frederikrothenberger: (sorry for the noise, I didn't read the description carefully enough.)

src/vc_util/src/lib.rs
przydatek: Indeed, done.

przydatek (commented):
Thanks!

PR #2263, opened 2024-02-12
Add inline script loader

nmattia:
This (re-)adds an inline script loader.

This is a workaround for [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) not working very well with whitelisted scripts in the CSP. See `inlineScriptsPlugin` for a description of the limitation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2210,6 +2210,7 @@ name = "internet_identity"
 version = "0.1.0"
 dependencies = [
  "asset_util",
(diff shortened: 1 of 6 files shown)


frederikrothenberger (approved):
Really unfortunate and surprising for me that this CSP hash whitelisting is not better supported given that SRI is really not a new thing... 🙁 

Thanks for the fix!

PR #2265, opened 2024-02-12
Remove custom-submodule from vc_util

przydatek:
Remove `vc_util::custom::*`, as the corresponding functionality is now provided by a generic validation function `validate_ii_presentation_and_claims`.  Update `vc-util-js` accordingly.


src/vc_util_js/src/lib.rs
```diff
@@ -20,10 +22,19 @@ pub fn validate_verified_adult_presentation(
         issuer_canister_id: Principal::from_slice(issuer_canister_id),
         issuer_origin: String::from_utf8(issuer_origin.to_vec()).expect("wrong issuer origin"),
     };
-    vc_util::custom::validate_verified_adult_presentation(
+    let credential_spec = {
```
frederikrothenberger, line 25: So this function will be removed as well at some point? And allow to pass through the credential spec from JS? Or is `vc_util_js` never intended to become the JS equivalent of `vc_util`?

frederikrothenberger (approved):
LGTM, thanks!

src/vc_util_js/src/lib.rs
przydatek: Yes, the function will be replaced with the generic validator that takes a spec.  But the js module is not indented to be fully equivalent to `vc_util`, at least for now.

przydatek:
Thanks!

PR #2266, opened 2024-02-13
Move VC/JWT-handling from the issuer into helpers in vc_util

przydatek:
Encapsulates the creation of VCs/JWTs in `vc_util`, so that the issuer does not directly depend on the underlying VC/JWT-handling libraries. This simplifies the dependencies, and allows for an easier swap of the VC/JWT-library under the API of `vc_util`.


src/vc_util/src/lib.rs
```diff
@@ -1367,4 +1407,32 @@ mod tests {
...
+        let params = CredentialParams {
+            spec: verified_adult_vc_spec(),
+            subject_id: did_for_principal(id_dapp),
+            credential_id_url: "https://age_verifier.info/credentials/42".to_string(),
+            issuer_url: "https://age_verifier.info".to_string(),
+            expiration_timestamp_s: (CURRENT_TIME_BEFORE_EXPIRY_NS / 1_000_000_000) as u32,
+        };
+        let credential = build_credential_jwt(params);
+        assert_eq!(credential.len(), example_jwt.len());
+        assert_ne!(credential, example_jwt);
```
frederikrothenberger, line 1434: As discussed, I misread the "not equal" part and was then confused about the removal of `nbf`. Maybe add a comment here that explains the two assertions? 😉

frederikrothenberger (approved):
Very nice!

src/vc_util/src/lib.rs
przydatek: Done.

przydatek (commented):
Thanks!

PR #2268, opened 2024-02-13
Remove redundant memory reserve

frederikrothenberger:
When canister memory still was small (and unmanaged) it was sensible to reserve some space at the end. Now that the canister could in theory address over 400GB of space _and_ we have more convenient ways to access that than write into the end adresses, that complexity can be removed.

The WASM page size constant is also redundant, as it is now provided by the CDK.


src/internet_identity/tests/integration/http.rs
```diff
@@ -171,7 +171,11 @@ fn ii_canister_serves_http_metrics() -> Result<(), CallError> {
 #[test]
 fn metrics_should_list_expected_user_range() -> Result<(), CallError> {
     let env = env();
-    let canister_id = install_ii_canister(&env, II_WASM.clone());
+    let canister_id = install_ii_canister_with_arg(
+        &env,
+        II_WASM.clone(),
+        arg_with_anchor_range((10_000, 8_188_860)),
```
przydatek, line 177: nit: maybe explain the constant `8_188_860` in a comment?  (it matches the *previous* number of max entries, right?)
or maybe extend the test to work e.g. for various anchor ranges, to cover also the *new* max number of entries?

frederikrothenberger: It matches the previous default, except that the configured upper bound is exclusive, while the reported max identity number is inclusive.
I added a test for the default range.

PR #2271, opened 2024-02-15
Remove low-level stable memory tests

frederikrothenberger:
This PR removes storage unit tests as well as II integration tests that make too strong assumptions about the underlying memory layout. These tests were very useful at the time when we were still operating manually on the raw memory and also during the migration to managed memory to make sure that all the bits and bytes end up exactly where we want them to.

Now, that the memory manager is well established, these tests are a hindrance to fully using its capabilities. As such, it is now acceptable to remove these tests and rely on the stable-structures crate itself being thoroughly tested.


przydatek (approved):
LGTM, thanks!
(great line-count for this PR!)

PR #2272, opened 2024-02-15
Fix: Improve vc-spec docs

lmuntaner:
Add related Candid types in the first section of the vc-spec


lmuntaner:
@przydatek please review

przydatek (approved):
LGTM, thanks a lot for catching and fixing these!

PR #2274, opened 2024-02-15
Move archive buffer to stable memory for freshly initialized II

frederikrothenberger:
This implements the archive entries buffer directly in stable memory, but only if II is initialized from a clean slat.
If the layout v7 is already there, it is maintained and the buffer in stable memory is just prepared.

A release containing this change can be used as a rollback version after migration.


src/internet_identity/src/storage/tests.rs
```diff
@@ -29,15 +29,27 @@ fn should_report_max_number_of_entries_for_32gb() {
...
     let memory = VectorMemory::default();
     let mut storage = Storage::new((1, 2), memory.clone());
     storage.update_salt([5u8; 32]);
     storage.flush();
 
+    assert_eq!(storage.version(), 8);
     let mut buf = vec![0; HEADER_SIZE];
     memory.read(0, &mut buf);
-    assert_eq!(buf, hex::decode("494943070000000001000000000000000200000000000000001005050505050505050505050505050505050505050505050505050505050505050000020000000000").unwrap());
+    assert_eq!(buf, hex::decode("494943080000000001000000000000000200000000000000001005050505050505050505050505050505050505050505050505050505050505050000020000000000").unwrap());
```
frederikrothenberger, line 52: Note: in only flipped the `version` byte from `7` to `8` in this hex string.

src/internet_identity/src/storage/tests.rs
```diff
@@ -29,15 +29,27 @@ fn should_report_max_number_of_entries_for_32gb() {
 }
 
 #[test]
-fn should_serialize_header_v7() {
+fn should_deserialize_header_v7() {
```
przydatek, line 32: nit: the same test for v7 is named `should_recover_header_from_memory_v7` below, maybe use a consistent pattern for both tests?

src/internet_identity/src/storage.rs
```diff
@@ -375,6 +420,38 @@ impl<M: Memory + Clone> Storage<M> {
         self.flush();
     }
 
+    /// Add a new archive entry to the buffer.
+    pub fn add_archive_entry(&mut self, entry: BufferedEntry) {
+        self.archive_entries_buffer
```
przydatek, line 425: would it make sense to (double) check here, and in the new helpers below, that we are indeed using v8?

przydatek (approved):
LGTM, thanks!
(nice tests, btw 🎉 )

src/internet_identity/src/storage/tests.rs
frederikrothenberger: I realized, the test is duplicate anyway. So I added `should_recover_header_from_memory_v8` but removed `should_deserialize_header_v7` as it is a duplicate of `should_recover_header_from_memory_v7`.

src/internet_identity/src/storage.rs
frederikrothenberger: Yes, that makes sense. Thanks!

PR #2277, opened 2024-02-16
Remove changes link in RC message

nmattia:
This removes the `changes: ...` link in the slack message sent upon RC release. It was slightly incorrect and not used.


frederikrothenberger (approved):
Thanks

PR #2278, opened 2024-02-16
Push newAnchor into LoginData

nmattia:
The authentication flow returns whether the user authenticated with a new or existing anchor. This data is used to tailor subsequent flows to new vs returning users.

Originally this information was added to all return types (success & error alike) for simplicity, although it only ever makes sense to return it upon successful authentication (until now we just defaulted to "not a new anchor" upon failures).

As we start passing more data through the flow (like whether the user used PIN or not) this shortcut became a bit problematic, so these changes push the `newAnchor` information deeper into the success state.

### DIFF
--- a/src/frontend/src/components/authenticateBox.ts
+++ b/src/frontend/src/components/authenticateBox.ts
@@ -92,11 +92,11 @@ export const authenticateBox = async ({
 
   // Retry until user has successfully authenticated
   for (;;) {
-    const { newAnchor, ...result } = await promptAuth();
+    const result = await promptAuth();
     const loginData = await handleLoginFlowResult(result);
 
     if (nonNullish(loginData)) {
-      return { ...loginData, newAnchor };
+      return loginData;
     }
   }
 };
@@ -163,7 +163,7 @@ export const authenticateBoxFlow = async <T, I>({
     pinIdentityMaterial: I;
   }) => Promise<"valid" | "expired">;
   registerFlowOpts: Parameters<typeof registerFlow<T>>[0];
-}): Promise<LoginFlowResult<T> & { newAnchor: boolean }> => {
+}): Promise<LoginFlowResult<T, { newAnchor: boolean }>> => {
   const pages = authnScreens(i18n, { ...templates });
 
   // The registration flow for a new identity
@@ -197,7 +197,7 @@ export const authenticateBoxFlow = async <T, I>({
 
   // Prompt for an identity number
   const doPrompt = async (): Promise<
-    LoginFlowResult<T> & { newAnchor: boolean }
+    LoginFlowResult<T, { newAnchor: boolean }>
   > => {
     const result = await pages.useExisting();
     if (result.tag === "submit") {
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/authenticateBox.ts
```diff
@@ -163,7 +163,7 @@ export const authenticateBoxFlow = async <T, I>({
     pinIdentityMaterial: I;
   }) => Promise<"valid" | "expired">;
   registerFlowOpts: Parameters<typeof registerFlow<T>>[0];
-}): Promise<LoginFlowResult<T> & { newAnchor: boolean }> => {
```
lmuntaner, line 166: Before it returned a promise with the parameter `newAnchor`?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -219,11 +219,17 @@ export const authenticateBoxFlow = async <T, I>({
...
+
+    const recoverResult = await recover();
+    if (recoverResult.tag === "ok") {
+      // If an anchor was recovered, then it's _not_ a new anchor
+      return {
+        newAnchor: false,
+        ...recoverResult,
+      };
+    } else {
+      return recoverResult;
```
lmuntaner, line 231: Does the `recoverResult` have a `newAnchor` field?

The type is `LoginFlowResult<T, E = object>`, isn't it? It's no assurance that `newAnchor` is present.

src/frontend/src/utils/flowResult.ts
```diff
@@ -2,19 +2,19 @@ import { DynamicKey } from "$src/utils/i18n";
...
   | LoginFlowCanceled;
 
-export type LoginFlowSuccess<T = AuthenticatedConnection> = {
+export type LoginFlowSuccess<T = AuthenticatedConnection, E = object> = {
   tag: "ok";
-} & LoginData<T>;
+} & LoginData<T, E>;
 
-export type LoginData<T = AuthenticatedConnection> = {
+export type LoginData<T = AuthenticatedConnection, E = object> = {
```
lmuntaner, line 14: Why don't you add `newAnchor` directly here?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -683,7 +686,12 @@ const useIdentityFlow = async <T, I>({
   result satisfies { kind: "pin" };
   const { result: pinResult } = result;
 
-  return { newAnchor: false, ...pinResult };
+  if (pinResult.tag === "ok") {
+    // We log in with an existing PIN anchor, meaning it is _not_ a new anchor
+    return { newAnchor: false, ...pinResult };
+  } else {
+    return pinResult;
```
lmuntaner, line 693: Ditto, does this result have `newAnchor` field?

lmuntaner (commented):
Thanks for sharing, I added a few questions. Sorry if they are stupid, I'm still learning :)

src/frontend/src/components/authenticateBox.ts
nmattia: only in the success case! that's the idea

src/frontend/src/components/authenticateBox.ts
nmattia: before it returned a `newAnchor` field always, now only in the success case (see description)

src/frontend/src/components/authenticateBox.ts
nmattia: it does not!

PR #2279, opened 2024-02-16
Change VC-id/jti to data-URL

przydatek:
Change the `credential_id`/`jti`-value used by II from a `https`-URL to `data`-URL. The `https`-URL is somewhat misleading as it suggests that some data can be fetched, while the main purpose of the id/jti-entry is to guarantee uniqueness.


frederikrothenberger (approved):
LGMT, thanks!

frederikrothenberger:
Should there be a test for that somewhere?

PR #2281, opened 2024-02-16
Add JS-API for validation of delegations

przydatek:
Add `validate_delegation_and_get_principal`-helper that thoroughly validates the given delegation and upon success returns a Principal that identifies the user (cf. description of the helper for details). 

There are some outstanding tasks for follow-up PRs (marked with TODO in comments):
- add a "deeper" expiration check for of the delegation (to check the "freshness" of the delegation)
- replace the added of code for computing representation independent hash with a crate (the crate exists, but is currently incomplete)
- add tests in JS for the wrapped functionality

### DIFF
--- a/src/canister_sig_util/src/lib.rs
+++ b/src/canister_sig_util/src/lib.rs
@@ -4,6 +4,7 @@ use lazy_static::lazy_static;
 use serde::Serialize;
 use serde_bytes::ByteBuf;
 use sha2::{Digest, Sha256};
+use std::collections::HashMap;
 
 pub mod signature_map;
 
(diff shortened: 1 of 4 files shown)


src/canister_sig_util/src/lib.rs
```diff
@@ -135,6 +136,133 @@ pub fn hash_bytes(value: impl AsRef<[u8]>) -> Hash {
     hasher.finalize().into()
 }
 
+pub fn delegation_signature_msg(
```
frederikrothenberger, line 139: Does this belong in the `canister_sig_utils` crate? I.e. shouldn't this crate provide the utilities around canister sigs that are independent of the signed message?

Not feeling super strongly, just wondering... Or is there a reason why _not_ to put it in `sig-verifier-js`?

frederikrothenberger (approved):
LGTM, thanks! Left one comment.

src/canister_sig_util/src/lib.rs
przydatek: Also not feeling strongly -- I thought that it fits to the crate, as verifying delegations is one of the main usages of canister sigs, and building this message is somewhat involved.  Let's ponder a bit more, and maybe move it in a follow up PR (that will remove the `hash`-copy).

przydatek:
Thanks!

PR #2285, opened 2024-02-20
Simplify auth flow types

nmattia:
This deduplicates the LoginSuccess & LoginFlow(Success,Result,Error) types and brings all error clarification to the top level.

Before this, several types (esp. `LoginSuccess` & `LoginFlowSuccess`) were used for similar purposes, with a translation layer (`apiResultToLoginFlowResult`) where necessary.

Moreover, a hierarchy of types existed (`ApiResult`, `RegisterResult`, ...) which obfuscated the actual subtypes and caused more error handling than actually necessary (by bundling types and surfacing errors that could never be triggered).

The changes implemented here do away with the "FlowResult" types (`LoginFlowResult`, `flowResult.ts`) and removes the artificial hierarchy between types (though introducing a new `FlowError` type representing a group of error types commonly found together).

Moreover the translation between error type and humand readable errors is left to the top-level caller, meaning error types stay in a more precise (i.e. non-string) format for longer.

These changes lead to less error handling and more precise error reporting, as well as fewer types to handle and translate.


src/frontend/src/components/authenticateBox.ts
```diff
@@ -256,27 +280,83 @@ export const authenticateBoxFlow = async <T, I>({
...
+
+// Makes the error human readable
+const clarifyError: {
+  [K in FlowError["kind"]]: (err: AssociatedError<K>) => {
+    title: string;
+    message: string;
+    detail?: string;
+  };
+} = {
+  authFail: (err) => ({
```
nmattia, line 309: All these messages were copied from now-gone `flowResult.ts`

src/frontend/src/flows/recovery/recoverWith/phrase.ts
```diff
@@ -312,36 +303,40 @@ export const wordTemplate = ({
...
+        const result = await withLoader(() =>
+          connection.fromSeedPhrase(userNumber, words.join(" "))
         );
-        return result;
+        if (result.kind !== "loginSuccess") {
+          const msg = {
+            noSeedPhrase:
+              "No recovery phrase: there is no recovery phrase associated with this identity. Cancel and recover another way.",
+            seedPhraseFail:
+              "Invalid Seed Phrase: Failed to authenticate using this seed phrase. Did you enter it correctly?",
```
nmattia, line 332: Taken from `flowResult.ts` where they could have actually never been triggered (there errors were not allowed to bubble up) and added here

src/frontend/src/components/authenticateBox.ts
```diff
@@ -256,27 +280,83 @@ export const authenticateBoxFlow = async <T, I>({
...
+  | WebAuthnFailed
+  | UnknownUser
+  | ApiError
+  | RegisterNoSpace;
+
+// Type machinery to allow translating error code (e.g. kind: "authFail") using the actual data
+// contained in the error objects (e.g. authFail: (authFail: AuthFail) => ...).
+type ErrorKind = FlowError["kind"]; // all known error tags for FlowError
+type KindToError = { [P in ErrorKind]: Omit<FlowError & { kind: P }, "kind"> }; // Look up from err. kind to object
+type AssociatedError<K extends ErrorKind> = {
```
nmattia, line 297: I wonder if we can squash those two

src/frontend/src/components/authenticateBox.ts
```diff
@@ -256,27 +280,83 @@ export const authenticateBoxFlow = async <T, I>({
...
+  | BadPin
+  | BadChallenge
+  | WebAuthnFailed
+  | UnknownUser
+  | ApiError
+  | RegisterNoSpace;
+
+// Type machinery to allow translating error code (e.g. kind: "authFail") using the actual data
+// contained in the error objects (e.g. authFail: (authFail: AuthFail) => ...).
+type ErrorKind = FlowError["kind"]; // all known error tags for FlowError
```
lmuntaner, line 295: Do they all the FlowError have a `"kind"` property?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -256,27 +280,83 @@ export const authenticateBoxFlow = async <T, I>({
...
+type KindToError = { [P in ErrorKind]: Omit<FlowError & { kind: P }, "kind"> }; // Look up from err. kind to object
```
lmuntaner, line 296: Could you add a comment of what this translates to with an example?

I had to ask chat gpt to translate it for me 😅

src/frontend/src/components/authenticateBox.ts
```diff
@@ -256,27 +280,83 @@ export const authenticateBoxFlow = async <T, I>({
...
+type AssociatedError<K extends ErrorKind> = {
+  [P in K]: { kind: P } & KindToError[P];
+}[K]; // Helper for adding the kind to the object (otherwise TS gets a bit lost)
```
lmuntaner, line 299: I'm a bit lost also 😂

src/frontend/src/components/authenticateBox.ts
nmattia: Yes! Otherwise this would be a type error I think

lmuntaner (approved):
Rubber stamp!

PR #2289, opened 2024-02-21
Clean up storage constants and remove 32GB limitation

frederikrothenberger:
This removes outdated constants in the `storage` module and removes an obsolete element from the II stable memory header. The 32GB memory limit is dropped as well. Due to limitations of the memory manager (to manage at most 256GB given the bucket size used) we still maintain a reserve of ~144GB stable memory that cannot be used for identities.


przydatek (approved):
LGTM, thanks!

PR #2290, opened 2024-02-22
Report virtual memory sizes on stats endpoint

frederikrothenberger:
Now that the memory manager manages multiple memories, it is no longer sufficient to report the total stable memory size only.

This PR introduces labeled metrics for each individual virtual memory.


przydatek (approved):
LGTM, thanks!

PR #2291, opened 2024-02-22
Add JS-tests for validateDelegationAndGetPrincipal

przydatek:
Add tests for `validateDelegationAndGetPrincipal`, and adapt the API slightly, to avoid unnecessary conversions.


frederikrothenberger (approved):
LGMT, thanks.

przydatek:
> Thanks!
> 
> This looks like it's re-running the rust tests. Are we really testing the implementation of the sig-verifier or are we testing the wasm bindgen implementation?
These tests are indeed performing checks similar to these in Rust, but they checking that whatever is running by calling the JS-API, we get the right responses.

src/sig-verifier-js/tests.ts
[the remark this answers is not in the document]
nmattia: Ack. I'll add a build & test separately

src/sig-verifier-js/tests.ts
```diff
@@ -1,18 +1,142 @@
...
+      {
+        "delegation":{
+          "expiration":"17b5b384762bfd21",
+          "pubkey":"e7875e69ce7beda6fc7b6d"
+        },
+      }
+    ],
+    "publicKey":"deadbeef"
+  }`; // missing "signature" after "delegation"
+  const call = async () => {
```
nmattia, line 85: here and elsewhere you could drop the `async/await`:

`const call = () => validateDelegationAndGetPrincipal(...)`

or even

`const result = validateDelegationAndGetPrincipal(...)`

as far as I'm concerned you can even inline the whole thing in `expect(validateDelegationAndGetPrincipal(...)).etc`

src/sig-verifier-js/tests.ts
```diff
@@ -1,18 +1,142 @@
...
+  0xb7, 0x5b, 0xe1, 0xc6, 0xf6, 0xd5, 0xde, 0xba, 0xe3, 0xae, 0x1e, 0xd7, 0xc7,
+  0xf8, 0x73, 0xde, 0x1b, 0x6f, 0x9f, 0x75, 0xe9, 0xe7, 0xdc, 0xdd, 0xcf, 0x37,
+  0xef, 0xad, 0xdc, 0xdf, 0x6f, 0x7b, 0x69, 0xa7, 0xb5, 0x73, 0x77, 0xb5, 0xdd,
+  0xae, 0xf8, 0x7d, 0xee, 0x38, 0x6d, 0xdd, 0x75, 0xe3, 0x9e, 0x9c, 0xd3, 0x9d,
+  0x7d, 0x77, 0xde, 0xbc, 0x79, 0xdf, 0x1b, 0x7b, 0x46, 0x9d, 0xf3, 0x6e, 0xb8,
+  0xe7, 0xce, 0xf4, 0x7b, 0x4d, 0x5c, 0xef, 0xa7, 0xf5, 0xdf, 0x67, 0xdb, 0xef,
+  0xc7, 0x3d, 0xeb, 0xdf, 0x5c,
+]);
+
+export var ROOT_PUBLIC_KEY_RAW = new Uint8Array([
```
nmattia, line 36: please use `const` when possible. Also I think you might be mutating this  in `Should fail validateDelegationAndGetPrincipal with wrong IC root public key` by settings the 42nd byte. Rearranging the tests might surface the issue.

nmattia (commented):
Thanks for addressing the comments!


> These tests are indeed performing checks similar to these in Rust, but they checking that whatever is running by calling the JS-API, we get the right responses.

ok, I'm not sure the signal we get from this is worth the test setup cost (which is a tad brittle due to hard coded challenge, delegations, etc) but I guess you'll be the one maintaining those tests

src/sig-verifier-js/tests.ts
przydatek: changed to `const`

src/sig-verifier-js/tests.ts
przydatek: removed `async/await`, thanks!

przydatek (commented):
Thanks, PTAL.

nmattia (approved):
thanks!

przydatek:
> thanks!

Thank you, @nmattia!

PR #2296, opened 2024-02-23
Migrate from storage v7 to v8 on upgrade

frederikrothenberger:
This PR introduces an automatic migration from v7 to v8 on upgrade. Since the actual amount of work required for the migration is very small, it can all be handled in the `post_upgrade` hook.

It is safe to automatically upgrade to v8, because the current release (`release-2024-02-23 `) is already compatible with v8.

This PR will be followed-up with separate changes to clean-up the legacy code, once the migration is complete.


przydatek (approved):
LGTM, thanks!

PR #2298, opened 2024-02-23
Allow authenticating dapps to disable PIN auth

nmattia:
This extends the II authentication API (postMessage interface) to include a new boolean field `allowPinAuthentication`. When set to false (default being true) Internet Identity will not allow the user to authenticating to the dapp using a PIN identity, and will not allow the user to register using a PIN identity.

This field is marked as experimental in the spec for now.


demos/test-app/src/index.tsx
```diff
@@ -203,11 +206,16 @@ const init = async () => {
     const derivationOrigin =
       derivationOriginEl.value !== "" ? derivationOriginEl.value : undefined;
 
+    let allowPinAuthentication = allowPinAuthenticationEl.checked
```
nmattia, line 209: use `const`

src/frontend/src/components/displayError.ts
```diff
@@ -31,7 +33,11 @@ ${options.detail}</pre
       })}
 
       <div class="l-stack">
-        <button id="displayErrorPrimary" class="c-button c-button--primary">
+        <button
+          id="displayErrorPrimary"
+          data-error-code=${ifDefined(options.errorCode)}
```
lmuntaner, line 38: Is this related to disabling the PIN?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -699,6 +720,7 @@ const useIdentityFlow = async <T, I>({
   | UnknownUser
   | ApiError
   | BadPin
+  | { kind: "pinNotAllowed" }
```
lmuntaner, line 723: Why not created a new type like `BadPin`?

src/frontend/src/test-e2e/flows.ts
```diff
@@ -104,19 +104,51 @@ export const FLOWS = {
...
+      pinAuthView.waitForDisplay().then(() => "pin_allowed" as const),
+      browser
+        .$('#errorContainer [data-error-code="pinNotAllowed"]')
+        .waitForDisplayed()
+        .then(() => "pin_disallowed" as const),
+    ]);
+    if (result === "pin_disallowed") {
+      return "pin_disallowed";
+    }
+    result satisfies "pin_allowed";
```
nmattia, line 128: Add a comment to explain what's happening

src/frontend/src/components/authenticateBox.ts
```diff
@@ -734,6 +756,11 @@ const useIdentityFlow = async <T, I>({
   }
   isValid satisfies "valid";
 
+  // if there is a PIN but allowPinAuth is false, then error out
```
lmuntaner, line 759: Where is the "PIN" in this function? Is it that `pinIdentityMaterial` is not nullish and `isValid` has `"valid"`?

lmuntaner: Could we also have checked this before checking the `isValid`?

lmuntaner (approved):
Thanks for sharing this!

I added some comments, but LGTM!

src/frontend/src/components/authenticateBox.ts
nmattia: Since they're all just type aliases it doesn't matter too much. Would you prefer having `PinNotAllowed`?

lmuntaner: No, I wanted to understand the patterns you use in the code.

src/frontend/src/components/authenticateBox.ts
nmattia: The flow first checks whether or not there's stored PIN material. If there's some, then it means the user is using a PIN and not webauthn. So if the PIN material exists (and is valid, i.e. not expired) then we check if PIN auth is allowed. It would not make sense to check if PIN auth is allowed because it's only relevant in case there PIN material exists.

Does this make sense?

src/frontend/src/components/displayError.ts
nmattia: It's related to the tests for disabling the PIN; I wanted to check that the error message shown was indeed that PIN auth is not allowed

src/frontend/src/components/authenticateBox.ts
lmuntaner: Ah, yes. Sorry, I got confused because I saw `isNullish(pinIdentityMaterial)` above, and I thought that was all the check needed.

I see that's just one exit in case there is no `pinIdentityMaterial`. But the user might still use a passkey even with a pin material.

peterpeterparker (approved):
LGTM, thx

marydwyer:
If it meets the request from @peterpeterparker, then it LGTM! Thanks!

PR #2300, opened 2024-02-23
Return information about authentication method

nmattia:
This adds a new field `authnMethod` to the auth API response. This field informs the authenticating dapp of which authentication method the user uesd: `passkey`, `pin` or `recovery`.

### DIFF
--- a/src/frontend/src/flows/authorize/postMessageInterface.ts
+++ b/src/frontend/src/flows/authorize/postMessageInterface.ts
@@ -53,6 +53,7 @@ export type AuthResponse =
       kind: "authorize-client-success";
       delegations: Delegation[];
       userPublicKey: Uint8Array;
+      authnMethod: "pin" | "passkey" | "recovery";
     };
 
 /**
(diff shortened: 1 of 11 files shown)


src/frontend/src/flows/authorize/postMessageInterface.ts
```diff
@@ -53,6 +53,7 @@ export type AuthResponse =
       kind: "authorize-client-success";
       delegations: Delegation[];
       userPublicKey: Uint8Array;
+      authnMethod: "pin" | "passkey" | "recovery";
```
lmuntaner, line 56: The type `AuthResponse` is not directly imported by the apps connecting to II, right?

It's just the type they receive. But adding a new mandatory field doesn't break anything.

lmuntaner (approved):
LGTM! Just one comment.

src/frontend/src/flows/authorize/postMessageInterface.ts
nmattia: > not directly imported by the apps connecting to II

Can you clarify what you mean by imported?

There is currently no type library exposed by II, right. And it's just the return type, so this should be backwards compatible!

lmuntaner: That's exactly what I meant. Thanks for clarifying.

PR #2307, opened 2024-02-26
Move compatibilityNotice showcase to astro

lmuntaner:
This migrates the `compatibilityNotice` page to a static Astro page to avoid having it in the hardcoded list of showcase components.


lmuntaner:
Current dependencies on/for this PR:
* **#2307** <a href="https://app.graphite.dev/github/pr/dfinity/internet-identity/2307?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/internet-identity/2307?utm_source=stack-comment).

lmuntaner:
@nmattia please review

src/showcase/src/layouts/Screen.astro
```diff
@@ -0,0 +1,11 @@
+---
+import Layout from "../layouts/Layout.astro";
+
+const { title, pageName } = Astro.props;
```
nmattia, line 4: do you know if Astro types the props if you don't specify the types? or does is default to `any`?

lmuntaner: Good point! Yes, I found it here: https://docs.astro.build/en/guides/typescript/#component-props

Done

src/showcase/src/layouts/Screen.astro
```diff
@@ -0,0 +1,19 @@
+---
+import Layout from "../layouts/Layout.astro";
+
+interface Props {
```
nmattia, line 4: nit: we use `type` pretty much everywhere in the codebase

src/showcase/src/layouts/Screen.astro
nmattia: Thanks! As far as I'm concerned you can even drop the `children` field

nmattia (approved):

src/showcase/src/layouts/Screen.astro
lmuntaner: Sounds good

src/showcase/src/layouts/Screen.astro
lmuntaner: The recommendation is that if the component requires a "slot" we should add it in the props.

This way I got an error when I used `<Screen>` without a child.

PR #2309, opened 2024-02-26
Remove support for storage v7

frederikrothenberger:
This PR removes II support for storage `v7`.
The previous release automatically migrates II to `v8`, so the code to handle `v7` can now be removed.


przydatek (approved):
LGTM, thanks!

PR #2310, opened 2024-02-26
Use HeaderField where applicable

nmattia:
A couple of signatures were using `(String, String)` instead of the more appropriate type alias `HeaderField = (String, String)`.


frederikrothenberger (approved):
LGTM, thanks!

PR #2312, opened 2024-02-27
Add derivation_origin-API to the demo issuer

przydatek:
Add `derivation_origin`-method to the demo issuer. Also, adapt issuer's initialisation to enable custom configuration for testing.

### DIFF
--- a/demos/vc_issuer/provision
+++ b/demos/vc_issuer/provision
@@ -17,9 +17,11 @@ Usage:
   $0 [--ii-canister-id CANISTER_ID] [--dfx-network NETWORK]
 
 Options:
-  --ii-canister-id CANISTER_ID  The canister ID to use as IDP, defaults to the local internet_identity canister
-  --dfx-network NETWORK         The network to use (typically "local" or "ic"), defaults to "local"
-  --issuer-canister CANISTER    The canister to configure (name or canister ID), defaults to "issuer"
+  --ii-canister-id CANISTER_ID    The canister ID to use as IDP, defaults to the local internet_identity canister
+  --dfx-network NETWORK           The network to use (typically "local" or "ic"), defaults to "local"
+  --issuer-canister CANISTER      The canister to configure (name or canister ID), defaults to "issuer"
+  --issuer-derivation-origin URL  The issuer's derivation origin, defaults to <issuer-canister-id>.ic0.app
+  --issuer-frontend-hostname URL  The issuer's frontend hostname, defaults to issuer's derivation origin
 EOF
 }
 
@@ -58,6 +60,16 @@ do
             shift; # shift past --issuer-canister & value
             shift;
             ;;
+        --issuer-derivation-origin)
+            DERIVATION_ORIGIN="${2:?missing value for '--issuer-canister'}"
+            shift; # shift past --issuer-canister & value
+            shift;
+            ;;
+        --issuer-frontend-hostname)
+            FRONTEND_HOSTNAME="${2:?missing value for '--issuer-canister'}"
+            shift; # shift past --issuer-canister & value
+            shift;
+            ;;
         *)
             echo "ERROR: unknown argument $1"
             usage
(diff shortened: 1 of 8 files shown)


demos/vc_issuer/vc_demo_issuer.did
```diff
@@ -102,6 +113,7 @@ service: (opt IssuerConfig) -> {
     vc_consent_message : (Icrc21VcConsentMessageRequest) -> (variant { Ok : Icrc21ConsentInfo; Err : Icrc21Error;});
     prepare_credential : (PrepareCredentialRequest) -> (variant { Ok : PreparedCredentialData; Err : IssueCredentialError;});
     get_credential : (GetCredentialRequest) -> (variant { Ok : IssuedCredentialData; Err : IssueCredentialError;}) query;
+    derivation_origin : (text) -> (variant {Ok: DerivationOriginData; Err: DerivationOriginError});
```
frederikrothenberger, line 116: What do you think, making the method also take a `record`? This would allow to name the `text` parameter and would make the method future proof to the addition of further arguments.

frederikrothenberger (approved):
This feature needs to be documented as well (in the VC spec). Will this be done in a separate PR?

demos/vc_issuer/src/main.rs
```diff
@@ -75,6 +76,10 @@ struct IssuerConfig {
     ic_root_key_raw: Vec<u8>,
     /// List of canister ids that are allowed to provide id alias credentials.
     idp_canister_ids: Vec<Principal>,
+    /// The derivation origin to be used by the issuer.
+    derivation_origin: String,
+    /// Frontend hostname be used by the issuer.
```
nmattia, line 81:
```suggestion
    /// Frontend hostname to be used by the issuer.
```

demos/vc_issuer/vc_demo_issuer.did
nmattia: 👍 or additionally explain what the parameter is/does

demos/vc_issuer/provision
```diff
@@ -58,6 +60,16 @@ do
             shift; # shift past --issuer-canister & value
             shift;
             ;;
+        --issuer-derivation-origin)
+            DERIVATION_ORIGIN="${2:?missing value for '--issuer-canister'}"
+            shift; # shift past --issuer-canister & value
```
nmattia, line 65:
```suggestion
            shift; # shift past --issuer-derivation-origin & value
```

nmattia: + "missing value for ..." both here and in the next param

demos/vc_issuer/vc_demo_issuer.did
nmattia: Actually, what is the parameter for?

nmattia (commented):
Thanks! Couple questions:
* Should the VC spec be updated as well?
* Is there an issuer without `get_derivation_origin` that we could use to test the no-derivation-origin case?

demos/vc_issuer/provision
przydatek: Done.

demos/vc_issuer/src/main.rs
przydatek: Done.

demos/vc_issuer/vc_demo_issuer.did
przydatek: A single issuer-BE could be serving several frontends.

przydatek: Good idea, changed to a record.

przydatek (commented):
Thanks, PTAL.

przydatek:
> Thanks! Couple questions:
> 
> * Should the VC spec be updated as well?
Yes, in a follow-up PR.
> * Is there an issuer without `get_derivation_origin` that we could use to test the no-derivation-origin case?
We can take a wasm built before this PR, or guard the feature with a flag.  What would be more convenient for CI?

nmattia:
> * Is there an issuer without get_derivation_origin that we could use to test the no-derivation-origin case?
> We can take a wasm built before this PR, or guard the feature with a flag. What would be more convenient for CI?

The feature flag should work nicely, good idea! I'll add it when I add the tests

PR #2320, opened 2024-02-28
Look up replica port in dev server

nmattia:
This ensures the correct replica port is used even if the user uses a different port than the default `4943` port.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -5,6 +5,7 @@ import {
   minifyHTML,
   replicaForwardPlugin,
 } from "@dfinity/internet-identity-vite-plugins";
+import { readReplicaPort } from "@dfinity/internet-identity-vite-plugins/utils";
 import basicSsl from "@vitejs/plugin-basic-ssl";
 import { resolve } from "path";
 import { AliasOptions, UserConfig, defineConfig } from "vite";
@@ -18,7 +19,7 @@ export const aliasConfig: AliasOptions = {
   $showcase: resolve(__dirname, "src/showcase/src"),
 };
 
-export default defineConfig(({ mode }: UserConfig): UserConfig => {
+export default defineConfig(({ command, mode }): UserConfig => {
   // Expand process.env variables with default values to ensure build reproducibility
   process.env = {
     ...process.env,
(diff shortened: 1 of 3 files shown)


vite.config.ts
```diff
@@ -28,6 +29,13 @@ export default defineConfig(({ mode }: UserConfig): UserConfig => {
     II_VERSION: `${process.env.II_VERSION ?? ""}`,
   };
 
+  // Astro gets a bit confused and tries to load the server config (including devserver-only
+  // plugins) on build. This ensures that plugins that need dfx are only invoked on serve.
+  // https://github.com/withastro/astro/issues/10262
+  const isAstro =
+    process.env["npm_lifecycle_script"]?.includes("astro") === true;
+  const isServe = command === "serve" && !isAstro;
```
lmuntaner, line 37: What does `"serve"` means here? To serve only static assets?

lmuntaner (approved):
Thanks! LGTM.

Just a question about a concept.

lmuntaner (changes requested):
Removing approval. I tested it locally and it doesn't work. I'll try to debug it.

vite.config.ts
```diff
@@ -72,24 +80,29 @@ export default defineConfig(({ mode }: UserConfig): UserConfig => {
...
+                      "identity.ic0.app",
+                      "identity.internetcomputer.org",
+                    ],
+                    canisterName: "internet_identity",
+                  },
+                ]
+              : []),
+          ],
+        }),
+        apply: () => isServe,
```
lmuntaner, line 104: What does this do?

vite.config.ts
```diff
@@ -98,11 +111,13 @@ export default defineConfig(({ mode }: UserConfig): UserConfig => {
         },
       },
     },
-    server: {
-      https: process.env.TLS_DEV_SERVER === "1",
-      proxy: {
-        "/api": "http://127.0.0.1:4943",
-      },
-    },
+    server: !isServe
```
lmuntaner, line 114: Why not start with the affirmative? `isServer ? ... : ...`?

vite.config.ts
nmattia: I suspect it's "serve" as in "serve dev server": https://vitejs.dev/config/#conditional-config

it's not only the static assets but the whole server (whatever might be served, in our case incl. serving the IC API & proxying to canisters)

vite.config.ts
nmattia: Is applies the plugin only when `true`: https://vitejs.dev/guide/api-plugin#conditional-application

vite.config.ts
lmuntaner: Ok, I might in another PR

vite.config.ts
lmuntaner: Ok, thanks!

vite.config.ts
lmuntaner: ok, thanks!

lmuntaner (approved):

PR #2321, opened 2024-02-28
Add support for persistent state in separate virtual memory

frederikrothenberger:
This PR adds support for the storage layout `v9` which keeps the persistent state in a separate virtual memory.
The PR does _not_ include any migration. Existing II installations will keep layout `v8`.


src/internet_identity/src/storage.rs
```diff
@@ -145,7 +146,7 @@ pub const MAX_ENTRIES: u64 = (MAX_MANAGED_WASM_PAGES - BUCKET_SIZE_IN_PAGES as u
 
 pub type Salt = [u8; 32];
 
-type ArchiveBufferMemory<M> = RestrictedMemory<VirtualMemory<RestrictedMemory<M>>>;
+type SingleBucketMemory<M> = RestrictedMemory<VirtualMemory<RestrictedMemory<M>>>;
```
frederikrothenberger, line 149: It would be cleaner to wrap the underlying memory such that:
* the memory is guaranteed to be limited to a single bucket
* the memory cannot be written to

However, I'll do this in a separate PR.

src/internet_identity/tests/integration/archive_integration.rs
```diff
@@ -17,20 +17,38 @@ use std::collections::HashMap;
...
+    assert_eq!(
+        ii_api::stats(env, ii_canister)
+            .unwrap()
+            .storage_layout_version,
+        9
+    );
+    ii_canister
+}
+
+fn ii_canisters_under_test(
```
przydatek, line 45: A nice improvement! 🎉

src/internet_identity/src/storage/tests.rs
```diff
@@ -29,16 +29,16 @@ fn should_report_max_number_of_entries_for_256gb() {
 }
 
 #[test]
-fn should_serialize_header_v8() {
+fn should_serialize_header_v9() {
```
przydatek, line 32: nit: shouldn't we still keep the test for `v8`?  (i.e. add a new one for `v9`)

przydatek (approved):
LGTM, thanks!

src/internet_identity/src/storage/tests.rs
frederikrothenberger: New storage is now automatically initialized in `v9`. The ability to initialize `v8` is removed.

Compatibility with existing `v8` storage is tested in the (unchanged) `should_recover_header_from_memory_v8` test.

PR #2324, opened 2024-02-29
Wrap memories in helper struct to prevent accidental writes

frederikrothenberger:
This PR wraps the memory fields of the `storage` struct with a helper struct to better express the purpose of that field and prevent accidental misuse.
Additionally, the `SingleBucketMemory` is renamed to `NestedRestrictedMemory` to better convey what it actually represents.

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -102,7 +102,7 @@ use internet_identity_interface::internet_identity::types::*;
 
 use crate::state::PersistentState;
 use crate::storage::anchor::Anchor;
-use crate::storage::size_ref::MemorySizeRef;
+use crate::storage::memory_wrapper::MemoryWrapper;
 use crate::storage::storable_anchor::StorableAnchor;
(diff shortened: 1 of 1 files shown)


src/internet_identity/src/storage.rs
```diff
@@ -635,3 +630,24 @@ impl fmt::Display for StorageError {
...
     }
 }
+
+/// Helper module to hide internal memory of the memory size reference.
+mod size_ref {
+    use ic_stable_structures::Memory;
+
+    /// Struct that holds a memory with the sole purpose to provide a function to get
+    /// the size of the memory.
+    pub struct MemorySizeRef<M: Memory> {
```
przydatek, line 640: nit: maybe `MemoryWrapper` or something similar?  `MemorySizeRef` sounds very cryptic, and may become "outdated" if another method is added to the impl 😉

przydatek (approved):
Nice clarification! LGTM, thanks!

PR #2326, opened 2024-03-01
Decouple business logic from stored persistent state

frederikrothenberger:
This PR introduces an indirection between the `PersistentState` used in business logic and the one written to stable memory (in layout version v9). This allows decoupling refactoring the business logic from storage layout migrations.

Further, this PR makes sure that all the optionally configurable fields are initialized and serialized with a sensible default. This allows removing a lot of `Option` handling in the future.


src/internet_identity/src/state.rs
```diff
@@ -12,22 +12,26 @@ use candid::{CandidType, Deserialize};
...
 use std::collections::HashMap;
 use std::ops::{Deref, DerefMut};
 use std::time::Duration;
 
 mod temp_keys;
 
-// Default value for max number of delegation origins to store in the list of latest used delegation origins
-const MAX_NUM_DELEGATION_ORIGINS: u64 = 1000;
+/// Default value for max number of delegation origins to store in the list of latest used delegation origins
+pub const DEFAULT_MAX_NUM_DELEGATION_ORIGINS: u64 = 1000;
```
przydatek, line 25: nit: there is a slight inconsistency: here we have `DEFAULT_MAX_NUM_..._ORIGINS`, and below we have `DEFAULT_MAX_..._CAPTCHAS` (i.e. without `NUM`), maybe remove the inconsistency? (either way)

src/internet_identity/src/storage/storable_persistent_state.rs
```diff
@@ -0,0 +1,126 @@
...
+            domain_active_anchor_stats: Some(storable_persistent_state.domain_active_anchor_stats),
+            active_authn_method_stats: Some(storable_persistent_state.active_authn_method_stats),
+            latest_delegation_origins: Some(storable_persistent_state.latest_delegation_origins),
+            max_num_latest_delegation_origins: Some(
+                storable_persistent_state.max_num_latest_delegation_origins,
+            ),
+            max_inflight_captchas: Some(storable_persistent_state.max_inflight_captchas),
+        }
+    }
+}
```
przydatek, line 126: maybe add some basic test(s)? (e.g. a round-trip test for for the `From`-impls, and a stability test for the `Default`-impl?)

przydatek (approved):
LGTM, thanks!

src/internet_identity/src/storage/storable_persistent_state.rs
frederikrothenberger: Added tests for both. Good point.

PR #2331, opened 2024-03-04
Fix wrong link in add device instructions

frederikrothenberger:
The link on the add tentative device disabled page was always pointing to `identity.ic0.app`. This PR fixes it to point to the same origin instead. This way it is also correct for dev setups, etc.

### DIFF
--- a/src/frontend/src/flows/addDevice/welcomeView/deviceRegistrationModeDisabled.ts
+++ b/src/frontend/src/flows/addDevice/welcomeView/deviceRegistrationModeDisabled.ts
@@ -74,7 +74,7 @@ export const deviceRegistrationDisabledInfo = (
   return new Promise((resolve) =>
     deviceRegistrationDisabledInfoPage({
       userNumber,
-      iiUrl: window.origin,
+      origin: window.origin,
       cancel: () => resolve("canceled"),
       retry: () => resolve("retry"),
     })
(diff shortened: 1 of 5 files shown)


lmuntaner (commented):
It seems `LEGACY_II_URL_NO_PROTOCOL` is not used anymore. Could it be removed?

lmuntaner (approved):
Already approving though.

src/frontend/src/flows/addDevice/welcomeView/deviceRegistrationModeDisabled.ts
```diff
@@ -22,8 +21,8 @@ const deviceRegistrationDisabledInfoTemplate = ({
     </hgroup>
     <ol class="c-list c-list--numbered l-stack">
       <li>
-        Connect to ${LEGACY_II_URL_NO_PROTOCOL} on a recognized device using
-        Internet Identity ${userNumber}
+        Connect to ${window.origin} on a recognized device using Internet
```
nmattia, line 24: note: this now includes the protocol. I think originally we assumed this was purely informative and not really relevant in the dev build. Not saying we shouldn't make the change.

It's also slightly unfortunate that the screen (e.g. in showcase) is now dynamic and depends on where the page is served. Might be nice to pass in the origin as a parameter instead.

frederikrothenberger: I think including the protocol is ok. It's also nice, that it points to the correct location for staging, etc. deployments.

I added the parameter.

lmuntaner (approved):
Thanks!

PR #2334, opened 2024-03-05
Simplify canister sig public keys in VC demo issuer

przydatek:
The canister sigs public keys used by VC demo issuer were unnecessary "individualised" (i.e. a new seed was used for each request), which is not really needed in this context. This PR simplifies the code by switching to a single static public key.


frederikrothenberger (approved):
LGTM, thanks a lot! 🙂

PR #2335, opened 2024-03-05
Add support for issuer custom/alternative origin

nmattia:
This adds frontend support and end-to-end testing for custom issuer derivation origins. In particular:

* Updates the VC flow to support looking up & validating the derivation origin (see spec for details)
* Fixes a couple of bugs in VC issuer canister provisioning
* Adds support for building the VC issuer canister _with_ and _without_ custom origin support (through feature flag) (incl. provisioning, CI build, etc)
* Adds custom principal support in issuer frontend (as opposed to always adding the principal resulting from auth), necessary for testing with other origins
* Add a dfx.json entry for the new canister (`issuer_custom_origin`) + rounting in `vite.config.ts`
* Various error logging improvements in the issuer & the vc flow


demos/vc_issuer/src/main.rs
```diff
@@ -293,6 +295,7 @@ async fn vc_consent_message(
     )
 }
 
+#[cfg(feature = "custom_origin")]
```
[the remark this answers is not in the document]
nmattia: thanks @przydatek ! I'm not very familiar with the issuer and its test suite. Is this something you can take over?

> We need the build without the custom origin only for FE testing, right?

we need both actually, since we test both with and without custom origin (and the design relies on the method existing or not on the canister to differentiate the cases)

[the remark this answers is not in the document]
przydatek: I can take over, but you're very close already (and have made the necessary changes to `build.sh` and added `vc_demo_issuer_custom_origin-build`-section) 😄, so the only changes needed from here are:
- rename the feature flag to `exclude_custom_origin`
- change the cfg-guards to `#[cfg(not(feature = "exclude_custom_origin))]`
- rename  `vc_demo_issuer_custom_origin-build` to `vc_demo_issuer_exclude_custom_origin-build` in canister-tests.yml
- undo changes to `vc-issuer-test:` in canister-tests.yml
- rename in `build.sh` the parameter `--custom-origin` to `--exclude-custom-origin` (and flip the logic)

This way you'll have the existing tests working, and a new wasm (that excludes custom origin) for FE-tests.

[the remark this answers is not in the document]
przydatek: ah, the right syntax for the cfg-guard seems to be `#[cfg_attr(not(feature = "exclude_custom_origin"), ignore)]` as you already use in line 621 below, sorry! 
(btw, with a change to `exclude_custom_origin` the line 621 won't be needed)

.github/workflows/canister-tests.yml
```diff
@@ -263,7 +290,7 @@ jobs:
           mkdir dist
           touch dist/index.{html,css,js}
           touch dist/index2.js
-          cargo test
+          cargo test --features 'custom_origin'
```
przydatek, line 293: If you don't want to flip logic of the feature flag, please copy vc_demo_issuer_custom_origin.wasm.gz to vc_demo_issuer.wasm.gz before running the tests, so that the tests can find the wasm.

nmattia: > If you don't want to flip logic of the feature flag

I really don't care! Up to you. This was the easiest thing to do with what I had.

dfx.json
```diff
@@ -20,6 +20,14 @@
       "build": "demos/vc_issuer/build.sh",
       "post_install": "demos/vc_issuer/provision",
       "dependencies": [ "internet_identity" ]
+    },
+    "issuer_custom_origin": {
+      "type": "custom",
+      "candid": "demos/vc_issuer/vc_demo_issuer.did",
+      "wasm": "demos/vc_issuer/vc_demo_issuer_custom_origin.wasm.gz",
+      "build": "demos/vc_issuer/build.sh --custom-origin -o demos/vc_issuer/vc_demo_issuer_custom_origin.wasm.gz",
+      "post_install": "bash -c 'demos/vc_issuer/provision --issuer-frontend-hostname https://nice-issuer-custom-orig.com --issuer-canister issuer_custom_origin --issuer-derivation-origin https://$(dfx canister id test_app).icp0.io'",
```
nmattia, line 29: I've just read that `dfx` sets these environment variables: `CANISTER_ID_<CANISTER_NAME_UPPERCASE>`

could switch to that

lmuntaner: Yes!

I also asked about env vars in dfx and Max shared this https://internetcomputer.org/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-envars with me

src/frontend/src/flows/verifiableCredentials/vcIssuer.ts
```diff
@@ -107,4 +109,51 @@ export class VcIssuer {
...
+    | { kind: "use_default" }
+    | { kind: "error" }
+  > => {
+    const actor = await this.createActor();
+
+    const methodName = "derivation_origin" as const;
+
+    let result;
+    try {
+      result = await actor[methodName]({
```
lmuntaner, line 128: Why do you do it like this instead of `actor.derivatoin_origin`?

src/frontend/src/flows/verifiableCredentials/vcIssuer.ts
```diff
@@ -107,4 +109,51 @@ export class VcIssuer {
...
+        frontend_hostname: origin,
+      });
+    } catch (e: unknown) {
+      try {
+        // XXX: this "ensures" that the error is the one expected by the issuer spec.
+        // This unfortunately depends on the exact method name, only works for query calls and
```
lmuntaner, line 134: The standard doesn't say it should be marked a query call: https://github.com/dfinity/internet-identity/blob/36621c439f3eef17ed4c2fba9b2f8160b656b95e/docs/vc-spec.md#L4

Is that a problem? Can any method be called as query even it's not marked as query?

src/vc-api/src/generated/vc_issuer_types.ts
lmuntaner, line 1: Which is the script to generate this file?

src/frontend/src/flows/verifiableCredentials/vcIssuer.ts
nmattia: mostly because this is brittle af because we have to check the error message; here at least if the method name ever changes then the error message check will get updated too

src/frontend/src/flows/verifiableCredentials/vcIssuer.ts
nmattia: good catch, that's a typo. I meant "update".

nmattia: Mh actually I'm not sure what you mean. This is an update call, right? and called as such in this comment?

dfx.json
nmattia: actually I'll wait a bit until new versions of dfx are a bit more widespread (mine doesn't support it for instance). Feel free to track this somewhere!

src/frontend/src/flows/verifiableCredentials/vcIssuer.ts
nmattia: Done!

nmattia:
Superseded by #2356

PR #2336, opened 2024-03-05
Move identityBackground

lmuntaner:
Final motivation is to remove the `iiPages` from `showcase.ts` to the `index.astro` by reading the page files.

I found that I would need to import a constant in the server side, but there was an error because the constans.ts file was also importing the `identityBackground` which doesn't work in the server.

Therefore, I move the `identityBackground` to a new file `backgrounds.ts`.

### DIFF
--- a/src/showcase/src/pages/displayManage.astro
+++ b/src/showcase/src/pages/displayManage.astro
@@ -5,9 +5,11 @@ import Screen from "../layouts/Screen.astro";
 <Screen title="Display Manage" pageName="displayManage">
   <script>
     import { toast } from "$src/components/toast";
-    import { userNumber, identityBackground } from "../constants";
+    import { userNumber } from "../constants";
+    import { identityBackground } from "../backgrounds";
     import { dapps } from "../constants";
     import { displayManagePage } from "$src/flows/manage";
+    import { html } from "lit-html";
 
     displayManagePage({
       identityBackground,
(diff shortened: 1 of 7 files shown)


lmuntaner:
* **#2336** <a href="https://app.graphite.dev/github/pr/dfinity/internet-identity/2336?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @lmuntaner and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/showcase/src/pages/displayManage.astro
```diff
@@ -28,7 +30,7 @@ import Screen from "../layouts/Screen.astro";
             alias: "Yubikey Blue",
             remove: () => toast.info("remove"),
             rename: () => toast.info("rename"),
-            warn: "Something is rotten in the state of Device",
+            warn: html`Something is rotten in the state of Device`,
```
lmuntaner, line 33: This was a TS type issue that was not caught. `warn` expects a template, not a string.

nmattia: probably worth adding an `astro check` or something right? separately

nmattia (approved):
Rubber stamp; note there seems to be some issue with file endings

src/showcase/src/pages/displayManage.astro
lmuntaner: Yes, true, maybe.

PR #2337, opened 2024-03-06
Bump version of upload-pages-artifcat due to using node16

frederikrothenberger:
The upload-pages-artifcat action used node16 which is deprecated. Upgrading to the latest version fixes the issue.


lmuntaner (approved):
Thanks!

PR #2338, opened 2024-03-06
Fix dead-end in device add flow

frederikrothenberger:
This PR fixes a dead end where users would end up stuck after adding a new passkey to an identity from the sign-in screen.

Note: This PR does not yet include the new confirmation screens as per figma. This will be done separately.


lmuntaner (approved):
I'm not yet 100% onboarded here, but this looks good and it's tested.

PR #2339, opened 2024-03-06
Remove persistent state size restriction

frederikrothenberger:
This PR removes the size restriction on the persistent state memory. The reason is, that the persistent state is saved during the `pre_upgrade` hook. If the canister hits the size restriction, it will panic and hence the canister would be un-upgradeable.

The size restriction included a safety margin of 20x, but given that the limitation is only enforced in specific places, not having a size limit at all is much safer.
We will add an alert that triggers on the memory size instead. Additionally, we should get rid of the `pre_upgrade` hook altogether.


przydatek (approved):
LGTM, thanks for catching the potential issue well in advance!

PR #2340, opened 2024-03-07
Change default frontend domain for issuer provisioning

przydatek:
Change vc_issuer's `provision`-script to use `icp0.io`-domain by default.


frederikrothenberger (approved):
LGTM, thanks!

PR #2342, opened 2024-03-07
Change feature II_INSECURE_REQUESTS to II_DEV_CSP and allow connections to localhost

frederikrothenberger:
This PR removes the feature II_INSECURE_REQUESTS in favour of a more general II_DEV_CSP feature. That more general feature now allows II to connect to localhost and its subdomains on both http and https, which is useful when developing an issuer and using the II dev build to test the integration.


przydatek (approved):
LGTM, thanks!

PR #2345, opened 2024-03-08
Bump astro to 4.4.15

nmattia:
This bumps astro to the latest version. The latest version includes a fix for an issue we had where astro loaded our `serve` build config. The workaround we implemented is removed.


nmattia:
Turns out it's 4.4.16 I want

PR #2348, opened 2024-03-08
Add feature flag to VC issuer to exclude custom origin

przydatek:
Add a feature flag to vc_issuer, to enable build without support for custom origin/`derivation_origin`-API.

### DIFF
--- a/demos/vc_issuer/src/main.rs
+++ b/demos/vc_issuer/src/main.rs
@@ -14,11 +14,12 @@ use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::HashSet;
 use vc_util::issuer_api::{
-    ArgumentValue, CredentialSpec, DerivationOriginData, DerivationOriginError,
-    DerivationOriginRequest, GetCredentialRequest, Icrc21ConsentInfo, Icrc21Error,
+    ArgumentValue, CredentialSpec, GetCredentialRequest, Icrc21ConsentInfo, Icrc21Error,
     Icrc21VcConsentMessageRequest, IssueCredentialError, IssuedCredentialData,
     PrepareCredentialRequest, PreparedCredentialData, SignedIdAlias,
 };
+#[cfg(not(feature = "exclude_custom_origin"))]
+use vc_util::issuer_api::{DerivationOriginData, DerivationOriginError, DerivationOriginRequest};
 use vc_util::{
     build_credential_jwt, did_for_principal, get_verified_id_alias_from_jws, vc_jwt_to_jws,
     vc_signing_input, vc_signing_input_hash, AliasTuple, CredentialParams,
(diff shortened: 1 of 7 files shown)


demos/vc_issuer/src/main.rs
```diff
@@ -172,14 +173,15 @@ fn authorize_vc_request(
         let config = config.get();
 
         for idp_canister_id in &config.idp_canister_ids {
-            if let Ok(alias_tuple) = get_verified_id_alias_from_jws(
+            match get_verified_id_alias_from_jws(
```
frederikrothenberger, line 176: Why was this logic changed? Is it somehow related to the derivation origin?

demos/vc_issuer/tests/issue_credential.rs
```diff
@@ -349,6 +352,7 @@ fn should_return_derivation_origin() {
 }
 
 #[test]
+#[cfg_attr(feature = "exclude_custom_origin", ignore)]
```
frederikrothenberger, line 355: Integration tests are currently only run on the "normal" version _with_derivation origin support. Do we need the switches here? It would also eliminate the `allow(unused)` above.

demos/vc_issuer/src/main.rs
przydatek: The logic remains the same, the change is just to print the error message in case of verification error (previously the error message was ignored).

demos/vc_issuer/tests/issue_credential.rs
przydatek: Good point, we run these test on a wasm with the feature, so no need to ignore them. Removed now the annotations, incl.  the `unused`-one.

przydatek (commented):
Thanks, PTAL.

dfx.json
```diff
@@ -17,8 +17,16 @@
       "type": "custom",
       "candid": "demos/vc_issuer/vc_demo_issuer.did",
       "wasm": "demos/vc_issuer/vc_demo_issuer.wasm.gz",
-      "build": "demos/vc_issuer/build.sh",
-      "post_install": "demos/vc_issuer/provision",
+      "build": "demos/vc_issuer/build.sh -o demos/vc_issuer/vc_demo_issuer.wasm.gz",
+      "post_install": "bash -c 'demos/vc_issuer/provision --issuer-frontend-hostname https://nice-issuer-custom-orig.com --issuer-canister issuer --issuer-derivation-origin https://$(dfx canister id test_app).icp0.io'",
+      "dependencies": [ "internet_identity" ]
+    },
+    "issuer_exclude_custom_origin": {
```
nmattia, line 24: you use a mix of `without` (test files) and `exclude` (here), would it make sense to keep it consistent?

demos/vc_issuer/src/main.rs
nmattia: @frederikrothenberger sorry this is my code, the original PR description including this:

> Various error logging improvements in the issuer & the vc flow


see here: https://github.com/dfinity/internet-identity/pull/2335

I don't feel too strongly about it

przydatek:
We decided make `derivation_origin`-API mandatory, so this PR is obsolete.

PR #2357, opened 2024-03-12
Update spec on VC issuer's derivation_origin

przydatek:
Update VC-spec making `derivation_origin` mandatory.  This simplifies the overall flow and avoids dependencies on exact wording of error messages.  Although the lack of the method is clearly distinguishable from other errors, per [IC wiki](https://wiki.internetcomputer.org/wiki/Error_Codes_returned_by_Internet_Computer), it is somewhat disguised by `agent.js`-API and needs matching on error messages, which is not quite future proof.  Also, most or even all real-life issuers will use the feature, so it is better to make it mandatory from the beginning, as long as there are only few issuers in development.
Also, fix the demo issuer's default domain for canister URLs.

### DIFF
--- a/demos/vc_issuer/src/main.rs
+++ b/demos/vc_issuer/src/main.rs
@@ -101,7 +101,7 @@ impl Storable for IssuerConfig {
 
 impl Default for IssuerConfig {
     fn default() -> Self {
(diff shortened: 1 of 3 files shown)


frederikrothenberger (approved):
LGTM, thanks!

PR #2358, opened 2024-03-12
Refactor tentative device add flow to no longer reload the page

frederikrothenberger:
This PR refactors the tentative device add flow to no longer call `window.location.reload()` which was an ugly hack to avoid implementing the cancel action properly.


lmuntaner (approved):
Do I understand correctly that there are no tests for these flows at the moment?

frederikrothenberger:
> Do I understand correctly that there are no tests for these flows at the moment?

There are tests, but only e2e tests (which don't care about whether the page reloads or not). Also, the `cancel` actions are only sparsely tested. But there are tests for the happy cases. 😉

PR #2359, opened 2024-03-12
Fix identity background in Showcase

lmuntaner:
The identity background was not loaded in the showcase pages.

For example: https://dfinity.github.io/internet-identity/displayManage/

The problem was that the import of images in Astro imports an object, not a string like it does in the rest of the project.

To fix it, I duplicated the import within the showcase folder and cast it to the Astro type.

I also had to add a parameter to `IdentityBackground` to allow using a different image than the one imported in the file.


lmuntaner:
* **#2359** <a href="https://app.graphite.dev/github/pr/dfinity/internet-identity/2359?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @lmuntaner and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

frederikrothenberger (approved):
LGTM, thanks. Is there a plan to deduplicate the image again (i.e. when the showcase refactoring is done)?

src/frontend/src/components/identityCard.ts
```diff
@@ -30,14 +30,14 @@ export class IdentityBackground {
 
   // Only the src attribute is used so the same Element can be shared across Templates
   public static singleton?: IdentityBackground;
-  public static getSingleton(): IdentityBackground {
-    IdentityBackground.singleton ??= new IdentityBackground();
+  public static getSingleton(image?: string): IdentityBackground {
```
nmattia, line 33: what happens if you call `getSingleton` twice in a row with different parameters?

src/showcase/src/backgrounds.ts
```diff
@@ -1,3 +1,8 @@
-import { loadIdentityBackground } from "$src/components/identityCard";
+import { IdentityBackground } from "$src/components/identityCard";
+import { ImageMetadata } from "astro";
+import backgroundImage from "./assets/identityCard.png";
 
-export const identityBackground = loadIdentityBackground();
+export const identityBackground = IdentityBackground.getSingleton(
+  // When running the build, this is also build but not in an Astro envioronment
```
nmattia, line 6: typos ("built", "environment")

src/showcase/src/backgrounds.ts
```diff
@@ -1,3 +1,8 @@
...
+  (backgroundImage as unknown as ImageMetadata).src
```
nmattia, line 7: why is the casting necessary?

nmattia (commented):
it's a bit unfortunate that production code is made less correct/safe for the showcase's sake! I'm also not convinced that it's worth paying that price if anyway the production image is now duplicated. How about at least making it a symlink?

lmuntaner:
> it's a bit unfortunate that production code is made less correct/safe for the showcase's sake! I'm also not convinced that it's worth paying that price if anyway the production image is now duplicated. How about at least making it a symlink?

I was able to remove it. I pushed it now #2361

src/showcase/src/backgrounds.ts
lmuntaner: because when building II it also check this import, whose time is not of ImageMetadata from Astro.

Once we move this to its own module maybe we'll be able to remove the casting.

src/frontend/src/components/identityCard.ts
lmuntaner: Good point, the first one will be used. Let me improve it in another PR.

nmattia: I'm not sure how a singleton would work with an argument but I'll stay tuned and learn something!

PR #2363, opened 2024-03-13
Ignore messages from own origin

frederikrothenberger:
Currently, II prints a warning to the console (or in case of the VC-flow even shows a toast) when it receives an unexpected message. Messages might get sent from browser extension (i.e. the MetaMask extension does that), which can be recognized by having the same origin as II itself. Since II does not send messages to itself, we can safely ignore all of these messages (and the messages do not warrant an `error`).


lmuntaner (approved):
LGTM!

PR #2364, opened 2024-03-14
Infer issuer canisterId from origin on localhost

frederikrothenberger:
For issuers hosted on `localhost` (using the a canister id as a subdomain) II will no longer make a network request to get the canister ID during the attribute sharing flow. Instead, the canister id will be parsed from the origin directly.

This change makes it possible to integrate issuers with II in a local dev setup if canisterId subdomain
of `localhost` is used as the issuer (derivation) origin.

### DIFF
--- a/src/frontend/src/flows/verifiableCredentials/index.ts
+++ b/src/frontend/src/flows/verifiableCredentials/index.ts
@@ -17,6 +17,7 @@ import {
   CredentialSpec,
   IssuedCredentialData,
 } from "@dfinity/internet-identity-vc-api";
+import { Principal } from "@dfinity/principal";
 import { isNullish, nonNullish } from "@dfinity/utils";
 import { abortedCredentials } from "./abortedCredentials";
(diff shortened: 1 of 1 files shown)


src/frontend/src/flows/verifiableCredentials/index.ts
```diff
@@ -334,28 +365,28 @@ const getValidatedIssuerDerivationOrigin = async ({
     }
   | { kind: "ok"; origin: string }
 > => {
-  const result = await vcIssuer.getDerivationOrigin({ origin: issuerOrigin });
+  const derivationOriginResult = await vcIssuer.getDerivationOrigin({
```
frederikrothenberger, line 368: I took the liberty of renaming a few variables in this function because I think it makes following the logic a bit easier.

przydatek (commented):
LGTM for the logic, thanks!
(I suppose @lmuntaner will take a look also on TS-changes)

lmuntaner (approved):
LGTM, thanks!

PR #2365, opened 2024-03-14
Remove Promise.race() in e2e test

frederikrothenberger:
This PR refactors the e2e tests to not use `Promise.race`. This  `Promise.race` is problematic because one of the Promises keeps polling the chromedriver until the timeout is reached.

The observed flakiness of the 
"Register with PIN then log into client application" test (due to chromedriver becoming unresponsive, presumably crashing) during polling when waiting for auth to finish seems to be reduced with this change.


lmuntaner (approved):
I like the change, it's more readable as well.

PR #2371, opened 2024-03-18
Mention proposal payload in hash verification

frederikrothenberger:
This change adds a line to the hash verification instructions to remind people to also check the proposal payload.


przydatek (approved):
LGTM, thanks!

PR #2373, opened 2024-03-20
Upgrade agent-js

### LINKED ISSUE #2235: Failures when using password manager Bitwarden to to store passkeys
The password manager Bitwarden has the ability to store and authenticate using FIDO2-compliant passkeys.  However, when creating a passkey to register an identity on `identity.internetcomputer.org`, after the passkey is stored in Bitwarden, the website returns the following error message (as reported [here](https://community.bitwarden.com/t/unable-to-create-passkeys/62644)):

`Failed to create identity: An error occurred during construction: Error: Could not create credentials.`

Moreover, if an identity has been registered by storing the passkey on a Yubikey, then attempting to create a "Recovery Device" and storing the recovery passkey in Bitwarden fails silently (the passkey is created and stored in Bitwarden, but does not register as a "Recovery Device" on identity.internetcomputer.org).

Furthermore, if an identity has been registered by storing the passkey on a Yubikey, then attempting to use <kbd>Add new passkey</kbd> fails with the following error message after the FIDO2 passkey is stored in Bitwarden:

```
Failed to add new device
We failed to add the new device to this Identity Anchor. Please try again.
Error details:
Could not create credentials.
```

frederikrothenberger:
This PR upgrades `agent-js` to the latest version to fix a bug related to WebAuthn registrations.
Fixes #2235.


demos/test-app/src/index.tsx
```diff
@@ -347,11 +347,13 @@ init();
 
 whoamiBtn.addEventListener("click", async () => {
   const canisterId = Principal.fromText(readCanisterId());
+  let agent = new HttpAgent({
+    host: hostUrlEl.value,
+    identity: delegationIdentity,
+  });
+  await agent.fetchRootKey();
```
frederikrothenberger, line 354: Fetching the root key is now necessary because the agent now verifies the replica signature on queries too.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -43,7 +43,6 @@ import {
 } from "@dfinity/identity";
 import { Principal } from "@dfinity/principal";
 import { isNullish, nonNullish } from "@dfinity/utils";
-import * as tweetnacl from "tweetnacl";
```
lmuntaner, line 46: Wasn't this in the dependencies of package.json?

lmuntaner (approved):
Just a questoin, but already approving, thanks!

demos/test-app/src/index.tsx
```diff
@@ -347,11 +347,13 @@ init();
 
 whoamiBtn.addEventListener("click", async () => {
   const canisterId = Principal.fromText(readCanisterId());
+  let agent = new HttpAgent({
```
peterpeterparker, line 350: `const`

src/frontend/src/utils/iiConnection.ts
```diff
@@ -719,7 +718,7 @@ export const creationOptions = (
       name: "Internet Identity Service",
     },
     user: {
-      id: tweetnacl.randomBytes(16),
+      id: window.crypto.getRandomValues(new Uint8Array(16)),
```
peterpeterparker, line 721: Wasn't the underlying implementation of tweetnacl adding some complexity to `new Unit8Array(16)` or it's enough?

https://github.com/dchest/tweetnacl-js/blob/fecde6ecf0eb81e31d54ca0509531ab1b825f490/nacl.js#L1160

src/frontend/src/crypto/ed25519.test.ts
```diff
@@ -66,9 +66,9 @@ const testVectorsSLIP10 = [
 test("derive Ed25519 via SLIP 0010", async () => {
   await Promise.all(
     testVectorsSLIP10.map(async (testVector: TestVector) => {
-      const seedBlob = fromHexString(testVector.seed);
-      const expectedPrivateKey = fromHexString(testVector.privateKey);
-      const expectedPublicKey = fromHexString(testVector.publicKey);
+      const seedBlob = fromHex(testVector.seed);
```
peterpeterparker, line 69: Had a quick look to `fromHex` which seems a bit less efficient than `fromHexString`. Of course it's relative, it's just an additional loop but, out of curiosity any reason to change those?

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: It was transitively available and never accurately reflected in the `package.json`.

src/frontend/src/crypto/ed25519.test.ts
frederikrothenberger: Yes, `fromHexString` is no longer available. And this is what the `agent-js` internal calls were changed to too, so I figured that this is the official replacement.

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: This does not add complexity. It just invokes the browser API in batches of 65536 bytes because that is the limit of the browser API.
But we don't need that here.

src/frontend/src/crypto/ed25519.test.ts
peterpeterparker: Actually `fromHexString` is still available in `@dfinity/candid`. You used to include it using an incorrect path `@dfinity/identity/lib/cjs/buffer` which I guess already referenced the inherited implemented (just a guess). From having a look at the compiled code, it seems to confirm it so, long story short, if you just want to adapt the code without modifying the feature, I've got the feeling that you should stick to `fromHexString` but imported from `@dfinity/candid`.

That said, `fromHex` might work, just pointing out that those are two different implementations.

src/frontend/src/utils/iiConnection.ts
peterpeterparker: Gotcha, thanks for the feedback.

src/frontend/src/crypto/ed25519.test.ts
frederikrothenberger: Ok, thanks. I switched to the `candid` implementation now.

peterpeterparker (approved):
LGTM, thx

PR #2374, opened 2024-03-21
Allow longer credential_id values to accommodate Titan security key

frederikrothenberger:
As per [report from a user](https://forum.dfinity.org/t/error-when-adding-fido-device/28523/4), the [Titan security key](https://store.google.com/us/product/titan_security_key) offered by Google generates `credential_id` values of 304 bytes. This PR increases the limit of that field up to 350 bytes to accommodate it. However, the limits for a single identity in general have not been raised. This means that the potential for abuse is negligible as users now just have more flexibility _how_ to allocate their space among the different fields.


przydatek (approved):
LGTM, thanks!

PR #2377, opened 2024-03-22
Refactor add device success screen to be shown consistently

frederikrothenberger:
This PR refactors the add device success screen to be shown consistently (regardless from where the flow is started) by making it part of the tentative device flow rather than having that logic in the top-level `index.ts` file.

Given the screen is part of both sides of the flow (the new device _and_ the existing one) I moved the corresponding files out of the `manage` folder).


lmuntaner (approved):
Nice! LGTM!

PR #2378, opened 2024-03-22
Remove II's direct dependency on identity.rs library for handling VCs

przydatek:
Until now II's code was depending on a fork of `identity.rs` in two ways: directly, to create id-alias VCs, and indirectly, via the dependency on `vc_util`-crate.  This PR removes the first (direct) dependency, by extending `vc_util`, and changing the format of the id-alias VCs, to follow [the recommended convention](https://github.com/dfinity/internet-identity/blob/main/docs/vc-spec.md#recommended-convention-connecting-credential-specification-with-the-returned-credentials) for general VCs.  This completes the removal of direct dependency on `identity.rs` for II and VC-issuers -- now the dependency is encapsulated in `vc_util`-crate.

NOTE: a side-effect of this PR is that once it is deployed in production, II will start issuing id-alias VCs in the new format, so any dapp that verifies these VCs should be updated as well.  If a dapp is using `vc_util` to verify these VCs, it should be re-compiled with the updated version  of `vc_util`.


src/vc_util/src/lib.rs
```diff
@@ -94,6 +98,44 @@ pub fn vc_jwt_to_jws(
...
+    let header = parsed_signing_input
+        .protected_header()
+        .expect("internal: failed getting protected header");
+
+    let encoder: CompactJwsEncoder = CompactJwsEncoder::new(parsed_signing_input.claims(), header)
+        .map_err(|e| format!("internal: failed creating JWS encoder: {:?}", e))?;
+    Ok(encoder.into_jws(sig))
+}
+
+/// Extracts the canister signature public key from the given signing_input.
```
frederikrothenberger, line 121: Maybe expand the comment a bit on what `signing input` is? I.e. a JWT without signature.

src/internet_identity/Cargo.toml
```diff
@@ -53,6 +49,7 @@ canister_tests.workspace = true
 hex-literal = "0.4"
 regex.workspace = true
 ic-response-verification.workspace = true
+identity_jose = { git = "https://github.com/frederikrothenberger/identity.rs.git", branch = "frederik/wasm-test", default-features = false, features = ["iccs"]}
```
frederikrothenberger, line 52: Ideally, this would be removed too. Is the infrastructure provided by `vc_util` not good enough for the tests?

frederikrothenberger (approved):
LGTM, thanks. Two minor comments.

src/vc_util/src/lib.rs
przydatek: Done.

src/internet_identity/Cargo.toml
przydatek: `vc_util` doesn't have currently the corresponding functionality.  I'll add it in a follow-up PR.

przydatek (commented):
Thanks!

PR #2379, opened 2024-03-22
Use ic_representation_independent_hash instead of a copied hash code

przydatek:
`ic_representation_independent_hash`-crate from version 2.5 supports hashing of arrays, so we can now removed a copy of hashing code added previously, and instead use the crate.


frederikrothenberger (approved):
Thanks, LGTM!

PR #2386, opened 2024-03-26
Move derivation origin validation into utils

frederikrothenberger:
This moves the derivation origin validation into utils as it is used both in the authorize and the VC flow.
This is PR will be followed up with a refactoring to bring the canister id resolution already present in the VC flow to the authorize flow as well.


lmuntaner (approved):
Thanks for the small PRs

PR #2387, opened 2024-03-26
Allow special characters in passkey alias

frederikrothenberger:
This PR loosens the restrictions on passkey aliases to allow for special characters.


src/frontend/src/utils/validateAlias.ts
```diff
@@ -1,29 +0,0 @@
-// Regex Pattern for input: All characters, must be alphabet or number
-// Can have hyphen(s), space(s) or underscore(s) in the middle.
-// Good examples: "2019_macbook", "2019-Macbook", "2019 Macbook"
-// Bad examples: "2019 macbook!", "2010 macbook_", "space trails at end "
-export const validateAlias = (
```
frederikrothenberger, line 5: Given this got even simpler and was not used anywhere else I moved the function into the alias component.

lmuntaner (approved):
I guess there are no issues in the backend with special characters, right?

Are spaces at the end trimmed? Should we do so?

It might be confusing if a user adds white spaces at the end and then forgets about it...

frederikrothenberger:
> I guess there are no issues in the backend with special characters, right?
> 
> Are spaces at the end trimmed? Should we do so?
> 
> It might be confusing if a user adds white spaces at the end and then forgets about it...

1. Yes, the back-end does not care
2. Spaces are trimmed (front and back). The alternative would be to show an error and have the user manually trim the value, but honestly that is just tedious. There is a check to make sure that the value is not all white space (and gets trimmed to empty). Given it is only used as a display value, I highly doubt users care (or even think about intentionally) about trailing white space.

PR #2390, opened 2024-03-27
Provide error message for files unsupported by asset_util

przydatek:
`asset_util` does not support files without extensions, and this PR provides information about offending files in a case of a failure.

### DIFF
--- a/src/asset_util/src/lib.rs
+++ b/src/asset_util/src/lib.rs
@@ -463,7 +463,14 @@ fn collect_assets_rec(dir: &Dir, assets: &mut Vec<Asset>) {
 /// Otherwise the content type is determined by the text after the last dot in the file name,
 /// and the encoding is `ContentEncoding::Identity`.
 fn content_type_and_encoding(asset_path: &Path) -> (ContentType, ContentEncoding) {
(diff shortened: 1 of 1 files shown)


frederikrothenberger (approved):
LGTM, thanks!

peterpeterparker:
Out of curiosity, will this message just end up beding displayed ash "Body does not pass certification" or mainnet, or do you have a way in II to certify error messages on the fly?

frederikrothenberger:
> Out of curiosity, will this message just end up beding displayed as "Body does not pass certification" or mainnet, or do you have a way in II to certify error messages on the fly?

This is the message that will be printed as II panicks during `init` or `post_upgrade`. 😉

frederikrothenberger:
I think it is more relevant for other devs that start reusing the `asset_util` crate.

peterpeterparker:
> This is the message that will be printed as II panicks during...

Gotcha! 👍

> I think it is more relevant for other devs...

If the goal is other devs to use it, maybe worth extending `asset_util` to support file without extension bc that's quite a common use case particularly for `.well-known` file. Just saying out at loud, thanks for the feedback!

PR #2391, opened 2024-03-27
Rename id_alias-VC-related fields

przydatek:
Rename id_alias-VC fields in `PreparedIdAlias` and `GetIdAliasRequest` to match their actual contents.  The renamed fields contained initially JWT-values, but later the contents changed to JWS signing input, so the new names match the current content.


lmuntaner:
Not needed anymore

PR #2398, opened 2024-04-02
Enable serving of well-known assets that don't have extensions

przydatek:
`asset_util` requires that served files do have extensions (so that the files' type & encoding can be established). However, some special, well-known files by definition have no extensions, so to enable serving such files, a "white-list" of known files is added.


frederikrothenberger (approved):
LGTM, thanks!

PR #2399, opened 2024-04-03
Allow custom domains to use alternative origins

frederikrothenberger:
This PR lifts the restriction on the alternative origins feature to only allow canister id based subdomains of `ic0.app` and `icp0.io`.

The canister id resolution is now the same for the session authorization flow as it is for verifiable credentials.

The following restriction is still in place: the alternative origins file needs to be served from a canister accessible on the `icp0.io` domain.

Allowances are made for local development using `localhost`, `0.0.0.0`, `127.0.0.1` and custom HTTP gateways.

This PR also adds a unit test suite for the derivation origin validation.


src/frontend/src/utils/validateDerivationOrigin.ts
```diff
@@ -1,12 +1,8 @@
+import { resolveCanisterId as resolveCanisterIdFn } from "$src/utils/canisterIdResolution";
```
lmuntaner, line 1: Did you do this to be able to test the function?

src/frontend/src/utils/validateDerivationOrigin.test.ts
```diff
@@ -0,0 +1,228 @@
...
+  setupMocks({
+    iiUrl: "https://identity.ic0.app",
+    response: Response.json({
+      notAlternativeOrigins: ["https://example.com"],
+    }),
+  });
+
+  const result = await validateDerivationOrigin({
+    requestOrigin: "https://example.com",
+    derivationOrigin: `https://some-url.com`, // different from requestOrigin so that we need to fetch the alternative origins
```
lmuntaner, line 177: Why is it using backticks instead of double quotes? Isn't prettier installed in the repo?

src/frontend/src/utils/validateDerivationOrigin.test.ts
```diff
@@ -0,0 +1,228 @@
...
+
+test("should not validate on alternative origins redirect", async () => {
+  setupMocks({
+    iiUrl: "https://identity.ic0.app",
+    response: Response.redirect("https://some-evil-url.com"),
+  });
+
+  const result = await validateDerivationOrigin({
+    requestOrigin: "https://example.com",
+    derivationOrigin: `https://some-url.com`, // different from requestOrigin so that we need to fetch the alternative origins
```
lmuntaner, line 192: ditto back ticks

src/frontend/src/utils/validateDerivationOrigin.test.ts
```diff
@@ -0,0 +1,228 @@
...
+
+test("should not validate on alternative origins error", async () => {
+  setupMocks({
+    iiUrl: "https://identity.ic0.app",
+    response: new Response(undefined, { status: 404 }),
+  });
+
+  const result = await validateDerivationOrigin({
+    requestOrigin: "https://example.com",
+    derivationOrigin: `https://some-url.com`, // different from requestOrigin so that we need to fetch the alternative origins
```
lmuntaner, line 207: ditto

lmuntaner (approved):
LGTM! Thanks!

src/frontend/src/utils/validateDerivationOrigin.ts
frederikrothenberger: Yes. I could also provide a wrapper function so that caller don't need to provide the `canisterIdResolution` as a parameter. Should I do that / change something here?

src/frontend/src/utils/validateDerivationOrigin.test.ts
frederikrothenberger: We do have `prettier` but it doesn't care. Feel free to change the config. I'll fix the backticks here.

src/frontend/src/utils/validateDerivationOrigin.test.ts
frederikrothenberger: dito

src/frontend/src/utils/validateDerivationOrigin.ts
lmuntaner: Maybe you could add the imported function as default parameter and substitute it in the tests.

This way you don't need to change the calls anywhere.

PR #2404, opened 2024-04-04
Handle error when parsing origin for canister id resolution

frederikrothenberger:
This PR improves error handling when passing invalid origins for canister id resolution.


lmuntaner (approved):
Thanks!

PR #2405, opened 2024-04-04
chore: update content for mdx v3 parser

dfx-json:
This PR makes some minor modifications to a couple of markdown files to support the [migration to docusaurus v3](https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3#try-docusaurus-v3-today) in the portal. v3 of Docusaurus uses mdx v3, which has different parsing rules for markdown.

PR #2409, opened 2024-04-05
chore: update content for mdx v3 parser

dfx-json:
Additional updates as a continuation of https://github.com/dfinity/internet-identity/pull/2405


dfx-json:
@frederikrothenberger this may just be a false positive then. I ran the tool that docusaurus provides to check compatibility `npx docusaurus-mdx-checker` and this file was flagged. but if it's excluded from the docusaurus compilation then no need to merge this PR

frederikrothenberger:
Ok, thanks --> closed.

PR #2414, opened 2024-04-10
Manage and improve VC flow edge cases and fix relying party and issuer card

lmuntaner:
The main motivation is to have the edge cases of the VC flow more under control.

In this PR:
* Changed on showcase page and added two more to control how the page should look like in different edge cases.
* Changed the error page title requested by product. More copy changes might come in this page, but I'm waiting for confirmation.

<hr/><details><summary>🟢 Some screens were added</summary></details><details><summary>🟡 Some screens were changed</summary></details><details><summary>🔴 Some screens were removed</summary></details>

PR #2416, opened 2024-04-10
Add issuer and RP expect in VC e2e flow

lmuntaner:
Yesterday I introduced a bug where the relying party was in the issuer card and viceversa. This morning I fixed it in #2414 but I didn't add a test for it.

In this PR, I introduce two expects that that would have caught the bug.


frederikrothenberger (approved):
Thanks, LGTM!

PR #2424, opened 2024-04-12
Simplify using-dev-build tests

frederikrothenberger:
This PR simplifies the tests of the `using-dev-build` demo by no longer using the `dfx-proxy` server. Instead, we rely on the fact that the tests use Chrome and simply use the canisterId base localhost URLs.

The readme is adjusted accordingly.

### DIFF
--- a/demos/using-dev-build/README.md
+++ b/demos/using-dev-build/README.md
@@ -138,7 +138,6 @@ _If you actually use the webapp, make sure that the "Internet Identity URL" fiel
 Figuring the canister IDs, and using the `canisterId=...` query parameter is all a bit cumbersome. Here are some commands you might like:
 
 - `npm run dev`: Build the app and serve it on `localhost:5173` with hot reload on code changes, ideal for hacking on the webapp.
-- `npm run proxy`: Start a proxy that serves Internet Identity on `localhost:8086` and the webapp on `localhost:8087` for easy access.
-- `npm run test`: Start the proxy and run browser tests against the `internet_identity` canister.
+- `npm run test`: Run browser tests against the `internet_identity` canister.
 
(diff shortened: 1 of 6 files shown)


demos/using-dev-build/README.md
```diff
@@ -138,7 +138,6 @@ _If you actually use the webapp, make sure that the "Internet Identity URL" fiel
 Figuring the canister IDs, and using the `canisterId=...` query parameter is all a bit cumbersome. Here are some commands you might like:
 
 - `npm run dev`: Build the app and serve it on `localhost:5173` with hot reload on code changes, ideal for hacking on the webapp.
-- `npm run proxy`: Start a proxy that serves Internet Identity on `localhost:8086` and the webapp on `localhost:8087` for easy access.
-- `npm run test`: Start the proxy and run browser tests against the `internet_identity` canister.
+- `npm run test`: Run browser tests against the `internet_identity` canister.
```
lmuntaner, line 141: Do you want this section within the `## Helpers` section?

I think that this section might not make sense anymore and we could have two different sectoins: `## Local Development` and `## Testing`. What do you think?

demos/using-dev-build/package.json
```diff
@@ -13,13 +13,14 @@
     "proxy": "npm run opts -- ./test.ts --no-run"
```
lmuntaner, line 13: Do you still need this script?

lmuntaner (approved):
A couple of comments but already approving, thanks!

I wonder whether we could use a more common wdio configuration file instead of having to use a custom file to execute it.

demos/using-dev-build/package.json
frederikrothenberger: No, good catch. Thanks!

demos/using-dev-build/README.md
frederikrothenberger: Yes, changed.

frederikrothenberger:
> I wonder whether we could use a more common wdio configuration file instead of having to use a custom file to execute it.

Yeah, I think we should switch to that in a follow-up PR.

PR #2431, opened 2024-04-15
Increase archive buffer memory limit further

frederikrothenberger:
This PR increases the archive buffer limit further to about ~8GB to meet current demand.


przydatek (approved):
LGTM, thanks!

PR #2432, opened 2024-04-16
Exclude dapps without website entry from dapps explorer

frederikrothenberger:
The dapps explorer is meant for deployed applications that users can visit. If a dapp does not have a website to visit, it will be filtered out from the list included in II.

This fixes automatic dapps update (#2426) which fails due to a missing `website` entry.

### DIFF
--- a/scripts/update-dapps
+++ b/scripts/update-dapps
@@ -83,6 +83,7 @@ function update() {
             | select(.usesInternetIdentity == true) # ensure the dapp actually uses II
             | select(.name != "Internet Identity") # skip II itself
             | select(.logo != null) # only use dapps with logo
+            | select(.website != null) # only use dapps with website
             | .logo |= sub("^/" + $portal_logo_prefix + "/"; "") # swap the logo for our prefix
             | with_entries( # drop entries we do not care about to reduce size
                 select(


lmuntaner (approved):
LGTM

lmuntaner:
Question though, why don't we filter them in the code? Maybe we find other UX for the ones without a website, no?

frederikrothenberger:
> Question though, why don't we filter them in the code? Maybe we find other UX for the ones without a website, no?

We don't have arbitrary amount of space for dapps. That's why we have so far focused on dapps that you can sign into with II. If they don't have a website, then naturally that's not the case.

lmuntaner:
> > Question though, why don't we filter them in the code? Maybe we find other UX for the ones without a website, no?
> 
> We don't have arbitrary amount of space for dapps. That's why we have so far focused on dapps that you can sign into with II. If they don't have a website, then naturally that's not the case.

Ah, that makes sense.

lmuntaner:
> > Question though, why don't we filter them in the code? Maybe we find other UX for the ones without a website, no?
> 
> We don't have arbitrary amount of space for dapps. That's why we have so far focused on dapps that you can sign into with II. If they don't have a website, then naturally that's not the case.

Ah, that makes sense.

PR #2433, opened 2024-04-16
Remove IdentityBackground class, use string instead

lmuntaner:
Main motivation is to remove the IdentityBackground class because it increases more the complexity than the value it adds.

The problem is that importing an asset in the showcase (Astro) and the application means different types of the imported value.

Before, we had a workaround that another instance of the class was created for the showcase and for the application.

Now, `identityCard` component expects a string and the image is imported directly in the showcase page or in the application code.


src/frontend/src/flows/register/finish.ts
```diff
@@ -119,7 +116,6 @@ export const displayUserNumberWarmup = (): OmitParams<
   typeof displayUserNumber,
   "identityBackground"
 > => {
-  const identityBackground = loadIdentityBackground();
```
frederikrothenberger, line 122: This would have triggered some preloading of the image in the client, right?
So presumably, the image is now loaded later. Is this noticeable for the clients?

lmuntaner: Good point. It might be noticeable. I didn't realize this functionality. Thanks for bringing it up.

Let me close this PR and I'll fix the issue in another way.

PR #2434, opened 2024-04-16
Add PreLoadImage class to substitute IdentityBackground

lmuntaner:
Main motivation is to remove the IdentityBackground class because it increases more the complexity than the value it adds.

The problem is that importing an asset in the showcase (Astro) and the application means different types of the imported value.

Before, we had a workaround that another instance of the class was created for the showcase and for the application.

IdentityBackground had two functionalities:
* Ensure only one image is used.
* Preload the image.

This PR introduces a new class `PreLoadImage` that has only the preloading functionality and expects an image as input.

This allows us to use the image differently in Astro and app environments.


src/showcase/src/pages/displayManage.astro
```diff
@@ -6,10 +6,13 @@ import Screen from "../layouts/Screen.astro";
   <script>
     import { toast } from "$src/components/toast";
     import { userNumber } from "../constants";
-    import { identityBackground } from "../backgrounds";
     import { dapps } from "../constants";
     import { displayManagePage } from "$src/flows/manage";
     import { html } from "lit-html";
+    import identityCardBackground from "$src/assets/identityCardBackground.png";
```
lmuntaner, line 12: Importing it here we ensure that it doesn't get mixed with the application code and we don't need to cast types.

Ideally, we will move the showcase to its own library with its own package.json and then we will be able to separate import statements better.

src/showcase/src/backgrounds.ts
```diff
@@ -1,8 +0,0 @@
-import { IdentityBackground } from "$src/components/identityCard";
-import backgroundImage from "$src/components/identityCard.png";
-import { ImageMetadata } from "astro";
-
-export const identityBackground = IdentityBackground.getSingleton(
-  // When running the build, this is also built but not in an Astro environment
-  (backgroundImage as unknown as ImageMetadata).src
```
lmuntaner, line 7: I wanted to remove this ugly line.

lmuntaner:
@frederikrothenberger let me know what you think now.

frederikrothenberger (approved):
Nice, lgtm!

PR #2435, opened 2024-04-17
Simplify archive integration tests

frederikrothenberger:
This PR removes the abstractions we introduced to simultaneously test multiple II memory layouts. It is no longer needed and makes the test more cumbersome.

This PR will facilitate upcoming changes to this test suite.


przydatek (approved):
LGTM, thanks!

PR #2425, opened 2024-04-12
Update dfx to version 0.19.0

frederikrothenberger:
Updates `dfx` in the II repository. This has the advantage that `blob` hex bytes no longer need to be preceded by `\`.

<hr/><details><summary>🟡 Some screens were changed</summary></details>

PR #2436, opened 2024-04-17
Revert erroneous candid blob encoding change

frederikrothenberger:
In #2425 a change was introduced assuming that `blob` types are handled differently with `dfx` 0.19.0. However, this change only affected printing hex values and not encoding. Hence the change introduced a bug that is reverted with this PR.


lmuntaner (approved):
Thanks!

PR #2440, opened 2024-04-22
Upgrade stable structures

frederikrothenberger:
This upgrades the `ic-stable-structures` crate to fix a memory leak in the `BTreeMap`.


lmuntaner (approved):
Thanks!

PR #2445, opened 2024-04-25
Improve error messages for passkey errors

frederikrothenberger:
We have users complaining about errors that are hard to diagnose because we don't know the source (e.g. "internal error").

This PR adds helpful context to errors thrown out of passkey interactions.


lmuntaner (approved):
Thanks!

PR #2446, opened 2024-04-26
Upgrade agent-js

frederikrothenberger:
This PR upgrades `agent-js` to the latest version because it includes an improvement to the errors thrown when failing to decode cbor.


lmuntaner (approved):
Thanks!

peterpeterparker:
If interesting, in ic-js I did not upgraded yet agent-js because an issue was reported:

https://github.com/dfinity/ic-js/pull/593#issuecomment-2033637984

I asked @krpeacock if the related patch would be merged soon but, it seems to not have been patched yet so, I'm still waiting.

Don't know though if the issue is a blocker or a race condition in some particular use case.

frederikrothenberger:
@peterpeterparker: Thanks for your input. I'll merge this and then set the retry count to 10 in the agent, which seems to fix this (albeit in an inelegant way).

krpeacock:
Yeah, it's taken me longer than I intended, and I still don't have it working against prod. Having the conference come up didn't help either

PR #2447, opened 2024-04-26
Configure up to 10 retries on agent failure

frederikrothenberger:
This PR configures the number of retries to 10 (as opposed to just 3, which is the default). This makes the agent more resilient against watermark check failures, because the likelihood of hitting only replicas that are behind gets lowered a lot and retrying takes time too.

Addresses [this comment](https://github.com/dfinity/internet-identity/pull/2446#issuecomment-2078795472).


lmuntaner (approved):
Thanks!

PR #2449, opened 2024-04-29
Collect anonymous prepare_delegation events for stats purposes

frederikrothenberger:
This PR adds storage to stable memory to keep anonymous(!) events about `prepare_delegation` calls. This will allow II to create the following stats:
* number of sign-ins per dapp per ii domain for the last
  * 24h
  * 30d
* cumulative seconds of session time per dapp per ii domain for the last
  * 24h
  * 30d

The latter is necessary to have some comparability between dapps that use a different session timeout.

The stats are not yet exposed by II. In subsequent PRs:
1. the old collection of observed front-end origins will be removed
2. the new data will be exposed
3. integration tests will be added
4. additional performance metrics will be collected in order to better judge the impact it has on the `prepare_delegation` call.

### DIFF
--- a/src/internet_identity/src/activity_stats/event_stats.rs
+++ b/src/internet_identity/src/activity_stats/event_stats.rs
@@ -31,10 +31,11 @@
 //! Example of event_aggregations:
 //! ```
 //! event_aggregations: {
-//!   "PD_count_24h_ic0.app_https://dapp.example.com": 5,
-//!   "PD_sess_sec_24h_ic0.app_https://dapp.example.com": 15,
-//!   "PD_count_30d_internetcomputer.org_https://anotherapp.example.com": 20,
-//!   "PD_sess_sec_30d_internetcomputer.org_https://anotherapp.example.com": 60
+//!   "PD_count>24h>ic0.app>https://dapp.example.com": 5,
+//!   "PD_sess_sec>24h>ic0.app>https://dapp.example.com": 15,
+//!   "PD_count>30d>internetcomputer.org>https://anotherapp.example.com": 20,
+//!   "PD_sess_sec>30d>internetcomputer.org>https://anotherapp.example.com": 60
+//!   //...
 //! }
 //! ```
 
--- a/src/internet_identity/src/activity_stats/event_stats.rs
+++ b/src/internet_identity/src/activity_stats/event_stats.rs
@@ -43,7 +43,7 @@ use crate::activity_stats::event_stats::event_aggregations::AGGREGATIONS;
 use crate::ii_domain::IIDomain;
 use crate::state::storage_borrow_mut;
 use crate::storage::Storage;
-use crate::DAY_NS;
+use crate::{state, DAY_NS};
 use ic_cdk::api::time;
 use ic_stable_structures::storable::Bound;
 use ic_stable_structures::{Memory, StableBTreeMap, Storable};
@@ -57,6 +57,34 @@ mod event_aggregations;
 
 #[cfg(test)]
 mod tests;
+
+#[derive(Deserialize, Serialize, Clone, Eq, PartialEq, Debug, Ord, PartialOrd)]
+pub struct EventKey {
+    /// Timestamp of the event.
+    pub time: Timestamp,
+    /// A counter providing uniqueness for events with the same timestamp.
+    pub counter: u16,
+}
+
+impl EventKey {
+    pub fn new(time: Timestamp) -> Self {
+        Self {
+            time,
+            counter: state::get_and_inc_event_data_counter(),
+        }
+    }
+
+    pub fn min_key(time: Timestamp) -> Self {
+        Self { time, counter: 0 }
+    }
+
+    pub fn max_key(time: Timestamp) -> Self {
+        Self {
+            time,
+            counter: u16::MAX,
+        }
+    }
+}
 
 #[derive(Deserialize, Serialize, Clone, Eq, PartialEq, Debug)]
 pub struct EventData {
@@ -73,6 +101,32 @@ pub struct PrepareDelegationEvent {
     pub ii_domain: Option<IIDomain>,
     pub frontend: FrontendHostname,
     pub session_duration_ns: u64,
+}
+
+const EVENT_KEY_SIZE: usize = 10;
+impl Storable for EventKey {
+    fn to_bytes(&self) -> Cow<[u8]> {
+        let mut buf = Vec::with_capacity(EVENT_KEY_SIZE);
+        buf.extend(self.time.to_be_bytes());
+        buf.extend(self.counter.to_be_bytes());
+        Cow::Owned(buf)
+    }
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        Self {
+            time: u64::from_be_bytes(
+                TryFrom::try_from(&bytes[0..8]).expect("failed to read event key timestamp"),
+            ),
+            counter: u16::from_be_bytes(
+                TryFrom::try_from(&bytes[8..10]).expect("failed to read event key counter"),
+            ),
+        }
+    }
+
+    const BOUND: Bound = Bound::Bounded {
+        is_fixed_size: true,
+        max_size: EVENT_KEY_SIZE as u32,
+    };
 }
 
 impl Storable for EventData {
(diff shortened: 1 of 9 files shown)


src/internet_identity/src/activity_stats/event_stats/event_aggregations.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Type definition for an aggregation function.
+///
+/// Each aggregation is defined as a function that takes a bucket size label (i.e. "24h" or "30d")
+/// and a tuple of `(timestamp, event_data)` and maps it to an optional [WeightedAggregation]
+/// (i.e. a key for the aggregation and a weight for the event).
+/// If the event is not part of the aggregation, the function should return [None].
+type Aggregation = fn(&str, &(u64, EventData)) -> Option<WeightedAggregation>;
+
+/// All aggregations currently maintained over event data.
+pub const AGGREGATIONS: [Aggregation; 2] =
```
lmuntaner, line 16: Sorry, but I don't understand this part. Is this a list of two functions?

src/internet_identity/src/activity_stats/event_stats.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        serde_cbor::from_slice(&bytes).unwrap()
+    }
+
+    const BOUND: Bound = Bound::Unbounded;
+}
+
+struct WeightedAggregation {
+    key: String,
```
lmuntaner, line 50: Does it make sense to add a comment of how this key looks like and how it's built?

src/internet_identity/src/activity_stats/event_stats/event_aggregations.rs
```diff
@@ -0,0 +1,80 @@
...
+            event,
+        )
+    };
+
+fn prepare_delegation_aggregation(
+    key_prefix: &str,
+    key_duration: &str,
+    weight_func: fn(&PrepareDelegationEvent) -> u64,
+    event: &EventData,
+) -> Option<WeightedAggregation> {
```
lmuntaner, line 56: Why does this return an `Option<WeightedAggregation>`? I don't see it returning `None` in any case.

src/internet_identity/src/activity_stats/event_stats.rs
```diff
@@ -0,0 +1,200 @@
...
+    new_event: EventData,
+    pruned_events: &[(Timestamp, EventData)],
+    db: &mut StableBTreeMap<String, u64, M>,
+) where
+    F: Fn(&(u64, EventData)) -> Option<WeightedAggregation>,
+{
+    // Process the pruned events to calculate the pruned weight by aggregation key
+    let mut pruned_weight_by_key = pruned_events
+        .iter()
+        .filter_map(&aggregation_filter_map)
```
lmuntaner, line 150: Why a `filter_map` and not a `map`?

src/internet_identity/src/activity_stats/event_stats/event_aggregations.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// Aggregates the count of prepare delegation events
+/// - per dapp frontend
+/// - per II domain
+/// - over 24h and 30d buckets
+///
+/// Results in labels like "PD_count_<bucket_length>_<ii_domain>_<frontend_hostname>"
+/// e.g "PD_count_24h_ic0.app_https://dapp.example.com"
+const PREPARE_DELEGATION_COUNT: fn(&str, &(u64, EventData)) -> Option<WeightedAggregation> =
+    |key_duration: &str, (_timestamp, event): &(u64, EventData)| -> Option<WeightedAggregation> {
```
lmuntaner, line 27: Sorry, but I don't see this `_timestamp` being used. Why is it part of the parameters?

src/internet_identity/src/activity_stats/event_stats/tests.rs
```diff
@@ -0,0 +1,364 @@
...
+use std::cell::RefCell;
+use std::rc::Rc;
+use std::time::Duration;
+
+const TIMESTAMP: u64 = 1_714_387_451 * 10u64.pow(9);
+const EXAMPLE_URL: &str = "https://example.com";
+const EXAMPLE_URL_2: &str = "https://other-example.com";
+
+#[test]
+fn should_store_event_and_add_to_aggregations() {
```
lmuntaner, line 17: Would it be useful to have a test with different frontend hostnames to check that the buckets are created properly per frontend hostname?

src/internet_identity/src/activity_stats/event_stats/tests.rs
```diff
@@ -0,0 +1,364 @@
...
+        storage
+            .event_aggregations
+            .get(&"PD_sess_sec_24h_ic0.app_https://example.com".to_string())
+            .unwrap(),
+        1
+    );
+}
+
+#[test]
+fn should_remove_zero_weighted_aggregations() {
```
lmuntaner, line 312: Sorry, but I don't understand what is being tested here. Why are the aggregations for example.com removed?

Is it because the next events added are older than 30 days?

src/internet_identity/src/activity_stats/event_stats/tests.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    let event2 = EventData {
+        event: Event::PrepareDelegation(PrepareDelegationEvent {
+            ii_domain: Some(IIDomain::Ic0AppDomain),
+            frontend: EXAMPLE_URL_2.to_string(),
+            session_duration_ns: Duration::from_secs(900).as_nanos() as u64,
+        }),
+    };
+
+    // after this update, the previous aggregations should be removed as their weight is 0
```
lmuntaner, line 343: What does it mean that the "weight is 0"?

src/internet_identity/src/activity_stats/event_stats/tests.rs
```diff
@@ -0,0 +1,364 @@
...
+    update_events_internal(event2, TIMESTAMP + 30 * DAY_NS, &mut storage);
```
lmuntaner, line 344: Should this 30 be part of the name of the function maybe? To check that events are pruned after 30 days.

src/internet_identity/src/activity_stats/event_stats.rs
lmuntaner, line 1: I believe it would be nice to have an example of how the storage data looks like:
```
event_data: {
  "1640995200000": {
    "event": {
      "PrepareDelegation": {
        "ii_domain": "Ic0AppDomain",
        "frontend": "https://dapp.example.com",
        "session_duration_ns": 3000000000
      }
    }
  },
  "1640998800000": {
    "event": {
      "PrepareDelegation": {
        "ii_domain": "InternetComputerOrgDomain",
        "frontend": "https://anotherapp.example.com",
        "session_duration_ns": 4500000000
      }
    }
  },
  //...
},
event_aggregations: {
  "PD_count_24h_ic0.app_https://dapp.example.com": 5,
  "PD_sess_sec_24h_ic0.app_https://dapp.example.com": 15,
  "PD_count_30d_internetcomputer.org_https://anotherapp.example.com": 20,
  "PD_sess_sec_30d_internetcomputer.org_https://anotherapp.example.com": 60
}
```

lmuntaner (approved):
A few questions and comments, but I'm rubber stamping because I don't see a blocker.

My comments are more about the design than Rust. I leave that to you.

Btw, I realized that since both are not so familiar with the other's language (specially me), should we first write how we are going to implement a feature in pseudo-code and discuss that? This way, I would have understood the PR better and also we might catch improvements before spending time to implement them.

frederikrothenberger:
Update regarding performance of event pruning: pruning 10k events takes about 24B instructions.
We can introduce a timer that prompts a pruning at least once per second. That prevents the to be pruned amount of events to ever grow to more than about ~1000 (at max subnet performance of ca. 1000 update calls / second), hence remaining well below the instruction limit.

I will introduce this pruning timer in a separate PR though.

src/internet_identity/src/activity_stats/event_stats.rs
frederikrothenberger: Great idea, added.

lmuntaner:
> Update regarding performance of event pruning: pruning 10k events takes about 24B instructions. We can introduce a timer that prompts a pruning at least once per second. That prevents the to be pruned amount of events to ever grow to more than about ~1000 (at max subnet performance of ca. 1000 update calls / second), hence remaining well below the instruction limit.
> 
> I will introduce this pruning timer in a separate PR though.

Is 24B a lot or not? Sorry, I'm not familiar with the limits in IC canisters. Why do you want to have a maximum of 1k events (2.4B instructions)?

frederikrothenberger:
> > Update regarding performance of event pruning: pruning 10k events takes about 24B instructions. We can introduce a timer that prompts a pruning at least once per second. That prevents the to be pruned amount of events to ever grow to more than about ~1000 (at max subnet performance of ca. 1000 update calls / second), hence remaining well below the instruction limit.
> > I will introduce this pruning timer in a separate PR though.
> 
> Is 24B a lot or not? Sorry, I'm not familiar with the limits in IC canisters. Why do you want to have a maximum of 1k events (2.4B instructions)?

On regular app subnets 20B is the instruction limits for update calls. We want to stay far away from that. Also above 4B we run into deterministic time slicing, which increases latency for users, so that should be avoided as well.

lmuntaner:
> On regular app subnets 20B is the instruction limit for update calls. We want to stay far away from that. Also above 4B we run into deterministic time slicing, which increases latency for users, so that should be avoided as well.

Thanks for the explanation!

PR #2451, opened 2024-05-02
Refactor IIDomain enum for conciseness

frederikrothenberger:
Refactors IIDomain to be more concise because the type will be used more often after #2449.


lmuntaner (approved):
LGTM

PR #2457, opened 2024-05-06
Expose gauge for number of entries in event_data and event_aggregations

frederikrothenberger:
Add counters for the event_data and event_aggregations BTreeMaps because these data structures can only be counted by iterating over them, which is not efficient enough for our use-case.

### DIFF
--- a/src/internet_identity/src/activity_stats/event_stats.rs
+++ b/src/internet_identity/src/activity_stats/event_stats.rs
@@ -186,6 +186,9 @@ fn update_events_internal<M: Memory>(event: EventData, now: Timestamp, s: &mut S
         "event already exists for key {:?}",
         current_key
     );
+    state::persistent_state_mut(|s| {
+        s.event_data_count += 1;
+    });
 
     let pruned_24h = if let Some((prev_key, _)) = s.event_data.iter_upper_bound(&current_key).next()
     {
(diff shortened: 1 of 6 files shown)


src/internet_identity/src/activity_stats/event_stats.rs
```diff
@@ -248,23 +251,54 @@ fn prune_events<M: Memory>(
...
     }
-
+    state::persistent_state_mut(|s| {
+        s.event_data_count -= pruned_events.len() as u64;
+    });
     pruned_events
 }
 
+/// Helper struct for updating the count of event aggregations.
+struct CountingAggregationsWrapper<'a, M: Memory>(&'a mut StableBTreeMap<AggregationKey, u64, M>);
```
lmuntaner, line 268: Why didn't you do something similar for the `event_data`?

lmuntaner (approved):
LGTM! Just a minor question

src/internet_identity/src/activity_stats/event_stats.rs
frederikrothenberger: Because there is exactly one place where we insert (asserting the insert happened) and one bulk removal (where we do not need to modify the counter one by one). So it seemed more straight forward to just manipulate the counter in place.
But the logic gets any more complicated, then yes, we should add such a wrapper as well.

lmuntaner: Ok, makes sense. Thanks for the explanation!

PR #2459, opened 2024-05-07
Upgrade agent-js

lmuntaner:
This PR upgrades agent-js to the latest version and removes the custom retry argument when creating an agent because the issue that required extra retris is fixed in the new version.


frederikrothenberger (approved):
Thanks!

PR #2462, opened 2024-05-07
Add II upgrade integration test for event stats

frederikrothenberger:
This adds an integration test to check that event stats are preserved across upgrades.


lmuntaner (approved):
LGTM!

PR #2464, opened 2024-05-08
Create `stats` as parent module for both activity and event stats

frederikrothenberger:
Currently, the event based stats are located within the activity stats which is misleading in terms of hierarchy.

This PR restructures the stats module as follows:
- stats
  - activity stats
  - event stats

This PR only _moves_ code and changes imports.


lmuntaner (approved):
Makes sense to me!

PR #2466, opened 2024-05-08
Simplify II init

frederikrothenberger:
A bug was introduced in #2460 that did not correctly restart timers after an upgrade.
This PR simplifies the initialization to share more code and adds a test to catch this bug in the future.

### DIFF
--- a/src/internet_identity/tests/integration/aggregation_stats.rs
+++ b/src/internet_identity/tests/integration/aggregation_stats.rs
@@ -177,11 +177,25 @@ fn should_keep_aggregations_across_upgrades() -> Result<(), CallError> {
 
 #[test]
 fn should_prune_automatically_after_upgrade() -> Result<(), CallError> {
+    const II_ORIGIN: &str = "ic0.app";
     let env = env();
     let canister_id = install_ii_canister(&env, II_WASM.clone());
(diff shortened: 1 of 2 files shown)


src/internet_identity/tests/integration/aggregation_stats.rs
```diff
@@ -122,19 +122,6 @@ fn should_report_at_most_100_entries() -> Result<(), CallError> {
     Ok(())
 }
 
-fn assert_expected_aggregation(
```
frederikrothenberger, line 125: This was simply moved down, after all the tests.

src/internet_identity/tests/integration/aggregation_stats.rs
```diff
@@ -188,6 +175,48 @@ fn should_keep_aggregations_across_upgrades() -> Result<(), CallError> {
     Ok(())
 }
 
+#[test]
+fn should_prune_automatically_after_upgrade() -> Result<(), CallError> {
```
lmuntaner, line 179: Why don't you add some events and assert they exist before upgrading?

frederikrothenberger: The event was there, but I didn't assert. Good idea, added.

lmuntaner (approved):
Thanks!

PR #2467, opened 2024-05-08
Drop archive memory restriction and simplify types

frederikrothenberger:
This PR removes the memory restriction on the archive buffer memory. It is replaced with monitoring and alerts.

Due to the incident, the limit was way too big to be useful anyway.

Since this change dropped the last usage of the type alias `NestedRestrictedMemory`, it is removed as well. It is replace with a much more useful type alias `ManagedMemory` which is applicable to all managed memories.


lmuntaner (approved):
Rubber stamp.

PR #2468, opened 2024-05-08
Report heap size on /metrics

frederikrothenberger:
Currently, there is no reporting of the heap size and that metric has to be obtained from the respective replica dashboard. This PR makes the heap size more easily accessible so that we can put it onto the Grafana dashboard too.


lmuntaner (approved):
LGTM!

PR #2470, opened 2024-05-13
Update readme

frederikrothenberger:
Because the readme was outdated (due to a previuos CI failure), the current release job did not correctly create the PR to update the readme automatically. With this manual update, the job should again work as expected.


lmuntaner (approved):
Could we add a slack message when the job fails to know when we have an outdated readme?

frederikrothenberger:
> Could we add a slack message when the job fails to know when we have an outdated readme?

We already have that. I just didn't have the time to deal with it until now.

PR #2473, opened 2024-05-16
Enable dev CSP by default when running dfx deploy

frederikrothenberger:
The dev CSP feature flag makes it possible to connect to II using http (no upgrade to https) which is required when using safari.

In addition, II may connect to other services using http which is also useful for development purposes.


lmuntaner (approved):
Thanks!

PR #2475, opened 2024-05-16
Switch vc_util to a minimal iota-fork

przydatek:
Switch `vc_util` to depend on a new iota-VC-library fork that contains minimal changes to work with IC/canister signatures.


frederikrothenberger (approved):
Thanks! Do you know why there are so many additional dependencies in the `Cargo.lock` files?

przydatek:
> Thanks! Do you know why there are so many additional dependencies in the `Cargo.lock` files?

Probably because the iota-library has evolved since August 2023 (our old fork is 92 commits and 10 releases behind).  I believe we can cut some options in the crates we're using, and so reduce the dependencies.  I'll do some experiments.
Currently the switch to the updated fork increases the size of `vc_issuer`-binary by about 33kB (roughly 3%):
```
-rw-r--r-- 1 890774 May 29 15:32 vc_demo_issuer--old-iota.wasm.gz
-rw-r--r-- 1 923056 May 28 16:15 vc_demo_issuer--updated-iota.wasm.gz
```

PR #2478, opened 2024-05-21
Fix dapp logo showcase

lmuntaner:
There was an issue in the home page's showcase of logos when a logo was not square and was instead a rectangle.

In this PR, I add a container to make sure the images have a square ration and the logo is contained within it.


lmuntaner:
The changes in the screenshots are because the logos are moved slightly to the left. I don't rely on the "left" to set the width of the container and instead I use the aspect-ratio.

peterpeterparker:
This logo (see screenshot) was too big before or becomes too big?

displayManage

peterpeterparker:
The Juno icon becomes a bit smaller but, I allow it 😉

PR #2482, opened 2024-05-29
Remove recursive self-call and add explicit iteration limit

frederikrothenberger:
This makes the event pruning safer by removing potentially dangerous constructs with self-calls such as endless loops and recursive self-calls.


lmuntaner (approved):
Thanks! I read the thread in the execution team to understand the context.

PR #2487, opened 2024-06-03
Add test with stable memory data that caused the stats incident

frederikrothenberger:
This PR adds a test that runs on a stable memory backup that reproduces the incident with `release-2024-05-13`.
For this test makes sure, that II can recover from such a state.


lmuntaner (approved):
Nice!

PR #2488, opened 2024-06-03
Replace assert with warning

frederikrothenberger:
This replaces another consistency check with a warning, such that II would continues to operate if there are duplicate keys for some reason.


przydatek (approved):
LGTM, thanks!

PR #2489, opened 2024-06-03
Add explicit bound to amortized clean-up

frederikrothenberger:
This adds an explicit limit to the number of events processed during amortized clean-up (100 events per time window). If more than 100 events need to be cleaned up, then the next call that adds an event will continue where the previous one left off.

Using that, we can (in a future PR) remove the timer based mechanism for pruning data.

### DIFF
--- a/src/internet_identity/src/stats/event_stats.rs
+++ b/src/internet_identity/src/stats/event_stats.rs
@@ -367,7 +367,7 @@ fn events_to_remove_from_24h_aggregations<M: Memory>(
     /// |  └ now - 24h                                        └ previous event
     /// └ previous event timestamp - 24h
     /// ```
-    fn removal_start_from_current_key<M: Memory>(
+    fn window_start_from_current_key<M: Memory>(
         current_key: &EventKey,
         db: &StableBTreeMap<EventKey, EventData, M>,
     ) -> Option<EventKey> {
@@ -377,14 +377,14 @@ fn events_to_remove_from_24h_aggregations<M: Memory>(
     }
 
     let window_start =
-        // Continue pruning from the last key. This value will be set if the 24h window has been pruned
-        // before.
+        // Load the window start from persistent state. This value will be set if events have been
+        // removed from the 24h window before.
         state::persistent_state(|s| s.event_stats_24h_start.clone()).or_else(|| {
             // Alternatively, calculate it from the current key. This is necessary in two cases:
-            // - the events have never been pruned before because they are not yet 24h old.
+            // - the events have never been removed before because they are not yet 24h old.
             // - this is the first event after an II upgrade from a version that did not have this
             //   state variable to track the beginning of the 24h window.
-            removal_start_from_current_key(current_key, db)
+            window_start_from_current_key(current_key, db)
         });
 
     let Some(start_key) = window_start else {
--- a/src/internet_identity/src/stats/event_stats.rs
+++ b/src/internet_identity/src/stats/event_stats.rs
@@ -77,7 +77,9 @@ mod event_aggregations;
 
 pub use event_aggregations::*;
 
-const MAX_EVENTS_TO_PRUNE: usize = 100;
+/// Number of past events to process per aggregation window in a single update call
+/// This limit will be applied twice, to the 24h and 30d aggregations.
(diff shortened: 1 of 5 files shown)


src/internet_identity/src/stats/event_stats.rs
```diff
@@ -276,69 +293,130 @@ fn update_events_internal<M: Memory>(event: EventData, now: Timestamp, s: &mut S
...
 }
 
-/// Adds an event to the event_data map and simultaneously removes events older than the retention period (30d).
+/// Collects events older than 24h that need to be removed from the 24h aggregations.
+/// Given events are kept for 30 days, the events are not deleted from the supplied `db`.
+/// Instead, this function simply updates the `event_stats_24h_pruning_start` state variable
+/// that denotes the first event that should be pruned in the next call.
+///
+/// Returns a vec of tuples of the pruned events and their timestamps, at most [MAX_EVENTS_TO_PRUNE].
+fn prune_events_24h<M: Memory>(
```
lmuntaner, line 345: This function doesn't actually prune any event, right? Could we rename it? `events_to_prune_24h` or `get_prune_events_24h`, does that make sense?

lmuntaner: Or even remove the "prune" altogether? It's more a list of events to update the aggregations rather than pruning them, right? `events_older_24h`?

src/internet_identity/src/stats/event_stats/tests.rs
```diff
@@ -397,6 +397,90 @@ fn should_prune_monthly_events_after_30d() {
...
+    );
+    update_events_internal(event.clone(), TIMESTAMP + 2 * DAY_NS, &mut storage);
+    assert_event_count_consistent(&mut storage);
+    // Prune again, after another 24h leaving only the event that triggered the pruning
+    // --> 1 expected events
+    assert_eq!(storage.event_aggregations.get(&aggregation_key).unwrap(), 1);
+    assert_eq!(
+        persistent_state(|s| s.event_stats_24h_pruning_start.clone()).unwrap(),
+        EventKey {
+            time: TIMESTAMP + DAY_NS,
```
lmuntaner, line 442: Maybe explain why is expected one day from now. It's because the event that triggers it is two days from now, right?

src/internet_identity/src/stats/event_stats/tests.rs
```diff
@@ -397,6 +397,90 @@ fn should_prune_monthly_events_after_30d() {
     );
 }
 
+#[test]
+fn should_prune_at_most_100_events_24h() {
```
lmuntaner, line 401: As mentioned, maybe we shouldn't call "pruning" to the process of updating the aggregations?

src/internet_identity/src/stats/event_stats.rs
```diff
@@ -73,12 +74,15 @@ use std::time::Duration;
 
 /// This module defines the aggregations over the events.
 mod event_aggregations;
+
 pub use event_aggregations::*;
 
+const MAX_EVENTS_TO_PRUNE: usize = 100;
```
lmuntaner, line 80: This assumes that the difference between days will never be in the order of magnitude larger than 100, right?

For example, if one day we see 100.000 events, and the next day we have 10. Not all events will be pruned, right?

lmuntaner (commented):
Looking good! I added a few comments to better understand the change.

src/internet_identity/src/stats/event_stats.rs
frederikrothenberger: Yes, if there is a huge spike in activity followed by no / very low activity then, for some time, the aggregations will be over a window that is actually longer than 24h. Eventually, the clean-up will catch up though, as for every event added, 100 are cleaned-up.

src/internet_identity/src/stats/event_stats/tests.rs
frederikrothenberger: Yes exactly. Added a comment.

src/internet_identity/src/stats/event_stats.rs
frederikrothenberger: I renamed it to be clearer. 👍

lmuntaner (approved):
Thanks! LGTM!

PR #2491, opened 2024-06-05
Remove timer-based pruning

frederikrothenberger:
This removes the timer based pruning that is no longer needed due to https://github.com/dfinity/internet-identity/pull/2489.
The prune event aggregations will be removed once II reports their value
to be 0 (i.e. in a subsequent release after the next one).


lmuntaner (approved):
LGTM!

PR #2493, opened 2024-06-05
Add to vc-issuer a default root key, guard configure()

przydatek:
Change the demo VC issuer to use the mainnet IC root public key by default, and guard `configure()` endpoint, so that only canister controller(s) can change the configuration.


frederikrothenberger (approved):
Makes sense, lgtm! Thanks.

PR #2495, opened 2024-06-06
Tweak event aggregations exposed on /metrics

frederikrothenberger:
This PR changes the aggregations exposed on /metrics to be more useful:
* remove the metrics for the non-official / other domain
* add prepare delegation count for https://identity.ic0.app

This allow to have both, aggregated session time _and_ count be displayed on the Grafana dashboard, while keeping the number of metrics tracked by victoria metrics low.

### DIFF
--- a/src/internet_identity/src/http/metrics.rs
+++ b/src/internet_identity/src/http/metrics.rs
@@ -4,13 +4,13 @@ use crate::state::PersistentState;
 use crate::stats::activity_stats::activity_counter::ActivityCounter;
 use crate::stats::activity_stats::ActivityStats;
 use crate::stats::event_stats::AggregationWindow::{Day, Month};
-use crate::stats::event_stats::{retrieve_aggregation, Aggregation, PD_SESS_SEC};
+use crate::stats::event_stats::{retrieve_aggregation, Aggregation, PD_COUNT, PD_SESS_SEC};
 use crate::{state, IC0_APP_DOMAIN, INTERNETCOMPUTER_ORG_DOMAIN};
 use ic_cdk::api::stable::stable64_size;
 use ic_cdk::api::time;
 use ic_metrics_encoder::{LabeledMetricsBuilder, MetricsEncoder};
 use std::time::Duration;
-use IIDomain::{Ic0App, InternetComputerOrg};
+use IIDomain::Ic0App;
 
 /// Collects the various metrics exposed by the Internet Identity canister.
 /// Returns a ascii-encoded string of the metrics in the Prometheus exposition format.
@@ -152,6 +152,11 @@ fn event_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> Result<(), std::io::Error> 
         "Cumulative authenticated session duration in seconds for the given front-end origin.",
     )?;
     pd_aggregation_metrics(PD_SESS_SEC, sess_sec_builder)?;
+    let count_builder = w.gauge_vec(
+        "internet_identity_prepare_delegation_count",
+        "Number of authenticated sessions for the given front-end origin.",
+    )?;
+    pd_aggregation_metrics(PD_COUNT, count_builder)?;
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


src/internet_identity/src/http/metrics.rs
```diff
@@ -152,6 +152,11 @@ fn event_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> Result<(), std::io::Error>
         "Cumulative authenticated session duration in seconds for the given front-end origin.",
     )?;
     pd_aggregation_metrics(PD_SESS_SEC, sess_sec_builder)?;
+    let count_builder = w.gauge_vec(
```
lmuntaner, line 155: Why wasn't this before exposed in /metrics ?

src/internet_identity/src/http/metrics.rs
```diff
@@ -160,19 +165,17 @@ fn pd_aggregation_metrics(
     mut metrics_builder: LabeledMetricsBuilder<Vec<u8>>,
 ) -> Result<(), std::io::Error> {
     for window in &[Day, Month] {
-        for domain in &[None, Some(Ic0App), Some(InternetComputerOrg)] {
```
lmuntaner, line 163: Is there a specific reason to remove the metrics from the non-official domain?

src/internet_identity/tests/integration/http.rs
```diff
@@ -609,9 +609,9 @@ fn should_list_aggregated_session_seconds_and_event_data_counters() -> Result<()
     api::prepare_delegation(
         &env,
         canister_id,
-        test_authn_method().principal(),
+        authn_method_ic0.principal(),
```
lmuntaner, line 612: Why this change?

src/internet_identity/tests/integration/http.rs
```diff
@@ -609,9 +609,9 @@ fn should_list_aggregated_session_seconds_and_event_data_counters() -> Result<()
...
         user_number_1,
-        "https://some-dapp-1.com",
+        "https://some-dapp-2.com",
```
lmuntaner, line 614: Why this change?

src/internet_identity/tests/integration/http.rs
```diff
@@ -637,33 +637,23 @@ fn should_list_aggregated_session_seconds_and_event_data_counters() -> Result<()
...
     assert_metric(
         &metrics,
-        "internet_identity_prepare_delegation_session_seconds{dapp=\"https://some-dapp-1.com\",window=\"24h\",ii_origin=\"other\"}",
-        5400f64,
-    );
-    assert_metric(
-        &metrics,
-        "internet_identity_prepare_delegation_session_seconds{dapp=\"https://some-dapp-1.com\",window=\"30d\",ii_origin=\"other\"}",
+        "internet_identity_prepare_delegation_session_seconds{dapp=\"https://some-dapp-2.com\",window=\"24h\",ii_origin=\"ic0.app\"}",
         5400f64,
```
lmuntaner, line 641: Sorry, but why is it 5400?

I see that the session lasts 3600 one time and None the second.

src/internet_identity/src/http/metrics.rs
frederikrothenberger: Because there is non-negligible cost to the time-series kept in victoria metrics. So we only expose a specific subset to be available for historical data.

And this PR changes the subset to be more useful.

src/internet_identity/src/http/metrics.rs
frederikrothenberger: Yes, because of the non-negligible storage cost and because the data is not that interesting.

src/internet_identity/tests/integration/http.rs
frederikrothenberger: So that the counter of `"https://some-dapp-2.com"` is not just 1, but a higher number so that it couldn't just be an off by one error 😉

src/internet_identity/tests/integration/http.rs
frederikrothenberger: see above

src/internet_identity/tests/integration/http.rs
frederikrothenberger: Because `None` defaults to 1800 seconds (30 minutes). I can also set an explicit value if you prefer.

src/internet_identity/src/http/metrics.rs
lmuntaner: Ah, ok, understood.

src/internet_identity/src/http/metrics.rs
lmuntaner: Thanks for the explanation!

src/internet_identity/tests/integration/http.rs
lmuntaner: Ok thanks for the explanation.

src/internet_identity/tests/integration/http.rs
lmuntaner: Ah, ok, got it.

lmuntaner (approved):

PR #2496, opened 2024-06-06
Reset event data on empty persistent state counts

frederikrothenberger:
This is necessary because the persistent state will reinitialize the event and aggregation count to 0, if rolled back across a version that did not have the event data feature yet.

### DIFF
--- a/src/internet_identity/src/state.rs
+++ b/src/internet_identity/src/state.rs
@@ -264,7 +264,17 @@ pub fn save_persistent_state() {
 
 pub fn load_persistent_state() {
     STATE.with(|s| {
-        storage_borrow(|storage| *s.persistent_state.borrow_mut() = storage.read_persistent_state())
+        let loaded_state = storage_borrow(|storage| storage.read_persistent_state());
+
+        // Reset the event_data and event_aggregations if they are reported as empty from the persistent state
+        // Necessary to handle rollback across the versions where the event_data and event_aggregations were introduced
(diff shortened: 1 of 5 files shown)


src/internet_identity/src/state.rs
```diff
@@ -264,7 +264,17 @@ pub fn save_persistent_state() {
 
 pub fn load_persistent_state() {
     STATE.with(|s| {
-        storage_borrow(|storage| *s.persistent_state.borrow_mut() = storage.read_persistent_state())
+        let loaded_state = storage_borrow(|storage| storage.read_persistent_state());
+
+        // Reset the event_data and event_aggregations if they are reported as empty from the persistent state
```
lmuntaner, line 269: What is the problem if we don't do this?

That there would be stale events and aggregations in the state that won't be pruned?

lmuntaner (approved):
LGTM! Just one small question

src/internet_identity/src/state.rs
frederikrothenberger: Then the exported metrics will be off. I.e. we will not know how many events / aggregations are currently being managed by II. The only efficient way to make this consistent again is to reset both to zero.

The stats per domain would still be accurate though.

lmuntaner: Why will they be off? Because we have been two weeks without it?

lmuntaner: Why is the value initialized at 0 if the previous didn't have a schema for it?

PR #2508, opened 2024-06-14
Increase stack size to 3MB

frederikrothenberger:
This PR increases the stack size from the 1MB default to 3MB as per recommendation [here](https://dfinity.slack.com/archives/CH4CADCJX/p1717748706241909?thread_ts=1716464873.584739&cid=CH4CADCJX).

I did some tests with very small stack sizes to see when we hit errors to get a feeling for the current stack usage of II. The tests indicate that we currently use more than 48KB but fewer than 200KB of stack.

Since we are also very lean on heap usage (~6.5 MB currently) we can afford to increase the stack size to have more leeway with regards to the stack.


przydatek (approved):
LGTM, thanks!

PR #2515, opened 2024-06-21
chore: close img tag

dfx-json:
This is to ensure the docusaurus compiler does not throw any errors during compilation for the portal


frederikrothenberger (approved):
Thanks

PR #2517, opened 2024-06-25
Do not show recovery page everytime

lmuntaner:
Main motivation is to not nag the user so much with adding a recovery phrase.

This PR changes to logic to show it only once a week.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -366,13 +366,6 @@ export const displayManage = (
         onAddDevice,
         addRecoveryPhrase,
         addRecoveryKey: async () => {
-          const confirmed = confirm(
-            "Add a Recovery Device\n\nUse a FIDO Security Key, like a YubiKey, as an additional recovery method."
-          );
-          if (!confirmed) {
-            // No resolve here because we don't need to reload the screen
-            return;
-          }
           await setupKey({ connection });
           resolve();
         },
--- a/src/frontend/src/flows/recovery/recoveryWizard.ts
+++ b/src/frontend/src/flows/recovery/recoveryWizard.ts
@@ -7,6 +7,10 @@ import { AuthenticatedConnection } from "$src/utils/iiConnection";
 import { setupRecovery } from "./setupRecovery";
 
 import { infoScreenTemplate } from "$src/components/infoScreen";
+import {
+  getLastShownWarningPageTimestamp,
+  setShownRecoveryWarningPage,
+} from "$src/storage";
 import copyJson from "./recoveryWizard.json";
 
 /* Phrase creation kick-off screen */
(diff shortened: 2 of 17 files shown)


src/frontend/src/test-e2e/flows.ts
```diff
@@ -80,16 +80,11 @@ export const FLOWS = {
...
-    deviceName: string,
     browser: WebdriverIO.Browser
   ): Promise<void> => {
     const welcomeView = new WelcomeView(browser);
     await welcomeView.waitForDisplay();
     await welcomeView.login(userNumber);
-    // This flow assumes no recovery phrase, so we explicitly skip the recovery nag here
-    await FLOWS.skipRecoveryNag(browser);
-    const mainView = new MainView(browser);
-    await mainView.waitForDeviceDisplay(deviceName);
```
lmuntaner, line 92: I added the await where `loginWelcomeView` is used.

I wanted to avoid coupling the logic of what comes after the flow from the flow.

lmuntaner: Another idea I had was to add a boolean parameter to control whether it should skip the recovery warning page or not.

src/frontend/src/flows/manage/index.ts
```diff
@@ -366,13 +366,6 @@ export const displayManage = (
         onAddDevice,
         addRecoveryPhrase,
         addRecoveryKey: async () => {
-          const confirmed = confirm(
```
lmuntaner, line 369: We agreed with Mary that this didn't add any value to the user.

lmuntaner:
We agreed with Andronikos to keep this in the user metadata in the canister.

@frederikrothenberger do you want to review and merge this so that we have it for the next release in the local storage or we skip that, and I implement it in the backend metadata directly?

frederikrothenberger:
> We agreed with Andronikos to keep this in the user metadata in the canister.
> 
> @frederikrothenberger do you want to review and merge this so that we have it for the next release in the local storage or we skip that, and I implement it in the backend metadata directly?

As discussed, let's not merge in this state, but rather do the backend integration directly.

src/frontend/src/flows/register/index.ts
```diff
@@ -37,7 +36,7 @@ import { displayUserNumberWarmup } from "./finish";
 import { savePasskeyOrPin } from "./passkey";
 
 /** Registration (anchor creation) flow for new users */
-export const registerFlow = async <T>({
+export const registerFlow = async ({
```
lmuntaner, line 39: The generic types was used in the showcase and was set to `null`.

But now, I need the connection to be of type AuthenticatedConnection because it's used.

This type was the `T` within the app, but not in the showcase. So, no extra changes were needed in the app part except removing the generic.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -473,6 +479,60 @@ export class AuthenticatedConnection extends Connection {
...
+  };
+
+  setIdentityMetadata = async (
+    metadata: MetadataMapV2
+  ): Promise<{ Ok: null } | { Err: IdentityMetadataReplaceError }> => {
+    const actor = await this.getActor();
+    return await actor.identity_metadata_replace(this.userNumber, metadata);
+  };
+
+  setShownRecoveryWarningPage = async (): Promise<void> => {
```
lmuntaner, line 496: I decided to put the logic of how the timestamp is saved within the `AuthenticatedConnection`. Not a strong opinion, but I didn't find another place.

My idea would be to have different methods to manage different metadata keys. And they might share some logic, but at least have a simple API with a complex implementation.

The downside is that I need to make two calls when setting the metadata.

* To get the metadata
* To replace it.

lmuntaner:

src/frontend/src/flows/recovery/recoveryWizard.ts
```diff
@@ -86,10 +86,20 @@ export const recoveryWizard = async (
...
   );
-  if (recoveries.length === 0) {
+  const ONE_WEEK_MILLIS = 7 * 24 * 60 * 60 * 1000;
+  const nowInMillis = Date.now();
+  const oneWeekAgoTimestamp = nowInMillis - ONE_WEEK_MILLIS;
+  const hasNotSeenRecoveryPageLastWeek =
+    (lastTimestampShown ?? 0) < oneWeekAgoTimestamp;
+  if (recoveries.length === 0 && hasNotSeenRecoveryPageLastWeek) {
+    // No need to await for this, we don't need the result.
+    void connection.setShownRecoveryWarningPage();
```
frederikrothenberger, line 102: Ideally we would still make sure this operation is actually completed, before the II window gets closed...

src/frontend/src/utils/iiConnection.ts
frederikrothenberger: What about switching to a model where the metadata map is loaded (asynchronously) when constructing the `AuthenticatedConnection`, and then modifications (maybe multiple) happen on the cached copy. And then, at some distinct point (in parallel with something that takes a while anyway, e.g. `prepare_delegation`) we write it back?

frederikrothenberger:
@lmuntaner: this PR is now obsolete due to #2523, right?

lmuntaner:
> @lmuntaner: this PR is now obsolete due to #2523, right?

That's correct. Closing this one because implementation was done in #2523 and #2520

PR #2518, opened 2024-06-25
Bump vitest and webdriverio

frederikrothenberger:
Upgrading these dependencies fixes dependabot alerts. Requires pinning a specific version of `ws` though. This should be removed in the future.

Includes changes to the vitest arguments to run tests sequentially, because the options changed in the upgrade to `1.x.x`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "preview:showcase": "astro preview --root ./src/showcase",
     "screenshots": "npm run opts -- ./src/frontend/screenshots.ts",
     "test": "vitest --config ./vitest.config.ts",
-    "test:e2e": "vitest --mode e2e --config ./vitest.config.ts --threads=false",
+    "test:e2e": "VITEST_MAX_THREADS=1 vitest --mode e2e --config ./vitest.config.ts",
     "test:e2e-desktop": "SCREEN=desktop npm run test:e2e",
     "test:e2e-mobile": "SCREEN=mobile npm run test:e2e",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -75,6 +75,9 @@
     "ua-parser-js": "^1.0.35",
     "zod": "^3.22.3"
   },
+  "overrides": {
+    "ws": "8.17.1"
```
lmuntaner, line 79: This was necessary in the end?

Which are the packages we need to watch to upgrade to be able to remove this?

Could you add it in the PR description so that we can come back to it to better understand it?

lmuntaner (approved):
LGTM, just one comment.

package.json
frederikrothenberger: Yes, it was. `puppeteer-core` seems to be the culprit:

```
+-- vitest@1.6.0
| `-- jsdom@20.0.3
|   `-- ws@8.17.1 overridden
`-- webdriverio@8.39.0
  +-- devtools@8.39.0
  | `-- puppeteer-core@20.3.0
  |   `-- ws@8.13.0 invalid: "8.17.1" from node_modules/devtools/node_modules/puppeteer-core overridden
  +-- puppeteer-core@20.9.0
  | `-- ws@8.13.0 invalid: "8.17.1" from node_modules/puppeteer-core overridden
  `-- webdriver@8.39.0
    `-- ws@8.17.1 deduped
```

PR #2519, opened 2024-06-26
Add metric for number of successful prepare_id_alias calls

frederikrothenberger:
This PR adds a metric for how many prepare_id_alias calls have been handled. This number can be used to infer the number of VCs presented using Internet Identity.


lmuntaner (approved):
LGTM!

Ideally could we also store the app where it's coming from like we do for logins?

Even more ideally, two new metrics per dapps, one per relying party and one per issuer.

But definitely not in this PR.

frederikrothenberger:
> LGTM!
> 
> Ideally could we also store the app where it's coming from like we do for logins?
> 
> Even more ideally, two new metrics per dapps, one per relying party and one per issuer.
> 
> But definitely not in this PR.

Yes, that would be interesting. Agreed! However, we should be careful about more targeted stats, as long as the volume of VC flows is low (i.e. to not leak information about the context).

PR #2520, opened 2024-06-27
Identity Metadata Repository setup

lmuntaner:
Main motivation is to set up the module to manage the anchor metadata.

### DIFF
--- a/src/frontend/src/flows/authorize/index.ts
+++ b/src/frontend/src/flows/authorize/index.ts
@@ -212,6 +212,7 @@ const authenticate = async (
   const derivationOrigin =
     authContext.authRequest.derivationOrigin ?? authContext.requestOrigin;
 
+  // TODO: Commit state
   const result = await withLoader(() =>
     fetchDelegation({
       connection: authSuccess.connection,
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -250,6 +250,7 @@ export const renderManage = async ({
   for (;;) {
     let anchorInfo: IdentityAnchorInfo;
     try {
+      // TODO: commit state
       anchorInfo = await withLoader(() => connection.getAnchorInfo());
     } catch (error: unknown) {
       await displayFailedToListDevices(
--- a/src/frontend/src/repositories/anchorMetadata.test.ts
+++ b/src/frontend/src/repositories/anchorMetadata.test.ts
@@ -1,0 +1,223 @@
+import { MetadataMapV2 } from "$generated/internet_identity_types";
+import {
+  AnchorMetadata,
+  AnchorMetadataRepository,
+  RECOVERY_PAGE_SHOW_TIMESTAMP_MILLIS,
+} from "./anchorMetadata";
+
+const recoveryPageShownTimestampMillis = 1234567890;
+const mockRawMetadata: MetadataMapV2 = [
+  [
+    RECOVERY_PAGE_SHOW_TIMESTAMP_MILLIS,
+    { String: String(recoveryPageShownTimestampMillis) },
+  ],
+];
+const mockAnchorMetadata: AnchorMetadata = {
+  recoveryPageShownTimestampMillis,
+};
+
+// Used to await that the getter has resolved.
+let getterResponse: MetadataMapV2 | null | undefined = null;
+const getterMockSuccess = vi.fn().mockImplementation(async () => {
+  // The `await` is necessary to make sure that the `getterResponse` is set before the test continues.
+  getterResponse = await mockRawMetadata;
+  return mockRawMetadata;
+});
+const getterMockError = vi.fn().mockImplementation(async () => {
+  // The `await` is necessary to make sure that the `getterResponse` is set before the test continues.
+  getterResponse = await null;
+  throw new Error("test error");
+});
+const setterMockSuccess = vi.fn();
+const setterMockError = vi.fn().mockRejectedValue("test error");
+
+beforeEach(() => {
+  getterResponse = undefined;
+  vi.clearAllMocks();
+  vi.spyOn(console, "warn").mockImplementation(() => {});
+});
+
+test("AnchorMetadataRepository loads data on init in the background", async () => {
+  const instance = AnchorMetadataRepository.init({
+    getter: getterMockSuccess,
+    setter: setterMockSuccess,
+  });
+
+  await vi.waitFor(() => expect(getterResponse).toEqual(mockRawMetadata));
+
+  expect(getterMockSuccess).toHaveBeenCalledTimes(1);
+  expect(await instance.getMetadata()).toEqual(mockAnchorMetadata);
+});
+
+test("AnchorMetadataRepository loads data again on init in the background", async () => {
+  const getterMockErrorOnce = vi
+    .fn()
+    .mockImplementationOnce(async () => {
+      // The `await` is necessary to make sure that the `getterResponse` is set before the test continues.
+      getterResponse = await null;
+      throw new Error("test error");
+    })
+    .mockResolvedValue(mockRawMetadata);
+
+  const instance = AnchorMetadataRepository.init({
+    getter: getterMockErrorOnce,
+    setter: setterMockSuccess,
+  });
+
+  // Wait for the first getter to fail.
+  await vi.waitFor(() => expect(getterResponse).toEqual(null));
+
+  // Error is not thrown, but a warning is logged.
+  expect(console.warn).toHaveBeenCalledTimes(1);
+  expect(getterMockErrorOnce).toHaveBeenCalledTimes(1);
+
+  expect(await instance.getMetadata()).toEqual(mockAnchorMetadata);
+  expect(getterMockErrorOnce).toHaveBeenCalledTimes(2);
+});
+
+test("AnchorMetadataRepository returns no metadata if fetching fails without raising an error", async () => {
+  const instance = AnchorMetadataRepository.init({
+    getter: getterMockError,
+    setter: setterMockSuccess,
+  });
+
+  // Wait for the first getter to fail.
+  await vi.waitFor(() => expect(getterResponse).toEqual(null));
+
+  // Error is not thrown, but warnings is logged.
+  expect(console.warn).toHaveBeenCalledTimes(1);
+  expect(getterMockError).toHaveBeenCalledTimes(1);
+
+  expect(await instance.getMetadata()).toEqual(undefined);
+  expect(getterMockError).toHaveBeenCalledTimes(2);
+  expect(console.warn).toHaveBeenCalledTimes(2);
+});
+
+test("AnchorMetadataRepository changes data in memory", async () => {
+  const instance = AnchorMetadataRepository.init({
+    getter: getterMockSuccess,
+    setter: setterMockSuccess,
+  });
+
+  await vi.waitFor(() => expect(getterResponse).toEqual(mockRawMetadata));
+
+  const newRecoveryPageShownTimestampMillis = 9876543210;
+  await instance.setPartialMetadata({
+    recoveryPageShownTimestampMillis: newRecoveryPageShownTimestampMillis,
+  });
+
+  expect(await instance.getMetadata()).toEqual({
+    recoveryPageShownTimestampMillis: newRecoveryPageShownTimestampMillis,
+  });
+});
+
+test("AnchorMetadataRepository commits updated metadata to canister", async () => {
+  const instance = AnchorMetadataRepository.init({
+    getter: getterMockSuccess,
+    setter: setterMockSuccess,
+  });
+
+  await vi.waitFor(() => expect(getterResponse).toEqual(mockRawMetadata));
+
+  const newRecoveryPageShownTimestampMillis = 9876543210;
+  await instance.setPartialMetadata({
+    recoveryPageShownTimestampMillis: newRecoveryPageShownTimestampMillis,
+  });
+
+  expect(setterMockSuccess).not.toHaveBeenCalled();
+  await instance.commitMetadata();
+
+  expect(setterMockSuccess).toHaveBeenCalledTimes(1);
+  expect(setterMockSuccess).toHaveBeenCalledWith([
+    [
+      RECOVERY_PAGE_SHOW_TIMESTAMP_MILLIS,
+      { String: String(newRecoveryPageShownTimestampMillis) },
+    ],
+  ]);
+});
+
+test("AnchorMetadataRepository doesn't commit to canister without changes", async () => {
+  const instance = AnchorMetadataRepository.init({
+    getter: getterMockSuccess,
+    setter: setterMockSuccess,
+  });
+
+  await vi.waitFor(() => expect(getterResponse).toEqual(mockRawMetadata));
+
+  expect(setterMockSuccess).not.toHaveBeenCalled();
+  await instance.commitMetadata();
+
+  expect(setterMockSuccess).not.toHaveBeenCalled();
+});
+
+test("AnchorMetadataRepository doesn't raise an error if committing fails", async () => {
+  const instance = AnchorMetadataRepository.init({
+    getter: getterMockSuccess,
+    setter: setterMockError,
+  });
+
+  await vi.waitFor(() => expect(getterResponse).toEqual(mockRawMetadata));
+
+  const newRecoveryPageShownTimestampMillis = 9876543210;
+  const newMetadata = {
+    recoveryPageShownTimestampMillis: newRecoveryPageShownTimestampMillis,
+  };
+  await instance.setPartialMetadata(newMetadata);
+
+  expect(setterMockError).not.toHaveBeenCalled();
+  const committed = await instance.commitMetadata();
+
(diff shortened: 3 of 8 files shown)


src/frontend/src/repositories/anchorMetadata.ts
```diff
@@ -0,0 +1,168 @@
+import { MetadataMapV2 } from "$generated/internet_identity_types";
+import type { AuthenticatedConnection } from "$src/utils/iiConnection";
+
+export type AnchorMetadata = {
+  recoveryPageShownTimestampMillis?: number;
```
frederikrothenberger, line 5: Doesn't this need to be `bigint`?

src/frontend/src/repositories/anchorMetadata.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+
+const RECOVERY_PAGE_SHOW_TIMESTAMP_MILLIS = "recoveryPageShownTimestampMillis";
+const METADATA_FETCH_TIMEOUT = 1_000;
+const METADATA_FETCH_RETRIES = 10;
```
frederikrothenberger, line 10: 10 seems like a lot, no?

src/frontend/src/repositories/anchorMetadata.ts
```diff
@@ -0,0 +1,168 @@
...
+        this.rawMetadata === "error" ||
+        this.rawMetadata === "not-loaded"
+      ) {
+        // Retry in case of error or not loaded.
+        void this.loadMetadata();
+      } else {
+        break;
+      }
+    }
+    if (
```
frederikrothenberger, line 89: I'd suggest to extract a helper for this condition.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -511,6 +521,34 @@ export class AuthenticatedConnection extends Connection {
     await actor.remove(this.userNumber, publicKey);
   };
 
+  getIdentityInfo = async (): Promise<
```
frederikrothenberger, line 524: A client should only use the `getAnchorMetadata`, `setPartialMetadata` and `commitMetadata` functions. `getIdentityInfo` and `setIdentityMetadata` should be `private` then, no?

frederikrothenberger (commented):
Yes, this is the right direction. Not sure about the complexity of retries though...
It makes the front-end more robust, but the PR #2517 didn't have it either?

src/frontend/src/repositories/anchorMetadata.ts
lmuntaner: I will get it from the browser with `Date.now()` which returns a `number`.

From chatGPT:

> Date.now() will return Number.MAX_SAFE_INTEGER sometime around the year 287,586

If we were using nanoseconds, then yes. But with milliseconds it should be enough.

src/frontend/src/repositories/anchorMetadata.ts
lmuntaner: makes sense

src/frontend/src/utils/iiConnection.ts
lmuntaner: Then the repository won't be able to use it...

I could pass the getter and setter as functions only.

lmuntaner:
> Yes, this is the right direction. Not sure about the complexity of retries though... It makes the front-end more robust, but the PR #2517 didn't have it either?

I was thinking and yes, I think we don't need retries. Even more, I think I want to swallow the errors and log a warning only.

The metadata is not crucial, we can find sensible defaults in case something isn't working as expected. Moreover, with the imperative code, I don't want an error to get metadata to disrupt a user flow.

@frederikrothenberger What do you think if I remove all the throws and instead swallow or return `undefined` or `false` accordingly?

frederikrothenberger:
> > Yes, this is the right direction. Not sure about the complexity of retries though... It makes the front-end more robust, but the PR #2517 didn't have it either?
> 
> I was thinking and yes, I think we don't need retries. Even more, I think I want to swallow the errors and log a warning only.
> 
> The metadata is not crucial, we can find sensible defaults in case something isn't working as expected. Moreover, with the imperative code, I don't want an error to get metadata to disrupt a user flow.
> 
> @frederikrothenberger What do you think if I remove all the throws and instead swallow or return `undefined` or `false` accordingly?

Yes, a warning should be sufficient. 👍

src/frontend/src/flows/authorize/index.ts
```diff
@@ -212,6 +212,7 @@ const authenticate = async (
   const derivationOrigin =
     authContext.authRequest.derivationOrigin ?? authContext.requestOrigin;
 
+  // TODO: Commit state
```
lmuntaner, line 215: To be done in another PR, but I want to remember exactly where to put it.

src/frontend/src/flows/manage/index.ts
```diff
@@ -250,6 +250,7 @@ export const renderManage = async ({
   for (;;) {
     let anchorInfo: IdentityAnchorInfo;
     try {
+      // TODO: commit state
```
lmuntaner, line 253: To be done in another PR, but I want to remember exactly where to put it.

lmuntaner:

src/frontend/src/repositories/anchorMetadata.ts
```diff
@@ -53,114 +86,107 @@ export class AnchorMetadataRepository {
...
     this.rawMetadata = "loading";
-    const identityInfo = await this.connection.getIdentityInfo();
-    if ("Ok" in identityInfo) {
+    try {
       this.updatedMetadata = false;
-      this.rawMetadata = identityInfo.Ok.metadata;
-    } else {
+      this.rawMetadata = await this.getter();
+    } catch (error) {
+      // Do not throw the error becasue this is not critical for the application.
```
frederikrothenberger, line 93:
```suggestion
      // Do not throw the error because this is not critical for the application.
```

src/frontend/src/repositories/anchorMetadata.ts
```diff
@@ -0,0 +1,194 @@
...
+ * Class to manage the metadata of the anchor and interact with the canister.
+ *
+ * We decided to not throw any errors because this is non-critical for the application
+ * and we don't want to disrupt user flows if there is an error with the metadata.
+ *
+ * This class loads the metadata in the background when it's created.
+ * It can then be read and updated in memory.
+ * The metadata needs to be committed to the canister with the `commitMetadata` method to persist the changes.
+ */
+export class AnchorMetadataRepository {
```
frederikrothenberger, line 44: Quite a while ago we decided to sunset the term "anchor" because it is confusing. So on the UI, this term is no longer used. In the code, we didn't change things.
But for new code, we now use "identity" rather than anchor. Hence, why on the API level it is called `IdentityMetadata` rather than `AnchorMetadata`.
I think it would be better to use the new terminology here too and keep going with the slow journey of eventually replacing all the occurrences of "anchor".

src/frontend/src/utils/iiConnection.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    BigInt(1234),
+    mockActor
+  );
+
+  await vi.waitFor(() => expect(infoResponse).toEqual(mockRawMetadata));
+
+  expect(await connection.getAnchorMetadata()).toEqual(mockAnchorMetadata);
+});
+
+test("comits changes on anchor metadata", async () => {
```
frederikrothenberger, line 74:
```suggestion
test("commits changes on anchor metadata", async () => {
```

src/frontend/src/utils/iiConnection.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  get_anchor_info: vi.fn().mockResolvedValue(expectedAnchorInfo),
+  identity_info: vi.fn().mockImplementation(async () => {
+    // The `await` is necessary to make sure that the `getterResponse` is set before the test continues.
+    infoResponse = await mockRawMetadata;
+    return { Ok: { metadata: mockRawMetadata } };
+  }),
+  identity_metadata_replace: vi.fn().mockResolvedValue({ Ok: null }),
+} as unknown as ActorSubclass<_SERVICE>;
+
+test("gets anchor info from actor", async () => {
```
frederikrothenberger, line 47: While it is good to test things, I'm a bit confused as to how `getAnchorInfo` relates to the identity metadata functionality added in this PR?

src/frontend/src/repositories/anchorMetadata.ts
```diff
@@ -0,0 +1,194 @@
...
+  // The nice AnchorMetadata is exposed to the outside world, while the raw metadata is kept private.
+  // We keep all the raw data to maintain other metadata fields.
+  private rawMetadata: RawMetadataState;
+  // Flag to keep track whether we need to commit the metadata to the canister.
+  private updatedMetadata: boolean;
+  private getter: MetadataGetter;
```
frederikrothenberger, line 50: These should be `readonly`?

src/frontend/src/repositories/anchorMetadata.test.ts
```diff
@@ -0,0 +1,223 @@
...
+// Used to await that the getter has resolved.
+let getterResponse: MetadataMapV2 | null | undefined = null;
+const getterMockSuccess = vi.fn().mockImplementation(async () => {
+  // The `await` is necessary to make sure that the `getterResponse` is set before the test continues.
+  getterResponse = await mockRawMetadata;
+  return mockRawMetadata;
+});
+const getterMockError = vi.fn().mockImplementation(async () => {
+  // The `await` is necessary to make sure that the `getterResponse` is set before the test continues.
+  getterResponse = await null;
```
frederikrothenberger, line 28: Sorry, I'm missing some context here. Why does the `await` change anything at all? Shouldn't that instantly resolve anyway?

src/frontend/src/repositories/anchorMetadata.ts
```diff
@@ -0,0 +1,194 @@
...
+   *
+   * It tries to load the metadata if it's not loaded yet.
+   *
+   * It returns `undefined` if the metadata is not loaded.
+   *
+   * @returns {AnchorMetadata | undefined}
+   */
+  getMetadata = async (): Promise<AnchorMetadata | undefined> => {
+    if (!this.metadataIsLoaded(this.rawMetadata)) {
+      await this.loadMetadata();
```
frederikrothenberger, line 121: Shouldn't we simply await `this.rawMetadata` here? Otherwise you might load things twice / are actually implementing an implicit retry?

Given it's not critical, we should probably avoid that, no?

src/frontend/src/repositories/anchorMetadata.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    setter: setterMockSuccess,
+  });
+
+  await vi.waitFor(() => expect(getterResponse).toEqual(mockRawMetadata));
+
+  expect(getterMockSuccess).toHaveBeenCalledTimes(1);
+  expect(await instance.getMetadata()).toEqual(mockAnchorMetadata);
+});
+
+test("AnchorMetadataRepository loads data again on init in the background", async () => {
```
frederikrothenberger, line 52: See comment below, but I think the metadata should not be loaded _again_.

src/frontend/src/repositories/anchorMetadata.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  // Error is not thrown, but a warning is logged.
+  expect(console.warn).toHaveBeenCalledTimes(1);
+  expect(getterMockErrorOnce).toHaveBeenCalledTimes(1);
+
+  expect(await instance.getMetadata()).toEqual(mockAnchorMetadata);
+  expect(getterMockErrorOnce).toHaveBeenCalledTimes(2);
+});
+
+test("AnchorMetadataRepository returns no metadata if fetching fails without raising an error", async () => {
```
frederikrothenberger, line 78:
```suggestion
test("AnchorMetadataRepository returns undefined without raising an error if fetching fails", async () => {
```

src/frontend/src/repositories/anchorMetadata.test.ts
```diff
@@ -0,0 +1,223 @@
+import { MetadataMapV2 } from "$generated/internet_identity_types";
```
frederikrothenberger, line 1: I think it would be good to have a test to ensure that the repository leaves additional data unchanged in `rawMetadata`.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -511,6 +538,34 @@ export class AuthenticatedConnection extends Connection {
...
+  ): Promise<{ Ok: null } | { Err: IdentityMetadataReplaceError }> => {
+    const actor = await this.getActor();
+    return await actor.identity_metadata_replace(this.userNumber, metadata);
+  };
+
+  getAnchorMetadata = (): Promise<AnchorMetadata | undefined> => {
+    return this.metadataRepository.getMetadata();
+  };
+
+  setPartialMetadata = async (
```
frederikrothenberger, line 559:
```suggestion
  updateIdentityMetadata = async (
```

frederikrothenberger (commented):
Thanks! I still have a few questions.

src/frontend/src/repositories/anchorMetadata.ts
lmuntaner: Thanks!

src/frontend/src/repositories/anchorMetadata.ts
lmuntaner: Done!

src/frontend/src/utils/iiConnection.test.ts
lmuntaner: Thanks!

src/frontend/src/utils/iiConnection.test.ts
lmuntaner: I was trying that I could successfully test AuthenticatedConnection and then I thought it might be ok to leave it.

But let's do that in another PR if I want to add more tests. I removed it from this PR.

src/frontend/src/repositories/anchorMetadata.ts
lmuntaner: yes, done!

src/frontend/src/repositories/anchorMetadata.test.ts
lmuntaner: For some reason it wasn't waiting before. But I tried again and it was working now. Removed.

src/frontend/src/repositories/anchorMetadata.ts
lmuntaner: As agreed, there won't be a `loadMetadata` call again and this function just returns the pretty metadata.

src/frontend/src/repositories/anchorMetadata.test.ts
lmuntaner: Done.

src/frontend/src/repositories/anchorMetadata.test.ts
lmuntaner: Thanks!

src/frontend/src/repositories/anchorMetadata.test.ts
lmuntaner: I had a test for this: `"IdentityMetadataRepository commits additional metadata to canister after update"`.

I rephrased the description to use "additional", maybe that was the confusing part.

src/frontend/src/utils/iiConnection.ts
lmuntaner: done

lmuntaner:
@frederikrothenberger ready for another review

lmuntaner:
@frederikrothenberger wait a bit, I found a bug.

lmuntaner:
@frederikrothenberger ready. I forgot to await for the rawMetadata.

src/frontend/src/repositories/identityMetadata.ts
```diff
@@ -0,0 +1,195 @@
...
+   * This method won't throw an error if the metadata can't be loaded.
+   * Instead, it will set the instance variable `rawMetadata` to "error".
+   *
+   * @returns {Promise<void>} In case a client wants to wait for the metadata to be loaded.
+   */
+  loadMetadata = async (): Promise<void> => {
+    this.rawMetadata = "loading";
+    try {
+      this.updatedMetadata = false;
+      this.rawMetadata = await this.getter();
```
frederikrothenberger, line 91: You could also just put the `Promise` as a field. That way you could await it alongside a timer with `Promise.race(...)` in `getMetadata`, no?

Not sure about the error handling though.

frederikrothenberger (approved):
Thanks, lgtm! I left one suggestion but I don't feel strongly about it.

src/frontend/src/repositories/identityMetadata.ts
lmuntaner: Yes, if I put the promise there I lose the states and I don't see the advantage. Unless I don't put the promise of the getter directly, but I wrap the getter so that it returns a promise of the states.

What do you think it's the advantage? I can do it in another PR. We would also lose the possibility of a client waiting on `loadMetadata`.

PR #2494, opened 2024-06-06
Replace Logo

meodai:
On @artkorotkikh-dfinity's request, I replaced the Logo.

To make sure it is still accessible to people using a screenreader, Ive added a `<title>` to the group containing the text inside the SVG.

Screenshot 2024-06-06 at 13 44 57

<hr/><details><summary>🟡 Some screens were changed</summary></details>

PR #2522, opened 2024-06-28
Introduce new Landing Page layout

lmuntaner:
The main motivation is to introduce a new layout for the landing page of II.

In this PR, I introduce the new layouts and prepare the flag to enable it.

Yet, I only enable it in the showcase.

This is because we still need changes in the screens before being ready for release. Those changes will happen in another PR.

### DIFF
--- a/src/frontend/src/components/authenticateBox.ts
+++ b/src/frontend/src/components/authenticateBox.ts
@@ -48,6 +48,7 @@ import { promptUserNumber } from "./promptUserNumber";
 
 import { DerEncodedPublicKey } from "@dfinity/agent";
 import copyJson from "./authenticateBox.json";
+import { landingPage } from "./landingPage";
 
 /** Template used for rendering specific authentication screens. See `authnScreens` below
  * for meaning of "firstTime", "useExisting" and "pick". */
@@ -562,12 +563,22 @@ export const authnPages = (i18n: I18n, props: AuthnTemplates) => {
   const templates = authnTemplates(i18n, props);
 
   return {
-    firstTime: (opts: Parameters<typeof templates.firstTime>[0]) =>
-      page(templates.firstTime(opts)),
+    firstTime: (
+      opts: Parameters<typeof templates.firstTime>[0],
+      // TODO: remove this parameter once the landing page is ready.
+      landingPageTemplate: boolean = false
+    ) =>
+      // TODO: Use the landing page template always when ready.
+      page(templates.firstTime(opts), landingPageTemplate),
     useExisting: (opts: Parameters<typeof templates.useExisting>[0]) =>
       page(templates.useExisting(opts)),
-    pick: (opts: Parameters<typeof templates.pick>[0]) =>
-      page(templates.pick(opts)),
+    pick: (
+      opts: Parameters<typeof templates.pick>[0],
+      // TODO: remove this parameter once the landing page is ready.
+      landingPageTemplate: boolean = false
+    ) =>
+      // TODO: Use the landing page template always when ready.
+      page(templates.pick(opts), landingPageTemplate),
   };
 };
 
@@ -615,12 +626,16 @@ export const authnScreens = (i18n: I18n, props: AuthnTemplates) => {
 };
 
 // Wrap the template with header & footer and render the page
-const page = (slot: TemplateResult) => {
-  const template = mainWindow({
-    slot: html` <!-- The title is hidden but used for accessibility -->
-      <h1 data-page="authenticate" class="is-hidden">Internet Identity</h1>
-      ${slot}`,
-  });
+const page = (slot: TemplateResult, landingPageTemplate: boolean = false) => {
+  const template = landingPageTemplate
+    ? landingPage({
+        slot,
+      })
+    : mainWindow({
+        slot: html` <!-- The title is hidden but used for accessibility -->
+          <h1 data-page="authenticate" class="is-hidden">Internet Identity</h1>
+          ${slot}`,
+      });
   const container = document.getElementById("pageContent") as HTMLElement;
   render(template, container);
 };
--- a/src/frontend/src/components/landingPage.ts
+++ b/src/frontend/src/components/landingPage.ts
@@ -1,6 +1,8 @@
+import { I18n } from "$src/i18n";
 import { html, TemplateResult } from "lit-html";
 import { navigationLink } from "./footer";
 import { githubIcon, icLogo, questionIcon } from "./icons";
+import copyJson from "./landingPage.json";
 
 /**
  * Landing page template
@@ -16,6 +18,9 @@ export const landingPage = ({
 }: {
   slot: TemplateResult;
 }): TemplateResult => {
+  const i18n = new I18n();
+  const copy = i18n.i18n(copyJson);
+
   return html` <main class="c-landingPage">
     <div class="c-landingPage__logo">
       <div class="c-logo">${icLogo}</div>
@@ -27,13 +32,8 @@ export const landingPage = ({
     </section>
     <section class="c-landingPage__left">
       <div class="c-landingPage__left__content">
-        <h1 class="t-title t-title--main">
-          Secure, seamless & privacy-preserving digital identity
-        </h1>
-        <p class="t-paragraph">
-          Internet Identity is a decentralized identity solution running
-          end-to-end on the Internet Computer blockchain.
-        </p>
+        <h1 class="t-title t-title--main">${copy.title}</h1>
+        <p class="t-paragraph">{${copy.subtitle}}</p>
       </div>
       <div class="c-landingPage__left__footer">
         ${navigationLink({
--- a/src/frontend/src/components/landingPage.ts
+++ b/src/frontend/src/components/landingPage.ts
@@ -30,7 +30,7 @@ export const landingPage = ({
         <div class="c-landingPage__right__content--full-width">${slot}</div>
       </div>
     </section>
-    <section class="c-landingPage__left">
+    <section class="c-landingPage__left" aria-label="Action Pane">
       <div class="c-landingPage__left__content">
         <h1 class="t-title t-title--main">${copy.title}</h1>
         <p class="t-paragraph">{${copy.subtitle}}</p>
--- a/src/frontend/src/components/landingPage.ts
+++ b/src/frontend/src/components/landingPage.ts
@@ -1,0 +1,58 @@
+import { html, TemplateResult } from "lit-html";
+import { navigationLink } from "./footer";
+import { githubIcon, icLogo, questionIcon } from "./icons";
+
+/**
+ * Landing page template
+ *
+ * It is a component with split panes left and right in desktop view.
+ * To the left there is some static content, and to the right there is a slot.
+ *
+ * In movile view, the static content is below the slot.
+ *
+ */
+export const landingPage = ({
+  slot,
+}: {
+  slot: TemplateResult;
+}): TemplateResult => {
+  return html` <main class="c-landingPage">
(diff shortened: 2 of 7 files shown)


src/frontend/src/components/authenticateBox.ts
```diff
@@ -562,12 +563,22 @@ export const authnPages = (i18n: I18n, props: AuthnTemplates) => {
   const templates = authnTemplates(i18n, props);
 
   return {
-    firstTime: (opts: Parameters<typeof templates.firstTime>[0]) =>
-      page(templates.firstTime(opts)),
+    firstTime: (
+      opts: Parameters<typeof templates.firstTime>[0],
+      // TODO: remove this parameter once the landing page is ready.
+      landingPageTemplate: boolean = false
```
lmuntaner, line 569: I needed this to use it in the showcase for now.

In a following PR, I'll remove it and the `firstTime` and `pick` will always come with the new landing page, whereas the `useExisting` comes with the old layout.

src/showcase/src/pages/manageNewLanding.astro
lmuntaner, line 1: This will be removed in a following PR because the `manage` will already be with the new landing.

But I wanted to have both in the showcase for now to share with product and design as well as myself when developing.

src/frontend/src/components/landingPage.ts
```diff
@@ -0,0 +1,58 @@
...
+        <div class="c-landingPage__right__content--full-width">${slot}</div>
+      </div>
+    </section>
+    <section class="c-landingPage__left">
+      <div class="c-landingPage__left__content">
+        <h1 class="t-title t-title--main">
+          Secure, seamless & privacy-preserving digital identity
+        </h1>
+        <p class="t-paragraph">
+          Internet Identity is a decentralized identity solution running
```
frederikrothenberger, line 34: Could you extract the text into an i18n file?

src/frontend/src/styles/main.css
```diff
@@ -131,6 +131,7 @@
   --vc-brand-alt: rgba(215, 205, 235, 1);
   --vc-grey: grey;
   --vc-light-grey: #d4d8e5;
+  --vc-brand-gradient: linear-gradient(90deg, #4b19d6 0%, #890eed 100%);
```
frederikrothenberger, line 134: It looks great. But I'm absolutely hopeless with CSS so it might be good to get a second opinion form @peterpeterparker or @meodai.

src/frontend/src/components/landingPage.ts
```diff
@@ -0,0 +1,58 @@
...
+    </div>
+    <section class="c-landingPage__right">
+      <div class="c-landingPage__right__content">
+        <div class="c-landingPage__right__content--full-width">${slot}</div>
+      </div>
+    </section>
+    <section class="c-landingPage__left">
+      <div class="c-landingPage__left__content">
+        <h1 class="t-title t-title--main">
+          Secure, seamless & privacy-preserving digital identity
```
frederikrothenberger, line 31: In the other layout there is a special (hidden) title for accessibility. I think for screen readers the landing page should still say "Internet Identity" first (which it presumably does in the old layout due to the hidden title?).

Or is the accessibility concern appropriately addressed by [this](https://github.com/dfinity/internet-identity/pull/2494#issuecomment-2152207910)?

lmuntaner: It was addressed by https://github.com/dfinity/internet-identity/pull/2494#issuecomment-2152207910, yes.

The logo appears first with the title Internet Identity.

src/frontend/src/components/landingPage.ts
lmuntaner: Oh, yes! Thanks!

lmuntaner:
@peterpeterparker @meodai could you take a look at the CSS of the new landing page? Thanks!

src/frontend/src/styles/main.css
lmuntaner: Done!

peterpeterparker:
I don't really know the CSS of II, better if @meodai does have a look.

frederikrothenberger (approved):
LGTM, thanks! Please wait for feedback from @meodai before merging.

src/frontend/src/styles/main.css
meodai: @lmuntaner: I would use vars here: `linear-gradient(90deg, var(--vc-brand-purple), var(--vc-brand-purple-light));` and the `%` are not needed if you are going to split them evenly.

src/frontend/src/styles/main.css
```diff
@@ -203,6 +204,9 @@
 
   --rc-button--disabled: var(--vc-shadow);
 
+  --rc-landing-page-background: var(--vc-brand-gradient);
```
meodai, line 207: --rc is for color, not a gradient. it would be rg or something I guess.

But I think  `--rc-landing-page-background` is too specific anyway. It allows for almost no reuse. Maybe calling it something like --rg-highlight? I would discuss with Artem for what reason the gradients are used and use this as the name,

src/frontend/src/styles/main.css
```diff
@@ -254,6 +258,8 @@
   --rg-brand-bruised: var(--vc-brand-blue) 50%, var(--vc-brand-purple) 90%;
 
   /* reference tokens: sizes */
+  --rs-main-content-width: 40rem;
```
meodai, line 261: reference tokens have to point to value tokens.

src/frontend/src/styles/main.css
```diff
@@ -313,6 +319,12 @@
   --rs-line--thick: calc(var(--rs-line) * 4);
 
   --rs-footer-height: 4rem;
+
+  --rs-landing-page-padding--mobile: calc(var(--vs-stack) * 3);
```
meodai, line 323: these are also too specific I think. Maybe calling it something like split layout would allow for some reuse?

src/frontend/src/styles/main.css
```diff
@@ -949,6 +961,159 @@ a:hover,
...
+/**
+ * LandingPage component
+ */
+
+.c-landingPage {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--rs-main-content-width), 1fr);
+  grid-template-areas: "left right";
+
+  height: 100vh;
```
meodai, line 973: should this not be min-height?

meodai: And dvh instead of vh?

src/frontend/src/styles/main.css
```diff
@@ -949,6 +961,159 @@ a:hover,
...
+
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+
+  padding: 0 20%;
+
+  /* To compensate for the IC logo that is outisde the normal flow with position absolute */
+  --logo-height: 3rem;
```
meodai, line 1012: should this var not be set more globally or not at all if its only used here?

src/frontend/src/styles/main.css
```diff
@@ -949,6 +961,159 @@ a:hover,
...
+
+    background: var(--rc-landing-page-background);
+
+    height: auto;
+    min-height: 100vh;
+  }
+
+  .c-landingPage__logo {
+    display: flex;
+    position: static;
```
meodai, line 1074: why position static? its the default anyway

src/frontend/src/components/landingPage.ts
```diff
@@ -0,0 +1,58 @@
...
+      <div class="c-landingPage__right__content">
+        <div class="c-landingPage__right__content--full-width">${slot}</div>
+      </div>
+    </section>
+    <section class="c-landingPage__left">
+      <div class="c-landingPage__left__content">
+        <h1 class="t-title t-title--main">${copy.title}</h1>
+        <p class="t-paragraph">{${copy.subtitle}}</p>
+      </div>
+      <div class="c-landingPage__left__footer">
```
meodai, line 38: this the main I think not the thing around...

src/frontend/src/components/landingPage.ts
```diff
@@ -0,0 +1,58 @@
...
+  slot: TemplateResult;
+}): TemplateResult => {
+  const i18n = new I18n();
+  const copy = i18n.i18n(copyJson);
+
+  return html` <main class="c-landingPage">
+    <div class="c-landingPage__logo">
+      <div class="c-logo">${icLogo}</div>
+    </div>
+    <section class="c-landingPage__right">
```
meodai, line 28: semantic elements are supposed to have either or an aria-label

meodai (changes requested):

src/frontend/src/styles/main.css
lmuntaner: I didn't know about `dvh` thanks!

I tried with `min-height`. But it creates a weird two scroll with the right page is higher than the screen. With `height` the scroll is smooth.

src/frontend/src/styles/main.css
lmuntaner: Because I set it to `absolute` for desktop and I want the default for mobile.

src/frontend/src/styles/main.css
lmuntaner: Done! Good point, I didn't realize they were the same colors, thanks!

src/frontend/src/styles/main.css
lmuntaner: I prefer to keep it as `rc`. A color could easily replace the gradient in the future and I don't want to change the name of the variable then.

src/frontend/src/styles/main.css
lmuntaner: Good point, done!

src/frontend/src/styles/main.css
lmuntaner: Good idea, done!

src/frontend/src/components/landingPage.ts
lmuntaner: what do you mean?

src/frontend/src/components/landingPage.ts
lmuntaner: either what?

I added an `aria-label` makes sense to me.

lmuntaner:
@meodai ready for another review. Thanks!

src/frontend/src/components/landingPage.ts
meodai: either a title (h1, h2, h3...) or aria-labael. sorry

src/frontend/src/components/landingPage.ts
meodai: main is only supposed to be the actual main content of the page. Not everything including the logo and this marketing stuff on the left

src/frontend/src/styles/main.css
meodai: But then something else is the issue. i you use height and the content happens to be higher as the actual layout, this might lead to issues

meodai (approved):

src/frontend/src/components/landingPage.ts
lmuntaner: Ahh, ok. In that case, yes, the section will have a title already. The component inside always has an `h1`.

src/frontend/src/styles/main.css
lmuntaner: The UI is ok, it's the scroll that seems weird.

This is with `min-height` (notice the pause in the scroll):

https://github.com/dfinity/internet-identity/assets/4550653/b70f5023-ab15-4f43-a1c8-6569f3e50537

This is with `height`:

https://github.com/dfinity/internet-identity/assets/4550653/8e284eee-0104-452f-bbd7-dfcdfb8a8c8a

PR #2523, opened 2024-07-02
Use identity metadata to not show recovery warning page

lmuntaner:
Main motivation is to use the IdentityMetadataRepository functionality in the application to avoid showing the recovery warning page always. Instead, it's shown once a week.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -250,8 +250,10 @@ export const renderManage = async ({
   for (;;) {
     let anchorInfo: IdentityAnchorInfo;
     try {
-      // TODO: commit state
-      anchorInfo = await withLoader(() => connection.getAnchorInfo());
+      // Ignore the `commitMetadata` response, it's not critical for the application.
+      [anchorInfo] = await withLoader(() =>
+        Promise.all([connection.getAnchorInfo(), connection.commitMetadata()])
+      );
     } catch (error: unknown) {
       await displayFailedToListDevices(
         error instanceof Error ? error : unknownError()
@@ -367,13 +369,6 @@ export const displayManage = (
         onAddDevice,
         addRecoveryPhrase,
         addRecoveryKey: async () => {
-          const confirmed = confirm(
-            "Add a Recovery Device\n\nUse a FIDO Security Key, like a YubiKey, as an additional recovery method."
-          );
-          if (!confirmed) {
-            // No resolve here because we don't need to reload the screen
-            return;
-          }
           await setupKey({ connection });
           resolve();
         },
--- a/src/frontend/src/repositories/identityMetadata.ts
+++ b/src/frontend/src/repositories/identityMetadata.ts
@@ -101,9 +101,9 @@ export class IdentityMetadataRepository {
     metadata: RawMetadataState
   ): metadata is MetadataMapV2 => {
     return (
-      this.rawMetadata !== "loading" &&
-      this.rawMetadata !== "error" &&
-      this.rawMetadata !== "not-loaded"
+      metadata !== "loading" &&
+      metadata !== "error" &&
+      metadata !== "not-loaded"
     );
   };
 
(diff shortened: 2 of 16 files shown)


src/frontend/src/flows/manage/index.ts
```diff
@@ -367,13 +369,6 @@ export const displayManage = (
         onAddDevice,
         addRecoveryPhrase,
         addRecoveryKey: async () => {
-          const confirmed = confirm(
```
lmuntaner, line 370: We agreed with product to remove this extra step.

src/frontend/src/repositories/identityMetadata.ts
```diff
@@ -101,9 +101,9 @@ export class IdentityMetadataRepository {
     metadata: RawMetadataState
   ): metadata is MetadataMapV2 => {
     return (
-      this.rawMetadata !== "loading" &&
-      this.rawMetadata !== "error" &&
-      this.rawMetadata !== "not-loaded"
+      metadata !== "loading" &&
```
lmuntaner, line 104: It works anyway, but I feld that it was better to do the checks on the passed parameter.

I needed to pass it so that I could do the type guard `metadata is MetadataMapV2`.

lmuntaner:

src/frontend/src/test-e2e/flows.ts
```diff
@@ -178,11 +167,6 @@ export const FLOWS = {
     await addDeviceSuccessView.waitForDisplay();
     await addDeviceSuccessView.continue();
   },
-  skipRecoveryNag: async (browser: WebdriverIO.Browser): Promise<void> => {
```
frederikrothenberger, line 181: Is there any test left that still goes through the recovery nag? I think there should be one e2e test that does the recovery nag flow. You could trigger that by artificially setting the metadata to some value in the past.

src/showcase/src/flows.ts
```diff
@@ -26,7 +31,37 @@ export const flowsPage = () => {
...
+      Ok: {
+        authn_methods: [],
+        metadata: [],
+        authn_method_registration: [],
+      },
+    };
+  },
+} as unknown as ActorSubclass<_SERVICE>;
+
+class MockAuthenticatedConnection extends AuthenticatedConnection {
```
frederikrothenberger, line 47: Not in this PR: We should extract an interface rather than extending the real implementation...

frederikrothenberger (commented):
LGTM, thanks. Let's have a discussion about the e2e test for the recovery nag to decide how we want to proceed with that.

src/frontend/src/test-e2e/flows.ts
lmuntaner: No, there is no test that has this flow anymore. I checked a way to advance time, but that's not possible for e2e.

How could I artificially set the metadata to some value in the past?

Shall I expose the function in the console and trigger it? It seems dangerous.

src/showcase/src/flows.ts
lmuntaner: Yes, that sounds better.

src/frontend/src/test-e2e/flows.ts
lmuntaner: Do we do that somewhere else already?

frederikrothenberger: No, but we do similar things when we modify alternative origins though.

lmuntaner: I'm sorry, but I don't find it. Why should the alternative origins create an identity? Where does it do that?

lmuntaner: As agreed, I tested the flow manually by changing the waiting time to one minute. I was able to test that before one minute the user doesn't see the page and after one minute the page appears. I did it three time to test that updating the metadata also works as expected.

lmuntaner:

src/frontend/src/test-e2e/flows.ts
frederikrothenberger: Thanks @lmuntaner and as discussed, the e2e test requires some infrastructure so we will do this in a separate PR.

PR #2526, opened 2024-07-04
Read canister id from 500 status response

lmuntaner:
Yesterday in a debugging session with Civic I realized that the problem was that they had as a derivation origin a custom domain pointing to a backend canister that was only serving the alternative domains path. Therefore, II couldn't resolve the canister id because it kept getting a 500 error in the HEAD request.

I asked them to change the derivation origin to the canister id url <canister-id>.icp0.io to unblock them.

Even though it might be an edge case to have a derivation origin a backend canister with a custom domain, I don't see why we shouldn't support it.

Conclusion: I removed the status check from the response to read the canister id from the headers.


frederikrothenberger (approved):
While the whole mechanism would be much better replaced by an actual API to query, this makes sense as it does address a real problem.
We should keep pushing the boundary nodes team to offer a better solution though.

PR #2527, opened 2024-07-04
UI fixes landing page

lmuntaner:
Main motivation is to fix a few UI changes requested by design:

* Change title styles for the marketing copy in the landing page.
* Add more space between footer links and change the hover color in the landing page.
* Improve the responsiveness by setting the breakpoint at 1024 and setting a max width in the content.
* Change the vertical padding for the right pane on mobile.

### DIFF
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -336,6 +336,7 @@
     var(--vs-stack) * 2
   );
   --rs-split-page-stack: calc(var(--vs-stack) * 4);
+  --rs-split-page-stack--large: calc(var(--vs-stack) * 6);
 
   /* This is not used in CSS, but it's used to compensate for the logo size. */
   --rs-logo-height: 3rem;
@@ -1124,7 +1125,7 @@ a:hover,
     margin-top: var(--rs-split-page-stack);
     padding-bottom: var(--rs-split-page-stack);
 
-    gap: var(--rs-split-page-stack);
+    gap: var(--rs-split-page-stack--large);
   }
 
   .c-landingPage__left__content {
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -329,7 +329,12 @@
   --rs-split-page-padding--mobile: calc(var(--vs-stack) * 3);
   --rs-split-page-left-pane-padding: calc(var(--vs-gutter) * 8);
   --rs-split-page-right-pane-padding--desktop: 10vmax;
-  --rs-split-page-right-pane-padding--mobile: calc(var(--vs-stack) * 2);
+  --rs-split-page-right-pane-padding--mobile--vertical: calc(
+    var(--vs-stack) * 3
+  );
+  --rs-split-page-right-pane-padding--mobile--horizontal: calc(
+    var(--vs-stack) * 2
+  );
   --rs-split-page-stack: calc(var(--vs-stack) * 4);
 
   /* This is not used in CSS, but it's used to compensate for the logo size. */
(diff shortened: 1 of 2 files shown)


lmuntaner:
New UI:

Screenshot 2024-07-04 at 11 19 49
Screenshot 2024-07-04 at 11 20 00
Screenshot 2024-07-04 at 11 20 14

lmuntaner:
@frederikrothenberger PR approved by Artem and ready for review

lmuntaner:
@meodai could you take a look?

src/frontend/src/styles/main.css
```diff
@@ -329,8 +329,14 @@
   --rs-split-page-padding--mobile: calc(var(--vs-stack) * 3);
   --rs-split-page-left-pane-padding: calc(var(--vs-gutter) * 8);
   --rs-split-page-right-pane-padding--desktop: 10vmax;
-  --rs-split-page-right-pane-padding--mobile: calc(var(--vs-stack) * 2);
+  --rs-split-page-right-pane-padding--mobile--vertical: calc(
```
meodai, line 332: my preference would be: `--rs-split-page-right-pane-padding-x--mobile` double modifiers in names a bit weird inho

src/frontend/src/styles/main.css
```diff
@@ -975,11 +981,21 @@ a:hover,
  */
 
 .c-landingPage {
+  height: 100dvh;
```
meodai, line 984: min-height? what happens if the content is higher then c-landingPage?

src/frontend/src/styles/main.css
```diff
@@ -975,11 +981,21 @@ a:hover,
...
   display: grid;
   grid-template-columns: 1fr minmax(var(--rs-main-content-width), 1fr);
   grid-template-areas: "left right";
 
-  height: 100dvh;
+  height: 100%;
+}
+
+.c-landingPage__title {
+  font-size: calc(2.8rem + 1vw);
```
meodai, line 996: could we use a token here? Like marketing title or something?

src/frontend/src/styles/main.css
meodai: I notice now that the variables names say "padding". I usually try to avoid technical implementation in the var name, because it limits reuse. I usually use:

- `somethin-stack` for spaces above or below outside of the element
- `something-inline` for spaces left or right outside of the element
- `something-bezzel` or `gutter` for spaces withing (or bezzel-x bezzel-y) 

https://codepen.io/meodai/pen/zNaEyM

But feel free to change this an other time if at all since its a bit out of scope for this PR. And I did not notice last time

src/frontend/src/styles/main.css
lmuntaner: I don't want it taller than the screen.

The page has two panes. The left one should match the height of the screen, while the right one is allowed to have scroll if the content doesn't fit.

This is with `height`:

https://github.com/dfinity/internet-identity/assets/4550653/577c4732-197b-43f7-892c-5b2c22d01863

This is with `min-height`:


https://github.com/dfinity/internet-identity/assets/4550653/5b5c5b01-1c35-4bed-8f14-20c88b184ae8

Does this make sense?

src/frontend/src/styles/main.css
lmuntaner: True, I renamed all the new vars to use the convention. Makes sense to me.

src/frontend/src/styles/main.css
lmuntaner: Done, thanks!

lmuntaner:
@meodai ready for another review! Thanks!

src/frontend/src/styles/main.css
meodai: aha I did not get the the other pane was scrollable on its own! Thanks for clearing that up

PR #2528, opened 2024-07-04
Add support to reconfigure derivation_origin for VC issuer

frederikrothenberger:
This PR allows to reconfigure the expected frontend hostname as well as the derivation origin. This will be used in the future to simplify the e2e tests.

Since this is most useful in e2e tests if the configuration is _not_ guarded by the controller, the guard is removed for the `configure` method too for consistency.

### DIFF
--- a/demos/vc_issuer/tests/issue_credential.rs
+++ b/demos/vc_issuer/tests/issue_credential.rs
@@ -6,9 +6,7 @@ use canister_sig_util::{extract_raw_root_pk_from_der, CanisterSigPublicKey};
 use canister_tests::api::http_request;
 use canister_tests::api::internet_identity::vc_mvp as ii_api;
 use canister_tests::flows;
-use canister_tests::framework::{
-    env, get_wasm_path, principal_1, principal_2, test_principal, time, II_WASM,
-};
+use canister_tests::framework::{env, get_wasm_path, principal_1, test_principal, time, II_WASM};
 use ic_cdk::api::management_canister::provisional::CanisterId;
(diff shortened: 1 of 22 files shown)


demos/vc_issuer/tests/issue_credential.rs
```diff
@@ -113,10 +111,9 @@ mod api {
     pub fn configure(
         env: &StateMachine,
         canister_id: CanisterId,
-        sender: Principal,
         config: &IssuerInit,
     ) -> Result<(), CallError> {
-        call_candid_as(env, canister_id, sender, "configure", (config,))
+        call_candid(env, canister_id, "configure", (config,))
```
frederikrothenberger, line 116: I removed the need to specify the `sender` as these methods can now be called anonymously again.

demos/vc_issuer/tests/issue_credential.rs
```diff
@@ -353,12 +362,34 @@ fn should_return_derivation_origin() {
...
+    api::set_derivation_origin(
+        &env,
+        canister_id,
+        frontend_hostname.as_str(),
+        derivation_origin,
+    )
+    .expect("failed to set derivation_origin");
+
+    let req = DerivationOriginRequest { frontend_hostname };
+    let response = api::derivation_origin(&env, canister_id, &req)
```
lmuntaner, line 387: Could you check this before calling `set_derivation_origin` to confirm that it has indeed changed?

lmuntaner (approved):
LGTM! Just one comment.

PR #2530, opened 2024-07-04
Disable ic-commit auto-update

frederikrothenberger:
The StateMachine binary has been broken since a while. But rather than investing time to fix it, we should simply migrate to PocketIC. PocketIC will is bundled with the next release of DFX.
Tracked in https://dfinity.atlassian.net/browse/GIX-2266.


lmuntaner (approved):
LGTM!

PR #2535, opened 2024-07-08
Make canisterId mandatory for VC flow

frederikrothenberger:
This PR makes it mandatory for the relying party to specify the canisterId when initiating the VC flow.

The change brings the following advantages:
* simpler logic on the II side
* support issuers with different front-end and back-end canisters without having them resort to alternative origins
* make the flow easier to understand for developers

### DIFF
--- a/docs/vc-spec.md
+++ b/docs/vc-spec.md
@@ -259,9 +259,8 @@ After receiving the notification that II is ready, the relying party can request
 * Method: `request_credential`
 * Params:
   * `issuer`: An issuer that the relying party trusts. It has the following properties:
-    * `origin`: The origin of the issuer.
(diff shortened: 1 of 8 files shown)


docs/vc-spec.md
```diff
@@ -321,7 +320,9 @@ After receiving the notification that II is ready, the relying party can request
   "method": "request_credential",
   "params": {
     "issuer": {
-        "origin": "https://kyc-resident-info.org"
+        "type": "IC",
```
lmuntaner, line 323: Is this needed?

frederikrothenberger: No, I missed that one when removing `type` again. Cood catch!

lmuntaner (approved):
Thanks!

PR #2541, opened 2024-07-17
Disable macos release build check due to flakyness

frederikrothenberger:
The release build check on macos was very flaky: one in 3 builds fails due to timeout.
Hence MacOs builds are disabled for now until there is time to look into a better solution.


lmuntaner (approved):
Thanks

PR #2543, opened 2024-07-18
fix: Candid file rendering in spec

jessiemongeon1:
Adds support to import the Candid interface file as a codeblock in the spec document


frederikrothenberger (approved):
LGTM, thanks!

PR #2550, opened 2024-07-26
Create warning page to add a device

lmuntaner:
The main motivation is to stop using "Recovery" concept.

In this PR, we create a page that will substitute the warning page shown to users with a pin key once a week.

At the moment, the warning page suggests the users to add a recovery phrase.

The new page will have two flavors:
* User has only pin key: It will suggest to add a device.
* User has only one passkey: It will suggest to add another device.


anedos-dfinity (approved):
LGTM

PR #2563, opened 2024-08-20
Implement API to auto-select identity based on known principal

frederikrothenberger:
This adds the option to specify a principal when requesting a session such that II auto-select the identity if the principal is known.

If the principal is not known, II will fall back to default behaviour (same as now) with manual identity selection.

### DIFF
--- a/docs/ii-spec.mdx
+++ b/docs/ii-spec.mdx
@@ -159,6 +159,7 @@ This section describes the Internet Identity Service from the point of view of a
       maxTimeToLive?: bigint;
       allowPinAuthentication?: boolean;
       derivationOrigin?: string;
+      autoSelectMatchingIdentity?: string
     }
     ```
 
@@ -171,6 +172,8 @@ This section describes the Internet Identity Service from the point of view of a
     -   the `allowPinAuthentication` (EXPERIMENTAL), if present, indicates whether or not the Identity Provider should allow the user to authenticate and/or register using a temporary key/PIN identity. Authenticating dapps may want to prevent users from using Temporary keys/PIN identities because Temporary keys/PIN identities are less secure than Passkeys (webauthn credentials) and because Temporary keys/PIN identities generally only live in a browser database (which may get cleared by the browser/OS).
 
     -   the `derivationOrigin`, if present, indicates an origin that should be used for principal derivation instead of the client origin. Internet Identity will only accept values that are also listed in the HTTP resource `/.well-known/ii-alternative-origins` of the corresponding canister (see [Alternative Frontend Origins](#alternative-frontend-origins)).
+
+    -   the `autoSelectMatchingIdentity`, if present, indicates the textual representation of this dapp's principal for which the delegation is requested. If it is known to Internet Identity, it will skip the identity selection and immediately prompt for authentication. This feature can be used to streamline re-authentication after a session expiry.
 
 
 6.  Now the client application window expects a message back, with data `event`.
--- a/docs/ii-spec.mdx
+++ b/docs/ii-spec.mdx
@@ -159,7 +159,7 @@ This section describes the Internet Identity Service from the point of view of a
       maxTimeToLive?: bigint;
       allowPinAuthentication?: boolean;
       derivationOrigin?: string;
-      autoSelectMatchingIdentity?: string
+      autoSelectionPrincipal?: string
     }
     ```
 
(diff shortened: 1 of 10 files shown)


docs/ii-spec.mdx
```diff
@@ -172,6 +173,8 @@ This section describes the Internet Identity Service from the point of view of a
 
     -   the `derivationOrigin`, if present, indicates an origin that should be used for principal derivation instead of the client origin. Internet Identity will only accept values that are also listed in the HTTP resource `/.well-known/ii-alternative-origins` of the corresponding canister (see [Alternative Frontend Origins](#alternative-frontend-origins)).
 
+    -   the `autoSelectMatchingIdentity`, if present, indicates the textual representation of this dapp's principal for which the delegation is requested. If it is known to Internet Identity, it will skip the identity selection and immediately prompt for authentication. This feature can be used to streamline re-authentication after a session expiry.
```
lmuntaner, line 176: Do you have a list of changes downstream for this?

First thoughts are auth-client and [II docs](https://internetcomputer.org/docs/current/references/ii-spec/#client-authentication-protocol)

Anything else?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -565,16 +576,29 @@ export const authnScreens = (i18n: I18n, props: AuthnTemplates) => {
             resolve({ tag: "recover", userNumber }),
         })
       ),
-    pick: (pickProps: { anchors: NonEmptyArray<bigint> }) =>
+    pick: (pickProps: {
+      anchors: NonEmptyArray<bigint>;
+      autoSelect?: bigint;
```
lmuntaner, line 581: `autoSelect` seems like a boolean type of parameter. How about `selectedIdentity`?

src/frontend/src/components/authenticateBox.ts
```diff
@@ -565,16 +576,29 @@ export const authnScreens = (i18n: I18n, props: AuthnTemplates) => {
...
+    }) =>
       new Promise<
         { tag: "more_options" } | { tag: "pick"; userNumber: bigint }
-      >((resolve) =>
+      >((resolve) => {
+        // render page first so that when the identity is picked and the passkey
```
lmuntaner, line 586: What is it that we need to wait? Couldn't we show a spinner while we wait for it and then resolve?

src/frontend/src/flows/authorize/index.ts
```diff
@@ -188,6 +188,14 @@ const authenticate = async (
     };
   }
 
+  let autoSelectIdentity = undefined;
+  if (nonNullish(authContext.authRequest.autoSelectMatchingIdentity)) {
+    autoSelectIdentity = await getAnchorByPrincipal({
```
lmuntaner, line 193: Same here, this is an identity, now? Why call it `auto`?

src/frontend/src/flows/authorize/postMessageInterface.ts
```diff
@@ -1,5 +1,6 @@
 // Types and functions related to the window post message interface used by
 // applications that want to authenticate the user using Internet Identity
+import { zodPrincipal } from "@dfinity/internet-identity-vc-api";
```
lmuntaner, line 3: Should we put it somewhere else? `zodPrincipal` doesn't seem related to `internet-identity-vc-api` to me. I don't think I would remember to import it from here.

Maybe not in this PR though.

lmuntaner (commented):
Why do you call it `auto`, it sounds like a boolean to me? Why not only `selectedIdentity`?

Using the `matching` part, it seems as if one could pass a regex.

I also added some minor comments.

src/frontend/src/components/authenticateBox.ts
frederikrothenberger: Well, normally we wait for the user to chose. But if we have an a `autoSelect` identity, then the action completes immediately. I am just threading it through here, because the identity picker should still be rendered, before automatically selecting something. Otherwise you have the passkey interaction from a completely blank page, which is pretty weird. This way, you have at least a (albeit blurred out) version of the II page in the background.

src/frontend/src/flows/authorize/index.ts
frederikrothenberger: Well, yes. `getAnchorByPrincipal` resolves the supplied Principal to an identity (number).

src/frontend/src/flows/authorize/postMessageInterface.ts
frederikrothenberger: Yeah, I thought about it. But then we need to have yet another package that only provides the `zodPrincipal`. Maybe we could put that into `ic-js` AFAIK @peterpeterparker also needs this definition in the signer library.

docs/ii-spec.mdx
frederikrothenberger: As discussed, we only need to have the docs rebuilt after merging.

frederikrothenberger:
> Why do you call it `auto`, it sounds like a boolean to me? Why not only `selectedIdentity`?

Because it automatically selects the identity when loading the page (in the future). And it is not guaranteed that it finds the apropriate identity to select. Also, the user can override that selection.
 
> Using the `matching` part, it seems as if one could pass a regex.

The rp can only supply a principal (that we then need to match against known values). We can also call it `autoSelectionPrincipal`?

src/frontend/src/components/authenticateBox.ts
lmuntaner: We could render something else if we want to improve it I guess. But definitely not a priority.

src/frontend/src/flows/authorize/postMessageInterface.ts
lmuntaner: Seems weird there as well. I was thinking more in II directly. I think that repeating this code in multiple places is also ok.

lmuntaner (approved):
LGTM! Thanks for the answers. I don't have a better name, so let's not wait and merge this.

frederikrothenberger:
> LGTM! Thanks for the answers. I don't have a better name, so let's not wait and merge this.

After some consideration I like `autoSelectionPrincipal` better. I have renamed the field and duplicated the `zodPrincipal`. The rest is still the same.

Thanks for the feedback!

PR #2567, opened 2024-08-22
Simplify identity metadata repository

frederikrothenberger:
The identity metadata repository had special polling logic to time out loading the metadata after 10 seconds.
This introduced flakiness in the end-to-end test because sometimes loading would be too slow resulting in the recoveryWizard showing the recovery reminder page instead.

The PR makes the following changes:
* remove the polling, simply `await` the promise when needed
  * this will fix e2e tests and simplify the code while it should not impact user experience too much (10 seconds timeout was almost never hit, otherwise users would have complained).
* do not wait for the commitMetadata to complete


lmuntaner (approved):
LGTM!

PR #2575, opened 2024-08-27
Fix errors in the alternative origins documentation

frederikrothenberger:
The alternative origins documenation had some errors and unclear sections. This PR improves the documentation.

### DIFF
--- a/docs/ii-spec.mdx
+++ b/docs/ii-spec.mdx
@@ -226,8 +226,8 @@ The Internet Identity frontend will use `event.origin` as the "Frontend URL" to 
 
 ## Alternative Frontend Origins
 
-To allow flexibility regarding the canister frontend URL, the client may choose to provide another frontend URL as the `derivationOrigin` (see [Client authentication protocol](#client-authentication-protocol)). This means that Internet Identity will issue the same principals to the frontend (which uses a different origin) as it would if it were using the `derivationOrigin` directly.
-This feature works for `canisterId` based URLs and [ICP custom domains](https://internetcomputer.org/docs/current/developer-docs/web-apps/custom-domains/using-custom-domains) backed by canisters.
+To allow flexibility regarding the canister frontend URL, the client may choose to provide another URL as the `derivationOrigin` (see [Client authentication protocol](#client-authentication-protocol)). This means that Internet Identity will issue the same principals to the frontend (which uses a different origin) as it would if it were using the `derivationOrigin` directly.
+This feature works for `canisterId` based URLs and [ICP custom domains](https://internetcomputer.org/docs/current/developer-docs/web-apps/custom-domains/using-custom-domains) (backed by canisters) as `derivationOrigin`.
 
 :::caution
 This feature is intended to allow more flexibility with respect to the origins of a _single_ service. Do _not_ use this feature to allow _third party_ services to use the same principals. Only add origins you fully control to `/.well-known/ii-alternative-origins` and never set origins you do not control as `derivationOrigin`!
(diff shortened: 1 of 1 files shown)


timohanke:
```
To allow flexibility regarding the canister frontend URL, the client may choose to provide another frontend URL as the `derivationOrigin` (see [Client authentication protocol](#client-authentication-protocol)). This means that Internet Identity will issue the same principals to the frontend (which uses a different origin) as it would if it were using the `derivationOrigin` directly.
This feature works for `canisterId` based URLs and [ICP custom domains](https://internetcomputer.org/docs/current/developer-docs/web-apps/custom-domains/using-custom-domains) backed by canisters.
```

In the last sentence: This feature works for [..] URLs and ICP custom domains as what? 
I think we have to say "works for canisterId based URLs and ICP custom domains as canister frontend URL". To make sure people don't read it as "works for canisterId based URLs and ICP custom domains as derivationOrigin".

In the first sentence: "To allow flexibility regarding the canister frontend URL, the client may choose to provide another frontend URL as the `derivationOrigin`" shouldn't we say "To allow flexibility regarding the canister frontend URL, the client may choose to provide another URL as the `derivationOrigin`" because the derivationOrigin doesn't actually have to host a frontend? It can host one but doesn't have to.

timohanke:
> In order for Internet Identity to accept the `derivationOrigin` the origin of the client application must be listed in the JSON object served on the URL `https://<canister_id>.icp0.io/.well-known/ii-alternative-origins` (i.e. the file must be hosted by an ICP canister that must implement the `http_request` query call as specified [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec)).

I think here we should say what `<canister_id>` is. Because in the paragraph that precedes this paragraph `<canister_id>` is used in more generality. Here in this paragraph actually `<canister_id>.icp0.io` is the derivationOrigin.

What about:
In order for Internet Identity to accept the `derivationOrigin` the `derivationOrigin` must have the form `<canister_id>.icp0.io` and the origin of the client application must be listed in the JSON object served on the URL `https://<canister_id>.icp0.io/.well-known/ii-alternative-origins` (i.e. the file must be hosted by an ICP canister that must implement the `http_request` query call as specified [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec)).

frederikrothenberger:
Thanks for the feedback!

> > To allow flexibility regarding the canister frontend URL, the client may choose to provide another frontend URL as the `derivationOrigin` (see [Client authentication protocol](#client-authentication-protocol)). This means that Internet Identity will issue the same principals to the frontend (which uses a different origin) as it would if it were using the `derivationOrigin` directly.
> > This feature works for `canisterId` based URLs and [ICP custom domains](https://internetcomputer.org/docs/current/developer-docs/web-apps/custom-domains/using-custom-domains) backed by canisters.
> 
> In the last sentence: This feature works for [..] URLs and ICP custom domains [..] as what? I think we have to say "works for canisterId based URLs and ICP custom domains as canister frontend URL". To make sure people don't read it as "works for canisterId based URLs and ICP custom domains as derivationOrigin".

But that would actually be correct: there is no restriction on the front-end URL, and the restriction to canisterId base URLs / custom domains for the `derivationOrigin`. II is only picky about where it fetches the alternative origins file from, not about who sends the tab postMessages.

Anyway, I'll clarify.

> In the first sentence:
> 
> > To allow flexibility regarding the canister frontend URL, the client may choose to provide another frontend URL as the `derivationOrigin`
> 
> shouldn't we say "To allow flexibility regarding the canister frontend URL, the client may choose to provide another URL as the `derivationOrigin`" because the derivationOrigin doesn't actually have to host a frontend? It can host one but doesn't have to.

Yes, thanks.

timohanke:
> But that would actually be correct: there is no restriction on the front-end URL, and the restriction to canisterId base URLs / custom domains for the derivationOrigin. II is only picky about where it fetches the alternative origins file from, not about who sends the tab postMessages.

Right, so the sentence ""works for canisterId based URLs and ICP custom domains as canister frontend URL" would be misleading because there is no restriction whatsoever on the frontend URL.

And the sentence "works for canisterId based URLs and ICP custom domains as derivationOrigin" is correct? I can set derivationOrigin to a custom domain (which is off the IC)?

frederikrothenberger:
> I can set derivationOrigin to a custom domain (which is off the IC)?

It's not possible to register an ICP custom domain that does _not_ point to a canister. II will validate that the custom domain resolves to a canisterId.

PR #2578, opened 2024-08-29
Separate principal and origin information in IndexedDB storage

frederikrothenberger:
This PR introduces the storage layout v4 which keeps separate digests for principals and origins. That way it will be possible in the future to determine whether a specific principal was last derived for a particular origin.

### DIFF
--- a/src/frontend/src/storage/index.test.ts
+++ b/src/frontend/src/storage/index.test.ts
@@ -89,7 +89,7 @@ test(
         after: (storage) => {
           // Written to V3
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          const storageV3: any = storage["ii-storage-v3"];
+          const storageV3: any = storage["ii-storage-v4"];
           expect(storageV3).toBeTypeOf("object");
 
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -146,7 +146,46 @@ test(
 );
 
 test(
-  "anchors are also written to V2",
+  "V3 anchors are migrated",
+  withStorage(
+    async () => {
+      expect(await getAnchors()).toContain(BigInt(10000));
+      expect(await getAnchors()).toContain(BigInt(10001));
+      expect(await getAnchors()).toContain(BigInt(10003));
+    },
+    {
+      indexeddb: {
+        before: {
+          /* V3 layout */
+          "ii-storage-v3": {
+            anchors: {
+              "10000": { lastUsedTimestamp: 0, knownPrincipals: [] },
+              "10001": { lastUsedTimestamp: 0, knownPrincipals: [] },
+              "10003": { lastUsedTimestamp: 0, knownPrincipals: [] },
+            },
+            hasher: await crypto.subtle.generateKey(
+              { name: "HMAC", hash: "SHA-512" },
+              false /* not extractable */,
+              ["sign"] /* only used to "sign" (e.g. produce a digest ) */
+            ),
+          },
+        },
+        after: (storage) => {
+          // Written to V3
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const storageV4: any = storage["ii-storage-v4"];
+          expect(storageV4).toBeTypeOf("object");
+          expect(storageV4.anchors["10000"]).toBeDefined();
+          expect(storageV4.anchors["10001"]).toBeDefined();
+          expect(storageV4.anchors["10003"]).toBeDefined();
+        },
+      },
+    }
+  )
+);
+
+test(
+  "anchors are also written to V3",
   withStorage(
     async () => {
       await setAnchorUsed(BigInt(10000));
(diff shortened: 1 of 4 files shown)


src/frontend/src/storage/index.test.ts
```diff
@@ -89,7 +89,7 @@ test(
         after: (storage) => {
           // Written to V3
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          const storageV3: any = storage["ii-storage-v3"];
+          const storageV3: any = storage["ii-storage-v4"];
```
lmuntaner, line 92: Maybe rename to `storageV4`?

src/frontend/src/storage/index.test.ts
```diff
@@ -146,7 +146,46 @@ test(
...
+            },
+            hasher: await crypto.subtle.generateKey(
+              { name: "HMAC", hash: "SHA-512" },
+              false /* not extractable */,
+              ["sign"] /* only used to "sign" (e.g. produce a digest ) */
+            ),
+          },
+        },
+        after: (storage) => {
+          // Written to V3
```
lmuntaner, line 174: What does this comment mean?

src/frontend/src/storage/index.test.ts
```diff
@@ -146,7 +146,46 @@ test(
...
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const storageV4: any = storage["ii-storage-v4"];
+          expect(storageV4).toBeTypeOf("object");
+          expect(storageV4.anchors["10000"]).toBeDefined();
+          expect(storageV4.anchors["10001"]).toBeDefined();
+          expect(storageV4.anchors["10003"]).toBeDefined();
```
lmuntaner, line 180: Could you also check here that `getAnchorByPrincipal` works as expected?

lmuntaner (approved):
LGTM, just minor comments

src/frontend/src/storage/index.test.ts
frederikrothenberger: Ah sorry, this is misleading it is read from v3 and written to v4

src/frontend/src/storage/index.test.ts
frederikrothenberger: No, known principals are _not_ migrated. They cant be migrated, because the digest scheme has changed.

lmuntaner: Using `getAnchorByPrincipal` on `10003` won't work in this scenario?

frederikrothenberger: No, it will return undefined. But the VC flow is currently the only usage of that (and it will be restored after the next login) so the impact for users should be acceptable.

lmuntaner: Ok, understood. Thanks for the explanation.

src/frontend/src/storage/index.ts
```diff
@@ -269,18 +280,46 @@ const nowMillis = (): number => {
...
   hasher: CryptoKey;
+}): Promise<string> => {
+  // Encode each element & size
+  const enc = new TextEncoder();
+  const principal_ = principalObj.toText();
+
+  const principal = enc.encode(principal_);
+  const principalLen = Uint8Array.from([principal_.length]);
+
+  // Create a buffer with all four elements
```
cybrowl, line 300: @frederikrothenberger `with two elements`

src/frontend/src/storage/index.ts
```diff
@@ -294,21 +333,16 @@ const computePrincipalDigest = async ({
...
   const enc = new TextEncoder();
-  const principal_ = principalObj.toText();
 
   const origin = enc.encode(origin_);
   const originLen = Uint8Array.from([origin_.length]);
 
-  const principal = enc.encode(principal_);
-  const principalLen = Uint8Array.from([principal_.length]);
-
   // Create a buffer with all four elements
```
cybrowl, line 340: @frederikrothenberger `with two elements`

src/frontend/src/storage/index.ts
```diff
@@ -533,18 +547,97 @@ const StorageV3 = z.object({
...
+    migratedAnchors[userNumber] = { ...oldAnchor, knownPrincipals: [] };
+  }
+
+  return { anchors: migratedAnchors, hasher: storage.hasher };
+};
+
+/**
+ * V4, indexeddb["ii-storage-v4"] = { anchors: { 20000: ... } }
+ * */
+type StorageV4 = z.infer<typeof StorageV4>;
```
cybrowl, line 589: @frederikrothenberger Why is there a V3 and a V4?

src/frontend/src/storage/index.ts
frederikrothenberger: Thanks, addressed in #2581.

src/frontend/src/storage/index.ts
frederikrothenberger: The requirements have changed. The V3 (which was in use since December 2023) no longer stored enough information for the auto-selection for the most recently used principal.

Hence it was changed to V4, which stores the principal and origin digests separately. We still maintain compatibility with old storage layouts to prevent people losing their prepopulated identity numbers.

PR #2579, opened 2024-08-29
Only allow the most recently used identity to be auto-selected

frederikrothenberger:
This is an enhancement to the auto-selection feature introduced in #2563: In order to not confuse users, only the most recently used identity can be auto-selected (i.e. when refreshing sessions). This way, a dapp cannot make a user _switch_ identities without them having the identity selected explicitly.

### DIFF
--- a/src/frontend/src/components/authenticateBox/index.ts
+++ b/src/frontend/src/components/authenticateBox/index.ts
@@ -431,7 +431,11 @@ export const authnTemplates = (i18n: I18n, props: AuthnTemplates) => {
 
         <ul class="c-link-group">
           <li>
-            <button @click=${() => useExistingProps.register()} class="t-link">
+            <button
+              @click=${() => useExistingProps.register()}
+              id="registerButton"
+              class="t-link"
+            >
               Create New
             </button>
           </li>
--- a/src/frontend/src/storage/index.test.ts
+++ b/src/frontend/src/storage/index.test.ts
@@ -334,13 +334,51 @@ test(
   "should not retrieve anchor if not most recent",
   withStorage(async () => {
     const origin = "https://example.com";
+    const oldPrincipal = Principal.fromText(
+      "hawxh-fq2bo-p5sh7-mmgol-l3vtr-f72w2-q335t-dcbni-2n25p-xhusp-fqe"
+    );
+    vi.useFakeTimers().setSystemTime(new Date(0));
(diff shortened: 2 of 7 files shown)


src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -431,7 +431,11 @@ export const authnTemplates = (i18n: I18n, props: AuthnTemplates) => {
 
         <ul class="c-link-group">
           <li>
-            <button @click=${() => useExistingProps.register()} class="t-link">
+            <button
+              @click=${() => useExistingProps.register()}
+              id="registerButton"
```
lmuntaner, line 436: Why the `id`?

src/frontend/src/storage/index.test.ts
```diff
@@ -311,6 +312,70 @@ test(
...
+test(
+  "should not retrieve anchor if not most recent",
+  withStorage(async () => {
+    const origin = "https://example.com";
+    const principal1 = Principal.fromText(
+      "hawxh-fq2bo-p5sh7-mmgol-l3vtr-f72w2-q335t-dcbni-2n25p-xhusp-fqe"
+    );
+    vi.useFakeTimers().setSystemTime(new Date(0));
+    await setKnownPrincipal({
+      userNumber: BigInt(10000),
```
lmuntaner, line 342: Can we put `BitInt(10000)` and `BigInt(1000)` in variables to better read the expectations?

src/frontend/src/storage/index.test.ts
```diff
@@ -311,6 +312,70 @@ test(
...
+      await getAnchorIfLastUsed({ principal: principal1, origin })
+    ).not.toBeDefined();
+    // most recent principal still works
+    expect(await getAnchorIfLastUsed({ principal: principal2, origin })).toBe(
+      BigInt(10001)
+    );
+  })
+);
+
+test(
```
lmuntaner, line 367: Could you add a test adding the principal in another origin?

lmuntaner (approved):
LBTM! Just a few comments

src/frontend/src/components/authenticateBox/index.ts
frederikrothenberger: I added the id there so that the e2e test button click logic for the register button works the same regardless of the template (`firstTime` / `useExisting`).

PR #2583, opened 2024-09-03
Remove unused vc_util_js crate

frederikrothenberger:
The crate is removed because it was not used and only contained a single dummy function.


lmuntaner (approved):
Thanks!

PR #2586, opened 2024-09-04
Remove dependency on the `ic-crypto-standalone-sig-verifier`

frederikrothenberger:
This PR removes the `ic-crypto-standalone-sig-verifier` dependency, because it was one of the last dependency that II had on the ic repository. There are now properly published alternatives available (also maintained by the crypto team).

However, the published package only supports canister signatures at the moment, the feature set of `sig-verifier-js` is reduced to only verifying canister signatures too. This is sufficient to integrate with II in a web2 context (which is what this library was initially intended for).


lmuntaner (approved):
LGTM!

PR #2591, opened 2024-09-05
Update rust dependencies

frederikrothenberger:
This updates all the locked dependencies to the latest compatible version. It has been a while since we have last done this.


peterpeterparker (approved):
Rubber stamp

PR #2592, opened 2024-09-05
Update vc_issuer dependencies

frederikrothenberger:
This updates all the dependencies of the vc_issuers to the latest version. And removes the unused dependency on `ic-crypto-standalone-sig-verifier`.

Unfortunately, I had to add `ic-crypto-getrandom-for-wasm` as we now have a transitive dependency on `getrandom`.


lmuntaner (approved):
Thanks!

PR #2599, opened 2024-09-10
Update ic-wasm

frederikrothenberger:
`ic-wasm` has not been updated for a long time and has become outdated.

### DIFF
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -37,7 +37,7 @@ runs:
         path: |
           ~/.cargo/bin/ic-wasm
           ~/.rustup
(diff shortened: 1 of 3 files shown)


peterpeterparker (approved):
LGTM, thanks

That said, from ic-wasm [CHANGELOG](https://github.com/dfinity/ic-wasm) it's really hard to guess what changed and what are the new features. Not even sure if there wasn't a breaking change.

frederikrothenberger:
> That said, from ic-wasm [CHANGELOG](https://github.com/dfinity/ic-wasm) it's really hard to guess what changed and what are the new features. Not even sure if there wasn't a breaking change.

Agreed, but given the tests still run and compression still works, I don't think anything broke. If the release candidate also still works, then we should be good.

PR #2600, opened 2024-09-10
Remove archive stable memory limit

frederikrothenberger:
For historical reasons, the archive stable memory was limited to 32 GB. Since in practice the canister has access to much more storage, this artificial limit is now dropped.


lmuntaner (approved):
Thanks!

PR #2602, opened 2024-09-11
Add config endpoint

frederikrothenberger:
Adds an endpoint to query the currently active config (i.e. `InternetIdentityInit`).


lmuntaner (approved):
Interesting that the `InternetIdentityInit` is not stored all in one place but multiple ones.

frederikrothenberger:
> Interesting that the `InternetIdentityInit` is not stored all in one place but multiple ones.

Yeah, that should probably be cleaned up 🙃

PR #2610, opened 2024-09-13
Add configuration options for upcoming dynamic captcha feature

frederikrothenberger:
This PR adds the required configuration options to make the captcha
(on registration) dynamic: once implemented, if the dynamic config is
enabled, the captcha will only be required on unusually high rate of new
registrations.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -209,6 +209,31 @@ type RateLimitConfig = record {
     max_tokens: nat64;
 };
 
+// Captcha configuration
+type CaptchaConfig = record {
+    // Maximum number of unsolved captchas.
+    // Default: 500
+    max_unsolved_captchas : nat64;
+    // Configuration for when captcha protection should kick in.
+    captcha_trigger: variant {
+      // Based on the rate of registrations compared to some reference time frame and allowing some leeway.
+      Dynamic: record {
+        // Percentage of increased registration rate observed in the current rate sampling interval (compared to
+        // reference rate) at which II will enable captcha for new registrations.
+        threshold_pct: nat16;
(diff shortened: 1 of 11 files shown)


src/internet_identity/internet_identity.did
```diff
@@ -209,6 +209,31 @@ type RateLimitConfig = record {
     max_tokens: nat64;
 };
 
+// Captcha configuration
+type CaptchaConfig = record {
+    // Maximum number of unsolved captchas.
+    // Default: 500
```
lmuntaner, line 215: If there is a default, shouldn't it be optional?

src/internet_identity/tests/integration/config.rs
```diff
@@ -21,7 +21,14 @@ fn should_retain_anchor_on_user_range_change() -> Result<(), CallError> {
             time_per_token_ns: 99,
             max_tokens: 874,
         }),
-        max_inflight_captchas: Some(456),
+        captcha_config: Some(CaptchaConfig {
+            max_unsolved_captchas: 788,
```
lmuntaner, line 25: Do the numbers mean anything? it was `456` before... I guess they don't, right?

lmuntaner (approved):
Looking good!

I just added a couple of questions.

src/internet_identity/internet_identity.did
frederikrothenberger: Ah, yes. Moved the default documentation to the entire `CaptchaConfig`.

src/internet_identity/tests/integration/config.rs
frederikrothenberger: No, it's just different from the default, making sure that the values are not accidentally the same.

Thompson1985 (approved):

PR #2611, opened 2024-09-13
Fix broken note blocks in documentation

frederikrothenberger:
The "API V2" note blocks are not rendered correctly on the official docs page: https://internetcomputer.org/docs/current/references/ii-spec#the-internet-identity-service-backend-interface

The reason is that the note block start and terminator apparently cannot be on the same line.


lmuntaner (approved):
Thanks!

PR #2616, opened 2024-09-18
Derive default for `State`

frederikrothenberger:
There is no need for a manual implementation of `Default` for `State`. It existing is distracting, hence it is removed.


lmuntaner (approved):
Thanks!

PR #2618, opened 2024-09-20
Refactor temp_keys in preparation for dynamic captcha registration flow

frederikrothenberger:
For the dynamic captcha registration flow, temp_keys need to be tracked before the identity number or authn method public key is known. For that reason, this refactoring abstracts out the `TempKeyId` such that we can introduce a different variant to track the keys.
For example a registration flow id.

### DIFF
--- a/src/internet_identity/src/state/temp_keys.rs
+++ b/src/internet_identity/src/state/temp_keys.rs
@@ -1,11 +1,35 @@
 use crate::MINUTE_NS;
 use candid::Principal;
 use ic_cdk::api::time;
-use internet_identity_interface::internet_identity::types::{AnchorNumber, DeviceKey, Timestamp};
+use internet_identity_interface::internet_identity::types::{IdentityNumber, PublicKey, Timestamp};
 use std::collections::{HashMap, VecDeque};
 
 // Expiration for temp keys, the same as the front-end delegation expiry
 const TEMP_KEY_EXPIRATION_NS: u64 = 10 * MINUTE_NS;
+
+/// A temp key identifier
+#[derive(Debug, Clone, Eq, PartialOrd, PartialEq, Hash)]
+pub enum TempKeyId {
+    /// A temp key that is tied to an existing identity and authn_method.
+    /// We need to track the authn_method because the associated temp_key
+    /// must become invalid if the authn_method is removed from the identity.
+    IdentityAuthnMethod {
+        identity_number: IdentityNumber,
+        authn_method_pubkey: PublicKey,
+    },
+}
+
+impl TempKeyId {
+    pub fn from_identity_authn_method(
+        identity_number: IdentityNumber,
+        authn_method_pubkey: PublicKey,
+    ) -> Self {
+        TempKeyId::IdentityAuthnMethod {
+            identity_number,
+            authn_method_pubkey,
+        }
+    }
+}
 
 #[derive(Default, Debug)]
 pub struct TempKeys {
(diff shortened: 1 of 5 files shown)


lmuntaner (commented):
Can you explain me the reason for these changes? I don't understand why this is needed when the captcha is removed.

Some questions:
* Why "temp_keys need to be tracked"?
* What does "a different variant to track the keys" mean? A variant where? of what?

And in general, the concept of TempKeys and its flow.

frederikrothenberger:
So the `temp_keys` were introduced when the second passkey interaction during registration was removed. The WebAuthn `create` and `get` calls both need the user to interact with the Passkey, but produce slightly different signatures. The ICP only accepts the signatures from `get`.

To not require a second interaction, the ephemeral session key pair is added as a `temp_key` at the same time as the first passkey is registered and it acts as a replacement for a delegation from the passkey for the first sessions (for calls going to the II canister). For subsequent sessions, the Passkey `get` operation will sign the delegation for the ephemeral session key and thus no temp_key is needed anymore.

Now, to figure out, whether a particular `caller` corresponds to a temp_key for any given identity, we need to be able to look up temp_keys by that (plus the device key). This is what I meant by "tracking the temp_keys" by some id.

Now, with the new registration flow, where the captcha is optional, the different pieces of information are split into different calls (as opposed to the `register` call, that allocates the user, checks the captcha and records the temp_key, all in one go).

The registration flow is _started_ by submitting a temp_key which will be tied to a registration flow id (as opposed to an identity number and public key). Only the `caller` that corresponds to the temp_key is allowed to make progress on that particular registration flow.

The second step (which might be skipped) is solving the captcha.

The third is to transition from a registration flow id to and identity with a passkey. In that step, the temp_key will be rebound to the identity number and passkey pubkey to continue the session.

So, why not switch the order of things and check the captcha after the passkey has been added? This is because we would need to reserve a identity number for that, which would use stable memory, as the current storage model does not allow for gaps in identity numbers. As a consequence, stable memory would be wasted by registrations being started but never getting past the captcha.

Does this make sense?

src/internet_identity/src/state/temp_keys.rs
```diff
@@ -1,12 +1,36 @@
...
-use internet_identity_interface::internet_identity::types::{AnchorNumber, DeviceKey, Timestamp};
+use internet_identity_interface::internet_identity::types::{IdentityNumber, PublicKey, Timestamp};
 use std::collections::{HashMap, VecDeque};
 
 // Expiration for temp keys, the same as the front-end delegation expiry
 const TEMP_KEY_EXPIRATION_NS: u64 = 10 * MINUTE_NS;
 
+/// A temp key identifier
+#[derive(Debug, Clone, Eq, PartialOrd, PartialEq, Hash)]
+pub enum TempKeyId {
```
frederikrothenberger, line 12: Having an enum here, allows to introduce a `FlowId(String)` variant later to be able to look up `temp_key` by registration flow.

src/internet_identity/src/state/temp_keys.rs
```diff
@@ -17,59 +41,49 @@ pub struct TempKeys {
...
     /// authenticate if the actual delegated WebAuthn device is removed).
     ///
     /// In addition, the temporary key is also linked to an anchor so that even if the same device is
     /// added to multiple anchors, it can only be used in the context of the anchor it was created for.
     ///
-    /// Since temp keys can only be added during registration, the the max number of temp keys is
+    /// Since temp keys can only be added during registration, the max number of temp keys is
     /// bounded by the registration rate limit.
-    temp_keys: HashMap<(AnchorNumber, DeviceKey), TempKey>,
+    temp_keys: HashMap<TempKeyId, TempKey>,
```
lmuntaner, line 53: This is stored in the heap and reset after every upgrade, right?

lmuntaner (approved):
LGTM! Just one small question/confirmation

src/internet_identity/src/state/temp_keys.rs
frederikrothenberger: Yes, exactly.

PR #2619, opened 2024-09-23
Extract captcha logic into new module

frederikrothenberger:
In preparation for the dynamic captcha feature, the captcha creation logic is extracted into it's own module. This makes it easier to reuse the specific parts needed.

This PR only _moves_ code.


lmuntaner (approved):
Thanks!

PR #2624, opened 2024-09-24
Update agent-js

frederikrothenberger:
`npm update` clashes with incompatible peer dependencies due to agent-js being on version 1.x.x but `@dfinity/utils` wanting 2.x.x. Hence agent-js and related dependencies are upgraded.


peterpeterparker (commented):
There were breaking changes in the way Agent creation works, introduced in v2. Doesn't II's codebase need some adaptation for this? Did you double-check?

frederikrothenberger:
> There were breaking changes in the way Agent creation works, introduced in v2. Doesn't II's codebase need some adaptation for this? Did you double-check?

Yes, you are right. The old way still works as it was just deprecated. But switching is definitely better. Thanks.

demos/using-dev-build/webapp/index.ts
```diff
@@ -61,7 +61,7 @@ document.getElementById("loginBtn")?.addEventListener("click", async () => {
   // At this point we're authenticated, and we can get the identity from the auth client:
   const identity = authClient.getIdentity();
   // Using the identity obtained from the auth client, we can create an agent to interact with the IC.
-  const agent = new HttpAgent({ identity });
+  const agent = await HttpAgent.create({ identity, shouldFetchRootKey: true });
```
peterpeterparker, line 64: Always `shouldFetchRootKey`?

frederikrothenberger: Yes, this is in the dev build demo, which runs on the local replica and hence always needs fetching the root key.

frederikrothenberger: Additional context: there was no `fetchRootKey` before in the example. However, that only worked because the `whoami` call was a query call, which is uncertified.

Anyone expanding that example with an update call would immediately run into an issue. Hence the `fetchRootKey` is added now.

PR #2627, opened 2024-09-25
Simplify authorization error

frederikrothenberger:
There is no good reason for the (internal) `AuthorizationEror` to be an enum and it complicates things. Hence it is changed to a struct.


lmuntaner (approved):
LGTM

PR #2628, opened 2024-09-25
Consistently use // instead of /// in candid files

frederikrothenberger:
In candid /// has no additional meaning over //.
Given // is shorter and used more prevalently in the codebase we switch everything to //.


lmuntaner (approved):
Thanks!

PR #2629, opened 2024-09-26
Change registration on API v2 to flow with 3 steps

frederikrothenberger:
The registration for new identities using API v2 is now split into 3 different steps:
1. start a flow, providing a principal to track progress on that flow
2. solve captcha
3. submit authn_method to tie the identity to

This PR is a preparation to making the step 2 dynamically remove step 2 on low load.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -955,8 +955,23 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-didc
+      - name: 'Get latest release'
+        uses: actions/github-script@v7
+        id: latest-release-tag
+        with:
+          result-encoding: string
+          script: return (await github.rest.repos.getLatestRelease({owner:"dfinity", repo:"internet-identity"})).data.tag_name;
       - name: "Check canister interface compatibility"
         run: |
+          release="release-2024-09-17"
+          # undo the breaking changes that we _explicitly_ made
+          # remove after the next release
+          # if we accidentally introduced other breaking changes, the patch would no longer apply / fix them
+          # making this job fail.
+          if [ "${{ steps.latest-release-tag.outputs.result }}" == "$release" ]; then
+            echo "Rolling back intentionally made breaking changes $release"
+            git apply allowed_breaking_change.patch
+          fi
           curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity.did -o internet_identity_previous.did
           didc check src/internet_identity/internet_identity.did internet_identity_previous.did
 
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -435,17 +435,6 @@ type IdentityInfoError = variant {
     InternalCanisterError: text;
 };
 
-
-
-type IdentityRegisterError = variant {
-    // No more registrations are possible in this instance of the II service canister.
-    CanisterFull;
-    // The captcha check was not successful.
-    BadCaptcha;
-    // The metadata of the provided authentication method contains invalid entries.
-    InvalidMetadata: text;
-};
-
 type AuthnMethodAddError = variant {
     InvalidMetadata: text;
 };
@@ -532,6 +521,72 @@ type SignedIdAlias = record {
     id_dapp : principal;
 };
 
+type IdRegNextStepResult = record {
+    // The next step in the registration flow
+    next_step: RegistrationFlowNextStep;
+};
+
+type IdRegStartError = variant {
+    // The method was called anonymously, which is not supported.
+    InvalidCaller;
+    // Too many registrations. Please try again later.
+    RateLimitExceeded;
+    // A registration flow is already in progress.
+    AlreadyInProgress;
+};
+
+// The next step in the registration flow:
+// - CheckCaptcha: supply the solution to the captcha using `check_captcha`
+// - Finish: finish the registration using `identity_registration_finish`
+type RegistrationFlowNextStep = variant {
+    // Supply the captcha solution using check_captcha
+    CheckCaptcha: record {
+        captcha_png_base64: text;
+    };
+    // Finish the registration using identity_registration_finish
+    Finish;
+};
+
+type CheckCaptchaArg = record {
+    solution : text;
+};
+
+type CheckCaptchaError = variant {
+    // The supplied solution was wrong. Try again with the new captcha.
+    WrongSolution: record {
+        new_captcha_png_base64: text;
+    };
+    // This call is unexpected, see next_step.
+    UnexpectedCall: record {
+        next_step: RegistrationFlowNextStep;
+    };
+    // No registration flow ongoing for the caller.
+    NoRegistrationFlow;
+};
+
+type IdRegFinishArg = record {
+    authn_method: AuthnMethodData;
+};
+
+type IdRegFinishResult = record {
+    identity_number: nat64;
+};
+
+type IdRegFinishError = variant {
+    // The configured maximum number of identities has been reached.
+    IdentityLimitReached;
+    // This call is unexpected, see next_step.
+    UnexpectedCall: record {
+        next_step: RegistrationFlowNextStep;
+    };
+    // No registration flow ongoing for the caller.
+    NoRegistrationFlow;
+    // The supplied authn_method is not valid.
+    InvalidAuthnMethod: text;
+    // Error while persisting the new identity.
+    StorageError: text;
+};
+
 service : (opt InternetIdentityInit) -> {
     // Legacy identity management API
     // ==============================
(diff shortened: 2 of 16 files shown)


src/internet_identity/src/state/flow_states.rs
```diff
@@ -0,0 +1,131 @@
...
+use candid::Principal;
+use ic_cdk::api::time;
+use internet_identity_interface::internet_identity::types::Timestamp;
+use std::collections::{HashMap, VecDeque};
+
+// Expiration for a registration flow, the same as the captcha expiry
+const FLOW_EXPIRATION_NS: u64 = 5 * MINUTE_NS;
+
+#[derive(Default)]
+pub struct FlowStates {
```
frederikrothenberger, line 12: @lmuntaner: Ultimately, I decided against mixing flow states with temp_keys. I feel this is cleaner in terms of separation.
But it means that the "expiration logic" here and temp_keys is duplicated. I don't think it is big enough to warrant extracting that, but up to you.

src/internet_identity/src/state/flow_states.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// In-progress registration flows tied to the respective principal. The principal will become
+    /// a temp key upon successful completion of the registration flow.
+    /// Registration flows are removed on three conditions:
+    /// * The registration flow is completed successfully
+    /// * Once the flow expires
+    /// * The Internet Identity canister is upgraded
+    ///
+    /// The total number of flow states is limited by the registration rate limit.
```
frederikrothenberger, line 20: Given the captcha is now tied to the flow_state and the number of flow_states is limited by the registration rate limit, the `max_unsolved_captchas` config option in the captcha config only has an effect on the legacy registration flow.

.github/workflows/canister-tests.yml
```diff
@@ -955,8 +955,23 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-didc
+      - name: 'Get latest release'
```
lmuntaner, line 958: Does this needs to be done every time we introduce a breaking change?

Do you have a script or how do you remember how to do it?

src/internet_identity/internet_identity.did
```diff
@@ -532,6 +521,72 @@ type SignedIdAlias = record {
     id_dapp : principal;
 };
 
+type IdRegNextStepResult = record {
```
lmuntaner, line 524: Why not `IdRegStartResult`?

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -0,0 +1,172 @@
...
+    .map_err(|_| IdRegStartError::AlreadyInProgress)?;
+
+    Ok(IdRegNextStepResult {
+        next_step: RegistrationFlowNextStep::CheckCaptcha {
+            captcha_png_base64: captcha_png_base64.0,
+        },
+    })
+}
+
+async fn captcha_flow_state(flow_created_timestamp_ns: u64) -> (Base64, RegistrationFlowState) {
```
lmuntaner, line 55: Will you add that this returns `Option<Base64>` afterwards?

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    let RegistrationFlowState::FinishRegistration { .. } = current_state else {
+        return Err(IdRegFinishError::UnexpectedCall {
+            next_step: RegistrationFlowNextStep::from(current_state),
+        });
+    };
+
+    let identity_number = create_identity(&arg)?;
+
+    let caller = caller();
```
lmuntaner, line 130: Is there a reason not to create this var the first thing and use it in line 117?

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -0,0 +1,172 @@
...
+            },
+            caller,
+        )
+    });
+
+    Ok(IdRegFinishResult { identity_number })
+}
+
+fn create_identity(arg: &IdRegFinishArg) -> Result<IdentityNumber, IdRegFinishError> {
+    let Some(mut identity) = state::storage_borrow_mut(|s| s.allocate_anchor()) else {
```
lmuntaner, line 149: Is this a thing we will need to change to allow creating the identity number at the beginning of the flow instead of the end?

src/internet_identity/src/state/flow_states.rs
lmuntaner: So now the temp keys won't be used to track the registration flow, right?

Do we need the enum for the TempKeyId still?

src/internet_identity/src/state/flow_states.rs
lmuntaner: Doesn't that depend on what comes furst, the registration rate limit or the `max_unsolved_captchas`?

src/internet_identity/src/state/flow_states.rs
```diff
@@ -0,0 +1,131 @@
...
+        captcha_solution: String,
+        captcha_png_base64: Base64,
+    },
+    /// Client is expected to finish the registration (supply an [authn_method](internet_identity_interface::internet_identity::types::AuthnMethodData))
+    /// using `identity_registration_finish`.
+    FinishRegistration { flow_created_timestamp_ns: u64 },
+}
+
+impl RegistrationFlowState {
+    pub fn create_at_timestamp_ns(&self) -> u64 {
```
lmuntaner, line 45: With this names it seems it's creating it.

How about `created_at_timestamp_ns`? with `created` instead of `create`.

src/internet_identity/src/state/flow_states.rs
```diff
@@ -0,0 +1,131 @@
...
+        self.registration_flows.insert(principal, flow_state);
+        Ok(())
+    }
+
+    pub fn remove_registration_flow(&mut self, principal: &Principal) {
+        self.prune_expired_flows();
+        self.registration_flows.remove(principal);
+    }
+
+    pub fn registration_flow_state(&self, principal: &Principal) -> Option<RegistrationFlowState> {
```
lmuntaner, line 96: Should we also prune here?

src/internet_identity/src/state/flow_states.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn prune_expired_flows(&mut self) {
+        const MAX_TO_PRUNE: usize = 100;
+
+        let now = time();
+        for _ in 0..MAX_TO_PRUNE {
+            let Some(expiration) = self.expirations.front() else {
+                break;
+            };
+
+            // The expirations are sorted by expiration time because the expiration is constant
```
lmuntaner, line 113: I would add this line comment above 109.

src/internet_identity/tests/integration/v2_api/identity_register.rs
```diff
@@ -37,6 +38,40 @@ fn should_register_multiple_identities() {
     assert_ne!(identity_number_1, identity_number_2);
 }
 
+#[test]
+fn should_transition_flow_principal_to_temp_key() {
```
lmuntaner, line 42: Why this name for this test?

It's to make sure that the temp_key is expired after the required time, right?

lmuntaner (approved):
I'm happy because after our conversation I was able to follow this PR pretty well. Thanks!

Minor comments, already approving.

.github/workflows/canister-tests.yml
frederikrothenberger: I usually pull it out again from git history whenever I need it. My hope was that it would not be worth it improving the infrastructure to do that, as we would eventually stop making breaking changes (as APIv2 would be finished)...

src/internet_identity/internet_identity.did
frederikrothenberger: I reuse this result for `_start` and `check_captcha`. It is essentially a result to indicate a step has been completed correctly and what the next step is.

So if there are more / different steps in the flow in the future, they can also use this.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
frederikrothenberger: You mean, once the captcha might be skipped? No, I think I will `identity_registration_start` with some logic to decide whether this function should be called or if the next step is `Finish`.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
frederikrothenberger: Actually no, I'll move it up.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
frederikrothenberger: Yes, there will definitely be changes here. But there need to be many more. Maybe its best we have a separate discussion about it.

src/internet_identity/src/state/flow_states.rs
frederikrothenberger: No, I will revert #2618. I'll do that after this is merged.

src/internet_identity/src/state/flow_states.rs
frederikrothenberger: No, `max_unsolved_captchas` is not used at all in the context of the V2 registration flow, as we store the captcha with the flow state.

src/internet_identity/src/state/flow_states.rs
frederikrothenberger: Ah, yes. That is a typo. Thanks!

src/internet_identity/src/state/flow_states.rs
frederikrothenberger: I don't think it is necessary, as we explicitly check `expiration` before returning anything.
IMHO it is more natural to prune alongside other modifications, rather than on read.

src/internet_identity/tests/integration/v2_api/identity_register.rs
frederikrothenberger: The test is there for two things:
1. the principal used in the flow can still be used after the registration (e.g. to add additional devices / get a delegation, etc.)
2. the principal used in the flow expires after the expected amount of time

PR #2631, opened 2024-09-27
Use stable BTreeMap directly to report stats

frederikrothenberger:
When implementing the original stats feature, I missed that `len` exists on the stable `BTreeMap`, leading to a much more complicated implementation for keeping track of the number of event stats related data.

This is the first step of simplifying again (we need to wait one release before removing the duplicated counters for rollback compatibility).


lmuntaner (approved):
Nice!

PR #2633, opened 2024-09-30
Allow fractions in  helper `assert_metric` helper

frederikrothenberger:
The `assert_metric` helper only parses integers right now, even though the metrics could be fractions.
This is required once the registrations rates are exposed as a metric, hence this PR to fix it.


lmuntaner (approved):
Thanks!

Do we have tests for this helper? Could we add some test cases with fractions?

PR #2638, opened 2024-10-02
Only require captcha if threshold rate is exceeded

frederikrothenberger:
This PR enables the dynamic captcha feature: I.e. if the captcha is configured to be dynamic _and_ the configured threshold has been exceeded, then require a captcha to be solved. Otherwise skip the captcha entirely.

### DIFF
--- a/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
+++ b/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
@@ -1,7 +1,8 @@
 use crate::anchor_management::registration::captcha::{
-    check_captcha_solution, create_captcha, make_rng, Base64,
+    check_captcha_solution, create_captcha, make_rng,
 };
 use crate::anchor_management::registration::rate_limit::process_rate_limit;
+use crate::anchor_management::registration::Base64;
 use crate::anchor_management::{activity_bookkeeping, post_operation_bookkeeping};
 use crate::state;
 use crate::state::flow_states::RegistrationFlowState;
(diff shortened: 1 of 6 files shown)


src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -38,17 +52,40 @@ pub async fn identity_registration_start() -> Result<IdRegNextStepResult, IdRegS
         return Err(IdRegStartError::InvalidCaller);
     }
 
-    let (captcha_png_base64, flow_state) = captcha_flow_state(time()).await;
+    let flow_state = match captcha_required() {
+        true => captcha_flow_state(time()).await.1,
```
lmuntaner, line 56: Does it make sense to create a `let now = time();` variable? Or there is no need?

src/internet_identity/tests/integration/v2_api/identity_register/dynamic_captcha.rs
lmuntaner, line 1: Why did you created the new `identity_register` directory? Are there any best practices on how to split this? WHy not `v2_api/dynamic_captcha_register`?

src/internet_identity/tests/integration/v2_api/identity_register/dynamic_captcha.rs
```diff
@@ -0,0 +1,68 @@
...
+    // initialize a base rate of one registration every 2 seconds
+    for _ in 0..10 {
+        create_identity_with_authn_method(&env, canister_id, &authn_method);
+        env.advance_time(Duration::from_secs(2))
+    }
+
+    // Double the rate of registrations to one per second
+    // The 20% threshold rate should allow 5 registrations before the captcha kicks in
+    let flow_principal = test_principal(0);
+    for _ in 0..5 {
```
lmuntaner, line 46: Why don't you use `create_identity_with_authn_method` here?

src/internet_identity/tests/integration/v2_api/identity_register/dynamic_captcha.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    // initialize a base rate of one registration every 2 seconds
+    for _ in 0..10 {
+        create_identity_with_authn_method(&env, canister_id, &authn_method);
+        env.advance_time(Duration::from_secs(2))
+    }
+
+    // Double the rate of registrations to one per second
+    // The 20% threshold rate should allow 5 registrations before the captcha kicks in
+    let flow_principal = test_principal(0);
```
lmuntaner, line 45: It doesn't matter that the principal for the registration is the same?

lmuntaner (approved):
A few comments, already approving.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
frederikrothenberger: Ah, yes. Thanks.

src/internet_identity/tests/integration/v2_api/identity_register/dynamic_captcha.rs
frederikrothenberger: I did it because the dynamic captcha is a subset of the whole identity registration tests. But I can also flatten the structure if you prefer.

src/internet_identity/tests/integration/v2_api/identity_register/dynamic_captcha.rs
frederikrothenberger: If all previous flows are completed, then it does not matter (because flows are cleaned up on `finish`). But you're right, it's cleaner to use different ones.

src/internet_identity/tests/integration/v2_api/identity_register/dynamic_captcha.rs
frederikrothenberger: Because the `create_identity_with_authn_method` helper works with or without captcha. And I need to assert here that the captcha is _not_ required.

PR #2639, opened 2024-10-02
Remove unnecessary counters from persistent state

frederikrothenberger:
The `StableBTreeMap` tracks the number of elements on its own, so there is no need for an external counter.

Removing the extra counters also makes the stats resetting obsolete and hence allows to remove the pre_stats II version from the integration tests.


frederikrothenberger:
> How about the future stable memory. Is removing keys frontwards compatible? Could we rollback afterwards to the memory with the counting fields?

Yes, because the fields are optional. Note that we do have explicit upgrade / rollback tests to verify that.

lmuntaner (approved):
Thanks!

lmuntaner:
> Yes, because the fields are optional. Note that we do have explicit upgrade / rollback tests to verify that.

True, I forgot about them.

PR #2640, opened 2024-10-02
Remove interface check breaking change exception

frederikrothenberger:
This PR removes the breaking change exception from the CI interface compatibility check. The exception is no longer necessary as the breaking changes have been released.


lmuntaner (approved):
Thanks!

PR #2650, opened 2024-10-15
Add new metrics to monitor memory

lmuntaner:
# Motivation

To introduce new metrics that will be used to monitor the memory usage as requested by security.

# Changes

* Introduce `ii_archive_stable_memory_bytes`, `ii_archive_heap_pages` and `ii_archive_heap_memory_bytes` metrics to the archive canister.
* Introduce `internet_identity_heap_memory_bytes` and `internet_identity_stable_memory_bytes` metrics to main II canister.


src/archive/src/main.rs
```diff
@@ -575,6 +575,7 @@ fn http_request(req: HttpRequest) -> HttpResponse {
 }
 
 fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
+    const WASM_PAGE_SIZE_IN_BYTES: f64 = 65536.0;
```
frederikrothenberger, line 578: There should be a constant for that in the `cdk`. Also, as a side-note: I would not do floating point calculations unless absolutely necessary. Floating point calculations could introduce rounding errors.

src/archive/src/main.rs
```diff
@@ -634,6 +635,23 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
         stable_size() as f64,
         "Number of stable memory pages used by this canister.",
     )?;
+    w.encode_gauge(
+        "ii_archive_stable_memory_bytes",
```
frederikrothenberger, line 639: Is this useful? Do you need both pages and bytes?

frederikrothenberger (commented):
Honestly, I would just go with `ii_archive_heap_pages` which is clearly missing. The rest can be deduced from `pages` when processing the data.

src/archive/src/main.rs
lmuntaner: Yes, there was the agreement to expose the bytes and alarms will be created for us.

See the thread: https://dfinity.slack.com/archives/C063Z1ZBL8P/p1728300749385249?thread_ts=1727341770.248009&cid=C063Z1ZBL8P

src/archive/src/main.rs
lmuntaner: Nice! Fixed!

lmuntaner:
> Honestly, I would just go with `ii_archive_heap_pages` which is clearly missing. The rest can be deduced from `pages` when processing the data.

Alarms for memory threshold will be added if we expose this metrics.

src/archive/src/main.rs
frederikrothenberger: Ah, TIL. That makes sense 👍

frederikrothenberger (approved):
LGTM, thanks!

PR #2651, opened 2024-10-15
Make identity metadata update synchronous

frederikrothenberger:
There is no need to `await` metadata if some value needs to be written. Instead we just update the internal promise of the metadata repository to update the value once it is loaded.

This also makes a change to the registration flow to make sure that when
the success page is shown that the metadata is committed to make sure
the identity is fully set up. This should finally resolve the observed
flakiness in the e2e tests.

### DIFF
--- a/src/frontend/src/flows/register/index.ts
+++ b/src/frontend/src/flows/register/index.ts
@@ -194,12 +194,16 @@ export const registerFlow = async ({
   await finalizeIdentity?.(userNumber);
   // We don't want to nudge the user with the recovery phrase warning page
   // right after they've created their anchor.
-  // The metadata starts to fetch when the connection is created.
-  // But it might not have finished yet, so we `await` for `updateIdentityMetadata` to also wait for it.
-  await result.connection.updateIdentityMetadata({
+  result.connection.updateIdentityMetadata({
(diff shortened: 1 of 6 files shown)


src/frontend/src/flows/register/index.ts
```diff
@@ -194,12 +194,16 @@ export const registerFlow = async ({
...
-  await result.connection.updateIdentityMetadata({
+  result.connection.updateIdentityMetadata({
     recoveryPageShownTimestampMillis: Date.now(),
   });
-  await setAnchorUsed(userNumber);
+  // Immediately commit (and await) the metadata, so that the identity is fully set up when the user sees the success page
+  // This way, dropping of at that point does not negatively impact UX with additional nagging
+  // To the user this just looks like the captcha takes a bit longer to verify.
+  await Promise.all([
+    result.connection.commitMetadata(),
```
lmuntaner, line 204: Otherwise, the metadata was committed when the user visited the manage page, right?

lmuntaner (approved):
Nice!

src/frontend/src/flows/register/index.ts
frederikrothenberger: Yes, but in many of our tests that page was never visited...

PR #2652, opened 2024-10-15
Add script test canisters

lmuntaner:
# Motivation

Make it easy to run the canisters tests.

Tu run all the tests successfully, the environment needs to be set up. There was a readme file explaining part of it.

# Changes

I moved the explanation to a script that can be run and it will download the necessary artifacts needed to run all the tests.


lmuntaner:

scripts/test.canisters.sh
```diff
@@ -0,0 +1,49 @@
...
+# Build II
+II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=1 ./scripts/build --internet-identity
+
+# Build Archive
+./scripts/build --archive
+
+# Run tests
+
+echo "Running integration tests."
+cargo test "${@}"
```
frederikrothenberger, line 49:
```suggestion
cargo test "${@}"

```

scripts/test.canisters.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
frederikrothenberger, line 1: This is the only script with multiple dots in the name. What about calling it `test-canisters.sh`?

scripts/test.canisters.sh
```diff
@@ -0,0 +1,49 @@
...
+
+POCKET_IC_SERVER_VERSION=6.0.0
+POCKET_IC_SERVER_PATH="pocket-ic"
```
frederikrothenberger, line 4: Given you are dealing with paths here, what about making sure that the working directory really is the repository directory?

scripts/test.canisters.sh
```diff
@@ -0,0 +1,49 @@
...
+PREVIOUS_II_WASM_PATH="internet_identity_previous.wasm.gz"
+PREVIOUS_ARCHIVE_WASM_PATH="archive_previous.wasm.gz"
+
+if [[ $OSTYPE == "linux-gnu"* ]] || [[ $RUNNER_OS == "Linux" ]]; then
```
frederikrothenberger, line 8: Did you intend this to be used by CI too?
I like that idea, but in that case it would be great if you actually used it in CI too. 😉

scripts/test.canisters.sh
frederikrothenberger: We also usually set flags to fail early on error:
```
set -euo pipefail
```

frederikrothenberger (approved):
Useful script 👍 
Left a few comments but already approving.

scripts/test.canisters.sh
lmuntaner: Yes, that's the idea but I'll do it in another PR.

scripts/test.canisters.sh
lmuntaner: Yes, good point.

scripts/test.canisters.sh
lmuntaner: Good idea, done!

scripts/test-canisters.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+set -euo pipefail
+
+POCKET_IC_SERVER_VERSION=6.0.0
+POCKET_IC_SERVER_PATH="pocket-ic"
+PREVIOUS_II_WASM_PATH="internet_identity_previous.wasm.gz"
+PREVIOUS_ARCHIVE_WASM_PATH="archive_previous.wasm.gz"
+
+# Check if the script is run from the root of the project
+project_root=$(git rev-parse --show-toplevel 2>/dev/null)
```
frederikrothenberger, line 10: We normally just switch to the appropriate directory:
```
SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
cd "$SCRIPTS_DIR/.."
```

lmuntaner: Ah, ok, changed!

PR #2655, opened 2024-10-17
Add test-canisters.sh script to CI

lmuntaner:
# Motivation

Make sure that test-canisters.sh script is kept up-to-date in case something else is needed to run tests.

# Changes

Add a test in CI to run the tests with this script.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -394,6 +394,8 @@ jobs:
     needs: [cached-build, docker-build-archive]
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
+
       - name: "Download II wasm"
         uses: actions/download-artifact@v4
         with:
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -412,6 +412,8 @@ jobs:
       - name: Run tests
         run: |
           pwd
+          ls -l
+          ls -l scripts
           ./scripts/test-canisters.sh --no-build
 
   ######################
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -417,11 +417,7 @@ jobs:
           touch dist/index.html
 
       - name: Run tests
-        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -391,6 +390,20 @@ jobs:
         env:
           RUST_BACKTRACE: 1
 
+  test-canisters-script:
```
frederikrothenberger, line 393: This job takes more than 10 minutes, so it is immediately the slowest thing in the whole pipeline. I would not add it without any caching.

For reference, the normal pipeline with caching takes ~6m 30s.

frederikrothenberger (commented):
Could you please split the changes and the reformatting into different PRs?

.github/workflows/canister-tests.yml
lmuntaner: I added the `--no-build` parameter and used the built wasms. It takes 6'30" now. Is this reasonable?

It depends on a few other jobs though. Therefore, the whole CI is probably slower now.

I could add pocket-ic and the other dependencies into a cache and use it as well. Maybe in another PR?

lmuntaner:
@frederikrothenberger ready for another review.

The duration is now down to 2min.

.github/workflows/canister-tests.yml
```diff
@@ -391,6 +390,49 @@ jobs:
...
+        run: mv internet_identity_test.wasm.gz internet_identity.wasm.gz
+
+      - name: Create dummy assets
+        run: |
+          mkdir dist
+          touch dist/index.html
+
+      - name: Run tests
+        run: |
+          ./scripts/test-canisters.sh --no-build should_keep_new_anchor_across_rollback
```
frederikrothenberger, line 428: Maybe add a comment on how you chose these particular tests?

frederikrothenberger (approved):
LGTM, thanks!

.github/workflows/canister-tests.yml
lmuntaner: done

PR #2656, opened 2024-10-17
Run e2e tests with captcha disabled

frederikrothenberger:
This PR adds another setting to the e2e tests: captcha enabled / disabled All tests are now also run against II deployed with captcha disabled.

In order to keep the number of tests run still to a somewhat reasonable size not all combinations are run anymore: i.e. there is one domain with captcha enabled and one domain with captcha disabled

Since the domain and the captcha are independent features, this should not reduce the signal gained by running e2e tests.


lmuntaner (approved):
True, this was easy.

PR #2658, opened 2024-10-17
Add loader animation on registration completion

frederikrothenberger:
There is currently a bit of an awkward wait time when finalizing the registration without a captcha. This adds a loader animation to convey that something is happening in the background.

I also remove the comment about captcha as it is not necessarily shown.


lmuntaner (approved):
thanks!

PR #2660, opened 2024-10-18
Change how dynamic captcha stores data

lmuntaner:
# Motivation

We realized that when the config changes, the stored data still takes into account the old config because it was stored by expiration date which took into account the config.

In this PR, the PR is stored with the create timestamp and pruned taking into account the config intervals.

There is a new test to check that the old config doesn't affect the new rates (given enough time to prune the old data).

### DIFF
--- a/src/internet_identity/src/storage/registration_rates.rs
+++ b/src/internet_identity/src/storage/registration_rates.rs
@@ -112,6 +112,8 @@ fn calculate_registration_rate<M: Memory>(now: u64, data: &MinHeap<Timestamp, M>
         .peek()
         // calculate the time window length with respect to the current time
         .map(|ts| now - ts)
+        // the value _could_ be 0 if the oldest timestamp was added in the same execution round
+        .filter(|val| *val != 0)
         // use the value to calculate the rate per second
         .map(|val| rate_per_second(data.len(), val))
         // if we don't have data, the rate is 0
--- a/src/internet_identity/src/storage/registration_rates.rs
+++ b/src/internet_identity/src/storage/registration_rates.rs
@@ -9,10 +9,10 @@ use std::time::Duration;
 
 pub struct RegistrationRates<M: Memory> {
     /// Min heap of registration timestamps to calculate the reference registration rate.
-    /// [prune_data] removes values that  are in the past whenever new data is added.
+    /// [prune_data] removes values that are older than now minus the reference rate interval whenever new data is added.
     reference_rate_data: MinHeap<Timestamp, M>,
     /// Min heap of registration timestamps to calculate the current registration rate.
-    /// [prune_data] removes values that  are in the past whenever new data is added.
+    /// [prune_data] removes values that are older than now minus the current rate interval whenever new data is added.
     current_rate_data: MinHeap<Timestamp, M>,
 }
 
@@ -47,16 +47,17 @@ impl<M: Memory> RegistrationRates<M> {
     }
     pub fn new_registration(&mut self) {
         self.prune_expired();
-        let Some(data_retention) = dynamic_captcha_config() else {
+        // Return if no dynamic captcha
+        if dynamic_captcha_config().is_none() {
             return;
         };
 
         let now = time();
         self.reference_rate_data
-            .push(&(now + data_retention.reference_rate_retention_ns))
+            .push(&now)
             .expect("out of memory");
         self.current_rate_data
-            .push(&(now + data_retention.current_rate_retention_ns))
+            .push(&now)
             .expect("out of memory");
     }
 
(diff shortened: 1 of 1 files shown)


lmuntaner:

src/internet_identity/src/storage/registration_rates.rs
```diff
@@ -108,20 +106,13 @@ impl<M: Memory> RegistrationRates<M> {
...
-        // get the oldest expiration timestamp
+        // get the oldest value
         .peek()
-        // calculate the registration timestamp from expiration
-        .map(|ts| ts - data_retention_ns)
         // calculate the time window length with respect to the current time
         .map(|ts| now - ts)
-        // the value _could_ be 0 if the oldest timestamp was added in the same execution round
-        // -> filter to avoid division by 0
+        // There could be negative values if they haven't been pruned yet.
```
frederikrothenberger, line 115: The values are `u64` so the cannot be negative. But also `ts` is never bigger than `now` so the minimum value is 0, no?

src/internet_identity/src/storage/registration_rates.rs
```diff
@@ -281,6 +272,71 @@ mod test {
...
+        for _ in 0..100 {
+            registration_rates.new_registration();
+            registration_rates.new_registration();
+            TIME.with_borrow_mut(|t| *t += Duration::from_secs(1).as_nanos() as u64);
+        }
+        assert_eq!(
+            registration_rates.registration_rates().unwrap(),
+            NormalizedRegistrationRates {
+                current_rate_per_second: 2.0,
+                reference_rate_per_second: 2.0, // increases too, but more slowly
```
frederikrothenberger, line 334:
```suggestion
                reference_rate_per_second: 2.0, // equal to current rate because window lenght is the same
```

src/internet_identity/src/storage/registration_rates.rs
```diff
@@ -281,6 +272,71 @@ mod test {
...
+        TIME.with_borrow_mut(|t| *t += Duration::from_secs(100).as_nanos() as u64);
+
+        // Set current rate to 2 registrations per seconds for 100 seconds
+        // which is the current and reference rate.
+        for _ in 0..100 {
+            registration_rates.new_registration();
+            registration_rates.new_registration();
+            TIME.with_borrow_mut(|t| *t += Duration::from_secs(1).as_nanos() as u64);
+        }
+        assert_eq!(
```
frederikrothenberger, line 330: Maybe change it to assert equality between `current_rate_per_second` and `reference_rate_per_second`? Because that is the important property of the test, not the actual resulting values.

src/internet_identity/src/storage/registration_rates.rs
```diff
@@ -281,6 +272,71 @@ mod test {
...
+                    current_rate_sampling_interval_s: 100,
+                    reference_rate_sampling_interval_s: 100,
+                },
+            }
+        });
+
+        // Needed to prune the data after new config
+        // It prunes 100 for each new registration
+        // we added 1100 before. Remove them with 110, plus 2 to remove the new 110 added.
+        for _ in 0..112 {
```
frederikrothenberger, line 316: I'm confused by this. It only takes 11 registrations to prune 1100 entries if every entry prunes 100 elements.

But also, the whole block is not necessary as the subsequent setup of the new rates will more than prune all of the existing ones.

src/internet_identity/src/storage/registration_rates.rs
lmuntaner: Oh, sorry, yes. I did some changes here and I didn't realize that it doesn't make sense this anymore.

src/internet_identity/src/storage/registration_rates.rs
lmuntaner: good point, done!

src/internet_identity/src/storage/registration_rates.rs
lmuntaner: Oh, yes, true. Removed.

lmuntaner:
@frederikrothenberger ready for another review

src/internet_identity/src/storage/registration_rates.rs
```diff
@@ -112,8 +112,6 @@ fn calculate_registration_rate<M: Memory>(now: u64, data: &MinHeap<Timestamp, M>
         .peek()
         // calculate the time window length with respect to the current time
         .map(|ts| now - ts)
-        // There could be negative values if they haven't been pruned yet.
-        .filter(|val| *val != 0)
```
frederikrothenberger, line 125: No,no you still need the 0 check. It's just the comment that was wrong. I.e. it should say filter out 0 rather than talk about negative values.

lmuntaner: oops, yes. Good that there was a test for it

frederikrothenberger (approved):
Thanks, LGTM! 👍

PR #2664, opened 2024-10-22
Add "derivationOrigin" to id_alias credential

lmuntaner:
# Motivation

Fix issue raised here: https://dfinity.atlassian.net/browse/FOLLOW-1651

# Changes

* Add the `derivationOrigin` field to the id_alias credential.
* Do not import `AliasTuple` from vc-sdk. It created a dependency when I wanted to change it. Therefore, I assumed it should not come from the sdk because it's just a helper type within the II implementation.

I will add `derivationOrigin` to the `AliasTuple` in the sdk as well, but I believe those two types shouldn't be reused because it creates this dependency where the SDK needs to change for II to use the change, and then the SDK needs to use the new change again. There is no need for this flow with this shared type because it's just a helper type here.

# Testing

I deployed locally and checked with the dummy relying party and issuer that the new field is added.

However, the integration tests rely on the VC SDK which still doesn't validate the new field. As soon as we add this validation, we should upgrade the dependency in II and the new field would be validated as well.

### DIFF
--- a/src/internet_identity/src/vc_mvp.rs
+++ b/src/internet_identity/src/vc_mvp.rs
@@ -60,11 +60,11 @@ pub async fn prepare_id_alias(
     };
 
     // rp_signing_input is sent to the relying party.
-    // Therefore, it needs to have the rp origin to confirm from where the principal is derived from.
+    // The relying party needs to have the derivation origin to confirm where the principal is derived from.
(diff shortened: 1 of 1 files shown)


lmuntaner:

frederikrothenberger (approved):
LGTM, thanks!

PR #2667, opened 2024-10-23
Update ii-spec.mdx

LXIF:
Remove query from 'replace_...' methods in spec as these are update methods


frederikrothenberger (approved):
Thanks!

PR #2671, opened 2024-10-25
remove first_time_subtitle from first-time known-dapp auth modal

LXIF:
Removed the subtitle from first-time known-dapp auth modal.


lmuntaner:
@LXIF we usually review when specifically asked for it. You can add me as "Reviewer" on the right bar.

Many times we create PRs that are not ready and we want to test something in CI or ask a different person. Therefore, most won't review unless ask for it directly.

lmuntaner (approved):

PR #2672, opened 2024-10-25
Remove mobile footer title and text

LXIF:
Remove mobile footer and text as per https://dfinity.atlassian.net/browse/GIX-3120 (Note: this removes it also from the landing page. If we need to handle this case-by-case, please let me know)

### DIFF
--- a/src/frontend/src/styles/main.css
+++ b/src/frontend/src/styles/main.css
@@ -1136,6 +1136,7 @@ a:hover,
     padding: 0;
     /* We don't need to account for the logo with absolute position */
     margin-top: 0;
+    display: none;
   }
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/styles/main.css
```diff
@@ -1136,6 +1136,7 @@ a:hover,
     padding: 0;
```
lmuntaner, line 1136: We can remove the previous rules as well then, right?

lmuntaner (approved):
To trigger a new review, you need to click on the "circle" icon next to my name in "Reviewers" list. And ideally also mention me in a comment so that I know that this can be reviewed again.

Thanks!

PR #2680, opened 2024-10-31
Display timestamp of last usage for each Passkey device in PAsskey device selector

LXIF:
Replaced type DeviceData with type DeviceWithUsage (this is in accordance with the data that was effectively already available, DeviceData was not really the correct type to use)

### DIFF
--- a/src/frontend/src/flows/manage/authenticatorsSection.ts
+++ b/src/frontend/src/flows/manage/authenticatorsSection.ts
@@ -47,6 +47,7 @@ export const authenticatorsSection = ({
   ];
 
   const authenticators = dedupLabels(authenticators_);
+  console.log(authenticators);
 
   return html`
     <aside class=${wrapClasses.join(" ")} data-role="passkeys">
@@ -109,7 +110,7 @@ export const authenticatorsSection = ({
 };
 
 export const authenticatorItem = ({
-  authenticator: { alias, dupCount, warn, remove, rename },
+  authenticator: { alias, last_usage, dupCount, warn, remove, rename },
   index,
   icon,
 }: {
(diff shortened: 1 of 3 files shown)


LXIF:
Screenshot 2024-10-31 at 15 11 58

It looks like this now^

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -125,6 +125,20 @@ export const authenticatorItem = ({
...
 
+  let lastUsageTimeStamp: Date | undefined;
+  let lastUsageFormattedString: string | undefined;
+
+  if (last_usage.length > 0 && typeof last_usage[0] === "bigint") {
+    lastUsageTimeStamp = new Date(Number(last_usage[0] / BigInt(1000000)));
+  }
+
+  if (lastUsageTimeStamp) {
+    lastUsageFormattedString = new Intl.DateTimeFormat(undefined, {
```
lmuntaner, line 136: How about using `Intl.RelativeTimeFormat`? So that it says "Yesterday"

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -134,6 +148,11 @@ export const authenticatorItem = ({
         ${nonNullish(dupCount) && dupCount > 0
           ? html`<i class="t-muted">&nbsp;(${dupCount})</i>`
           : undefined}
+        ${nonNullish(lastUsageFormattedString)
```
lmuntaner, line 151: I like the current text style but I would put them in two lines, what do you think?

lmuntaner (commented):
Thanks! Sorry for the delay, I didn't clean up my email this morning.

A couple of comments that we can discuss.

LXIF:
Screenshot 2024-11-01 at 14 01 12

Rearranged and using relative time format now.

LXIF:
Screenshot 2024-11-01 at 14 22 38

Added helper function and tests to display decreasing precision the longer ago the last usage was.

lmuntaner (approved):
Thanks!

PR #2681, opened 2024-11-01
Change landing page title

lmuntaner:
# Motivation

Change the copy of the landinga page as requested.

# Changes

* Change the copy of the landing page title.
* Split the copy in two to make sure it's always shown separately. It was ugly with any other split.
* The second line was sometimes overflowing a big just before the breakpoint and turning into two lines (three in total). This was ugly and instead of changing the padding to adapt to it, I used word-wrap to ensure the line is kept whole even when overflowing. It only overflows a bit, so it's not noticeable.

### DIFF
--- a/src/frontend/src/components/landingPage.json
+++ b/src/frontend/src/components/landingPage.json
@@ -1,6 +1,7 @@
 {
   "en": {
-    "title": "Identity that's private and decentralized.",
+    "title_1": "Decentralized.",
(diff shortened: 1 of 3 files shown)


lmuntaner:
New UI:

Screenshot 2024-11-01 at 13 21 12

LXIF (approved):

PR #2684, opened 2024-11-05
Change dynamic captcha registration rates calculation

lmuntaner:
# Motivation

The dynamic captcha is triggered without user registration spikes.

This is due to the current calculation of the current rate which doesn't use the current interval, and instead uses as interval the difference between now and the oldest point.

For example:
* current_rate_interval is 5 minutes.
* There are no registrations in the last 5 minutes.
* Then two new registrations are added in 2 seconds.
* The calculation right afterwards will be 2 seconds. Therefore, the rate would be 2 registrations / 2 seconds -> 1. Instead of 2 registrations / 5 minutes.

# Changes

* Change the rate calculation in `registration_rates`.

# Tests

* Change the test expectations.

### DIFF
--- a/src/internet_identity/src/storage/registration_rates.rs
+++ b/src/internet_identity/src/storage/registration_rates.rs
@@ -57,12 +57,19 @@ impl<M: Memory> RegistrationRates<M> {
         self.current_rate_data.push(&now).expect("out of memory");
     }
 
+    /// Calculates the registration rates for the current and reference intervals along with the threschold.
+    /// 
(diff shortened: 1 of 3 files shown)


sea-snake (approved):
As discussed, this change will affect the behavior when window size configuration has changed since this change doesn't take into account how to handle missing (pruned) data points.

Overall, as long as we don't change the window size configuration till we have an implementation handling the missing data points, we shouldn't run into issues.

PR #2688, opened 2024-11-07
Fix dynamic captcha when config changes to longer intervals

lmuntaner:
# Motivation

With the current registration rate calculation, when the interval is increased through a config change, the rate will be underestimated and could mean that the captcha is not enabled, when it should.

We want to fix this and to be on the safe side, enable captcha when this happens.

# Changes

* Add a new stored field to keep track of the interval of the pruned data.
* Calculate whether the data we have is sufficient or not to compare rates by checking whether the interval of the pruned data is longer than the config intervals.


lmuntaner:
@sea-snake please review

sea-snake (approved):

PR #2691, opened 2024-11-13
Speed up the onboarding process by parallelizing calls

LXIF:
Moved the recoveryWizard call and the commitMetadata call below the fetchDelegation call. This way they can actually run in parallel.

The reason they ran sequentially before is that the IdentityMetadataRepository in the AuthenticatedConnection class instantiates its metadata with the Promise that wraps the result of the identity_info call. This makes sense, as it allows it to load asynchronously when the connection gets established.

The rub comes when you try to access the data the promise resolves into before it's there - it then blocks access to anything else in the object - and the fetchDelegation call needs to use the AuthenticatedConnection object, though not the metadata inside it. [I made a codepen to demonstrate this](https://codepen.io/LXIF/pen/WNVPLNX?editors=1111).

Here's the result in terms of the sequencing of the network requests:

Before:
Screenshot 2024-11-13 at 16 41 22

After:
Screenshot 2024-11-13 at 16 40 44

### DIFF
--- a/src/frontend/src/flows/authorize/index.ts
+++ b/src/frontend/src/flows/authorize/index.ts
@@ -210,11 +210,9 @@ const authenticate = async (
     autoSelectionIdentity: autoSelectionIdentity,
   });
 
-  
   // at this point, derivationOrigin is either validated or undefined
   const derivationOrigin =
-  authContext.authRequest.derivationOrigin ?? authContext.requestOrigin;
-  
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks for the research and fix!

Could you change the title to something more user friendly.

The commit messages go to the proposal, so it's nicer if we have them "more or less" understandable for people not familiar with the code. Something like "Speed up onboarding process by parallelizing calls".

PR #2700, opened 2024-11-21
change path in frontend checks to new icons location

LXIF:
Change frontend check to account for the new icons location introduced by cacheable icons

### DIFF
--- a/.github/workflows/frontend-checks.yml
+++ b/.github/workflows/frontend-checks.yml
@@ -29,7 +29,7 @@ jobs:
         run: |
           while read -r logo
           do
-            if [ ! -f src/frontend/assets/icons/"$logo" ]
+            if [ ! -f src/frontend/assets/src/assets/icons/"$logo" ]
             then
               echo "Logo not found: $logo"
               exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/frontend-checks.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: |
           while read -r logo
           do
-            if [ ! -f src/frontend/assets/icons/"$logo" ]
+            if [ ! -f src/frontend/assets/src/assets/icons/"$logo" ]
```
lmuntaner, line 32: Is this correct? I thought it was `src/frontend/src/assets/icons`

LXIF:
> Sorry, I don't think we can do this separately. We need to do it in the other PR, right?

This one fails because the scripts are changed but the file locations aren't. I could change the file locations in this one already, but I have to assume in that case other tests would fail without updating the imports etc.

lmuntaner:
> > Sorry, I don't think we can do this separately. We need to do it in the other PR, right?
> 
> This one fails because the scripts are changed but the file locations aren't. I could change the file locations in this one already, but I have to assume in that case other tests would fail without updating the imports etc.

No, it's ok. Let's close this one and do it in the other PR.

PR #2701, opened 2024-11-21
change paths in update-dapps.yml and scripts/update-dapps

LXIF:
change paths in update-dapps.yml and scripts/update-dapps to new icon location that allows dynamic loading and caching

### DIFF
--- a/.github/workflows/update-dapps.yml
+++ b/.github/workflows/update-dapps.yml
@@ -4,7 +4,7 @@ name: Dapps Update
 on:
   schedule:
     # Check for updates daily
-    - cron:  '0 0 * * *'
+    - cron: "0 0 * * *"
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-dapps.yml
```diff
@@ -32,13 +32,14 @@ jobs:
           base: main
           add-paths: |
             src/frontend/src/flows/dappsExplorer/dapps.json
-            src/frontend/assets/icons
+            src/frontend/assets/src/assets/icons
```
lmuntaner, line 35: Ditto, is this correct? I was expecting it to be `frontend/src/assets/icons`

Could you try running this workflow and see that it works as expected?

LXIF: @lmuntaner happy to do that, need to check how to do that though

lmuntaner: I believe that it's also better to introduce this change along with the PR that introduces caching.

LXIF: Yeah makes sense at the end of the day

lmuntaner:
Closing because it's added in #2697

PR #2702, opened 2024-11-21
Show support links on loading and error screens.

sea-snake:
Right now both loading and error screens don't link users to support, this PR adds support links on both these screens.
- The loading overlay shows a "Ongoing issues" link if it is shown longer than 10s.
- The error screen has a secondary button "Go to support" below the "Try again" button.

Both links go to the [same support page](https://identitysupport.dfinity.org/hc/en-us/articles/32301362727188), this page will be updated to document common errors and ongoing issues over time.

### DIFF
--- a/src/frontend/src/components/displayError.ts
+++ b/src/frontend/src/components/displayError.ts
@@ -1,3 +1,4 @@
+import { ERROR_SUPPORT_URL } from "$src/config";
 import { TemplateElement } from "$src/utils/lit-html";
 import { nonNullish } from "@dfinity/utils";
 import { html, render } from "lit-html";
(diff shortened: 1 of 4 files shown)


LXIF (approved):

lmuntaner (approved):
Thanks!

Could you provide a screenshot of how the new loader screen looks like?

sea-snake:
> Could you provide a screenshot of how the new loader screen looks like?

The link would only appear below the astronaut in the case, the user is stuck for a while (after 10 sec).

image

lmuntaner:
> > The link would only appear below the astronaut in the case, the user is stuck for a while (after 10 sec).
> 
> Thanks! Does that feel like a button? How about changing the color on hover?

More than the style, I think it's the text that doesn't make it look like a button. It needs a verb.

How about "Go to ongoing issues"

PR #2710, opened 2024-11-27
Fix loader in showcase and add it to screenshots

sea-snake:
- Fix loader in showcase, make sure the astronaut image loads and doesn't animate.
- Add the loader screen to screenshots, including the "taking forever" variant that shows after 10 seconds.

### DIFF
--- a/src/frontend/src/components/loader.ts
+++ b/src/frontend/src/components/loader.ts
@@ -7,7 +7,7 @@ const loaderUrl = import.meta.glob("./loader.svg", {
   eager: true,
   query: "?url",
   import: "default",
-})["./loader.svg"];
+})["./loader.svg"] as string;
 
 // Duration in milliseconds a user considers as taking forever
 const TAKING_FOREVER = 10000;
--- a/src/frontend/src/components/loader.ts
+++ b/src/frontend/src/components/loader.ts
@@ -1,13 +1,20 @@
 import { ERROR_SUPPORT_URL } from "$src/config";
 import { html, render } from "lit-html";
-import loaderUrl from "./loader.svg";
+
+// Use same import approach as in 'src/frontend/src/flows/dappsExplorer/dapps.ts'
+// this makes the import the same format (url string) in both the build and showcase.
+const loaderUrl = import.meta.glob("./loader.svg", {
+  eager: true,
+  query: "?url",
+  import: "default",
+})["./loader.svg"];
 
 // Duration in milliseconds a user considers as taking forever
 const TAKING_FOREVER = 10000;
 
 const loader = (takingForever = false) =>
   html` <div id="loader" class="c-loader">
-    <img class="c-loader__image" src=${loaderUrl} alt="loading" />
+    <img class="c-loader__image" src="${loaderUrl}" alt="loading" />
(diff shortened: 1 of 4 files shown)


src/frontend/src/components/loader.ts
```diff
@@ -1,13 +1,20 @@
 import { ERROR_SUPPORT_URL } from "$src/config";
 import { html, render } from "lit-html";
-import loaderUrl from "./loader.svg";
+
+// Use same import approach as in 'src/frontend/src/flows/dappsExplorer/dapps.ts'
```
lmuntaner, line 4: Ah, nice, maybe we can do the same in the other and it fixes the issue with the types?

src/frontend/src/styles/main.css
```diff
@@ -3149,6 +3149,7 @@ input[type="checkbox"] {
   animation-duration: 5s;
   animation-iteration-count: infinite;
   animation-timing-function: ease-in-out;
+  transform: translate(-50%, -50%);
```
lmuntaner, line 3152: Why is this necessary?

src/frontend/src/components/loader.ts
```diff
@@ -33,8 +40,11 @@ const startLoader = () => {
   };
 };
 
-export const withLoader = async <A>(action: () => Promise<A>): Promise<A> => {
-  const endLoader = startLoader();
+export const withLoader = async <A>(
+  action: () => Promise<A>,
+  showCheckOngoingIssues?: boolean
```
lmuntaner, line 45: Where do you pass this? I don't see it used in any showcase page nor anywhere else...

lmuntaner (commented):
Thanks! A few questions but it's looking very good!

src/frontend/src/styles/main.css
sea-snake: When I disable the animation in the showcase (for screenshots), the logo isn't centered anymore because `translate(-50%, -50%)` was part of the animation. With this change the logo is always centered even if the animation is disabled..

src/frontend/src/components/loader.ts
sea-snake: It's used in the `loaderTakingForever.astro` file, to avoid having to wait the 10 seconds, there's two loading screen screenshots, one with button and one without.

src/frontend/src/styles/main.css
lmuntaner: noted, thanks!

src/frontend/src/components/loader.ts
sea-snake: This doesn't fix the issue with types, it fixes the issue with the actual imported data being inconsistent between showcase and build.

src/frontend/src/components/loader.ts
lmuntaner: Oh, sorry. The `true` was on the other line in my screen and I missed it

lmuntaner (approved):

src/frontend/src/components/loader.ts
lmuntaner: thanks for the explanation!

PR #2687, opened 2024-11-06
Remove screen to select authentication method: passkey or pin

lmuntaner:
# Motivation

Simplify onboarding new users.

# Changes

Skip the page that allows users to select the authentication method, passkey or pin.

# Tests

* Skip tests for pin functionality.
* Remove the code that awaits for that screen and clicks create with passkey.

I created two tickets to clean up the code and enable the tests for the login with PIN. We need to think how we want to do that. But I'd rather wait for the community feedback to this change before we remove this path.

<hr/><details><summary>🟡 Some screens were changed</summary></details>

PR #2711, opened 2024-11-27
Enable PIN registration with query param

lmuntaner:
# Motivation

In the PR where we deprecated registering with PIN we disabled the e2e tests to login (and register) with PIN.

However, current PIN users can still login with with the PIN and will continue to do so for a while.

Therefore, to keep the e2e coverage we had before, we want to reintroduce the e2e tests that login with the PIN. For that, we also need to reintroduce registering with PIN. We use a query param `enablePin` to enable registering with a PIN.

# Changes

* Revert most #2687 changes 
* If PIN is disabled, do not show page to choose and create passkey identity directly.

# Tests

* Change the URL used for PIN tests to add the needed query param to register with PIN.

### DIFF
--- a/src/frontend/src/flows/register/index.ts
+++ b/src/frontend/src/flows/register/index.ts
@@ -98,7 +98,6 @@ export const registerFlow = async ({
   | RateLimitExceeded
   | "canceled"
 > => {
-  console.log("in da registerFlow");
   if (!registrationAllowed) {
     const result = await registerDisabled();
     result satisfies { tag: "canceled" };
--- a/src/frontend/src/flows/register/index.ts
+++ b/src/frontend/src/flows/register/index.ts
@@ -1,5 +1,6 @@
 import { AuthnMethodData } from "$generated/internet_identity_types";
 import { withLoader } from "$src/components/loader";
+import { ENABLE_PIN_QUERY_PARAM_KEY } from "$src/config";
 import {
   PinIdentityMaterial,
   constructPinIdentity,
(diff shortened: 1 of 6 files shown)


src/frontend/src/flows/register/index.ts
```diff
@@ -98,24 +109,82 @@ export const registerFlow = async ({
...
+          pubKey: identity.getPublicKey().toDer(),
+        }),
+        finalizeIdentity: (userNumber: bigint) =>
+          storePinIdentity({ userNumber, pinIdentityMaterial }),
+        finishSlot: tempKeyWarningBox({ i18n: new I18n() }),
+        authnMethod: "pin" as const,
+      };
+    } else {
+      const identity = savePasskeyResult;
+      // TODO: Return something meaningful if getting the passkey identity fails
```
lmuntaner, line 149: This is the only new line that was not just reverted. Because the view can return `undefined`.

This is in the current code as well. The way this error was handled before was to never resolve the promise and show a toast error.

src/frontend/src/flows/register/passkey.ts
```diff
@@ -97,11 +97,32 @@ const savePasskeyTemplate = ({
 export const savePasskeyPage = renderPage(savePasskeyTemplate);
 
 // Prompt the user to create a WebAuthn identity or a PIN identity (if allowed)
-export const savePasskeyOrPin = async (): Promise<
-  IIWebAuthnIdentity | undefined
-> => {
+export const savePasskeyOrPin = async ({
```
lmuntaner, line 100: This was reverted to the old logic if `pinAllowed` true and skip to render the page if not enabled.

src/frontend/src/test-e2e/pinAuth.test.ts
```diff
@@ -123,32 +117,11 @@ test.skip("Should not prompt for PIN after deleting temp key", async () => {
   }, APPLE_USER_AGENT);
 }, 300_000);
 
-test.skip("Log into client application using PIN registration flow", async () => {
```
lmuntaner, line 126: This won't be enabled in the foreseeable future.

src/frontend/src/flows/register/index.ts
```diff
@@ -87,6 +97,7 @@ export const registerFlow = async ({
   | RateLimitExceeded
   | "canceled"
 > => {
+  console.log("in da registerFlow");
```
sea-snake: This probably should be removed.

sea-snake (approved):

PR #2714, opened 2024-11-28
Clean up and enable tests after PIN teste enabled

lmuntaner:
# Motivation

We commented some tests and code that needed to cleaned up once we confirmed that we are not reintroducing pin registration.

# Changes

* Remove commented code.
* Enable or remove some e2e tests.

# Tests

Changes only in test files.


lmuntaner:
@sea-snake please review

This is the continuation PR of #2711

sea-snake (approved):

PR #2719, opened 2024-12-02
Improve Verifiable Credential docs

lmuntaner:
# Motivation

It was not explained anywhere that the canister signature of the VC needed a specific domain.

In this PR, I add a note in the `prepare_credential` step of the VC docs to explain it.

This markdown document was not created with lint before, and that's why there are so many changes. I added a comment for the actual change. I can also create two different PRs if preferred.

### DIFF
--- a/docs/vc-spec.md
+++ b/docs/vc-spec.md
@@ -169,7 +169,7 @@ the context contains a time-stamped yet unsigned VC, for which the canister sign
 available only at `get_credential`-call.
 
 **NOTE:**
-The signed credential must use the verifialbe credential domain separator `"iccs_verifiable_credential"` as an array of bytes.
+The signed credential must use the verifiable credential domain separator `"iccs_verifiable_credential"` as an array of bytes.
 
 ### 4: Get Credential
 
--- a/docs/vc-spec.md
+++ b/docs/vc-spec.md
@@ -1,6 +1,5 @@
 # II Verifiable Credential Spec (MVP)
 
-
 ## Issuer API
 
 This section describes the (Candid) interface to be implemented by an issuer of verifiable credentials on the IC.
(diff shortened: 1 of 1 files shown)


docs/vc-spec.md
```diff
@@ -156,19 +155,22 @@ type ArgumentValue = variant { "Int" : int32; String : text };
...
 The value of `prepared_context` is basically used to transfer information between `prepare_credential`
 and `get_credential` steps, and it is totally up to the issuer to decide on the content of
-that field.  That is, the issuer creates `prepared_context`, and is the only entity that
-consumes it.  For example, when using [canister signatures](https://internetcomputer.org/docs/current/references/ic-interface-spec#canister-signatures)
+that field. That is, the issuer creates `prepared_context`, and is the only entity that
+consumes it. For example, when using [canister signatures](https://internetcomputer.org/docs/current/references/ic-interface-spec#canister-signatures)
 the context contains a time-stamped yet unsigned VC, for which the canister signature will be
 available only at `get_credential`-call.
 
+**NOTE:**
```
lmuntaner, line 171: This is the only actual change in this document. The rest are formatting changes.

lmuntaner:

LXIF (approved):

docs/vc-spec.md
```diff
@@ -156,19 +155,22 @@ type ArgumentValue = variant { "Int" : int32; String : text };
...
+The signed credential must use the verifiable credential domain separator `"iccs_verifiable_credential"` as an array of bytes.
```
przydatek, line 172: In principle, the domain separator does not have to be used, but rather this is a convention that is used by `ic-verifiable-credentials` (via the constant `VC_SIGNING_INPUT_DOMAIN`). I guess we cans till write that the domain separator **must be used** (to be compatible with the convention), but maybe add that the domain separation is achieved by adding a prefix that consists of a 1-byte length of the domain separator, followed by the actual separator (with a reference to [vc_signing_input_hash](https://docs.rs/ic-verifiable-credentials/1.0.1/ic_verifiable_credentials/fn.vc_signing_input_hash.html)

przydatek (commented):
Sorry for the late feedback, I missed the PR yesterday.

docs/vc-spec.md
lmuntaner: How about the following: https://github.com/dfinity/internet-identity/pull/2721/files

PR #2720, opened 2024-12-03
Add well-known asset for Related Origin Requests

lmuntaner:
# Motivation

Support passkeys from multiple domains.

In this PR, we implement the first step which is an asset: `.well-known/webauthn` as requested by the standard.

# Changes

* Add new path `.well-known/webauth`. The value depends on the value in the config.

# Tests

* Tested locally that the file is present.
* New tests to http to check the new asset is present in different scenarios.

### DIFF
--- a/src/internet_identity/src/assets.rs
+++ b/src/internet_identity/src/assets.rs
@@ -6,14 +6,15 @@ use crate::state;
 use asset_util::{collect_assets, Asset, CertifiedAssets, ContentEncoding, ContentType};
 use base64::engine::general_purpose::STANDARD as BASE64;
 use base64::Engine;
-use ic_cdk::api;
+use ic_cdk::{api, println};
 use include_dir::{include_dir, Dir};
+use serde_json::json;
 use sha2::Digest;
 
 // used both in init and post_upgrade
-pub fn init_assets() {
+pub fn init_assets(maybe_related_origins: Option<Vec<String>>) {
     state::assets_mut(|certified_assets| {
-        let assets = get_static_assets();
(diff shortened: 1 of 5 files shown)


lmuntaner:

src/internet_identity/src/assets.rs
```diff
@@ -58,6 +58,16 @@ pub fn get_static_assets() -> Vec<Asset> {
         encoding: ContentEncoding::Identity,
         content_type: ContentType::OCTETSTREAM,
     });
+
+    // Required to share passkeys with the different domains.
+    // See https://web.dev/articles/webauthn-related-origin-requests#step_2_set_up_your_well-knownwebauthn_json_file_in_site-1
+    // Maximum of 5 labels. If we want to support more, we'll need to set them in the configuration.
+    assets.push(Asset {
+        url_path: "/.well-known/webauthn".to_string(),
+        content: b"{\"origins\":[\"https://beta.identity.ic0.app\",\"https://beta.identity.internetcomputer.org\",\"https://identity.internetcomputer.org\",\"https://identity.ic0.app\",\"https://identity.icp0.io\"]}".to_vec(),
```
sea-snake, line 67: Since the beta version of II has a different salt from production and thus results in different identities, shouldn't passkeys also be isolated between beta and mainnet version of II?

lmuntaner: Yes, they are. But this asset is also deployed to the beta version. Therefore, it will allow to test the compatibility in between the two beta versions. Which is why I'm adding it here.

sea-snake (approved):
LGTM!

PR #2725, opened 2024-12-04
Add Related Origins Requests to Config

lmuntaner:
# Motivation

Support passkeys from multiple domains.

In this PR, I add a new field in the config that will be used in further PRs to set the needed `.well-known/webauthn` asset.

# Changes

* Add `related_origins` to the multiple Config and State related types.

# Tests

* I tested the whole flow with the branch #2720 to make sure it worked.
* I also added the field in the necessary tests

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -346,6 +346,7 @@ fn config() -> InternetIdentityInit {
         canister_creation_cycles_cost: Some(persistent_state.canister_creation_cycles_cost),
         register_rate_limit: Some(persistent_state.registration_rate_limit.clone()),
         captcha_config: Some(persistent_state.captcha_config.clone()),
+        related_origins: persistent_state.related_origins.clone(),
     })
 }
 
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -416,6 +416,11 @@ fn apply_install_arg(maybe_arg: Option<InternetIdentityInit>) {
                 persistent_state.captcha_config = captcha_config;
             })
         }
+        if let Some(related_origins) = arg.related_origins {
+            state::persistent_state_mut(|persistent_state| {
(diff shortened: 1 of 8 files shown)


lmuntaner:
@sea-snake please review

src/internet_identity/src/main.rs
```diff
@@ -346,6 +346,7 @@ fn config() -> InternetIdentityInit {
         canister_creation_cycles_cost: Some(persistent_state.canister_creation_cycles_cost),
         register_rate_limit: Some(persistent_state.registration_rate_limit.clone()),
         captcha_config: Some(persistent_state.captcha_config.clone()),
+        related_origins: persistent_state.related_origins.clone(),
```
sea-snake, line 349: Any reason why all the above are wrapped with `Some` and this one isn't? Apparently the ones above are also optional similar to this one.

src/internet_identity/src/main.rs
```diff
@@ -415,6 +416,11 @@ fn apply_install_arg(maybe_arg: Option<InternetIdentityInit>) {
                 persistent_state.captcha_config = captcha_config;
             })
         }
+        if let Some(related_origins) = arg.related_origins {
+            state::persistent_state_mut(|persistent_state| {
+                persistent_state.related_origins = Some(related_origins);
```
sea-snake, line 421: Same as above comment, though here it's the other way around.

sea-snake: Edit: nevermind, the persistent state and init aren't equal types.

src/internet_identity/src/main.rs
sea-snake: Edit: nevermind, the persistent state and init aren't equal types.

lmuntaner: Yes, the rest of the configs are not optional in the state. There is a default captcha for example. I could use a default empty array instead as well, but I thought it was meaningful to be an option of list of origins instead.

PR #2730, opened 2024-12-05
Runtime frontend feature flags.

sea-snake:
Runtime frontend feature flags, uses local storage for sync read/write of flags.

### DIFF
--- a/src/frontend/src/featureFlags/index.ts
+++ b/src/frontend/src/featureFlags/index.ts
@@ -54,7 +54,9 @@ const initializedFeatureFlags = Object.fromEntries(
 );
 
 // Make feature flags configurable from browser console
-Object.assign(window, "__featureFlags", initializedFeatureFlags);
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+window.__featureFlags = initializedFeatureFlags;
 
 // Export initialized feature flags as named exports
 export const { DOMAIN_COMPATIBILITY } = initializedFeatureFlags;
--- a/src/frontend/src/featureFlags/index.ts
+++ b/src/frontend/src/featureFlags/index.ts
@@ -20,10 +20,14 @@ export class FeatureFlag {
     this.#key = key;
     this.#defaultValue = defaultValue;
     const storedValue = this.#storage.getItem(this.#key);
-    this.#value =
-      storedValue === null
-        ? this.#defaultValue
-        : Boolean(JSON.parse(storedValue));
+    try {
+      this.#value =
+        storedValue === null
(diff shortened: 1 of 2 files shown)


sea-snake:
@lmuntaner Please review

src/frontend/src/featureFlags/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    defaultValue: boolean
+  ) {
+    this.#storage = storage;
+    this.#key = key;
+    this.#defaultValue = defaultValue;
+    const storedValue = this.#storage.getItem(this.#key);
+    this.#value =
+      storedValue === null
+        ? this.#defaultValue
+        : Boolean(JSON.parse(storedValue));
```
lmuntaner, line 26: Do we want to add a try/catch here in case there is a problem parsing the stored value? And revert to the default value instead?

src/frontend/src/featureFlags/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  constructor(
+    storage: Pick<Storage, "getItem" | "setItem" | "removeItem">,
+    key: string,
+    defaultValue: boolean
+  ) {
+    this.#storage = storage;
+    this.#key = key;
+    this.#defaultValue = defaultValue;
+    const storedValue = this.#storage.getItem(this.#key);
+    this.#value =
```
lmuntaner, line 23: When we enable a flag by changing the default value, the users that disabled it, won't see it.

Is this what we want?

Shouldn't setting the value to `true` as default set it to `true` in all users?

lmuntaner (commented):
Code looks good!

I have a question about the functionality rather than the code.

src/frontend/src/featureFlags/index.ts
sea-snake: Good catch, will add it.

src/frontend/src/featureFlags/index.ts
sea-snake: I think for (testing) purposes (we or users) might still want to disable a feature that's enabled by default at some point. By having the default value always enforced when it's true, you can't disable it anymore.

If we want all users to have a feature enabled without a possibility to disable it, the flag should be removed instead.

lmuntaner: We consider this as a testing only feature then, not to enable / disable features for a/b testing or to do partial roll outs of features.

Makes sense to me.

lmuntaner (approved):
LGTM!

@sea-snake FYI, when you want the review again, you need to ask for it explicitly:

Screenshot 2024-12-06 at 08 01 34

PR #2734, opened 2024-12-09
Add origin field to WebAuthn

lmuntaner:
# Motivation

This is part of the Domains Compatibility.

To find the right RP ID, I need the origin in the devices. The origin is already available in another endpoint in the legacy API. But I'd rather add it to the v2 API `identity_authn_info` and start using it.

I assume this is not a security issue because the endpoint [lookup](https://github.com/dfinity/internet-identity/blob/0a0ec83077d404fd759b1927015b330d6594bec7/src/internet_identity/internet_identity.did#L606) already returns the `origin` and much more.

# Changes

* Add `origin` field to `WebAuthn` type.
* Use it in the `identity_authn_info` endpoint.

# Tests

* Add new field in test mockups and expect it from the response.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -335,6 +335,7 @@ type MetadataMapV2 = vec record {
 type WebAuthn = record {
     credential_id: CredentialId;
     pubkey: PublicKey;
+    origin: opt text;
 };
 
 // Authentication method using generic signatures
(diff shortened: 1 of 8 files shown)


lmuntaner:
Closing in favor of #2737 

Adding the origin in the V2 API means more changes than expected. And we might anyway need to change the API with the new passkey flow, I'd rather not do more changes now.

PR #2735, opened 2024-12-09
Fix log in with PIN in Internet Identity page

lmuntaner:
# Motivation

Bug: User can't log in `https://identity.ic0.app/` if they registered with a PIN. It doesn't happen when logging in from a dapp, only directly in II.

# Changes

The problem was that the disabled flag was passed to the helper used to login. To make the fix obvious, I refactored the code and introduced two separate flags: `allowPinLogin` and `allowPinRegistration`.

I didn't change the API that clients can use `allowPinAuthentication`.

* Either split `allowPinAuthentication` into `allowPinLogin` and `allowPinRegistration`.
* Or rename to `allowPinLogin` or `allowPinRegistration` depending on the meaning in that function.

# Tests

The e2e test didn't catch the bug because they used the query param flag when logging in as well.

* Use the `enablePin` flag only when registering, not when logging in.

### DIFF
--- a/src/frontend/src/components/authenticateBox/index.ts
+++ b/src/frontend/src/components/authenticateBox/index.ts
@@ -71,13 +71,15 @@ export const authenticateBox = async ({
   connection,
   i18n,
   templates,
-  allowPinAuthentication,
+  allowPinLogin,
+  allowPinRegistration,
   autoSelectionIdentity,
 }: {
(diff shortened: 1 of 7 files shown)


lmuntaner:

roelstorms (approved):
LGTM, thanks for the quick fix.

PR #2737, opened 2024-12-10
Use lookupAll instead of lookupCredentials in recovery flow

lmuntaner:
# Motivation

This is part of the Domains Compatibility.

I'd like to use the `origin` of the device when logging in. However, the endpoint used by `lookupCredentials` doesn't offer that field.

Therefore, I change where this is used and in another PR I will change `fromWebauthnCredentials` to accept also the origin as part of the webauthn method and then I'll be able to add the RP ID dynamically when logging in.

# Changes

* Use `lookupAll` instead of `lookupCredentials` in `attemptRecovery` function.

# Tests

* No new functionality, no new tests.
* I tested the flow locally and it worked.


lmuntaner:

PR #2742, opened 2024-12-11
Share CredentialData param type of fromCredentials

lmuntaner:
# Motivation

To support domain compatibility, we need to use the helper to find the RP ID. 

The idea is to use the helper `findWebAuthnRpId` within `fromCredentials` from `MultiWebAuthnIdentity`.

Therefore, I reused the type expected there in all the chain that uses it.

At the moment, we could have used `DeviceData`, but API v2 doesn't use this type. Therefore, I didn't want to rely on this and reuse the custom type that holds the necessary fields for the logic.

# Changes

* Move `CredentialData` to its own module and add a converter to be used.
* Change parameter type to `CredentialData` in `findWebAuthnRpId` and `fromWebauthnCredentials`.

# Tests

* Change the tests for `findWebAuthnRpId` to use the new parameter type.


lmuntaner:

sea-snake (approved):

LXIF (approved):
LGTM

PR #2744, opened 2024-12-11
Change findWebAuthnRpId signature and return

lmuntaner:
# Motivation

Support multiple domains for passkeys.

Improve the helper `findWebAuthnRpId`:

# Changes

Changes to `findWebAuthnRpId`:
* Return the hostname which is what is needed as RP ID, not the full origin.
* Instead of expecting a list of domains with two parts, it expects full origins. It's more maintenable and we can differentiate between `beta.identity.ic0.app` and `https://fgte5-ciaaa-aaaad-aaatq-cai.ic0.app`.
* Third parameter relatedDomains is mandatory.
* Returns `undefined` when no related domains. This will be the case for local development and II deployed individually to mainnet.
* Complementary helper `relatedDomains`, which returns the set of domains related to the current one.

# Tests

* Change tests to new return type and parameters.

### DIFF
--- a/src/frontend/src/utils/findWebAuthnRpId.test.ts
+++ b/src/frontend/src/utils/findWebAuthnRpId.test.ts
@@ -1,15 +1,15 @@
 import { CredentialData } from "./credential-devices";
-import { findWebAuthnRpId } from "./findWebAuthnRpId";
+import {
+  BETA_DOMAINS,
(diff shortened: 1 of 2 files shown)


lmuntaner:

LXIF (approved):
LGTM, tests also check out locally

PR #2751, opened 2024-12-17
Refactor finding rpId for login

lmuntaner:
# Motivation

There are scenarios where the RP ID selected is not the correct one for the device the user is in. When that happens, we want to tell the user to try again, but this time, we will select another RP ID.

To be able to select the right RP ID the second time, we need to remember which RP ID was selected in the first place (or second place). We can't store that in the `MultiWebAuthnIdentity` instance, because there is a new instance every time the user tries to log in.

Instead, I want to store this info in the `Connection` instance, which is the same across tries.

In this PR, I move the logic of selecting the RP ID to the `Connection.fromWebauthnCredentials`.

# Changes

* Move logic of selecting RP ID to `Connection.fromWebauthnCredentials`.
* Add `rpId` parameter to `MultiWebAuthnIdentity` constructors.

# Tests

* Add tests for the `login` method of `Connection` class to check that rpID is passed as expected.

### DIFF
--- a/src/frontend/src/utils/iiConnection.test.ts
+++ b/src/frontend/src/utils/iiConnection.test.ts
@@ -1,12 +1,29 @@
-import { MetadataMapV2, _SERVICE } from "$generated/internet_identity_types";
+import {
+  DeviceData,
+  MetadataMapV2,
+  _SERVICE,
+} from "$generated/internet_identity_types";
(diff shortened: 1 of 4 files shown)


lmuntaner:

LXIF (approved):
LGTM :)

PR #2754, opened 2024-12-18
Add nice UX when passkey is not found in that RP ID

lmuntaner:
# Motivation

The algorithm to choose RP ID is not bulletproof. For those scenarios that the user needs to retry, we want to guide the user a bit.

In this PR, the UX is improved by showing an info toast encouraging the user to try again.

# Changes

* Add new return type `PossiblyWrongRPID` to the `login` and `fromWebauthnCredentials` methods.
* Manage the new return type and show a toast info if so.

# Tests

* Tested in beta domains. See screen video.
* Change current test for the scenario that has been changed.

### DIFF
--- a/src/frontend/src/components/authenticateBox/index.ts
+++ b/src/frontend/src/components/authenticateBox/index.ts
@@ -33,6 +33,7 @@ import {
   InvalidCaller,
   LoginSuccess,
   NoRegistrationFlow,
+  PossiblyWrongRPID,
   RateLimitExceeded,
   RegisterNoSpace,
   UnexpectedCall,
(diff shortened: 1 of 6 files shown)


lmuntaner:
UX:

https://github.com/user-attachments/assets/498c359b-6faa-41d5-bc6b-071b8eb369d4

New toast:

Screenshot 2024-12-18 at 11 57 57

lmuntaner:

LXIF (approved):
LGTM technically, though admittedly wording is confusing - users may not know we have multiple domains and that passkeys are separated by domain.

lmuntaner:
> LGTM technically, though admittedly wording is confusing - users may not know we have multiple domains and that passkeys are separated by domain.

I agree. Thankfully, this is an edge case that I hope not many people will encounter.

I'm also open to other wordings, but this was the best I could think of. We can always change before enabling the flag in mainnet.

PR #2755, opened 2024-12-18
Lm fix convert rp id to url

lmuntaner:
# Motivation

There was a bug in the code for dynamic RP IDs becuase the cancelled RP IDs were only hostnames, not origins. Therefore, they couldn't be converted to `URL` to be compared with the origins of the devices.

# Changes

* Append `https://` in front of the RP ID when present.

# Tests

* Change tests to have as input hostnames only without the schema.


lmuntaner:

LXIF (approved):
LGTM

PR #2757, opened 2024-12-19
UI Page to replace current device's origin

lmuntaner:
# Motivation

As part of the migration plan for Domains Compatibility, we want to offer the user the possibility to change the origin of a device.

This is needed to avoid having a bad UX when the user logs in and the initial RP ID selected by the algorithm is incorrect and the user needs to try again. For example, the user has a device registered in the current domain, but is currently on another device.

# Changes

* New page `replaceCurrentDeviceOrigin` with its templates and pages.

# Tests

* Added a showcase page to check how it looks. See screenshot attached.

### DIFF
--- a/src/frontend/src/flows/addDevice/replaceCurrentDeviceOrigin.json
+++ b/src/frontend/src/flows/addDevice/replaceCurrentDeviceOrigin.json
@@ -1,0 +1,9 @@
+{
+  "en": {
+    "title": "Change Current Device's Origin",
+    "paragraph": "Your current device is stored with a different origin. To avoid a bad user experience, you need to change the current device's origin with the new one.",
+    "skip": "Skip",
+    "replace_origin": "Change Origin",
+    "label_icon": "Recommended Action"
+  }
+}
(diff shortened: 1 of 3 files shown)


lmuntaner:
New Page:

Screenshot 2024-12-19 at 10 53 23

LXIF (approved):
LGTM

sea-snake (changes requested):

src/frontend/src/flows/addDevice/addCurrentDevice.json
```diff
@@ -0,0 +1,9 @@
+{
+  "en": {
+    "title": "Add Current Device",
+    "paragraph": "Please add the current device to your identity to improve the user experience with Internet Identity.",
+    "skip": "Skip",
+    "replace_origin": "Add",
```
sea-snake, line 6: One more

sea-snake, line 6:
```suggestion
    "add": "Add",
```

src/frontend/src/flows/addDevice/addCurrentDevice.ts
```diff
@@ -0,0 +1,46 @@
...
+  skip: () => void;
+  i18n: I18n;
+}): TemplateResult => {
+  const copy = i18n.i18n(copyJson);
+
+  return infoScreenTemplate({
+    cancel: skip,
+    cancelText: copy.skip,
+    next: add,
+    nextText: copy.replace_origin,
```
sea-snake, line 22:
```suggestion
    nextText: copy.add,
```

sea-snake (approved):

PR #2760, opened 2024-12-27
Fix authentication with (older) identities that have devices without a (valid) credential id.

sea-snake:
Fix authentication when (older) identities have devices without a (valid) credential id.

# Changes

- Reimplement old behavior from previous release where devices without a (valid) credential id are filtered out.
- Improve guarantees during build time by moving credential id checks into `convertToCredentialData` method itself.

# Added Tests

- Check if devices are ignored with missing credential id
- Check if devices are ignored with invalid credential id

### DIFF
--- a/src/frontend/src/utils/credential-devices.ts
+++ b/src/frontend/src/utils/credential-devices.ts
@@ -13,8 +13,22 @@ const derFromPubkey = (pubkey: DeviceKey): DerEncodedPublicKey =>
 
 export const convertToCredentialData = (
   device: Omit<DeviceData, "alias">
-): CredentialData => ({
-  credentialId: Buffer.from(device.credential_id[0] ?? []),
-  pubkey: derFromPubkey(device.pubkey),
-  origin: device.origin[0],
-});
+): CredentialData | undefined => {
+  // In certain cases, e.g. Chrome on Windows 10, an invalid credential id is
+  // not ignored but instead will result in a WebAuthn error that prevents a
+  // user from authenticating with any of their registered devices.
+  //
+  // Instead of throwing an error, we return `undefined` to make sure that the
+  // device will be filtered out to unblock the user from authenticating.
+  if (
+    device.credential_id.length !== 1 ||
+    device.credential_id[0].length === 0
+  ) {
+    return;
+  }
+  return {
+    credentialId: Buffer.from(device.credential_id[0]),
+    pubkey: derFromPubkey(device.pubkey),
+    origin: device.origin[0],
+  };
+};
--- a/src/frontend/src/utils/credential-devices.ts
+++ b/src/frontend/src/utils/credential-devices.ts
@@ -11,7 +11,7 @@ export type CredentialData = {
 const derFromPubkey = (pubkey: DeviceKey): DerEncodedPublicKey =>
   new Uint8Array(pubkey).buffer as DerEncodedPublicKey;
 
-export const convertToCredentialData = (
+export const convertToValidCredentialData = (
   device: Omit<DeviceData, "alias">
 ): CredentialData | undefined => {
   // In certain cases, e.g. Chrome on Windows 10, an invalid credential id is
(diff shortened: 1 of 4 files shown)


src/frontend/src/utils/credential-devices.ts
```diff
@@ -13,8 +13,22 @@ const derFromPubkey = (pubkey: DeviceKey): DerEncodedPublicKey =>
 
 export const convertToCredentialData = (
   device: Omit<DeviceData, "alias">
-): CredentialData => ({
-  credentialId: Buffer.from(device.credential_id[0] ?? []),
-  pubkey: derFromPubkey(device.pubkey),
-  origin: device.origin[0],
-});
+): CredentialData | undefined => {
+  // In certain cases, e.g. Chrome on Windows 10, an invalid credential id is
```
lmuntaner, line 17: Why did you move this logic in `convertToCredentialData`? It feels they are different conceptual things.

src/frontend/src/utils/credential-devices.ts
```diff
@@ -13,8 +13,22 @@ const derFromPubkey = (pubkey: DeviceKey): DerEncodedPublicKey =>
...
+  // not ignored but instead will result in a WebAuthn error that prevents a
+  // user from authenticating with any of their registered devices.
+  //
+  // Instead of throwing an error, we return `undefined` to make sure that the
+  // device will be filtered out to unblock the user from authenticating.
+  if (
+    device.credential_id.length !== 1 ||
+    device.credential_id[0].length === 0
```
lmuntaner, line 25: Was this the error? To not check the value inside the `credential_id`?

src/frontend/src/utils/credential-devices.ts
sea-snake: That was intentional, by moving it into the same function it guarantees credential data _always_ has a valid credential id.

I initially experimented with a `hasValidCredentialId` function which with the correct typing can guarantee that the credential id is not undefined in the device. But this didn't give the guarantee that the credential id is valid (more than 0 bytes) when it is set in the device. 

Checking the latter with a type guard would be difficult during TS compile - since we want to avoid runtime errors not create them by throwing an error - hence I opted instead to move the check into the function itself and force the caller to handle the case where a credential doesn't have a valid credential id.

src/frontend/src/utils/credential-devices.ts
sea-snake: The current implementation falls back to an empty buffer when a device doesn't have a credential id:

https://github.com/dfinity/internet-identity/blob/680f8a0aca77de5bf6f97b33daf98b51a3f2ac7b/src/frontend/src/utils/credential-devices.ts#L17

Previously this never happened, since a flatmap would filter out devices without a credential id before they're converted into credential data:

https://github.com/dfinity/internet-identity/blob/5a685b03d58f79f189a0964e73970fdbcd3c7b2c/src/frontend/src/utils/iiConnection.ts#L212

Since the current implementation falls back to an empty buffer, the WebAuthn API on some devices will throw an exception instead of ignoring these in the `allowedCredentials` list.

The changes and added tests in this PR, attempt to guarantee that credential data will never contain an empty buffer as credential id.

lmuntaner: Thanks for the clarification!

src/frontend/src/utils/credential-devices.ts
lmuntaner: My issue is that we might need "valid" credentials somewhere else in the code and in this case we just happen to be using the same converting function in those two places we also need valid credentials.

The same way I didn't realize the flatMap was important filtering out those credentials. I'm afraid the bug might be introduced again in the future. How about renaming the function to `convertToValidCredentialData`?

lmuntaner:
Already approving, but please check my comment about renaming instead of creating another function.

lmuntaner (approved):

src/frontend/src/utils/credential-devices.ts
sea-snake: Good idea, I'll rename the function.

PR #2762, opened 2024-12-31
Implement OpenID add/remove accounts in identity management

sea-snake:
Implement OpenID add/remove accounts in identity management

# Changes

- Add `OPENID_AUTHENTICATION` feature flag
- Add `II_OPENID_GOOGLE_CLIENT_ID` environment variable
- Copy over redirect in popup implementation from prototype
- Copy over request JWT implementation from prototype
- Implement "Linked Accounts" section to Identity manage (behind above flag)
- Implement add/remove account with mock actor and request JWT method
- Implement `hasOtherAuthMethods` to allow removal of all passkeys/credentials if one still has credentials/keys., since a user only needs to have at least 1 passkey or linked account.

image


src/frontend/src/styles/main.css
```diff
@@ -284,7 +284,7 @@
 
   --rs-list-item-stack: calc(var(--vs-stack) * 1.5);
   --rs-list-parcel-bezel: var(--vs-bezel);
-  --rs-list-gutter: var(--vs-gutter);
+  --rs-list-gutter: calc(var(--vs-gutter) * 2);
```
sea-snake, line 287: Horizontal spacing within list item was doubled as seen in the screenshot changes of this PR, previously this was rather narrow.

This increases the horizontal space between e.g. the icons and labels to be less cramped as seen in below example:

image

package.json
```diff
@@ -7,8 +7,8 @@
   "private": true,
   "license": "SEE LICENSE IN LICENSE.md",
   "scripts": {
-    "dev": "II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=1 vite",
-    "host": "II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=1 vite --host",
+    "dev": "II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=1 II_OPENID_GOOGLE_CLIENT_ID=45431994619-cbbfgtn7o0pp0dpfcg2l66bc4rcg7qbu.apps.googleusercontent.com vite",
```
lmuntaner, line 10: Is this the production client id? will we have only one or one per environment? What's the recommended way?

src/canister_tests/src/framework.rs
```diff
@@ -433,7 +433,7 @@ xr-spatial-tracking=()",
     let rgx = Regex::new(
         "^default-src 'none';\
 connect-src 'self' https:;\
-img-src 'self' data:;\
+img-src 'self' data: https://\\*.googleusercontent.com;\
```
lmuntaner, line 436: Is this to be able to show the image from the google account?

src/frontend/src/featureFlags/index.ts
```diff
@@ -1,6 +1,7 @@
 // Feature flags with default values
 const FEATURE_FLAGS_WITH_DEFAULTS = {
   DOMAIN_COMPATIBILITY: false,
+  OPENID_AUTHENTICATION: false,
```
lmuntaner, line 4: Will we want a flag per open id provider or can we enable other providers with one PR?

src/frontend/src/flows/manage/index.ts
```diff
@@ -245,7 +269,7 @@ export const renderManage = async ({
   // There's nowhere to go from here (i.e. all flows lead to/start from this page), so we
   // loop forever
   for (;;) {
-    let anchorInfo: IdentityAnchorInfo;
+    let anchorInfo: IdentityAnchorInfo & { credentials: OpenIDCredential[] };
```
lmuntaner, line 272: Do we want to move this type to it's own file to then easily change the import to the type created by the declaration?

We can do it in another PR though, not necessarily now.

src/frontend/src/flows/manage/index.ts
```diff
@@ -334,6 +369,30 @@ export const displayManage = (
...
+    const onLinkAccount = async () => {
+      const { jwt, salt } = await requestJWT();
+      const { iss, sub } = decodeJWT(jwt);
+      if (
+        credentials.find(
+          (credential) => credential.iss === iss && credential.sub === sub
+        )
+      ) {
+        toast.error("This account has already been linked");
+        return;
```
lmuntaner, line 381: Don't you need to `resolve()` here?

src/frontend/src/flows/manage/index.ts
```diff
@@ -334,6 +369,30 @@ export const displayManage = (
...
+      }
+      await connection.addJWT(jwt, new Uint8Array(salt));
+      resolve();
+    };
+    const onUnlinkAccount = async (credential: OpenIDCredential) => {
+      const name =
+        getMetadataString(credential.metadata, "name") ?? credential.sub;
+      if (!confirm(`Do you really want to unlink the account "${name}"?`)) {
```
lmuntaner, line 389: Can you use i18n for this message?

src/frontend/src/flows/manage/index.ts
```diff
@@ -334,6 +369,30 @@ export const displayManage = (
...
 
+    const onLinkAccount = async () => {
+      const { jwt, salt } = await requestJWT();
+      const { iss, sub } = decodeJWT(jwt);
+      if (
+        credentials.find(
+          (credential) => credential.iss === iss && credential.sub === sub
+        )
+      ) {
+        toast.error("This account has already been linked");
```
lmuntaner, line 380: can you use i18n for this message?

src/frontend/src/flows/manage/index.ts
```diff
@@ -291,21 +316,31 @@ function isPinAuthenticated(
...
   devices_: DeviceWithUsage[],
+  credentials: OpenIDCredential[],
   identityBackground: PreLoadImage
 ): Promise<void | AuthenticatedConnection> => {
   // Fetch the dapps used in the teaser & explorer
   // (dapps are suffled to encourage discovery of new dapps)
   const dapps = shuffleArray(getDapps());
+
+  // Create method to initiate JWT request
+  const identity = await ECDSAKeyIdentity.generate();
```
lmuntaner, line 331: Do we need to generate the identity here? Won't this be created every time users access this page?

Why can't we do this within `createRequestJWT`?

src/frontend/src/flows/manage/linkedAccountsSection.ts
lmuntaner, line 1: Could you add a screenshot of how the new section looks like?

Could you also add it to the showcase?

src/frontend/src/flows/manage/linkedAccountsSection.ts
```diff
@@ -0,0 +1,132 @@
...
+  const unlinkAvailable = credentials.length > 1 || hasOtherAuthMethods;
+
+  return html` <aside
+    class="l-stack c-card c-card--narrow"
+    data-role="linked-accounts"
+  >
+    <h2 class="t-title">Linked Accounts</h2>
+    <p class="t-paragraph t-lead">
+      ${credentials.length === 0
+        ? "Link your online account to hold assets and securely sign into dapps."
```
lmuntaner, line 35: Could you use i18n keys?

src/frontend/src/flows/manage/linkedAccountsSection.ts
```diff
@@ -0,0 +1,132 @@
...
+        : "Use your online accounts to hold assets and securely sign into dapps."}
```
lmuntaner, line 36: Ditto i18n keys

src/frontend/src/flows/manage/linkedAccountsSection.ts
```diff
@@ -0,0 +1,132 @@
...
+      </ul>
+      <div class="c-action-list__actions">
+        <button
+          ?disabled=${credentials.length >= MAX_CREDENTIALS}
+          class="c-button c-button--primary c-tooltip c-tooltip--onDisabled c-tooltip--left"
+          @click="${() => onLinkAccount()}"
+          id="linkAdditionalAccount"
+        >
+          <span class="c-tooltip__message c-card c-card--tight"
+            >You can link up to ${MAX_CREDENTIALS} online accounts. You must
```
lmuntaner, line 57: Ditto i18n keys

src/frontend/src/flows/manage/linkedAccountsSection.ts
```diff
@@ -0,0 +1,132 @@
...
+            unlink an account before you can link another.</span
+          >
+          <span
+            >${credentials.length === 0
+              ? "Link account"
```
lmuntaner, line 62: Ditto i18n keys

src/frontend/src/flows/manage/linkedAccountsSection.ts
```diff
@@ -0,0 +1,132 @@
...
+              : "Link additional account"}</span
```
lmuntaner, line 63: Ditto i18n keys

src/frontend/src/flows/manage/linkedAccountsSection.ts
```diff
@@ -0,0 +1,132 @@
...
+              ? settingsDropdown({
+                  alias: credential.sub,
+                  id: `account-${index}`,
+                  settings,
+                })
+              : ""
+          }
+        </div>
+        <div>
+          <div class="t-muted">Last used: ${lastUsageFormattedString}</div>
```
lmuntaner, line 127: Ditto i18n keys

src/frontend/src/flows/redirect.ts
```diff
@@ -0,0 +1,40 @@
+import { withLoader } from "$src/components/loader";
+
+export const callbackFlow = (): Promise<never> => {
+  window.opener.postMessage(window.location.href, window.location.origin);
```
lmuntaner, line 4: Why is this needed? Sorry, I forgot.

src/frontend/src/utils/openID.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  return id_token;
+};
+
+export const createRequestJWT = async (
+  config: RequestConfig,
+  options: RequestOptions
+): Promise<() => Promise<{ jwt: string; salt: ArrayBuffer }>> => {
+  const salt = new Uint8Array(32);
```
lmuntaner, line 124: Could we explain here why we use the `salt`, it might be useful when we lose the design doc :)

src/frontend/src/utils/openID.ts
```diff
@@ -0,0 +1,159 @@
...
+  configURL: "https://accounts.google.com/gsi/fedcm.json",
+};
+
+const requestWithCredentials = async (
+  config: Omit<RequestConfig, "authURL">,
+  options: Omit<RequestOptions, "principal">,
+  nonce: ArrayBuffer
+): Promise<string | undefined> => {
+  try {
+    const identityCredential = await navigator.credentials.get({
```
lmuntaner, line 37: Can we add a comment that this is `FedCM` request? My first thought was, why are you requesting a passkey again 😅

src/frontend/src/utils/mockOpenID.ts
```diff
@@ -43,19 +32,15 @@ export class MockOpenID {
         ["picture", { String: picture }],
       ],
     });
-    return Promise.resolve({ Ok: null });
+    return Promise.resolve();
```
lmuntaner, line 35: Why did you change the return type?

src/frontend/src/utils/mockOpenID.ts
```diff
@@ -17,20 +17,9 @@ export interface OpenIDCredential {
...
-  add_jwt(
-    _userNumber: UserNumber,
-    jwt: JWT,
-    _salt: Salt
-  ): Promise<{ Ok: null } | { Err: string }> {
+  add_jwt(_userNumber: UserNumber, jwt: JWT, _salt: Salt): Promise<void> {
     const [_header, body, _signature] = jwt.split(".");
     const { iss, sub, aud, email, name, picture } = JSON.parse(atob(body));
-    if (
-      this.#credentials.find(
```
lmuntaner, line 28: Why did you remove this check?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -918,17 +918,11 @@ export class AuthenticatedConnection extends Connection {
   };
 
   addJWT = async (jwt: JWT, salt: Salt): Promise<void> => {
-    const result = await this._mockOpenID.add_jwt(this.userNumber, jwt, salt);
```
lmuntaner, line 921: Why these changes? It made sense to have the `Ok` and `Err` return types.

src/frontend/src/styles/main.css
lmuntaner: Looks good to me

src/frontend/src/index.ts
```diff
@@ -43,6 +44,9 @@ void createSpa(async (connection) => {
   if (url.hash === "#authorize") {
     // User was brought here by a dapp for authorization
     return authFlowAuthorize(connection);
+  } else if (url.pathname === "/callback") {
```
lmuntaner, line 47: Can we add a comment when is this used?

lmuntaner (commented):
Thanks! Could you also provide some screenshots?

A few comments and questions. I like where this is going!

package.json
sea-snake: I would recommend one per environment (local, beta and prod), this would prevent JWT tokens from being usable across environments, similar to how passkeys are different between environments.

sea-snake: For now, I set the local environment client id in the dev script as default for development purposes. But since this client id is restricted to @dfinity accounts, which makes it unusable for non DFINITY devs, this should be replaced at a later point with instructions to creating and setting your own local client id for dev purposes.

src/canister_tests/src/framework.rs
sea-snake: Yes exactly, currently the CSP only allowed images from it's own canister and data uris. I considered fetching the image and storing it as datauri instead but that would add quite some complexity and prevent it from showing the latest image if the user changes their profile image.

src/frontend/src/featureFlags/index.ts
sea-snake: We can do either, by the time we get to other providers, this flag has already been removed and other more specific flags can be added (if needed). I made this flag less specific since it not just Google but OpenID as a whole behind it.

src/frontend/src/flows/manage/index.ts
sea-snake: Once the backend changes are made, the `IdentityAnchorInfo` itself should be updated to include this new property. For now, with the mocking approach, I added the type manually where needed.

src/frontend/src/flows/manage/index.ts
sea-snake: You could do it in `createRequestJWT` but that would remove the flexibility of choosing which identity and identity type you'd want to use. Actually here, the identity from the authenticated connection can be used, I'll update it to use that instead.

src/frontend/src/flows/manage/linkedAccountsSection.ts
sea-snake: I'll update the showcase to also include a manage flow with the flag enabled.

src/frontend/src/flows/redirect.ts
sea-snake: I'll add some comments, before I forget myself one day and run into this code 😅

This redirect in a popup flow avoids the need and complexity for II to navigate to a OpenID provider in the middle of a flow and navigate back, restoring the flow state from storage. Instead it opens a popup and receives the callback url from this popup through a postmessage, these post messages are restricted to the II origin to keep this callback data secure.

src/frontend/src/utils/mockOpenID.ts
sea-snake: Looking at the current passkey add/remove canister methods, I noticed they didn't have Ok/Err return types either and trap instead when something happens. So I wasn't sure at this point if we'd plan to follow this existing approach or if we'd implement Ok/Err return types. For now, until we get to the backend code, I removed the types to keep it simple until we have these details.

src/frontend/src/utils/iiConnection.ts
sea-snake: See my comment here: https://github.com/dfinity/internet-identity/pull/2762#discussion_r1900706820

src/frontend/src/flows/manage/index.ts
sea-snake: The state hasn't changed so there's no need to re-render the page (which is what `resolve()` does).

src/frontend/src/flows/manage/index.ts
sea-snake: Looked into it, and it seems like the i18n implementation in II is completely custom without support for interpolation :(

I'll change the translation label to "Do you really want to unlink this account?", I think this is worse UX but implementing string interpolation within the current custom i18n implementation isn't feasible either.

src/frontend/src/flows/manage/linkedAccountsSection.ts
sea-snake: As mentioned above, the current custom i18n implementation doesn't support interpolation, I've hardcoded the value (100) in the translation label for now.

src/frontend/src/index.ts
sea-snake: I've extended the comment to explain when and how it's used.

src/frontend/src/utils/openID.ts
sea-snake: I've added a TSDoc to the function :D

src/frontend/src/utils/openID.ts
sea-snake: I've split the salting into it's own function and added a TSDoc to explain the details.

src/frontend/src/flows/manage/index.ts
lmuntaner: Yes, I understand, my worry is that we might forget to remove the extra type everywhere were not necessary anymore after it's added to `IdentityAnchorInfo`.

package.json
lmuntaner: Ok, thanks for the clarification.

src/canister_tests/src/framework.rs
lmuntaner: Yes, this is fine. Maybe check with security by sending them a message, but it shouldn't block this PR.

src/frontend/src/featureFlags/index.ts
lmuntaner: Ok, thanks for the clarification.

src/frontend/src/flows/manage/index.ts
lmuntaner: Thanks for the clarification!

src/frontend/src/flows/manage/index.ts
lmuntaner: Yes, I know. Thanks!

src/frontend/src/flows/redirect.ts
lmuntaner: And then that popup is closed and disappears and the user is "redirected" back to where it all started?

src/frontend/src/utils/openID.ts
```diff
@@ -0,0 +1,195 @@
...
+ * @param options for the JWT request
+ */
+const requestWithRedirect = async (
+  config: Omit<RequestConfig, "configURL">,
+  options: RequestOptions
+): Promise<string> => {
+  const state = toBase64URL(window.crypto.getRandomValues(new Uint8Array(12)));
+
+  const redirectURL = new URL(window.location.origin);
+  redirectURL.pathname = "/callback";
```
lmuntaner, line 98: Do we want to create a constant `/callback` to be used here and in the index.ts?

src/frontend/src/utils/openID.ts
lmuntaner: Thanks!

src/frontend/src/utils/openID.ts
lmuntaner: Thanks!

src/frontend/src/utils/mockOpenID.ts
lmuntaner: Noted. This is one of the reasons to move the API v2, current one has no error management. I wonder whether we want to move to use the new endpoints before implementing google sign in. Wdyt?

src/frontend/src/utils/mockOpenID.ts
lmuntaner: Ok, makes sense.

lmuntaner (approved):
Thanks! A couple more comments but already approving!

src/frontend/src/flows/manage/index.ts
sea-snake: That's why the type is located in `mockOpenID.ts`, we'll get TS errors once this file is deleted until we remove the extra type everywhere :D

src/frontend/src/utils/mockOpenID.ts
sea-snake: I agree we should look into the new endpoints, no idea what their current status is. The Google sign in shouldn't rely on the new endpoints though since it would consist of additional endpoints and updating the auth guards (likely shared between new and old API).

PR #2769, opened 2025-01-09
Yubikey test feature

LXIF:
# Motivation

Recent changes in how macOS and Chrome appear to be handling hardware FIDO2 devices, we need to run tests with users who are experiencing major issues.

# Changes

Add HARDWARE_KEY_TEST feature flag, filter out all key_type='unknown'… authenticators if feature flag is enabled. At the very least in identities with only on-device and yubikey authenticators, this filters out all but the yubikey. The effect is that, since only the yubikey is requested by the browser, the UX is restored.

To activate the feature flag, run ```window.__featureFlags.HARDWARE_KEY_TEST.set(true)``` in the browser console.

# Tests

I have tested it manually.


lmuntaner (approved):
Thanks!

PR #2774, opened 2025-01-10
Implement timers to fetch Google certs.

sea-snake:
Implement timers to fetch Google certs.

# Changes

- Immediate and hourly timer to fetch Google certs, responses are valid for 5 hours so that should be enough margin.
- Exponential backoff retries if fetching Google certs has failed e.g. responses weren't all the same.
- Method to transform HTTP outcall responses so they're all the same and the outcall can succeed.

# Tests

Tested the timer (with smaller time values to test all cases) and it seems to correctly schedule the Google certificates fetching.


src/internet_identity/src/main.rs
```diff
@@ -371,6 +373,12 @@ fn acknowledge_entries(sequence_number: u64) {
     archive::acknowledge_entries(sequence_number)
 }
 
+/// This query method name _must_ match `google::TRANSFORM_CERTS_METHOD`.
+#[query]
+fn transform_google_certs(raw: http_request::TransformArgs) -> http_request::HttpResponse {
+    google::transform_certs(&raw)
+}
+
```
jasonz-dfinity, line 381: I think if you use [`http_request_with_closure`](https://docs.rs/ic-cdk/latest/ic_cdk/api/management_canister/http_request/fn.http_request_with_closure.html) (instead of `http_request`), you don't have to define your own canister method, and ic-cdk will take care of that for you.

Otherwise, you might want to use `#[query(hidden)]` since you probably don't want that on your .did file (since it's an internal method)

src/internet_identity/src/openid.rs
```diff
@@ -1 +1,37 @@
...
+    if let Ok(google_certs) = google::get_certs().await {
+        GOOGLE_CERTS.replace(google_certs);
+    } else {
+        // Try again with backoff if fetch failed, the HTTP outcall responses aren't the same
+        // across nodes when we fetch the certificates just at the moment of key rotation.
+        let backoff = Option::map(previous_delay, |value| value * 2).unwrap_or(60);
+        if backoff < FETCH_GOOGLE_CERTS_INTERVAL.as_secs() {
+            set_timer(Duration::from_secs(backoff), move || {
+                spawn(fetch_google_certs(Some(backoff)));
+            });
```
jasonz-dfinity, line 34: Calling `set_timer` here, combined with using `set_timer_interval` for regular operation, will make it possible for 2 invocations to interleave.

Example:
 * t = 0, `fetch_google_certs` is called, and `google::get_certs` is called
 * t = 5, `google::get_certs` resolves to failure, `fetch_google_certs` is scheduled to t=65
 * t = 65, `fetch_google_certs` is called, and `google::get_certs` is called
 * t = 70, `google::get_certs` resolves to failure, `fetch_google_certs` is scheduled to t=190
 * ...
 * t = 1940, `google::get_certs` resolves to failure, `fetch_google_certs` is scheduled to t=3860
 * t = 3605, fetch_google_certs` is called (because of interval), and `google::get_certs` is called

Around t = 3860, the backoff calls triggered the second interval might start to interleave with the calls triggered by the backoff of the first interval (although it's the last one). 


I think a simpler pattern is:

```
  fn schedule_fetch_google_certs(delay: Option<u64>) {
    ic_cdk_timers::set_timer(delay, || {
        spawn(async {
              if let Ok(google_certs) = google::get_certs().await {
                GOOGLE_CERTS.replace(google_certs);
                schedule_fetch_google_certs(Some(FETCH_GOOGLE_CERTS_INTERVAL));
              } else {
                let new_delay = min(FETCH_GOOGLE_CERTS_INTERVAL, delay.unwrap_or(60) * 2);
                schedule_fetch_google_certs(Some(new_delay));
              }
       })
    );
  }

  pub fn setup_timers() {
    schedule_fetch_google_certs(None);
  }
```

This way, there is no interleaving possible, while retaining a similar backoff mechanism.

sea-snake: Good idea, I've updated the code and used a match instead so I didn't have to repeat the `schedule_fetch_google_certs` fn call :D

src/internet_identity/src/main.rs
sea-snake: I saw the thread in the conversation in the sdk channel, I've updated the code to use the closure variant, had to manually enable this feature in `cargo.toml` to make it available.

src/internet_identity/src/openid/google.rs
```diff
@@ -56,14 +84,14 @@ pub async fn get_certs(transform_method: &str) -> Result<Vec<Jwk>, String> {
 // so we deserialize, sort the keys and serialize to make the response the same across all nodes.
 //
 // This function traps since HTTP outcall transforms can't return or log errors anyway.
-#[allow(unused)]
-pub fn transform_certs(raw: &TransformArgs) -> HttpResponse {
-    if raw.response.status != HTTP_STATUS_OK {
+#[allow(clippy::needless_pass_by_value)]
```
lmuntaner, line 87: Why can't you expect a reference? Is this type in the library?

lmuntaner (approved):
Thanks! That looks very nice,

I guess the comment on the query call and the exponential backoff in the description can be editted/removed, right?

src/internet_identity/src/openid/google.rs
sea-snake: > I guess the comment on the query call and the exponential backoff in the description can be editted/removed

Which comments are you referring to? The query call should have been replaced with a closure and comments are updated where needed.

sea-snake: > Why can't you expect a reference? Is this type in the library?

`http_request_with_closure` expects a method that takes a value instead of reference.

lmuntaner: > Which comments are you referring to?

The description of the PR. Sorry, I see now why you got confused.

src/internet_identity/src/openid/google.rs
```diff
@@ -1,36 +1,64 @@
...
+
+pub fn setup_timers() {
+    // Fetch the certs directly after canister initialization.
+    schedule_fetch_certs(None);
+}
+
+fn schedule_fetch_certs(delay: Option<u64>) {
+    set_timer(Duration::from_secs(delay.unwrap_or(0)), move || {
+        spawn(async move {
+            let new_delay = if let Ok(google_certs) = fetch_certs().await {
```
LXIF, line 42: I think a match statement would be more idiomatic and easier to read here:

```rust
let new_delay = match fetch_certs().await {
        Ok(google_certs) => {
            CERTS.replace(google_certs);
            FETCH_CERTS_INTERVAL
        },
        Err(_) => {
            // Try again earlier with backoff if fetch failed, the HTTP outcall responses
            // aren't the same across nodes when we fetch at the moment of key rotation.
            min(FETCH_CERTS_INTERVAL, delay.unwrap_or(60) * 2)
        }
```

src/internet_identity/src/openid/google.rs
```diff
@@ -56,14 +84,14 @@ pub async fn get_certs(transform_method: &str) -> Result<Vec<Jwk>, String> {
...
         trap("Invalid response status")
     };
 
-    let certs: GoogleCerts = serde_json::from_slice(raw.response.body.as_slice())
-        .unwrap_or_else(|_| trap("Invalid JSON"));
+    let certs: GoogleCerts =
+        serde_json::from_slice(response.body.as_slice()).unwrap_or_else(|_| trap("Invalid JSON"));
 
     let mut sorted_keys = certs.keys.clone();
     sorted_keys.sort_by_key(|key| key.kid().unwrap().to_owned());
```
LXIF, line 97: Assuming .kid() cannot fail (otherwise a naked unwrap would be ill-advised for prod)

LXIF (approved):
Looks good, optional changes see comments

src/internet_identity/src/openid/google.rs
sea-snake: In theory it could if Google were to send invalid certs, I'll update it to unwrap or else trap.

src/internet_identity/src/openid/google.rs
sea-snake: According to clippy if else was more idiomatic since there were only 2 cases and Err value is ignored 😅

But I agree match is more readable since it puts the fetch_certs and Ok on separate lines, so I've updated the code to use a match.

PR #2778, opened 2025-01-13
Functionality to add current device to current origin

lmuntaner:
# Motivation

When the user is accessing II from a device which was registered in another domain, while having another device in the current docmain, it will have to retry the login.

Therefore, we added a new step that requests the user to add the current device to the current origin as well.

The screen was already developed, this PR uses the screen and adds the functionality.

# Changes

* Rename `addFIDODevice` to `addCurrentDevice` because it wasn't really only FIDO device.
* Use `addCurrentDevice` helper in `addCurrentDeviceScreen` when the user selects to add the device.
* Show the new step if `showAddCurrentDevice` is `true` from the `authenticatedBox` function. Both in the manage and login flow.

# Tests

Tested in beta environment. See video attached.

### DIFF
--- a/src/frontend/src/flows/addDevice/manage/addCurrentDevice.ts
+++ b/src/frontend/src/flows/addDevice/manage/addCurrentDevice.ts
@@ -27,17 +27,21 @@ const displayFailedToAddDevice = (error: Error) =>
   });
 
 /**
- * Add a new device (i.e. a device connected to the browser the user is
- * currently using, like a YubiKey, or FaceID, or, or. Not meant to be used to
- * add e.g. _another_ browser, macbook or iPhone.)
+ * Add a new device form the current browser.
+ *
+ * Used to add a device connected to the browser the user is
+ * currently using, like a YubiKey, or FaceID.
+ *
+ * Used to add current device in the current origin.
+ *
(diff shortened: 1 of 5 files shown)


lmuntaner:
New UX:

https://github.com/user-attachments/assets/0a683495-22da-4b7f-97b9-0c8c334eb023

lmuntaner:

src/frontend/src/flows/addDevice/manage/addCurrentDevice.ts
```diff
@@ -27,17 +27,21 @@ const displayFailedToAddDevice = (error: Error) =>
...
  * @param userNumber anchor to add the device to
  * @param connection authenticated II connection
  * @param devices already existing devices
  */
-export const addFIDODevice = async (
+export const addCurrentDevice = async (
   userNumber: bigint,
   connection: AuthenticatedConnection,
-  devices: DeviceData[]
+  devices: Omit<DeviceData, "alias">[]
```
lmuntaner, line 44: This was needed to be able to use the devices from `lookUpCredentials`. And anyway the field `"alias"` was not used within this function.

sea-snake (approved):

PR #2783, opened 2025-01-16
Wrap Google open id config in option so it can be disabled.

sea-snake:
Wrap Google open id config in option so it can be disabled. Additionally, change client id to struct to allow multiple properties besides client id for each open id client.

# Tests

- Updated existing test to go from enabled -> disabled

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -241,8 +241,8 @@ type CaptchaConfig = record {
 // Each field is wrapped is `opt` to indicate whether the field should
 // keep the previous value or update to a new value (e.g. `null` keeps the previous value).
 //
-// Some fields, like `openid_google`, have an additional nested `opt`,
-// this indicates enable/disable status (e.g. `opt null` disables a feature).
+// Some fields, like `openid_google`, have an additional nested `opt`, this indicates
+// enable/disable status (e.g. `opt null` disables a feature while `null` leaves it untouched).
 type InternetIdentityInit = record {
     // Set lowest and highest anchor
     assigned_user_number_range : opt record {
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -237,7 +237,12 @@ type CaptchaConfig = record {
 };
 
 // Init arguments of II which can be supplied on install and upgrade.
-// Setting a value to null keeps the previous value.
+//
+// Each field is wrapped is `opt` to indicate whether the field should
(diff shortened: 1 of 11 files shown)


src/internet_identity/internet_identity.did
```diff
@@ -237,7 +237,12 @@ type CaptchaConfig = record {
 };
 
 // Init arguments of II which can be supplied on install and upgrade.
-// Setting a value to null keeps the previous value.
+//
+// Each field is wrapped is `opt` to indicate whether the field should
+// keep the previous value or update to a new value (e.g. `null` keeps the previous value).
+//
+// Some fields, like `openid_google`, have an additional nested `opt`,
+// this indicates enable/disable status (e.g. `opt null` disables a feature).
```
lmuntaner, line 245: Shall we add that `null` doesn't change the feature, whereas `opt null` disabled it.

src/internet_identity/tests/integration/config.rs
lmuntaner, line 1: Should we add a test that when `Some(None)` is passed it "disables" the open id? Could we use another endpoint instead of the `config` one?

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -188,6 +188,13 @@ pub struct AnchorCredentials {
     pub recovery_phrases: Vec<PublicKey>,
 }
 
+/// Init arguments of II which can be supplied on install and upgrade.
+///
+/// Each field is wrapped in `Option<>` to indicate whether the field should
+/// keep the previous value or update to a new value (e.g. `None` keeps the previous value).
+///
+/// Some fields, like `openid_google`, have an additional nested `Option<>`,
+/// this indicates enable/disable status (e.g. `Some(None)` disables a feature).
```
lmuntaner, line 197: Ditto to say that `Some(None) disables a feature whereas None leaves it untouched`

lmuntaner (approved):
I understand the point, it makes sense for nicer API to disable / enable different Open ID providers.

Approving, but take a look whether you can add the test and also improve the comment.

src/internet_identity/internet_identity.did
lmuntaner: Oh, and thanks for adding this explanation!

src/internet_identity/tests/integration/config.rs
sea-snake: As discussed, this is covered in the integration test. In another PR I'll attempt to clear/split up these tests.

PR #2784, opened 2025-01-17
Add storage utils to persist cancelled RP IDs

lmuntaner:
# Motivation

We want to give a second good UX to users who go through a failed attempt to login and need to retry.

In this PR, I introduce the storage utils needed to persist the cancelled RP IDs that will be reused to improve the experience.

# Changes

* Add three storage utils to add, read and clean up cancelled RP IDs per user number.

# Tests

* Add tests for the three new utils.

### DIFF
--- a/src/frontend/src/storage/index.ts
+++ b/src/frontend/src/storage/index.ts
@@ -197,15 +197,15 @@ export const setKnownPrincipal = async ({
  */
 export const cleanUpRpIdMapper = async (userNumber: bigint) => {
   await withStorage((storage) => {
-    const ix = userNumber.toString();
+    const anchorIndex = userNumber.toString();
     const anchors = storage.anchors;
-    const oldAnchor = anchors[ix];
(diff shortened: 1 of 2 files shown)


lmuntaner:

src/frontend/src/storage/index.ts
```diff
@@ -189,6 +189,94 @@ export const setKnownPrincipal = async ({
...
 
+/**
+ * Sets the `cancelledRpIdsMapper` field to an empty object.
+ *
+ * This is needed when a user removes a device from their account.
+ * @param userNumber {bigint} The anchor number.
+ */
+export const cleanUpRpIdMapper = async (userNumber: bigint) => {
+  await withStorage((storage) => {
+    const ix = userNumber.toString();
```
sea-snake, line 200: nit: I would call this `anchorIndex` to clarify that this is used as index for the anchors array.

lmuntaner: Thanks, done!

PR #2785, opened 2025-01-17
Refactor config integration tests.

sea-snake:
Refactor config integration tests to clarify what cases are covered, utility methods are implemented for repetitive checks e.g. assert config with config returned from query.


src/internet_identity/tests/integration/config.rs
```diff
@@ -2,189 +2,216 @@ use canister_tests::api::internet_identity as api;
...
+/// this maps `None` to `Some(None)` for `Option<Option>` since it does not return `None`.
+fn assert_config(
+    env: &PocketIc,
+    canister_id: CanisterId,
+    expected_config: &InternetIdentityInit,
+) -> Result<(), CallError> {
+    assert_eq!(
+        api::config(env, canister_id)?,
+        InternetIdentityInit {
+            openid_google: expected_config.openid_google.clone().or(Some(None)),
```
lmuntaner, line 47: Why is this one necessary like this?

src/internet_identity/tests/integration/config.rs
```diff
@@ -2,189 +2,216 @@ use canister_tests::api::internet_identity as api;
...
+    config: &InternetIdentityInit,
+    expected_config: &InternetIdentityInit,
+) -> Result<(), CallError> {
+    upgrade_ii_canister_with_arg(env, canister_id, II_WASM.clone(), Some(config.clone()))?;
+    assert_config(env, canister_id, expected_config)?;
+    Ok(())
+}
+
+/// Utility method to upgrade with other config change and check if any other config hasn't changed
+fn unrelated_change_and_assert(
```
lmuntaner, line 77: `unrelated_change` to what? To another config? Which other config? One with optionals?

src/internet_identity/tests/integration/config.rs
```diff
@@ -2,189 +2,216 @@ use canister_tests::api::internet_identity as api;
...
+        canister_id,
+        II_WASM.clone(),
+        Some(other_config.clone()),
+    )?;
+    assert_config(env, canister_id, &expected_config)?;
+    Ok(canister_id)
+}
 
 #[test]
 fn should_retain_anchor_on_user_range_change() -> Result<(), CallError> {
```
lmuntaner, line 108: Wasn't there something missing in this test? which Anchor is retained?

src/internet_identity/tests/integration/config.rs
```diff
@@ -2,189 +2,216 @@ use canister_tests::api::internet_identity as api;
...
 
-    let canister_id = install_ii_canister_with_arg(&env, II_WASM.clone(), Some(config.clone()));
-
-    assert_eq!(api::config(&env, canister_id)?, config);
-
-    let _ = upgrade_ii_canister_with_arg(&env, canister_id, II_WASM.clone(), None);
-
-    assert_eq!(api::config(&env, canister_id)?, config);
+    let canister_id = install_and_assert(&env, &config)?;
+    upgrade_ii_canister_with_arg(&env, canister_id, II_WASM.clone(), Some(config.clone()))?;
```
lmuntaner, line 137: Shouldn't this upgrade using `None`?

src/internet_identity/tests/integration/config.rs
```diff
@@ -2,189 +2,216 @@ use canister_tests::api::internet_identity as api;
...
 
-    let canister_id = install_ii_canister_with_arg(&env, II_WASM.clone(), Some(config.clone()));
+    // Check if related origins are updated and then untouched
+    let canister_id_1 = install_and_assert(&env, &config_1)?;
+    upgrade_and_assert(&env, canister_id_1, &config_2, &config_2)?;
+    unrelated_change_and_assert(&env, canister_id_1)?;
 
-    assert_eq!(api::config(&env, canister_id)?, config);
+    // Check if related origins are disabled and then untouched
+    let canister_id_2 = install_and_assert(&env, &config_1)?;
```
lmuntaner, line 169: Why are there different canister ids? Shouldn't they all be the same?

src/internet_identity/tests/integration/config.rs
```diff
@@ -2,189 +2,216 @@ use canister_tests::api::internet_identity as api;
...
+    upgrade_and_assert(&env, canister_id_2, &disabled_config, &disabled_config)?;
```
lmuntaner, line 170: Wouldn't it be better if the confi passed to `upgrade_and_assert` had mostly `None` fields except for the changes we are interested and then we expect the fields to be present from before?

If config logic substituted a `None` for some value, we wouldn't catch it here, right?

src/internet_identity/tests/integration/config.rs
```diff
@@ -2,189 +2,216 @@ use canister_tests::api::internet_identity as api;
...
+    unrelated_change_and_assert(&env, canister_id_1)?;
 
-    let expected_config_3 = InternetIdentityInit {
-        related_origins: Some(related_origins_2.clone()),
-        openid_google: Some(openid_google2.clone()),
-        ..expected_config_2
-    };
+    // Check if open id config is disabled and then untouched
+    let canister_id_2 = install_and_assert(&env, &config_1)?;
+    upgrade_and_assert(&env, canister_id_2, &disabled_config, &disabled_config)?;
```
lmuntaner, line 208: Ditto as before

lmuntaner (commented):
Looks much nicer, thanks!

I added some comments.

sea-snake:
@lmuntaner

Please review, I've taken a different approach with separate tests per config option due to:
- Nuanced differences between options are missed otherwise.
- Explicitly checks all possible config state transitions for each config option.
- Adding a new options is more straightforward, just add another file based on any existing file and add additional options specific tests where needed.

src/internet_identity/tests/integration/config/register_rate_limit.rs
```diff
@@ -0,0 +1,153 @@
...
+        related_origins: Some(vec!["https://example.com".into()]),
+        ..Default::default()
+    };
+
+    let canister_id = install_ii_canister_with_arg(&env, II_WASM.clone(), Some(config.clone()));
+    config.related_origins = Some(vec![]);
+    upgrade_ii_canister_with_arg(&env, canister_id, II_WASM.clone(), Some(config.clone())).unwrap();
+    assert_eq!(
+        api::config(&env, canister_id).unwrap().related_origins,
+        config.related_origins
```
lmuntaner, line 82: Could we compare it to `vec![]` directly?

src/internet_identity/tests/integration/config/analytics_config.rs
```diff
@@ -0,0 +1,177 @@
...
+    config.analytics_config = Some(None);
+    upgrade_ii_canister_with_arg(&env, canister_id, II_WASM.clone(), Some(config.clone())).unwrap();
+    assert_eq!(
+        api::config(&env, canister_id).unwrap().analytics_config,
+        config.analytics_config
+    );
+}
+
+#[test]
+fn should_disable_config() {
```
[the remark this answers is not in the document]
sea-snake: > I think this is should_enable_config and the previous should_disable_config.

Ah good catch!

> As I understand, enabling is going from Some(Some)) to Some(None) and disabling the opposite, right?

Uhm, the other way around. This exactly how straightforward it is to make the mistake 😅

src/internet_identity/tests/integration/config/openid_google.rs
```diff
@@ -0,0 +1,142 @@
...
+        })),
+        ..Default::default()
+    };
+
+    let canister_id = install_ii_canister_with_arg(&env, II_WASM.clone(), Some(config.clone()));
+    config.openid_google = Some(None);
+    upgrade_ii_canister_with_arg(&env, canister_id, II_WASM.clone(), Some(config.clone())).unwrap();
+    assert_eq!(
+        api::config(&env, canister_id).unwrap().openid_google,
+        config.openid_google
```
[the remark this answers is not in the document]
sea-snake: Same issue as you mentioned above, should be handled with the same approach.

src/internet_identity/tests/integration/config/register_rate_limit.rs
sea-snake: To be consistent, I'd rather use the same approach as you've mentioned above.

lmuntaner: Sounds good

src/internet_identity/tests/integration/config/captcha_config.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    let canister_id = install_ii_canister_with_arg(&env, II_WASM.clone(), Some(config.clone()));
+    config.captcha_config = Some(CaptchaConfig {
+        max_unsolved_captchas: 300,
+        captcha_trigger: Static(CaptchaDisabled),
+    });
+    upgrade_ii_canister_with_arg(&env, canister_id, II_WASM.clone(), Some(config.clone())).unwrap();
+    assert_eq!(
+        api::config(&env, canister_id).unwrap().captcha_config,
+        config.captcha_config
```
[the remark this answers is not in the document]
lmuntaner: Thanks!

lmuntaner (approved):

PR #2781, opened 2025-01-14
Remember successful rpId per anchor

lmuntaner:
# Motivation

Draft PR to test storing the successful RP ID in the localstorage.

# Changes

# Tests

PR #2786, opened 2025-01-17
Clean up cancelled RP IDs after removing device

lmuntaner:
# Motivation

The calculation of the RP ID depends on the devices. When a device is removed, a cancelled RP ID could be the solution. Therefore, we clean up the mapper that stores RP IDs.

# Changes

* Add `userNumber` in `deleteDevice` to call `cleanUpRpIdMapper` in parallel to removing the device.
* Pass the `userNumber` when `deleteDevice` is called.

# Tests

Tested in beta with #2781 PR


lmuntaner:

LXIF (approved):
LGTM

PR #2787, opened 2025-01-17
Persist cancelled RP IDs

lmuntaner:
# Motivation

We want to improve the UX of users with credentials created in different domains. In their case, we might need to iterate through RP IDs to find the right one. Therefore, remembering which didn't work will avoid bad UX on returning users.

# Changes

* Move the stored cancelled RP IDs from an instance variable in `Connection` to browser storage.

# Tests

* Same tests apply. I added one extra test to check that a new `Connection` instance works the same way.


lmuntaner:

LXIF (approved):
LGTM

PR #2788, opened 2025-01-20
Expose origin of adding a device

lmuntaner:
# Motivation

We want to avoid a bad UX for users when we migrate to `internetcomputer.org`. The bad UX appears when there are devices registered in multiple origins.

In this PR, the origin used for the `addDevice` flow is exposed as an argument instead of hardcoded to `window.origin`. This will be used in upcoming PRs to configure the RP ID of where to register the new device.

# Changes

* Add `origin` parameter to `addDevice` flow and use it.
* Pass `origin` as `window.origin` for now to not change the functionality.

# Tests

No functionality changed.


lmuntaner:

PR #2792, opened 2025-01-20
Add device based on credentials' origin

lmuntaner:
# Motivation

Draft PR that enforces that new devices added through adding one device use the same origin as current credentials.

# Changes

# Tests

<hr/><details><summary>🟡 Some screens were changed</summary></details>

PR #2793, opened 2025-01-21
Util getCredentialsOrigin

lmuntaner:
# Motivation

Make sure that users don't have a bad UX when logging in from different domains. The bad UX comes from having devices registered in different domains. Therefore, we want to avoid that a user can register devices in different domains.

In this PR, a new util is introduced to get the common origin of all user's device.

# Changes

* New util `getCredentialsOrigin`.
* Remove `DEFAULT_ORIGIN` from `findWebauthnRpId` move it to a constants file.

# Tests

* Test new util.
* An example of how this is used can be found in #2792 

### DIFF
--- a/src/frontend/src/utils/credential-devices.ts
+++ b/src/frontend/src/utils/credential-devices.ts
@@ -1,5 +1,5 @@
 import { DeviceData, DeviceKey } from "$generated/internet_identity_types";
-import { iiLegacyOrigin } from "$showcase/constants";
+import { iiLegacyOrigin } from "$src/constants";
 import { DerEncodedPublicKey } from "@dfinity/agent";
 import { supportsWebauthRoR } from "./userAgent";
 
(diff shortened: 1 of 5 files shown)


lmuntaner:

src/frontend/src/utils/credential-devices.ts
```diff
@@ -32,3 +34,41 @@ export const convertToValidCredentialData = (
...
+ * it can lead to bad UX when calculating the RP ID.
+ *
+ * Therefore, we want to avoid devices registered in multiple origins.
+ *
+ * First, it checks whether the browser supports ROR.
+ * Second, it checks whether all the devices have the same origin.
+ * - If they do, it returns the origin.
+ * - If they don't, it returns `undefined`.
+ *
+ * @param credentials
```
sea-snake, line 52: Nit: the `userAgent` is missing from the doc.

src/showcase/src/constants.ts
```diff
@@ -3,6 +3,7 @@ import { getDapps } from "$src/flows/dappsExplorer/dapps";
 
 export const dapps = getDapps();
 
+// If credentials have no origin, it's because they were created in this domain
 export const iiLegacyOrigin = "https://identity.ic0.app";
```
sea-snake, line 7: Why is this in the showcase?

lmuntaner: Oops, thanks! wrong place indeed!

lmuntaner:

sea-snake (approved):

PR #2794, opened 2025-01-21
Pass all credentials when logging in

lmuntaner:
# Motivation

In case there is a bug when setting the origin of a device, we don't want to prevent that device from being used.

Therefore, we pass all the credentials when getting the `MultiWebAuthnIdentity` instead of the filtered ones. This doesn't hurt the UX.

# Changes

* Use `credentials` instead of `filteredCredentials` when creating `MultiWebAuthnIdentity`.

# Tests

* Tested in beta that UX is the same.


lmuntaner:

sea-snake (approved):

PR #2796, opened 2025-01-21
Show add current device screen once a week

lmuntaner:
# Motivation

In this PR, I change the logic to only show the page the first time and then only if the anchor was used more than a week ago.

# Changes

# Tests


lmuntaner:
Done in multiple PRs finishing in #2801

PR #2798, opened 2025-01-21
Add a new timestamp and make used one optional in anchor storage

lmuntaner:
# Motivation

In this PR, I add a new optional field `lastShownAddCurrentDevicePage`.

# Changes

# Tests

### DIFF
--- a/src/frontend/src/storage/index.ts
+++ b/src/frontend/src/storage/index.ts
@@ -244,7 +244,8 @@ export const addAnchorCancelledRpId = async ({
 
     storage.anchors[anchorIndex] = {
       ...oldAnchor,
-      lastUsedTimestamp: nowMillis(),
+      // Only set on successful login.
+      lastUsedTimestamp: undefined,
       cancelledRpIdsMapper: {
         ...cancelledRpIdsMapper,
         [origin]: originCancelledRpIds,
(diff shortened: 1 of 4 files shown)


src/frontend/src/storage/index.ts
```diff
@@ -177,14 +179,20 @@ export const setKnownPrincipal = async ({
...
     // Only keep the more recent N principals
     const prunedPrincipals = dedupedPrincipals.slice(0, MAX_SAVED_PRINCIPALS);
 
     // Here we try to be as non-destructive as possible and we keep potentially unknown
     // fields
-    storage.anchors[ix] = { ...oldAnchor, knownPrincipals: prunedPrincipals };
+    storage.anchors[ix] = {
+      ...oldAnchor,
+      knownPrincipals: prunedPrincipals,
+      lastUsedTimestamp: nowMillis(),
```
lmuntaner, line 194: This was missing. Setting a known principal didn't change the `lastTimeUsedTimestamp` which seems like a bug.

src/frontend/src/storage/index.ts
```diff
@@ -269,12 +278,38 @@ export const getCancelledRpIds = async ({
 }: {
   userNumber: bigint;
   origin: string;
-}): Promise<Set<string | undefined>> => {
+}): Promise<{
+  cancelledRpIds: Set<string | undefined>;
+  lastShownAddCurrentDevicePage: number | undefined;
```
lmuntaner, line 283: This new field will be used to calculate whether the screen to add current device should be shown or not.

src/frontend/src/storage/index.ts
```diff
@@ -677,7 +712,7 @@ const PrincipalDataV3 = z.object({
 
 const AnchorV3 = z.object({
   /** Timestamp (mills since epoch) of when anchor was last used */
-  lastUsedTimestamp: z.number(),
+  lastUsedTimestamp: z.number().optional(),
```
lmuntaner, line 715: For backwards compatibility, we have to also make the AnchorV3 optional.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -405,7 +405,7 @@ export class Connection {
     credentials: CredentialData[]
   ): Promise<LoginSuccess | WebAuthnFailed | PossiblyWrongRPID | AuthFail> => {
     // Get cancelled rpids for the user from local storage.
-    const cancelledRpIds = await getCancelledRpIds({
+    const { cancelledRpIds } = await getCancelledRpIds({
```
lmuntaner, line 408: Using the new field will be done in another PR.

PR #2800, opened 2025-01-21
Expose origin of add iiConnection

lmuntaner:
# Motivation

We will use RP IDs also when adding new devices.

That means that the origin when adding a new device might not be the same as the device.

In this PR, I create the parameter `origin` for the `add` method instead of using `readDeviceOrigin`.

Idea for another PR, move from list of arguments to an object argument to understand what is what when calling `add`.

# Changes

* New method `origin` in `add` method in iiConnection.
* Move `readDeviceOrigin` to its own utils file.
* Use `readDeviceOrigin` where `add` is called.

# Tests

No new functionality is added in this PR.


src/frontend/src/utils/readDeviceOrigin.ts
```diff
@@ -0,0 +1,14 @@
+import { isNullish } from "@dfinity/utils";
+
+// Reads the "origin" used to infer what domain a FIDO device is available on.
+// The canister only allow for 50 characters, so for long domains we don't attach an origin
+// (those long domains are most likely a testnet with URL like <canister id>.large03.testnet.dfinity.network, and we basically only care about identity.ic0.app & identity.internetcomputer.org).
+//
+// The return type is odd but that's what our didc version expects.
```
sea-snake, line 7: > The return type is odd but that's what our didc version expects.

This part of the comment is no longer relevant

src/frontend/src/utils/readDeviceOrigin.ts
```diff
@@ -0,0 +1,14 @@
+import { isNullish } from "@dfinity/utils";
+
+// Reads the "origin" used to infer what domain a FIDO device is available on.
+// The canister only allow for 50 characters, so for long domains we don't attach an origin
```
sea-snake, line 4: > The canister only allow for 50 characters, so for long domains we don't attach an origin

Maybe we should move this comment and check down to where we pass the origin as canister call arguments?Particularly since this function is not directly used there anymore either, origin is now coming from a param.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -736,7 +737,7 @@ export class AuthenticatedConnection extends Connection {
       key_type: keyType,
       purpose,
       protection,
-      origin: readDeviceOrigin(),
+      origin: origin === undefined ? [] : [origin],
```
sea-snake, line 740: Maybe move length check and comment to explain here instead?

src/frontend/src/utils/readDeviceOrigin.ts
lmuntaner: Yes, that makes sense.

LXIF (approved):
LGTM

src/frontend/src/utils/readDeviceOrigin.ts
lmuntaner: I removed the whole util.

lmuntaner:
Thanks @LXIF !

sea-snake (approved):

PR #2801, opened 2025-01-21
Show add current device page only once

lmuntaner:
# Motivation

In this PR, I change the logic to only show the page the first time and when the user hasn't seen it in a week.

# Changes

* Use the new field `lastShownAddCurrentDevicePage` to calculate `showAddCurrentDevice` value.
* Call `setLastShownAddCurrentDevicePage` when the add current device screen is shown.

# Tests

* Add a test to check how `showAddCurrentDevice` evolves with time.

### DIFF
--- a/src/frontend/src/flows/addDevice/addCurrentDevice.ts
+++ b/src/frontend/src/flows/addDevice/addCurrentDevice.ts
@@ -1,5 +1,6 @@
 import { infoScreenTemplate } from "$src/components/infoScreen";
 import { I18n } from "$src/i18n";
+import { setLastShownAddCurrentDevicePage } from "$src/storage";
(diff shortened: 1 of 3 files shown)


lmuntaner:
The page I reference is the following one:

Screenshot 2025-01-21 at 15 03 36

Which is shown after a user has to go through the passkey flow twice because the first RP ID was incorrect.

lmuntaner:

LXIF (approved):
LGTM - I would change the user prompt a bit though: "...to improve the user experience..." => "...to improve your user experience..." to make clear that the user does this for their own benefit.

lmuntaner:
Thanks @LXIF ! I'll do the text changes in another PR.

PR #2802, opened 2025-01-21
Add origin to addCurrentDevice function

lmuntaner:
# Motivation

We want to avoid bad UX for users logging in from different domains. The bad UX comes when the user has devices registered in multiple domains. Therefore, we want to keep the new devices in the same origin as the old devices.

In this PR, we add the functionality to choose an origin when using the `addCurrentDevice` function.

# Changes

* Add `origin` parameter to `addCurrentDevice` which is then converted to RP ID and passed to `createOptions`.
* Pass the `origin` to `addCurrentDevice` in `addDevice` flow. This still doesn't change the functoinality because the origin passed is `window.origin` and it needs to be calculated depending on the devices origins.
* Rename `addCurrentDevice` page to `registerCurrentDeviceCurrentOrigin` to make it clear that the device will be registered in the current origin.

# Tests

This PR doesn't include new functionality. It allows future functionality when we choose the origin based on the devices.

<hr/><details><summary>🟢 Some screens were added</summary></details><details><summary>🟡 Some screens were changed</summary></details><details><summary>🔴 Some screens were removed</summary></details>


lmuntaner:

lmuntaner:
Thanks! Ditto, I'll do that in anothe PR.

PR #2806, opened 2025-01-22
Change the warning icon in devices list

lmuntaner:
# Motivation

At the moment, there is a warning icon saying that the current device won't work if the origin where it was registered it's different than the current origin. This is not the case anymore with the domains compatibility.

In this PR, we show an info icon of where the device was registered, only if the devices where not registered all in the same origin.

# Changes

* Add new field `info` to `Authenticator`.
* Calculate new field with `domainInfo` in `devicesFromDevicesWithUsage`.
* Templating helper `itemInfo`.
* Use new template helper to render the new icon in `authenticatorItem`.
* Improve documentation of the fields.

# Tests

Add tests for `devicesFromDevicesWithUsage`. Not complete, but at least to test the current behavior I added.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -628,7 +628,7 @@ export const domainWarning = (
   }
 };
 
-export const domainInfo = (
+const domainInfo = (
   device: DeviceData,
   allDevices: DeviceData[]
 ): TemplateResult | undefined => {
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -30,6 +30,7 @@ import {
 import { addPhrase, recoveryWizard } from "$src/flows/recovery/recoveryWizard";
 import { setupKey, setupPhrase } from "$src/flows/recovery/setupRecovery";
 import { I18n } from "$src/i18n";
+import { getCredentialsOrigin } from "$src/utils/credential-devices";
 import { AuthenticatedConnection, Connection } from "$src/utils/iiConnection";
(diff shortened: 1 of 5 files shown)


lmuntaner:
New UI:

Screenshot 2025-01-22 at 10 13 29

lmuntaner:

sea-snake:
> New UI:

Isn't a green info icon confusing? We already show a green icon when a user has added a recovery method. Maybe it should be a neutral grey color or blue (like the info toast).

src/frontend/src/flows/manage/index.ts
```diff
@@ -593,27 +599,48 @@ export const domainWarning = (
...
+    window.location.origin === LEGACY_II_URL &&
+    deviceOrigin !== window.location.origin
+  ) {
     return html`This Passkey may not be usable on the current URL
-    (${window.origin})`;
+    (${window.location.origin})`;
   }
 
   // In general, if this is _not_ the _old_ II, then it's most likely the _new_ II, meaning all devices should have an origin attached.
   if (isNullish(deviceOrigin)) {
```
sea-snake, line 616: Shouldn't this be `window.origin !== LEGACY_II_URL && isNullish(deviceOrigin)`, if I read the comment above correctly?

lmuntaner: This is part of the warning icon. I didn't touch the logic. Just added the flag guard and changed to `window.location.origin`.

lmuntaner:
@sea-snake yes, true. Better grey. Here it is:

Screenshot 2025-01-22 at 17 22 11

lmuntaner:

sea-snake (approved):

PR #2810, opened 2025-01-22
Add and remove OpenID credentials from anchor

sea-snake:
# Changes

- OpenID credentials in stable structures.
- Lookup anchor number by OpenID credential in stable structures.
- Anchor management methods to add/remove OpenID credentials.
- Anchor management method to lookup anchor with OpenID credential.
- Canister methods to add/remove OpenID credentials.
- Replaced frontend mock OpenID implementation with above canister methods.
- Use canister config endpoint to get Google client id for FedCM interaction. 

# Tests

Added additional anchor tests:
- should_add_openid_credential
- should_remove_openid_credential
- should_update_openid_credential

Added additional storage test:
- should_write_and_update_openid_credential_lookup


src/internet_identity/src/storage.rs
```diff
@@ -128,13 +130,15 @@ const EMPTY_SALT: [u8; 32] = [0; 32];
...
 
 /// MemoryManager parameters.
-const ANCHOR_MEMORY_INDEX: u8 = 0u8;
-const ARCHIVE_BUFFER_MEMORY_INDEX: u8 = 1u8;
-const PERSISTENT_STATE_MEMORY_INDEX: u8 = 2u8;
-const EVENT_DATA_MEMORY_INDEX: u8 = 3u8;
-const STATS_AGGREGATIONS_MEMORY_INDEX: u8 = 4u8;
-const REGISTRATION_REFERENCE_RATE_MEMORY_INDEX: u8 = 5u8;
-const REGISTRATION_CURRENT_RATE_MEMORY_INDEX: u8 = 6u8;
+const ANCHOR_MEMORY_INDEX: u8 = 0;
```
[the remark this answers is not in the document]
sea-snake: I removed explicit type in the value since the variable is already typed: `const X: u8 = 5u8` -> `const X: u8 = 5`.

But yeah, I'll revert it, PR is already big enough as is that these nits are just adding confusion.

src/internet_identity/src/storage.rs
```diff
@@ -372,7 +397,9 @@ impl<M: Memory + Clone> Storage<M> {
     /// Writes the data of the specified anchor to stable memory.
     pub fn write(&mut self, data: Anchor) -> Result<(), StorageError> {
         let anchor_number = data.anchor_number();
-        let storable_anchor = StorableAnchor::from(data);
+
+        // Write fixed 4KB anchor
```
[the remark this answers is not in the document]
sea-snake: It's defined in the top of the file:

```
//! This module implements all the stable memory interactions of Internet Identity.
//! It uses the [Reader] and [Writer] implementations of the `stable_structures` crate.
//!
//! ## Stable Memory Layout
//!
//! Variables used below:
//! * HEADER_SIZE: 66 bytes
//! * ENTRY_OFFSET: 131 072 bytes = 2 WASM Pages
//! * Anchor size: 4096 bytes
```

src/internet_identity/src/storage/anchor.rs
```diff
@@ -293,6 +333,43 @@ impl Anchor {
...
+            .openid_credentials
+            .iter()
+            .position(|entry| entry.iss == iss && entry.sub == sub)
+        else {
+            return Err(AnchorError::OpenIdCredentialNotFound);
+        };
+        Ok(index)
+    }
+
+    pub fn add_openid_credential(
```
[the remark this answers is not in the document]
sea-snake: It's used by a method annotated with `#[allow(unused)]` in `anchor_management.rs`.

src/internet_identity/src/storage/stable_anchor.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::storage::anchor::Anchor;
+use candid::CandidType;
+use ic_stable_structures::storable::Bound;
+use ic_stable_structures::Storable;
+use serde::Deserialize;
+use std::borrow::Cow;
+
+#[derive(CandidType, Deserialize, Clone)]
+pub struct StableAnchor {
+    pub openid_credentials: Option<Vec<OpenIdCredential>>,
```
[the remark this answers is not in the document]
sea-snake: I made it a vector since every entry struct already contains the `iss` and `sub`, repeating these two values as a tuple key when making it a map would only add a risk of this tuple key not matching it's value, basically not having a single source of truth. It follows the same approach as devices, which also could be a map based on e.g. `pubkey`.

And you're right, this one doesn't need to optional indeed, I'll update it.

src/internet_identity/src/openid.rs
```diff
@@ -1,19 +1,27 @@
...
+#[derive(Debug, PartialEq, Eq, CandidType, Deserialize, Clone)]
 pub struct OpenIdCredential {
-    pub iss: String,
-    pub sub: String,
-    pub aud: String,
-    pub principal: Principal,
+    pub iss: Iss,
+    pub sub: Sub,
+    pub aud: Aud,
+    pub delegation_principal: Principal,
```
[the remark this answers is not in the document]
sea-snake: To clarify that this principal is the delegation principal created based on this OpenID credential data, previously the name "principal" wasn't explicit in this regard. Also practically speaking, the did file didn't like `principal` as record key, I could escape it, but I'd rather avoid it by not naming it `principal` to begin with.

src/internet_identity/src/openid.rs
```diff
@@ -57,6 +65,34 @@ pub fn verify(jwt: &str, salt: &[u8; 32]) -> Result<OpenIdCredential, String> {
     })
 }
 
+#[allow(clippy::cast_possible_truncation)]
+fn calculate_seed(client_id: &str, iss: &Iss, sub: &Sub) -> Hash {
```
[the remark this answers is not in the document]
sea-snake: Will do :D

src/internet_identity/src/storage.rs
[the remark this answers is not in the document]
lmuntaner: Thanks!

src/internet_identity/src/storage.rs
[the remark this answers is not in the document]
lmuntaner: Ah, I don't know I always thought it was 7. I forgot it was 4. Thanks!

src/internet_identity/src/storage/anchor.rs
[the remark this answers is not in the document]
lmuntaner: Ok, thanks!

src/internet_identity/src/storage/anchor.rs
```diff
@@ -293,6 +333,43 @@ impl Anchor {
...
+            .iter()
+            .any(|entry| entry.iss == openid_credential.iss && entry.sub == openid_credential.sub)
+        {
+            return Err(AnchorError::DuplicateOpenIdCredential);
+        }
+        self.openid_credentials.push(openid_credential);
+        Ok(())
+    }
+
+    pub fn remove_openid_credential(&mut self, iss: &str, sub: &str) -> Result<(), AnchorError> {
```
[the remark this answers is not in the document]
lmuntaner: ok, thanks!

src/internet_identity/src/storage/stable_anchor.rs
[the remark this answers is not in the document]
lmuntaner: Ok, makes sense.

Please check with Jason about the new data structures.

I will also give it another review tomorrow morning with a fresh head.

jasonz-dfinity (approved):

lmuntaner:
@sea-snake after sleeping over this new memory, I can't find a reason not to move forward. Looks good to me!

Let me know when the PR is ready for the final review.

src/internet_identity/src/storage.rs
```diff
@@ -387,20 +414,78 @@ impl<M: Memory + Clone> Storage<M> {
 
         writer.write_all(&buf).expect("memory write failed");
         writer.flush().expect("memory write failed");
+
+        // Write unbounded stable structures anchor
+        let previous_stable_anchor = self.stable_anchor_memory.get(&anchor_number);
+        let stable_anchor = StableAnchor::from(data);
+        self.stable_anchor_memory
+            .insert(anchor_number, stable_anchor.clone());
```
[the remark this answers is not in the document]
sea-snake: Thanks!

src/internet_identity/src/storage.rs
```diff
@@ -387,20 +414,78 @@ impl<M: Memory + Clone> Storage<M> {
...
+        anchor_number: AnchorNumber,
+        previous: Vec<OpenIdCredential>,
+        current: Vec<OpenIdCredential>,
+    ) {
+        let credential_to_be_removed = previous
+            .iter()
+            .filter(|p| current.iter().all(|c| c.iss == p.iss && c.sub == p.sub));
+        let credential_to_be_added = current
+            .iter()
+            .filter(|c| previous.iter().all(|p| p.iss == c.iss && p.sub == c.sub));
```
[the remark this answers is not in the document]
sea-snake: You're right, updated it to use diff (was actually looking for something this straightforward).

src/internet_identity/src/storage/anchor.rs
```diff
@@ -293,6 +333,43 @@ impl Anchor {
...
+
+    fn openid_credential_index(&self, iss: &str, sub: &str) -> Result<usize, AnchorError> {
+        let Some(index) = self
+            .openid_credentials
+            .iter()
+            .position(|entry| entry.iss == iss && entry.sub == sub)
+        else {
+            return Err(AnchorError::OpenIdCredentialNotFound);
+        };
+        Ok(index)
```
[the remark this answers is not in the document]
sea-snake: Thanks!

src/internet_identity/src/storage/stable_anchor.rs
```diff
@@ -0,0 +1,31 @@
...
+    const BOUND: Bound = Bound::Unbounded;
+}
+
+impl From<Anchor> for StableAnchor {
+    fn from(anchor: Anchor) -> Self {
+        Self {
+            openid_credentials: anchor.openid_credentials().clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
sea-snake: That indeed makes a lot more sense, managed to implement the 2nd option.

src/internet_identity/src/openid.rs
```diff
@@ -1,19 +1,25 @@
...
+use sha2::{Digest, Sha256};
 use std::cell::RefCell;
 use std::collections::HashMap;
 
 mod google;
 
-#[derive(Debug, PartialEq)]
+pub type Iss = String;
+pub type Sub = String;
+pub type Aud = String;
```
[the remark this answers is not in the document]
sea-snake: Good idea!

src/internet_identity/src/storage/stable_anchor.rs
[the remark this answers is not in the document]
jasonz-dfinity: Nice, then should `impl From<Anchor> for StableAnchor` be deleted? I believe clippy doesn't catch useless trait impl

src/internet_identity/src/storage.rs
```diff
@@ -387,20 +414,75 @@ impl<M: Memory + Clone> Storage<M> {
...
+            self.lookup_anchor_with_openid_credential_memory.remove(key);
+        });
+        credential_to_be_added.for_each(|key| {
+            self.lookup_anchor_with_openid_credential_memory
+                .insert(key.clone(), anchor_number);
+        });
+    }
+
+    #[allow(unused)]
+    pub fn lookup_anchor_with_openid_credential(
```
jasonz-dfinity, line 484: If I'm not mistaken, this function isn't tested, which means the logic in `update_lookup_anchor_with_openid_credential` is also not covered by any test?

src/internet_identity/src/storage/anchor/tests.rs
```diff
@@ -482,6 +486,94 @@ fn should_not_allow_reserved_metadata_key() {
...
+    assert_eq!(
+        anchor.add_openid_credential(openid_credential_1.clone()),
+        Ok(())
+    );
+
+    // Check if an already added OpenID credential results in an error
+    assert_eq!(
+        anchor.openid_credentials,
+        vec![openid_credential_0.clone(), openid_credential_1]
+    );
```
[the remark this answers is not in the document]
sea-snake: You're right, updated.

src/internet_identity/src/storage/stable_anchor.rs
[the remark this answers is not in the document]
sea-snake: Yeah, missed that one.

src/internet_identity/src/storage.rs
sea-snake: You're right, added a test that covers all the cases. Thanks!

src/internet_identity/src/anchor_management.rs
```diff
@@ -165,3 +187,38 @@ pub fn identity_metadata_replace(
...
+    Ok(Operation::RemoveOpenIdCredential { iss: iss.into() })
+}
+
+/// Updates an `OpenIdCredential` of the given anchor, used to update details like the metadata.
+/// Return an error if the `OpenIdCredential` to be updated does not exist.
+#[allow(unused)]
+pub fn update_openid_credential(
+    anchor: &mut Anchor,
+    openid_credential: OpenIdCredential,
+) -> Result<(), AnchorError> {
```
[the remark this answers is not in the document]
sea-snake: Yeah that's intentional, compared to updating a device, updating a credential doesn't result in any operation that needs to be stored in the archive canister. This is because a device its metadata can change after user action e.g. alias change while an OpenID credential will be updated on every JWT verification (to keep latest claims from OpenID provider stored). 

Additionally, I intentionally only keep track of the issuer in operations for the archive canister, anything beyond that would likely impact a user's privacy. This data is enough to keep track of adoption of a given OpenId without knowing details beyond that.

src/archive/archive.did
lmuntaner, line 1: Thanks, good thinking about the archive canister!

src/canister_tests/src/api/archive.rs
lmuntaner, line 1: Will you ad them in anoter PR?

src/internet_identity/internet_identity.did
```diff
@@ -293,6 +293,8 @@ type IdentityAnchorInfo = record {
     devices : vec DeviceWithUsage;
     // Device registration status used when adding devices, see DeviceRegistrationInfo
     device_registration: opt DeviceRegistrationInfo;
+    // OpenID accounts linked to this anchor
+    openid_credentials: opt vec OpenIdCredential;
```
lmuntaner, line 297: This is used in the old API. Could you add it in the v2 API as well?

I believe in `IdentityInfo`.

src/internet_identity/src/openid/google.rs
```diff
@@ -300,6 +303,15 @@ fn verify_claims(client_id: &String, claims: &Claims, salt: &[u8; 32]) -> Result
     if now < claims.iat * NANOSECONDS_PER_SECOND {
         return Err("JWT is not valid yet".into());
     }
+    if claims.email.as_ref().is_some_and(|val| val.len() > 256) {
```
lmuntaner, line 306: why 256? can we put it in a variable?

src/internet_identity/src/openid/google.rs
```diff
@@ -300,6 +303,15 @@ fn verify_claims(client_id: &String, claims: &Claims, salt: &[u8; 32]) -> Result
...
+        return Err("Email too long".into());
+    }
+    if claims.name.as_ref().is_some_and(|val| val.len() > 128) {
```
lmuntaner, line 309: Ditto, why 128 and can we put it in a variable?

src/internet_identity/src/storage.rs
```diff
@@ -387,20 +416,77 @@ impl<M: Memory + Clone> Storage<M> {
 
         writer.write_all(&buf).expect("memory write failed");
         writer.flush().expect("memory write failed");
+
+        // Write unbounded stable structures anchor
```
lmuntaner, line 420: `// Write and read old unbouded stable structures anchor`

You are also using the old one, right?

lmuntaner (approved):
Thanks!! LGTM.

This means we need to prepare the archive for the next release as well.

src/internet_identity/src/storage.rs
```diff
@@ -387,20 +416,77 @@ impl<M: Memory + Clone> Storage<M> {
...
+        // Update `OpenIdCredential` to `AnchorNumber` lookup map
+        let previous_openid_credentials = previous_stable_anchor
+            .map(|anchor| anchor.openid_credentials)
+            .unwrap_or_default();
+        let current_openid_credentials = stable_anchor.openid_credentials;
+        self.update_lookup_anchor_with_openid_credential(
+            anchor_number,
+            previous_openid_credentials,
+            current_openid_credentials,
+        );
```
jasonz-dfinity, line 434: It's not super clear to me which layer is responsible for making sure that there are no 2 anchors with the same OpenIdCredentialKey. I think either way works:

1. if it's on the user of the storage layer, perhaps the rustdoc should mention this responsibility of the caller 
2. if it's on the storage layer, then you might need to define another storage layer error, and check whether there is clobbering in the `update_lookup_anchor_with_openid_credential` function.

I don't think this issue necessarily blocks this PR, since it can be a future improvement (perhaps you already planned to do it)

jasonz-dfinity (approved):

src/internet_identity/src/storage.rs
```diff
@@ -203,6 +212,13 @@ pub struct Storage<M: Memory> {
     current_registration_rate_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
     /// Memory wrapper used to report the size of the reference registration rate memory.
     reference_registration_rate_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
+    /// Memory wrapper used to report the size of the stable anchor memory.
+    stable_anchor_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
+    stable_anchor_memory: StableBTreeMap<AnchorNumber, StableAnchor, ManagedMemory<M>>,
+    /// Memory wrapper used to report the size of the lookup anchor with OpenID credential memory.
+    lookup_anchor_with_openid_credential_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
+    lookup_anchor_with_openid_credential_memory:
+        StableBTreeMap<StorableOpenIdCredentialKey, AnchorNumber, ManagedMemory<M>>,
```
lmuntaner, line 221: Support multiple Anchor numbers for the same OpenId credential.

Storage layer should allow for it, but not the code. This way, we are not blocking us from doing it in the future.

lmuntaner (changes requested):

src/internet_identity/src/openid/google.rs
sea-snake: Done, it's just a sanity check with a value that has a reasonable margin for the given data.

src/internet_identity/src/openid/google.rs
sea-snake: Done, it's just a sanity check with a value that has a reasonable margin for the given data.

src/internet_identity/src/storage.rs
sea-snake: To be more clear, I'll call it the previous value. Using the word "old" might be confusing, for a moment I thought you referred to the 4KB anchor here 😅

src/internet_identity/src/storage.rs
sea-snake: You're right, based on your comment and the comment from @lmuntaner above regarding keeping the storage flexible I added it to storage layer for now. The stored data supports multiple anchors but the storage layer implementation will limit it to one anchor number only.

src/canister_tests/src/api/archive.rs
sea-snake: Added them in this PR

lmuntaner (approved):
Thanks!

dfx.json
```diff
@@ -5,7 +5,7 @@
       "candid": "src/internet_identity/internet_identity.did",
       "wasm": "internet_identity.wasm.gz",
       "build": "bash -c 'II_DEV_CSP=1 II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=${II_DUMMY_CAPTCHA:-1} scripts/build'",
-      "init_arg": "(opt record { captcha_config = opt record { max_unsolved_captchas= 50:nat64; captcha_trigger = variant {Static = variant {CaptchaDisabled}}}})",
+      "init_arg": "(opt record { captcha_config = opt record { max_unsolved_captchas= 50:nat64; captcha_trigger = variant {Static = variant {CaptchaDisabled}}}; openid_google = opt opt record { client_id = \"45431994619-cbbfgtn7o0pp0dpfcg2l66bc4rcg7qbu.apps.googleusercontent.com\" }})",
```
lmuntaner, line 8: Is this our dev environment? Did you check with IT about the prod one?

src/frontend/src/flows/manage/index.ts
```diff
@@ -402,9 +416,14 @@ export const displayManage = async (
     };
 
     const onLinkAccount = async () => {
+      const googleClientId = getGoogleClientId();
+      if (isNullish(googleClientId)) {
+        toast.error(copy.linking_google_accounts_is_unavailable);
```
lmuntaner, line 421: Does this message already exist? I can't find it in `main`.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -952,12 +951,26 @@ export class AuthenticatedConnection extends Connection {
...
+    ]);
+    if ("Err" in res) throw new CanisterError(res.Err);
   };
 
-  removeJWT = async (iss: string, sub: string): Promise<void> => {
-    await this._mockOpenID.remove_jwt(this.userNumber, iss, sub);
+  // Get previously fetched config, else fetch it
+  getConfig = (): Promise<InternetIdentityInit> => {
+    this.configPromise =
+      this.configPromise ?? this.getActor().then((actor) => actor.config());
```
lmuntaner, line 972: Does this do a query or an update call?

src/internet_identity/src/anchor_management.rs
```diff
@@ -165,3 +188,47 @@ pub fn identity_metadata_replace(
     anchor.replace_identity_metadata(metadata)?;
     Ok(Operation::IdentityMetadataReplace { metadata_keys })
 }
+
+/// Adds an `OpenIdCredential` to the given anchor and returns the operation to be archived.
+/// Returns an error if the `OpenIdCredential` already exists in this or another anchor.
+#[allow(unused)]
```
lmuntaner, line 194: This can be removed, right? Isn't it used?

src/internet_identity/src/anchor_management.rs
```diff
@@ -165,3 +188,47 @@ pub fn identity_metadata_replace(
...
+        return Err(AnchorError::OpenIdCredentialAlreadyRegistered);
+    }
+    Ok(Operation::AddOpenIdCredential {
+        iss: openid_credential.iss,
+    })
+}
+
+/// Removes an `OpenIdCredential` of the given anchor and returns the operation to be archived.
+/// Return an error if the `OpenIdCredential` to be removed does not exist.
+#[allow(unused)]
```
lmuntaner, line 210: Ditto

src/internet_identity/src/anchor_management.rs
```diff
@@ -165,3 +188,47 @@ pub fn identity_metadata_replace(
...
+    key: &OpenIdCredentialKey,
+) -> Result<Operation, AnchorError> {
+    anchor.remove_openid_credential(key)?;
+    let (iss, _) = key;
+    Ok(Operation::RemoveOpenIdCredential { iss: iss.clone() })
+}
+
+/// Updates an `OpenIdCredential` of the given anchor, used to update details like the metadata.
+/// Return an error if the `OpenIdCredential` to be updated does not exist.
+#[allow(unused)]
```
lmuntaner, line 222: Ditto

src/internet_identity/src/anchor_management.rs
```diff
@@ -165,3 +188,47 @@ pub fn identity_metadata_replace(
...
+pub fn update_openid_credential(
+    anchor: &mut Anchor,
+    openid_credential: OpenIdCredential,
+) -> Result<(), AnchorError> {
+    anchor.update_openid_credential(openid_credential)
+}
+
+/// Lookup `AnchorNumber` for the given `OpenIdCredentialKey`.
+#[allow(unused)]
```
lmuntaner, line 231: Ditto

src/internet_identity/src/storage/anchor.rs
```diff
@@ -293,6 +336,50 @@ impl Anchor {
...
+    }
+
+    fn openid_credential_index(&self, key: &OpenIdCredentialKey) -> Result<usize, AnchorError> {
+        self.openid_credentials
+            .iter()
+            .position(|entry| &entry.key() == key)
+            .ok_or(AnchorError::OpenIdCredentialNotFound)
+    }
+
+    pub fn add_openid_credential(
```
lmuntaner, line 351: Is it here that we check that only one per anchor is added?

src/internet_identity/src/storage/anchor/tests.rs
lmuntaner, line 1: Shoud we add a test here that the same open id credential can't be added to multiple anchors?

lmuntaner (changes requested):
Thanks! I left a few comments.

This PR is too big for my taste, it's hard to keep track of all the implications. Please test it carefully before merging.

I'm requesting changes to prevent merging before the comments are addressed.

src/frontend/src/flows/manage/index.ts
sea-snake: It's in `linkedAccountsSection.json`.

src/frontend/src/utils/iiConnection.ts
sea-snake: Hmm, I assumed it was an update call since it was mentioned we use update calls to make sure it goes through consensus. But now I see it's a query call. What do you recommend as approach here?
- Force this query call to be ran as update call, no backend changes
- Update method to be an update call, breaking changes but method is mentioned in did as internal
- Other ideas?

src/internet_identity/src/storage/anchor.rs
sea-snake: Added this check in anchor management, not sure on which layer you'd recommend to add this check?
- anchor management (check is currently here)
- anchor
- storage

src/frontend/src/flows/manage/linkedAccountsSection.json
```diff
@@ -9,6 +9,10 @@
     "link_additional_account": "Link additional account",
     "max_linked_accounts_reached": "You can link up to 100 online accounts. You must unlink an account before you can link another.",
     "unlink": "Unlink",
-    "last_used": "Last used"
+    "last_used": "Last used",
+    "authentication_failed": "Authentication failed",
+    "jwt_signature_invalid": "The account could not be linked due to an invalid signature",
+    "account_not_found": "The account could not be found",
+    "linking_google_accounts_is_unavailable": "Linking Google accounts is unavailable"
```
lmuntaner, line 16: I would add: "Please try again later."

dfx.json
sea-snake: It's a dev environment I created, only works internally within DFINITY. This should be removed at a later point and replaced with instructions on creating a client id and configuring it with II.

Haven't gotten around to ask IT about the prod one yet, we'll need to also ask them to create others for our other environments.

src/frontend/src/flows/manage/linkedAccountsSection.json
```diff
@@ -9,6 +9,10 @@
     "link_additional_account": "Link additional account",
     "max_linked_accounts_reached": "You can link up to 100 online accounts. You must unlink an account before you can link another.",
     "unlink": "Unlink",
-    "last_used": "Last used"
+    "last_used": "Last used",
+    "authentication_failed": "Authentication failed",
+    "jwt_signature_invalid": "The account could not be linked due to an invalid signature",
```
lmuntaner, line 14: Can you add a "." at the end?

src/frontend/src/flows/manage/linkedAccountsSection.json
```diff
@@ -9,6 +9,10 @@
...
+    "account_not_found": "The account could not be found",
```
lmuntaner, line 15: Ditto "." at the end.

src/internet_identity/src/storage/anchor/tests.rs
sea-snake: See my above comment, if it's clear on which layer, I'll add a test there.

src/frontend/src/utils/iiConnection.ts
lmuntaner: I would discuss this with security.

Leave it as is and add a TODO to be discussed with them.

src/internet_identity/src/storage/anchor.rs
lmuntaner: No, that makes sense. I saw it now.

Do you have a test for this?

src/internet_identity/src/storage/anchor/tests.rs
lmuntaner: Yes, that makes sense to me. We need a diagram to better understand how the layers talk to each other to be honest.

Adding a test would be great, thanks!

src/frontend/src/utils/iiConnection.ts
sea-snake: Added TODO comment as suggested.

lmuntaner (approved):

PR #2811, opened 2025-01-22
Use devices origin when setting up recovery

lmuntaner:
# Motivation

Register recovery device in the same origin as the other credentials.

This is not strictly necessary, but it can be useful if in the future we want to consider the recovery device as one more authenticator.

# Changes

* Calculate the origin based on the current devices and use it if flag is enabled.

# Tests

* Tested in beta.


lmuntaner:

PR #2813, opened 2025-01-23
Do not use cancelled RP IDs for recovery device

lmuntaner:
# Motivation

If a user has a cancelledRpId and that is the origin where the recovery device was registered, then the recovery wouldn't work because it would be filtered out.

We don't track the cancelled RP IDs for the recovery devices and we don't need to because there is only one recovery device.

In this PR, I added a flag to `fromWebauthnCredentials` to skip reading and storing cancelled RP IDs.

# Changes

* Add new flag parameter in `fromWebauthnCredentials` and skip reading and storing cancelled RP IDs when set to `true`.
* Use the new flag in the recovery device flow.

# Tests

* Add tests for the new functionality.

### DIFF
--- a/src/frontend/src/flows/recovery/recoverWith/device.ts
+++ b/src/frontend/src/flows/recovery/recoverWith/device.ts
@@ -115,5 +115,12 @@ const attemptRecovery = async ({
     .map(convertToValidCredentialData)
     .filter(nonNullish);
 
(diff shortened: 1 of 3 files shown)


lmuntaner:

sea-snake (approved):

LXIF (approved):

PR #2814, opened 2025-01-23
Check devices origin in the post recovery flow

lmuntaner:
# Motivation

Users can also add a device after they use the recovery device.

In this PR, we use the devices to set the origin of the device that might be added after recovering with a device.

# Changes

* Use the devices and calculate their common origin to then use it when the device is added.

# Tests


lmuntaner:

lmuntaner:
UX:

In this video the user starts with an device registered in ic0.app and the recovery device also in ic0.app.

The user goes through the recovery flow in internetcomputer.org. After the recovery, the user registers the current device's passkey. However, the device is registered in ic0.app instead of the current domain internetcomputer.org.

This is what the PR is doing.

We can see that because if the passkey would be registered in internetcomputer.org, we would see the info icons in the list of devices specifying where each was registered. This info icons appear when the devices were registered in different domains but not if all the devices have the same domain.

I also tested this by logging in with the iCloud passkey afterwards and it used the RP ID ic0.app.

https://github.com/user-attachments/assets/bae8d99b-f7bb-42d5-a0c7-badaa5ed6e6a

sea-snake (approved):

PR #2815, opened 2025-01-23
Pass `origin` to `add` call in addCurrentDevice

lmuntaner:
# Motivation

This PR uses the origin of the devices in the flow to add a new device triggered from the Manage page.

# Changes

* Calculate the new device origin based on devices and pass it to `addDevice`.
* Pass `origin` if present in to `add` call in `addCurrentDevice`.

# Tests

* Tested in beta domains.

### DIFF
--- a/src/frontend/src/flows/addDevice/manage/addCurrentDevice.ts
+++ b/src/frontend/src/flows/addDevice/manage/addCurrentDevice.ts
@@ -81,7 +81,7 @@ export const addCurrentDevice = async (
         { authentication: null },
         newDevice.getPublicKey().toDer(),
         { unprotected: null },
-        window.origin,
+        origin ?? window.origin,
         newDevice.rawId
       )
     );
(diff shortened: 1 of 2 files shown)


lmuntaner:

lmuntaner:
@sea-snake @LXIF wait for the review. Not yet.

lmuntaner:
UX:

The user starts with a device in ic0.app and logs in internetcomputer.org.

Then, it clicks to add a new device. Where we can see that the URL to add it will be ic0.app and not internetcomputer.org.

The user adds a FIDO device which we can see that it was registered in ic0.app because if it had been in another domain we would see the info warnings.

I also tried logging in with the new device in ic0.app and it worked in the first try.

https://github.com/user-attachments/assets/ba99bf78-8cf7-4ded-9611-43d1854a9095

lmuntaner:

sea-snake (approved):

PR #2817, opened 2025-01-24
Use passkeys including recovery device for origin in recovery wizard

lmuntaner:
# Motivation

In the future, we might want to remove the idea of "recovery device" and use it as authenticator. Therefore, we should take it into account when calculating the origin of the new device to avoid bad UX for having devices registered in multiple origins.

# Changes

* Introduce a new method in Connection to return all the passkeys.
* Use the new method in recoveryWizard.

# Tests

PENDING


lmuntaner:
I don't like this idea. I'm closing it. I don't see how this can be helpful.

If the recovery device is in the same origin, this does nothing.
If the recovery device is in another origin, the new device might be added yet in another origin. It would be better if the new device is in the same origin as other devices.

What we want to do is that when a recovery device is created, it also checks the authenticators.

PR #2822, opened 2025-01-28
Clean up cancelled RP IDs if no filtered credentials

lmuntaner:
# Motivation

We were adding domains as cancelled when the user closed the webauthn modal. We were assuming that was because the RP ID was incorrect.

However, there is the edge case that the RP ID is correct and the user cancelled manually. We can't differentiate both cancels.

In this PR, we clean up the cancelled RP IDs if there are no credentials for the remaining origins.

# Changes

* Add a check before calling `findWebAuthnRpId` to clean up the persisted storage and the local variables.

# Tests

* Added a unit test for it and changed the expectation for another one.
* See screenshot for the UX.

### DIFF
--- a/src/frontend/src/utils/iiConnection.test.ts
+++ b/src/frontend/src/utils/iiConnection.test.ts
@@ -241,6 +241,69 @@ describe("Connection.login", () => {
       }
     });
 
+    it("connection cleans up cancelled RP IDs if no credentials when user cancels in a valid RP ID", async () => {
+      // This one would fail because it's not the device the user is using at the moment.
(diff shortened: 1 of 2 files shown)


lmuntaner:
Before this fix, the user would get stuck, being sent to an error page until they clean up the browser storage.

New UX:

https://github.com/user-attachments/assets/d241333c-de78-4766-a655-c4c82cc81f73

lmuntaner:
@sea-snake please review

sea-snake (approved):

PR #2823, opened 2025-01-29
Better error for PIN identities in other domains

lmuntaner:
# Motivation

If a user was registered with PIN in one domain and tries to log in another domain, it sees an "Unknown user" error.

This will be a problem if we migrate to another domain.

In this PR, a new return type is added to manage this specific case and show a better info message suggesting next steps.

# Changes

* Add a new error return to `login`: `PinUserOtherDomain`.
* If the login produces this error, show a toast with next steps.

# Tests

* Tested in beta as seen in the video.
* New test in `iiConnection.test.ts`.

### DIFF
--- a/src/frontend/src/components/authenticateBox/index.ts
+++ b/src/frontend/src/components/authenticateBox/index.ts
@@ -33,6 +33,7 @@ import {
   InvalidCaller,
   LoginSuccess,
   NoRegistrationFlow,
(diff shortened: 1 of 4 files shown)


lmuntaner:
@sea-snake please review

sea-snake (approved):

sea-snake (approved):

PR #2825, opened 2025-01-29
Iframe webauthn postmessage

sea-snake:
# Motivation

Implement WebAuthn through iframes to be able to authenticate from one origin while the user is on another. This makes it possible as a user to interact with the same identity on multiple origins with the same passkey.

# Changes

- Implement internal cross-origin II to II post message communication.
- Use this communication channel to implement method that requests WebAuthn credential within an iframe from another related origin.
- Update permission policy header to allow cross origin WebAuthn API access.
- Update csp to allow II to embed itself cross-origin

# Tests

- Updated tests to include permission policy and csp header changes


src/frontend/src/utils/internalPostMessage.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * Utility methods to reliably and securely establish an internal cross-origin
+ * post message communication channel between two related origins.
+ */
+
+const READY_INTERVAL = 100;
+const READY_TIMEOUT = 3000;
+
+interface ReadyRequest {
```
lmuntaner, line 9: We have zod installed, do you want to use it?

sea-snake: Wasn't aware of that, but it's internal calls from II to II, so the data is trusted and shouldn't need validation. 

The only risk here is browser extensions sending post messages (e.g. NordPass) on behalf of the II origin, those messages should be ignored, hence the top level identifying key e.g. `ii_ready_request` that can be used in run time to make sure that we're looking at the right messages.

lmuntaner: Fair enough.

src/frontend/src/flows/iframeWebAuthn.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (
+      event.source === targetWindow &&
+      event.origin === targetOrigin &&
+      isCredentialRequest(event.data)
+    ) {
+      try {
+        const credential = (await navigator.credentials.get(
+          event.data.ii_credential_request.options
+        )) as PublicKeyCredential;
+        window.parent.postMessage(
```
lmuntaner, line 88: Could you create the object result object before and type it with the expected type?

It's hard to read otherwise.

src/frontend/src/utils/internalPostMessage.ts
```diff
@@ -0,0 +1,78 @@
...
+        ),
+      READY_INTERVAL
+    );
+    const timeout = setTimeout(() => {
+      clearInterval(interval);
+      reject("Ready timed out");
+    }, READY_TIMEOUT);
+  });
+
+export const waitForWindowReadyRequest = (
```
lmuntaner, line 58: Could you add a comment on what this does?

src/frontend/src/utils/internalPostMessage.ts
```diff
@@ -0,0 +1,78 @@
...
+  ii_ready_response: string;
+}
+
+const isReadyRequest = (data: unknown): data is ReadyRequest =>
+  typeof data === "object" && data !== null && "ii_ready_request" in data;
+
+const isReadyResponse = (data: unknown): data is ReadyResponse =>
+  typeof data === "object" && data !== null && "ii_ready_response" in data;
+
+export const waitForWindowReadyResponse = (
```
lmuntaner, line 23: Could you add a comment on what this does?

src/frontend/src/utils/multiWebAuthnIdentity.ts
```diff
@@ -79,7 +81,12 @@ export class MultiWebAuthnIdentity extends SignIdentity {
         userVerification: "discouraged",
         rpId: this.rpId,
       },
-    })) as PublicKeyCredential;
+    };
+    const result = (
+      isNullish(this.rpId) || this.iframe !== true
```
lmuntaner, line 86: This is to avoid opening an ifram to the same domain, no?

lmuntaner: I find more readable `this.iframe && nonNullish(this.rpId)`, don't you?

We can do it in another PR, but should we change the parameters of the constructors to be: `webAuthRequest: { rpId: string, iframe } | undefined`?

This way, we can't have `rpId` as `undefined` with an `iframe`.

src/frontend/src/flows/iframeWebAuthn.ts
```diff
@@ -0,0 +1,188 @@
...
+  });
+
+export const webAuthnInIframeFlow = async (
+  connection: Connection
+): Promise<never> => {
+  // Establish cross-origin connection with parent window
+  const config = await connection.getConfig();
+  const targetOrigin = await waitForWindowReadyRequest(
+    window.parent,
+    config.related_origins[0] ?? []
```
lmuntaner, line 139: Should we add a comment why we allow only the related origins?

lmuntaner (approved):
Looking very good! Only a few comments but already approving.

src/frontend/src/utils/multiWebAuthnIdentity.ts
sea-snake: > This way, we can't have rpId as undefined with an iframe.
You can pass `undefined, true` but indeed passing a option/config object with optional properties is cleaner. I will not do this in this PR though, too many tests that will break again 😅.

sea-snake: > this.iframe && nonNullish(this.rpId), don't you?

I agree, I'll make this change by swapping the condition around.

src/frontend/src/flows/iframeWebAuthn.ts
sea-snake: I'll add a comment indeed, I'll mention that it's regarding security concerns.

PR #2826, opened 2025-01-29
DO NOT MERGE OpenID Login / Registration feature development PR

LXIF:
# Motivation

We can already add and remove OIDC accounts. Now we want to also be able to use them to sign into II / create an authenticated connection with II.

# Changes

- built backend functionality to create signed delegations based on JWTs (method on OpenIdCredential, endpoint in openid_api, some structs)
- built frontend functionality so that we can create an AuthenticatedConnection from a JWT using this backend functionality

# Tests

No tests yet. PR is up for preliminary review, tests will be added in time.


LXIF:
@sea-snake If you have some spare time, a preliminary review if I'm on the right path here would be appreciated. In principle we should be able to get an authenticated connection now (have to test it still though).

src/frontend/generated/internet_identity_idl.js
```diff
@@ -327,6 +327,14 @@ export const idlFactory = ({ IDL }) => {
   });
   const JWT = IDL.Text;
   const Salt = IDL.Vec(IDL.Nat8);
+  const OpenIdDelegationResponse = IDL.Record({
+    'anchor_number' : UserNumber,
+    'delegation_response' : GetDelegationResponse,
+  });
+  const OpenIdDelegationError = IDL.Variant({
```
sea-snake, line 334: See additional case in `OpenIdCredentialAddError`:
- `'InternalCanisterError' : IDL.Text` might be useful to cover all other errors as seen in v2 api

src/frontend/src/utils/iiConnection.ts
```diff
@@ -966,6 +969,60 @@ export class AuthenticatedConnection extends Connection {
...
+        signedDelegation.signature
+      ) as unknown as Signature,
+    };
+
+    const chain = DelegationChain.fromDelegations(
+      [transformedDelegation],
+      sessionIdentity.getPublicKey().toDer()
+    );
+
+    const JwtSignedIdentity = DelegationIdentity.fromDelegation(
```
sea-snake, line 1012: It's a variable instance not a class reference)
```suggestion
    const jwtSignedIdentity = DelegationIdentity.fromDelegation(
```

src/frontend/src/utils/iiConnection.ts
```diff
@@ -966,6 +969,60 @@ export class AuthenticatedConnection extends Connection {
...
+      sessionIdentity,
+      chain
+    );
+
+    return new AuthenticatedConnection(
+      this.canisterId,
+      sessionIdentity,
+      JwtSignedIdentity,
```
sea-snake, line 1020:
```suggestion
      jwtSignedIdentity,
```

src/frontend/src/utils/iiConnection.ts
```diff
@@ -966,6 +969,60 @@ export class AuthenticatedConnection extends Connection {
...
+      jwt,
+      salt,
+      new Uint8Array(sessionIdentity.getPublicKey().toDer()),
+      nonNullish(maxTimeToLive) ? [maxTimeToLive] : []
+    );
+    if ("Err" in res) throw new CanisterError(res.Err);
+
+    const { anchor_number, delegation_response } = res.Ok;
+
+    if ("no_such_delegation" in delegation_response) throw "No such delegation";
```
sea-snake, line 991: I suppose this is not within the error itself since you're reusing the DelegationResponse type. You could move this error up in the `OpenIdDelegationError` and use `SignedDelegation` instead directly. But since you'd need to split this into a call and query anyway, this issue should resolve itself already with those changes.

src/internet_identity/src/authz_utils.rs
```diff
@@ -111,6 +111,12 @@ pub fn check_authorization(
             return Ok((anchor.clone(), device.pubkey.clone()));
         }
     }
+    // check openid authorization
```
sea-snake, line 114: Make sure to update the comment at line 95 too, right now it says `and returns a reference to the device used.` but that's no longer true after this change. 

Also check if this device returned (now possibly OpenID) is used somewhere, where it would break if it's no longer a device. First thing that comes to mind would be possibly bookkeeping, storing it as a device login 😅

src/internet_identity/src/delegation.rs
```diff
@@ -18,11 +18,11 @@ use std::net::IpAddr;
 
 // The expiration used for delegations if none is specified
 // (calculated as now() + this)
-const DEFAULT_EXPIRATION_PERIOD_NS: u64 = 30 * MINUTE_NS;
+pub const DEFAULT_EXPIRATION_PERIOD_NS: u64 = 30 * MINUTE_NS;
 
 // The maximum expiration time for delegation
 // (calculated as now() + this)
-const MAX_EXPIRATION_PERIOD_NS: u64 = 30 * DAY_NS;
+pub const MAX_EXPIRATION_PERIOD_NS: u64 = 30 * DAY_NS;
```
sea-snake, line 25: Since JWT delegations are only used internally within the II frontend itself, I would recommend to simply hardcode the expiry to a sane value (30 min?) instead of allowing input (with a max of 30 days).

src/internet_identity/src/main.rs
```diff
@@ -807,6 +812,28 @@ mod openid_api {
                 })
         })
     }
+
+    //TODO: add tests for this
+    #[update]
+    async fn openid_create_delegation(
```
sea-snake, line 818: As discussed this probably needs to be split in prepare and get.

src/internet_identity/src/main.rs
```diff
@@ -807,6 +812,28 @@ mod openid_api {
...
+        jwt: String,
+        salt: [u8; 32],
+        session_key: SessionKey,
+        max_time_to_live: Option<u64>,
+    ) -> Result<OpenIdDelegationResponse, OpenIdDelegationError> {
+        let openid_credential = openid::verify(&jwt, &salt)
+            .map_err(|_| OpenIdDelegationError::JwtVerificationFailed)?;
+
+        match lookup_anchor_with_openid_credential(&openid_credential.clone().into()) {
```
sea-snake, line 827: This match can be rewritten with a `.ok_or(Err(OpenIdDelegationError::NoSuchAnchor))?;` to get the value or throw an error.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -648,6 +653,112 @@ export class Connection {
...
+    const actor = await this.createActor(sessionIdentity);
+    const sessionKey = new Uint8Array(sessionIdentity.getPublicKey().toDer());
+
+    const prepareDelegationResponse = await actor.openid_prepare_delegation(
+      jwt,
+      salt,
+      sessionKey
+    );
+
+    console.log(prepareDelegationResponse);
```
sea-snake, line 720: Just leaving this comment here as a reminder to not forget removing this and other `console.log` :D

src/frontend/src/utils/iiConnection.ts
```diff
@@ -648,6 +653,101 @@ export class Connection {
...
+
+    const signedDelegation = await withLoader(() =>
+      retryGetJwtDelegation(jwt, salt, sessionKey, timestamp, actor)
+    );
+
+    const transformedDelegation = {
+      delegation: new Delegation(
+        new Uint8Array(signedDelegation.delegation.pubkey),
+        signedDelegation.delegation.expiration,
+        signedDelegation.delegation.targets?.[0]?.map((p) =>
```
sea-snake, line 724: We don't assign targets so this could as well be `undefined` if I remember correctly.

Good catch finding this, it went completely past me that `[] !== undefined`, the first means the signature is valid for 0 canisters and the latter means the signature is valid for all canisters.

LXIF: I was grasping at straws tbh, but noticed that the Delegation expects `Principal[] | undefined`, so decided to give it a shot^^

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -237,11 +248,17 @@ export const authenticateBoxFlow = async <I>({
...
+
+  const getGoogleClientId = () =>
+    configRef.current?.openid_google[0]?.[0]?.client_id;
+
   // The registration flow for a new identity
   const doRegister = async (): Promise<
     | (LoginSuccess & {
         newAnchor: true;
-        authnMethod: "pin" | "passkey" | "recovery";
+        authnMethod: "pin" | "passkey" | "recovery" | "openid";
```
sea-snake, line 261: See other comment regarding `authnMethod`.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -274,11 +291,51 @@ export const authenticateBoxFlow = async <I>({
...
+      authnMethod: "openid" as const,
+    };
+  };
+
   // Prompt for an identity number
   const doPrompt = async (): Promise<
     | (LoginSuccess & {
         newAnchor: boolean;
-        authnMethod: "pin" | "passkey" | "recovery";
+        authnMethod: "pin" | "passkey" | "recovery" | "openid";
```
sea-snake, line 338: See other comment regarding `authnMethod`.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -728,7 +808,7 @@ const useIdentityFlow = async <I>({
 }): Promise<
   | (LoginSuccess & {
       newAnchor: boolean;
-      authnMethod: "pin" | "passkey" | "recovery";
+      authnMethod: "pin" | "passkey" | "recovery" | "openid";
```
sea-snake, line 811: See other comment regarding `authnMethod`.

src/frontend/src/flows/authorize/index.ts
```diff
@@ -161,7 +161,7 @@ const authenticate = async (
       kind: "success";
       delegations: Delegation[];
       userPublicKey: Uint8Array;
-      authnMethod: "pin" | "passkey" | "recovery";
+      authnMethod: "pin" | "passkey" | "recovery" | "openid";
```
sea-snake, line 164: See other comment regarding `authnMethod`.

src/frontend/src/flows/authorize/postMessageInterface.ts
```diff
@@ -90,7 +90,7 @@ export async function authenticationProtocol({
         kind: "success";
         delegations: Delegation[];
         userPublicKey: Uint8Array;
-        authnMethod: "pin" | "passkey" | "recovery";
+        authnMethod: "pin" | "passkey" | "recovery" | "openid";
```
sea-snake, line 93: See other comment regarding `authnMethod`.

src/frontend/src/flows/authorize/postMessageInterface.ts
```diff
@@ -71,7 +71,7 @@ export type AuthResponse =
       kind: "authorize-client-success";
       delegations: Delegation[];
       userPublicKey: Uint8Array;
-      authnMethod: "pin" | "passkey" | "recovery";
+      authnMethod: "pin" | "passkey" | "recovery" | "openid";
```
sea-snake, line 74: See other comment regarding `authnMethod`.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -569,6 +648,7 @@ export const authnScreens = (i18n: I18n, props: AuthnTemplates) => {
             resolve({ tag: "add_device", userNumber }),
           recover: (userNumber?: bigint) =>
             resolve({ tag: "recover", userNumber }),
+          loginOpenIDGoogle: () => resolve({ tag: "open_id_google" }),
```
sea-snake, line 651: It's ok for now to have something like `loginOpenIDGoogle`, but later in a future PR this will need to be made more generic e.g. `loginOpenID` with google as an argument.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -662,6 +667,96 @@ export class Connection {
...
+  fromJwt = async (
+    jwt: JWT,
+    salt: Salt,
+    sessionIdentity: SignIdentity
+  ): Promise<AuthenticatedConnection> => {
+    const retryGetJwtDelegation = async (
+      jwt: JWT,
+      salt: Salt,
+      sessionKey: SessionKey,
+      timestamp: bigint,
```
sea-snake, line 680: As mentioned in the other comment, we can omit the timestamp since it's fixed.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -662,6 +667,96 @@ export class Connection {
...
+          setInterval(resolve, 1000 * i);
+        });
+        const res = await actor.openid_get_delegation(
+          jwt,
+          salt,
+          sessionKey,
+          timestamp
+        );
+        if ("Err" in res) {
+          const errorKeys = Object.keys(res.Err);
```
sea-snake, line 696: If you're not going to handle all the error cases here individually with e.g.`new CanisterError(res.Err)`, then just log the error as is to the console. Probably makes more sense anyway since the error toasts on each try will only confuse the user.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -662,6 +667,96 @@ export class Connection {
...
+          const errorMessage =
+            errorKeys.length > 0 ? errorKeys[0] : "unknown error";
+          toast.error("Error while fetching delegation: " + errorMessage);
+          continue;
+        }
+
+        return res.Ok;
+      }
+      throw new Error(
```
sea-snake, line 705: I assume this is automatically caught and shown to the user on the error screen?

src/frontend/src/utils/iiConnection.ts
sea-snake: Maybe we could turn this in to a utility function `const transformSignedDelegation = (signedDelegation: SignedDelegation): Delegation => ...`?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -823,9 +918,6 @@ export class AuthenticatedConnection extends Connection {
   ): Promise<[PublicKey, bigint] | { error: unknown }> => {
     try {
       const origin = remapToLegacyDomain(origin_);
-      console.log(
-        `prepare_delegation(user: ${this.userNumber}, origin: ${origin}, session_key: ${sessionKey})`
-      );
```
sea-snake, line 828: Let's keep this and below console log for now, they've always been there, no need to remove them now.

src/internet_identity/internet_identity.did
```diff
@@ -13,32 +13,32 @@ type HeaderField = record {
 };
 
 type HttpRequest = record {
-    method: text;
-    url: text;
-    headers: vec HeaderField;
-    body: blob;
-    certificate_version: opt nat16;
+    method : text;
```
sea-snake, line 16: Can we avoid updating the whole did file whitespacing in this PR?

src/internet_identity/src/authz_utils.rs
sea-snake: The wrapping `authorization_key` approach seems to cover this now :D

src/internet_identity/src/authz_utils.rs
```diff
@@ -108,13 +133,26 @@ pub fn check_authorization(
...
-            return Ok((anchor.clone(), device.pubkey.clone()));
+            return Ok((
+                anchor.clone(),
+                AuthorizationKey::DevicePubKey(device.pubkey.clone()),
+            ));
+        }
+    }
+    // check openid authorization
+    for credential in anchor.openid_credentials() {
+        //TODO: handle temp keys
```
sea-snake, line 144: We should discuss probably how to handle temp keys (register flow) if we should add this to open id credentials or find a more generic approach. Leaving this as TODO looks good to me.

src/internet_identity/src/openid.rs
```diff
@@ -30,10 +46,65 @@ impl OpenIdCredential {
...
+
+    //TODO: maybe this can be improved
+    pub async fn prepare_jwt_delegation(
+        &self,
+        session_key: SessionKey,
+        frontend: FrontendHostname, //TODO: maybe add IIDomain
+    ) -> (UserKey, Timestamp) {
+        state::ensure_salt_set().await;
+
+        let session_duration_ns = 30 * MINUTE_NS;
```
sea-snake, line 65: This is a constant, so it probably should be moved out of this function to the top of the file with ideally a comment to explain details where needed.

src/internet_identity/src/openid.rs
```diff
@@ -30,10 +46,65 @@ impl OpenIdCredential {
...
+        (
+            ByteBuf::from(der_encode_canister_sig_key(seed.to_vec())),
+            expiration,
+        )
+    }
+
+    pub fn get_jwt_delegation(
+        &self,
+        session_key: SessionKey,
+        expiration: Timestamp,
```
sea-snake, line 86: A constant was used to create the delegation, so this can be omitted.

sea-snake (changes requested):

src/frontend/generated/internet_identity_idl.js
```diff
@@ -542,6 +552,26 @@ export const idlFactory = ({ IDL }) => {
         [IDL.Variant({ 'Ok' : IDL.Null, 'Err' : OpenIdCredentialRemoveError })],
         [],
       ),
+    'openid_get_delegation' : IDL.Func(
+        [JWT, Salt, SessionKey, Timestamp],
```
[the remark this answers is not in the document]
LXIF: Admittedly I should rename it to expiration though.

src/frontend/generated/internet_identity_idl.js
```diff
@@ -339,7 +339,17 @@ export const idlFactory = ({ IDL }) => {
...
+  const OpenIdDelegationError = IDL.Variant({
+    'NoSuchDelegation' : IDL.Null,
+    'NoSuchAnchor' : IDL.Null,
+    'JwtVerificationFailed' : IDL.Null,
+  });
   const UserKey = PublicKey;
+  const OpenIdPrepareDelegationResponse = IDL.Record({
+    'user_key' : UserKey,
+    'anchor_number' : UserNumber,
+    'timestamp' : Timestamp,
```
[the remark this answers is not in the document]
LXIF: See above reasoning

src/frontend/generated/internet_identity_idl.js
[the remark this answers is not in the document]
sea-snake: You're right, didn't notice that the value is `time() + fixed value`, `openid_get_delegation` can't know this value without it being told.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -274,11 +291,51 @@ export const authenticateBoxFlow = async <I>({
       allowPinLogin,
     });
 
+  const doLoginWithGoogle = async (connection: Connection) => {
+    const googleClientId = getGoogleClientId();
+
+    if (isNullish(googleClientId)) {
+      toast.error("Google login is unavailable."); // TODO: put into copy / handle errors better
```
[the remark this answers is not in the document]
LXIF: done

src/frontend/src/utils/iiConnection.ts
LXIF: Screenshot 2025-02-10 at 16 07 30

Correct, you get to this screen.

LXIF: Could be prettier though. Refactoring into a toast.

LXIF: (This is the pattern used by the iiConnection in general, though admittedly it would be nice if there was a unified way of course)

src/internet_identity/internet_identity.did
LXIF: my code editor does it automatically, as this is the official candid syntax^^

LXIF: ..but removed it now

src/internet_identity/src/openid.rs
LXIF: done

src/frontend/src/utils/iiConnection.ts
LXIF: done

src/frontend/src/utils/iiConnection.ts
LXIF: done

src/frontend/src/utils/iiConnection.ts
LXIF: done

src/internet_identity/src/openid.rs
LXIF: It cannot be omitted as per above discussion, the expiration is required to retrieve the prepared delegation.

lmuntaner:
@LXIF can we close this PR?

PR #2827, opened 2025-01-30
Use RP ID in WebAuthIdentity sign

lmuntaner:
# Motivation

There was an issue with a user that logged in using a RP ID and then stayed idle so that a new signature from the passkey was required.

The issue was that the new request didn't use the initial RP ID.

# Changes

* Add `rpId` to the constructor of `WebAuthnIdentity`.
* Pass it where the constructor is used.

# Tests

I tested in beta domains by requesting a new signature on every call instead of using the delegation.

### DIFF
--- a/src/frontend/src/utils/webAuthnIdentity.ts
+++ b/src/frontend/src/utils/webAuthnIdentity.ts
@@ -153,13 +153,17 @@ export class WebAuthnIdentity extends SignIdentity {
    * @param json - json to parse
    */
   public static fromJSON(json: string): WebAuthnIdentity {
-    const { publicKey, rawId } = JSON.parse(json);
+    const { publicKey, rawId, rpId } = JSON.parse(json);
 
-    if (typeof publicKey !== "string" || typeof rawId !== "string") {
+    if (
+      typeof publicKey !== "string" ||
+      typeof rawId !== "string" ||
+      (typeof rpId !== "string" && rpId !== undefined)
+    ) {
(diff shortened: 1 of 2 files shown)


lmuntaner:
@sea-snake please review

src/frontend/src/utils/webAuthnIdentity.ts
```diff
@@ -153,13 +153,17 @@ export class WebAuthnIdentity extends SignIdentity {
...
   public static fromJSON(json: string): WebAuthnIdentity {
-    const { publicKey, rawId } = JSON.parse(json);
+    const { publicKey, rawId, rpId } = JSON.parse(json);
 
-    if (typeof publicKey !== "string" || typeof rawId !== "string") {
+    if (
+      typeof publicKey !== "string" ||
+      typeof rawId !== "string" ||
+      (typeof rpId !== "string" && rpId !== undefined)
+    ) {
```
sea-snake, line 162: Wouldn't this throw an error for stored passkeys without rp id? Which I assume we have?

lmuntaner: No, if rpId is `undefined`, then `rpID !== undefined` is `false`. Therefore `(typeof rpId !== "string" && rpId !== undefined)` is also `false`. A `rpId` of type `number` would throw an error.

lmuntaner:

sea-snake (approved):

PR #2834, opened 2025-02-05
Use iframe workaround when needed

lmuntaner:
# Motivation

Allow users to log in even if the first selected RP ID was the incorrect one.

In this PR, we use the new helper `findWebAuthnSteps` with the new parameters `rpId` and `iframe` of `MultiWebAuthnIdentity`.

In another PR: Change info message on first error.
In another PR: Rename `webAuthnSteps` to `webAuthnFlows`.

# Changes

* New instance variable `webAuthFlows` in iiConnection to keep track of the steps and current step index to try.
* If the new instance is not set, get the flows and set it.
* Use the RP ID and iframe in the current flow.
* Reset the flow index when out of bounds.
* Increase the step when needed.

# Tests

* Reintroduced a test that was removed before `"connection excludes rpId when user cancels"`.
* Remove duplicated test.
* Change expected frame parameter value in current tests.

### DIFF
--- a/src/frontend/src/utils/iiConnection.test.ts
+++ b/src/frontend/src/utils/iiConnection.test.ts
@@ -190,7 +190,59 @@ describe("Connection.login", () => {
         expect(MultiWebAuthnIdentity.fromCredentials).toHaveBeenCalledWith(
           [convertToValidCredentialData(mockDevice)],
           "identity.ic0.app",
-          undefined
+          true
+        );
+      }
+    });
+
+    it("connection excludes rpId when user cancels", async () => {
+      // This one would fail because it's not the device the user is using at the moment.
+      const currentOriginDevice: DeviceData = createMockDevice(currentOrigin);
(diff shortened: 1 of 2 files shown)


src/frontend/src/utils/iiConnection.test.ts
```diff
@@ -348,36 +400,7 @@ describe("Connection.login", () => {
             currentOriginCredentialData,
           ]),
           undefined,
-          undefined
-        );
-      }
-    });
-  });
-
-  describe("domains compatibility flag enabled and browser doesn't support", () => {
```
lmuntaner, line 357: This was duplicated

lmuntaner:
UX with a user that needs two tries:


https://github.com/user-attachments/assets/b89c6a2b-8261-41d9-a456-184d806b7e54

lmuntaner:
@sea-snake ready for review

sea-snake (approved):

PR #2835, opened 2025-02-05
Rename PossiblyWrongRPID to PossiblyWrongWebAuthnFlow

lmuntaner:
# Motivation

PossiblyWrongRPID is confusing now that we moved to the concept of WebAuthn flows.

# Changes

* Change `PossiblyWrongRPID` and i18n keys.

# Tests

No new functionality was added.

<hr/><details><summary>🟡 Some screens were changed</summary></details><details><summary>🔴 Some screens were removed</summary></details>

### DIFF
--- a/src/frontend/src/components/authenticateBox/index.ts
+++ b/src/frontend/src/components/authenticateBox/index.ts
@@ -379,10 +379,7 @@ export const handleLoginFlowResult = async <E>(
     toast.info(
       infoToastTemplate({
         title: copy.title_possibly_wrong_web_authn_flow,
-        messages: [
-          copy.message_possibly_wrong_web_authn_flow_1,
-          copy.message_possibly_wrong_web_authn_flow_2,
-        ],
+        messages: [copy.message_possibly_wrong_web_authn_flow_1],
       })
     );
     return undefined;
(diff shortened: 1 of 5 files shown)


lmuntaner:
@sea-snake please review

In the end, I didn't change the message. I think that the current is good enough. The only thing I would do is maybe to remove the second sentence: "Try again and another domain will be used." Just to make the toast smaller and have a higher chance that users will read it.

sea-snake:
@lmuntaner I agree that making the sentence shorter will help, but more importantly, the sentence you've mentioned starts with "Try again" which should be indeed at the start of the sentence, since this is the key takeaway and action for the user from the message.

lmuntaner:
> @lmuntaner I agree that making the sentence shorter will help, but more importantly, the sentence you've mentioned starts with "Try again" which should be indeed at the start of the sentence, since this is the key takeaway and action for the user from the message.

The title is already "Please try again". That's why I don't think it necessary to repeat it below.

sea-snake:
> The title is already "Please try again".

Ah in that case, it's indeed not necessary to repeat it again.

lmuntaner:
I removed the second sentence for simplicity.

PR #2836, opened 2025-02-05
Rename WebAuthnSteps to WebAuthnFlows

lmuntaner:
# Motivation

WebAuthnSteps is confusing because those are not a set of steps that the user needs to perform.

WebAuthnFlows is a better (but not perfect) name to understand that the user will perform one of those.

# Changes

* Rename file, util and type.

# Tests

* Rename test files and describe.

### DIFF
--- a/src/frontend/src/utils/findWebAuthnFlows.test.ts
+++ b/src/frontend/src/utils/findWebAuthnFlows.test.ts
@@ -1,9 +1,9 @@
 import { LEGACY_II_URL } from "$src/config";
 import { CredentialData } from "./credential-devices";
+import { findWebAuthnFlows } from "./findWebAuthnFlows";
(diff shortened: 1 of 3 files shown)


lmuntaner:
@sea-snake please review

sea-snake:
I agree flow is better than steps but still not entirely clear, what about `WebAuthnAttempts`?

sea-snake (approved):

lmuntaner:
> I agree flow is better than steps but still not entirely clear, what about `WebAuthnAttempts`?

I'm not convinced with `Attempt` because it feel weird that an attempt has properties. It sounds better the RP ID of the flow than the RP ID from the attempt.

sea-snake:
> it feel weird that an attempt has properties

I agree

PR #2838, opened 2025-02-05
Improve post message interface

sea-snake:
Improve post message interface, by removing and adding error screens:

- Removed: invalid data
- Added: invalid request
- Added: connection closed
- Added: wrong place

Besides the improved error messaging around orphan pages, it now also separates those from closed connections and connections that timed out (didn't receive a request in 10 seconds). This last point should avoid user experiences with spinners that never end.

Additionally it now shows an error when an invalid request has been received, previously this was ignored while waiting for a valid request, but since the protocol doesn't notify the relying party that the request is invalid, this will never happen in practice.

# Tests

- Removed: Should ignore invalid data and allow second valid message
- Added: Should show error for invalid message
- Added: Should show error after not receiving message for 10 seconds
- Added: Should show error after manually navigating to authorize url

### DIFF
--- a/src/canister_tests/src/framework.rs
+++ b/src/canister_tests/src/framework.rs
@@ -366,14 +366,16 @@ pub fn expect_user_error_with_message<T: std::fmt::Debug>(
     }
 }
 
-pub fn verify_security_headers(headers: &[HeaderField]) {
-    let expected_headers = vec![
-        ("X-Frame-Options", "DENY"),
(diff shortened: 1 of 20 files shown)


sea-snake:
Some screenshots of new error screens:

image
image
image

lmuntaner (approved):
Thanks! This is a great improvement!

PR #2840, opened 2025-02-06
Use devices origin if ROR is supported

lmuntaner:
# Motivation

ROR is not supported in all browsers. We can't use it when adding new devices on those browsers.

Not in this PR: Ignore it also when third-party password managers are installed.

# Changes

* Check ROR support when calculating the origin of a new device.

# Tests

* Tested in beta domains.


lmuntaner:
@sea-snake please review

sea-snake (approved):

PR #2842, opened 2025-02-07
Check for Password Manager extensions when adding new devices

lmuntaner:
# Motivation

Third-party password managers don't support yet ROR. Therefore, when we add a new device, we want to skip a RP ID if the user uses one.

# Changes

* New util `userSupportsWebauthRoR` that checks the browser version and tries to find extensions installed.
* Use new util `userSupportsWebauthRoR` instead of the user agent util `supportsWebauthRoR` when adding new devices.

# Tests

* Test the new util.
* Tested in beta.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -50,7 +50,7 @@ import {
   isRecoveryDevice,
   isRecoveryPhrase,
 } from "$src/utils/recoveryDevice";
-import { supportsWebauthRoR } from "$src/utils/userAgent";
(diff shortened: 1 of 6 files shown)


lmuntaner:
UX when adding from a user with the extension. See how after the device is added, the list shows the domain of each device.

https://github.com/user-attachments/assets/e8dc8aea-8357-4fbe-8626-661ba359e791

UX when adding from a user without the extension. See how after the devices is added, the list doesn't show the domains of each device. That's because they are all registered in the same domain.

https://github.com/user-attachments/assets/5a8c0c43-3808-409c-8775-289b49153582

lmuntaner:
@sea-snake please review

PR #2843, opened 2025-02-07
Fix forward error from iframe

lmuntaner:
# Motivation

The iframe workaround didn't have the same interface when throwing error and those were not properly caught.

# Changes

* Change the error return type in post message and use it to create a DOMException error.

# Tests

* Tested in beta domains.


lmuntaner:
@sea-snake please review

sea-snake (approved):

PR #2851, opened 2025-02-11
Add analytics config

lmuntaner:
# Motivation

Add analytics config to init parameters to make it possible to enable or disable analytics with a config change.

# Changes

* New `AnalyticsConfig` enum.
* Add `analytics_config` field to `InternetIdentityInit` and `PeristentState`.
* Set the persistent state field on init.

# Tests

* Extend a test to check that the new field is set.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -342,7 +342,12 @@ type OpenIdConfig = record {
 type OpenIdCredentialKey = record { Iss; Sub };
 
 type AnalyticsConfig = variant {
-    Plausible;
+    Plausible {
+        domain: opt text,
+        hash_mode: opt bool,
+        track_localhost: opt bool,
+        api_host: opt text,
+    };
 };
 
 type OpenIdCredential = record {
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -342,7 +342,7 @@ type OpenIdConfig = record {
 type OpenIdCredentialKey = record { Iss; Sub };
 
 type AnalyticsConfig = variant {
-    Plausible {
+    Plausible: record {
         domain: opt text,
         hash_mode: opt bool,
         track_localhost: opt bool,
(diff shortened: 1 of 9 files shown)


lmuntaner:

src/internet_identity/internet_identity.did
```diff
@@ -339,6 +341,10 @@ type OpenIdConfig = record {
 
 type OpenIdCredentialKey = record { Iss; Sub };
 
+type AnalyticsConfig = variant {
+    Plausible;
```
sea-snake, line 345: Should this variant maybe contain a Plausible config used by the Plausible lib in the frontend?

lmuntaner: Yes, but there is no config. Only if we wanted to have all in the same domain.

https://www.npmjs.com/package/plausible-tracker

According to chatGPT I can allow my site to allow multiple domains. This would mean that I see the aggregated data of all domains while keeping each individual as well.

But I can't find this in Plausible documentation now. I'll do some tests once I have the Plausible account.

lmuntaner: I added all the possible configuration to the config type.

I also did some tests with Plausible and I suggest we set one domain in the config. Then we can filter by hostname in the dashboard.

src/internet_identity/tests/integration/config.rs
```diff
@@ -77,6 +78,12 @@ fn should_retain_config_after_none() -> Result<(), CallError> {
         }),
         related_origins: Some(related_origins),
         openid_google: Some(Some(openid_google)),
+        analytics_config: Some(Some(AnalyticsConfig::Plausible {
```
LXIF, line 81: Why are we using Some(Some()) here? (And then opt opt in the .did)?

LXIF (approved):
LGTM in general, assuming there is a good reason to double-wrap the config.

src/internet_identity/tests/integration/config.rs
lmuntaner: Good question. The first Option means that the field is optional. The second means enabling or disabling the feature.

We though of renaming the second one: Option<Enabled<XXXX>>

sea-snake (approved):

PR #2852, opened 2025-02-11
OpenID Connect JWT Delegations

LXIF:
# Motivation

In order to implement OpenID Connect Login and Registration, we need to be able to create delegations based on OpenID JWTs. This PR implements the necessary backend functionality.

# Changes

Add functions to allow preparing and retrieving OpenID JWT based Delegations, as well as generalizing related functionality (e.g. DeviceKey is now AuthorizationKey with variants DevicePubKey and OpenIdPubKey).

# Tests

Integration tests are included in this PR, in addition to integration tests for adding and removing Google accounts (this is necessary to test the delegations).


src/internet_identity/tests/integration/openid.rs
```diff
@@ -0,0 +1,99 @@
...
+    // Generate random salt
+    let salt = salt();
+
+    // Create bytes array containing salt + principal
+    let mut bytes = Vec::with_capacity(32 + principal.as_slice().len());
+    bytes.extend_from_slice(&salt);
+    bytes.extend_from_slice(principal.as_slice());
+
+    // Calculate SHA-256 hash
+    let hash = ring::digest::digest(&ring::digest::SHA256, &bytes);
```
sea-snake, line 93: No need to add another Rust package for this, we already have a different package that was used here: https://github.com/dfinity/internet-identity/pull/2852/files#diff-d8bff67d07c9665c18043a61031542c1acb5b40e7289a117e764e321355a8cb6R166

src/internet_identity/src/main.rs
```diff
@@ -811,6 +815,43 @@ mod openid_api {
...
+        salt: [u8; 32],
+        session_key: SessionKey,
+    ) -> Result<OpenIdPrepareDelegationResponse, OpenIdDelegationError> {
+        let openid_credential = openid::verify(&jwt, &salt)
+            .map_err(|_| OpenIdDelegationError::JwtVerificationFailed)?;
+
+        let anchor_number = lookup_anchor_with_openid_credential(&openid_credential.clone().into())
+            .ok_or(OpenIdDelegationError::NoSuchAnchor)?;
+
+        let (user_key, expiration) = openid_credential.prepare_jwt_delegation(session_key).await;
```
jasonz-dfinity, line 831: It's probably OK, but just wanted to call out the fact that the `anchor_number` may no longer be associated with the credential (e.g. credential removed) after the `await`, if the `raw_rand` is indeed called during `ensure_salt_set`. Just making sure this edge case is considered.

src/canister_tests/src/framework.rs
```diff
@@ -166,6 +166,19 @@ pub fn install_ii_canister_with_arg(
     canister_id
 }
 
+pub fn install_ii_canister_with_arg_and_cycles(
+    env: &PocketIc,
+    wasm: Vec<u8>,
+    arg: Option<InternetIdentityInit>,
+    amount: u128,
+) -> CanisterId {
+    let byts = candid::encode_one(arg).expect("error encoding II installation arg as candid");
```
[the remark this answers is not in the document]
LXIF: Fixed ✅

src/internet_identity/src/authz_utils.rs
```diff
@@ -73,30 +75,53 @@ pub fn anchor_operation_with_authz_check<R, E>(
 where
     E: From<IdentityUpdateError>,
 {
-    let (mut anchor, device_key) = check_authorization(anchor_number)
+    let (mut anchor, authorization_key) = check_authorization(anchor_number)
         .map_err(|err| E::from(IdentityUpdateError::from(err)))?;
-    anchor_management::activity_bookkeeping(&mut anchor, &device_key);
 
-    let result = op(&mut anchor);
+    match authorization_key {
```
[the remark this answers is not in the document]
LXIF: Makes perfect sense, I wrote that part initially not knowing exactly what the difference would be and neglected to refactor it properly. Thanks!

src/internet_identity/src/openid.rs
```diff
@@ -1,21 +1,37 @@
...
 pub type Iss = String;
 pub type Sub = String;
 pub type Aud = String;
 
+impl From<OpenIdCredential> for OpenIdCredentialKey {
+    fn from(credential: OpenIdCredential) -> Self {
+        (credential.iss, credential.sub)
+    }
+}
+
```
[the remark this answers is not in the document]
LXIF: Yes, thanks! Was too quick and eager to write impls and didn't see it.

src/internet_identity/src/openid/google.rs
```diff
@@ -304,7 +304,7 @@ fn verify_claims(client_id: &String, claims: &Claims, salt: &[u8; 32]) -> Result
         return Err(format!("Invalid audience: {}", claims.aud));
     }
     if claims.nonce != expected_nonce {
-        return Err(format!("Invalid nonce: {}", claims.nonce));
+        return Err(format!("Invalid nonce: {}", claims.nonce,));
```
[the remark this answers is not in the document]
LXIF: Nope, had more logs there and cleanup was sloppy. Thanks!

src/internet_identity/tests/integration/openid.rs
```diff
@@ -0,0 +1,317 @@
...
+        archive_config: None,
+        canister_creation_cycles_cost: None,
+        register_rate_limit: None,
+        captcha_config: None,
+        related_origins: None,
+        openid_google: Some(Some(OpenIdConfig {
+            client_id: CLIENT_ID.to_string(),
+        })),
+        analytics_config: None,
+    };
```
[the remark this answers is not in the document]
LXIF: Ah, was not aware of this syntax in Rust yet! Elegant =)

src/internet_identity_interface/src/internet_identity/types/openid.rs
```diff
@@ -27,3 +28,22 @@ pub enum OpenIdCredentialRemoveError {
...
+pub enum OpenIdDelegationError {
+    NoSuchAnchor,
+    NoSuchDelegation,
+    JwtVerificationFailed,
+}
+
+pub type OpenIdCredentialKey = (Iss, Sub);
+pub type Iss = String;
+pub type Sub = String;
+pub type Aud = String;
```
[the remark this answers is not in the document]
LXIF: Yes, they are not part of the same project (one is `internet_identity` and the other `internet_identity_interface`), and I can't import from `internet_identity`. If I were to want to import from there, I would have to add `pub mod openid` to the lib.rs, but then I need to change pretty much every other import as well (wasn't aware of this before but compiler says no). So, in order to limit the amount of changes I'd need to make, I chose to duplicate it here.

src/canister_tests/src/framework.rs
```diff
@@ -166,6 +166,19 @@ pub fn install_ii_canister_with_arg(
     canister_id
 }
 
+pub fn install_ii_canister_with_arg_and_cycles(
```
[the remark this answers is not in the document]
LXIF: Not sure what you mean - I added this function because I need to install and add cycles before the canister is installed, as the openID setup immediately starts making outcalls. Do you mean using this one instead of `install_ii_canister_with_arg`? That's correct, that one was not added by me though and I would have to go change a lot of old integration tests. Would that be more standard practice?

src/internet_identity/src/main.rs
LXIF: Would you recommend checking the association again after the await?

LXIF: Added this. Do you know off the top of your head how I would run an integration test for this?

src/canister_tests/src/framework.rs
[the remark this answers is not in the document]
lmuntaner: Make a PR to change the old endpoint along with all the other tests. This could be done either after this PR or after. If after, add a TODO and let's make sure we do it right afterwards.

src/frontend/src/flows/manage/index.ts
```diff
@@ -453,6 +453,7 @@ export const displayManage = async (
             nonce,
           })
         );
+
```
lmuntaner, line 456: I would have reverted this line to make the PR cleaner.

src/frontend/src/utils/iiConnection.ts
```diff
@@ -498,7 +498,6 @@ export class Connection {
     }
 
     const actor = await this.createActor(delegationIdentity);
-
```
lmuntaner, line 501: ditto I would have reverted this line to make the PR cleaner.

src/internet_identity/Cargo.toml
```diff
@@ -53,6 +53,7 @@ hex-literal = "0.4"
 regex.workspace = true
 ic-response-verification.workspace = true
 flate2 = "1.0"
+getrandom = "0.2"
```
lmuntaner, line 56: Is this necessary? where is it used?

src/internet_identity/internet_identity.did
```diff
@@ -758,6 +770,8 @@ service : (opt InternetIdentityInit) -> {
     // =========================
     openid_credential_add : (IdentityNumber, JWT, Salt) -> (variant { Ok; Err : OpenIdCredentialAddError });
     openid_credential_remove : (IdentityNumber, OpenIdCredentialKey) -> (variant { Ok; Err : OpenIdCredentialRemoveError });
+    openid_prepare_delegation : (JWT, Salt, SessionKey) -> (variant { Ok : OpenIdPrepareDelegationResponse; Err : OpenIdDelegationError });
```
lmuntaner, line 773: We should also change the ii_spec file to include the new endpoints. But not in this PR, let's do it when we enable the flat for production.

src/internet_identity/src/openid.rs
```diff
@@ -30,10 +40,58 @@ impl OpenIdCredential {
...
-        let public_key = der_encode_canister_sig_key(seed.to_vec());
+        let public_key = self.public_key();
         Principal::self_authenticating(public_key)
     }
+    pub fn public_key(&self) -> PublicKey {
+        let seed = calculate_delegation_seed(&self.aud, &self.key());
+        der_encode_canister_sig_key(seed.to_vec()).into()
+    }
+
+    pub async fn prepare_jwt_delegation(&self, session_key: SessionKey) -> (UserKey, Timestamp) {
```
lmuntaner, line 51: Why don't we have a max_session_to_live parameter like in the normal delegation?

Will we add it later?

src/internet_identity/src/openid/google.rs
```diff
@@ -64,7 +64,7 @@ struct Certs {
     keys: Vec<Jwk>,
 }
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Debug)]
```
lmuntaner, line 67: Is this necessary? I'm ok leaving it, just curiosity

src/internet_identity/tests/integration/openid.rs
```diff
@@ -0,0 +1,311 @@
...
+        canister_id,
+        test_principal,
+        identity_number,
+        &claims.key(),
+    )?
+    .map_err(|e| CallError::Reject(format!("{:?}", e)))?;
+
+    // Try to get delegation based on credential, should fail now
+    // Create session key
+    let pub_session_key = ByteBuf::from("session public key");
```
lmuntaner, line 74: I would also use an endpoint that returns the access methods of the user.

If not part of this PR, it can be done in another one. We'll need that data anyway.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -0,0 +1,311 @@
...
+        Err(err) => match err {
+            OpenIdDelegationError::NoSuchAnchor => Ok(()),
+            _ => panic!("We should get a NoSuchAnchor error here!"),
+        },
+    }
+}
+
+/// Verifies that valid JWT delegations are issued based on added .
+#[test]
+fn can_get_valid_jwt_delegation() -> Result<(), CallError> {
```
lmuntaner, line 95: 🚀

src/internet_identity/src/openid.rs
sea-snake: As discussed previously (in another PR), this endpoint is used internally only, there is no dapp input regarding expiry. 

The existing delegation endpoint has the expiry input since a dapp is able to define the expiry. This endpoints would still be used even with google sign in: 
```
dapp sign in -> II sign in with google -> prepare/get_openid_delegation -> prepare/get_delegation -> dapp
```

lmuntaner: That makes sense, thanks for the explanation and sorry for missing the other discussion.

src/canister_tests/src/framework.rs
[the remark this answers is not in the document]
jasonz-dfinity: I guess @lmuntaner is thinking of replacing calls to `install_ii_canister_with_arg` with `install_ii_canister_with_arg_and_cycles` where `cycles = 0`. That works too. I was thinking `install_ii_canister_with_arg` could just call `install_ii_canister_with_arg_and_cycles`, similar to `install_ii_canister` calls `install_ii_canister_with_arg` where `args = None`. Either works, I just wanted to point out there is some duplicated code between `install_ii_canister_with_arg` and `install_ii_canister_with_arg_and_cycles`.

src/internet_identity/src/main.rs
jasonz-dfinity: Unfortunately, I'm not aware of a good way of testing interleaving messages reliably...

I think the current implementation is OK (perhaps adding some comment on why rechecking). It's a trade-off between more complexity (checking again certainly is more complex) and a potential TOCTOU issue.

Again, I wasn't saying you necessarily had a problem (without recheck), and just wanted to make sure the case is considered. For example, if the delegation is still valid after calling `openid_prepare_delegation` and then immediately removing the association afterwards (without interleaving), then it sounds like there wasn't a problem if you don't check again.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -0,0 +1,317 @@
...
+    email: Option<String>,
+    name: Option<String>,
+    picture: Option<String>,
+}
+
+impl From<Claims> for OpenIdCredentialKey {
+    fn from(value: Claims) -> Self {
+        (value.iss, value.sub)
+    }
+}
```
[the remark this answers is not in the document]
jasonz-dfinity: Exactly. The additional clones aren't really a problem in test code, but it's just a pattern that's often suboptimal.

jasonz-dfinity (approved):

src/internet_identity/src/main.rs
LXIF: So, the `openid_prepare_delegation` call does not actually return the delegation yet - the frontend needs to call `openid_get_delegation` afterwards to actually retrieve the delegation. This is because the delegation is served as a certified asset signed by the canister, so it needs to be prepared (and the root hash updated) in an update call, and then retrieved in a subsequent query call.

In the `openid_get_delegation` call, the association between the credential and the anchor is checked again. We don't currently check explicitly that it's the same anchor number, as the function fails out if it doesn't have the same salt, jwt and session key as the first call.

I'm admittedly not an expert on TOCTOU attacks, but I have created a new integration test to check this. Turns out that if the credential was removed and added to another account we can still get the delegation, even though it is now associated with a different anchor. This is certainly not desirable behavior. Will have to remediate that.

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -21,7 +21,7 @@ pub mod vc_mvp;
 
 // re-export v2 types without the ::v2 prefix, so that this crate can be restructured once v1 is removed
 // without breaking clients
-use crate::internet_identity::types::openid::OpenIdCredentialData;
+pub use crate::internet_identity::types::openid::*;
```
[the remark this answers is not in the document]
LXIF: Because we need a few things from the openid types in the integration test helper functions (under canister tests > src > api > internet_identity.rs)

src/internet_identity/tests/integration/openid.rs
LXIF: 🤙🏻

src/internet_identity_interface/src/internet_identity/types.rs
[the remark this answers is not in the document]
lmuntaner: Ok!

src/internet_identity/tests/integration/openid.rs
LXIF: Added!

src/internet_identity/tests/integration/openid.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    api::openid_credential_add(
+        &env,
+        canister_id,
+        test_principal,
+        identity_number,
+        &jwt,
+        &salt,
+    )?
+    .map_err(|e| CallError::Reject(format!("{:?}", e)))
```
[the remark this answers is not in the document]
LXIF: Added this as well!

src/internet_identity/src/openid/google.rs
LXIF: It was needed for debugging the integration test - removing it

src/internet_identity/src/authz_utils.rs
```diff
@@ -108,13 +119,26 @@ pub fn check_authorization(
...
-            return Ok((anchor.clone(), device.pubkey.clone()));
+            return Ok((
+                anchor.clone(),
+                AuthorizationKey::DevicePubKey(device.pubkey.clone()),
+            ));
+        }
+    }
+    // check openid authorization
+    for credential in anchor.openid_credentials() {
+        //TODO: handle temp keys
```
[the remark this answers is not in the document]
LXIF: Correct, to the best of my foresight we will require them there :)

src/internet_identity/src/main.rs
LXIF: (Related integration test is in another PR now)

src/frontend/src/utils/iiConnection.ts
LXIF: done

src/frontend/src/flows/manage/index.ts
LXIF: done

PR #2853, opened 2025-02-11
Implement canister config in html.

sea-snake:
Implement canister config in html.

# Changes

Changes were made to match the same approach as `data-canister-id` was added previously.

- Refactor `initialize` method to first `apply_install_arg` and then re-use `config` method to get the merged config.
- Pass this merged config to `init_assets` so that `fixup_html` can add the config as `data-canister-config` attribute.
- Implement `readCanisterConfig` method to decode this attribute on page load to `InternetIdentityInit`
- Update `Connection` to remove `getConfig` and have a `canisterConfig` attribute instead.
- Replace usages of `getConfig` with `canisterConfig`.

### DIFF
--- a/src/frontend/src/utils/openID.ts
+++ b/src/frontend/src/utils/openID.ts
@@ -1,5 +1,6 @@
 import { MetadataMapV2 } from "$generated/internet_identity_types";
 import { REDIRECT_CALLBACK_PATH, redirectInPopup } from "$src/flows/redirect";
+import { toBase64URL } from "$src/utils/utils";
 import { Principal } from "@dfinity/principal";
 import { isNullish, nonNullish } from "@dfinity/utils";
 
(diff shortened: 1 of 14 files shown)


src/frontend/src/utils/openID.ts
```diff
@@ -76,12 +77,6 @@ export const isNotSupportedError = (error: unknown) =>
 export const isPermissionError = (error: unknown) =>
   error instanceof Error && error.name === "NetworkError";
 
-const toBase64 = (bytes: ArrayBuffer): string =>
```
lmuntaner, line 79: Why these changes?

src/internet_identity/src/assets.rs
```diff
@@ -33,25 +35,28 @@ pub fn init_assets(maybe_related_origins: Option<Vec<String>>) {
...
         );
     });
 }
 
-// Fix up HTML pages, by injecting canister ID
-fn fixup_html(html: &str) -> String {
+// Fix up HTML pages, by injecting canister ID and canister config
+fn fixup_html(html: &str, config: &InternetIdentityInit) -> String {
     let canister_id = api::id();
+    let encoded_config = BASE64.encode(Encode!(&config).unwrap());
```
lmuntaner, line 46: Why base64 encoded?

PR #2863, opened 2025-02-13
Fix related origins edge case handling

lmuntaner:
# Motivation

There was an infinite loop in `findWebAuthnFlows` if there were no related origins and that's when I realized that the way to handle edge cases was probably not the right one.

I'd rather continue with the default flow if there is some inconsistency in the state rather than failing. It could be that there was a bug when origin was stored and a weird origin is there. Or to support Utopia the state might also be different than this.

# Changes

* Change the handling of edge cases in `findWebAuthnFlows` and `findWebAuthnRpId` and avoid throwing errors.
* Add a way out of an infinite loop by checking that step to be added is already included.

# Tests

* Add tests for missing use cases like no relying parties or "wrong" relying parties.

### DIFF
--- a/src/frontend/src/test-e2e/views.ts
+++ b/src/frontend/src/test-e2e/views.ts
@@ -239,8 +239,7 @@ export class MainView extends View {
   async waitForDisplay(): Promise<void> {
     await this.browser
       .$('[data-role="identity-management"]')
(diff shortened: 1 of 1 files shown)


lmuntaner:
@sea-snake please review

sea-snake (approved):

PR #2869, opened 2025-02-18
Add event per page

lmuntaner:
# Motivation

Better understand where users land when visiting Internt Identity.

# Changes

* Add different events depending on the flow that the user visits.

# Tests

No new functionality.

### DIFF
--- a/src/frontend/src/index.ts
+++ b/src/frontend/src/index.ts
@@ -18,6 +18,7 @@ void createSpa(async (connection) => {
   // Figure out if user is trying to add a device. If so, use the anchor from the URL.
   const addDeviceAnchor = getAddDeviceAnchor();
   if (nonNullish(addDeviceAnchor)) {
+    analytics.event("page-add-new-device");
     const userNumber = addDeviceAnchor;
     // Register this device (tentatively)
     const registerDeviceResult = await registerTentativeDevice(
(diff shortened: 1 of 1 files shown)


lmuntaner:
@LXIF please review

lmuntaner:
@sea-snake please review

src/frontend/src/index.ts
```diff
@@ -18,6 +18,7 @@ void createSpa(async (connection) => {
   // Figure out if user is trying to add a device. If so, use the anchor from the URL.
   const addDeviceAnchor = getAddDeviceAnchor();
   if (nonNullish(addDeviceAnchor)) {
+    analytics.event("page-add-new-device");
```
sea-snake, line 21: Looking at the Plausible API spec and tracking lib, shouldn't this be: `analytics.trackPageview({ url: 'add/device' })`. I'm afraid that using custom events instead of `pageView` events (according to doc a special event type) will cause issues later when we migrate to actual routing.

sea-snake: Fair enough, then for now custom events is fine indeed.

sea-snake (approved):

src/frontend/src/index.ts
lmuntaner: Not idea, but better than nothing

PR #2873, opened 2025-02-20
Let user pick RP id instead of automatically picking the RP id

sea-snake:
Let user pick RP id instead of automatically picking the RP id.

# Changes

- Pass optional `pickRpId` method to connection passkey login so that a screen is rendered where user picks RP id to authenticate with.
- Keep existing flow logic "as is", the flows of the picked RP id are prioritized within the existing generated flows.
- Updated e2e tests to automatically pick the first RP id from the list (results in same behavior as previously in e2e)

### DIFF
--- a/src/frontend/src/flows/authorize/index.json
+++ b/src/frontend/src/flows/authorize/index.json
@@ -40,6 +40,6 @@
     "pick_alternative_of": "Choose the Internet Identity that connects to",
     "pick_alternative_of_join": "and",
     "multiple_domains": "Multiple domains",
-    "multiple_domains_pick_to_continue": "Multiple domains found, pick one to continue"
+    "multiple_domains_pick_to_continue": "We found passkeys registered in multiple domains for this identity. In which domain did you register your current device?"
   }
 }
--- a/src/frontend/src/flows/authorize/index.json
+++ b/src/frontend/src/flows/authorize/index.json
@@ -38,6 +38,8 @@
     "pick_subtitle": "to connect",
     "pick_subtitle_join": "to",
     "pick_alternative_of": "Choose the Internet Identity that connects to",
-    "pick_alternative_of_join": "and"
+    "pick_alternative_of_join": "and",
+    "multiple_domains": "Multiple domains",
+    "multiple_domains_pick_to_continue": "Multiple domains found, pick one to continue"
   }
 }
--- a/src/frontend/src/test-e2e/views.ts
+++ b/src/frontend/src/test-e2e/views.ts
@@ -542,6 +542,12 @@ export class AuthenticateView extends View {
     await this.browser.$('[data-role="anchor-input"]').waitForDisplayed();
     await this.browser.$('[data-role="anchor-input"]').setValue(anchor);
     await this.browser.$('[data-action="continue"]').click();
+
+    // If there are multiple RP ids, always pick the first one
+    const selectRpId = await this.browser.$("[data-rp-id]");
+    if (await selectRpId.isExisting()) {
+      await selectRpId.click();
+    }
   }
 
   async expectAnchorInputField(): Promise<void> {
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -464,7 +464,7 @@ export class Connection {
       if (
         nonNullish(pickRpId) &&
         this.webAuthFlows.currentIndex === 0 &&
-        uniqueRpIds.length > 0
+        uniqueRpIds.length > 1
       ) {
         const result = await pickRpId(uniqueRpIds);
         if (result.kind === "rpIdPickCancelled") {
--- a/src/frontend/src/utils/iiConnection.ts
+++ b/src/frontend/src/utils/iiConnection.ts
@@ -30,6 +30,10 @@ import {
   UserNumber,
   VerifyTentativeDeviceResponse,
 } from "$generated/internet_identity_types";
+import {
+  RpIdPickCancelled,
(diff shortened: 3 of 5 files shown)


sea-snake:
image

src/frontend/src/flows/authorize/index.json
```diff
@@ -38,6 +38,8 @@
     "pick_subtitle": "to connect",
     "pick_subtitle_join": "to",
     "pick_alternative_of": "Choose the Internet Identity that connects to",
-    "pick_alternative_of_join": "and"
+    "pick_alternative_of_join": "and",
+    "multiple_domains": "Multiple domains",
+    "multiple_domains_pick_to_continue": "Multiple domains found, pick one to continue"
```
lmuntaner, line 43: I would explain a bit more. How about: "We found passkeys registered in multiple domains for this identity. In which domain did you register your current device?"

I'm not sure about using "passkeys" in the first sentence "we found passkeys" and maybe use "we found devices". But I'm assuming this is only for power users that know what passkeys are.

src/frontend/src/test-e2e/views.ts
```diff
@@ -542,6 +542,12 @@ export class AuthenticateView extends View {
     await this.browser.$('[data-role="anchor-input"]').waitForDisplayed();
     await this.browser.$('[data-role="anchor-input"]').setValue(anchor);
     await this.browser.$('[data-action="continue"]').click();
+
+    // If there are multiple RP ids, always pick the first one
+    const selectRpId = await this.browser.$("[data-rp-id]");
```
lmuntaner, line 547: Could we add an e2e test that will fail if the screen is not present?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -416,29 +427,61 @@ export class Connection {
...
-        relatedOrigins: this.canisterConfig.related_origins[0] ?? [],
-      });
-      this.webAuthFlows = {
-        flows,
-        currentIndex: 0,
-      };
+    if (DOMAIN_COMPATIBILITY.isEnabled()) {
+      // Create flows if not initialized yet
+      if (isNullish(this.webAuthFlows)) {
+        const flows = findWebAuthnFlows({
```
lmuntaner, line 447: Is your idea to remove the `findWebAuthnFlows` in the future when this becomes the default flow to pick RP ID?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -416,29 +427,61 @@ export class Connection {
...
+      if (
+        nonNullish(pickRpId) &&
+        this.webAuthFlows.currentIndex === 0 &&
+        uniqueRpIds.length > 1
+      ) {
+        const result = await pickRpId(uniqueRpIds);
+        if (result.kind === "rpIdPickCancelled") {
+          return { kind: "webAuthnFailed" };
+        }
+        void (result satisfies RpIdPickSuccess);
```
lmuntaner, line 473: Do we need the `void`?

src/frontend/src/utils/iiConnection.ts
```diff
@@ -416,29 +427,61 @@ export class Connection {
...
+
+        const pickedFlows = this.webAuthFlows.flows.filter(
+          (flow) => flow.rpId === result.rpId
+        );
+        const otherFlows = this.webAuthFlows.flows.filter(
+          (flow) => flow.rpId !== result.rpId
+        );
+        this.webAuthFlows.flows = [...pickedFlows, ...otherFlows];
```
lmuntaner, line 481: Why this? It's not clear to me why we would want to reassign the flows.

lmuntaner (commented):
Looks very good!

One minor question that I don't understand and a comment in an e2e test.

src/frontend/src/flows/authorize/index.json
sea-snake: We already renamed the device section to Passkey section so I think passkeys are fine. I'll update the description with your suggestion since that's indeed more clear.

src/frontend/src/utils/iiConnection.ts
sea-snake: 1. There might be some flows where we don't want this additional step of user interaction.
2. It's still used to pick the order of origins shown to the user (e.g. we can optimize the function to sort by last used)
3. It's still used after WebAuthn fails, it'll show the blue info tooltip and user can retry with next flow.

So technically, this screen only "prioritize a given flow first" but flows are still used, and considering flows are more than just origin, I would keep it that way.

src/frontend/src/utils/iiConnection.ts
sea-snake: See my comment here: https://github.com/dfinity/internet-identity/pull/2873#discussion_r1967332190

The origin picker is only prioritizing the first flow that the user gets, it's not replacing the flow implementation.

src/frontend/src/utils/iiConnection.ts
lmuntaner: Ah, I see that this is used only for the first step.

I'm not sure we should use the other flow afterwards.

We don't know for sure that the user cancelled because the credential was wrong. Having a following flow different than the first one without being chosen by the user seems a bit inconsistent and might be confusing to users.

I would always use the selector. Then, it should be their job to select the correct one...

sea-snake: > Then, it should be their job to select the correct one...
But the selector only selects origin, not a specific flow. There could be multiple flows for the same origin.

sea-snake:
Closing for now since we opted to go a different direction.

PR #2874, opened 2025-02-20
Show device domains if multiple are found

sea-snake:
Show device domains if multiple are found. 

# Changes

- Add optional `rpId` property to `Authenticator`.
- Assign value to this property using existing logic from `domainInfo`.
- Split `domainInfo` into `domainLabel` to just get the rp Id instead of a TemplateResult.
- Render `rpId` if it's non-nullish, see screenshot below.

### DIFF
--- a/src/frontend/src/flows/manage/index.ts
+++ b/src/frontend/src/flows/manage/index.ts
@@ -741,7 +741,7 @@ const domainLabel = (
   if (nonNullish(commonOrigin)) {
     return undefined;
   }
-  return device.origin[0] ?? LEGACY_II_URL;
+  return new URL(device.origin[0] ?? LEGACY_II_URL).hostname;
 };
 
 const unknownError = (): Error => {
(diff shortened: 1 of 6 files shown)


src/frontend/src/flows/manage/index.ts
```diff
@@ -647,6 +647,7 @@ export const devicesFromDevicesWithUsage = ({
 
       const authenticator = {
         alias: device.alias,
+        rpId: domainLabel(device, devices_),
         last_usage: device.last_usage,
         warn: domainWarning(device),
         info: domainInfo(device, devices_),
```
lmuntaner, line 653: Do we still want the `domainInfo`?

lmuntaner (approved):
Minor question, but looks good to me!

src/frontend/src/flows/manage/index.ts
sea-snake: Good point, I'll move this tooltip (explanation part of the tooltip) to this domain line itself instead.

PR #2881, opened 2025-02-25
Refactor removing devices

lmuntaner:
# Motivation

We want to add a confirmation screen instead of an alert when removing devices.

To do that, we'll need to use the same pattern as adding a recovery phrase from the manage page. This means, passing a function that will be triggered with the user interaction instead of calling the function in the Authenticator.

In this PR, I refactor the code so that in a subsequent PR I'll be able to render a new page and then send the user back to the manage page.

# Changes

* New fields in the `Authenticator`: `device` and `canBeRemoved`.
* New helper `onRemoveDevice` that calls `deleteDevice`.
* New parameter `onRemoveDevice` accordingly.

# Tests

* No new tests because there is no new functionality.
* Tested locally.

### DIFF
--- a/src/frontend/src/flows/manage/authenticatorsSection.ts
+++ b/src/frontend/src/flows/manage/authenticatorsSection.ts
@@ -1,3 +1,4 @@
+import { DeviceData } from "$generated/internet_identity_types";
 import {
   infoIcon,
   pulsatingCircleIcon,
@@ -44,12 +45,14 @@ export const dedupLabels = (
 export const authenticatorsSection = ({
   authenticators: authenticators_,
   onAddDevice,
+  onRemoveDevice,
   warnNoPasskeys,
   cleanupRecommended,
   i18n,
 }: {
   authenticators: Authenticator[];
   onAddDevice: () => void;
+  onRemoveDevice: (device: DeviceData) => void;
   warnNoPasskeys: boolean;
   cleanupRecommended: boolean;
   i18n: I18n;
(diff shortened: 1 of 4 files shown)


lmuntaner:
@sea-snake please review

sea-snake:
@lmuntaner please review

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -124,26 +127,33 @@ export const authenticatorItem = ({
...
   index,
   i18n,
   icon,
+  onRemoveDevice,
 }: {
   authenticator: DedupAuthenticator;
   index: number;
   i18n: I18n;
   icon?: TemplateResult;
+  onRemoveDevice: (device: DeviceData) => void;
```
sea-snake, line 144: If the `device` is passed into `authenticatorItem` already as argument and there is only a single `device` passed in, there's no need for the `device` argument in `onRemoveDevice`.

I would suggest keeping the method abstract:

```suggestion
  onRemove: () => void;
```

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -124,26 +127,33 @@ export const authenticatorItem = ({
...
     { action: "rename", caption: "Rename", fn: () => rename() },
   ];
 
-  if (nonNullish(remove)) {
-    settings.push({ action: "remove", caption: "Remove", fn: () => remove() });
+  if (canBeRemoved) {
+    settings.push({
+      action: "remove",
+      caption: "Remove",
+      fn: () => onRemoveDevice(device),
```
sea-snake, line 155:
```suggestion
      fn: onRemove,
```

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -124,26 +127,33 @@ export const authenticatorItem = ({
     dupCount,
     warn,
     info,
-    remove,
+    device,
```
sea-snake, line 130: Unless `authenticatorItem` needs to render the device, it shouldn't need it as argument.

```suggestion
```

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -96,7 +99,7 @@ export const authenticatorsSection = ({
         <div class="c-action-list">
           <ul>
           ${authenticators.map((authenticator, index) =>
-            authenticatorItem({ authenticator, index, i18n })
+            authenticatorItem({ authenticator, index, i18n, onRemoveDevice })
```
sea-snake, line 102: As suggested earlier, from a UI perspective we only have one thing to remove within the `authenticatorItem` component, so the following should suffice.

```suggestion
            authenticatorItem({ authenticator, index, i18n, onRemove: () => onRemoveDevice(authenticator. device) })
```

lmuntaner: Sure

src/frontend/src/flows/manage/authenticatorsSection.ts
lmuntaner: yep

src/frontend/src/flows/manage/authenticatorsSection.ts
lmuntaner: Yep

lmuntaner:

PR #2883, opened 2025-02-26
Create WebAuthn flows based on device order (recently used).

sea-snake:
Create WebAuthn flows based on device order (recently used).

# Changes

- Simplify `findWebAuthnFlows` to simple mapping over devices that are expected to be in recently used order.
- Remove `findWebAuthnRpId` and related methods since they're no longer needed

# Tests

- Added "should return flows in order of devices (recently used)" test
- Removed all "findWebAuthnRpId" tests since method no longer exists


src/frontend/src/utils/findWebAuthnFlows.ts
```diff
@@ -39,46 +35,35 @@ export const findWebAuthnFlows = ({
...
-      filteredCredentials,
-      relatedOrigins
-    );
+  // The devices are expected to be ordered by recently used first
+  const deviceRpIds = [
+    ...new Set(
+      devices
+        // Device origin to RP ID (hostname)
+        .map((device) =>
+          device.origin !== currentOrigin && currentOrigin !== II_LEGACY_ORIGIN
```
lmuntaner, line 49: Why check `currentOrigin !== II_LEGACY_ORIGIN`? Why not set `undefined` when currentOrigin is different than the legacy?

src/frontend/src/utils/findWebAuthnFlows.ts
```diff
@@ -39,46 +35,35 @@ export const findWebAuthnFlows = ({
...
+  );
 
-  while (filteredCredentials.length > 0) {
-    const rpId = findWebAuthnRpId(
-      currentOrigin,
-      filteredCredentials,
-      relatedOrigins
-    );
+  // The devices are expected to be ordered by recently used first
+  const deviceRpIds = [
```
lmuntaner, line 44: Shall we name it `orderedDeviceRpIds`? To make it clear that we care about the order.

lmuntaner (approved):
Nice! Already approving, minor comments

src/frontend/src/utils/findWebAuthnFlows.ts
sea-snake: Some tests are failing related to this, the following implementation should work as expected:

```ts
        // Device origin to RP ID (hostname)
        .map((device) =>
          device.origin === currentOrigin ||
          (currentOrigin === II_LEGACY_ORIGIN && isNullish(device.origin))
            ? undefined
            : new URL(device.origin ?? II_LEGACY_ORIGIN).hostname
        )
```

src/frontend/src/utils/findWebAuthnFlows.ts
sea-snake: Will do!

PR #2888, opened 2025-02-28
Fix recommend cleanup

sea-snake:
Fix recommend cleanup, only show when there are different origins across authenticators (exclude recovery devices).

### DIFF
--- a/src/frontend/src/flows/manage/authenticatorsSection.ts
+++ b/src/frontend/src/flows/manage/authenticatorsSection.ts
@@ -107,25 +107,21 @@ export const authenticatorsSection = ({
         <ul>
           ${authenticators.map((authenticator, index) =>
             authenticatorItem({
-              authenticator: {
-                ...authenticator,
-                // Only show rpId if cleanup is recommended
-                rpIdLabel: cleanupRecommended
-                  ? authenticator.rpIdLabel
-                  : undefined,
-              },
+              authenticator,
               index,
               i18n,
               onRemove: () => onRemoveDevice(authenticator.device),
+              showRpId: cleanupRecommended,
             })
-          )}</ul>
-          <div class="c-action-list__actions">
-            <button
-              .disabled=${authenticators.length >= MAX_AUTHENTICATORS}
-              class="c-button c-button--primary c-tooltip c-tooltip--onDisabled c-tooltip--left"
-              @click="${() => onAddDevice()}"
-              id="addAdditionalDevice"
-            >
+          )}
+        </ul>
+        <div class="c-action-list__actions">
+          <button
+            .disabled=${authenticators.length >= MAX_AUTHENTICATORS}
+            class="c-button c-button--primary c-tooltip c-tooltip--onDisabled c-tooltip--left"
+            @click="${() => onAddDevice()}"
+            id="addAdditionalDevice"
+          >
               <span class="c-tooltip__message c-card c-card--tight"
               >You can register up to ${MAX_AUTHENTICATORS} authenticator devices.
                 Remove a device before you can add a new one.</span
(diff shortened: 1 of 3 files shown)


src/frontend/src/flows/manage/index.ts
```diff
@@ -679,7 +679,10 @@ export const devicesFromDevicesWithUsage = ({
       const canBeRemoved = !(hasSingleDevice && !hasOtherAuthMethods);
       const authenticator: Authenticator = {
         alias: device.alias,
-        rpId: rpIdLabel(device, devices_),
+        rpId: rpIdLabel(
```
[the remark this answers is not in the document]
sea-snake: I think there might be a slight confusion, recovery devices are already handled separately above in this file in line 611 and returned in 612. I'm filtering the recovery devices out here since this line is already specifically setting a property on a authenticator (non recovery) device. Later when `cleanupRecommended ` is defined (in another file) this already solely checks authenticators, which are passed as a separate list besides the recovery devices list prop.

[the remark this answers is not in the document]
sea-snake: Renamed rpId to rpIdLabel and moved the code that decides if label should be rendered or not to parent component of authenticator.

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -107,7 +107,13 @@ export const authenticatorsSection = ({
         <ul>
           ${authenticators.map((authenticator, index) =>
             authenticatorItem({
-              authenticator,
+              authenticator: {
+                ...authenticator,
+                // Only show rpId if cleanup is recommended
+                rpIdLabel: cleanupRecommended
```
lmuntaner, line 113: What if we pass `cleanUpRecommended` to `authenticatorItem` and that one decides to render the `rpIdLabel` or not depending on the value?

I think it's more consistent meaning of `rpIdLabel` than removing it just because we don't want to render it.

lmuntaner: Yes, `showRpId` makes more sense to me.

I didn't like passing another authenticator data than is created.

src/frontend/src/flows/manage/types.ts
```diff
@@ -5,8 +5,9 @@ import { TemplateResult } from "lit-html";
 export type Authenticator = {
   alias: string;
   last_usage: [] | [bigint];
-  rpId?: string;
   rename: () => void;
+  // RP ID (shown to user if there are multiple different domains)
+  rpId?: string;
```
lmuntaner, line 10: Is this still optional?

src/frontend/src/flows/manage/authenticatorsSection.ts
```diff
@@ -245,15 +249,15 @@ const itemWarning = ({
...
+}): TemplateResult => html` <div class="c-action-list__action">
   <span class="c-tooltip c-icon c-icon--error" tabindex="0"
     >${warningIcon}<span class="c-tooltip__message c-card c-card--tight"
       >${warn}</span
     ></span
   >
 </div>`;
 
-const itemInfo = (msg: TemplateResult): TemplateResult => html`<div
+const itemInfo = (msg: TemplateResult): TemplateResult => html` <div
```
lmuntaner, line 260: Is this your linter?

lmuntaner (approved):
Looks good, thanks!

Minor question and we should also look into your linter that added the space after `html`. Ideally we all have the same rules.

src/frontend/src/flows/manage/authenticatorsSection.ts
sea-snake: I didn't notice this earlier, but after diving into it further, it seems to be an annoying thing my IDE does, looking into fixing this IDE config now.

lmuntaner: Thanks!

src/frontend/src/flows/manage/types.ts
sea-snake: For now unfortunately since pinAuthenticators are also using it.

PR #2891, opened 2025-03-03
Fix update AAGUID data script 

sea-snake:
Fix update AAGUID data script, after some debugging found and fixed two issues:
- cut was adding new lines, replaced with awk
- base64 on Ubuntu only supports base64 not base64url so had to add additional tr command

Removed the tmp files from previous PR, doesn't seem to be needed and it makes code harder to read.


lmuntaner (approved):
🤞

PR #2895, opened 2025-03-03
Omit PR author from job

sea-snake:
Omit PR author from job as suggested by IDX and shown here: https://github.com/dfinity/ic/blob/97b7ca0b0e6eb8bd769e0f9d6facff1239f6d0ae/.github/workflows/container-base-images.yml#L142


lmuntaner (approved):
Why is this necessary?

Already approving.

sea-snake:
> Why is this necessary?
> 
> Already approving.

I previously updated the gix author to the IDX author as mentioned in the IDX example. But that was a commit author, not PR author. When I asked for a recommended approach IDX suggested to remove author altogether from the PR action and pin the action version, as shown above.

PR #2898, opened 2025-03-04
Fix bootstrap script and upgrade Rust

lmuntaner:
# Motivation

Bootstrap script was broken for an upstream change in the rustup installer.

https://github.com/dfinity/internet-identity/actions/runs/13648511573/job/38152844935

We need to make rustup install the toolchain with `rustup show`.

# Changes

* Run `rustup show` after the rustup installation script.
* Upgrade Rust version.
* Fix clippy issue.

# Tests

CI tests pass.

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -21,6 +21,7 @@ echo "using rust version '$rust_version'"
 
 # here we set the toolchain to 'none' and rustup will pick up on ./rust-toolchain.toml
 run curl --fail https://sh.rustup.rs -sSf | run sh -s -- -y --default-toolchain "none" --no-modify-path
+rustup toolchain install nightly
 
 echo "looking for ic-wasm 0.8.5"
 if [[ ! "$(command -v ic-wasm)" || "$(ic-wasm --version)" != "ic-wasm 0.8.5" ]]
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -28,6 +28,9 @@ if ! rustup toolchain list | grep -q "$rust_version"; then
     run rustup toolchain install "$rust_version"
 fi
 
+# New changes also don't install targets
+run rustup target add wasm32-unknown-unknown --toolchain
+
 # Set the default toolchain to the required version
 run rustup default "$rust_version"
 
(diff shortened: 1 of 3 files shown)


lmuntaner:

src/internet_identity/src/storage.rs
```diff
@@ -223,7 +223,7 @@ pub struct Storage<M: Memory> {
         StableBTreeMap<StorableOpenIdCredentialKey, StorableAnchorNumberList, ManagedMemory<M>>,
 }
 
-#[repr(packed)]
+#[repr(C, packed)]
```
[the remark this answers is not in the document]
lmuntaner: This was a clippy issue:

```
error: item uses `packed` representation without ABI-qualification
```

Copilot explained it and helped me fix it:

occurs because Rust’s default behavior for #[repr(packed)] has changed. In recent versions of Rust and Clippy, when you use:
```
#[repr(packed)]
struct Header {
    magic: [u8; 3],
    version: u8,
    salt: [u8; 32],
}
```

it implicitly defaults to:
```
#[repr(Rust, packed)]
```
This has no stable ABI, meaning the memory layout may change between Rust versions, making the struct unsafe to use in FFI (Foreign Function Interface) or low-level memory manipulation.

How to Fix the Issue?

Solution 1: Explicitly Use #[repr(C, packed)]

If your struct is meant to have a well-defined memory layout (for example, if it will be used in binary formats, FFI, or networking):
```
#[repr(C, packed)]
struct Header {
    magic: [u8; 3],
    version: u8,
    salt: [u8; 32],
}
```
This ensures a stable memory layout similar to C structs.

sea-snake (approved):

scripts/bootstrap
```diff
@@ -22,6 +22,18 @@ echo "using rust version '$rust_version'"
 # here we set the toolchain to 'none' and rustup will pick up on ./rust-toolchain.toml
 run curl --fail https://sh.rustup.rs -sSf | run sh -s -- -y --default-toolchain "none" --no-modify-path
 
+# https://blog.rust-lang.org/2025/03/02/Rustup-1.28.0.html
+# Check if the required toolchain is installed; if not, install it
+if ! rustup toolchain list | grep -q "$rust_version"; then
```
nmattia, line 27: is the check necessary? wouldn't `rustup toolchain install ...` simply be a no-op if the toolchain is already installed?

lmuntaner: Yes, I'm just a nooby for now. I thought that this would do no harm when I saw it suggested somewhere and decided to add it.

We'll improve this to set the rustup version, so I'll remove this check.

PR #2903, opened 2025-03-05
OpenID Connect Google Login

LXIF:
# Motivation

This PR contains the necessary functionality to log into II using a previously connected google account.

# Changes

Primarily Frontend code to get the necessary delegations from the backend based on a google login.

# Tests

E2E tests are added in this PR (TODO).


src/frontend/src/utils/utils.ts
```diff
@@ -411,3 +414,24 @@ export const fromBase64URL = (base64Url: string): ArrayBuffer =>
       .replace(/_/g, "/")
       .padEnd(Math.ceil(base64Url.length / 4) * 4, "=")
   );
+
+export interface FrontendSignedDelegation {
```
sea-snake, line 418: Isn't this basically a `DelegationChain` instance?

sea-snake:
Can you add a screenshot of the login screen in the comments here?

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -274,6 +287,49 @@ export const authenticateBoxFlow = async <I>({
       allowPinLogin,
     });
 
+  const doLoginWithGoogle = async (connection: Connection) => {
```
lmuntaner, line 290: For better readability, having the return types defined would be nice.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -299,6 +358,12 @@ export const authenticateBoxFlow = async <I>({
       return await doRegister();
     }
 
+    if (result.tag === "open_id_google") {
+      const loginResult = await doLoginWithGoogle(connection);
+      if (!loginResult) throw toast.error(copy.failed_to_login_with_google);
```
lmuntaner, line 363: Have you seen the pattern of `throw toast.error` somewhere? What will it throw?

Looking to `authenticateBoxFlow`, I don't see any error being thrown. Instead, it returns a different value that is handled afterwards. Shouldn't we follow the same pattern?

Can you also use `isNullish`? I thought we had a lint rule to not use `!` unless it's a boolean value.

And this will also show two toast errors. The one within `doLoginWithGoogle` and one after.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -479,10 +545,22 @@ export const authnTemplates = (i18n: I18n, props: AuthnTemplates) => {
               useExistingProps.addDevice(userNumber)
             )}
           id="addNewDeviceButton"
-          class="c-button c-button--textOnly"
+          class="c-button c-button--secondary"
         >
           Continue with another device
         </button>
+        ${OPENID_AUTHENTICATION.isEnabled()
```
lmuntaner, line 552: Maybe in another PR, but do we want to also add a button in the list of Anchors?

LXIF: Would make sense, yes.

src/frontend/src/utils/utils.ts
LXIF: A `DelegationChain` instance has fields:

```typescript
    readonly delegations: SignedDelegation[];
    readonly publicKey: DerEncodedPublicKey;
```

This is a single signed delegation. This was previously declared ad-hoc in `fetchDelegation.ts`:

```typescript
export interface Delegation {
  delegation: {
    pubkey: Uint8Array;
    expiration: bigint;
    targets?: Principal[];
  };
  signature: Signature;
}
```

The reason this exists is that you can't pass the candid SignedDelegation you receive from the backend directly to the DelegationChain on the frontend, it needs to be transformed first. I decided to rename it to `FrontendSignedDelegation` because it's just a transformation of the `SignedDelegation` we get from the backend to be usable in the frontend.

src/frontend/src/components/authenticateBox/index.ts
LXIF: done

src/frontend/src/utils/utils.ts
sea-snake: I mean agent-js already exports an interface that could be used: https://github.com/dfinity/agent-js/blob/86a16f77868240de0918255f9d3dd18c4c984856/packages/identity/src/identity/delegation.ts#L83

src/frontend/src/components/authenticateBox/index.ts
lmuntaner: Will you create an issue to keep track of this?

src/frontend/src/components/authenticateBox/index.ts
lmuntaner: Thanks!

src/frontend/src/utils/utils.ts
LXIF: Replaced it now, there was a name collision with a generated type so I hadn't seen it - thx a lot!

LXIF:
> Can you add a screenshot of the login screen in the comments here?
Screenshot 2025-03-10 at 16 18 55

src/frontend/src/components/authenticateBox/index.ts
LXIF: Adjusted to regular pattern (returning a `FlowError`)

src/frontend/src/components/authenticateBox/index.ts
LXIF: discussed this with Erika today, we may want to wait a bit and definitely also do it in the google login styling that people are used to. but did create the issue in jira.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -279,6 +294,57 @@ export const authenticateBoxFlow = async <I>({
...
+      sessionIdentity
+    );
+
+    return {
+      kind: "loginSuccess" as const,
+      connection: authenticatedConnection,
+      userNumber: authenticatedConnection.userNumber,
+      showAddCurrentDevice: false,
+      newAnchor: false,
+      authnMethod: "passkey" as const, // we are returning passkey here because we don't want dapps to be able to block based on openID login
```
lmuntaner, line 344: Why are dapps receiving the return of this function?

Shouldn't this be changed when we send it to the dapp and not here?

I also wonder whether we should remove or change the values here, but this is another PR.

Maybe if we implement the new standard then we'll lose this field and we don't need to worry about it.

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -304,6 +370,13 @@ export const authenticateBoxFlow = async <I>({
       return await doRegister();
     }
 
+    if (result.tag === "open_id_google") {
+      const loginResult = await doLoginWithGoogle(connection);
+      if (isNullish(loginResult))
+        return { kind: "googleLoginFailed" } as FlowError;
```
lmuntaner, line 376: Why is casting `as FlowError` necessary?

src/frontend/src/components/authenticateBox/index.ts
```diff
@@ -484,10 +560,22 @@ export const authnTemplates = (i18n: I18n, props: AuthnTemplates) => {
...
         </button>
+        ${OPENID_AUTHENTICATION.isEnabled()
+          ? html`
+              <button
+                @click=${() =>
+                  withUserNumber(() => useExistingProps.loginOpenIDGoogle())}
+                id="addNewDeviceButton"
+                class="c-button c-button--secondary"
+              >
+                Continue with Google
```
lmuntaner, line 575: Don't we have i18n in this screen?

lmuntaner (commented):
Thanks looking good!

Can we have a video of how the flow looks like? Thanks!

src/frontend/src/components/authenticateBox/index.ts
LXIF: The `authenticationProtocol` in `postMessageInterface.ts` has:
```typescript
const authenticateResult = await authenticate(authContext);
// ...other code...
  window.opener.postMessage(
    {
      kind: "authorize-client-success",
      delegations: authenticateResult.delegations,
      userPublicKey: authenticateResult.userPublicKey,
      authnMethod: authenticateResult.authnMethod,
    } satisfies AuthResponse,
    authContext.requestOrigin
  );
```
...the `authenticate()` function passed to the `authenticationProtocol` in `flows/authorize/index.ts` also calls the `authenticateBox` and takes the `authnMethod` from there, then combines it with the result of the `fetchDelegation` call:

```typescript
  const authSuccess = await authenticateBox({
    connection,
    i18n,
    templates: authnTemplateAuthorize({
      origin: authContext.requestOrigin,
      derivationOrigin: authContext.authRequest.derivationOrigin,
      i18n,
      knownDapp: getDapps().find((dapp) =>
        dapp.hasOrigin(authContext.requestOrigin)
      ),
    }),
    allowPinLogin: authContext.authRequest.allowPinAuthentication ?? true,
    // Registration with PIN is not allowed anymore
    allowPinRegistration: false,
    autoSelectionIdentity: autoSelectionIdentity,
  });
  
  //...
  
    const result = await withLoader(() =>
    fetchDelegation({
      connection: authSuccess.connection,
      derivationOrigin,
      publicKey: authContext.authRequest.sessionPublicKey,
      maxTimeToLive: authContext.authRequest.maxTimeToLive,
    })
  );
  
  //...
  
 const [userKey, parsed_signed_delegation] = result;
    
  return {
    kind: "success",
    delegations: [parsed_signed_delegation],
    userPublicKey: Uint8Array.from(userKey),
    authnMethod: authSuccess.authnMethod,
  };
  ```
  
  So, this is how the authnMethod is passed to a dapp.

File: src/frontend/src/components/authenticateBox/index.ts (hunk 7 above)
Comment (LXIF) [reply src/frontend/src/components/authenticateBox/index.ts:575]: On main branch, `Continue with another device`, `Create new` and `Lost Access?` are hardcoded, but I can of course add it.

Comment (LXIF) [reply src/frontend/src/components/authenticateBox/index.ts]: done

File: src/frontend/src/components/authenticateBox/index.ts (hunk 5 above)
Comment (lmuntaner) [reply src/frontend/src/components/authenticateBox/index.ts:344]: Shouldn't we change this part?

```
window.opener.postMessage(
    {
      kind: "authorize-client-success",
      delegations: authenticateResult.delegations,
      userPublicKey: authenticateResult.userPublicKey,
      authnMethod: authenticateResult.authnMethod,
    } satisfies AuthResponse,
    authContext.requestOrigin
  );
```

Something like:

```
window.opener.postMessage(
    {
      kind: "authorize-client-success",
      delegations: authenticateResult.delegations,
      userPublicKey: authenticateResult.userPublicKey,
      // Do not send "openid" to prevent dapps from blocking it.
      authnMethod: authenticateResult.authnMethod === "openid" ? "passkey" : authenticateResult.authnMethod,
    } satisfies AuthResponse,
    authContext.requestOrigin
  );
```

It seems to me that we could want to use the authnMethod somehow else.

Although, I'm not sure about the advantage of having this anyway.

We should probably migrate to ICRC-34 and ICRC-29 at some point.

src/frontend/src/components/authenticateBox/index.ts
lmuntaner: Thanks! Could have been in another PR as well.

lmuntaner:
@LXIF thanks for the video!

There is a moment without a loader. No need to fix it now, but I guess it's under your radar to be fixed, right?

lmuntaner (approved):
Approving on my side, but wait for @sea-snake who is more involved than me in this feature.

src/frontend/src/components/authenticateBox/index.ts
sea-snake: > Why are dapps receiving the return of this function?

This also surprised me when I added the additional errors to the connect screen. Refactoring this will probably take some effort, so probably should be another PR, but will give us a better idea and control over what we're sending (exposing) to the dapp.

> new standard then we'll lose this field

ICRC-25 and ICRC-34 doesn't share these sorts of details indeed.

LXIF:
> @LXIF thanks for the video!
> 
> There is a moment without a loader. No need to fix it now, but I guess it's under your radar to be fixed, right?

making a ticket =)

PR #2919, opened 2025-03-12
Fix CI job to deploy test dapp and issuer

lmuntaner:
# Motivation

Deploying test dapp and issuer was also broken for the same reason that deploying II was broken.

# Changes

* Add `export DFX_WARNING=-mainnet_plaintext_identity` when deploying dapp and issuer as well.

# Tests

* It will be tested after merging.


lmuntaner:

LXIF (approved):
LGTM!

PR #2921, opened 2025-03-12
Certify /callback URL

sea-snake:
Also certify /callback path by adding an html file for it  so we canhandle callback flows.

# Changes

- Added `callback/index.html`

# Tests

- Updated `should_set_cache_control_for_icons` integration test to resolve failing tests due to test icon being referenced in another js file than expected in the original test implementation.

### DIFF
--- a/src/internet_identity/tests/integration/http.rs
+++ b/src/internet_identity/tests/integration/http.rs
@@ -433,99 +433,110 @@ fn should_set_cache_control_for_icons() -> Result<(), CallError> {
     let index_response = http_request(&env, canister_id, &index_request)?;
 
     // Convert body to string
-    let index_html =
-        String::from_utf8(index_response.body.into_vec()).expect("Failed to parse HTML");
-
-    // Find the SPA URL in the HTML
-    let spa_url = {
+    let index_html = String::from_utf8(index_response.body.into_vec()).expect("Failed to parse HTML");
+    
+    // Find all cacheable.js URLs in the HTML
+    let spa_urls: Vec<String> = {
+        let mut urls = Vec::new();
+        let mut search_start = 0;
         let spa_suffix = "cacheable.js";
-        let spa_end = index_html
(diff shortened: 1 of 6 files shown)


sea-snake:
@LXIF @lmuntaner Please review, I kept the change simple since handling other urls (from e.g. svelte routes) automatically is a bigger change which at that point we might as well consider switching to `ic-asset-certification`.

lmuntaner (commented):
Thanks!

How much work would it be to move it to its own HTML page? Similar to vc-flow.html?

I see that the callback doesn't really use anything from the index.html main flows.

sea-snake:
> Thanks!
> 
> How much work would it be to move it to its own HTML page? Similar to vc-flow.html?
> 
> I see that the callback doesn't really use anything from the index.html main flows.

This wouldn't be much work it's just a different approach. Right now, that's actually a better idea, this avoid the need to modify the `asset_util` dependency.

lmuntaner:
> > Thanks!
> > How much work would it be to move it to its own HTML page? Similar to vc-flow.html?
> > I see that the callback doesn't really use anything from the index.html main flows.
> 
> This wouldn't be much work it's just a different approach. Right now, that's actually a better idea, this avoid the need to modify the `asset_util` dependency.

Yes, I think it makes more sense. It can also be added to vite config the same way `vc-flow/index.html` is added.

sea-snake:
@lmuntaner PR is ready for another review

LXIF (approved):
LGTM!

src/internet_identity/tests/integration/http.rs
lmuntaner, line 1: Why are these changes necessary?

lmuntaner (commented):
Comments about the new test.

Should we add a new test to check that `/callback` is present as HTTP request?

lmuntaner (approved):

PR #2926, opened 2025-03-13
Add unlink account page

sea-snake:
Add unlink account page, will be used in another PR to unlink an account. This page is also needed to explain to the user that they will be signed out if they're currently signed in with this linked account.

# Changes

- Create unlink account template and page
- Add unlink account page to showcase


lmuntaner (approved):
LGTM! Thanks for creating small PRs!

PR #2933, opened 2025-03-14
Registration via Google OIDC backend and interface changes

LXIF:
# Motivation

In order to allow registration with openid, we need to adjust some types and implement the necessary backend functionality.

# Changes

Adjust registration finish types and implement identity creation from open id credential functionality.

# Tests

Added integration test.


src/internet_identity/index_debug.html
lmuntaner, line 1: Is this necessary?

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -168,26 +168,58 @@ pub fn identity_registration_finish(
...
+            // add temp key so the user can keep using the identity used for the registration flow
+            state::with_temp_keys_mut(|temp_keys| {
+                temp_keys.add_temp_key(
+                    &id_reg_finish_arg.authn_method.public_key(),
+                    identity_number,
+                    caller,
+                )
+            });
+        }
+        CreateIdentityData::OpenID(_) => {}
```
lmuntaner, line 182: Registration with Google doesn't use temp keys?

src/internet_identity/tests/integration/openid.rs
```diff
@@ -195,6 +197,28 @@ fn can_get_valid_jwt_delegation() -> Result<(), CallError> {
...
+
+    let (jwt, salt, _claims, test_time, test_principal, _test_authn_method) = openid_test_data();
+
+    let time_to_advance = Duration::from_millis(test_time) - Duration::from_nanos(time(&env));
+    env.advance_time(time_to_advance);
+
+    // Create identity (this will panic if it doesn't work)
+    // the test principal here is from webauthn
+    // this is a bit of a hack as in practice it would be a temporary frontend keypair
+    // but it is not relevant for the purposes of this test, as we just need some principal that verifies with the nonce
```
lmuntaner, line 215: Can you be more specific what it means: "some principal that verifies with the nonce"?

lmuntaner (commented):
A couple of questions but it's looking good!

src/internet_identity/index_debug.html
LXIF: Goddamit, I knew I was going to forget this 🙈

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
LXIF: Correct, to my understanding temp keys are only necessary because we can't create an authenticated connection out of an `AuthenticatorAttestationResponse` which we get from `navigator.credentials.create`. So as a workaround, we use temp keys when registering with WebAuthn for the first time. But shouldn't be necessary with OIDC.

src/internet_identity/tests/integration/openid.rs
LXIF: The nonce here is a hash of the principal and a salt (generated in the frontend). The nonce is sent to and signed by google. This signed message is passed to the backend with the salt, where it can be verified by hashing the caller principal with the salt and comparing the two hashes.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
lmuntaner: Ah, that makes sense.

Will you add the registration flow as well to the schema we are keeping?

https://www.figma.com/board/o4aDFNbwbPOUpguwbRMpHA/OIDC-Integration-Test-Notes?t=09BXrhVKJtdmkgwb-0

When we finish this part we should clean those docs a bit to make it clearer.

It should help us understand what's needed to implement the new passkey flow.

src/internet_identity/tests/integration/openid.rs
lmuntaner: Why not create a random principal then? To make it clear that they are not related.

And just mention precisely what you explained here next to the random principal?

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
sea-snake: A code comment to mention this would probably be helpful.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -168,26 +168,58 @@ pub fn identity_registration_finish(
...
+            activity_bookkeeping(&mut identity, &device.pubkey);
+
+            Operation::RegisterAnchor {
+                device: DeviceDataWithoutAlias::from(device),
+            }
+        }
+        CreateIdentityData::OpenID(openid_registration_data) => {
+            let OpenIDRegFinishArg { jwt, salt } = openid_registration_data;
+            let open_id_credential =
+                openid::verify(jwt, salt).map_err(IdRegFinishError::InvalidAuthnMethod)?;
```
sea-snake, line 211: The error code `InvalidAuthnMethod` seems to be regarding WebAuthn but is used here for OpenID. Probably another error should be added here instead.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -168,26 +168,58 @@ pub fn identity_registration_finish(
...
+
+            identity
+                .add_openid_credential(open_id_credential.clone())
+                .map_err(|err| IdRegFinishError::InvalidAuthnMethod(err.to_string()))?;
```
sea-snake, line 215: Same here

src/internet_identity/tests/integration/openid.rs
LXIF: Because then I would need to go extract test data again, only to arrive at something that is functionally identical. The comment makes it clear to me that the principal is arbitrary, is it unclear to you?

lmuntaner: And what would be the non-hack way? That the JWT is created with a nonce, whose principal will be the temporary keypair used in the frontend?

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
LXIF: @lmuntaner will do @sea-snake will do

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
LXIF: Correct - here `Authn` is meant as an abbreviation for `Authentication` and not for `WebAuthn`.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
LXIF: Hope this is better =)

sea-snake (approved):

PR #2949, opened 2025-03-19
Use is_production to hide or show banner

lmuntaner:
# Motivation

Be able to set beta domains with the not official II banner.

# Changes

Changes to warning banner logic:

* Modified the `showWarningIfNecessary` function to use the `is_production` flag from the configuration instead of checking the related_origins.
* Updated the URL in the warning message to use the first related origin from the configuration, if available, or fallback to the official URL.

# Tests

Tested locally.

### DIFF
--- a/src/frontend/src/banner.ts
+++ b/src/frontend/src/banner.ts
@@ -7,6 +7,7 @@ import { anyFeatures } from "./features";
 // is a flavored build, or if the is_production flag is not set to true.
 export const showWarningIfNecessary = (config: InternetIdentityInit): void => {
   const isProduction: boolean = config.is_production[0] ?? false;
+  const firstUrl: string = config.related_origins[0]?.[0] ?? OFFICIAL_II_URL;
   if (anyFeatures()) {
     showWarning(html`Test only. Do not use your regular Internet Identity!
       <a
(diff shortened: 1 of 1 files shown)


lmuntaner:
@LXIF this means that this week's release need to set the config field "is_production" to "[true]". Otherwise, the banner will be shown in production.

sea-snake (approved):

LXIF (approved):
LGTM!

src/frontend/src/banner.ts
```diff
@@ -16,18 +16,19 @@ export const showWarningIfNecessary = (config: InternetIdentityInit): void => {
...
       >`);
-  } else if (!isOfficialOrigin(window.location.origin, config)) {
+  } else if (!isProduction) {
     showWarning(html`This is not the official Internet Identity.
       <a
         class="features-warning-btn"
         target="_blank"
         rel="noopener noreferrer"
-        href=${OFFICIAL_II_URL}
+        href=${config.related_origins[0] ?? OFFICIAL_II_URL}
```
peterpeterparker, line 25: Randomly noticed this PR and just sharing in case it's interesting—feel free to ignore. But if set, isn’t `config.related_origins[0]` an array rather than a string? If so, won't you end up with something like `href="['hello.com']"`?

lmuntaner: Fixed

sea-snake: This is still one of the most confusing parts of agent-js and I remember the long discussion threads about it 😅

sea-snake: Ah nevermind, It's indeed an optional array, my bad.

PR #2950, opened 2025-03-19
OpenID Registration: add choose-registration-method view and copy

LXIF:
# Motivation

After implementing Register-with-OIDC, we need to let the users choose how to register.

# Changes

Add the view and related copy that lets the user choose their registration method.

# Tests

Current tests pass, no new tests were added. This PR only adds these files, they are still inactive though.


lmuntaner:
Didn't you use the savePasskeyPage? It's already in the flow...

You still need to keep backwards compatibility with registering with PIN to enable the e2e tests for login with PIN.

We still have users with PIN.

Did Erika review this new page or will she do it later?

src/frontend/src/flows/register/chooseRegistrationMethod.ts
lmuntaner, line 1: Can you add it to the showcase as well in case we want to use it.

LXIF:
> Didn't you use the savePasskeyPage? It's already in the flow...
> 
> You still need to keep backwards compatibility with registering with PIN to enable the e2e tests for login with PIN.
> 
> We still have users with PIN.
> 
> Did Erika review this new page or will she do it later?

I took some of the code from that page, but I couldn't use it 1-to-1, did you mean for me to update that one directly then?

LXIF:
Screenshot 2025-03-20 at 15 22 31

This is what it looks like now, the only change is one more button and I changed the "without passkey" to "with pin" for clarity.

src/frontend/src/flows/register/index.ts
```diff
@@ -89,6 +91,7 @@ export const registerFlow = async ({
   registrationAllowed: { isAllowed: boolean; allowedOrigins: string[] };
   pinAllowed: () => Promise<boolean>;
   uaParser: PreloadedUAParser;
+  connection: Connection;
```
lmuntaner, line 94: Wouldn't it be nicer to have the parameter `googleAllowed` directly here to avoid passing the whole "connection" object?

lmuntaner (approved):
Minor comment, looks good.

Can you also change the title and descript of the PR?

lmuntaner (approved):
Thanks!

PR #2951, opened 2025-03-19
OpenID Registration: Extend Connection class

LXIF:
# Motivation
In order to be able to register with an OpenID credential, the Connection class needs to be able to call the appropriate backend endpoints.

# Changes

Add the `openid_identity_registration_finish` method on the Connection class. I stayed with the rusty naming because the other identity registration related methods (`identity_registration_start` and `identity_registration_finish`) do the same. The error handling is the same as in `identity_registration_finish`, so I extracted it into a private method.

# Tests

No new tests were added.

### DIFF
--- a/src/frontend/src/utils/iiConnection.test.ts
+++ b/src/frontend/src/utils/iiConnection.test.ts
@@ -9,7 +9,12 @@ import {
   IdentityMetadata,
   RECOVERY_PAGE_SHOW_TIMESTAMP_MILLIS,
 } from "$src/repositories/identityMetadata";
-import { ActorSubclass, DerEncodedPublicKey, Signature } from "@dfinity/agent";
+import {
+  ActorSubclass,
+  DerEncodedPublicKey,
+  SignIdentity,
+  Signature,
+} from "@dfinity/agent";
 import { DelegationIdentity } from "@dfinity/identity";
 import { IDBFactory } from "fake-indexeddb";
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):

lmuntaner (changes requested):
Oh, I forgot. Question, could we add some unit tests for this function? We started doing that and it would be nice if possible. Mocking all the external dependencies of course.

LXIF:
> Oh, I forgot. Question, could we add some unit tests for this function? We started doing that and it would be nice if possible. Mocking all the external dependencies of course.

I can try doing that, though mocking all the external dependencies may not be trivial.

lmuntaner:
> > Oh, I forgot. Question, could we add some unit tests for this function? We started doing that and it would be nice if possible. Mocking all the external dependencies of course.
> 
> I can try doing that, though mocking all the external dependencies may not be trivial.

Give it a try and if it's too cumbersome add a comment on where you got stuck adding a unit test in the method for future reference.

lmuntaner (approved):
LGTM, thanks!

lmuntaner:
> Ideally Connection class shouldn't call flows, because this means you'd need to mock these flows as seen in this PR. So it might make more sense to implement `openid_identity_registration_finish` as flow instead. Overall this is a nitpick, something that can be ignored for now, but something we'll run into more when working on Svelte.

Agree. But let's use this knowledge for the Svelte rework.

PR #2954, opened 2025-03-19
OpenID credential bookkeeping

sea-snake:
Update device bookkeeping to also support OpenID credentials.

# Changes

- Update `AuthorizationKey` to use `OpenIdCredentialKey` instead of `PublicKey` for `OpenIdCredential`.
- Update `activity_bookkeeping` to accept `(&Anchor, &AurthorizationKey)` instead of `&Device`.
- Update `check_authorization` to forward  `AurthorizationKey` instead of `Device` so bookkeeping can be called without authentication method distinction.
- When an OpenID delegation is created, update anchor with latest OpenID credential from JWT so latest metadata is stored.
- Update anchor `last_activity` to check both device and OpenID credential timestamps.
- Add `openid_credential` to authentication method metrics

### DIFF
--- a/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
+++ b/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
@@ -13,9 +13,10 @@ use ic_cdk::caller;
 use internet_identity_interface::archive::types::{DeviceDataWithoutAlias, Operation};
 use internet_identity_interface::internet_identity::types::IdRegFinishError::IdentityLimitReached;
 use internet_identity_interface::internet_identity::types::{
-    CaptchaTrigger, CheckCaptchaArg, CheckCaptchaError, CreateIdentityData, DeviceData,
-    DeviceWithUsage, IdRegFinishError, IdRegFinishResult, IdRegNextStepResult, IdRegStartError,
-    IdentityNumber, OpenIDRegFinishArg, RegistrationFlowNextStep, StaticCaptchaTrigger,
+    AuthorizationKey, CaptchaTrigger, CheckCaptchaArg, CheckCaptchaError, CreateIdentityData,
+    DeviceData, DeviceWithUsage, IdRegFinishError, IdRegFinishResult, IdRegNextStepResult,
+    IdRegStartError, IdentityNumber, OpenIDRegFinishArg, RegistrationFlowNextStep,
+    StaticCaptchaTrigger,
 };
 
 impl RegistrationFlowState {
@@ -203,7 +204,10 @@ fn create_identity(arg: &CreateIdentityData) -> Result<IdentityNumber, IdRegFini
             identity
                 .add_device(device.clone())
                 .map_err(|err| IdRegFinishError::InvalidAuthnMethod(err.to_string()))?;
-            activity_bookkeeping(&mut identity, &device.pubkey);
+            activity_bookkeeping(
+                &mut identity,
+                &AuthorizationKey::DeviceKey(device.pubkey.clone()),
+            );
 
             Operation::RegisterAnchor {
                 device: DeviceDataWithoutAlias::from(device),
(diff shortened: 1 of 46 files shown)


src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -203,24 +204,30 @@ fn create_identity(arg: &CreateIdentityData) -> Result<IdentityNumber, IdRegFini
...
+            );
 
             Operation::RegisterAnchor {
                 device: DeviceDataWithoutAlias::from(device),
             }
         }
         CreateIdentityData::OpenID(openid_registration_data) => {
             let OpenIDRegFinishArg { jwt, salt } = openid_registration_data;
-            let open_id_credential =
+            let openid_credential =
```
LXIF, line 218: Screenshot 2025-03-20 at 14 00 19

LXIF (approved):
LGTM

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
sea-snake: It's one word everywhere so I couldn't help but see this space 😂

src/internet_identity/src/main.rs
```diff
@@ -861,6 +860,13 @@ mod openid_api {
         let anchor_number = lookup_anchor_with_openid_credential(&openid_credential.key())
             .ok_or(OpenIdDelegationError::NoSuchAnchor)?;
 
+        // Update anchor with latest OpenID credential from JWT so latest metadata is stored
```
lmuntaner, line 863: Do you mean the latest_used_timestamp is updated?

src/internet_identity/src/openid.rs
```diff
@@ -39,13 +39,11 @@ impl OpenIdCredential {
...
     }
+
     pub fn principal(&self) -> Principal {
-        let public_key = self.public_key();
-        Principal::self_authenticating(public_key)
-    }
-    pub fn public_key(&self) -> PublicKey {
         let seed = calculate_delegation_seed(&self.aud, &self.key());
-        der_encode_canister_sig_key(seed.to_vec()).into()
+        let public_key: PublicKey = der_encode_canister_sig_key(seed.to_vec()).into();
```
lmuntaner, line 45: Was this necessary or it's just an improvement?

src/internet_identity/src/stats/activity_stats.rs
```diff
@@ -82,27 +83,29 @@ pub struct OngoingActivityStats<T: ActivityCounter> {
...
-                .update_counters(&current_device);
+        // Anchor device activity across II domains
+        if let AuthorizationKey::DeviceKey(device_key) = current_authorization_key {
+            if let Some(domain) = anchor.device(device_key).and_then(Device::ii_domain) {
+                let context = DomainActivityContext {
+                    anchor,
+                    current_domain: domain,
+                };
+                persistent_state
+                    .domain_active_anchor_stats
```
lmuntaner, line 101: Don't we want to update this for open id?

Does it make sense to have an origin also for open id?

Otherwise, the total active will not match the sum of the domains, right? Or is this tracked somewhere else?

src/internet_identity/src/storage/anchor.rs
```diff
@@ -376,7 +388,25 @@ impl Anchor {
         openid_credential: OpenIdCredential,
     ) -> Result<(), AnchorError> {
         let index = self.openid_credential_index(&openid_credential.key())?;
-        self.openid_credentials[index] = openid_credential;
+        self.openid_credentials[index] = OpenIdCredential {
+            // Don't update last usage timestamp, below `set_openid_credential_usage_timestamp`
```
lmuntaner, line 392: Could you also add why this is the case? That's probably what we'll forget in a few months.

lmuntaner (approved):
Looks good!! A few questions. Feel free to address merge and we can discuss them separately.

src/internet_identity/src/openid.rs
sea-snake: It's not completely clear from the diff, but it's the removal of the public key method from the OpenIdCredential. I removed it since the public key is only needed to get the principal during authentication. Having this public key method only results in it accidentally being used over the key method, which this PR also resolved.

src/internet_identity/src/storage/anchor.rs
sea-snake: I've updated the comment to clarify this.

src/internet_identity/src/openid.rs
lmuntaner: It was just an improvement, then, no?

src/internet_identity/src/stats/activity_stats.rs
lmuntaner: I thought that this also was used for the active anchors, but after reviewing it I see this is done in before this in `active_anchor_stats`

src/internet_identity/src/main.rs
lmuntaner: Ok, thanks

lmuntaner (approved):
LGTM!

src/internet_identity/src/openid.rs
sea-snake: Within this PR I've already changed:
```rs
pub enum AuthorizationKey {
    DevicePubKey(DeviceKey),
    OpenIdPubKey(PublicKey),
}
```

To:

```rs
pub enum AuthorizationKey {
    DeviceKey(DeviceKey),
    OpenIdCredentialKey(OpenIdCredentialKey),
}
```

Since the credential key (iss, sub pair) should be used to reliably identify a credential instead of the public key from the delegation seed. The `OpenIdCredential` `public_key()` method was only used to create a `AuthorizationKey.OpenIdPubKey` which after moving to `AuthorizationKey.OpenIdCredentialKey` is no longer needed.

PR #2959, opened 2025-03-21
OpenID: Improve anchor management

LXIF:
# Motivation

Fixes a bug in the OIDC registration that would allow multiple identities to be registered with the same credential

# Changes

Use anchor_management instead of adding credential directly to anchor. Add error management in frontend to display error properly.

# Tests

No new tests are added.


LXIF:
diff is botched. redoing.

PR #2960, opened 2025-03-21
OpenID: Improve Anchor Management

LXIF:
# Motivation

Fixes a bug in the OIDC registration that would allow multiple identities to be registered with the same credential

# Changes

Use anchor_management instead of adding credential directly to anchor. Add error management in frontend to display error properly.

# Tests

No new tests are added.


src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -214,14 +214,17 @@ fn create_identity(arg: &CreateIdentityData) -> Result<IdentityNumber, IdRegFini
...
+                anchor_management::add_openid_credential(&mut identity, open_id_credential.clone())
+                    .map_err(|err| IdRegFinishError::InvalidAuthnMethod(err.to_string()))?;
 
-            Operation::RegisterAnchorWithOpenIdCredential {
-                iss: open_id_credential.iss.clone(),
-            }
+            //TODO: add activity bookkeeping
+
+            let Operation::AddOpenIdCredential { iss } = add_credential_operation else {
+                unreachable!()
```
sea-snake, line 224: Not sure about using `unreachable` here, yes you could argue that's the case but in practice you already managed to trigger this error by registering an II with the same JWT twice, so making this actually return an error (even if generic) probably makes more sense. I suppose there must be some error in `IdRegFinishError` that could be used?

LXIF: changed the syntax to be simpler ✅

lmuntaner (approved):

PR #2963, opened 2025-03-24
OpenID bookkeeping integration tests

sea-snake:
Implement OpenID bookkeeping integration tests. As a result of these tests, make last usage timestamp optional.

# Changes

- Create integration test that asserts both daily and monthly authentication stats for OpenID.
- Change `last_usage_timestamp` from `Timestamp` to `Option<Timestamp>` so that the `AuthnMethodCounter` can differentiate between new registrations and existing OpenID credentials.

# Tests

Tested to make sure that existing stable structures data correctly handles the type change (to an optional) with existing data.

### DIFF
--- a/src/internet_identity/src/stats/activity_stats/activity_counter/authn_method_counter.rs
+++ b/src/internet_identity/src/stats/activity_stats/activity_counter/authn_method_counter.rs
@@ -56,13 +56,13 @@ impl ActivityCounter for AuthnMethodCounter {
                 .and_then(|d| d.last_usage_timestamp),
             AuthorizationKey::OpenIdCredentialKey(openid_credential_key) => anchor
                 .openid_credential(openid_credential_key)
-                .map(|c| c.last_usage_timestamp),
+                .and_then(|c| c.last_usage_timestamp),
         };
 
         // only count authentications on devices that have not already been counted
         // i.e. the last usage timestamp is before the start of the window
         if let Some(timestamp) = last_usage_timestamp {
-            if timestamp >= self.start_timestamp {
+            if timestamp > self.start_timestamp {
                 return;
             }
         }
@@ -93,7 +93,9 @@ impl ActivityCounter for AuthnMethodCounter {
             }
             AuthorizationKey::OpenIdCredentialKey((iss, _)) => {
                 if let Some(map) = &mut self.openid_credential_auth_counter {
-                    map.insert(iss.clone(), map.get(iss).unwrap_or(&0) + 1);
+                    map.entry(iss.clone())
+                        .and_modify(|count| *count += 1)
+                        .or_insert(1);
                 }
             }
         }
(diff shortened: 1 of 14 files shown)


src/internet_identity/tests/integration/activity_stats/authn_methods.rs
```diff
@@ -242,6 +245,67 @@ fn should_keep_stats_across_upgrades() -> Result<(), CallError> {
...
+        "internet_identity_daily_active_authn_methods{type=\"openid\",issuer=\"https://accounts.google.com\"}",
+        1f64,
+    );
+
+    // Check monthly stats, some activity is required first to update the stats
+    env.advance_time(Duration::from_secs(MONTH_SECONDS - DAY_SECONDS));
+    create_identity_with_authn_method(&env, canister_id, &test_authn_method());
+    assert_metric(
+        &get_metrics(&env, canister_id),
+        "internet_identity_monthly_active_authn_methods{type=\"openid\",issuer=\"https://accounts.google.com\"}",
```
lmuntaner, line 304: Why isn't this 2?

To test that the counter actually acumulate, could we add one more identity daily check that it moves to 2, then pass a day add one after, check that daily is back to 1 and that monthly is 3?

lmuntaner (commented):
Looking good, I added a comment to improve tests.

src/internet_identity/tests/integration/activity_stats/authn_methods.rs
sea-snake: I wanted to do this, but that would require me to create the OpenID JWT data all over again from scratch (this would be a 3rd time).

If we want to do this, I'd suggest first making helpers in another PR to create OpenID test data on demand that can be used in these tests (instead of snapshots of Google data).

lmuntaner: Now I understand your comment on Slack.

Let's do this later and add a comment with the link to the issue.

But why isn't it 2? Shouldn't it?

We will be able to test this in prod next week as well.

src/internet_identity/src/stats/activity_stats/activity_counter/authn_method_counter.rs
```diff
@@ -93,7 +93,9 @@ impl ActivityCounter for AuthnMethodCounter {
             }
             AuthorizationKey::OpenIdCredentialKey((iss, _)) => {
                 if let Some(map) = &mut self.openid_credential_auth_counter {
-                    map.insert(iss.clone(), map.get(iss).unwrap_or(&0) + 1);
+                    map.entry(iss.clone())
+                        .and_modify(|count| *count += 1)
```
LXIF, line 97: Slick!

src/internet_identity/tests/integration/activity_stats/authn_methods.rs
lmuntaner: Oops, I didn't read well that the new identity created was using a normal method. My bad.

lmuntaner (approved):
Thanks for the clarifications!

src/internet_identity/tests/integration/activity_stats/authn_methods.rs
sea-snake: Added todo comment.

LXIF (approved):
LGTM!

PR #2964, opened 2025-03-25
Calculate and send flow duration in Funnel class

lmuntaner:
# Motivation

Track how long users take on II flows to better understand whether our developments improve or not the user experience.

Not everything is about making it faster, but if we don't track it, we don't even have this data point.

# Changes

Enhancements to event tracking:

* `src/frontend/src/utils/Funnel.ts`: Added a `#startTimestamp` property to the `Funnel` class to record the start time when `init` is called and calculate the duration when `close` is called. This ensures that the duration of the funnel session is tracked accurately.

Updates to test cases:

* `src/frontend/src/utils/Funnel.test.ts`: Updated the test cases to verify that the `analytics.event` is called with the correct duration when `close` is called after `init`. Added new test cases to ensure that `close` does not trigger any events if `init` was not called.

### DIFF
--- a/src/frontend/src/utils/Funnel.test.ts
+++ b/src/frontend/src/utils/Funnel.test.ts
@@ -105,12 +105,38 @@ describe("Funnel", () => {
     funnel.init();
     funnel.close();
 
+    expect(analytics.event).toHaveBeenCalledTimes(2); // the start-login and end-login events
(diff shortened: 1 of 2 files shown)


LXIF (approved):
LGTM!

PR #2968, opened 2025-03-28
Add handlers first in init test app function

lmuntaner:
# Motivation

The init function in the test application was first waiting on an async function making an HTTP request before setting the button handlers.

This means that there was a race condition before users (e2e user) would click the button and the action would happen.

# Changes

* Prettier fixes in test app file.
* Moved the `await updateAlternativeOriginsView()` call in the `init` function.

### DIFF
--- a/demos/test-app/src/index.tsx
+++ b/demos/test-app/src/index.tsx
@@ -21,58 +21,58 @@ import "./main.css";
 const signInBtn = document.getElementById("signinBtn") as HTMLButtonElement;
 const whoamiBtn = document.getElementById("whoamiBtn") as HTMLButtonElement;
 const updateAlternativeOriginsBtn = document.getElementById(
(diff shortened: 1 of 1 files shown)


demos/test-app/src/index.tsx
```diff
@@ -349,6 +348,7 @@ const init = async () => {
     await actor.update_alternative_origins(newAlternativeOriginsEl.value, mode);
     await updateAlternativeOriginsView();
   };
+  await updateAlternativeOriginsView();
```
lmuntaner, line 351: Second and last important change.

demos/test-app/src/index.tsx
lmuntaner, line 1: Only the two commented lines are important.

The rest are prettier fixes.

sea-snake (approved):

PR #2973, opened 2025-03-31
Format all files with updated prettier.

sea-snake:
Format all files with updated prettier.

# Changes

- Changes from running `npm run format`
- Adding `npm run format-check` back


lmuntaner (approved):
Thanks!

PR #2976, opened 2025-04-01
Move analytics and funnel files

lmuntaner:
# Motivation

Funnels will be how we trigger events. That means we will create funnel instances.

To keep it maintainable I suggest we have them all under the same directory: "utils/analytics".

# Changes

* New directory "analytics" under "utils".
* Move analytics.ts and Funnel.ts along the test.
* Update imports of analyitcs.

# Tests

Only files moved.


lmuntaner:
Closing, I forgot to pull latest changes.

PR #2977, opened 2025-04-01
New analytics directory

lmuntaner:
# Motivation

Funnels will be how we trigger events. That means we will create funnel instances.

To keep it maintainable I suggest we have them all under the same directory: "utils/analytics".

# Changes

* New directory "analytics" under "utils".
* Move analytics.ts and Funnel.ts along the test.
* Update imports of analyitcs.

# Tests

Only files moved.


src/frontend/src/callback-flow.ts
```diff
@@ -1,5 +1,5 @@
 import { callbackFlow } from "$src/flows/redirect";
-import { analytics, initAnalytics } from "$src/utils/analytics";
+import { analytics, initAnalytics } from "$src/utils/analyitcs/analytics";
```
LXIF, line 2: Nit: please rename analyitcs to analytics.

LXIF (dismissed):
Looks good apart from typo in folder name!

src/frontend/src/callback-flow.ts
lmuntaner: thanks!! Done!

src/frontend/src/callback-flow.ts
```diff
@@ -1,5 +1,5 @@
 import { callbackFlow } from "$src/flows/redirect";
-import { analytics, initAnalytics } from "$src/utils/analytics";
+import { analytics, initAnalytics } from "$src/utils/analytics/analytics";
```
sea-snake, line 2: While we're at it, it might be nice to add an index.ts to the folder to re-export all folder content that should be used and available outside the folder.

LXIF: Ah yes, good idea, I second this.

(was not aware initially of [this pattern](https://medium.com/@bobjunior542/maximize-typescript-productivity-use-the-index-file-now-cac2f4f4d1ed) but it makes sense, similar to how it is done in rust)

lmuntaner: We agreed to do it as part of the new Svelte project structure.

sea-snake (approved):

PR #2988, opened 2025-04-02
Add JWT verification to OpenID Connect registration flow

LXIF:
# Motivation

The ProdSec review had identified that we had missed to validate the OIDC JWT on `openid_identity_registration_finish`. This PR corrects this oversight.

# Changes

Add the check to the endpoint, add integration Test

# Tests

Added integration test.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -778,7 +778,7 @@ mod openid_api {
         remove_openid_credential, update_openid_credential,
     };
     use crate::authz_utils::{anchor_operation_with_authz_check, IdentityUpdateError};
-    use crate::openid::{self, OpenIdCredentialKey};
+    use crate::openid::{self, verify, OpenIdCredentialKey};
     use crate::storage::anchor::AnchorError;
     use crate::{
         state, IdentityNumber, OpenIdCredentialAddError, OpenIdCredentialRemoveError,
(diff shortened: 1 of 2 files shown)


src/internet_identity/src/main.rs
```diff
@@ -806,9 +806,12 @@ mod openid_api {
...
         arg: OpenIDRegFinishArg,
     ) -> Result<IdRegFinishResult, IdRegFinishError> {
-        registration::registration_flow_v2::identity_registration_finish(
-            CreateIdentityData::OpenID(arg),
-        )
+        match verify(&arg.jwt, &arg.salt) {
+            Ok(_) => registration::registration_flow_v2::identity_registration_finish(
+                CreateIdentityData::OpenID(arg),
+            ),
+            Err(err) => Err(IdRegFinishError::InvalidAuthnMethod(err)),
```
lmuntaner, line 813: Is this the right error to return?

LXIF: An invalid JWT is an invalid authentication method, no?

sea-snake (approved):

src/internet_identity/src/main.rs
```diff
@@ -806,9 +806,12 @@ mod openid_api {
     fn openid_identity_registration_finish(
         arg: OpenIDRegFinishArg,
     ) -> Result<IdRegFinishResult, IdRegFinishError> {
-        registration::registration_flow_v2::identity_registration_finish(
-            CreateIdentityData::OpenID(arg),
-        )
+        match verify(&arg.jwt, &arg.salt) {
```
sea-snake, line 809: Only nitpick I can find is that I would have written `openid::verify` instead to make it more clear this is an openid verification and to avoid any future conflict with other methods that might be called verify.

sea-snake (approved):

PR #2995, opened 2025-04-04
Override feature flags with query params.

sea-snake:
Override feature flags with query params.

# Changes

- Moved feature flags from legacy to utils
- Export initialized feature flags object so it can be accessed by key
- Add override method to FeatureFlag class
- Add init method to client hook that overrides feature flags based on query params


src/frontend/src/hooks.client.ts
```diff
@@ -0,0 +1,39 @@
...
+  // Override feature flags based on search params before any other code
+  // including other hooks runs that might depend on these feature flags.
+  //
+  // Example: ?feature_flag_openid_authentication=true
+  const url = new URL(window.location.href);
+  for (const [key, value] of url.searchParams.entries()) {
+    if (key.startsWith(FEATURE_FLAG_PREFIX)) {
+      const flag = key.slice(FEATURE_FLAG_PREFIX.length).toUpperCase();
+      if (!(flag in featureFlags)) {
+        console.warn(`Invalid feature flag received '${flag}'`);
```
lmuntaner, line 16: We also have the `enablePin`... Should we move that to this later?

src/frontend/src/lib/flows/recovery/useRecovery.ts
```diff
@@ -182,7 +182,7 @@ const enrollAuthenticator = async ({
       newDevice.getPublicKey().toDer(),
       { unprotected: null },
       newDeviceOrigin ?? window.location.origin,
-      newDevice.rawId,
+      newDevice.rawId
```
lmuntaner, line 185: Should we create a PR fixing all this?

src/frontend/src/lib/utils/featureFlags/index.ts
```diff
@@ -45,6 +45,10 @@ export class FeatureFlag {
     this.#value = this.#defaultValue;
     this.#storage.removeItem(this.#key);
   }
+
+  override(value: boolean) {
```
lmuntaner, line 49: Should we specify in the name that this is not persisted in local storage?

lmuntaner (approved):
A few comments but it looks good to me!

src/frontend/src/lib/flows/recovery/useRecovery.ts
sea-snake: Found the issue, created another PR to fix it.

src/frontend/src/lib/utils/featureFlags/index.ts
sea-snake: Ah yeah, I'll call it `temporaryOverride`.

PR #2998, opened 2025-04-04
Cache all assets with "immutable" in the url path instead of "cacheable"

sea-snake:
Cache all assets with "immutable" in the url path instead of "cacheable"

# Changes

- Change asset path check to look for "immutable" instead of "cacheable".


LXIF (approved):
LGTM!

PR #2999, opened 2025-04-04
Exclude frontend lib from .prettierignore and format all files

sea-snake:
Exclude frontend lib from .prettierignore and format all files again.


LXIF (approved):
LGTM! As per @sea-snake, these are exclusively formatting changes.

PR #3002, opened 2025-04-07
add memory and read/write methods for storing and looking up anchor n…

LXIF:
# Motivation

In order to allow logging in / signing up with discoverable passkeys, we need to be able to retrieve anchor number and pubkey by credential id.

# Changes

Add the map in storage, add read/write methods and related data types.

# Tests

No tests.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -28,7 +28,7 @@ use internet_identity_interface::internet_identity::types::vc_mvp::{
 use internet_identity_interface::internet_identity::types::*;
 use serde_bytes::ByteBuf;
 use std::collections::HashMap;
-use storage::{Salt, Storage};
+use storage::{DiscoverableCredentialData, Salt, Storage};
 
 mod anchor_management;
 mod archive;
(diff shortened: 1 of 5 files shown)


src/internet_identity/src/main.rs
```diff
@@ -905,6 +905,21 @@ mod openid_api {
...
 
+/// API required for the discoverable credentials flow
+mod discoverable_credentials {
+    use crate::anchor_management::lookup_anchor_number_and_pubkey_with_credential_id;
+    use crate::DiscoverableCredentialData;
+    use ic_cdk::query;
+    use internet_identity_interface::internet_identity::types::CredentialId;
+
+    #[query]
+    fn get_discoverable_credential_data(
```
lmuntaner, line 916: This was called `get_pubkey_by_credential_id` in the candid interface: https://github.com/dfinity/internet-identity/compare/main...to-be-shipped-hackathon#diff-623401f1d9297ecff25efc7dc8e6e87983a4d8e0b62e82b8e029375805708392R815

PR #3005, opened 2025-04-07
add mapping updating functionality and small bugfix

LXIF:
# Motivation

The map also needs to be updated when a credential gets added or removed.

# Changes

add functionality to update the mapping when a device is updated.
small refactoring, removing duplicate memory.

# Tests

Nuh-uh

### DIFF
--- a/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
+++ b/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
@@ -215,18 +215,16 @@ fn create_identity(arg: &CreateIdentityData) -> Result<IdentityNumber, IdRegFini
                 &AuthorizationKey::DeviceKey(device.pubkey.clone()),
             );
 
-            match &id_reg_finish_arg.authn_method.authn_method {
-                AuthnMethod::WebAuthn(WebAuthn {
-                    pubkey,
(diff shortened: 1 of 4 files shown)


src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -215,18 +215,16 @@ fn create_identity(arg: &CreateIdentityData) -> Result<IdentityNumber, IdRegFini
...
 
-            match &id_reg_finish_arg.authn_method.authn_method {
-                AuthnMethod::WebAuthn(WebAuthn {
-                    pubkey,
+            if let AuthnMethod::WebAuthn(WebAuthn {
+                pubkey,
+                credential_id,
+            }) = &id_reg_finish_arg.authn_method.authn_method
+            {
+                store_anchor_number_and_pubkey_with_credential_id(
```
lmuntaner, line 223: Didn't you say that this was not needed anymore because the `write` function will take care of it?

LXIF: Good Morning @lmuntaner - correct, this should not be needed anymore, this is the change I mentioned yesterday that I still need to do today.

LXIF:
Closed since @sea-snake is reimplementing and merging to main.

PR #3006, opened 2025-04-08
Lookup identity by device credential id in storage.

sea-snake:
Lookup identity by device credential id in storage.

# Changes

- Add `lookup_anchor_with_device_credential_memory`
- Read previous 4KB stable memory anchor in `write` to compare previous and current devices in below method
- Add `update_lookup_anchors_with_device_credential` method

# Tests

- Added `should_write_and_update_device_credential_lookup` test

### DIFF
--- a/src/internet_identity/src/anchor_management.rs
+++ b/src/internet_identity/src/anchor_management.rs
@@ -257,14 +257,14 @@ fn should_register_openid_credential_only_for_a_single_anchor() {
             iss: openid_credential.iss.clone()
         })
     );
-    storage_borrow_mut(|storage| storage.write(anchor_0.clone()).unwrap());
+    storage_borrow_mut(|storage| storage.write(anchor_0.clone(), false).unwrap());
 
     // Check if adding OpenID credential twice returns an error
     assert_eq!(
         add_openid_credential(&mut anchor_0, openid_credential.clone()),
         Err(AnchorError::OpenIdCredentialAlreadyRegistered)
     );
-    storage_borrow_mut(|storage| storage.write(anchor_0.clone()).unwrap());
+    storage_borrow_mut(|storage| storage.write(anchor_0.clone(), true).unwrap());
 
     // Check if adding OpenID credential to another anchor returns an error
     assert_eq!(
@@ -279,7 +279,7 @@ fn should_register_openid_credential_only_for_a_single_anchor() {
             iss: openid_credential.iss.clone()
         })
     );
-    storage_borrow_mut(|storage| storage.write(anchor_0.clone()).unwrap());
+    storage_borrow_mut(|storage| storage.write(anchor_0.clone(), true).unwrap());
     assert_eq!(
         add_openid_credential(&mut anchor_1, openid_credential.clone()),
         Ok(Operation::AddOpenIdCredential {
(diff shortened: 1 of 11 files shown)


src/internet_identity/src/storage.rs
```diff
@@ -221,6 +226,10 @@ pub struct Storage<M: Memory> {
     lookup_anchor_with_openid_credential_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
     lookup_anchor_with_openid_credential_memory:
         StableBTreeMap<StorableOpenIdCredentialKey, StorableAnchorNumberList, ManagedMemory<M>>,
+    /// Memory wrapper used to report the size of the lookup anchor with device credential memory.
+    lookup_anchor_with_device_credential_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
+    lookup_anchor_with_device_credential_memory:
+        StableBTreeMap<StorableCredentialId, AnchorNumber, ManagedMemory<M>>,
```
lmuntaner, line 232: We don't need the public key in the entry?

Is the idea to read the devices and compare the credential id to return the public key?

lmuntaner (approved):
LGTM! Just a question

src/internet_identity/src/storage.rs
sea-snake: Exactly, discussed this with @LXIF, since we have the credential id in devices, there's indeed no need to also store it in the lookup.

LXIF: Yes, that's the idea now.

src/internet_identity/src/anchor_management.rs
```diff
@@ -257,14 +257,14 @@ fn should_register_openid_credential_only_for_a_single_anchor() {
             iss: openid_credential.iss.clone()
         })
     );
-    storage_borrow_mut(|storage| storage.write(anchor_0.clone()).unwrap());
+    storage_borrow_mut(|storage| storage.write(anchor_0.clone(), false).unwrap());
```
lmuntaner, line 260: Did you push your changes? I thought you were going to use a `create` and `update` instead of the extra parameter.

sea-snake: Didn't push the changes yet, was finalizing the other PR changes 😅

src/internet_identity/src/state.rs
```diff
@@ -163,6 +163,7 @@ pub struct RateLimitState {
     pub token_timestamp: Timestamp,
 }
 
+#[allow(clippy::large_enum_variant)]
```
lmuntaner, line 166: Why is this needed?

lmuntaner (approved):
LGTM! Just a minor comment but already approving.

src/internet_identity/src/state.rs
sea-snake: Clippy was complaining that the enum was too large, likely due to the added memory pushing it past the clippy default threshold. This basically disables that clippy check since we're not changing this enum anytime soon. Also looking at [the lint description](https://rust-lang.github.io/rust-clippy/master/index.html#/large_enum_variant), this doesn't seem to be an actual issue in our use case.

PR #3007, opened 2025-04-08
Refactor Feature Flags functionality to be more in line with svelte patterns

LXIF:
# Motivation

We want to be able to lazily initialize the feature flags so we can safely use them inside load functions. Also, once we use proper svelte components, it would be nice to be able to use store syntax for the feature flags.

# Changes

Package the feature flags inside a readable store while retaining the `FeatureFlag` class so everything else runs the same.

# Tests

Update tests.


LXIF:
> If a class is inside the store, does it rerender when a field changes?
> 
> I think that it might with a `$state` but does it with a Svelte store?

Not sure, didn't test this right now (flows are still imperative) - but likely won't whether it's a state or a store because of how nested it is. In the future it can be refactored again into e.g. separate stores for each FeatureFlag, which should allow that. For now I implemented it to avoid proxies that return proxies.

> You also mention:
> 
> > Put the feature flag overrides into the +layout.ts load function.
> 
> Is that done in following PR?

Had that initially but the query param replacing didn't work and would've become a bit tedious, so we're sticking with the hooks file.

LXIF:
@lmuntaner now it should work as you asked.

src/frontend/src/lib/state/featureFlags.ts
```diff
@@ -0,0 +1,91 @@
...
+      set,
+      update,
+      getFeatureFlag: () => undefined,
+    };
+  }
+
+  // define getter function to pass to flag object
+  const get = () => {
+    let value: boolean = false;
+    update((oldVal) => {
```
lmuntaner, line 35: Are you using `update` to read the value? Did you try having a local variable within  `createFeatureFlagStore` and subscribing to the store and changing the value when the trigger of subscribe is called?

This way the `get` is simpler, no?

Or is there an advantage of using `update`?

src/frontend/src/lib/state/featureFlags.ts
```diff
@@ -0,0 +1,91 @@
...
+export const HARDWARE_KEY_TEST = createFeatureFlagStore(
+  "HARDWARE_KEY_TEST",
+  false,
+);
+
+export default {
+  DOMAIN_COMPATIBILITY,
+  OPENID_AUTHENTICATION,
+  HARDWARE_KEY_TEST,
+} as Record<string, Writable<boolean>>;
```
lmuntaner, line 91: Do we need to default export? Why not use `import * as featureFlags` if we need them all.

Also, is `Writable<boolean>` the right type? They also expose the `getFeatureFlag` method.

src/frontend/src/lib/state/featureFlags.ts
```diff
@@ -0,0 +1,91 @@
...
+    update((oldVal) => {
+      value = oldVal;
+      return oldVal;
+    });
+    return value;
+  };
+
+  // Initialize feature flag object with value from localstorage
+
+  const initializedFeatureFlag: FeatureFlag = new FeatureFlag(
```
lmuntaner, line 44: Do we still need the `FeatureFlag` class? Why not moving all the logic here if we are refactoring it anyway?

It seems that there is duplicated data that we are synchronizing between the instance and the store.

LXIF: There is no duplicated data. The class uses the store as storage.

src/frontend/src/lib/state/featureFlags.ts
LXIF: Corrected the type. Yes, we need the export so we can dynamically check the flags in `hooks.client.ts`.

src/frontend/src/lib/state/featureFlags.ts
LXIF: Having a local value would be duplicating data. This is a leftover from when I initialized the flags inside the store initialization function. Refactoring.

lmuntaner (approved):

src/frontend/src/lib/state/featureFlags.ts
LXIF: We still need some object to set on window.__featureFlags. In principle this could be put inside the store creation function, but having it separate is cleaner imo.

LXIF:
@lmuntaner refactored it to be cleaner now.

lmuntaner (approved):
Thanks, I approved before

PR #3009, opened 2025-04-09
Add name to anchor in storage

sea-snake:
Add name to anchor in storage

# Changes

- Add name to stable anchor
- Add set_name method to anchor, limit name length to max 128 characters (same as OpenID name)
- Add set_name method to anchor management


src/internet_identity/src/storage/anchor/tests.rs
lmuntaner, line 1: Do you have tests for the set_name method?

lmuntaner (commented):
Looking good! Just a question about maybe missing tests

src/internet_identity/src/storage/anchor/tests.rs
sea-snake: Good point, will add it!

PR #3012, opened 2025-04-09
Lookup device key with credential id

sea-snake:
Lookup device key with credential id

# Changes

- Add `lookup_device_key` method that looks up device key and anchor number for a given credential id

# Tests

- Add `should_lookup_device_key_with_credential_id` integration test for the above


lmuntaner (approved):
LGTM!

PR #3015, opened 2025-04-09
Andri/reroute to disco flow

LXIF:
# Motivation

We want to be able to reroute users to the new discoverable passkey flow based on feature flag and query param. The query param and route itself should disappear from the address bar. The new styling should be scoped to only the new routes via a routing group.

# Changes

Add new feature flag, rerouting logic, routing group and test.

# Tests

Added 2 new e2e tests.


src/frontend/src/routes/(new-styling)/new-authenticate/+page.svelte
```diff
@@ -0,0 +1,4 @@
+<!-- an element with id 'newAuthenticateTitle' is necessary for the e2e tests to pass -->
+<div data-role="new-authenticate-view">
+  <h1 class="text-2xl font-bold">Authenticate</h1>
```
lmuntaner, line 3: In Skeleton headings are "h1", "h2" as class: https://www.skeleton.dev/docs/design/typography#headings

src/frontend/tests/e2e/featureFlagRerouting.test.ts
```diff
@@ -0,0 +1,51 @@
+import { runInBrowser } from "./util";
+import { II_URL } from "./constants";
+import { NewAuthenticateView, AuthenticateView } from "./views";
+import { ElementArray } from "webdriverio";
+
+const checkIfHasTailwind = async (styles: ElementArray) => {
+  for (const style of styles) {
+    const text = await style.getHTML();
```
lmuntaner, line 8: There is a Github comment that it fails in this check.

Does it make sense to console log this whole text to understand what they are? The build in local and in CI is different.

src/frontend/src/routes/(new-styling)/new-authenticate/+page.svelte
LXIF: Hadn't looked into that yet as this PR is about the routing and scoping the styling to only this route group - I added the h1 as a placeholder to see that it works. Is it ok if we add the skeleton components/syntax in the PR where we implement the flow or do you want it here already?

src/frontend/tests/e2e/featureFlagRerouting.test.ts
LXIF: Yeah, will probably have to try that.

sea-snake (approved):

PR #3028, opened 2025-04-14
Dark design update

LXIF:
# Motivation

We have been asked to implement a purely dark theme for II.

# Changes

Styling has been adjusted as per Erika's Guerrilla design.

# Tests

No tests have been changed. Also, showcase is broken, so all testing was done manually.


sea-snake (approved):
No functionality seems to have changed besides the removal of the visual warning around the recovery methods section `c-card--warning`, but this doesn't affect the functionality for the end user besides giving it less emphasis from now on.

We should make sure that all flows are visually still functioning e.g. no invisible buttons or labels.
- Create identity
- Sign into II
- Sign into dapp
- VC issue credential

PR #3029, opened 2025-04-14
Update new authorize UX flow

sea-snake:
- Implement Google authentication
- Use dialog element for modal/bottomsheet
- Refactor authentication implementation, only rely on connection class at the very end.

### DIFF
--- a/src/frontend/src/lib/utils/authenticate/jwt.ts
+++ b/src/frontend/src/lib/utils/authenticate/jwt.ts
@@ -1,0 +1,36 @@
+import type { _SERVICE } from "$lib/generated/internet_identity_types";
+import { ActorSubclass, SignIdentity } from "@dfinity/agent";
+import { identityFromActor } from "$lib/utils/authenticate/actor";
+import {
+  throwCanisterError,
+  transformSignedDelegation,
+} from "$lib/utils/utils";
+import { DelegationChain, DelegationIdentity } from "@dfinity/identity";
+
+export const authenticateWithJWT = async ({
+  jwt,
+  salt,
+  actor,
+}: {
+  jwt: string;
+  salt: Uint8Array;
+  actor: ActorSubclass<_SERVICE>;
+}): Promise<{ identity: SignIdentity; anchorNumber: bigint }> => {
+  const identity = await identityFromActor(actor);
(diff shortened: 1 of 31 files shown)


src/frontend/src/lib/utils/authenticate/jwt.ts
```diff
@@ -0,0 +1,36 @@
...
+export const authenticateWithJWT = async ({
+  jwt,
+  salt,
+  actor,
+}: {
+  jwt: string;
+  salt: Uint8Array;
+  actor: ActorSubclass<_SERVICE>;
+}): Promise<{ identity: SignIdentity; anchorNumber: bigint }> => {
+  const identity = await identityFromActor(actor);
```
LXIF, line 19: This is of course fine and how we've been doing it - in the future, this logic could be put into a store so we can just always reference `$actor` and `$identity` - or `get(actor)` / `get(identity)` in this case (or, with some workarounds, it could also be done with state - the logic would be less elegant / self-contained but we could then access it without using `get()`).

src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
```diff
@@ -57,9 +82,145 @@
     showPasskeyCard = true;
   };
 
-  const handleContinueWithGoogle = () => {
-    //TODO
-    console.log("continuing with google");
+  const handleContinueWithGoogle = async () => {
+    const clientId = data.session.config.openid_google?.[0]?.[0]?.client_id;
```
LXIF, line 86: In the future, this too (or the whole config) would probably best be put into store/state. Not vital right now though.

LXIF (approved):
LGTM, fixed your linter warning for you :)

src/frontend/src/lib/utils/authenticate/jwt.ts
sea-snake: Actor is passed in here since it's a stateless utils function, the caller of this method may still grab the actor from a store. Passing in both an actor and identity is unnecessary since the actor already has one within.

sea-snake:
> Forgot to mention - the bottom slider / modal no longer has its internal swiping animation. Of course fine for now, will need to discuss animations with UX to have a coherent language at some point.

Added this back :D

PR #3032, opened 2025-04-15
Add "Manage your Identity" copy on top of the Identity picker

LXIF:
# Motivation

We have been asked to pop the "Manage your Internet Identity" copy on top of the identity picker on mobile. Also, the "Manage your Internet Identity" copy should only appear when visiting ii by itself, not when authenticating with a dapp.

# Changes

Added the necessary functionality.

# Tests

No tests were added.


sea-snake (approved):

src/frontend/src/lib/templates/landingPage.ts
```diff
@@ -16,30 +16,45 @@ import copyJson from "./landingPage.json";
...
   dataPage,
+  landingType,
 }: {
   slot: TemplateResult;
   dataPage: string;
+  landingType?: "firstTime" | "useExisting" | "pick";
 }): TemplateResult => {
   const i18n = new I18n();
   const copy = i18n.i18n(copyJson);
+  const isLanding = window.location.hash.length === 0;
```
lmuntaner, line 27: Will we be keeping the hash in the new flow as well?

How about adding it to a parameter to `landingPage` instead. Maybe we can then use something decoupled from the URL, and we can also create different templates in the showcase with the different scenarios.

However, we are moving away from this code, so I'm not sure we need to improve this...

lmuntaner: I know we need to keep at the beginning, but we could remove it the same way we remove the query params. That's why I was asking.

PR #3040, opened 2025-04-22
Page error handler

sea-snake:
Move page error handling into its own file, the +page.svelte file can then focus on only handling the errors that need to be explicitly handled while the `handleError` method can be used as catch all.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/new-authorize/error.ts
+++ b/src/frontend/src/routes/(new-styling)/new-authorize/error.ts
@@ -1,0 +1,68 @@
+import { isWebAuthnCancel } from "$lib/utils/webAuthnErrorUtils";
+import { toaster } from "$lib/utils/toaster";
+import { isCanisterError } from "$lib/utils/utils";
+import type {
+  CheckCaptchaError,
+  IdRegFinishError,
+  IdRegStartError,
+} from "$lib/generated/internet_identity_types";
+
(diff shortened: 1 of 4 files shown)


src/frontend/src/routes/(new-styling)/new-authorize/error.ts
```diff
@@ -0,0 +1,68 @@
...
+      case "InvalidAuthnMethod":
+      case "StorageError":
+        toaster.error({
+          title: "Something went wrong during registration",
+          description: error.value(error.type),
+        });
+        break;
+      case "AlreadyInProgress":
+      case "WrongSolution":
+        // Errors that should have been handled but weren't
```
LXIF, line 50: Nit: This comment seems a bit misplaced.

LXIF (approved):
LGTM! Nice and centralized handling for future extendability, very cool 😎

src/frontend/src/routes/(new-styling)/new-authorize/error.ts
sea-snake: It's intentional, but maybe not clearly explain. So what I intended to say is that these errors should never be shown to the user, instead they should have been caught and handled earlier in the call stack. Other errors on the other hand are not or sometimes not able to be handled and thus always end up in the `handleError` method.

PR #3043, opened 2025-04-23
Move state and error handling into the page itself.

sea-snake:
Move state and error handling into the page itself.

State: this state is specific to the page itself only anyway.

Error handling: same as above and considering it's calling toast.error(), having it inside a svelte file makes more sense.


sea-snake:
It's an idea, not sure if you're either against or in favor of this change. But wanted to present the idea at least for discussion.

LXIF (approved):
LGTM! Is there a particular reason for having two separate script tags?

sea-snake:
> LGTM! Is there a particular reason for having two separate script tags?

You can't export within a non module script tag

sea-snake:
Closing this PR since after weighing the pros and cons, separate files seems like the better approach for now.

PR #3045, opened 2025-04-23
Continue As supports Google

lmuntaner:
# Motivation

User have an easy way to authenticate with their last authentication method when that one was Google.

# Changes

### Updates to `LastUsedIdentity` type and store logic:

* Added a new `authMethod` field to the `LastUsedIdentity` type, which can be either `"passkey"` or `"google"`.
* Updated the `initLastUsedIdentitiesStore` function to include `authMethod` when adding or updating identities in the store.

### Test updates:

* Modified test cases in `last-used-identities.store.test.ts` to include the `authMethod` field in both input data and expected results.

### Updates to the `new-authorize` page:

* Adjusted the `lastUsedIdentitiesStore.addLatestUsed` call to populate the `authMethod` field based on the presence of a credential ID.
* Updated the logic for continuing authentication to handle different methods (`passkey` or `google`) based on the `authMethod` field.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
@@ -394,14 +394,17 @@
           lastUsedIdentitiesStore.addLatestUsed({
             identityNumber: authenticatedConnection.userNumber,
             name: anchorInfo.name[0],
-            credentialId: credentialId
+            credentialId: nonNullish(credentialId)
               ? new Uint8Array(credentialId)
               : undefined,
+            authMethod: nonNullish(credentialId) ? "passkey" : "google",
           });
           resolve({
             kind: "success",
             delegations: [parsed_signed_delegation],
             userPublicKey: new Uint8Array(userKey),
+            // This is a authnMethod forwarded to the app that requested authorization.
+            // We don't want to leak which authnMethod was used.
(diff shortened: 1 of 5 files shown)


src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
```diff
@@ -411,10 +414,12 @@
...
-                authenticateWithPasskey({
-                  anchorNumber: data.lastUsedIdentity.identityNumber,
-                  credentialId: data.lastUsedIdentity.credentialId,
-                }),
+                data.lastUsedIdentity.authMethod === "passkey"
+                  ? authenticateWithPasskey({
+                      anchorNumber: data.lastUsedIdentity.identityNumber,
+                      credentialId: data.lastUsedIdentity.credentialId,
+                    })
+                  : authenticateWithGoogle(),
```
sea-snake, line 422: I would also store the email of the Google account in the last used storage, so it can be used a loginHint here (requestJWT supports loginHint).

The `decodeJWT` method can be used to grab the email from a JWT string.

lmuntaner: Thanks for the early review. I was wondering how to do that.

LXIF (approved):
LGTM!

PR #3046, opened 2025-04-23
Implement warning if passkeys aren't supported.

sea-snake:
Implement warning if passkeys aren't supported.


lmuntaner (commented):
Can you add a screenshot?

sea-snake:
The color looks off since the background is blue, should be resolved once we have a theme.

image

PR #3047, opened 2025-04-23
Add OpenID credential limit

LXIF:
# Motivation

We want to limit the number of OpenID credentials a user can add to their identity for security purposes.

# Changes

Add the functionality to `anchor.rs` (It's a simple check right now, if it becomes more complicated we can extract it to mirror the `check_device_invariants` function).

# Tests

Add unit test.


sea-snake (approved):

lmuntaner (approved):
Thanks!

PR #3049, opened 2025-04-24
Did an iteration over the design and language of the new screens.

sea-snake:
Did an iteration over the design and language of the new screens.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
@@ -36,7 +36,6 @@
   import SolveCaptcha from "./components/SolveCaptcha.svelte";
   import ContinueAs from "./components/ContinueAs.svelte";
   import Dialog from "$lib/components/UI/Dialog.svelte";
-  import Button from "$lib/components/UI/Button.svelte";
   import { lastUsedIdentitiesStore } from "$lib/stores/last-used-identities.store";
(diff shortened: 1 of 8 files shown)


sea-snake:
image
image
image
image

src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
```diff
@@ -467,21 +470,21 @@
               <p class="text-sm">Please choose another sign-in method</p>
             </div>
           {/if}
-          <Button
+          <button
             onclick={connectOrCreatePasskey}
-            variant="primary"
-            disabled={!supportsPasskeys}>Continue with Passkey</Button
+            class="btn preset-filled py-2"
```
lmuntaner, line 475: Why not use the `Button`?

And I'd rather avoid setting margins in the component level. Instead, it should be the parent setting it.

These limitations should help us have a clear, objective and consistent language with design.

lmuntaner (approved):
LGTM for now.

We should start implementing the new component structure (atoms, etc) at some point. But let's wait until design also uses those atoms.

src/frontend/src/routes/(new-styling)/new-authorize/+page.svelte
sea-snake: As discussed in Slack, I agree, but we need a clear design system first before we can define it in code :D

PR #3050, opened 2025-04-24
Fix dialog on mobile, keyboard was overlapping the passkey name input.

sea-snake:
Fix dialog on mobile, keyboard was overlapping the passkey name input.


LXIF (approved):
LGTM!

PR #3051, opened 2025-04-25
Prevent an anchor from overwriting the credential id lookup of another.

sea-snake:
Prevent an anchor from overwriting the credential id lookup of another.

# Changes

- Added check to see if lookup has already been written

# Tests

- Added test to make sure lookup is not overwritten by another anchor


lmuntaner (approved):
Thanks!

PR #3053, opened 2025-04-25
Prevent an anchor from removing the credential id lookup of another

sea-snake:
Prevent an anchor from removing the credential id lookup of another.

# Changes

- Add check to removal of credential lookup

# Tests

- Updated test to cover this case


lmuntaner (approved):
Thanks!

PR #3058, opened 2025-04-30
First iteration on multiple-accounts frontend.

sea-snake:
First iteration on multiple-accounts frontend.

# Changes

- Update ContinueAs component
- Add PickAccount component
- Significant refactor to accommodate for multiple-account changes  
- Update mock account data to better enable testing frontend during implementation

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -4,8 +4,8 @@
       "type": "custom",
       "candid": "src/internet_identity/internet_identity.did",
       "wasm": "internet_identity.wasm.gz",
-      "build": "bash -c 'II_DEV_CSP=1 II_FETCH_ROOT_KEY=1 II_DUMMY_CAPTCHA=${II_DUMMY_CAPTCHA:-1} scripts/build'",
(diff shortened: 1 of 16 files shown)


dfx.json
lmuntaner, line 1: Can we revert this?

src/frontend/src/routes/(new-styling)/new-authorize/components/ContinueAs.svelte
```diff
@@ -1,22 +1,108 @@
...
+    >Continue</button
   >
 </div>
+{#if showOtherIdentities}
+  <Dialog
+    title={"Switch Internet Identity"}
+    onClose={() => (showOtherIdentities = false)}
+  >
+    <div class="h-4"></div>
+    {#each identities as identity}
```
lmuntaner, line 86: My idea for the last used identities was to also store the accounts (maybe we should rename it). Do we want to show all this here, or only identities? Did you check with design?

sea-snake: This is the dialog/bottom sheet in the low fidelity design to switch between accounts, it doesn't show the accounts within those identities directly.

image

sea-snake:
Last used storage has not been updated yet to keep track of the account in this PR, this is for another PR.

src/frontend/src/routes/(new-styling)/new-authorize/components/ContinueAs.svelte
lmuntaner: Yes, nevermind. I forgot now that the last identities store is indexed by identity number.

I think we should revisit that and store all the accounts there as well.

lmuntaner (approved):
Approving because it shuold all change in the refactor.

PR #3059, opened 2025-04-30
Storage layer for multiple accounts

LXIF:
# Motivation

This is a preliminary draft of the Multi-ACcount IDentity storage implementation.

# Changes

A lot! Created the storage structures and several structs and access methods and updated some of the lookup methods and such. This needs to be refactored though, and is definitely incomplete.

# Tests

Added one unit test, which runs fine.


src/internet_identity/src/storage.rs
```diff
@@ -597,6 +635,247 @@ impl<M: Memory + Clone> Storage<M> {
...
+                let new_application = Application {
+                    origin: origin.to_string(),
+                    total_accounts: 0u64,
+                };
+
+                self.stable_application_memory
+                    .insert(new_number, new_application);
+                Some(new_number)
+            })
+            .expect("This should not happen.")
```
lmuntaner, line 661: Avoid raising or raise something more explainable.

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +635,247 @@ impl<M: Memory + Clone> Storage<M> {
...
+    }
+
+    pub fn lookup_application_number_with_origin(
```
lmuntaner, line 664: Do we need all of these lookups?

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +635,247 @@ impl<M: Memory + Clone> Storage<M> {
...
+    ///   - If yes: Updates the `last_used` timestamp.
+    ///   - If no: Adds a new AccountReference and increments the application's `total_accounts`.
+    /// - Saves the updated StableAnchor and Application data.
+    /// - If the account has a `name` (unreserved account):
+    ///   - Creates a new permanent StorableAccount entry.
+    ///   - Assigns a globally unique AccountNumber.
+    ///   - Returns the Account with the new AccountNumber.
+    /// - Otherwise (reserved account): Returns the Account with `account_number = None`.
+    #[allow(dead_code)]
+    pub fn write_account(&mut self, acc: Account) -> Result<Account, StorageError> {
```
lmuntaner, line 699: See the comments below and agree on a way forward.

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +635,247 @@ impl<M: Memory + Clone> Storage<M> {
...
+            account_number: None,
+            anchor_number,
+            origin: acc.origin.clone(),
+            last_used: acc.last_used,
+            name: None,
+        })
+    }
+
+    #[allow(dead_code)]
+    pub fn get_account_by_id(&self, id: AccountNumber) -> Option<StorableAccount> {
```
lmuntaner, line 804: Do we need this?

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +635,247 @@ impl<M: Memory + Clone> Storage<M> {
...
+                                acc_ref.anchor_number,
+                                application.origin.clone(),
+                                acc_ref.last_used,
+                                Some(storable_acc.name),
+                            )
+                        });
+                }
+                None => {
+                    let anchor = self.read(acc_ref.anchor_number).unwrap();
+                    // TODO LM: Improve this part by looking up at the same time the application and the number
```
lmuntaner, line 833: Can we avoid the two lookups?

src/internet_identity/src/storage/tests.rs
```diff
@@ -316,6 +317,129 @@ fn should_not_overwrite_device_credential_lookup() {
     );
 }
 
+#[test]
+fn should_write_account() {
```
lmuntaner, line 321: Pending to add more tests for different edge cases.

lmuntaner:
@sea-snake this is finally ready for review. On the other hand, it includes all the planned functionality in the storage layer.

I finally decided to create a different struct for each method.

Here is the comparison table:
Screenshot 2025-05-07 at 17 10 20

This can be improved, but adding extra parameters or making them optional when we don't want them optional made the methods more complex and I think those should be handled gracefully in the business logic, not at the storage layer.

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +680,331 @@ impl<M: Memory + Clone> Storage<M> {
...
+    pub fn lookup_application_with_application_number(
+        &self,
+        application_number: &ApplicationNumber,
+    ) -> Option<Application> {
+        self.stable_application_memory.get(application_number)
+    }
+
+    // Add 1 to avoid having an account number 0
+    fn get_next_account_number(&self) -> AccountNumber {
+        AccountNumber::from(self.stable_account_memory.len() + 1)
```
sea-snake, line 737: We'll need a separate counter, else there's a risk of re-using an account number that was previously deleted. Considering that delegations are based on these numbers, we can't re-use deleted account numbers.

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +680,331 @@ impl<M: Memory + Clone> Storage<M> {
...
+        self.lookup_application_with_origin_memory
+            .get(&OriginHash::from_origin(origin))
+    }
+
+    pub fn lookup_account_references(
+        &self,
+        anchor_number: AnchorNumber,
+        application_number: ApplicationNumber,
+    ) -> Option<Vec<AccountReference>> {
+        self.stable_account_reference_list_memory
```
sea-snake, line 723: ~So we're moving the automatic return of a default account to a higher layer outside of storage?~

Edit: seems to be used internally in list_accounts, this method should probably be made non public?

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +680,331 @@ impl<M: Memory + Clone> Storage<M> {
...
+                Some(refs) => Ok(refs),
+            },
+        }
+    }
+
+    #[allow(dead_code)]
+    /// Returns the requested account.
+    /// If the account number doesn't esist, returns a default Account.
+    /// If the account number exists but the account doesn't exist, returns None.
+    /// If the account exists, returns it as Account.
```
sea-snake, line 863: Comment didn't match implementation, see below suggestion:

```suggestion
    /// If the application doesn't exist, returns a default Account.
    /// If the application exists but the account doesn't exist, returns None.
    /// If the account exists, returns it as Account.
```

src/internet_identity/src/storage.rs
```diff
@@ -245,6 +279,31 @@ struct Header {
     salt: [u8; 32],
 }
 
+pub struct CreateAccountParams {
```
sea-snake, line 282: These structs should probably be moved into a separate file?

src/internet_identity/src/storage.rs
```diff
@@ -245,6 +279,31 @@ struct Header {
...
+    anchor_number: AnchorNumber,
+    name: String,
+    origin: FrontendHostname,
+}
+
+pub struct UpdateAccountParams {
```
sea-snake, line 288: I would add an optional last used here, since it probably shouldn't be visible outside the storage layer that this property is stored differently from the other properties.

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +680,331 @@ impl<M: Memory + Clone> Storage<M> {
...
+        };
+        self.stable_account_memory
+            .insert(account_number, storable_account);
+
+        // Update application data
+        let app_num = self.lookup_or_insert_application_number_with_origin(origin);
+        let mut application_data = self
+            .lookup_application_with_application_number(&app_num)
+            .expect("Application data should exist after lookup_or_insert");
+        application_data.stored_accounts += 1;
```
sea-snake, line 787: This should be +2 if both a default and additional account are created.

sea-snake (commented):
I don't see a account limit and/or counter that keeps track of the number of stored default/additional accounts. 

Given that we'd need to read all account references (with a range) every time we'd want to check this, should we consider two separate counters in stable memory for this?

src/internet_identity/src/storage.rs
lmuntaner: Why do you say it doesn't match? This is the read_account, which checks the account_number. If None, returns a default account. Otherwise, it looks it up in memory.

src/internet_identity/src/storage.rs
lmuntaner: Let's handle last_used in a separate PR.

src/internet_identity/src/storage.rs
lmuntaner: Yes, good point. I moved them to the storage/accounts.rs file

src/internet_identity/src/storage.rs
lmuntaner: One stored account, two account references, indeed.

lmuntaner:
@sea-snake changes done and passing pipeline green.

Please take a look, thanks!

src/internet_identity/src/storage.rs
sea-snake:
```
/// Line 933-942: If the application doesn't exist, returns a default Account.
/// Line 944: If the application exists but the account doesn't exist, returns None.
/// Line 945-950: If the account exists, returns it as Account.
```

src/internet_identity/src/storage.rs
```diff
@@ -146,6 +157,12 @@ const REGISTRATION_CURRENT_RATE_MEMORY_INDEX: u8 = 6u8;
 const STABLE_ANCHOR_MEMORY_INDEX: u8 = 7u8;
 const LOOKUP_ANCHOR_WITH_OPENID_CREDENTIAL_MEMORY_INDEX: u8 = 8u8;
 const LOOKUP_ANCHOR_WITH_DEVICE_CREDENTIAL_MEMORY_INDEX: u8 = 9u8;
+const STABLE_ACCOUNT_MEMORY_INDEX: u8 = 10u8;
+const STABLE_APPLICATION_MEMORY_INDEX: u8 = 11u8;
+const LOOKUP_APPLICATION_WITH_ORIGIN_MEMORY_INDEX: u8 = 12u8;
+const STABLE_ACCOUNT_REFERENCE_LIST_MEMORY_INDEX: u8 = 13u8;
+const STABLE_ANCHOR_ACCOUNT_COUNTER_MEMORY_INDEX: u8 = 14u8;
+const STABLE_ACCOUNT_COUNTER_MEMORY_INDEX: u8 = 15u8;
```
sea-snake, line 165: Do we need a global counter besides the per anchor counter? Since we can check the stable account memory entries length already.

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +683,400 @@ impl<M: Memory + Clone> Storage<M> {
...
+                let default_account_reference = AccountReference {
+                    account_number: None,
+                    last_used: None,
+                };
+                self.stable_account_reference_list_memory.insert(
+                    (anchor_number, app_num),
+                    vec![default_account_reference, additional_account_reference].into(),
+                );
+                // Two new account references were created.
+                self.update_counters(app_num, anchor_number, AccountType::AccountReference)?;
```
sea-snake, line 876: We should always increase counters before inserting, increasing it without inserting an account due to an error is less critical than inserting an account and not increasing it due to an error.

lmuntaner: Good point, done!

sea-snake (approved):

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +683,400 @@ impl<M: Memory + Clone> Storage<M> {
...
+            .range(range_start..=range_end)
+        {
+            let storable_refs_vec: Vec<AccountReference> = storable_account_ref_list_val.into();
+
+            for storable_ref in storable_refs_vec {
+                all_accounts.push(AccountReference {
+                    account_number: storable_ref.account_number,
+                    last_used: storable_ref.last_used,
+                });
+            }
```
jasonz-dfinity, line 825: Nit: you can probably use `flat_map` to achieve the same thing

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +683,400 @@ impl<M: Memory + Clone> Storage<M> {
...
+                        break;
+                    }
+                }
+
+                // This could happen if the account was removed and now we try to update it.
+                if !found_and_updated {
+                    return Err(StorageError::MissingAccount {
+                        anchor_number: params.anchor_number,
+                        name: params.name.clone(),
+                    });
```
jasonz-dfinity, line 1070: At this exit point, is it OK that some mutations have happened? (`lookup_or_insert_application_number_with_origin` and `update_counters`

src/internet_identity/src/storage.rs
```diff
@@ -597,6 +683,400 @@ impl<M: Memory + Clone> Storage<M> {
...
+                let mut refs_vec: Vec<AccountReference> = existing_storable_list.into();
+                let mut found_and_updated = false;
+                for r_mut in refs_vec.iter_mut() {
+                    if r_mut.account_number.is_none() {
+                        // Found the default account reference.
+                        r_mut.account_number = Some(new_account_number);
+                        found_and_updated = true;
+                        break;
+                    }
+                }
```
jasonz-dfinity, line 1063: If you do `.iter_mut().find(...)`, I think it will be more apparent that no mutations have happened in the error case

src/internet_identity/src/storage.rs
lmuntaner: Yes, we will be adding a check to fix the counters in case they run out of sync.

And one more application isn't so bad either, it's just a bit more memory.

PR #3062, opened 2025-05-02
Refactor single authorize page into multiple pages and stores.

sea-snake:
Refactor single authorize page into multiple pages and stores.

# Changes

- Create session, authentication and authorization stores
- Moves initialization of there's store to init hook
- Read canister id and config only once in the init hook and then export in $lib/globals
- Additionally export anonymousAgent and anonymousActor from the $lib/globals export
- Split authorize page into multiple pages that use these stores and globals
- As a result, load functions are now solely being used for what they're intended: loading data
- This means that routes can be e.g. safely invalidated, the session, authentication and authorization are now all in stores

### DIFF
--- a/src/frontend/src/hooks.client.ts
+++ b/src/frontend/src/hooks.client.ts
@@ -1,9 +1,6 @@
 import type { ClientInit } from "@sveltejs/kit";
 import featureFlags from "$lib/state/featureFlags";
-import {
-  authenticatedStore,
-  authenticationStore,
-} from "$lib/stores/authentication.store";
+import { authenticationStore } from "$lib/stores/authentication.store";
 import { sessionStore } from "$lib/stores/session.store";
 import { initGlobals, canisterId, agentOptions } from "$lib/globals";
 


src/frontend/src/lib/stores/authentication.store.ts
```diff
@@ -0,0 +1,77 @@
...
+  return {
+    init: async ({ canisterId, agentOptions }) => {
+      const agent = await HttpAgent.create(agentOptions);
+      const actor = Actor.createActor<_SERVICE>(internet_identity_idl, {
+        agent,
+        canisterId,
+      });
+      store.set({ initialized: { agent, actor } });
+    },
+    isAuthenticated: derived(store, ({ authentication }) =>
```
lmuntaner, line 46: What would happen if you do `$authenticationStore.isAuthenticated`. Do you get a store that we need to subscribe again, or do you get the value within the derived stored?

Is it `$(authenticationStore.isAuthenticated)`, or `($authenticationStore).isAuthenticated`?

src/frontend/src/hooks.ts
```diff
@@ -16,6 +16,6 @@ export const reroute: Reroute = ({ url }) => {
     return "/vc-flow/index";
   }
   if (url.hash === "#authorize") {
-    return get(DISCOVERABLE_PASSKEY_FLOW) ? "/new-authorize" : "/authorize";
+    return get(DISCOVERABLE_PASSKEY_FLOW) ? "/authorize" : "/legacy/authorize";
```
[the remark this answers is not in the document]
sea-snake: Since the new routes will be replacing the old routes, it made more sense to move the old routes all under `/legacy` (if they overlap with new routes). The alternative of prefixing all new routes with `new-` doesn't make sense long term (considering that old routes will stay around for a while).

Also by making them a subroute instead of a prefix, it's easier to group them together, give them a shared layout if needed etc.

Overall, this change doesn't affect the initial routing since that goes to #authorize, but when a users decides to pick a different identity on `/authorize/continue` they see `/authorize` instead of `/new-authorize` in the url bar which is highly preferred.

src/frontend/src/lib/stores/authentication.store.ts
sea-snake: You'll get an error that `$authenticationStore` does not contain a value `isAuthenticated`, since the $ syntax subscribes to the store value. 

Instead you need to grab it from the store first, which is indeed `$(authenticationStore.isAuthenticated)`. Though that syntax didn't work for me in Svelte, had to do the following syntax:
```ts
const { isAuthenticated } = authenticationStore;
// Then use $isAuthenticated in code below
```

lmuntaner: That's also what I thought.

Why don't we do a normal derived store then?

```
const isAuthenticatedStore = derive(authenticationStore, ...);
```

This way we can easily use it in a component `$isAuthenticatedStore`?

src/frontend/src/hooks.ts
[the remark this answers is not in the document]
lmuntaner: It makes sense to start moving things to how we want them to be.

I just would have expected this in another PR, hehe.

lmuntaner (approved):

PR #3066, opened 2025-05-06
Avoid code splitting

lmuntaner:
# Motivation

There is a glitch in Internet Identity where the content appears in a white background and without any style. That's because the CSS is loaded too late.

This glitch also happened in OISY and they fixed by reducing the chunks.

# Changes

* Apply manual chunking to have two chunks. One from the dependencies and one from II code. To be determined whether we could keep the same name of the vendor chunk so that it can use the browser cache.

# Tests

* See screenshots with the chunks folder.


lmuntaner:
Before:
Screenshot 2025-05-06 at 20 28 14

Now:
Screenshot 2025-05-06 at 20 28 57

lmuntaner:
Not working as expected. It needs more research.

PR #3067, opened 2025-05-12
Implement new design (behind flag)

sea-snake:
# Motivation

# Changes

# Tests


LXIF:
Screenshot 2025-05-16 at 10 45 17

You may want to add a name length check in the frontend as well so a user doesn't create a passkey that then gets rejected by the backend (128 characters is hard to reach sanely but feels neater if it gets rejected right away) - that being said, not sure if this needs to be in this PR.

src/frontend/src/lib/components/layout/Footer.svelte
```diff
@@ -0,0 +1,22 @@
+<script lang="ts">
+  import type { HTMLAttributes } from "svelte/elements";
+  import { SOURCE_CODE_URL, SUPPORT_URL } from "$lib/config";
+
+  type Props = HTMLAttributes<HTMLHeadElement>;
```
LXIF, line 5: Nit: Feels weird, [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/footer) types footers as `HTMLElement`, but `HTMLHeadElement` [doesn't appear to have any additional properties](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHeadElement), so this should be entirely inconsequential.

src/frontend/src/lib/components/ui/AuthorizeHeader.svelte
```diff
@@ -0,0 +1,53 @@
...
+  import type { HTMLAttributes } from "svelte/elements";
+  import { GlobeIcon } from "@lucide/svelte";
+
+  type Props = HTMLAttributes<HTMLDivElement> & {
+    origin: string;
+  };
+
+  const { class: className, origin, ...props }: Props = $props();
+
+  const hostname = $derived(new URL(origin).hostname);
```
LXIF, line 15: Technically, `$derived` is only necessary when you need to re-render on a change. Since the hostname and dapp don't change during the login process, you could omit the `$derived` here. Not consequential though.

src/frontend/src/lib/components/ui/Badge.svelte
```diff
@@ -0,0 +1,37 @@
...
+  }: Props = $props();
+</script>
+
+<div
+  {...props}
+  class={[
+    "rounded-full border font-medium",
+    {
+      surface: "border-border-tertiary text-text-secondary bg-bg-primary",
+    }[color],
```
LXIF, line 27: Nit: Assuming this is for future-proofing/consistency? (The object indexing syntax with just a single possible property)

src/frontend/src/lib/components/ui/Identicon.svelte
```diff
@@ -0,0 +1,42 @@
...
+
+  type Props = Omit<SVGAttributes<SVGSVGElement>, "seed"> & {
+    seed: bigint;
+    elements?: number;
+  };
+
+  const { class: className, seed, elements = 4, ...props }: Props = $props();
+
+  const colors = [
+    "#29abe2",
```
LXIF, line 12: Nit: Are these hardcoded for a reason?

src/frontend/src/lib/components/ui/ProgressRing.svelte
```diff
@@ -0,0 +1,83 @@
...
+    r={(clientWidth - strokeWidth) / 2}
+    cx={clientWidth / 2}
+    cy={clientWidth / 2}
+    stroke-dasharray={circumference}
+    stroke-dashoffset={circumference - (circumference / 100) * (value ?? 0)}
+  />
+</svg>
+
+<style>
+  /*noinspection CssUnusedSymbol*/
```
LXIF, line 48: Is this an IDE-specific comment? (these classes are used above, but fine to leave the comments of course if there are issues otherwise)

sea-snake:
> You may want to add a name length check in the frontend as well

I'll add these in another PR, this one keeps growing other wise 😅

src/frontend/src/lib/utils/authentication/passkey.ts
```diff
@@ -40,21 +39,17 @@ export const authenticateWithPasskey = async ({
...
-      publicKey: nonNullish(credentialId)
-        ? {
-            allowCredentials: [{ type: "public-key", id: credentialId }],
-          }
-        : creationOptions([], undefined, undefined),
-    },
+  const passkeyIdentity = DiscoverablePasskeyIdentity.useExisting({
+    credentialId,
     getPublicKey: async (result) => {
+      console.log("lookup_device_key", result.rawId);
```
LXIF, line 45: Is there a reason to leave in these logs?

src/frontend/src/lib/utils/discoverablePasskeyIdentity.ts
```diff
@@ -202,44 +225,26 @@ export class DiscoverablePasskeyIdentity extends SignIdentity {
...
 
   async sign(blob: ArrayBuffer): Promise<Signature> {
-    const credential = await (this.#credentialCreationOptions &&
-    !this.#credentialId
+    const credential = await (nonNullish(this.#credentialCreationOptions)
       ? navigator.credentials.create({
           ...this.#credentialCreationOptions,
           publicKey: {
-            ...this.#credentialCreationOptions.publicKey,
+            ...this.#credentialCreationOptions!.publicKey,
```
LXIF, line 232: Nit: This `!` appears unnecessary, since you already checked `nonNullish(this.#credentialCreationOptions)`

LXIF (approved):
LGTM, only nits

src/frontend/src/lib/components/layout/Footer.svelte
sea-snake: Ah that's just a mistake on my end, will update in next PR.

src/frontend/src/lib/components/ui/Badge.svelte
sea-snake: Yes, there are multiple colors in the II design library.

src/frontend/src/lib/components/ui/Identicon.svelte
sea-snake: Yes, they're the color palette used for identicons, but the component isn't used right now, so I'll remove the component in the next PR.

src/frontend/src/lib/components/ui/ProgressRing.svelte
sea-snake: IDE comments yeah, since it wasn't recognized by the IDE as used.

src/frontend/src/lib/utils/authentication/passkey.ts
sea-snake: Nope, will remove them.

src/frontend/src/lib/utils/discoverablePasskeyIdentity.ts
sea-snake: Sadly, this doesn't work like that in this case since it's an or check. So added the ! to override it.

src/frontend/src/lib/components/ui/Alert.svelte
```diff
@@ -0,0 +1,94 @@
...
+    direction = "vertical",
+    onClose,
+    ...props
+  }: Props = $props();
+</script>
+
+<div
+  {...props}
+  class={[
+    "bg-bg-primary_alt border-border-tertiary grid grid-cols-[auto_1fr_auto] grid-rows-[auto_auto] items-start gap-4 rounded-xl border p-4",
```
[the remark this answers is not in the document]
sea-snake: As discussed, leaving it as is for now, will discuss at a later time.

src/frontend/src/lib/components/ui/AuthorizeHeader.svelte
sea-snake: Technically yes, but since this is a component prop, it could technically change e.g. when rendered in a component UI library at a future point.

PR #3068, opened 2025-05-13
Add get_accounts endpoint and related management functionality

LXIF:
# Motivation

We need to be able to retrieve accounts.

# Changes

Add `get_accounts` endpoint, add `get_accounts_for_origin` management function, update existing code.

# Tests

Unit tests for get_accounts are added in the create_accounts PR (integration tests for accounts CRUD will be separate PR)

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -30,6 +30,7 @@ use serde_bytes::ByteBuf;
 use std::collections::HashMap;
 use storage::{Salt, Storage};
 
+mod account_management;
 mod anchor_management;
 mod archive;
 mod assets;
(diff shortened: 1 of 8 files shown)


src/internet_identity/src/main.rs
```diff
@@ -298,21 +299,15 @@ fn get_delegation(
...
-            name: Some("Default Mock Account".to_string()),
-        },
-        AccountInfo {
-            account_number: Some(1),
-            origin: "example.com".to_string(),
-            last_used: Some(0u64),
-            name: Some("Additional Mock Account".to_string()),
-        },
-    ]
+fn get_accounts(anchor_number: AnchorNumber, origin: FrontendHostname) -> Vec<AccountInfo> {
```
sea-snake, line 302: This should return a result so it can return an unauthorized error.

src/internet_identity/src/storage/account.rs
```diff
@@ -29,10 +29,10 @@ pub struct UpdateExistinAccountParams {
...
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct ReadAccountParams {
-    pub account_number: Option<AccountNumber>,
-    pub anchor_number: AnchorNumber,
-    pub origin: FrontendHostname,
+pub struct ReadAccountParams<'a> {
+    pub account_number: &'a Option<AccountNumber>,
+    pub anchor_number: &'a AnchorNumber,
+    pub origin: &'a FrontendHostname,
```
sea-snake, line 35: Can we avoid lifecycle annotations here?

LXIF: Parameters for a read function should not be owned, but borrowing them requires lifecycle annotations.

PR #3069, opened 2025-05-13
Add create_account update method and related management function

LXIF:
# Motivation

We want to be able to create accounts. [NOTE: this PR branches off of #3068. That one should be reviewed and merged before this one.]

# Changes

Add create_account_for_origin function in account_management, add it to endpoint, add tests, add `MAX_ANCHOR_ACCOUNTS` const, use it to throw error above max account number and add test for that.

# Tests

Added unit tests for create and get accounts in account_management


src/internet_identity/src/main.rs
```diff
@@ -298,35 +299,25 @@ fn get_delegation(
...
-            name: Some("Default Mock Account".to_string()),
-        },
-        AccountInfo {
-            account_number: Some(1),
-            origin: "example.com".to_string(),
-            last_used: Some(0u64),
-            name: Some("Additional Mock Account".to_string()),
-        },
-    ]
+fn get_accounts(anchor_number: AnchorNumber, origin: FrontendHostname) -> Vec<AccountInfo> {
```
sea-snake, line 302: Should return a result so it can return proper e.g. unauthenticated errors.

LXIF: Given that this is not an expected error and we're using the unauthorized => trap pattern in many other endpoints, is there a specific reason you need this implemented with a `Result`?

src/internet_identity/src/account_management.rs
```diff
@@ -0,0 +1,168 @@
...
+            .collect()
+    })
+}
+
+pub fn create_account_for_origin(
+    anchor_number: AnchorNumber,
+    origin: FrontendHostname,
+    name: String,
+) -> Result<Account, CreateAccountError> {
+    storage_borrow_mut(|storage| {
```
sea-snake, line 33: Can we add a constant `MAX_ANCHOR_ACCOUNTS` and return an error here if the user already reached this limit? 

We could still have an additional limit per dapp, but that's something we could probably do on the frontend as UX while this global limit would be more of an error if they have too many additional accounts in total across dapps.

src/internet_identity/src/main.rs
LXIF: implemented ✅

LXIF:
Come to think of it, I need to add authorization and per-acc authorization as well

LXIF:
Ok, this should work (will need to add integration tests later)

src/internet_identity/src/account_management.rs
```diff
@@ -1,8 +1,30 @@
...
+    AccountNumber, AnchorNumber, CreateAccountError, FrontendHostname,
+};
+
+const MAX_ANCHOR_ACCOUNTS: usize = 500;
+
+#[allow(dead_code)]
+pub fn anchor_has_account(
+    anchor_number: &AnchorNumber,
+    origin: &FrontendHostname,
+    account_number: &Option<AccountNumber>,
```
jasonz-dfinity, line 17: Nit: I often find it somewhat confusing when `Copy` is passed by reference (for `anchor_number` and `account_number`)

jasonz-dfinity (approved):

src/internet_identity/src/account_management.rs
LXIF: Fair point, changing this :)

PR #3070, opened 2025-05-13
Add account updating functionality

LXIF:
# Motivation

We want to be able to update accounts. [NOTE: This branches off of #3069. Please review and merge that one before this one.]

# Changes

Added update endpoint and related account management functions. Added unit tests. Added per-account authorization checking to the update function in account_management.

# Tests

Added unit tests.


src/internet_identity/src/account_management.rs
```diff
@@ -0,0 +1,384 @@
...
+            .create_additional_account(CreateAccountParams {
+                anchor_number,
+                origin,
+                name,
+            })
+            .map_err(|err| CreateAccountError::InternalCanisterError(format!("{}", err)))
+    })
+}
+
+pub fn update_account_for_origin(
```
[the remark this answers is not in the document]
LXIF: Accurate, account limits are separate jira issue though and will be handled in a separate PR later this week. Nice catch though.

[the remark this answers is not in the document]
LXIF: done ✅

src/internet_identity/internet_identity.did
```diff
@@ -561,10 +561,14 @@ type IdentityMetadataReplaceError = variant {
 
 type CreateAccountError = variant {
     InternalCanisterError : text;
+    MaximumAccountNumberReached;
```
sea-snake, line 564: Giving that account numbers are a global increasing number, this might be confusing, maybe `AccountLimitReached`?

src/internet_identity/src/account_management.rs
```diff
@@ -22,3 +48,421 @@ pub fn get_accounts_for_origin(
...
+    anchor_number: AnchorNumber,
+    account_number: Option<AccountNumber>,
+    origin: FrontendHostname,
+    update: AccountUpdate,
+) -> Result<Account, UpdateAccountError> {
+    match update.name {
+        Some(name) => storage_borrow_mut(|storage| {
+            // Check against max if it is a
```
sea-snake, line 85: Is a what? I think this comment is cut short early.

LXIF: done ✅

src/internet_identity/src/main.rs
```diff
@@ -332,17 +332,18 @@ fn create_account(
...
-        origin: "example.com".to_string(),
-        last_used: Some(0u64),
-        name: Some("Default Mock Account".to_string()),
-    })
+    // If the anchor doesn't own this account, we return unauthorized.
+    if account_management::anchor_has_account(anchor_number, &origin, account_number).is_none() {
+        return Err(UpdateAccountError::Unauthorized(caller()));
+    }
+
+    account_management::update_account_for_origin(anchor_number, account_number, origin, update)
```
sea-snake, line 345: This method already accepts a anchor_number + account_number pair, so you'd expect it to throw in here already, avoiding the need for above method call.

On another note, I don't see an authentication check here, I only see a check that checks if an account belongs to an anchor.

LXIF: jfc, I need to make a post-it, this is the second time this happens.

LXIF: About the function arguments, the `update_accounts` in storage does not check whether an account actually belongs to the caller. At some point we do need to check this. I put it in the endpoint because we usually do our auth checks there, but it can of course also be in the account_management method.

sea-snake: Seems like these changes have been made, approving the PR.

sea-snake (approved):

PR #3071, opened 2025-05-14
Remove existing account endpoints

LXIF:
# Motivation

The interface compatibility check failed on the PR that actually implements these beyond dummy state, so I'm removing these here to add them back in later.

# Changes

Removed existing account endpoints

# Tests

-


LXIF:
This did not work.

LXIF:
This did not work.

PR #3075, opened 2025-05-15
Andri/implement create account

LXIF:
# Motivation

In order to be able to authenticate with accounts, we need to first prepare the delegation.

# Changes

Add necessary functions - `calculate_account_delegation seed` (which resolves to a regular anchor number based seed for default accounts and former default accounts), `prepare_account_delegation` which closely mirrors `prepare_delegation` but uses the different seed and checks whether the anchor has access to the account, as well as the `prepare_account_delegation` endpoint in main.rs which checks auth, records activity and returns the delegation, wrapped in a Result for frontend convenience. Also, related structs/methods etc.

# Tests

No additional tests at this time, created Jira tickets for the integration tests.


LXIF:
accidentally opened this for wrong branch.

PR #3076, opened 2025-05-15
Add account delegation preparation

LXIF:
# Motivation

In order to be able to authenticate with accounts, we need to first prepare the delegation.

# Changes

Add necessary functions - `calculate_account_delegation seed` (which resolves to a regular anchor number based seed for default accounts and former default accounts), `prepare_account_delegation` which closely mirrors `prepare_delegation` but uses the different seed and checks whether the anchor has access to the account, as well as the `prepare_account_delegation` endpoint in main.rs which checks auth, records activity and returns the delegation, wrapped in a Result for frontend convenience. Also, related structs/methods etc.

# Tests

No additional tests at this time, created Jira tickets for the integration tests.


src/internet_identity/internet_identity.did
```diff
@@ -571,6 +571,16 @@ type UpdateAccountError = variant {
     Unauthorized : principal;
 };
 
+type AccountDelegationError = variant {
+    InternalCanisterError : text;
+    Unauthorized : principal;
+};
+
+type PrepareAccountDelegationOk = record {
```
sea-snake, line 579: Nit: "ok" variants don't need a suffix, I've only seen a suffix used in the codebase for error variants.

src/internet_identity/src/account_management.rs
```diff
@@ -127,6 +137,94 @@ pub fn update_account_for_origin(
...
 }
 
+/// Create `Hash` used for a delegation that can make calls on behalf of an `Account`.
+/// If the `Account` is a non-stored default account or has a `seed_from_anchor` (and thus is a stored default account),
+/// the respective anchor number will be used as a seed input. Otherwise, the `AccountNumber` is used.
+///
+/// # Arguments
+///
+/// * `account` is the `Account` we're using for this delegation
+fn calculate_account_delegation_seed(account: Account) -> Hash {
```
sea-snake, line 147: Maybe this can be `calculate_seed()` in `impl Account`? That would avoid having to add a public getter for `get_seed_anchor`.

src/internet_identity/internet_identity.did
LXIF: Yeah, I changed it in my next PR (thought it looked strange) - changing this here as well

src/internet_identity/src/account_management.rs
LXIF: Nice idea, let me try that real quick

src/internet_identity/internet_identity.did
LXIF: done ✅

src/internet_identity/src/account_management.rs
LXIF: done ✅

src/internet_identity/src/storage/account.rs
```diff
@@ -159,4 +187,70 @@ impl Account {
...
+    /// the respective anchor number will be used as a seed input. Otherwise, the `AccountNumber` is used.
+    ///
+    /// # Arguments
+    ///
+    /// * `account` is the `Account` we're using for this delegation
+    pub fn calculate_seed(&self) -> Hash {
+        let salt = state::salt();
+        let mut blob: Vec<u8> = vec![];
+        blob.push(salt.len() as u8);
+        blob.extend_from_slice(&salt);
```
sea-snake, line 202: Are these lines still needed? Looks like this is already in the None case below.

src/internet_identity/src/account_management.rs
```diff
@@ -127,6 +134,53 @@ pub fn update_account_for_origin(
     }
 }
 
+pub async fn prepare_account_delegation(
```
sea-snake, line 137: Can we make the existing `prepare_delegation` method just call this method with account_number None?

sea-snake: Also I' missing an authentication check (again 😂)

src/internet_identity/src/main.rs
```diff
@@ -350,14 +351,27 @@ fn update_account(
...
 #[update]
-fn prepare_account_delegation(
-    _anchor_number: AnchorNumber,
-    _origin: FrontendHostname,
-    _account_number: Option<AccountNumber>,
-    _session_key: SessionKey,
-    _max_ttl: Option<u64>,
-) -> (UserKey, Timestamp) {
-    (ByteBuf::new(), 0)
+async fn prepare_account_delegation(
```
sea-snake, line 354: Can we make the existing `prepare_delegation` method call this method with account_number None?

src/internet_identity/src/storage/account.rs
LXIF: done ✅

src/internet_identity/src/main.rs
LXIF: done ✅

src/internet_identity/src/storage.rs
```diff
@@ -735,20 +735,68 @@ impl<M: Memory + Clone> Storage<M> {
             .map(|list| list.into())
     }
 
-    pub fn has_account_reference(
+    fn has_account_reference(
```
sea-snake, line 738: I don't think this method should be changed, the method explicitly is called `_reference` and non public. Instead  the public `anchor_has_account` should handle the default account logic instead. This is probably also why tests are failing now, since some other methods likely depend on `has_account_reference` being explicitly about "reserved" accounts.

src/internet_identity/internet_identity.did
lmuntaner, line 1: I think that you need to manually recreate the types. The task that used to do it got broken somehow, which we should think whether we want to fix or remove.

src/internet_identity/src/account_management.rs
```diff
@@ -82,49 +74,105 @@ pub fn update_account_for_origin(
...
-                } = storage.get_account_counter(anchor_number);
-
-                if stored_accounts >= MAX_ANCHOR_ACCOUNTS as u64 {
-                    return Err(UpdateAccountError::AccountLimitReached);
+    storage_borrow_mut(|storage| {
+        if storage
+            .anchor_has_account(anchor_number, &origin, account_number)
+            .is_none()
+        {
+            return Err(UpdateAccountError::Unauthorized(caller()));
```
lmuntaner, line 82: Why `Unauthorized` instead of a `NotFound`?

src/internet_identity/src/account_management.rs
```diff
@@ -82,49 +74,105 @@ pub fn update_account_for_origin(
...
+                // Check if whe have reached account limit
+                // Because editing a default account turns it into a stored account
+                if account_number.is_none() {
+                    let AccountsCounter {
+                        stored_accounts,
+                        stored_account_references: _,
+                    } = storage.get_account_counter(anchor_number);
+
+                    if stored_accounts >= MAX_ANCHOR_ACCOUNTS as u64 {
+                        return Err(UpdateAccountError::AccountLimitReached);
```
lmuntaner, line 96: Can you add a TODO to check the accounts for this anchor and reset the counter in case there is a problem? Also create a ticket for this so that we do it within this epic.

src/internet_identity/src/account_management.rs
```diff
@@ -82,49 +74,105 @@ pub fn update_account_for_origin(
...
+            return Err(AccountDelegationError::Unauthorized(caller()));
+        }
+
+        storage
+            .read_account(ReadAccountParams {
+                account_number,
+                anchor_number,
+                origin: &origin,
+            })
+            .ok_or(AccountDelegationError::InternalCanisterError(
```
lmuntaner, line 153: Why not return here the `Unauthorized` error? What's the difference between `anchor_has_account` and `read_account` returning `None`?

src/internet_identity/src/storage/account.rs
```diff
@@ -159,4 +187,65 @@ impl Account {
...
+
+    /// Create `Hash` used for a delegation that can make calls on behalf of an `Account`.
+    /// If the `Account` is a non-stored default account or has a `seed_from_anchor` (and thus is a stored default account),
+    /// the respective anchor number will be used as a seed input. Otherwise, the `AccountNumber` is used.
+    ///
+    /// # Arguments
+    ///
+    /// * `account` is the `Account` we're using for this delegation
+    pub fn calculate_seed(&self) -> Hash {
+        // If this is a non-stored default account, we derive from frontend and anchor
```
lmuntaner, line 199: Is it always "non-stored"? Default accounts might be stored, right?

src/internet_identity/src/storage/account.rs
```diff
@@ -159,4 +187,65 @@ impl Account {
...
+            return delegation::calculate_seed(self.anchor_number, &self.origin);
+        }
+
+        match self.get_seed_anchor() {
+            Some(seed_from_anchor) => {
+                // If this is a stored default account, we derive from frontend and anchor
+                delegation::calculate_seed(seed_from_anchor, &self.origin)
+            }
+            None => {
+                // If this is an added account, we derive from the account number.
```
lmuntaner, line 210: Could we add this in `delegation::calculate_account_seed`? This way we keep the logic of creating seeds in the same place. Wdyt?

src/internet_identity/internet_identity.did
LXIF: Not sure what you mean by that - which types do I need to recreate?

src/internet_identity/src/storage/account.rs
LXIF: @sea-snake thought it should be in `Account`. I don't have much of a preference either way, how should I proceed?

src/internet_identity/src/account_management.rs
LXIF: To me, if you try to log in with an account you do not own, you are not authorized. At the same time, it is of course also not found. I don't have a strong preference, though. Anyhow it is no longer relevant because we now just use `list_accounts`

src/internet_identity/src/account_management.rs
LXIF: done ✅

src/internet_identity/src/account_management.rs
LXIF: Good point

src/internet_identity/src/storage/account.rs
sea-snake: Given that we need private variables (`anchor_from_seed`) within the `Account`, it makes sense to implement it within the `Account`. 

But we could update the `delegation::calculate_seed` method to also be able to create primary account seeds, avoiding having the seed logic in two places (modules), it's just that we shouldn't call it directly then, instead always indirectly through the method in `Account`.

src/internet_identity/internet_identity.did
lmuntaner: When you change the candid file, we must also change the Typescript related files by running `npm run generate`

src/internet_identity/src/account_management.rs
lmuntaner: Good then

src/internet_identity/src/storage/account.rs
lmuntaner: I'm not saying to move all this to the delegation module, only the part of the seed from the account.

I would still keep the match here, but this method could call the delegation module. Similar to above with `delegation::calculate_seed(seed_from_anchor, &self.origin)`.

I just meant to have a match where one calls `delegation::calculate_seed(seed_from_anchor, &self.origin)` while the other branch calls `delegation::calculate_seed_for_account(...)`.

src/internet_identity/src/storage.rs
```diff
@@ -735,21 +735,21 @@ impl<M: Memory + Clone> Storage<M> {
...
```
lmuntaner, line 738: Should we rename this method to `find_account_reference`? It returns an AccountReference while the name `has_...` indicates a boolean.

I probably wrote this method, but I think it's confusing.

src/internet_identity/src/storage/account/tests.rs
lmuntaner, line 1: Can we add two more tests for `read_account` for the two new use cases?

lmuntaner (commented):
Looking good!

src/internet_identity/internet_identity.did
LXIF: ahh right, doing that

src/internet_identity/src/storage.rs
LXIF: Sure!

src/internet_identity/src/storage.rs
```diff
@@ -895,72 +894,136 @@ impl<M: Memory + Clone> Storage<M> {
...
+                    self.lookup_account_references(
+                        params.anchor_number,
+                        application_number.unwrap(),
+                    )
+                {
+                    // if the list exists but is empty, we should still return a synthetic default account
+                    // this should only happen if a named account was created, and then both it and the
+                    // default account references were moved/deleted.
+                    if acc_ref_vec.is_empty() {
+                        return Some(Account::new(
```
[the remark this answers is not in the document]
LXIF: Normally yes, but we do have to guarantee that a user can always log in. If we don't return a synthetic default account here, hypothetically a user could have hit the maximum after deleting a renamed default account. Maybe we should talk about this more though, basically transferring a default account and also having a synthetic default account here means the original identity retains access to the transferred default account.

In any case, this should not happen for now because we cannot currently move or delete accounts.

src/internet_identity/src/storage/account.rs
LXIF: Sure

src/internet_identity/src/storage/account/tests.rs
LXIF: I can add tests for the ownership checks - testing for the case where an account has been moved/removed would require implementing the actual functionality for moving/removing accounts, which I feel would be overkill.

LXIF: done ✅

src/internet_identity/src/storage.rs
LXIF: Done ✅

src/internet_identity/src/storage/account.rs
LXIF: Done ✅

lmuntaner (approved):
Looking good thanks!

Already approving, but please extend the comments. There are some edge cases that we should solve sooner rather than later.

src/internet_identity/src/account_management.rs
```diff
@@ -82,49 +62,90 @@ pub fn update_account_for_origin(
...
-            // Because editing a default account turns it into a stored account
-            if account_number.is_none() {
-                let AccountsCounter {
-                    stored_accounts,
-                    stored_account_references: _,
-                } = storage.get_account_counter(anchor_number);
-
-                if stored_accounts >= MAX_ANCHOR_ACCOUNTS as u64 {
-                    return Err(UpdateAccountError::AccountLimitReached);
+    storage_borrow_mut(|storage| {
```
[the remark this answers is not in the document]
LXIF: You and @sea-snake had asked to put `anchor_has_account` functionality into `read_account`, so I removed it here. Not sure why `storage_borrow_mut` got moved, moving it back now (no need to borrow storage if we're not using it).

src/internet_identity/src/delegation.rs
```diff
@@ -150,6 +119,21 @@ fn calculate_seed(anchor_number: AnchorNumber, frontend: &FrontendHostname) -> H
     hash_bytes(blob)
 }
 
+pub fn calculate_account_seed(account_number: AccountNumber) -> Hash {
```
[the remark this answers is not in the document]
LXIF: done ✅

src/internet_identity/src/storage.rs
```diff
@@ -895,72 +894,136 @@ impl<M: Memory + Clone> Storage<M> {
...
+                    })
+                    .collect(),
             },
         }
     }
 
-    #[allow(dead_code)]
     /// Returns the requested account.
+    /// If the anchor doesn't own this account, returns None.
+    /// If the account is default but has been moved/deleted, returns None.
```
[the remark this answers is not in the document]
LXIF: Right, updating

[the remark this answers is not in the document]
LXIF: It is correct but incomplete - if the default has been moved or deleted but there are other accounts, we return None. If there are no accounts left (all accounts moved/deleted), we return a synthetic default account. Updating comments.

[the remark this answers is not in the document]
LXIF: (I do agree that this is not a great solution - we currently cannot actually get to this state, but we do need to think about this at some point)

src/internet_identity/src/main.rs
```diff
@@ -275,14 +276,23 @@ async fn prepare_delegation(
...
         &ii_domain,
     )
     .await
+    .map(
+        |PrepareAccountDelegation {
+             user_key,
+             timestamp,
+         }| (user_key, timestamp),
+    )
+    .unwrap_or_else(|err| trap(&format!("{:?}", err)))
```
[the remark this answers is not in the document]
lmuntaner: I guess that's fine for now. I was wondering whether you discussed it already.

src/internet_identity/src/storage/account/tests.rs
```diff
@@ -498,3 +496,68 @@ fn should_count_accounts_different_anchors() {
         "Total counters after anchor 2 additional account mismatch"
     );
 }
+
+#[test]
+fn should_read_default_account_with_empty_reference_list() {
```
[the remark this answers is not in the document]
LXIF: done ✅

src/internet_identity/src/storage/account.rs
```diff
@@ -159,4 +204,55 @@ impl Account {
...
+        }
+
+        match self.get_seed_anchor() {
+            Some(seed_from_anchor) => {
+                // If this is a stored default account, we derive from frontend and anchor
+                delegation::calculate_anchor_seed(seed_from_anchor, &self.origin)
+            }
+            None => {
+                // If this is an added account, we derive from the account number.
+                delegation::calculate_account_seed(self.account_number.unwrap())
```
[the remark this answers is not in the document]
lmuntaner: You don't need four branches. Three should be enough and it's maybe clearer the error.

```
(Some(seed_from_anchor), _) => ...
(None, Some(account_number) => delegation::calculate_account_seed(account_number)
(None, None) => trap("Missing both seed anchor and account number in this account")
```

Just a nitpick though. No need to implement it. I like match :)

PR #3079, opened 2025-05-16
Fix feedback from previous design impl PR

sea-snake:
Fix feedback from previous PR.

- Props typing of Footer component
- Add max length to identity name input
   - Limit is set to 64 to make sure that e.g. YubiKeys won't throw an error
   - Storage limit in the canister is meanwhile higher to handle longer names coming from OpenID
- Add max length to account name input, matches storage limit (32).
- Remove console logs
- Avoid the need for `!` in discoverable passkey identity implementation


LXIF (approved):
LGTM!

PR #3081, opened 2025-05-19
Fix Google authentication and related changes based on feedback

sea-snake:
Fix Google authentication and related changes based on feedback

- Update `dfx.json` init args to use Google dev client instead of prod client config
- Use `active` mode with FedCM, this centers the modal and avoids being blocked by a cooldown period
- Add and render `SystemOverlayBackdrop` component in FedCM flow, to align the FedCM UX with Passkey UX
- Store account name after Google registration in last used identity storage.
- Fix error in `Dialog` component, the `updateKeyboardInset` function shouldn't be called on cleanup
- Add disabled style to `RadioCard` component
- Show loader indicator on "Continue with Google" button
- Rename `isWebAuthnCancel` to `isWebAuthnCancelError` to align with other error type method namings
- Trim identity and account name input values
- Preselect first account if user reaches account selection screen from a new sign in
- Throw error if window is closed in `redirectInPopup`, previously the user got stuck after closing the Google sign-in window

### DIFF
--- a/src/frontend/src/lib/components/utils/SystemOverlayBackdrop.svelte
+++ b/src/frontend/src/lib/components/utils/SystemOverlayBackdrop.svelte
@@ -1,0 +1,14 @@
+<script lang="ts">
+  import { fade } from "svelte/transition";
+  import { type HTMLAttributes } from "svelte/elements";
+
+  type Props = HTMLAttributes<HTMLElement>;
+
+  const { class: className, ...props }: Props = $props();
+</script>
+
+<div
+  class={[`bg-bg-overlay/80 fixed inset-0 z-999`, className]}
+  transition:fade={{ duration: 0.2 }}
+  {...props}
+></div>
--- a/src/frontend/src/lib/components/views/PickAuthenticationMethod.svelte
+++ b/src/frontend/src/lib/components/views/PickAuthenticationMethod.svelte
@@ -4,13 +4,25 @@
   import GoogleIcon from "$lib/components/icons/GoogleIcon.svelte";
   import PasskeyIcon from "$lib/components/icons/PasskeyIcon.svelte";
   import Alert from "$lib/components/ui/Alert.svelte";
+  import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
 
   interface Props {
     setupOrUseExistingPasskey: () => void;
-    continueWithGoogle: () => void;
+    continueWithGoogle: () => Promise<void>;
   }
 
   const { setupOrUseExistingPasskey, continueWithGoogle }: Props = $props();
+
+  let googleLoading = $state(false);
+
+  const handleContinueWithGoogle = async () => {
+    googleLoading = true;
+    try {
+      await continueWithGoogle();
+    } finally {
+      googleLoading = false;
+    }
+  };
 
   const supportsPasskeys = nonNullish(window.PublicKeyCredential);
 </script>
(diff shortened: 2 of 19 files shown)


src/frontend/src/lib/components/utils/SystemOverlayBackdrop.svelte
lmuntaner, line 1: Why is this inside `utils`?

src/frontend/src/lib/components/views/PickAuthenticationMethod.svelte
```diff
@@ -26,15 +38,25 @@
...
-    <Button onclick={continueWithGoogle} variant="secondary" size="xl">
-      <GoogleIcon />
-      Continue with Google
+    <Button
+      onclick={handleContinueWithGoogle}
+      variant="secondary"
+      size="xl"
+      disabled={googleLoading}
+    >
+      {#if googleLoading}
```
lmuntaner, line 53: Probably not worth adding a prop to `Button`: `isLoading` and `loadingText`, right?

src/frontend/src/lib/flows/redirect.ts
```diff
@@ -17,20 +17,27 @@ export const callbackFlow = (): Promise<never> => {
 };
 
 export const redirectInPopup = (url: string): Promise<string> => {
-  const callbackPromise = new Promise<string>((resolve) => {
-    const listener = (event: MessageEvent) => {
+  const callbackPromise = new Promise<string>((resolve, reject) => {
+    const closeInterval = setInterval(() => {
```
lmuntaner, line 21: Shall we add why we added this? Is it to check whether there user closed the window?

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
```diff
@@ -28,8 +29,15 @@
   const hostname = $derived(new URL(origin).hostname);
   const dapps = getDapps();
   const dapp = $derived(dapps.find((dapp) => dapp.hasOrigin(origin)));
+  const preselectedAccount = untrack(() =>
+    "preselectAccount" in page.state && page.state.preselectAccount === true
```
lmuntaner, line 33: Isn't `preselectAccount` already a boolean type?

lmuntaner (approved):
Thanks for the fixes!

A few comments, but already approving.

src/frontend/src/lib/components/utils/SystemOverlayBackdrop.svelte
sea-snake: Since it's not a UI component (follow design system), but a "util" component (mimic backdrop from passkey modal). 

Not entirely sure about the component folder structure yet, feel free to think of better ways to organize them.

src/frontend/src/lib/components/views/PickAuthenticationMethod.svelte
sea-snake: I considered it, but considering the simplicity of passing it like children like this, it seemed to make things rather more complicated than simple.

src/frontend/src/lib/flows/redirect.ts
sea-snake: Yes, I'll add a comment.

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
sea-snake: Navigation state is untyped, and here I've only checked if the type contains the property, I haven't checked the type yet (so it's unknown). 

Instead of checking the property type, it seemed more straightforward to just check it's equality to true.

PR #3083, opened 2025-05-20
Add canonical link

lmuntaner:
# Motivation

To avoid id.ai being crawled by bots, we can point them to identity.internetcomputer.org.

This is a simpler solution than telling the bot to ignore the site when it's id.ai with a header or meta tag only for id.ai.

The only problem is that we will also be pointing the testing domains to the canonical URL. But that is probably also correct.

# Changes

* Add canonical link to identity.internetcomputer.org in the app.html

# Tests

Tested locally. It's added.

### DIFF
--- a/src/frontend/src/app.html
+++ b/src/frontend/src/app.html
@@ -8,6 +8,7 @@
       name="viewport"
       content="width=device-width, initial-scale=1, viewport-fit=cover"
     />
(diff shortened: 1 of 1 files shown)


lmuntaner:
> LGTM - not sure if this will actually repel crawlers though.

Not exactly. But adding a noindex has shown very complex for the extra value we can add compared to this. Specially taking into account it's only temporary.

PR #3085, opened 2025-05-20
Implement minicbor for storage

sea-snake:
Implement minicbor for storage.

- Implement Storable{NameOfType} for everything we store, they should only be used within the storage module, public methods should always receive/return non Storable types instead.
- Implement minicbor for these storable types.
- Deprecate memory id 7 and 8.


src/internet_identity/src/storage/storable/fixed_anchor.rs
```diff
@@ -7,12 +7,12 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 
 #[derive(Clone, Debug, Default, CandidType, Deserialize, Eq, PartialEq)]
-pub struct StorableAnchor {
+pub struct StorableFixedAnchor {
```
lmuntaner, line 10: This is just a rename eh, no changes were done to this struct, right?

src/internet_identity/src/storage.rs
```diff
@@ -716,11 +721,14 @@ impl<M: Memory + Clone> Storage<M> {
         origin: &FrontendHostname,
     ) -> Option<ApplicationNumber> {
         self.lookup_application_with_origin_memory
-            .get(&OriginHash::from_origin(origin))
+            .get(&StorableOriginHash::from_origin(origin))
     }
 
     #[allow(dead_code)]
-    pub fn lookup_application_with_origin(&self, origin: &FrontendHostname) -> Option<Application> {
```
lmuntaner, line 723: This is used in only one test, should we add a TODO to remove it so that it doesn't break the rule or being public and returning an internal type?

lmuntaner: Or add an `into` to make it return an `Application` again?

src/internet_identity/src/storage/account/tests.rs
lmuntaner, line 1: Good sign that these tests didn't need to change!!

src/internet_identity/src/storage/storable/openid_credential.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[n(0)]
+    pub iss: Iss,
+    #[n(1)]
+    pub sub: Sub,
+    #[n(2)]
+    pub aud: Aud,
+    #[n(3)]
+    pub last_usage_timestamp: Option<Timestamp>,
+    #[n(4)]
+    pub metadata: HashMap<String, StorableMetadataEntryV2>,
```
lmuntaner, line 23: I forgot we had this. Why do we need this?

lmuntaner (commented):
Looking super good, thanks!!

Will you try deploying this to beta before merging?

I have one anchor with a Google account linked. This is the only thing I would expect to lose.

src/internet_identity/src/storage/storable/fixed_anchor.rs
sea-snake: Just rename to make it more clear compared to the other struct. I'd rather not touch this struct 😅

src/internet_identity/src/storage/storable/openid_credential.rs
sea-snake: Details we decide to store from the JWT token like name, email and photo.

src/internet_identity/src/storage.rs
sea-snake: I removed the `pub`, it's not needed for that test since that's in the storage module scope.

LXIF (approved):
LGTM, assuming that no migration is necessary for these changes.

src/internet_identity/src/storage/anchor.rs
```diff
@@ -128,35 +128,47 @@ impl From<OpenIdCredentialData> for OpenIdCredential {
...
                 devices: anchor.devices,
                 metadata: anchor.metadata,
             },
-            StableAnchor {
-                openid_credentials: anchor.openid_credentials,
+            StorableAnchor {
+                openid_credentials: anchor
+                    .openid_credentials
+                    .iter()
+                    .cloned()
```
jasonz-dfinity, line 142: The cloning looks wasteful since `anchor.openid_credentials` are destroyed afterwards anyway. Is `.into_iter` somehow not possible here?

src/internet_identity/src/storage/anchor.rs
```diff
@@ -128,35 +128,47 @@ impl From<OpenIdCredentialData> for OpenIdCredential {
...
             anchor_number,
             devices: storable_anchor.devices,
             openid_credentials: stable_anchor
                 .clone()
-                .map(|anchor| anchor.openid_credentials)
+                .map(|anchor| {
+                    anchor
+                        .openid_credentials
+                        .iter()
+                        .cloned()
```
jasonz-dfinity, line 168: Same here

PR #3087, opened 2025-05-21
Redirect on load main page if new domain

lmuntaner:
# Motivation

We are not yet launching the new domain's landing page.

Therefore, we agreed to redirect to the current domains when the home is accessed directly.

# Changes

* Add redirection code in the load function of the home page.

# Tests

Tested locally by changing the domain that needs to be redirected to localhost


src/frontend/src/routes/+page.ts
```diff
@@ -0,0 +1,12 @@
+import { redirect, type Load, type LoadEvent } from "@sveltejs/kit";
+
+export const load: Load = (event: LoadEvent) => {
+  // TODO: Remove when we launch landing page in id.ai
+  const hostname = event.url.hostname;
+  if (hostname === "id.ai") {
```
sea-snake, line 6: I think you also need to add a check for the path to be equal to `/`, else it'll also redirect on sub paths e.g. `/authorize`.

lmuntaner: It's not even run when an app send the user here using the old flow. That's sent to `/legacy/authorize`.

What I realize and don't understand, is why the path is changed for `/authorize/continue`, but not for `/legacy/authorize`.

lmuntaner: Apparently, the `reroute` function redirects without changing the URL, but `throw redirect(302, "/authorize/continue")` changes the URL.

lmuntaner: Found the documentation: https://svelte.dev/docs/kit/hooks#Universal-hooks-reroute

```
Using reroute will not change the contents of the browser’s address bar, or the value of event.url.
```

lmuntaner:
@sea-snake ready for another review.

After thinking about it, I think I'd rather keep it specific to the domain instead of more generic. We anyway will remove this code in the coming weeks and I'd rather not have surprises for edge cases we didn't think of.

sea-snake (approved):

PR #3089, opened 2025-05-21
Use new account delegation methods.

sea-snake:
Use new account delegation methods.

- Replace connection fetchDelegation with prepare and get account delegation calls instead.
- Make sure legacy domains (ic0.app) are mapped to new domains (icp0.io).

### DIFF
--- a/src/frontend/src/lib/stores/authorization.store.ts
+++ b/src/frontend/src/lib/stores/authorization.store.ts
@@ -1,14 +1,16 @@
 import { type Readable, derived, writable, get } from "svelte/store";
-import { isNullish } from "@dfinity/utils";
-import type { InternetIdentityInit } from "$lib/generated/internet_identity_types";
-import { Principal } from "@dfinity/principal";
+import { isNullish, nonNullish } from "@dfinity/utils";
+import type { SignedDelegation } from "$lib/generated/internet_identity_types";
 import {
   authenticationProtocol,
   AuthRequest,
 } from "$lib/flows/authorize/postMessageInterface";
 import { authenticatedStore } from "$lib/stores/authentication.store";
-import { Connection } from "$lib/utils/iiConnection";
-import { fetchDelegation } from "$lib/flows/authorize/fetchDelegation";
+import { remapToLegacyDomain } from "$lib/utils/iiConnection";
+import {
+  throwCanisterError,
+  transformSignedDelegation,
+} from "$lib/utils/utils";
 
 export type AuthorizationContext = {
   authRequest: AuthRequest;
@@ -31,10 +33,7 @@ type AuthorizationStore = Readable<{
   context?: AuthorizationContext;
   status: AuthorizationStatus;
 }> & {
-  init: (params: {
-    canisterId: Principal;
-    canisterConfig: InternetIdentityInit;
-  }) => Promise<void>;
+  init: () => Promise<void>;
   authorize: (
     accountNumber: bigint | undefined,
     artificialDelay?: number,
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -52,40 +51,75 @@ let authorize: (
...
+                derivationOrigin,
+                nonNullish(accountNumber) ? [accountNumber] : [],
+                context.authRequest.sessionPublicKey,
+                nonNullish(context.authRequest.maxTimeToLive)
+                  ? [context.authRequest.maxTimeToLive]
+                  : [],
+              )
+              .then(throwCanisterError);
+            let delegation: SignedDelegation;
+            for (let i = 1; i <= 5; i++) {
```
lmuntaner, line 91: Can we add this logic in a separate util for readability?

Or something like:

```
delegation = retry(() => await.actor..., { opts });
```

I find that more readable.

src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -52,40 +51,75 @@ let authorize: (
...
+              } catch {
+                if (i === 5) {
+                  resolve({
+                    kind: "failure",
+                    text: "Couldn't fetch delegation",
+                  });
+                  return;
+                }
+                // Linear backoff
+                await new Promise((resolve) => setTimeout(resolve, i * 1000));
```
lmuntaner, line 112: Ditto on an util for readability:

```
await waitFor(i * 1000);
```

src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -52,40 +51,75 @@ let authorize: (
...
+            },
+            requestOrigin: remapToLegacyDomain(context.requestOrigin),
+          },
+          status: "authenticating",
+        });
         return new Promise((resolve) => {
-          authorize = async (_accountNumber, artificialDelay) => {
+          authorize = async (accountNumber, artificialDelay) => {
             internalStore.set({ context, status: "authorizing" });
             const artificialDelayPromise = new Promise((resolve) =>
```
lmuntaner, line 72: We could use the util here: `const artificialDelayPromise = waitFor(artificialDelay);`

src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -52,40 +51,75 @@ let authorize: (
...
-              publicKey: context.authRequest.sessionPublicKey,
-              maxTimeToLive: context.authRequest.maxTimeToLive,
-            });
-            await artificialDelayPromise;
-            if ("error" in result) {
-              resolve({ kind: "failure", text: "Couldn't fetch delegation" });
-              return;
+            const { identityNumber, actor } = get(authenticatedStore);
+            const derivationOrigin = remapToLegacyDomain(
+              context.authRequest.derivationOrigin ?? context.requestOrigin,
```
lmuntaner, line 77: Why do we need both here and not before? Can't we create one variable to keep the `derivationOrigin` and keep using it?

lmuntaner (commented):
Thanks!! a few comment!

lmuntaner (approved):

src/frontend/src/lib/stores/authorization.store.ts
sea-snake: Done

src/frontend/src/lib/stores/authorization.store.ts
sea-snake: Also done

src/frontend/src/lib/stores/authorization.store.ts
sea-snake: Same here

src/frontend/src/lib/stores/authorization.store.ts
sea-snake: Created `effectiveOrigin` variable in context and using that everywhere. Also added comments to context to explain what each variable is.

PR #3090, opened 2025-05-21
Fix last used account check.

sea-snake:
Fix last used account check.

- Clarify error, "Unreachable" is too vague to easily pinpoint when needed.
- Fix load function, make sure that it checks correctly if a last used identity exists for the given anchor.
- Fallback to default account if last used account is missing for last used identity.


lmuntaner (commented):
a question

src/frontend/src/routes/(new-styling)/authorize/+page.ts
```diff
@@ -3,20 +3,22 @@ import { get } from "svelte/store";
...
+import { authorizationContextStore } from "$lib/stores/authorization.store";
 
 let redirected = false;
 
 export const load: PageLoad = () => {
   if (!redirected) {
     // Only redirect on first visit
+    const context = get(authorizationContextStore);
+    const origin =
+      context.authRequest.derivationOrigin ?? context.requestOrigin;
```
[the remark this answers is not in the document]
sea-snake: I agree that this is getting repetitive. I'd like to refactor/cleanup the authorize flow implementation and this part indeed, something for another PR.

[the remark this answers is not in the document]
sea-snake: As mentioned in the other PR, added `effectiveOrigin` to authorization context that should be used for most use cases.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -2,34 +2,50 @@
...
 <CenterLayout data-page="new-authorize-view">
-  {#if status === "init" || status === "waiting" || status === "validating" || status === "authorizing"}
-    <div class="flex flex-col items-center justify-center gap-4">
-      <ProgressRing class="text-fg-primary size-14" />
-      <p class="text-text-secondary text-lg">
-        {status === "authorizing" ? "Redirecting to the app" : "Loading"}
-      </p>
-    </div>
-  {:else if status === "authenticating"}
+  {#if status === "authenticating" && navigating.to?.url.pathname !== "/authorize/continue"}
```
lmuntaner, line 40: How about the other statuses?

"init", "waiting" and "validating"? Should we show a loading spinner?

And what if "authenticating" and navigating to the URL? I guess that's so short it doesn't make sense, right?

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -2,34 +2,50 @@
...
+  import { goto } from "$app/navigation";
 
   const { children }: LayoutProps = $props();
 
-  const status = $derived($authorizationStatusStore);
+  const { context, status } = $derived($authorizationStore);
+  const lastUsedAccountAvailable = $derived(
+    nonNullish(context) &&
+      nonNullish(
+        Object.values($lastUsedIdentitiesStore).sort(
```
lmuntaner, line 19: Do we need to sort? We care that there is a last used, not which one, right?

sea-snake: We do, without sorting it's possibly not the last used but the one before that. We could make this smarter, look for the first last used with an account for this origin, but that's a bigger UX topic that needs more looking into.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
sea-snake: For now I hid the loading spinner in these initial statuses to avoid the 2~3 ms flicker of it being visible and then hidden again. Ideally, it should never get stuck for long into any of these statuses before going to an error status.

The navigating state is sub milliseconds, long enough to see a flicker of the previous page, too short to show any indicator.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
lmuntaner: So a user if a user has two identities in local storage being used in different applications, they won't see a Continue As in one of the applications?

lmuntaner (approved):

PR #3091, opened 2025-05-21
Implement changes from feedback in storage layer.

sea-snake:
Implement changes from feedback in storage layer.

- Double check and adjust where needed the variables in memories
- Use `into_iter()` to avoid unnecessarily cloning the whole vector


lmuntaner (approved):
Thanks!

PR #3095, opened 2025-05-22
Add prefix and origin to account seed.

sea-snake:
As discussed, this PR adds a prefix and origin to seed calculations of additional accounts, making sure they're tied to a given origin while avoiding collisions with existing primary account seed calculations.


LXIF (approved):
LGTM

PR #3096, opened 2025-05-22
Add account delegation integration tests

LXIF:
# Motivation

We want to make sure that our account delegations have good integration test coverage. Also the API should make sense.

# Changes

Add integration tests. Change `timestamp` to the more correct `expiration` in `PrepareDelegation` struct.

# Tests

Yes!


lmuntaner (approved):
Thanks!!! Nice!

Do we have a test that ensures that open id can also get delegations for different accounts? Not for this PR, just checking and the open_id integration tests don't use accounts. Should we also add those?

LXIF:
@lmuntaner Ok, creating a JIRA ticket and will do.

PR #3098, opened 2025-05-23
Store all last used accounts in localstorage

sea-snake:
Store all last used accounts in localstorage, instead of only the "last" one. This local data can be later used when needed in possible future design changes.

### DIFF
--- a/src/frontend/src/lib/stores/last-used-identities.store.ts
+++ b/src/frontend/src/lib/stores/last-used-identities.store.ts
@@ -37,6 +37,8 @@ type LastUsedIdentitiesStore = Readable<LastUsedIdentities> & {
   reset: () => void;
 };
 
+export const PRIMARY_ACCOUNT_KEY = "primary";
+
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/stores/last-used-identities.store.ts
```diff
@@ -64,7 +66,14 @@ export const initLastUsedIdentitiesStore = (): LastUsedIdentitiesStore => {
         if (isNullish(identity.accounts)) {
           identity.accounts = {};
         }
-        identity.accounts[params.origin] = {
+        if (isNullish(identity.accounts[params.origin])) {
+          identity.accounts[params.origin] = {};
+        }
+        identity.accounts[params.origin][
+          isNullish(params.accountNumber)
+            ? "primary"
```
lmuntaner, line 74: Can you create a constant for this? This way it's more obvious that it's hardcoded and we can explain why.

lmuntaner (approved):
Nice!

Minor comment, already approving.

PR #3099, opened 2025-05-23
Fix update account guard

sea-snake:
Make sure that update account, only updates an account if that account belongs to that anchor.

- Always read account reference first before updating an existing account in storage
- Return the updated account from storage, avoids reading an account and a false results in tests.

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -1064,7 +1064,8 @@ impl<M: Memory + Clone> Storage<M> {
         &mut self,
         params: UpdateExistingAccountParams,
     ) -> Result<Account, StorageError> {
-        // Check if account exists for given anchor number, origin and account number
+        // Check if account reference exists for given anchor number, origin and account number,
+        // if the account refence exists for a given anchor, that means the anchor has access.
         let application_number = self.lookup_application_number_with_origin(&params.origin);
         let account_reference = self
             .find_account_reference(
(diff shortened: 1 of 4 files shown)


src/internet_identity/src/storage.rs
```diff
@@ -1063,20 +1062,45 @@ impl<M: Memory + Clone> Storage<M> {
...
-            Some(storable_account) => {
-                let mut storable_account = storable_account.clone();
-                storable_account.name = params.name;
-                self.stable_account_memory
-                    .insert(params.account_number, storable_account);
-                Ok(params.account_number)
-            }
-        }
+            })?;
+        // Check if account is an existing account (reserved with an account number)
```
lmuntaner, line 1078: It's not checking if is an existing account, it already is, you are in `update_existing_account`. Maybe just check whether it exists.

src/internet_identity/src/storage.rs
```diff
@@ -1063,20 +1062,45 @@ impl<M: Memory + Clone> Storage<M> {
     /// Used in `update_account` to update an existing account.
     fn update_existing_account(
         &mut self,
-        params: UpdateExistinAccountParams,
-    ) -> Result<AccountNumber, StorageError> {
-        match self.stable_account_memory.get(&params.account_number) {
-            None => Err(StorageError::AccountNotFound {
+        params: UpdateExistingAccountParams,
+    ) -> Result<Account, StorageError> {
+        // Check if account exists for given anchor number, origin and account number
```
lmuntaner, line 1067: Exists and the anchor has access to it, right?

And do you mean account reference?

lmuntaner (approved):
Thanks! Just a couple of comments, but already approving.

src/internet_identity/src/storage.rs
sea-snake: Updated comment to clarify.

PR #3103, opened 2025-05-28
Add discrepancy counter and expose as metric

LXIF:
# Motivation

We want to ascertain correctness of our account counters.

# Changes

Implemented a discrepancy counter functionality.

# Tests

Added a unit test, updated existing integration test to ascertain metric presence.

### DIFF
--- a/src/internet_identity/tests/integration/http.rs
+++ b/src/internet_identity/tests/integration/http.rs
@@ -1355,6 +1355,11 @@ fn should_report_total_account_metrics() -> Result<(), CallError> {
         "internet_identity_total_application_count",
         0f64,
     );
+    assert_metric(
+        &initial_metrics,
+        "internet_identity_account_counter_discrepancy_count",
+        0f64,
+    );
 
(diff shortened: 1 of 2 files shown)


src/internet_identity/tests/integration/http.rs
```diff
@@ -1355,6 +1355,11 @@ fn should_report_total_account_metrics() -> Result<(), CallError> {
         "internet_identity_total_application_count",
         0f64,
     );
+    assert_metric(
```
lmuntaner, line 1358: I meant one extra test similar to `should_report_total_account_metrics`, to check that the new metric is also added when such a scenario appears.

lmuntaner (commented):
Thanks! Looking good, one more comment about the http test,

src/internet_identity/tests/integration/http.rs
lmuntaner: Oh, of course. That's true.

In that case the effort and hack is not worth the value. Ignore it.

PR #3108, opened 2025-06-02
Add .windsurf to gitignore

lmuntaner:
# Motivation

Be able to customize windsurf for each developer

# Changes

Add .windsurf to gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .vscode/
 .idea/
 **/*~
+.windsurf/
 
 # Mac OSX temporary files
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 .vscode/
 .idea/
 **/*~
+.windsurf/
```
sea-snake, line 5: Ah, at least the typo isn't where it matters!

lmuntaner: oops, thanks! haha

PR #3110, opened 2025-06-02
Implement E2E tests for all possible new authorization flows.

sea-snake:
Implement E2E tests for all possible new authorization flows.

# Changes

- Implement PlayWright for new authorization flow e2e tests 
- Add pipeline for PlayWright e2e tests
- Implement E2E tests covering all user flows across the new authorize routes.

# Tests

- [x] Added 8 tests across 3 files.
- [x] Manually ran all these tests on both desktop and mobile to visually verify they behaved as expected.


src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
```diff
@@ -109,6 +109,8 @@
       <RadioCard
         onclick={() => (selectedAccountNumber = account.account_number[0])}
         checked={account === selectedAccount}
+        role="radio"
```
lmuntaner: In which html tag are these being added when passed to a Svelte component that renders other components?

Why did you move `role` and `aria-checked` from `RadioCard` to here? Isn't that something we want in all the `RadioCard`s?

src/frontend/tests/e2e/util.ts
```diff
@@ -556,3 +556,59 @@ export const createActor = async (
...
+
+/**
+ * Workaround for Discoverable Passkeys in E2E tests,
+ * current VirtualWebAuth implementation seems to throw
+ * when a discoverable passkey is requested or created.
+ *
+ * So as a workaround, we remove the resident key
+ * parameters from the webauthn request for now.
+ *
+ * Also, we add the `allowCredentials` parameter
```
lmuntaner: Expand why we do that than just saying what we do. That would help us better in the future.

src/frontend/tests/e2e/webDriverPageObjectElement.ts
```diff
@@ -0,0 +1,113 @@
...
+      // the matching role with the `getComputedRole` method.
+      const elements = await (await this.#getElement()).$$(`aria/${name}`);
+      for (const element of elements) {
+        try {
+          if ((await element.getComputedRole()) === role) {
+            return element;
+          } else if (
+            (await element.parentElement().getComputedRole()) === role
+          ) {
+            // `aria/Name of button` unfortunately selects the span within,
```
lmuntaner: What do you mean of `aria/Name of button`? Is it a role button with `aria/{name}`?

It selects it if the button has a span within it, no?

Maybe it's better to give an example of the html structure for this case?

src/frontend/tests/e2e/(new-styling)/authorize/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+import { WebAuthnCredential } from "../../../e2e-setup";
+import { nonNullish } from "@dfinity/utils";
+import { expect } from "vitest";
+
+/**
+ * Authorize with test app
+ * @returns passkey credential
+ */
+export const authorize = async (
+  authenticate: (authenticator: string) => Promise<void>,
```
lmuntaner: What is this `authenticator: string`?

src/frontend/tests/e2e/(new-styling)/authorize/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+import { II_URL, TEST_APP_NICE_URL } from "../../constants";
+import { WebDriverPageObjectElement } from "../../webDriverPageObjectElement";
+import { DemoAppView } from "../../views";
+import { WebAuthnCredential } from "../../../e2e-setup";
+import { nonNullish } from "@dfinity/utils";
+import { expect } from "vitest";
+
+/**
+ * Authorize with test app
+ * @returns passkey credential
```
lmuntaner: Not true anymore, it returns principal and credential 😄

src/frontend/tests/e2e/(new-styling)/authorize/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  // Assert that the user is authenticated
+  const principal = await demoAppView.getPrincipal();
+  expect(principal).not.toBe("");
+
+  return { principal, credential: credentials[0] };
+};
+
+/**
+ * Authorize with test app to create a passkey
+ * @returns passkey credential
```
lmuntaner: ditto

src/frontend/tests/e2e/(new-styling)/authorize/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+// This assumes that they have been successfully dfx-deployed
+import { II_URL, TEST_APP_NICE_URL } from "../../constants";
+import { WebDriverPageObjectElement } from "../../webDriverPageObjectElement";
+import { DemoAppView } from "../../views";
+import { WebAuthnCredential } from "../../../e2e-setup";
+import { nonNullish } from "@dfinity/utils";
+import { expect } from "vitest";
+
+/**
+ * Authorize with test app
```
lmuntaner: Could you expand a bit what the `authenticate` parameter does and why it's there?

src/frontend/tests/e2e/(new-styling)/authorize/continue.test.ts
lmuntaner, line 1: Does it make sense to add a test to check that when using the "Use existing passkey" it's also added to Continue As and not only when registering?

src/frontend/tests/e2e/(new-styling)/authorize/index.test.ts
```diff
@@ -0,0 +1,84 @@
+import { runInBrowser, wipeStorage } from "../../util";
+
+// Read canister ids from the corresponding dfx files.
+// This assumes that they have been successfully dfx-deployed
+import { II_URL } from "../../constants";
+import { WebDriverPageObjectElement } from "../../webDriverPageObjectElement";
+import { expect } from "vitest";
+import { authorize, createPasskeyIdentity } from "./utils";
+
+test("Authorize with a new passkey", async () => {
```
lmuntaner: "with a new passkey" it's not clear it's a new user. How about "Register and authorize".

src/frontend/tests/e2e/(new-styling)/authorize/index.test.ts
lmuntaner, line 1: Should we also check that identities are also stored in the browser?

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
sea-snake: Visually RadioCard might look like a radio item, but from an accessible perspective that's not always the case. For example the "Create additional account" radio item, opens a popup immediately when clicked, which is now how radio items would behave from an accessible perspective.

src/frontend/tests/e2e/(new-styling)/authorize/continue.test.ts
sea-snake: Yes, definitely a good idea!

src/frontend/tests/e2e/(new-styling)/authorize/index.test.ts
sea-snake: Good idea, I'll rephrase things to either register or sign in.

src/frontend/tests/e2e/(new-styling)/authorize/index.test.ts
sea-snake: As discussed, will add a test.

LXIF (approved):
LGTM in general, are the TODOs still TODO or just leftover comments?

PR #3112, opened 2025-06-04
Update spec for accounts

LXIF:
# Motivation

We want our spec to be up to date with recent developments.

# Changes

Add information about accounts and their related management endpoints, add info about account delegation endpoints below the regular delegation endpoints.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -566,12 +566,14 @@ type CreateAccountError = variant {
     InternalCanisterError : text;
     AccountLimitReached;
     Unauthorized : principal;
+    NameTooLong;
 };
 
 type UpdateAccountError = variant {
     InternalCanisterError : text;
     AccountLimitReached;
     Unauthorized : principal;
+    NameTooLong;
 };
 
 type AccountDelegationError = variant {
--- a/src/internet_identity/src/account_management.rs
+++ b/src/internet_identity/src/account_management.rs
@@ -9,8 +9,8 @@ use crate::{
     state::{self, storage_borrow, storage_borrow_mut},
     storage::{
         account::{
-            Account, AccountDelegationError, AccountsCounter, CreateAccountParams,
-            PrepareAccountDelegation, ReadAccountParams, UpdateAccountParams,
+            validate_account_name, Account, AccountDelegationError, AccountsCounter,
+            CreateAccountParams, PrepareAccountDelegation, ReadAccountParams, UpdateAccountParams,
         },
         Storage,
     },
@@ -44,6 +44,7 @@ pub fn create_account_for_origin(
     origin: FrontendHostname,
     name: String,
 ) -> Result<Account, CreateAccountError> {
+    validate_account_name(&name).map_err(Into::<CreateAccountError>::into)?;
     let created_account = storage_borrow_mut(|storage| {
         check_or_rebuild_max_anchor_accounts(
             storage,
@@ -80,6 +81,7 @@ pub fn update_account_for_origin(
 ) -> Result<Account, UpdateAccountError> {
     match update.name {
         Some(new_name) => {
+            validate_account_name(&new_name).map_err(Into::<UpdateAccountError>::into)?;
             let (updated_account, old_account_name) =
                 // Type annotation was necessary for the compiler to infer the correct type
                 storage_borrow_mut(|storage| -> Result<(Account, Option<String>), UpdateAccountError> {
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -104,6 +104,7 @@ use ic_stable_structures::{
 };
 use internet_identity_interface::archive::types::BufferedEntry;
 
+use crate::delegation::check_frontend_length;
 use crate::openid::OpenIdCredentialKey;
 use crate::state::PersistentState;
 use crate::stats::event_stats::AggregationKey;
@@ -929,6 +930,7 @@ impl<M: Memory + Clone> Storage<M> {
         &mut self,
         params: CreateAccountParams,
     ) -> Result<Account, StorageError> {
+        check_frontend_length(&params.origin);
         let anchor_number = params.anchor_number;
         let origin = &params.origin;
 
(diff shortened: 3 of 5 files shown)


LXIF:
FYI: The document has also been autoformatted - Hope that's okay, as the changes are minor.

I added some general info about Accounts as well as all new endpoints, and changed the word "Account" to "Identity" whenever used for an Identity, to prevent confusion.

docs/ii-spec.mdx
```diff
@@ -468,22 +477,130 @@ This call is used to supply a solution to the captcha challenge returned from `i
...
+### Account Management
+
+Internet Identity supports Accounts. Accounts are subordinate entities to an Identity and allow a user to appear to any given dApp as a completely different user/principal while using the same Identity and authentication methods. Accounts are separated per origin - in practice this means that you will have one set of accounts for each dapp. They also each have a unique account number assigned to them.
+
+There are two flavors of Accounts:
+
+- Primary Accounts are not named and always available. They show up as the same principal to any given dApp as your Identity did prior to the introduction of Accounts.
+- Secondary Accounts have custom names and are currently limited to 500 total, across origins. They show up as a new and unique principal to a dApp.
+
+Accounts can be renamed. Renaming a Primary Account turns them into a Secondary Account. Accounts can not currently be deleted.
```
lmuntaner: Does it turn them into secondary? Since we don't have this concept in the code, I wouldn't even mention the part of "turns them into a Secondary Account".

I'd rather just:
> Accounts can be renamed. Accounts can not currently be deleted.

docs/ii-spec.mdx
```diff
@@ -468,22 +477,58 @@ This call is used to supply a solution to the captcha challenge returned from `i
...
+
+**Authorization**: This request must be sent to the canister with `caller` that is the self-authenticating id derived from any of the public keys of devices associated with the user before this call.
+
+The `create_account` method causes the Internet Identity Service backend to create a new Secondary account for the user for a specific origin. This is counted against the total accounts limit.
+
+#### The `update_account` method
+
+**Authorization**: This request must be sent to the canister with `caller` that is the self-authenticating id derived from any of the public keys of devices associated with the user before this call.
+
+The `update_account` method causes the Internet Identity Service backend to update an existing account for the user. Currently, only account names can be updated. Updating the name of a Primary account turns it into a Secondary account and is counted against the total accounts limit. It has no influence on the principal of the account.
```
lmuntaner: Ditto about "turning it into a secondary account". I would just mention that it will start counting against the accounts limit. I wouldn't call it either "total accounts limit", just "accounts limit".

lmuntaner: The part of `"It has no influence on the principal of the account."` is a good point. I would put it in a separate paragraph and be more explicit: calls to "update_account" have no influence on the principal of the account.

docs/ii-spec.mdx
```diff
@@ -497,12 +542,39 @@ The method returns the expiration timestamp of the delegation. This is returned
...
 
 #### The `get_delegation` query method
+
 **Authorization**: This request must be sent to the canister with `caller` that is the self-authenticating id derived from any of the public keys of devices associated with the user before this call.
 
 For a certain amount of time after a call to `prepare_delegation`, a query call to `get_delegation` with the same arguments, plus the timestamp returned from `prepare_delegation`, actually fetches the delegation.
 
 Together with the `UserKey` returned by `prepare_delegation`, the result of this method is used by the Frontend to pass to the client application as per the [client authentication protocol](#client-authentication-protocol).
 
+### Account Delegations
```
lmuntaner: Should this be at the same level as "### Authentication protocol"?

Maybe at the same level as "delegation methods"?

Or should we put the "account delegatoin methods" at the same level as "delegation methods" and add the intro sentence to both methods?

docs/ii-spec.mdx
```diff
@@ -497,12 +542,39 @@ The method returns the expiration timestamp of the delegation. This is returned
...
+
+The `prepare_account_delegation` method causes the Internet Identity Service backend to prepare a delegation from the user identity associated with the given Identity Anchor, Account Number and Client Application Frontend Hostname to the given session key. The account number is an Option parameter. If a Secondary account is to be used, the Option should be `Some(account_number)`. If a Primary account is to be used, the parameter should be `None`. When using a Primary account, this endpoint is functionally equivalent to `prepare_delegation`.
+
+This method returns the user's identity that's associated with the given Client Application Frontend Hostname. By returning this here, and not in the less secure `get_account_delegation` query, we prevent attacks that trick the user into using a wrong identity. It also returns the expiration timestamp of the delegation. This is returned purely so that the client can feed it back to the backend in `get_account_delegation`.
+
+This method needs to be called before the delegation can be fetched using `get_account_delegation`.
```
lmuntaner, line 559: I guess it's obvious, but should we say that only accounts belonging the the `caller` can be used here?

lmuntaner (commented):
Thanks! Looking very good! Just a few comments.

docs/ii-spec.mdx
LXIF: @lmuntaner But they are functionally different, as they don't count against the account limit and don't have an account number or name. Don't we need some way of differentiating them? Or do we want to keep this internal?

docs/ii-spec.mdx
LXIF: Ok, will try rephrasing that.

docs/ii-spec.mdx
LXIF: Currently there's no way of transferring accounts, but good to add it for sure.

docs/ii-spec.mdx
LXIF: They are turned into a stored account with an account number, but they retain their principal because they have the added seed_from_anchor.

docs/ii-spec.mdx
LXIF: Technically, the new account delegations are also part of the authentication protocol

docs/ii-spec.mdx
lmuntaner: It might be worth mentioning that some accounts do not count towards the limit, maybe by saying that some don't have a number and that's how they can differentiate them?

As we discussed, this is for other clients. They don't need to know the details, only what we expose in the API.

docs/ii-spec.mdx
lmuntaner: I agree, they are part. That's why I was asking whether they should be at the same level or inside.

At the moment it's "###", same as `### Authentication protocol`.

LXIF: 🤙🏻

lmuntaner (approved):
Thanks!

PR #3113, opened 2025-06-05
Add OpenID to spec

LXIF:
# Motivation

Our OpenID Implementation is not in the spec yet. We want our spec to be exhaustive at least in terms of the API.

# Changes

Added OpenID info and endpoints to spec.

### DIFF
--- a/docs/ii-spec.mdx
+++ b/docs/ii-spec.mdx
@@ -7,21 +7,21 @@ import IICandidInterface from "!!raw-loader!./internet_identity.did";
 
 This document describes and specifies Internet Identity from various angles and at various levels of abstraction, namely:
 
--   High level goals, requirements and use cases.
+- High level goals, requirements and use cases.
 
--   Overview of the security and identity machinery, including the interplay of identities, keys, and delegations.
+- Overview of the security and identity machinery, including the interplay of identities, keys, and delegations.
 
--   Interface as used by client applications frontends, i.e., our [client authentication protocol](#client-authentication-protocol).
+- Interface as used by client applications frontends, i.e., our [client authentication protocol](#client-authentication-protocol).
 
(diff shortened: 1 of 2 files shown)


docs/ii-spec.mdx
```diff
@@ -497,12 +511,53 @@ The method returns the expiration timestamp of the delegation. This is returned
...
 
+### OpenID Connect Protocol
+
+Internet Identity supports authentication using OpenID Connect credentials. This allows users to authenticate using credentials from supported OpenID Connect providers.
+
+#### The `openid_identity_registration_finish` method
+
+**Authorization**: Any non-anonymous identity can call this
+
+Completes the registration of a new identity using OpenID Connect credentials. This is part of the identity registration flow that starts with `identity_registration_start`. The salt needs to match the salt that was used to anonymize the principal when fetching the JWT from the frontend (in this case an anonymous principal).
```
lmuntaner, line 529: Looking at the spec, I see that it's a bit confusing. 

For example `### Identity management (API V2 only)` is about all APIs for identities, except the delegation ones. And it doesn't have a specific section on Registration. It is also not so obvious that that registration is the passkey registration.

Not sure we want to do this in this PR though...

lmuntaner: The explanation is good enough, at the same level of the other explanations. I wonder how useful this can be to be honest.

But probably not a priority to improve at the moment.

lmuntaner (approved):
Thanks! LGTM!

Once we have the account delegation we could consider shuffling things around to better explain how to use them together.

docs/ii-spec.mdx
LXIF: Maybe after the dashboard is done we can do a bit of an overhaul if it becomes a priority. I initially had more extensive text including description of all the arguments and return values, but it felt bloated by comparison.

PR #3115, opened 2025-06-06
Add name and frontend string length checking to account management functions.

LXIF:
# Motivation

Because of how the account seed is constructed, we need to make sure that the length of the frontend is no longer than 255. Also, it makes sense to limit the length of account names on the backend.

# Changes

Add the `check_frontend_length` function to `create_account_for_origin`, `update_account_for_origin` and `get_accounts_for_origin`. Add the `check_name_length` function (following a slightly different pattern to return an error instead of trapping), to limit the account names to the same length (seems like a good length).

# Tests

No manual tests were done.

### DIFF
--- a/src/internet_identity/src/account_management.rs
+++ b/src/internet_identity/src/account_management.rs
@@ -46,7 +46,7 @@ pub fn create_account_for_origin(
     name: String,
 ) -> Result<Account, CreateAccountError> {
     check_frontend_length(&origin);
-    check_name_length(&name);
+    check_name_length(&name).map_err(Into::<CreateAccountError>::into)?;
     let created_account = storage_borrow_mut(|storage| {
         check_or_rebuild_max_anchor_accounts(
             storage,
@@ -84,7 +84,7 @@ pub fn update_account_for_origin(
     match update.name {
         Some(name) => {
             check_frontend_length(&origin);
-            check_name_length(&name);
+            check_name_length(&name).map_err(Into::<UpdateAccountError>::into)?;
             storage_borrow_mut(|storage| {
                 // If the account to be updated is a default account
                 // Check if whe have reached account limit
--- a/src/internet_identity/src/account_management.rs
+++ b/src/internet_identity/src/account_management.rs
@@ -79,7 +79,6 @@ pub fn update_account_for_origin(
     origin: FrontendHostname,
     update: AccountUpdate,
 ) -> Result<Account, UpdateAccountError> {
-    check_frontend_length(&origin);
     match update.name {
         Some(new_name) => {
             validate_account_name(&new_name).map_err(Into::<UpdateAccountError>::into)?;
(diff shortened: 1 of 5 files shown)


src/internet_identity/src/account_management.rs
```diff
@@ -36,6 +36,7 @@ pub fn get_accounts_for_origin(
     anchor_number: AnchorNumber,
     origin: &FrontendHostname,
 ) -> Vec<Account> {
+    check_frontend_length(origin);
     storage_borrow(|storage| storage.list_accounts(anchor_number, origin))
```
sea-snake: I would recommend moving `check_frontend_length` into the storage layer within `list_accounts`, else this method could still be called somewhere else in the future without this case being guarded.

src/internet_identity/src/account_management.rs
```diff
@@ -44,6 +45,8 @@ pub fn create_account_for_origin(
     origin: FrontendHostname,
     name: String,
 ) -> Result<Account, CreateAccountError> {
+    check_frontend_length(&origin);
+    check_name_length(&name).map_err(Into::<CreateAccountError>::into)?;
```
sea-snake: It's probably more consistent to make both these two method return a "too long" error, then you can either map the error or trap depending on what you intend to do in the place you're calling it. 

If the methods returns a "too long" error variant, it probably makes sense to call the methods "validate_frontend_origin" and "validate_account_name". This way, we can add other validation checks and error variants in the future e.g. "forbidden characters".

LXIF: I really don't want to bloat this PR beyond its own scope, changing legacy patterns is definitely outside. I can rename it though.

LXIF: Renamed ✅

src/internet_identity/src/account_management.rs
LXIF: Done ✅

src/internet_identity/src/account_management.rs
```diff
@@ -78,8 +79,10 @@ pub fn update_account_for_origin(
     origin: FrontendHostname,
     update: AccountUpdate,
 ) -> Result<Account, UpdateAccountError> {
+    check_frontend_length(&origin);
```
sea-snake: This should also be in storage layer, since storing a too long frontend name breaks the delegations while a too long account name doesn't.

sea-snake: Maybe we can move it right where we write the frontend to storage (probably where we create the application id)?

LXIF: Nah, I would check it at the top of the update and create public storage functions - no need to perform storage reads if the inputs are invalid.

LXIF: There ✅

PR #3116, opened 2025-06-06
dedupe read_account in list_accounts

LXIF:
# Motivation

When `list_accounts` is called, it already looks up the application number. It is thus unnecessary to call it again in `read_account` in this specific case.

# Changes

Add an optional `known_app_num` to `ReadAccountParams`. Set it to `None` everywhere but in `list_accounts`

# Tests

No new tests were added.

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -1014,6 +1014,7 @@ impl<M: Memory + Clone> Storage<M> {
                             account_number: acc_ref.account_number,
                             anchor_number,
                             origin,
+                            known_app_num: Some(app_num),
                         })
                     })
                     .collect(),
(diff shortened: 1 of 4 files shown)


src/internet_identity/src/storage.rs
```diff
@@ -1028,8 +1029,13 @@ impl<M: Memory + Clone> Storage<M> {
...
     /// If the `Account` number exists but the `Account` doesn't exist, returns None.
     /// If the `Account` exists, returns it as `Account`.
+    /// Optionally an application number can be passed if it is already known, so we don't look it up more than necessary.
     pub fn read_account(&self, params: ReadAccountParams) -> Option<Account> {
-        let application_number = self.lookup_application_number_with_origin(params.origin);
+        let application_number = if params.known_app_num.is_none() {
+            self.lookup_application_number_with_origin(params.origin)
+        } else {
+            params.known_app_num
+        };
```
sea-snake: Nit, can be simplified to: 
```suggestion
        let application_number = params.known_app_num.unwrap_or_else(|| {
            self.lookup_application_number_with_origin(params.origin)
        });
```

sea-snake (approved):

src/internet_identity/src/storage.rs
LXIF: These are not equivalent / implementing this syntax would throw an error.

Unwrapping `params.known_app_num` results in a `u64`, but `self_lookup_application_number_with_origin(params.origin)` results in an `Option<u64>` (as it should). We need to preserve the possibility to have the lookup return a `None` so we can return synthetic primary accounts if the application number does not exist.

LXIF: (changing it to `or_else` instead)

PR #3120, opened 2025-06-10
Change origin hash to 32 bytes

LXIF:
# Motivation

To address ProdSec finding F01, we are changing the OriginHash to be the full 32-byte 

# Changes

Change the StorableOriginHash to 32 bytes, update its functions accordingly.

# Tests

No new tests were added.

### DIFF
--- a/src/internet_identity/src/storage/storable/application.rs
+++ b/src/internet_identity/src/storage/storable/application.rs
@@ -31,7 +31,7 @@ impl Storable for StorableApplication {
 
 #[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct StorableOriginHash {
-    hash: [u8; 8],
+    hash: [u8; 32],
 }
 
 impl StorableOriginHash {
@@ -39,13 +39,10 @@ impl StorableOriginHash {
         let mut hasher = Sha256::new();
         hasher.update(origin.as_bytes());
         let full_hash_result = hasher.finalize();
-        // Truncate the 32-byte SHA-256 hash to the first 8 bytes.
-        let truncated_hash_slice: &[u8] = &full_hash_result[0..8];
-        let hash_8_bytes: [u8; 8] = truncated_hash_slice
-            .try_into()
-            .expect("Failed to truncate SHA256 hash to 8 bytes; slice length should be 8.");
 
-        Self { hash: hash_8_bytes }
+        Self {
+            hash: full_hash_result.into(),
+        }
     }
 }
 
@@ -61,7 +58,7 @@ impl Storable for StorableOriginHash {
     }
 
     const BOUND: Bound = Bound::Bounded {
-        max_size: 8,
+        max_size: 32,
         is_fixed_size: true,
     };
 }


src/internet_identity/src/storage.rs
```diff
@@ -749,6 +751,19 @@ impl<M: Memory + Clone> Storage<M> {
             .and_then(|application_number| self.stable_application_memory.get(&application_number))
     }
 
+    /// Used for migrating from 8-byte to 32-byte origin hash
+    pub fn rebuild_lookup_application_with_origin_memory(&mut self) {
+        // Check that the memory is empty, otherwise panic
+        assert_eq!(self.lookup_application_with_origin_memory.len(), 0);
```
sea-snake: Given that running below insert is a noop, I don't think we need this guard.

LXIF: I interpret this as meaning that it does not destroy anything - true, I had this in so we definitely don't forget to remove it afterwards, but I can also take out the check.

LXIF: Removed ✅

src/internet_identity/src/storage.rs
```diff
@@ -163,7 +163,9 @@ const REGISTRATION_CURRENT_RATE_MEMORY_INDEX: u8 = 6u8;
 const LOOKUP_ANCHOR_WITH_DEVICE_CREDENTIAL_MEMORY_INDEX: u8 = 9u8;
 const STABLE_ACCOUNT_MEMORY_INDEX: u8 = 10u8;
 const STABLE_APPLICATION_MEMORY_INDEX: u8 = 11u8;
-const LOOKUP_APPLICATION_WITH_ORIGIN_MEMORY_INDEX: u8 = 12u8;
+// This memory index has been abandoned, do not use it
+// const LOOKUP_APPLICATION_WITH_ORIGIN_MEMORY_INDEX: u8 = 12u8;
+const LOOKUP_APPLICATION_WITH_ORIGIN_MEMORY_INDEX: u8 = 19u8;
```
lmuntaner, line 168: Should we have this after the `STABLE_ACCOUNT_COUNTER_DISCREPANCY_COUNTER_MEMORY_INDEX` which is the 18 index?

Otherwise I can see ourselves adding the 19 again...

src/internet_identity/src/storage.rs
```diff
@@ -749,6 +751,16 @@ impl<M: Memory + Clone> Storage<M> {
             .and_then(|application_number| self.stable_application_memory.get(&application_number))
     }
 
+    /// Used for migrating from 8-byte to 32-byte origin hash
+    pub fn rebuild_lookup_application_with_origin_memory(&mut self) {
+        self.stable_application_memory
```
lmuntaner, line 756: Won't this be pointing to the new stable memory with index 19?

We need to keep both accessible until we finish the migration, right?

lmuntaner (commented):
Could we have a short document of how we expect to do this migration.

This is touching the storage layer and it's important because it uses post_upgrade, which if we mess up we are very much screwed.

Therefore, I think that a short design doc explaining what we want to do is time well spent.

src/internet_identity/src/storage.rs
LXIF: You're right. This needs to be split up into two separate PRs.

src/internet_identity/src/storage.rs
LXIF: Sure, why not.

src/internet_identity/src/storage.rs
LXIF: `self.stable_application_memory` points to memory ID 11. We only need to rebuild the lookup table, as it is the only place the origin hash is used. The old lookup table can be discarded.

Rwilliamson832 (approved):

PR #3121, opened 2025-06-10
use checked adds in counters

LXIF:
# Motivation

To address ProdSec finding F08, we should use checked adds on our counters. Realistically, as these are u64, they should never overflow though.

# Changes

Add checked adds to counters.

# Tests

No new tests were added.


lmuntaner (approved):
Thanks!

PR #3123, opened 2025-06-12
Update design based on feedback.

sea-snake:
Update design based on feedback. 

# Changes

- Identity and accounts have been separated, the identity switcher has moved to the top right of the screen.
- The centered card is solely focused on your dapp accounts, phrasing has been updated to add additional clarity in that regard e.g. "Choose your KongSwap account".
- After choosing "Create or use another account", a subtle animation indicates to the user that they've navigated and the dapp logo and url remain prominently visible to retain the context for the end user.
- All "Continue" buttons have been removed, instead things happen immediately after being pressed.

# Tests

- [x] Verified that identity can be created (both passkey and google)
- [x] Verified that identity can be sign in with (both passkey and google)
- [x] Verified identity can be created from the "user another" dialog  (both passkey and google)
- [x] Verified another identity can be signed in with from the "user another" dialog (both passkey and google)
- [x] Verified that accounts can be selected to sign in with
- [x] Verified that accounts can be created to later sign in with


src/frontend/src/lib/stores/current-identity.store.ts
```diff
@@ -0,0 +1,5 @@
+import { get, type Writable, writable } from "svelte/store";
+import { lastUsedIdentityStore } from "$lib/stores/last-used-identities.store";
+
+export const currentIdentityNumberStore: Writable<bigint | undefined> =
+  writable(get(lastUsedIdentityStore)?.identityNumber);
```
lmuntaner: Is this a good pattern?

Why not adding a new field "selected" to the used identities store?

src/frontend/src/lib/components/layout/Header.svelte
```diff
@@ -1,21 +1,96 @@
...
+  );
+  const currentIdentity = $derived(
+    nonNullish($currentIdentityNumberStore)
+      ? lastUsedIdentities.find(
+          (identity) => identity.identityNumber === $currentIdentityNumberStore,
+        )
+      : undefined,
+  );
+
+  let identityPopover = $state(false);
```
lmuntaner: Name is a big confusing. How about "isIdentityPopoverOpen"?

src/frontend/src/lib/components/layout/Header.svelte
```diff
@@ -1,21 +1,96 @@
...
+  let signInDialog = $state(false);
```
lmuntaner: Ditto "isSignInDialogOpen"

src/frontend/src/lib/components/ui/Dialog.svelte
lmuntaner, line 1: Are the changes here necessary?

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
```diff
@@ -0,0 +1,187 @@
...
+      </Button>
+    {:else}
+      <Button onclick={useAnotherIdentity} variant="secondary">
+        <span>Use another identity</span>
+      </Button>
+    {/if}
+  </div>
+  <hr class="border-t-border-tertiary mb-4" />
+  <div class="flex gap-4">
+    {#each links as { href, label }}
```
lmuntaner, line 108: Why these links here? Aren't they already in the footer of the page?

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
lmuntaner, line 1: UX question. Why do we need two screens here?

Screenshot 2025-06-12 at 17 37 37
Screenshot 2025-06-12 at 17 37 42

Why not show directly the Switch identity view?

I understand this two screens from the dashboard where more things might be available in the first screen, but not from the authentication flow.

src/frontend/src/lib/components/layout/AuthPanel.svelte
```diff
@@ -9,11 +11,23 @@
...
+    "sm:bg-bg-secondary sm:border-border-secondary sm:border sm:px-6",
     className,
   ]}
 >
-  {@render children?.()}
+  {#if !("disableNavigationAnimation" in page.state)}
+    {#key page.route.id}
+      <div
+        class="col-start-1 row-start-1 flex flex-col max-sm:flex-1"
+        in:fly={{ x: 200 * (navigating.delta ?? 1), duration: 200, delay: 80 }}
```
lmuntaner, line 24: Super cool animation! 🚀

src/frontend/src/lib/components/ui/Popover.svelte
```diff
@@ -0,0 +1,100 @@
...
+      }
+    };
+    element?.addEventListener("toggle", listener);
+    return () => element?.removeEventListener("toggle", listener);
+  });
+</script>
+
+<svelte:window bind:innerWidth={windowWidth} />
+
+{#if windowWidth >= 480}
```
lmuntaner: Can we use a variable for the 480 to understand what it is?

src/frontend/src/lib/components/views/UseAnotherIdentity.svelte
lmuntaner, line 1: Silly question, but is this being used?

src/frontend/src/lib/stores/last-used-identities.store.ts
```diff
@@ -82,6 +88,46 @@ export const initLastUsedIdentitiesStore = (): LastUsedIdentitiesStore => {
         return lastUsedIdentities;
       });
     },
+    // TODO: Update this method once we store usage timestamps in the canister
```
lmuntaner: Can you also add a todo to add tests for this store?

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
```diff
@@ -1,50 +1,41 @@
...
-  let selectedAccountNumber = $state<bigint | undefined | null>(
-    preselectedAccount,
-  );
-  const selectedAccount = $derived(
-    accounts.find(
-      (account) => account.account_number[0] === selectedAccountNumber,
-    )!,
-  );
-  let dialog = $state(false);
+  let createAccountDialog = $state(false);
```
lmuntaner, line 37: ditto `isCreateAccountDialogOpen`?

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.ts
```diff
@@ -3,12 +3,20 @@ import { authenticatedStore } from "$lib/stores/authentication.store";
...
+import { lastUsedIdentitiesStore } from "$lib/stores/last-used-identities.store";
 
 export const load: PageLoad = async () => {
   const { identityNumber, actor } = get(authenticatedStore);
   const { effectiveOrigin } = get(authorizationContextStore);
-  const accounts = await actor
+  let accounts = await actor
     .get_accounts(identityNumber, effectiveOrigin)
     .then(throwCanisterError);
+  // TODO: currently makes sure that local last used sorting is applied
```
lmuntaner, line 14: What is the todo here?

I like the explanation. But does that mean we'll remove the sync once we have last used in the canister?

src/frontend/src/lib/stores/current-identity.store.ts
sea-snake: Since used identities was written to storage and this isn't, I made it a separate store. 

But you're right, I can just instantiate another writable within the last used store instead.

src/frontend/src/lib/components/layout/Header.svelte
sea-snake: Good point, I'll rename this and others.

src/frontend/src/lib/components/ui/Dialog.svelte
sea-snake: Good catch, they were initially, but not anymore.

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
sea-snake: This was discussed with design, showing the list here immediately would be confusing to the end user (particularly if they have many identities). They would see two lists at once which brings back the confusion between identity and account. By splitting it into two screens, this confusion is reduced since it's more clear what their current identity is.

sea-snake: Also on another note, if there aren't multiple identities, the "Switch identity" button will be labeled "Use another identity" instead, skipping the switch identity screen.

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
sea-snake: This was discussed, ultimately we might remove them from the footer of the page, but not right now yet.

src/frontend/src/lib/components/ui/Popover.svelte
sea-snake: Yes, I'll make a variable.

src/frontend/src/lib/components/views/UseAnotherIdentity.svelte
sea-snake: Yes, it's used within Header.svelte, shown as a dialog when a user click "Use another identity". This component has quite a bit of code duplication with the authorize page, hope to improve upon that in another PR.

src/frontend/src/lib/stores/last-used-identities.store.ts
sea-snake: I'll add the todo, though I wouldn't implement the tests until we've implemented last used account timestamps in the canister. This current implementation is more of a workaround to satisfy "show all accounts sorted by last used" requirement without having the timestamps in the canister yet.

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.ts
sea-snake: Yes, once we have it in the canister, we no longer need the sync since the canister will be the source of truth instead of the local storage.

sea-snake: I'll explicitly mention in the todo that this can be removed once that is the case.

src/frontend/src/lib/components/views/UseAnotherIdentity.svelte
lmuntaner: Ok, thanks!

Add a todo and the issue in Jira, thanks!

PR #3127, opened 2025-06-16
Enable dummy auth through canister config

sea-snake:
Add dummy auth to canister config so it can be enabled in e2e tests that rely on production builds. Additionally, add optional `prompt_for_index` config to make multiple dummy auth identities possible.

# Tests

- [x] Verified that dummy auth is only enabled when either the build flag or canister config is set.
- [x] Verified that dummy auth with `prompt_for_index` allows for multiple dummy auth identities.


LXIF (approved):
LGTM

PR #3133, opened 2025-06-19
Implement sign in page for dashboard

sea-snake:
Implement sign in page for dashboard.

# Changes

- Implement sign in page at /
- Implement placholder "work in progress" page at /manage
- Move sign in with last used into it's own flow so it can be re-used
- Refactor UseAnotherIdentity component into generic AuthDialog component

# Tests

- Added e2e tests for authentication flows


src/frontend/src/hooks.client.ts
```diff
@@ -62,19 +62,7 @@ const maybeSetDiscoverablePasskeyFlowFlag = () => {
   featureFlags.DISCOVERABLE_PASSKEY_FLOW.set(true);
 };
 
-const redirectIfNecessary = (): void => {
```
LXIF, line 65: Oh wow, are we going live on main?

PR #3134, opened 2025-06-23
Implement tooltip, danger button and fixes

sea-snake:
Implement tooltip, danger button and fixes

# Changes

- Implement tooltip component
- Add danger prop to button component
- Fix popover component positioning
- Fix dialog component dismiss on outside click

### DIFF
--- a/src/frontend/src/lib/components/ui/Tooltip.svelte
+++ b/src/frontend/src/lib/components/ui/Tooltip.svelte
@@ -121,7 +121,7 @@
 >
   <div
     class={[
-      "bg-fg-primary relative flex max-w-80 flex-col items-start rounded-lg p-3 shadow-lg",
+      "bg-fg-primary relative flex max-w-80 flex-col items-start rounded-lg p-3 drop-shadow-lg",
       {
         up: {
           start: "origin-[1.6rem_100%]",
--- a/src/frontend/src/lib/components/ui/Tooltip.svelte
+++ b/src/frontend/src/lib/components/ui/Tooltip.svelte
@@ -1,0 +1,226 @@
+<script lang="ts">
+  import type { HTMLAttributes } from "svelte/elements";
+  import { nonNullish } from "@dfinity/utils";
+
+  type Direction = "up" | "right" | "down" | "left";
+  type Align = "start" | "center" | "end";
+
+  type Props = HTMLAttributes<HTMLElement> & {
+    label: string;
+    description?: string;
+    direction?: Direction;
+    align?: Align;
+    distance?: string;
+  };
+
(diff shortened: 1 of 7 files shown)


src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -66,7 +66,6 @@
         <SolveCaptcha {...authFlow.captcha} />
       {:else}
         <h1 class="text-text-primary my-2 self-start text-2xl font-medium">
-          Manage your identity
```
sea-snake, line 69: Seems like a merge conflict mistake in a previous PR where both old and new title were rendered, fixed it here.

src/frontend/src/lib/components/ui/Tooltip.svelte
lmuntaner, line 1: Did you look into the Tooltip that was [implemented in the gix components](https://github.com/dfinity/gix-components/blob/d661ced95fff84a29d02b64a7c73839b87ab59c1/src/lib/components/Tooltip.svelte#L4) already?

That one has already been battle tested by NNS Dapp.

sea-snake: Wasn't aware of that tooltip implementation, at first glance while making a comparison:

- Also uses clientBoundingBox calculations to position it, though it doesn't seem to keep it in sync with its anchor
- Uses z-index instead of the "top layer" with popover, which means it will get clipped by parent boxes which it tries to avoid by moving it to the html body, which is kind of a hack since that disconnects the DOM element from where Svelte expects it to be.
- Doesn't show on focus or touch tap, only on mouseenter/leave, not sure if it works with a keyboard and mobile?

Overall, with modern CSS creating a tooltip should have been trivial and barely require any JS. Unfortunately the CSS Anchor spec isn't supported yet beyond Chrome, thus I had to use clientBoundingBox similarly to the implementation above.

With CSS anchor positioning you'd get many nuanced behavioral details like repositioning on scroll/window resize etc https://developer.chrome.com/blog/anchor-positioning-api

sea-snake: Oh and looking further into the GIX implementation, it doesn't seem to support the direction and align options we have in the II design system. It only seems to support rendering centered above it's anchor.

lmuntaner: I mentioned in case we could get something valuable. That tooltip has been changed and improved a few times.

If you think your implementation covers everything that that one does, we should be good.

LXIF (approved):
LGTM

PR #3138, opened 2025-06-24
Frontend account limit per dapp

sea-snake:
Frontend account limit per dapp, disable button and show tooltip when account limit has been reached.

# Changes

- Disable button when account limit has been reached
- Add tooltip to button when disabled (anchored to the inner FeatureIcon within)
- Add bindable element prop to FeaturedIcon component (so it can be anchored to)
- Add optional anchor prop to Tooltip component (so it can anchor to elements other than its direct child component)
- Fix cursor style on button, only apply it when the button isn't disabled.

# Tests

- Added e2e test "Can't create more than 5 accounts"

# Screenshot

image

### DIFF
--- a/src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
@@ -31,13 +31,13 @@
     ),
   );
 
-  const maxAccountsReached = $derived(accounts.length >= 5);
+  const isAccountLimitReached = $derived(accounts.length >= 5);
   const origin = $derived($authorizationContextStore.requestOrigin);
   const dapps = getDapps();
   const dapp = $derived(dapps.find((dapp) => dapp.hasOrigin(origin)));
 
-  let createAccountDialog = $state(false);
-  let loading = $state(false);
+  let isCreateAccountDialogVisible = $state(false);
+  let isAuthorizing = $state(false);
   let tooltipAnchorRef = $state<HTMLElement>();
 
   const createAccount = async (name: string) => {
@@ -53,12 +53,12 @@
     } catch (error) {
       handleError(error);
     } finally {
-      createAccountDialog = false;
+      isCreateAccountDialogVisible = false;
     }
   };
 
   const continueAs = async (account: AccountInfo) => {
-    loading = true;
+    isAuthorizing = true;
     lastUsedIdentitiesStore.addLastUsedAccount({
       origin: $authorizationContextStore.effectiveOrigin,
       identityNumber: $authenticatedStore.identityNumber,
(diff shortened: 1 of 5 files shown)


lmuntaner:
Could you share a screenshot of how the tooltip looks like?

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
```diff
@@ -103,16 +109,32 @@
...
-        <PlusIcon size="1.25rem" />
-      </FeaturedIcon>
-      <span>Create additional account</span>
-    </ButtonCard>
+    <Tooltip
+      label="Limit reached"
+      description="You’ve reached the maximum of 5 accounts."
+      direction="down"
+      align="start"
+      anchor={tooltipAnchorRef}
```
lmuntaner, line 117: How come that you bind it to the featured icon?

src/frontend/tests/e2e-playwright/authorize/account.spec.ts
```diff
@@ -63,3 +63,25 @@ test("Create and authorize with additional account, then switch back to primary
...
+      await authPage
+        .getByRole("button", { name: "Create additional account" })
+        .click();
+      await authPage.getByLabel("Account name").fill(`Additional account ${i}`);
+      await authPage.getByRole("button", { name: "Create account" }).click();
+    }
+    await expect(
+      authPage.getByRole("button", { name: "Create additional account" }),
+    ).toBeDisabled();
+    await authPage.getByRole("button", { name: "Primary account" }).click();
```
lmuntaner, line 86: Why clicking here if we don't expect anything later anyway?

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
sea-snake: That's how it was designed visually.

src/frontend/tests/e2e-playwright/authorize/account.spec.ts
sea-snake: Because the utility `authorize()` function this is called within will timeout otherwise (it expects the user to authenticate).

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
sea-snake: To clarify, the tooltip is anchored (positioning) to the featured icon but still has it's hover/focus events bound to it's direct children (the whole button).

By default a tooltip is anchored to it's direct children, in this example that's overridden with the anchor prop.

src/frontend/tests/e2e-playwright/authorize/account.spec.ts
lmuntaner: Add a comment then to understand why we need it if we do nothing with it.

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/account/+page.svelte
lmuntaner: Yes, I understood that. I didn't know that was a requirement from design, noted.

lmuntaner (approved):
Thanks!

Could you also share a screenshot of how it looks like when we do UI changes?

PR #3141, opened 2025-06-25
Add Dashboard 2.0

LXIF:
# Motivation

II 2.0 needs a new dashboard.

# Changes

Implement the new dashboard.

# Tests

- [x] Switch between tabs
- [x] Using different authentication methods updates correctly
- [x] Switching identities works and updates correctly

E2e tests will be added in another PR.


src/frontend/src/lib/components/ui/ListItem.svelte
```diff
@@ -0,0 +1,30 @@
...
+  }: {
+    children: Snippet;
+    href?: string;
+    hrefAriaLabel?: string;
+    class?: string;
+  } = $props();
+</script>
+
+{#if href}
+  <a {href} class="contents h-full w-full" aria-label={hrefAriaLabel}>
```
lmuntaner, line 17: Could we put it inside the"li"? It's more semantically correct to put the "li" as a direct child of "ul".

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -0,0 +1,59 @@
...
+  DeviceWithUsage,
+  OpenIdCredential,
+} from "$lib/generated/internet_identity_types";
+import { authenticatedStore } from "./authentication.store";
+import { nonNullish } from "@dfinity/utils";
+
+const fetchIdentityInfo = async () => {
+  let authenticated = get(authenticatedStore);
+
+  return await authenticated.actor.get_anchor_info(
```
lmuntaner: The plan is to use v2 api in another PR?

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  reset = () => {
+    this.name = "";
+    this.devices = [];
+    this.openIdCredentials = [];
+    this.deviceRegistration = undefined;
+    this.loaded = false;
+  };
+
+  constructor() {}
```
lmuntaner: Is this necessary?

src/frontend/src/lib/stores/identity-info.state.svelte.ts
lmuntaner, line 1: Interesting pattern!

As we add more state in the dashbord, we need to agree on what should be the unit for a state class. For example, do we want each class to represent only an API endpoint? Or should we keep all togeter? Or grouped by specific things. How should we update the state after an action, should actions also be part of the class?

Those are just random questions I had, but let's think about it to implement them in a consistent way afterwards.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -0,0 +1,245 @@
...
+    </TabElementGroup>
+  {/snippet}
+
+  {#snippet content()}
+    {#key page.url.pathname}
+      <div
+        bind:this={divRef}
+        in:fly={{
+          y:
+            animationDirection === "up"
```
lmuntaner: How about creating a helper  or state for the `y` and `x` values?  It would be easier to read the render part this way.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -0,0 +1,245 @@
...
+    identityInfo.fetch();
+    await gotoManage();
+    isAuthDialogOpen = false;
+  };
+
+  const handleSwitchIdentity = async (identityNumber: bigint) => {
+    const authLastUsedFlow = new AuthLastUsedFlow();
+    const chosenIdentity =
+      $lastUsedIdentitiesStore.identities[Number(identityNumber)];
+    await authLastUsedFlow.authenticate(chosenIdentity).catch((e) => {
```
lmuntaner: Why this `catch` here if it just throws an error?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.ts
lmuntaner, line 1: lint had something to say in this file 😂

src/frontend/src/lib/components/ui/AccessMethod.svelte
```diff
@@ -0,0 +1,72 @@
...
+    type DeviceWithUsage,
+  } from "$lib/generated/internet_identity_types";
+  import { formatLastUsage } from "$lib/utils/time";
+  import { nonNullish } from "@dfinity/utils";
+  import { fade } from "svelte/transition";
+  import PlaceHolder from "./PlaceHolder.svelte";
+
+  let {
+    accessMethod,
+  }: { accessMethod: DeviceWithUsage | OpenIdCredential | null } = $props();
```
lmuntaner: Is the `null` used only to show a placeholder for the last used access method while the info is loading?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,206 @@
...
+  import Tooltip from "$lib/components/ui/Tooltip.svelte";
+
+  let removableOpenIdCredential = $state<OpenIdCredential | null>(null);
+  let displayAddCredentialDialog = $state(false);
+
+  const totalAccessMethods = $derived<number>(
+    identityInfo.devices.length + identityInfo.openIdCredentials.length,
+  );
+
+  const handleAddGoogle = async () => {
```
lmuntaner: Shuold we add this functinoality to `IdentityInfo` class?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,206 @@
...
+      identityInfo.fetch();
+    } else {
+      identityInfo.openIdCredentials = identityInfo.openIdCredentials.filter(
+        (cred) => !(cred.iss === iss && cred.sub === sub),
+      );
+      throw new Error(Object.keys(await googleAddResult.Err)[0]);
+    }
+  };
+
+  const handleRemoveGoogle = async (credential: OpenIdCredential) => {
```
lmuntaner: Ditto on addint it to `IdentityInfo` class

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,206 @@
...
+        <h4>Manage your passkeys, security keys, and linked accounts.</h4>
+      </div>
+
+      <div>
+        <Button
+          onclick={() => {
+            displayAddCredentialDialog = true;
+          }}
+          class="bg-bg-brand-solid text-text-primary-inversed text-[] top-0 flex w-full items-center justify-center gap-1 rounded-sm px-3.5 py-2 font-semibold md:max-w-fit"
+          >Add <Plus size="1.25rem" /></Button
```
lmuntaner, line 36: Should we name it "Link Google Account" for now?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,206 @@
...
+        >
+      </div>
+    </div>
+    <ul>
+      {#each identityInfo.devices as device}
+        <ListItem>
+          <div class="min-w-8">
+            <AppleIcon />
```
lmuntaner: Why an AppleIcon for the passkeys? Shuoldn't we use the PasskeyIcon?

src/frontend/src/lib/components/ui/ListItem.svelte
lmuntaner, line 1: Should we adapt this ListItem to the Figma component "ListItem"?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,206 @@
...
+            </Button>
+          {/if}
+        </ListItem>
+      {/each}
+    </ul>
+  </Panel>
+</div>
+
+{#if removableOpenIdCredential}
+  <Dialog onClose={() => (removableOpenIdCredential = null)}>
```
lmuntaner: This could be its own component, right?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,206 @@
...
+        variant="primary"
+        danger>Unlink Google Account</Button
+      >
+      <Button variant="tertiary">Keep linked</Button>
+    </div>
+  </Dialog>
+{/if}
+
+{#if displayAddCredentialDialog}
+  <Dialog onClose={() => (displayAddCredentialDialog = false)}>
```
lmuntaner: Ditto its own component

lmuntaner (commented):
Thanks code is looking very good!

I guess that you didn't mean to push all the changes in `try-ii/src` right? Can we add those to gitignore so that it doesn't happen again?

Did you align the "components/ui" new components with design? As I mentioned in one of them, they don't seen to be matching exactly.

I think that they also need to make some decisions in Figma, each of those components in Figma has many variants and we should stick with one when possible.

LXIF:
> Thanks code is looking very good!
> 
> I guess that you didn't mean to push all the changes in `try-ii/src` right? Can we add those to gitignore so that it doesn't happen again?
> 
> Did you align the "components/ui" new components with design? As I mentioned in one of them, they don't seen to be matching exactly.
> 
> I think that they also need to make some decisions in Figma, each of those components in Figma has many variants and we should stick with one when possible.

I removed the cached version of try-ii from my git and then pushed again and it disappeared. It does have a gitignore file, not sure what went wrong there. As for components, I tried to balance alignment with figma and actually building a functional frontend (1-to-1 isn't really possible but am open to suggestions).

src/frontend/src/lib/components/ui/ListItem.svelte
LXIF: Yes, but then it won't react to clicking the entire li. However, with `contents`, this is ignored by screen readers etc.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
LXIF: Nope, leftover. Removed.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
LXIF: Yeah, I wanted to try this out - it's functionally very similar to a custom store but has somewhat more flexibility as it can have multiple separate reactive fields. I think we can keep adding fields to this one as necessary, I don't think it needs to be all separated, unless it is functionally / semantically different.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
LXIF: Done ✅

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
LXIF: I copy-pasted this from sea-snake's implementation, true it doesn't make much sense here.

src/frontend/src/lib/components/ui/AccessMethod.svelte
LXIF: Correct

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: Sure, why not

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: Sure, why not

src/frontend/src/lib/components/ui/ListItem.svelte
lmuntaner: Should we build using `role` attribute instead of ul and li?

This way we can use an anchor tag and keep the semantics.

We did this in [NNS Dapp](https://github.com/dfinity/nns-dapp/blob/61b1aeccf33802aa6925421ff21c34c8b4d0df66/frontend/src/lib/components/ui/ResponsiveTableRow.svelte#L51)

Maybe not for this version, we can talk about this with design. They should be the ones deciding whether this is a table or list items.

If they want aligned columns, it should be a table and we can use Grid CSS and subgrid for it.

If they don't want aligned columns, we can keep it as list items.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: We are using the PasskeyIcon ✅

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: Sure, but why not

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: Sure, why not

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: The button? That would be incongruent with Figma, make the button enormous. I advise against it - adding passkeys is coming very soon.

Screenshot 2025-06-26 at 14 01 39

src/frontend/src/lib/components/ui/ListItem.svelte
LXIF: Can't find any `ListItem` in Figma. Not sure what you mean.

LXIF:
>Let's review this later and along the other components. The idea was always to keep them in sync. That might also mean that Figma needs to adapt.

I had a call with @sea-snake about this - I do think it makes sense to do this best as we can, but right now it's not 1-to-1. He did mention maybe we could implement a storybook.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
lmuntaner: True, that's too big.

But it might be confusing to have "Add" and that trigger the Google link.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,91 @@
...
+  </h2>
+  <Panel>
+    <div class="flex flex-col justify-between gap-5 p-4 md:flex-row">
+      <div>
+        <h3 class="mb-2 text-lg font-semibold">Access methods</h3>
+        <h4>Manage your passkeys, security keys, and linked accounts.</h4>
+      </div>
+
+      <div>
+        <Button
```
lmuntaner, line 31: I understood from Erika that the functionality of the button depends on the state.

For example, if there is already a google account linked, it will show passkey directly.

Do we want to disable the button if the user already has a Google account? Maybe in another PR?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+page.svelte
```diff
@@ -1 +1,119 @@
...
+  import { isNullish, nonNullish } from "@dfinity/utils";
+  import {
+    type OpenIdCredential,
+    type AuthnMethodData,
+  } from "$lib/generated/internet_identity_types";
+  import PlaceHolder from "$lib/components/ui/PlaceHolder.svelte";
+  import { fade } from "svelte/transition";
+  import AccessMethod from "$lib/components/ui/AccessMethod.svelte";
+
+  const getLastUsedAccessMethod = (
```
lmuntaner: Could we create this as an util and add some unit tests for it?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -0,0 +1,91 @@
...
+  import GoogleIcon from "$lib/components/icons/GoogleIcon.svelte";
+  import identityInfo from "$lib/stores/identity-info.state.svelte";
+  import AccessMethod from "$lib/components/ui/AccessMethod.svelte";
+  import PasskeyIcon from "$lib/components/icons/PasskeyIcon.svelte";
+  import RemoveOpenIdCredential from "$lib/components/views/RemoveOpenIdCredential.svelte";
+  import AddOpenIdCredential from "$lib/components/views/AddOpenIdCredential.svelte";
+
+  let displayAddCredentialDialog = $state(false);
+
+  const totalAccessMethods = $derived<number>(
```
lmuntaner: nitpick: adding the derived variable in `identityInfo` directly.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: I'll ask Erika

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: Before the google link there's a selector - currently there's only google in that, but will be extended to passkeys.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: Did we decide to only allow a single google account to be linked?

lmuntaner (approved):
Can you share some videos of the functionality and screenshots?

Only nitpicks, so already approving.

src/frontend/src/lib/components/ui/ListItem.svelte
LXIF: I initially had this with Grid, but it became a bit tedious. We can go back to it if necessary of course.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: Discussing this with Erika - if anything, the CTA on the modal would differ, but for now the "Add" remains unchanged.

lmuntaner: This is what I learned today by reviewing the UX flow to add/remove passkeys. I thought you discussed it with her

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+page.svelte
LXIF: done ✅

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: done ✅

PR #3142, opened 2025-06-26
Migration flow on authentication

lmuntaner:
# Motivation

Users should be able to migrate during the authentication flow.

# Changes

* Create a new "components/wizards" folder and moved the logic of migration steps there.
* I removed the success step and instead forward the user to their destination, I'm assuming that will be the desired UX later.
* Use the new MigrationWizard in the `/migrate` route used for visitors of the landing page.
* Use the new MigrationWizard in the `/authorize/migrate` route used for users during authentication flow.
* Change the URL of the Migrate button in the footer depening on current path.  

# Tests

Test manually the following steps (see video attached)
1. Create an identity in the old flow.
2. Go to an app to authenticate.
3. Click on the "Migrate button"
4. Follow the migration steps.
5. I see a toast that the migration was successful, and I'm redirected back to the application.

Also tested the migration when the user lands in the main page:
1. Create an identity in the old flow.
2. Go to II new flow.
3. Click on the "Migrate button"
4. Follow the migration steps.
5. I see a toast that the migration was successful, and I'm redirected to the manage page.

### DIFF
--- a/src/frontend/src/lib/components/wizards/MigrationWizard.svelte
+++ b/src/frontend/src/lib/components/wizards/MigrationWizard.svelte
@@ -18,30 +18,17 @@
   const migrationFlow = new MigrationFlow();
 
   const handleSubmit = async () => {
+    // Button is disabled if identityNumber is null or undefined so no need to manage that case.
     if (nonNullish(identityNumber)) {
-      try {
-        await migrationFlow
-          .authenticateWithIdentityNumber(BigInt(identityNumber))
-          .catch(handleError);
-      } catch (e) {
-        migrationFlow.view = "enterNumber";
-        console.error(e);
-      }
+      await migrationFlow
+        .authenticateWithIdentityNumber(BigInt(identityNumber))
+        .catch(handleError);
     }
   };
 
   const handleCreate = async (name: string) => {
-    try {
-      if (isNullish(migrationFlow.identityNumber)) {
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/wizards/MigrationWizard.svelte
```diff
@@ -0,0 +1,92 @@
...
+  import { lastUsedIdentitiesStore } from "$lib/stores/last-used-identities.store";
+  import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
+  import { handleError } from "$lib/components/utils/error";
+
+  const { onSuccess }: { onSuccess: () => void } = $props();
+
+  let identityNumber: number | undefined = $state(undefined);
+  const migrationFlow = new MigrationFlow();
+
+  const handleSubmit = async () => {
```
LXIF, line 20: Nit: Since the `MigrationFlow` is generally in control of this wizard, these handlers might also be moved there.

LXIF (approved):
LGTM

src/frontend/src/lib/components/wizards/MigrationWizard.svelte
lmuntaner: Good point, and I realized I don't even need the try/catch here. I remove the try/catch and moved some checks and functionality into the MigrationFlow

PR #3144, opened 2025-06-26
Use RP ID and auth flows when authenticating in migration flow

lmuntaner:
# Motivation

The migration needs to request passkeys from the old domains.

Therefore, we take advantage of the logic we already had for the domains compatibility for the migratino flow.

# Changes

* Add the same logic we had in iiConnection to the MigrationFlow about keepint authFlows and finding the RP ID based on the authenticators of the user.

# Tests

Deployed to beta domains.
Manually tested:
1. Pick an identity created in the old domain.
2. Clicked on "Migrate"
3. Followed the steps
4. I was successfully redirected to the dashboard in progress with a success toast.

See video for more details.

Also tested the retry of another RP ID:
1. Create an identity in beta.identity.internetcomputer.org
2. Log in with that identity in beta.identity.ic0.app
3. Add a ledger device as passkey (this means I'll have two passkeys in two different domains).
4. Log in with the ledger device in beta.identity.ic0.app (this will make that passkey the last used).
5. Disconnect the ledger device from the computer (now I don't have access to that passkey)
6. Go to beta.id.ai
7. Click "migrate"
8. Enter the identity number and click authenticate. I get the QR code of passkey (it's trying RP ID ic0.app). I cancel.
9. I see the message: Try again
10. I click authenticate. I am requested to use Touch ID (it uses beta.identity.internetcomputer.org and I have a passkey there).
11. I continue with the flow
12. I am successfully redircted to the dashboard in progress.

### DIFF
--- a/src/frontend/src/lib/flows/migrationFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/migrationFlow.svelte.ts
@@ -19,11 +19,16 @@ import { inferPasskeyAlias, loadUAParser } from "$lib/legacy/flows/register";
 import { lastUsedIdentitiesStore } from "$lib/stores/last-used-identities.store";
 import { throwCanisterError } from "$lib/utils/utils";
 import { toaster } from "$lib/components/utils/toaster";
+import { findWebAuthnFlows, WebAuthnFlow } from "$lib/utils/findWebAuthnFlows";
+import { supportsWebauthRoR } from "$lib/utils/userAgent";
+import { canisterConfig } from "$lib/globals";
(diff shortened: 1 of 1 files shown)


lmuntaner:
UX:


https://github.com/user-attachments/assets/95789ff8-5c0e-478c-9ec4-f89766fbf116

LXIF (approved):
If would have to go deeper into the functions called here to do a deep review, but don't see anything egregious so LGTM

lmuntaner:
> If would have to go deeper into the functions called here to do a deep review, but don't see anything egregious so LGTM

It's using the same logic as in the old flow and the old flow has been throughly tested the last months in production. So we shuold be good.

PR #3148, opened 2025-06-27
Fix mobile dashboard layout

LXIF:
# Motivation

There's a small layout bug in the current minimal dashboard on some mobile phones.

# Changes

Small css changes to alleviate the bug. Will need to check with Design about potentially changing font size on mobile.

Before:
Screenshot 2025-06-27 at 14 58 32

After:
Screenshot 2025-06-27 at 14 58 10

# Tests

- [x] Looked at it with my eyeballs

### DIFF
--- a/src/frontend/src/lib/components/ui/AccessMethod.svelte
+++ b/src/frontend/src/lib/components/ui/AccessMethod.svelte
@@ -141,6 +141,6 @@
   .foldable-subgrid {
     display: grid;
     /* We may need to update the minimum width for really long emails, but it seems fine for now.  */
-    grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
+    grid-template-columns: repeat(auto-fit, minmax(180px, 1fr));
   }
 </style>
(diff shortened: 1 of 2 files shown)


LXIF:
Hold off on reviewing, Layout needs to be aligned with figma first.

src/frontend/src/lib/components/ui/AccessMethod.svelte
```diff
@@ -141,6 +141,6 @@
   .foldable-subgrid {
     display: grid;
     /* We may need to update the minimum width for really long emails, but it seems fine for now.  */
-    grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
+    grid-template-columns: repeat(auto-fit, minmax(180px, 1fr));
```
lmuntaner, line 144: Why was it overflowing?

Why does this fix it?

Why "auto-fit" and not "auto-fill" or why not controlling each column and discussing with design how to align it?

LXIF: I need to redo it anyhow, see above comment (was different from Figma, hadn't checked before fixing)

It was overflowing because 200px was too wide. Auto-fill would create implicit columns, which I didn't want.

PR #3150, opened 2025-06-30
Align error handling with pre-existing pattern

LXIF:
# Motivation

Align error handling with pre-existing pattern

# Changes

Aligned error handling with pre-existing pattern

# Tests

- [x] tried to add an already linked account. error still gets displayed.

### DIFF
--- a/src/frontend/src/lib/components/utils/error.ts
+++ b/src/frontend/src/lib/components/utils/error.ts
@@ -66,6 +66,7 @@ export const handleError = (error: unknown) => {
         });
         break;
       case "Unauthorized":
+        break;
       case "InternalCanisterError":
         toaster.error({
           title: "An internal error occurred",
--- a/src/frontend/src/lib/components/utils/error.ts
+++ b/src/frontend/src/lib/components/utils/error.ts
@@ -66,10 +66,6 @@ export const handleError = (error: unknown) => {
         });
         break;
       case "Unauthorized":
-        toaster.error({
-          title: "You are not authorized to add this credential",
-        });
-        break;
       case "InternalCanisterError":
         toaster.error({
           title: "An internal error occurred",
--- a/src/frontend/src/lib/components/utils/error.ts
+++ b/src/frontend/src/lib/components/utils/error.ts
@@ -62,7 +62,7 @@ export const handleError = (error: unknown) => {
         break;
       case "OpenIdCredentialAlreadyRegistered":
         toaster.error({
-          title: "This credential is already linked to another identity",
+          title: "This account is already linked to another identity",
         });
         break;
       case "Unauthorized":
--- a/src/frontend/src/lib/components/utils/error.ts
+++ b/src/frontend/src/lib/components/utils/error.ts
@@ -5,6 +5,8 @@ import type {
   CheckCaptchaError,
   IdRegFinishError,
   IdRegStartError,
+  OpenIdCredentialAddError,
+  OpenIdCredentialRemoveError,
 } from "$lib/generated/internet_identity_types";
 import { isOpenIdCancelError } from "$lib/utils/openID";
 
@@ -21,9 +23,13 @@ export const handleError = (error: unknown) => {
 
   // Handle canister errors
   if (
-    isCanisterError<IdRegStartError | IdRegFinishError | CheckCaptchaError>(
-      error,
-    )
+    isCanisterError<
+      | IdRegStartError
+      | IdRegFinishError
+      | CheckCaptchaError
+      | OpenIdCredentialAddError
+      | OpenIdCredentialRemoveError
+    >(error)
   ) {
     switch (error.type) {
       case "RateLimitExceeded":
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/utils/error.ts
```diff
@@ -54,6 +60,32 @@ export const handleError = (error: unknown) => {
           description: error.type,
         });
         break;
+      case "OpenIdCredentialAlreadyRegistered":
+        toaster.error({
+          title: "This credential is already linked to another identity",
```
sea-snake: I would call it "account" instead of credential.

src/frontend/src/lib/components/utils/error.ts
```diff
@@ -54,6 +60,32 @@ export const handleError = (error: unknown) => {
...
+        });
+        break;
+      case "Unauthorized":
+        toaster.error({
+          title: "You are not authorized to add this credential",
```
sea-snake: Unauthorized variant could also be thrown by other errors e.g. IdentityInfoError, so the current message is too specific. 

On another note, this error should never be shown to the user since authorized actions should only be performed when the user is actually authorized, so if it reaches here, it's basically an unhandled error. 

Therefore I would recommend moving it to the "Unhandled error" case above.

src/frontend/src/lib/components/views/AddOpenIdCredential.svelte
```diff
@@ -3,8 +3,22 @@
...
+  import { throwCanisterError } from "$lib/utils/utils";
 
   const { onClose } = $props();
+
+  const handleAddCredential = async () => {
+    try {
+      let res = await identityInfo.addGoogle();
+
+      if ("Err" in res) {
+        throwCanisterError(res);
```
sea-snake: I would make `identityInfo.addGoogle` throw a canister error.

src/frontend/src/lib/components/utils/error.ts
```diff
@@ -54,6 +60,32 @@ export const handleError = (error: unknown) => {
...
+        break;
+      case "InternalCanisterError":
+        toaster.error({
+          title: "An internal error occurred",
+          description: error.value(error.type),
+        });
+        break;
+      case "JwtVerificationFailed":
+        toaster.error({
+          title: "The JWT is invalid",
```
sea-snake: This error message is probably too technical, see suggestion below for an alternative.

```suggestion
          title: "Session token invalid",
          description: "It might have expired, please try again",
```

src/frontend/src/lib/components/utils/error.ts
LXIF: Wouldn't that risk confusion with our concept of "Account"?

src/frontend/src/lib/components/utils/error.ts
LXIF: This is still technical. If we want it to be user friendly we should avoid talking about tokens altogether.

src/frontend/src/lib/components/views/AddOpenIdCredential.svelte
LXIF: Sure, why not

src/frontend/src/lib/components/utils/error.ts
LXIF: If it shouldn't be shown, shouldn't we just remove any handling?

src/frontend/src/lib/components/utils/error.ts
LXIF: Anyhow, adjusted ✅

src/frontend/src/lib/components/utils/error.ts
LXIF: Removed ✅

src/frontend/src/lib/components/views/AddOpenIdCredential.svelte
LXIF: Done ✅

src/frontend/src/lib/components/utils/error.ts
LXIF: Done ✅

src/frontend/src/lib/components/utils/error.ts
```diff
@@ -54,6 +60,29 @@ export const handleError = (error: unknown) => {
...
+      case "OpenIdCredentialAlreadyRegistered":
+        toaster.error({
+          title: "This account is already linked to another identity",
+        });
+        break;
+      case "Unauthorized":
+      case "InternalCanisterError":
+        toaster.error({
+          title: "An internal error occurred",
+          description: error.value(error.type),
```
sea-snake, line 73: Looking at the `Unauthorized` variant in the did file, this would put the principal object as description value instead of a string.

sea-snake: <s>Also `Unauthorized` isn't specified on line 26.</s>

Edit: My bad, `Unauthorized` is a variant within e.g. `OpenIdCredentialAddError`.

LXIF: The Unauthorized case doesn't do anything, the Internal canister error should put the text description of the error. Am I missing something or did you forget to approve? @sea-snake

sea-snake: Both case statements for Unauthorized and InternalCanisterError are right above one another (line 68-69) and thus will run the same code that sets a description based on the variant value.

LXIF: Oh, I forgot to add `break`. Sure

LXIF: There.

sea-snake: I don't think the error should be swallowed, but it's fine for now, I have another PR pending on the same file with more changes (error handling of global max additional accounts).

sea-snake (approved):

PR #3153, opened 2025-07-01
Fix popover shadows

LXIF:
# Motivation

The popover had a css bug that clipped its shadows.

# Changes

Move the shadow to the parent element, add some transition adjustments so it fades in smoothly.

Old:

Screenshot 2025-07-01 at 10 57 09

New:

Screenshot 2025-07-01 at 10 57 19
<video src="https://github.com/user-attachments/assets/02af5ef4-7046-443c-9d24-9748f58b9098" />

# Tests

- [x] Tested with my eyes.

### DIFF
--- a/src/frontend/src/lib/components/ui/Popover.svelte
+++ b/src/frontend/src/lib/components/ui/Popover.svelte
@@ -103,7 +103,6 @@
       popoverRef?.classList.add("shadow-xl");
     }}
     onoutrostart={() => {
-      popoverRef?.classList.remove("shadow-xl");
       popoverRef?.hidePopover();
     }}
     class="popover slow-shadow fixed rounded-xl bg-transparent"
--- a/src/frontend/src/lib/components/ui/Popover.svelte
+++ b/src/frontend/src/lib/components/ui/Popover.svelte
@@ -100,16 +100,15 @@
     out:fade|global={{ delay: 160, duration: 1 }}
     onintrostart={() => {
       popoverRef?.showPopover();
-      popoverRef?.classList.add("shadow-xl");
     }}
     onoutrostart={() => {
       popoverRef?.hidePopover();
     }}
-    class="popover slow-shadow fixed rounded-xl bg-transparent"
+    class="popover fixed overflow-visible bg-transparent"
   >
     <div
       class={[
-        "bg-bg-primary_alt border-border-secondary flex w-90 flex-col rounded-xl border p-4",
+        "bg-bg-primary_alt border-border-secondary flex w-90 flex-col rounded-xl border p-4 shadow-xl",
         {
           up: {
             start: "origin-bottom-left",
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/ui/Popover.svelte
```diff
@@ -98,13 +98,19 @@
...
     out:fade|global={{ delay: 160, duration: 1 }}
-    onintrostart={() => popoverRef?.showPopover()}
-    onoutrostart={() => popoverRef?.hidePopover()}
-    class="popover fixed bg-transparent"
+    onintrostart={() => {
+      popoverRef?.showPopover();
+      popoverRef?.classList.add("shadow-xl");
+    }}
+    onoutrostart={() => {
+      popoverRef?.classList.remove("shadow-xl");
```
lmuntaner: Why is removing and adding necessary? Is it to create the animation?

LXIF: Yes, though arguably the removal can be removed because the out-transition of the element is basically immediate. The adding is necessary so we can transition the shadow in, and since the div changes size on its intro but the shadow does not, if it's just there already we get an ugly "pre-shadow" in the shape that the div grows into.

LXIF: Done ✅

src/frontend/src/lib/components/ui/Popover.svelte
```diff
@@ -98,13 +98,18 @@
     popover={closeOnOutsideClick ? "auto" : "manual"}
     in:fade|global={{ duration: 1 }}
     out:fade|global={{ delay: 160, duration: 1 }}
-    onintrostart={() => popoverRef?.showPopover()}
-    onoutrostart={() => popoverRef?.hidePopover()}
-    class="popover fixed bg-transparent"
+    onintrostart={() => {
+      popoverRef?.showPopover();
+      popoverRef?.classList.add("shadow-xl");
```
sea-snake: This parent element doesn't animate it's transform scale, which was intentional since that would break the measuring during animation in the positioning implementation.

Which does mean that the shadow is now detached from the actual inner popover div that has a border and background during the animation. Due to the animation speed and the transparency of the shadow it's not that visible, but the gap between the popover and the shadow is there:

image

Previously the shadow on the inner div was apparently clipped by the overflow of the outer div (Good catch, didn't see it since I mostly worked in dark mode). That can be resolved by adding `overflow-visible` to the outer div and `overflow-hidden` on the inner div.

To animate the shadow you can probably also do that by overriding the box-shadow in `@starting-style .popover:popover-open > div`.

LXIF: Fixed, now looks like this


https://github.com/user-attachments/assets/2881458d-1887-49f0-80e3-d41added72f5

LXIF: (With the caveat that the shadow now fades in at the same speed as the inner div, I like when it's a bit slower as it lets the div come out slightly more dynamically but i'll leave that to design)

PR #3154, opened 2025-07-01
Add logout button

LXIF:
# Motivation

We want users to be able to log out.

# Changes

Add logout button and functionality.

Looks like this as per Design:
Screenshot 2025-07-01 at 14 11 44

# Tests

- [x] I logged out manually.

### DIFF
--- a/src/archive/src/main.rs
+++ b/src/archive/src/main.rs
@@ -53,7 +53,7 @@ use ic_stable_structures::{
 };
 use internet_identity_interface::archive::types::*;
 use internet_identity_interface::http_gateway::{HttpRequest, HttpResponse};
-use internet_identity_interface::internet_identity::types::*;
+use internet_identity_interface::internet_identity::types::{AnchorNumber, Timestamp};
 use serde_bytes::ByteBuf;
 use std::borrow::Cow;
 use std::cell::RefCell;


src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
```diff
@@ -112,6 +115,10 @@
     </FeaturedIcon>
     <span>Use another identity</span>
   </ButtonCard>
+  {#if onLogout}
+    <Button onclick={onLogout} variant="tertiary"><LogOutIcon />Sign Out</Button
```
sea-snake: Nitpick: icon size in design is 1rem for default button size if I remember correctly.

```suggestion
    <Button onclick={onLogout} variant="tertiary"><LogOutIcon size="1rem" />Sign Out</Button
```

LXIF: It's 1.25rem/20px in the design but nice catch regardless - changed ✅

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -149,6 +151,12 @@ class IdentityInfo {
     await throwCanisterError(googleRemoveResult);
   };
 
+  logout = () => {
+    this.reset();
+    void authorizationStore.init();
+    void goto("/");
```
sea-snake, line 159: Why is the authorization store reset? shouldn't this be the authentication store?

LXIF: Now that you mention it, it does look like it yes.

LXIF: done ✅

sea-snake (approved):

PR #3155, opened 2025-07-01
Better UX on openID credential removal

LXIF:
# Motivation

When a user removes an OpenID credential they are currently logged in with, we want to log the user out and remove the identity from localstorage. This way we can at least locally avoid broken UX with unusable "continue as" state.

[NOTE]: We had discussed removing the identity from localstorage on any access method mutation. However, this would lead to bad UX and after a chat with Erika, the entire UX implications of adding/removing access methods require proper UX attention. Therefore, I just implemented a minimal solution for the most egregious case here.

# Changes

- Add removal functionality to lastUsedIdentitiesStore
- Add the necessary checks and functionality on openID credential removal
- Add an additional conditional warning in the openID credential removal modal 

# Tests

Upon manual testing I:
- [x] get logged out and the identity disappears from the "continue as" modal when removing an OpenID credential I am currently logged in with
- [x] stay logged in and the identity remains in the "continue as" modal when removing an OpenID credential I am not currently logged in with

### DIFF
--- a/src/frontend/src/lib/stores/identity-info.state.svelte.ts
+++ b/src/frontend/src/lib/stores/identity-info.state.svelte.ts
@@ -176,9 +176,10 @@ class IdentityInfo {
 
     if ("Ok" in googleRemoveResult) {
       void this.fetch();
-      // lastUsedIdentitiesStore.removeIdentity(
-      //   get(authenticatedStore).identityNumber,
-      // );
+      lastUsedIdentitiesStore.removeIdentity(
+        get(authenticatedStore).identityNumber,
+      );
+      // TODO: if we are signed in via google, we should sign out here.
     } else {
       this.openIdCredentials.push(temporaryCredential);
       toaster.error({
--- a/src/frontend/src/lib/stores/identity-info.state.svelte.ts
+++ b/src/frontend/src/lib/stores/identity-info.state.svelte.ts
@@ -3,8 +3,6 @@ import {
   AuthnMethodData,
   AuthnMethodRegistrationInfo,
   OpenIdCredential,
-  OpenIdCredentialAddError,
-  OpenIdCredentialRemoveError,
 } from "$lib/generated/internet_identity_types";
 import { authenticatedStore } from "./authentication.store";
 import { isNullish, nonNullish } from "@dfinity/utils";
@@ -15,8 +13,13 @@ import {
   decodeJWTWithNameAndEmail,
   requestJWT,
 } from "$lib/utils/openID";
-import { toaster } from "$lib/components/utils/toaster";
-import { lastUsedIdentitiesStore } from "./last-used-identities.store";
+import { throwCanisterError } from "$lib/utils/utils";
+import {
+  lastUsedIdentitiesStore,
+  lastUsedIdentityStore,
+} from "./last-used-identities.store";
+import { authorizationStore } from "./authorization.store";
+import { goto } from "$app/navigation";
 
 const fetchIdentityInfo = async () => {
   const authenticated = get(authenticatedStore);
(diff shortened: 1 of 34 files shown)


src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -141,6 +145,20 @@ class IdentityInfo {
...
+      // If we just deleted the method we are logged in with, we log the user out.
+      const lastUsedAuthMethod = get(lastUsedIdentityStore)?.authMethod;
+      if (
+        lastUsedAuthMethod &&
+        "openid" in lastUsedAuthMethod &&
+        lastUsedAuthMethod.openid.sub === temporaryCredential.sub
+      ) {
+        this.logout();
+        lastUsedIdentitiesStore.removeIdentity(
+          get(authenticatedStore).identityNumber,
```
lmuntaner, line 152: Why don't yuo use the identityNumber you extract in line 94?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -82,6 +83,8 @@
 
 {#if identityInfo.removableOpenIdCredential}
   <RemoveOpenIdCredential
+    lastUsedCredential={$lastUsedIdentityStore?.authMethod}
+    credentialToBeRemoved={identityInfo.removableOpenIdCredential}
```
lmuntaner, line 88: How about adding a `isOpenIdCurrentAccessMethod` in `identityInfo` to keep the logic of comparing credentials there?

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -141,6 +145,20 @@ class IdentityInfo {
     const googleRemoveResult = await googleRemovePromise;
 
     if ("Ok" in googleRemoveResult) {
+      // If we just deleted the method we are logged in with, we log the user out.
+      const lastUsedAuthMethod = get(lastUsedIdentityStore)?.authMethod;
+      if (
+        lastUsedAuthMethod &&
+        "openid" in lastUsedAuthMethod &&
+        lastUsedAuthMethod.openid.sub === temporaryCredential.sub
```
lmuntaner: How about creating the variable `isOpenIdCurrentAccessMethod` before we make the call to google and remove it so that we can use it here instead of all these lines? You can take the value from the method `isOpenIdCurrentAccessMethod` that I propose to add to `identityInfo` as well.

lmuntaner (commented):
LGTM!

Mainly small code changes to keep business logic together.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
LXIF: Because that's in a different method and thus out of scope.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
LXIF: That feels a bit too specific, but I now added a general `isCurrentAccessMethod` that we'll be able to reuse for passkeys as well as when we add more openID providers.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
LXIF: ^See above

src/frontend/src/lib/stores/identity-info.state.svelte.ts
lmuntaner: Oh, I see now, yes.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -149,6 +162,38 @@ class IdentityInfo {
     await throwCanisterError(googleRemoveResult);
   };
 
+  logout = () => {
+    this.reset();
+    void authorizationStore.init();
+    void goto("/");
+  };
+
+  isCurrentAccessMethod = (
```
lmuntaner, line 171: Nice! Good idea passing the access method.

lmuntaner (approved):
Thanks!

PR #3162, opened 2025-07-03
Add missing lib to try-ii

LXIF:
# Motivation

`try-ii` accidentally had its lib removed by our global gitignore file.

# Changes

Add exception for try-ii to global gitignore file, add lib I had locally.

# Tests

- [x] Was able to push, so it should work now.


.gitignore
```diff
@@ -18,6 +18,7 @@ dist-showcase/
 .env
 lib
 !src/frontend/src/lib
+!src/try-ii/src/try-ii-frontend/src/lib
```
lmuntaner, line 21: why this?

LXIF: We have `lib` in our gitignore, so we have to unignore our sveltekit libs

PR #3167, opened 2025-07-08
Samsung Browser doesn't support FedCM with our params

lmuntaner:
# Motivation

User got stuck in a loading state when trying to authenticate with Google with a Samsung browser.

Samsung browser might use an older version of FedCM not compatible with our parameters.

# Changes

* Fallback to authentication with redirect for Samsung browsers.
* Created helper to check for FedCM support.

# Tests

Deployed to beta.
Tried to authenticate on beta with Google on Samsung browser after deploying.


LXIF (approved):
LGTM

PR #3168, opened 2025-07-08
Safari blocking Google popup

lmuntaner:
# Motivation

During my testing round, Safari blocked the popup to authenticate with Google.

The reason must be similar to why II pop up is blocked in Safari. In our case, we were triggering analytics before opening the popup.

# Changes

* Move analytics call to after the popup is open.

# Tests

Deployed to beta, but I couldn't replicate the blocking anymore, so I can't test that this is indeed the solution.

According to the docs, it's not clear what is necessary for Safari to block or not block: https://webkit.org/blog/13862/the-user-activation-api/

Deployed to beta and tested that registering and signing it works.

### DIFF
--- a/src/frontend/src/lib/flows/authFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authFlow.svelte.ts
@@ -142,7 +142,6 @@ export class AuthFlow {
       });
       this.#options.onSignIn(identityNumber);
     } catch (error) {
-      this.systemOverlay = false;
       if (
         isCanisterError<OpenIdDelegationError>(error) &&
         error.type === "NoSuchAnchor" &&
@@ -157,6 +156,8 @@ export class AuthFlow {
       }
       this.view = "chooseMethod";
       throw error;
+    } finally {
+      this.systemOverlay = false;
     }
   };
 
--- a/src/frontend/src/lib/flows/authFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authFlow.svelte.ts
@@ -106,13 +106,14 @@ export class AuthFlow {
       throw new Error("Google is not configured");
     }
     try {
-      authenticationV2Funnel.trigger(AuthenticationV2Events.ContinueWithGoogle);
       const requestConfig = createGoogleRequestConfig(clientId);
       this.systemOverlay = true;
       jwt = await requestJWT(requestConfig, {
         nonce: get(sessionStore).nonce,
         mediation: "required",
       });
+      // Moved after `requestJWT` to avoid Safari from blocking the popup.
+      authenticationV2Funnel.trigger(AuthenticationV2Events.ContinueWithGoogle);
       this.systemOverlay = false;
       const { identity, identityNumber, iss, sub } = await authenticateWithJWT({
         canisterId,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -134,6 +135,9 @@ export class AuthFlow {
       });
       this.#options.onSignIn(identityNumber);
     } catch (error) {
+      // We also want to trigger it if `requestJWT` fails to have the drop-off in the analytics.
+      // Double-triggering is not a problem as it's considered by the same user in Plausible.
+      authenticationV2Funnel.trigger(AuthenticationV2Events.ContinueWithGoogle);
```
sea-snake: If we want to trigger it in both try and catch, shouldn't we then use `finally`?

LXIF: That would indeed be neat.

lmuntaner: The order of triggering events matter for the funnel and in this flow we trigger other events such as:

```
authenticationV2Funnel.trigger(AuthenticationV2Events.RegisterWithGoogle);
// or
authenticationV2Funnel.trigger(AuthenticationV2Events.LoginWithGoogle);
```

That's why I need to trigger it twice to make sure it comes before those two.

sea-snake: Ah that makes sense, but wouldn't triggering it in catch already be putting it in the wrong order if an error is thrown at a later point?

sea-snake: Maybe moving the requestJWT into it's own try catch finally above the current try catch would avoid ordering issues?

lmuntaner: two try/catch sounds like a good idea!

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -105,14 +105,23 @@ export class AuthFlow {
...
+    } catch (error) {
+      this.systemOverlay = false;
+      this.view = "chooseMethod";
+      throw error;
+    } finally {
+      // Moved after `requestJWT` to avoid Safari from blocking the popup.
+      authenticationV2Funnel.trigger(AuthenticationV2Events.ContinueWithGoogle);
+    }
+    try {
       this.systemOverlay = false;
```
sea-snake: This line can be removed
```suggestion
```

sea-snake: There's another one a few lines down that I can't seem to comment on.

lmuntaner: thanks!

src/frontend/src/lib/flows/authFlow.svelte.ts
lmuntaner: Done!

src/frontend/src/lib/flows/authFlow.svelte.ts
lmuntaner: Maybe I removed it already. I pushed changes now.

PR #3169, opened 2025-07-08
Share error message on JWT verification

lmuntaner:
# Motivation

I got a few times an error when authenticating with Google that the verification failed.

However, I didn't get any reason for it.

In this PR, I add the text with the reason and send that to Plausible for debugging.

# Changes

* Add "String" to `JwtVerificationFailed` error types.
* Add event in `AuthenticationV2Events`.
* Trigger the error event when handling the error along with the error text.

# Tests

Hardcoded the following error in `throwCanisterError`: `throw new CanisterError({ JwtVerificationFailed: "Test error" });` and then triggered a Google auth and got the expected error and a console.log of the "Test error" text.

### DIFF
--- a/src/frontend/src/lib/components/utils/error.ts
+++ b/src/frontend/src/lib/components/utils/error.ts
@@ -10,6 +10,10 @@ import type {
   OpenIdCredentialRemoveError,
 } from "$lib/generated/internet_identity_types";
 import { isOpenIdCancelError } from "$lib/utils/openID";
+import {
+  AuthenticationV2Events,
+  authenticationV2Funnel,
+} from "$lib/utils/analytics/authenticationV2Funnel";
 
 export const handleError = (error: unknown) => {
   // Handle browser errors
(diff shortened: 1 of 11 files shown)


src/frontend/src/lib/components/utils/error.ts
```diff
@@ -68,8 +72,22 @@ export const handleError = (error: unknown) => {
       case "JwtVerificationFailed":
         toaster.error({
           title: "Authorization invalid",
-          description: "It may have expired — please try again",
+          description:
+            "There was an error verifying your account — please try again.",
         });
+        authenticationV2Funnel.trigger(
```
sea-snake, line 81: Not sure if `AuthenticationV2Events` funnel makes the most sense here since this error could also be thrown when adding a google account from the dashboard, polluting the funnel events.

Instead if we want to track this specifically in the authentication flow we can probably add this in `authFlow.ts`.

lmuntaner: Good point.

It's still ok. The funnels in Plausible are filtered by the same user triggering the previous event. Therefore, a user triggering the error from the dashboard would not impact the user triggering it from the auth.

Unless it's the same browser within the same day (what plausible considers a "user").

But it might be worth to make a comment to understand it.

If we keep it here. We will also gather the errors from the dashboard, which I wasn't thinking, but it's also good.

sea-snake: > If we keep it here. We will also gather the errors from the dashboard, which I wasn't thinking, but it's also good.

But if we want to gather errors from the dashboard I'd say that those errors should probably outside the funnel.

Overall, I think it's a larger discussion, generic error handling (with plausible) and funnels, in the long run, I'd imagine these being separate concepts and events.

lmuntaner: Yeah, good point. I also thought of putting them outside.

That depends whether we care about errors within a funnel or we want to analyze them separately.

Food for thought and another discussion. Let's keep them here and clean up afterwards while keeping the same event name and nothing should be lost in Plausible.

PR #3170, opened 2025-07-08
Add random id based tentative device flows

LXIF:
# Motivation

For II 2.0 we want to enable random id keyed tentative device flows to improve UX when adding new passkeys on other devices.

# Changes

Add lookup table from random registration id to identity number, add convenience functions to interact with state, add more functions (incomplete as of now, this is a draft)

# Tests

Added new integration tests for two new endpoints.

### DIFF
--- a/src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
+++ b/src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
@@ -49,7 +49,8 @@ fn should_require_authentication_to_enter_authn_method_registration_mode() {
         Principal::anonymous(),
         identity_number,
         Some(registration_mode_id),
-    ).expect("authn_method_registration_mode_enter failed");
+    )
+    .expect("authn_method_registration_mode_enter failed");
 
     assert!(matches!(
         result,


src/internet_identity/src/main.rs
```diff
@@ -858,8 +863,12 @@ mod v2_api {
     #[update]
     fn authn_method_registration_mode_enter(
         identity_number: IdentityNumber,
+        id: RegistrationId,
     ) -> Result<RegistrationModeInfo, ()> {
-        let timeout = enter_device_registration_mode(identity_number);
+        check_authz_and_record_activity(identity_number)
```
lmuntaner, line 868: Wow, does that mean that an anonymous caller could start the registration mode on any identity?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -45,6 +48,43 @@ pub fn enter_device_registration_mode(anchor_number: AnchorNumber) -> Timestamp
...
+                        id,
+                        (
+                            identity_number,
+                            TentativeDeviceRegistration {
+                                expiration,
+                                state: DeviceRegistrationModeActive,
+                            },
+                        ),
+                    );
+                    lookup;
```
lmuntaner: What does this do?

Shouldn't we insert something to the lookup table?

src/internet_identity/src/state.rs
```diff
@@ -190,6 +217,11 @@ struct State {
     // tentative device registrations, not persisted across updates
     // if an anchor number is present in this map then registration mode is active until expiration
     tentative_device_registrations: RefCell<HashMap<AnchorNumber, TentativeDeviceRegistration>>,
+    // if an anchor number + id combo is present in this map then registration mode is active until expiration
+    tentative_device_registrations_v2:
+        RefCell<HashMap<RegistrationId, (IdentityNumber, TentativeDeviceRegistration)>>,
+    // lookup table so we can easily return a user's active registrations in identity_info
+    lookup_tentative_device_registration_v2: RefCell<HashMap<IdentityNumber, Vec<RegistrationId>>>,
```
lmuntaner: Should we allow multiple registration ids at the same time for one identity?

II 1.0 wasn't allowing that, right?

lmuntaner: Could we use `tentative_device_registrations` along a lookup table of `RegistrationId` to `AnchorNumber`?

Requirements:
- Enter registration mode.
- Retrieve registration mode based on registration id.

src/internet_identity/src/main.rs
LXIF: It shouldn't, we are checking the auth with `check_authz_and_record_activity`.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
LXIF: Ah right, I knew I forgot something

src/internet_identity/src/state.rs
LXIF: II 1.0 was not allowing that, correct. I thought we had mentioned it during a meeting but admittedly I don't recall the exact use case. I can refactor it to use the old memory, yes.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
LXIF: done ✅

src/internet_identity/src/state.rs
LXIF: done ✅

src/internet_identity/internet_identity.did
```diff
@@ -839,7 +851,7 @@ service : (opt InternetIdentityInit) -> {
     // confirmed before it can be used for authentication on this identity.
     // The registration mode is automatically exited after the returned expiration timestamp.
     // Requires authentication.
-    authn_method_registration_mode_enter : (IdentityNumber) -> (variant { Ok : record { expiration : Timestamp }; Err });
+    authn_method_registration_mode_enter : (IdentityNumber, text) -> (variant { Ok : record { expiration : Timestamp }; Err : AuthnMethodRegistrationModeEnterError });
```
lmuntaner: For backwards compatibility we should keep the second parameter optional, no?

Can we create a dummy type to make clear what the `text` means? Something like `type RegistrationId = text;`

src/internet_identity/src/anchor_management.rs
```diff
@@ -38,42 +35,15 @@ pub fn get_anchor_info(anchor_number: AnchorNumber) -> IdentityAnchorInfo {
...
-            None | Some(_) => IdentityAnchorInfo {
-                devices,
-                device_registration: None,
-                openid_credentials,
-                name,
-            },
-        }
-    })
+    let tentative_device_registration =
+        state::get_tentative_device_registration_by_identity(anchor_number);
```
lmuntaner, line 39: Where did all the logic go? Isn't it needed?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -144,38 +187,57 @@ fn get_verified_device(
...
             }
-        }
+        })
+    })
+}
+
+/// Checks whether a tentative device has been verified without mutating anything
+/// This is so that on the new client we can prompt for adding the final passkey as soon as
+/// on the old client we have verified the temporary key
+pub fn check_tentative_device(identity_number: IdentityNumber) -> bool {
```
lmuntaner: I think that the design doc expected to use `identity_authn_info` in the end. Any reason to use something else?

The final backend changes should add a session and check that, not a device registration, so we will change this anyway.

src/internet_identity/src/main.rs
```diff
@@ -858,8 +865,14 @@ mod v2_api {
     #[update]
     fn authn_method_registration_mode_enter(
         identity_number: IdentityNumber,
-    ) -> Result<RegistrationModeInfo, ()> {
-        let timeout = enter_device_registration_mode(identity_number);
+        id: String,
```
lmuntaner: Ditto, should we make this optional to keep backwards compatibility or isn't it backwards compatible anyway with an optional field?

lmuntaner: If we break backwards compatibility, we need to also change the frontend code in this PR. But we still don't have the random code generated in the frontend. So it makes sense to make it optional anyway to merge these changes.

src/internet_identity/src/main.rs
lmuntaner: Yes, but I don't see it in the old code, that's why I was asking...

src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
```diff
@@ -335,3 +353,441 @@ fn should_reject_confirmation_with_wrong_code() -> Result<(), CallError> {
...
+
+    assert!(matches!(
+        result,
+        Err(LookupByRegistrationIdError::InvalidId(_))
+    ));
+    Ok(())
+}
+
+#[test]
+fn should_handle_multiple_registration_ids() -> Result<(), CallError> {
```
lmuntaner, line 690: For multiple users, right?

Should we also check that exiting the mode on one doesn't exit the other registration?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -204,3 +266,46 @@ fn prune_expired_tentative_device_registrations(
...
+}
+
+impl From<AuthorizationError> for CheckTentativeDeviceError {
+    fn from(_err: AuthorizationError) -> Self {
+        CheckTentativeDeviceError::Unauthorized
+    }
+}
+
+#[derive(CandidType, Clone, Eq, PartialEq, Hash)]
+pub struct RegistrationId(String);
```
lmuntaner: As the bot code reviewer commented, can't we use the type in api_v2.rs?

lmuntaner (commented):
Can you ask Jason to take a look as well?

src/internet_identity/internet_identity.did
LXIF: No, because the old version of the frontend uses the v1 endpoint `enter_device_registration_mode`. This endpoint is not yet being used, so nothing is broken by changing it. Yes, we can make a type alias.

src/internet_identity/src/anchor_management.rs
LXIF: I abstracted the logic into the method `to_maybe_info` in `impl TentativeDeviceRegistration`.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
LXIF: You should be able to use either. I already had this endpoint and it is currently faster as it is query. The frontend can decide which one to use.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
LXIF: Then we would have to add the `impl` in `api_v2.rs` as well, which would go against the idea of that file being part of `internet_identity_interface`.

src/internet_identity/src/main.rs
LXIF: As I said, the frontend does not use the v2 api endpoint, it uses the v1 api endpoint which remains unchanged. We can keep using the v1 endpoint and just switch to the v2 endpoint when we add the random code.

src/internet_identity/src/main.rs
LXIF: The old code just calls the v1 update method, which then performs the auth check. In this new code, I directly call `enter_device_registration_mode_v2` in `tentative_device_registration.rs`, so I need to first check the auth.

src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
LXIF: Yes, this is for two separate identities. I can add another test for this.

src/internet_identity/src/anchor_management.rs
LXIF: I renamed it now to `to_info_if_still_valid` to be more expressive.

src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
LXIF: done ✅

src/internet_identity/internet_identity.did
LXIF: Ok, after merging main I see @sea-snake now implemented it using the new endpoint. Making it optional for now.

src/internet_identity/src/main.rs
LXIF: Ok, after merging main I see @sea-snake now implemented it using the new endpoint. Making it optional for now.

LXIF: done ✅

src/internet_identity/internet_identity.did
LXIF: done ✅

src/internet_identity/src/anchor_management/tentative_device_registration.rs
LXIF: As for this session, the design doc doesn't really talk in more detail about it, functionally it seems to do the same as the tentative device though.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -45,10 +46,52 @@ pub fn enter_device_registration_mode(anchor_number: AnchorNumber) -> Timestamp
...
+                        identity_number,
+                        TentativeDeviceRegistration {
+                            expiration,
+                            state: DeviceRegistrationModeActive,
+                            id: Some(id.clone()),
+                        },
+                    );
+                    lookup.insert(id, identity_number);
+
+                    expiration
```
jasonz-dfinity: Nit: not sure if you considered using the Entry API: `registrations.entry(...).or_insert_with(...).expiration`

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -45,10 +46,52 @@ pub fn enter_device_registration_mode(anchor_number: AnchorNumber) -> Timestamp
...
-        registrations.remove(&anchor_number)
+        state::lookup_tentative_device_registration_mut(|lookup| {
+            prune_expired_tentative_device_registrations_v2(registrations, lookup);
+            if let Some(TentativeDeviceRegistration {
+                id: Some(reg_id), ..
+            }) = registrations.get(&anchor_number)
+            {
+                lookup.remove(reg_id);
+            }
+            registrations.remove(&anchor_number);
```
jasonz-dfinity: Nit: you could avoid `get()` by using the removed value returned by `remove`

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -144,38 +187,57 @@ fn get_verified_device(
...
+        })
+    })
+}
+
+/// Checks whether a tentative device has been verified without mutating anything
+/// This is so that on the new client we can prompt for adding the final passkey as soon as
+/// on the old client we have verified the temporary key
+pub fn check_tentative_device(identity_number: IdentityNumber) -> bool {
+    state::tentative_device_registrations(|registrations| {
+        registrations.get(&identity_number).is_some()
```
jasonz-dfinity: contains_key?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -204,3 +266,46 @@ fn prune_expired_tentative_device_registrations(
...
+) {
+    let now = time();
+
+    registrations.retain(|_, TentativeDeviceRegistration { expiration, id, .. }| {
+        if *expiration > now {
+            true
+        } else {
+            if id.is_some() {
+                lookup.remove(&id.clone().unwrap());
+            }
```
jasonz-dfinity: Nit: `if let Some(id) = id { lookup.remove(&id); }`? Mainly, the `unwrap()` might unnecessarily spook a reader when it's in fact safe.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -204,3 +266,46 @@ fn prune_expired_tentative_device_registrations(
...
+        CheckTentativeDeviceError::Unauthorized
+    }
+}
+
+#[derive(CandidType, Clone, Eq, PartialEq, Hash)]
+pub struct RegistrationId(String);
+
+impl RegistrationId {
+    pub fn new(s: String) -> Result<Self, String> {
+        if s.chars().count() == 5 {
```
jasonz-dfinity: Sometimes strings can be tricky because of encoding, but I'm almost sure that `s.chars().count()` is just `s.len()`?

src/internet_identity/src/state.rs
```diff
@@ -313,6 +345,42 @@ pub fn tentative_device_registrations_mut<R>(
...
+                get_tentative_device_registration_by_identity(identity_number)
+            {
+                if expiration > time() {
+                    Some(identity_number)
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
```
jasonz-dfinity: Nit: you can reduce the nesting a little bit by:

```
            let TentativeDeviceRegistration { expiration, .. } =
                get_tentative_device_registration_by_identity(identity_number)?;
            if expiration > time() {
                Some(identity_number)
            } else {
                None
            }
```

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -204,3 +266,46 @@ fn prune_expired_tentative_device_registrations(
...
+    fn from(_err: AuthorizationError) -> Self {
+        CheckTentativeDeviceError::Unauthorized
+    }
+}
+
+#[derive(CandidType, Clone, Eq, PartialEq, Hash)]
+pub struct RegistrationId(String);
+
+impl RegistrationId {
+    pub fn new(s: String) -> Result<Self, String> {
```
jasonz-dfinity: Very nit: I think the idiomatic name here would be `try_new`

src/internet_identity/src/anchor_management.rs
lmuntaner: Thanks!

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: The reason for the session is to allow the user to access the dashboard without a new face id interaction.

For that, the temporary key created in the frontend will be added to the temp_keys of that user. Like in the registration flow.

But yes, we can use this endpoint instead. It would have been also nicer to discuss that in the design doc instead of in the PR.

sea-snake: To clarify `identity_authn_info` is also a query and public (currently used) while `identity_info` is private and an update method. The `identity_authn_info` endpoint polling should already suffice as far as I understand: https://github.com/dfinity/internet-identity/blob/main/src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts#L69

Also as mentioned in the design doc, you shouldn't be able to authenticate using the session key with any methods yet until the registration has been closed with passkey data (call signed with session key). 

This guarantees that there never will be a session key within the canister state without a corresponding passkey registered for it (as defined in the existing temp_keys implementation).

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: Not according to AI:

> s.len() returns the length of the string in bytes
> s.chars().count() counts the number of Unicode characters in the string

src/internet_identity/src/state.rs
lmuntaner: Cool!

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: Done!

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: good point!

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: thanks!

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: Thanks!

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: Done!

sea-snake:
@lmuntaner 

I see two new error variants `Unauthorized: null` and `AuthorizationFailure: text`. Since we already use `Unauthorized: Principal` in some other methods, can we stay consistent and use that variant instead? This would keep the `handleError` implementation in `error.ts` more straightforward (else you'll have duplicates and even union types there).

src/internet_identity/internet_identity.did
```diff
@@ -513,6 +515,19 @@ type AuthnMethodConfirmationError = variant {
...
+    InvalidId : text;
+    AuthorizationFailure : text;
+};
+
+type CheckTentativeDeviceError = variant {
+    Unauthorized;
+};
+
+type LookupByRegistrationIdError = variant {
+    InvalidId : text;
```
sea-snake: I would be more explicit here e.g. `InvalidRegistrationId` since it's an error variant that's explicitly about registration ids, not re-used in other methods outside the registration flows.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -144,38 +183,60 @@ fn get_verified_device(
...
-                if user_verification_code == verification_code {
-                    return Ok(tentative_device);
-                }
+            // Check if the registration exists
+            if !registrations.contains_key(&anchor_number) {
+                return Err(VerifyTentativeDeviceError::DeviceRegistrationModeOff);
+            }
 
-                let failed_attempts = failed_attempts + 1;
-                if failed_attempts < MAX_DEVICE_REGISTRATION_ATTEMPTS {
```
sea-snake, line 165: Why was this removed? I'd assume this was here for security considerations?

Though this code is only called by the authenticated principal, so no idea why this needed a max attempt implementation in the first place :S

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -144,38 +183,60 @@ fn get_verified_device(
...
+                        if user_verification_code == *verification_code {
+                            // Verification successful - we'll remove the registration
+                            (true, Ok(tentative_device.clone()))
+                        } else {
+                            // Increment failed attempts counter
+                            *failed_attempts += 1;
+
+                            // Check if max attempts reached
+                            let should_remove =
+                                *failed_attempts >= MAX_DEVICE_REGISTRATION_ATTEMPTS;
```
sea-snake, line 222: Ignore my previous comment, seems that the code was moved here.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
sea-snake: We only allow human readable UTF-8 single byte characters, allowing for multi byte characters could be misused if someone were to enter a malicious character code of many bytes. 

On a related note, we probably want to have a character whitelist here (I'd recommend [base62](https://en.wikipedia.org/wiki/Base62) in this particular use case).

sea-snake: Nit: I'd consider naming this struct something along the lines of `ValidatedRegistrationId`, to make it's purpose more clear.

src/internet_identity/internet_identity.did
lmuntaner: Sounds good, I changed it in the other type as well

src/internet_identity/src/anchor_management/tentative_device_registration.rs
lmuntaner: AI suggests to use `is_ascii_alphanumeric`

src/internet_identity/internet_identity.did
```diff
@@ -513,6 +515,20 @@ type AuthnMethodConfirmationError = variant {
...
 };
 
+type AuthnMethodRegistrationModeEnterError = variant { 
+    InvalidRegistrationId : text;
+    Unauthorized : principal;
+    InternalError : text;
+};
+
+type CheckTentativeDeviceError = variant {
+    Unauthorized;
```
sea-snake: Could this be changed to `Unauthorized : principal` to match other unauthorized errors?

lmuntaner: Oops, we don't need this. It's the error from the removed endpoint.

Cleaning!

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -45,10 +46,56 @@ pub fn enter_device_registration_mode(anchor_number: AnchorNumber) -> Timestamp
...
+            prune_expired_tentative_device_registrations_v2(registrations, lookup);
+            if registrations.len() >= MAX_ANCHORS_IN_REGISTRATION_MODE {
+                return Err(AuthnMethodRegistrationModeEnterError::InternalError(
+                    "too many anchors in device registration mode".to_string(),
+                ));
+            }
+
+            let registration = registrations.entry(identity_number).or_insert_with(|| {
+                let expiration = time() + REGISTRATION_MODE_DURATION;
+                lookup.insert(id.clone(), identity_number);
```
jasonz-dfinity, line 66: The fact that any lookup table entry always has a corresponding entry in the registration table, is not that obvious. I wonder if those 2 maps should be represented by one struct in `State`, and mutations all go through that struct, so the consistency between the maps can be more "local" (to that struct). More importantly, it will be more obvious that the lookup table is also bounded by `MAX_ANCHORS_IN_REGISTRATION_MODE`

lmuntaner: That's a good idea. We have something similar in the storage where the storage layer is the responsible for keeping two parts in sync. We could that something similar here. But I consider that a refactor after this PR.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -45,10 +46,48 @@ pub fn enter_device_registration_mode(anchor_number: AnchorNumber) -> Timestamp
...
+                .or_insert_with(|| {
+                    let expiration = time() + REGISTRATION_MODE_DURATION;
+                    lookup.insert(id.clone(), identity_number);
+                    TentativeDeviceRegistration {
+                        expiration,
+                        state: DeviceRegistrationModeActive,
+                        id: Some(id.clone()),
+                    }
+                })
+                .expiration;
```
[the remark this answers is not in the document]
lmuntaner: Good catch! Added a test and fixed it!

src/canister_tests/src/api/internet_identity/api_v2.rs
```diff
@@ -298,6 +299,23 @@ pub fn authn_method_confirm(
     .map(|(x,)| x)
 }
 
+pub fn authn_method_lookup_by_registration_mode_id(
```
sea-snake: In the doc this method is without the `authn_method_` prefix if I remember correctly, with this prefix it reads as if we're looking up an authn method instead of identity number.

src/internet_identity/internet_identity.did
```diff
@@ -855,6 +868,8 @@ service : (opt InternetIdentityInit) -> {
     // Requires authentication.
     authn_method_confirm : (IdentityNumber, confirmation_code : text) -> (variant { Ok; Err : AuthnMethodConfirmationError });
 
+    authn_method_lookup_by_registration_mode_id : (RegistrationId) -> (variant { Ok : opt IdentityNumber; Err : LookupByRegistrationIdError }) query;
```
sea-snake: Since this is just a public query lookup, why would we return ok/err instead of an `opt nat`?

lmuntaner: Why wouldn't we?  Even if it's public we'll use it in our frontend and following the same pattern might be easier for us. We might want to show a different error to the user depending on the error.

src/canister_tests/src/api/internet_identity/api_v2.rs
lmuntaner: done!

src/internet_identity/internet_identity.did
sea-snake: I mean, why would we even have an error? 

We're not mutating state since it's a query and we're not authenticating, we only do a lookup that either finds something or it doesn't.

Having it validate and throw an error if the registration id format is incorrect is helpful but at the same time, I don't really see how this additional detail helps either the user or the implementer. 

This is rather something that would be communicated in typing (sadly we can't do template literal types in candid) or in a candid type comment.

sea-snake: To clarify, we already use the `opt X` approach in the other lookup ( `lookup_device_key`).

src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
```diff
@@ -43,19 +41,20 @@ fn should_require_authentication_to_enter_authn_method_registration_mode() {
...
 
-    expect_user_error_with_message(
-        result,
-        CanisterCalledTrap,
-        Regex::new("[a-z0-9-]+ could not be authenticated.").unwrap(),
-    );
+    assert!(matches!(
+        result.unwrap(),
+        Err(AuthnMethodRegistrationModeEnterError::Unauthorized(_))
+    ))
```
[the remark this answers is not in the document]
lmuntaner: done!

PR #3172, opened 2025-07-09
Show add access method button

sea-snake:
Show add access method button.

# Changes

- Show add access method button if either there are less than 8 passkeys or no google account linked yet.
- Rename some boolean variable to `isForExampleOpenOrVisible`
- Adjust some css classes to align with Figma e.g. subtitle font size
- Change h1, h2, h3, h4 to h1, p, h2, p to improve accessibility
- Remove redundant css classes from add button

# Tests

- Checked that add access method button is visible when:
   -  No google account is linked and passkey limit is not reached
   -  No google account is linked and passkey limit is reached
   - Google account is linked and passkey limit is not reached
- Checked that add access method button is hidden when:
   - Google account is linked and passkey limit is reached

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
@@ -9,7 +9,6 @@
   import RemoveOpenIdCredential from "$lib/components/views/RemoveOpenIdCredential.svelte";
   import AddOpenIdCredential from "$lib/components/views/AddOpenIdCredential.svelte";
   import { ADD_ACCESS_METHOD } from "$lib/state/featureFlags";
-  import { get } from "svelte/store";
 
   const MAX_PASSKEYS = 8;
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -10,78 +10,90 @@
   import AddOpenIdCredential from "$lib/components/views/AddOpenIdCredential.svelte";
   import { lastUsedIdentityStore } from "$lib/stores/last-used-identities.store";
 
-  let displayAddCredentialDialog = $state(false);
+  let isAddAccessMethodDialogOpen = $state(false);
+
+  const isMaxOpenIdCredentialsReached = $derived(
+    identityInfo.openIdCredentials.length >= 1,
+  );
+  const isMaxPasskeysReached = $derived(identityInfo.authnMethods.length >= 8);
```
lmuntaner: Could we create a constant for the 8?

lmuntaner (approved):
Thanks!

In general, let's also use a feature flag in case we would need to do a release in the middle or for some reason we can't make it to the release.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
sea-snake: I'll add a feature flag and define the constants.

PR #3176, opened 2025-07-09
Add access method dialog with passkey on current device.

sea-snake:
Tested adding google account

# Changes

- Implement add passkey dialog following design
- Implement add passkey within flow class to sync state changes with dialog interface

# Tests

- Tested adding another local passkey


src/frontend/src/lib/components/views/AddAccessMethodDialog.svelte
```diff
@@ -0,0 +1,167 @@
...
+          <span>Create passkey</span>
+        {/if}
+      </Button>
+      <Button
+        variant="tertiary"
+        size="lg"
+        disabled={addAccessMethodFlow.isCreatingPasskey}
+      >
+        Continue on another device
+      </Button>
```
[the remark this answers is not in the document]
sea-snake: I'm aware, click handler will be added in another PR.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -34,6 +39,11 @@
     openIdCredentials.push(credential);
     invalidateAll();
   };
+
+  const handlePasskeyRegistered = (authnMethod: AuthnMethodData) => {
+    authnMethods.push(authnMethod);
```
lmuntaner, line 44: Is this updating the "identityInfo" state?

When the user navigates between screens, are the methods still there?

I think it would make more sense to add a "addAccessMethod" to the "IdentityInfo" class.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
```diff
@@ -34,6 +39,11 @@
...
+    invalidateAll();
```
lmuntaner, line 45: What does this do?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
lmuntaner: I see it's also done for openIdCredentials... I didn't see that...

lmuntaner (approved):
Looks good!

We still need to agree on how we keep and update state. I guess we can continue as is for now though.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
sea-snake: Yeah I didn't update the identity info state intentionally, to keep the changes to a minimum for now, this `$derived` approach allows for temporarily overriding the derived state until the "IdentityInfo" state is updated.

You're right that if a user navigates away, this temporary override is no longer there, but I don't think that's a concern given the 2s window in which this could happen.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
sea-snake: Call all the load functions of the current page, which in this particular case calls the identityinfo fetch method.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/security/+page.svelte
sea-snake: Once we agree on an approach, it should be trivial to clean up this code though.

PR #3182, opened 2025-07-11
Remove passkeys

sea-snake:
Initial implementation of remove passkey UX following existing remove OpenId credential implementation.

# Changes

- Implement `RemoveAuthnMethodDialog`.
- Implement `removePasskey` method in `identityInfo`.
- Update `security/+page`:
   - Show delete button only if there is _more than 1_ access method.
   - Show delete button for passkeys and show confirmation dialog on click

# Tests

- Verified that the delete button is only shown when there is _more than 1_ access method.
- Verified that passkey can be removed successfully
- Verified that the _correct_ passkey is removed
- Verified that user is logged out and identity is removed from last used when the passkey is currently in use.

### DIFF
--- a/src/frontend/src/lib/stores/identity-info.state.svelte.ts
+++ b/src/frontend/src/lib/stores/identity-info.state.svelte.ts
@@ -185,11 +185,25 @@ class IdentityInfo {
       await actor
         .authn_method_remove(identityNumber, publicKey)
         .then(throwCanisterError);
+
+      if (
+        "WebAuthn" in authnMethod.authn_method &&
+        this.isCurrentAccessMethod({
+          passkey: {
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -160,6 +163,48 @@ class IdentityInfo {
     await throwCanisterError(googleRemoveResult);
   };
 
+  async removePasskey(): Promise<void> {
```
lmuntaner, line 166: I would have felt easier to expect the authnMethod in the parameter.

Even though we also have it in the state.

lmuntaner (approved):
nitpick, but already approving

PR #3189, opened 2025-07-15
Rename passkey

lmuntaner:
# Motivation

Allow users to rename their passkeys.

# Changes

* New component RenamePasskeyDialog to rename passkey.
* New method `renamePasskey` to IdentityInfo class.
* New util to get public key of WebAuthn method and moved helper to get aliast to utils.
* Add a new pencil button next to remove button on passkeys
* Add errors to handleError util.

# Tests

Tested locally see videos shared.

Tried also:
* Pressing "Esc" exits editing mode without saving.
* Pressing "Enter" saves new name.
* Tried with mobile screen and it looks good.


src/frontend/src/lib/utils/webAuthn.ts
```diff
@@ -93,3 +93,13 @@ export const authnMethodEqual = (
...
 };
+
+export const getPublicKey = (authnMethod: AuthnMethodData): Uint8Array => {
+  if ("WebAuthn" in authnMethod.authn_method) {
+    return new Uint8Array(authnMethod.authn_method.WebAuthn.pubkey);
+  }
+  if ("PubKey" in authnMethod.authn_method) {
+    return new Uint8Array(authnMethod.authn_method.PubKey.pubkey);
+  }
+  throw new Error("Unknown authn method");
```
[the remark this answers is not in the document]
lmuntaner: thansk!

src/frontend/src/lib/components/ui/EditText.svelte
```diff
@@ -0,0 +1,78 @@
...
+  const handleSave = async () => {
+    if (!isEditing || inputValue === text) {
+      isEditing = false;
+      return;
+    }
+
+    isEditing = false;
+    isLoading = true;
+    await onSave(inputValue);
+    isLoading = false;
```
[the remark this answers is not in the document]
lmuntaner: thanks!

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -205,6 +206,44 @@ class IdentityInfo {
...
+        newMetadata,
+      )
+      .then(throwCanisterError);
+    const index = this.authnMethods.findIndex((value) =>
+      authnMethodEqual(value, authnMethod),
+    );
+    this.authnMethods[index] = {
+      ...authnMethod,
+      metadata: newMetadata,
+    };
```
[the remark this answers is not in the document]
lmuntaner: thanks!

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -205,6 +206,48 @@ class IdentityInfo {
...
+    );
+    if (index < 0) {
+      throw new Error("Authentication method not found in authnMethods.");
+    }
+
+    this.authnMethods[index] = {
+      ...authnMethod,
+      metadata: newMetadata,
+    };
+    await this.fetch();
```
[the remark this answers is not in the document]
lmuntaner: Not true

src/frontend/src/lib/components/ui/AccessMethod.svelte
```diff
@@ -51,6 +54,15 @@
     return undefined;
   };
 
+  const renamePasskeyFactory =
```
sea-snake: Nit: Not sure if a factory pattern is more clear than a fn called with an arrow fn e.g.
```ts
const handleRenamePasskey = async (authnMethod: AuthnMethodData, name: string) => {
  try {
    await identityInfo.renamePasskey(authnMethod, name);
  } catch (error) {
    handleError(error);
  }
};

<EditName
  text={getAuthnMethodAlias(accessMethod) ?? ""}
  onSave={(name) => handleRenamePasskey(accessMethod, name)}
/>
```

src/frontend/src/lib/components/ui/EditText.svelte
```diff
@@ -0,0 +1,81 @@
...
+      bind:element={inputElement}
+      bind:value={inputValue}
+      onblur={handleSave}
+      onkeydown={handleKeyDown}
+      type="text"
+      disabled={isLoading}
+      size="sm"
+    />
+  {:else if isLoading}
+    <PlaceHolder class="w-30" />
```
sea-snake: Nit: Instead of a placeholder (which in my opinion indicates data isn't available yet), maybe showing the updated text with a slow blinking opacity to it similar to the placeholder is more clear?

src/frontend/src/lib/components/ui/EditText.svelte
```diff
@@ -0,0 +1,81 @@
...
+  {:else}
+    <span class="name-display">{text}</span>
+    <Button
+      variant="tertiary"
+      iconOnly
+      onclick={startEditing}
+      aria-label="Edit name"
+    >
+      <Edit size={16} />
```
sea-snake: Icons should scale with their button size based on screen scaling size instead of having an absolute size value.
```suggestion
      <Edit size={'1.25rem'} />
```

src/frontend/src/lib/components/utils/error.ts
```diff
@@ -126,6 +128,16 @@ export const handleError = (error: unknown) => {
         });
         console.error(error);
         break;
+      case "AuthnMethodNotFound":
```
sea-snake: Could you move these above `NameTooLong`? The errors below with comments and console logs are less specific more generic catch all cases.

I'd recommend to check if these two errors are expected or unexpected, in the latter case, you could add them to the less specific more generic catch all cases.

Overall, we still need to discuss our approach to error handling, so this comment can be considered mostly a nit.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -205,6 +206,48 @@ class IdentityInfo {
     }
   }
 
+  async renamePasskey(
+    authnMethod: AuthnMethodData,
+    newName: string,
+  ): Promise<void> {
+    if (newName === "") {
+      throw new CanisterError({ InvalidMetadata: "Name cannot be empty." });
```
sea-snake: Let's not throw canister errors that aren't actually coming from the canister, also I'd assume the canister will already throw this error for an empty passkey name to begin with. Combined with the fact that the UI shouldn't allow a user to confirm an empty name input, this wouldn't need the additional check in the frontend.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
```diff
@@ -205,6 +206,48 @@ class IdentityInfo {
...
+      .authn_method_metadata_replace(
+        identityNumber,
+        getPublicKey(authnMethod),
+        newMetadata,
+      )
+      .then(throwCanisterError);
+    const index = this.authnMethods.findIndex((value) =>
+      authnMethodEqual(value, authnMethod),
+    );
+    if (index < 0) {
```
sea-snake: Nit: you could probably use `find` instead of `findIndex` and update the metadata property directly and doing a `nonNullish` check.

src/frontend/src/lib/utils/webAuthn.ts
```diff
@@ -93,3 +93,15 @@ export const authnMethodEqual = (
   }
   return false;
 };
+
+export const getPublicKey = (authnMethod: AuthnMethodData): Uint8Array => {
```
sea-snake: Nit: I would call it `toPublicKey` or `authnMethodToPublicKey` instead to clarify we're mapping data instead of getting it from somewhere.

src/frontend/src/lib/components/ui/AccessMethod.svelte
lmuntaner: Not a strong opinion. They both create a function.

Maybe it's easier to understand yours?

src/frontend/src/lib/components/ui/EditText.svelte
lmuntaner: Oh, that would be cool. I'll wait until Erika approves this. We might do the rename with a modal...

src/frontend/src/lib/stores/identity-info.state.svelte.ts
lmuntaner: No, empty names are ok.

I can add the check in the input, yes, that makes more sense.

I'll wait until I have the UX confirmed.

src/frontend/src/lib/stores/identity-info.state.svelte.ts
lmuntaner: Good point

src/frontend/src/lib/utils/webAuthn.ts
lmuntaner: naming is hard... but I like yuor idea, yes

src/frontend/src/lib/components/utils/error.ts
```diff
@@ -118,6 +120,16 @@ export const handleError = (error: unknown) => {
...
+        toaster.error({
+          title: "Authentication method not found",
+        });
+        break;
+      case "InvalidMetadata":
+        toaster.error({
+          title: `Invalid metadata. ${error.value(error.type)}`,
```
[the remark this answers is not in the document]
lmuntaner: nops

src/frontend/src/lib/components/views/RenamePasskeyDialog.svelte
```diff
@@ -0,0 +1,88 @@
...
+    Give your passkey a memorable name to help you identify it.
+  </p>
+
+  <form onsubmit={handleSave} class="flex flex-col gap-6">
+    <Input
+      bind:element={inputElement}
+      bind:value={inputValue}
+      onkeydown={handleKeyDown}
+      type="text"
+      disabled={isLoading || inputValue.trim() === ""}
```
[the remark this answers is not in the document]
lmuntaner: Good catch!

lmuntaner:
@sea-snake UX suggested by design

src/frontend/src/lib/components/views/RenamePasskeyDialog.svelte
```diff
@@ -0,0 +1,88 @@
...
+<Dialog onClose={isLoading ? undefined : onClose}>
+  <FeaturedIcon variant="info" size="lg" class="mb-3">
+    <EditIcon size="1.5rem" />
+  </FeaturedIcon>
+  <h1 class="text-text-primary mb-3 text-2xl font-medium">Rename passkey</h1>
+  <p class="text-text-tertiary mb-6 font-medium">
+    Give your passkey a memorable name to help you identify it.
+  </p>
+
+  <form onsubmit={handleSave} class="flex flex-col gap-6">
```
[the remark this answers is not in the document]
lmuntaner: fixed!

sea-snake (approved):

src/frontend/src/lib/components/views/RenamePasskeyDialog.svelte
```diff
@@ -0,0 +1,89 @@
...
+  let inputElement = $state<HTMLInputElement>();
+
+  const handleSave = async (e: SubmitEvent) => {
+    e.preventDefault();
+    // Button is disabled so this shouldn't happen.
+    if (inputValue.trim() === "") {
+      return;
+    }
+
+    if (inputValue === currentName) {
```
sea-snake: Nit:

This instant UX (no "saving..." text shown for ~2s) since no request is made, probably feels off to the user, I would recommend removing this condition check, so the UX of the save button is always the same.

Updating a name to the same value it already was isn't a bad thing.

src/frontend/src/lib/components/views/RenamePasskeyDialog.svelte
```diff
@@ -0,0 +1,89 @@
...
+  <h1 class="text-text-primary mb-3 text-2xl font-medium">Rename passkey</h1>
+  <p class="text-text-tertiary mb-6 font-medium">
+    Give your passkey a memorable name to help you identify it.
+  </p>
+
+  <form onsubmit={handleSave} class="flex flex-col gap-6">
+    <Input
+      bind:element={inputElement}
+      bind:value={inputValue}
+      onkeydown={handleKeyDown}
```
sea-snake: Why is this needed? Isn't this already handled by the dialog itself?

lmuntaner: To exit on `ESC`

sea-snake: But that should already work without this additional code when you're in a dialog.

lmuntaner: ah, yes!

lmuntaner: done

PR #3190, opened 2025-07-15
Update register passkey flow

sea-snake:
Update register passkey flow

# Changes

- Handle invalid confirmation code.
- Render 6 digit code input instead of text input for confirmation code.
- Split `new-device` page into multiple views.
- Simplify `registerAccessMethodFlow`, only keep a boolean `unableToContinue`.
- Simplify the render of above error, render a error dialog instead similar to `authorize` flow.

# Tests

- Verified that invalid confirmation codes shows an error to the user.
- Verified that the `CodeInput` component works as expected (special keys, pasting and focus management).
- Verified that the error dialog with instructions is shown when the user can't continue on the new device.

### DIFF
--- a/src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
+++ b/src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
@@ -1,9 +1,12 @@
 <script lang="ts">
   import { onMount } from "svelte";
   import Button from "$lib/components/ui/Button.svelte";
-  import Input from "$lib/components/ui/Input.svelte";
   import ConfirmDeviceIllustration from "$lib/components/illustrations/ConfirmDeviceIllustration.svelte";
   import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
+  import CodeInput from "$lib/components/ui/CodeInput.svelte";
+  import { RotateCcwIcon } from "@lucide/svelte";
+
+  const CODE_LENGTH = 6;
 
   interface Props {
     confirm: (confirmationCode: string) => Promise<void>;
@@ -13,9 +16,9 @@
   const { confirm, restart }: Props = $props();
 
   let inputRef = $state<HTMLInputElement>();
-  let confirmationCode = $state("");
+  let confirmationCode = $state<string>("");
   let isConfirming = $state(false);
-  let isInvalidCode = $state(true);
+  let isInvalidCode = $state(false);
 
   const handleSubmit = async () => {
     isConfirming = true;
@@ -23,13 +26,15 @@
       await confirm(confirmationCode);
     } catch (error) {
       isInvalidCode = true;
+      confirmationCode = "";
+      inputRef?.focus();
     } finally {
       isConfirming = false;
     }
   };
 
   $effect(() => {
-    if (confirmationCode.length > 0) {
+    if (confirmationCode) {
       isInvalidCode = false;
     }
   });
@@ -54,19 +59,21 @@
     <br /><br />
     <b class="text-text-primary">Never</b> enter a code from another source.
   </p>
-  <Input
+  <CodeInput
+    bind:element={inputRef}
     bind:value={confirmationCode}
-    placeholder="XXXXXX"
-    size="md"
-    class="mb-8"
-    error={isInvalidCode ? "Invalid code, please try again" : undefined}
+    length={CODE_LENGTH}
+    class="mb-1"
   />
+  <div class="text-text-error-primary mb-3 h-5 text-sm">
+    {isInvalidCode ? "Invalid code. Please check and try again." : ""}
+  </div>
   <Button
     onclick={handleSubmit}
     variant="primary"
     size="xl"
     type="submit"
-    disabled={confirmationCode.length === 0 || isConfirming}
+    disabled={confirmationCode.length < CODE_LENGTH || isConfirming}
     class="mb-3"
   >
     {#if isConfirming}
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -11,6 +11,7 @@
     inputClass?: ClassValue;
     size?: Size;
     error?: string;
+    errorBorder?: boolean;
```
lmuntaner, line 14: If it has a default, it's not optional, right?

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
```diff
@@ -1,25 +1,44 @@
...
+      isInvalidCode = true;
+      confirmationCode = "";
+      inputRef?.focus();
+    } finally {
+      isConfirming = false;
+    }
   };
 
+  $effect(() => {
+    if (confirmationCode) {
```
lmuntaner, line 35: Don't you get a lint error? We are supposed to use `isNullish` to transform it to boolean for conditionals

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
```diff
@@ -40,24 +59,39 @@
...
   </p>
-  <Input
+  <CodeInput
+    bind:element={inputRef}
     bind:value={confirmationCode}
-    placeholder="XXXXXX"
-    size="md"
-    class="mb-8"
+    length={CODE_LENGTH}
+    class="mb-3"
```
lmuntaner, line 66: I would prefer to add an extra element to this one to manage the outer space. But something to discuss in our best practices list.

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
```diff
@@ -40,24 +59,39 @@
...
+    hint={"\u00a0"}
   />
   <Button
     onclick={handleSubmit}
     variant="primary"
     size="xl"
     type="submit"
-    disabled={confirmationCode.length === 0 || loading}
+    disabled={confirmationCode.length < CODE_LENGTH || isConfirming}
     class="mb-3"
```
lmuntaner, line 79: ditto

lmuntaner (approved):
Thanks! Looking good!

Some minor comments, but already approving!

src/frontend/src/lib/components/ui/Input.svelte
sea-snake: It still is optional, it uses the default when it's value is not defined. In most cases you'd use the `error` prop instead of `errorBorder` prop directly, unless you need to show the red error border without an error message, which is an exception case.

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
sea-snake: It's never nullish, it's a string with initial value set to `""`.

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
sea-snake: The `CodeInput` component has a single wrapping div where the margin in the class prop would be applied to. Am I missing something?

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
sea-snake: > I would prefer to add an extra element to this one to manage the outer space.
I'd assume you're referring to this. 

Maybe you mean an extra element around the buttons, titles, paragraphs etc, to use a gap instead of margin on the elements themselves? I did consider that, but given that all elements are vertically stacked, it seems a bit unnecessary to group things together just because they have different gaps between one another.

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
lmuntaner: Yes, I don't like the props to manage outer space of a component.

Instead, I prefer to do:

```
<div class="mb-3">
  <CodeInput ... />
</div>
```

Or use flex and gap in the parent, or grid and gap. But have that in the parent and not use props for that.

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
lmuntaner: I meant more to have "props" to change outer space. Outer space should not be the responsibility of the component (with or without props) but always of the parent.

lmuntaner: Using flex or margins it's ok for me. It's where this is handled that I mean.

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
sea-snake: Ah like that, I actually prefer to add margins directly on the component props than adding a wrapper, which is why I always consider it frustrating when component don't expose their class/style prop for example.

But I prefer flex with gap over margins indeed, let's do that in another PR though so we can solely focus on that in that other PR (also including existing view components within this flow).

src/frontend/src/lib/components/views/AuthorizeNewDevice.svelte
sea-snake: I saw your other comment, will do it in another PR though.

PR #3195, opened 2025-07-17
Fix alternative origins for 2.0

lmuntaner:
# Motivation

Derivation origin isn't working on id.ai because it tries to use id.ai as a host to get the well-known file.

This is due to the logic to infer the alternative origins url.

# Changes

* Add "id.ai" to the list of hostnames that use the expected `<canister-id>.icp0.io<alternative-origins-path>`

Relying on the host name is technical debt that we should fix.

# Tests

* Changed and added unit tests.

## Manual test

Deployed to beta.

Successful
1. Go to `https://o6tmp-fqaaa-aaaad-aak5q-cai.icp0.io` and add `https://vt36r-2qaaa-aaaad-aad5a-cai.icp0.io` as alternative origin
2. Go to `https://vt36r-2qaaa-aaaad-aad5a-cai.icp0.io` and login with derivation origin`https://o6tmp-fqaaa-aaaad-aak5q-cai.icp0.io`.
3. User is able to log in.
4. I did this test with beta.id.ai, beta.identity.internetcomputer.org and beta.identity.ic0.app

Error
1. Go to `https://vt36r-2qaaa-aaaad-aad5a-cai.icp0.io` and login with derivation origin`https://nns.ic0.app/`.
2. Error showed in UI
3. I did this test with beta.id.ai, beta.identity.internetcomputer.org and beta.identity.ic0.app

### DIFF
--- a/src/frontend/src/lib/utils/validateDerivationOrigin.ts
+++ b/src/frontend/src/lib/utils/validateDerivationOrigin.ts
@@ -140,7 +140,8 @@ const inferAlternativeOriginsUrl = ({
   if (
     location.hostname.endsWith("icp0.io") ||
     location.hostname.endsWith("ic0.app") ||
-    location.hostname.endsWith("internetcomputer.org")
+    location.hostname.endsWith("internetcomputer.org") ||
+    location.hostname.endsWith("id.ai")
   ) {
     // If this is a canister running on one of the official IC domains, then return the
     // official canister id based API endpoint
--- a/src/frontend/src/lib/utils/validateDerivationOrigin.ts
+++ b/src/frontend/src/lib/utils/validateDerivationOrigin.ts
@@ -137,17 +137,6 @@ const inferAlternativeOriginsUrl = ({
     return `https://${canisterId.toText()}.${IC_HTTP_GATEWAY_DOMAIN}${ALTERNATIVE_ORIGINS_PATH}`;
   }
 
-  if (
-    location.hostname.endsWith("icp0.io") ||
-    location.hostname.endsWith("ic0.app") ||
-    location.hostname.endsWith("internetcomputer.org") ||
-    location.hostname.endsWith("id.ai")
-  ) {
-    // If this is a canister running on one of the official IC domains, then return the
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/utils/validateDerivationOrigin.ts
```diff
@@ -140,7 +140,8 @@
   if (
     location.hostname.endsWith("icp0.io") ||
     location.hostname.endsWith("ic0.app") ||
-    location.hostname.endsWith("internetcomputer.org")
+    location.hostname.endsWith("internetcomputer.org") ||
```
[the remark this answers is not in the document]
lmuntaner: Done

src/frontend/src/lib/utils/validateDerivationOrigin.ts
```diff
@@ -162,9 +152,6 @@ const inferAlternativeOriginsUrl = ({
     return `${location.protocol}//${location.host}${ALTERNATIVE_ORIGINS_PATH}?canisterId=${canisterId}`;
   }
 
-  // Otherwise assume it's a custom setup expecting the gateway to
```
lmuntaner, line 165: I don't see the reason to have the mainnet hostnames hardcoded.

When we have a use case for the custom setup, we can discuss it.

sea-snake: Wouldn't utopia be an example use case? Are you sure removing the custom setup doesn't break it?

lmuntaner: I asked them.

If they use the canister id as query param yes it would.

lmuntaner: Confirmed with Utopia team that they are not using this feature and that if used, the way to get the file would be different than the fallback used here.

It's <canister-id>.<domain> where domain is different and would need to be added in the config of II if we want to support it.

Therefore, the right solution is to move `IC_HTTP_GATEWAY_DOMAIN` to config of II.

PR #3196, opened 2025-07-17
E2E tests for adding passkeys from the dashboard

lmuntaner:
# Motivation

Make sure we don't break current functionality.

In this PR, two new E2E tests are added:
* `User can log in to the dashboard and add a new passkey from the same device`
* `User can log in to the dashboard and add a new passkey from another device`

# Changes

* Add aria-labels to make content selectable by label.
* Add E2E helper to create identity in II landing page.
* Add two E2E tests.

### DIFF
--- a/src/frontend/tests/e2e-playwright/dashboard/addPasskeys.spec.ts
+++ b/src/frontend/tests/e2e-playwright/dashboard/addPasskeys.spec.ts
@@ -12,7 +12,9 @@ test("User can log in the dashboard and add a new passkey from the same device",
   page,
 }) => {
   const auth = dummyAuth();
+  await page.goto(II_URL);
   await createNewIdentityInII(page, TEST_USER_NAME, auth);
+  await page.waitForURL(II_URL + "/manage");
   await clearStorage(page);
   await page.goto(II_URL);
   await page.getByRole("button", { name: "Continue with Passkey" }).click();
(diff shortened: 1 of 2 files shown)


src/frontend/tests/e2e-playwright/dashboard/addPasskeys.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  // Start the "add passkey" flow
+  await page.getByRole("button", { name: "Add" }).click();
+  await page.getByRole("button", { name: "Continue with Passkey" }).click();
+  await page
+    .getByRole("button", { name: "Continue on another device" })
+    .click();
+
+  const linkToPair = await page.getByLabel("Pairing link").innerText();
+
+  const linkPage = await context.newPage();
```
sea-snake: I would create a new context instead of a new page, that would mimic the concept of a new device better, I think you'd still have browser data shared across pages within the same context.

https://playwright.dev/docs/api/class-browsercontext

src/frontend/tests/e2e-playwright/utils.ts
```diff
@@ -56,6 +56,27 @@ export const authorize = async (
...
 };
 
+/**
+ * Creates a new identity in II page assuming there is no stored identity
+ *
+ * @param page
+ * @param name
+ * @param dummyAuth
+ */
+export const createNewIdentityInII = async (
```
sea-snake, line 66: Isn't this method doing the same as `createIdentity` below? I know that the latter does it within the authorization flow, since we didn't have landing page at the time, but that doesn't make any difference from the outside.

lmuntaner: Good point, I kept this helper with the code from the other one and used it in `createIdentity`.

We might want to rename them, but I'll do that in another PR to keep it clean.

src/frontend/tests/e2e-playwright/dashboard/addPasskeys.spec.ts
lmuntaner: done!

PR #3197, opened 2025-07-17
Refactor authentication screens into a single AuthWizard

sea-snake:
Refactor authentication screens into a single AuthWizard.

# Changes

- Create `AuthWizard` component that can be used either outside or within a dialog.
- Remove callbacks from `AuthFlow` constructor, instead return identity number from method calls.
- Make state variables in `AuthFlow` read only, state changes should always go through a method call.
- Remove `AuthDialog`, instead use the `AuthWizard` within a dialog using `isAuthenticating` binding to update dialog close controls.
- Co-locate `AuthWizard` specific views.

# Tests

- Verified that both authentication and authorization flows are still working as expected (sign in/up).
- Verified that sign up with captcha works as expected.
- Verified that switching identities works as expected.


src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -50,6 +43,11 @@
   );
 
   let isAuthDialogOpen = $state(false);
+  let isAuthenticating = $state(false);
+
+  $effect(() => {
+    console.log("isAuthenticating", isAuthenticating);
```
lmuntaner: To be removed I guess 😅

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -63,89 +61,86 @@
   <div class="flex flex-1 flex-col items-center justify-center">
     <AuthPanel class="sm:max-w-100">
       <div class="flex-1"></div>
-      {#if nonNullish(authFlow.captcha)}
-        <SolveCaptcha {...authFlow.captcha} />
-      {:else}
+      {#if lastUsedIdentities.length > 0}
         <h1 class="text-text-primary my-2 self-start text-2xl font-medium">
```
lmuntaner, line 61: It would be nice to have this as a separate view as well. For readibility purposes mostly.

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -63,89 +61,86 @@
...
-              {/if}
-            </Dialog>
-          {/if}
+              onError={(error) => {
+                isAuthDialogOpen = false;
+                handleError(error);
+              }}
+              withinDialog
+            >
+              <h1
```
lmuntaner, line 120: The children of `AuthWizard` is the header at the top, right?

Would it make it clearer if we used a snippet "header" instead of children?

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
```diff
@@ -0,0 +1,105 @@
...
+  {#if authFlow.view === "chooseMethod" || !withinDialog}
+    {@render children?.()}
+    <PickAuthenticationMethod
+      setupOrUseExistingPasskey={authFlow.setupOrUseExistingPasskey}
+      continueWithGoogle={handleContinueWithGoogle}
+    />
+  {/if}
+  {#if authFlow.view !== "chooseMethod"}
+    {#if !withinDialog}
+      <Dialog
```
lmuntaner, line 90: I understand this is the UX, but doesn't this raise an issue that only some part of the flow is rendered within a Dialog? Instead of all or nothing?

I mean, shouldn't it be instead a wizard that users can move back and forth instead of closing a dialog?

Not for this PR, just an open question.

lmuntaner (approved):
Did you test switching identity?
Did you test the captcha?

Very nice refactor!

I added a nitpick, a comment for the future and please remove a console.log, hehe.

But already approving.

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
sea-snake: As discussed, not in this PR.

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
```diff
@@ -0,0 +1,105 @@
...
+  const handleContinueWithExistingPasskey = async () => {
+    isAuthenticating = true;
+    try {
+      onSignIn(await authFlow.continueWithExistingPasskey());
+    } catch (error) {
+      onError(error);
+    } finally {
+      isAuthenticating = false;
+    }
+  };
```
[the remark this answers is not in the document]
sea-snake: We assume that the `onError` method never throws itself

PR #3198, opened 2025-07-17
Andri/wave background

LXIF:
# Motivation

We want to add some flair.

# Changes

Build preliminary wave background. Demo available at `http://localhost:5173/fancy-backgrounds/wave-grid`

<video src="https://github.com/user-attachments/assets/2a81ff13-ad3d-464c-868e-3781db3744c1" />

# Tests

Moved and clicked my mouse.


LXIF:
> Looks super cool!
> 
> One question, did you consider doing it with ?
> 
> I'm thinking that for some of the other examples (like the matrix one) a canvas would be better suited and I wonder whether we could use have always one canvas background and change the animation there.
> 
> I think that regarding performance with the current example the DOM manipulation is fine, but I was wondering how you decided to do it with DOM elements.

I replaced it with SVG for now - canvas is also a possibility but unfortunately a bit finicky at times, especially if you want to get pixel-perfect rendering on all screens. But refactoring for canvas should not be too hard if we do decide for it. SVG still has performance benefits over divs without the overhead and imperative style of canvas.

lmuntaner:
Already merged in a few PRs

PR #3206, opened 2025-07-22
Create register access method wizard

sea-snake:
Create re-usable register access method wizard.

# Changes

- Create new `RegisterAccessMethodWizard` from `routes/pair/+page.svelte`.
- Refactor `RegisterAccessMethodFlow`.
- Add `WaitingForExistingDevice` view to the wizard.
- Add `ContinueOnExistingDevice` view to the wizard for when the flow isn't started from the dashboard.

# Tests

- Verified that the continue on another device flow in the dashboard is still working as expected.

### DIFF
--- a/src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
@@ -1,5 +1,5 @@
 import { anonymousActor, canisterConfig } from "$lib/globals";
-import { isCanisterError, throwCanisterError, waitFor } from "$lib/utils/utils";
+import { secureRandomId, throwCanisterError, waitFor } from "$lib/utils/utils";
 import { passkeyAuthnMethodData } from "$lib/utils/authnMethodData";
 import { authenticateWithSession } from "$lib/utils/authentication";
 import { sessionStore } from "$lib/stores/session.store";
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
```diff
@@ -15,24 +15,58 @@ import { DiscoverableDummyIdentity } from "$lib/utils/discoverableDummyIdentity"
...
+    const registrationId = existingRegistrationId ?? secureRandomId(5);
+    if (isNullish(existingRegistrationId)) {
+      this.#existingDeviceLink = new URL(
+        `/activate#${registrationId}`,
+        window.location.origin,
+      );
     }
+    this.#view = "continueOnExistingDevice";
+
+    const expiration = new Date(Date.now() + 300000).getTime(); // 5 min
```
lmuntaner, line 55: Shall we add the units?

src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
```diff
@@ -15,24 +15,58 @@ import { DiscoverableDummyIdentity } from "$lib/utils/discoverableDummyIdentity"
...
+  }
+
+  get existingDeviceLink() {
+    return this.#existingDeviceLink;
+  }
+
+  waitForExistingDevice = async (existingRegistrationId?: string) => {
+    const registrationId = existingRegistrationId ?? secureRandomId(5);
+    if (isNullish(existingRegistrationId)) {
+      this.#existingDeviceLink = new URL(
```
lmuntaner, line 48: Won't this show the link in the flow started from the dashboard?

src/frontend/src/lib/components/wizards/registerAccessMethod/RegisterAccessMethodWizard.svelte
```diff
@@ -0,0 +1,54 @@
...
+      await registerAccessMethodFlow.waitForExistingDevice(registrationId);
+    } catch (error) {
+      onError(error);
+    }
+  });
+</script>
+
+{#if registerAccessMethodFlow.view === "continueOnExistingDevice"}
+  {#if registerAccessMethodFlow.existingDeviceLink}
+    <ContinueOnExistingDevice
```
lmuntaner, line 39: Why are `ContinueOnExistingDevice` and `WaitingForExistingDevice` the same view?

src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
```diff
@@ -41,26 +75,14 @@ export class RegisterAccessMethodFlow {
...
-            this.isUnableToComplete = true;
-            return;
-        }
-      }
-      throw error;
-    }
+    const confirmation = await anonymousActor
+      .authn_method_register(identityNumber, authnMethodData)
+      .then(throwCanisterError);
+    const expiration = confirmation.expiration;
```
lmuntaner, line 81: Can we add units here?

src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
sea-snake: Both flows have an optional `existingRegistrationId` param, if not set, it will generate a registration id and show the QR code. If set, it will use it as is instead and not show the QR code but a waiting screen instead.

This basically means that the `RegisterAccessMethodFlow/Wizard` and `ConfirmAccessMethodFlow/Wizard` can be used in both directions (starting on authorized or unauthorized device).

src/frontend/src/lib/components/wizards/registerAccessMethod/RegisterAccessMethodWizard.svelte
sea-snake: See my comment here: https://github.com/dfinity/internet-identity/pull/3206#discussion_r2224742052

I'll add some more code comments at a later point (follow up PRs) to all flows and wizards to clarify these things.

src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
lmuntaner: Ah, wait, now I read well the `if (isNullish(existingRegistrationId))` that is pointing to the parameter, not the `registrationId` that it will always be present. Got it!

src/frontend/src/lib/components/wizards/registerAccessMethod/RegisterAccessMethodWizard.svelte
lmuntaner: I understand that, but should they be considered the same view and then render a different component depending on another parameter?

Maybe it depends on what we consider a "view". If it was a step, I understand is the same step. Maybe we could rename it like that.

PR #3209, opened 2025-07-23
Redirect activate links to dashboard

sea-snake:
Redirect activate links to dashboard.

# Changes

- Add `CONTINUE_FROM_ANOTHER_DEVICE` feature flag.
- Add `/activate#{code}` route that redirects to `/manage?activate={code}`.
- Add `next` search param when navigating to `/manage/**` while not authenticated.
- Navigate to `next` search param after authentication.

# Tests

- Verified that authentication from the landing page still redirects to `/manage` by default.
- Verified that `/activate#{code}` redirects to `/manage?activate={code}` after authentication on the landing page.

### DIFF
--- a/src/frontend/src/lib/state/featureFlags.ts
+++ b/src/frontend/src/lib/state/featureFlags.ts
@@ -88,6 +88,11 @@ export const ADD_ACCESS_METHOD = createFeatureFlagStore(
   true,
 );
 
+export const CONTINUE_FROM_ANOTHER_DEVICE = createFeatureFlagStore(
+  "CONTINUE_FROM_ANOTHER_DEVICE",
+  false,
+);
+
 export default {
   DOMAIN_COMPATIBILITY,
   OPENID_AUTHENTICATION,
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/state/featureFlags.ts
lmuntaner, line 1: Is this related to the params?

lmuntaner (approved):
This PR has three features, right?

* "next" param to redirect after authentication. Why is this needed for this flow?
* New feature flag.
* Redirect from "/activate"

The changes are fine, I am just confused how these are related to adding a passkey.

src/frontend/src/lib/state/featureFlags.ts
sea-snake: > I am just confused how these are related to adding a passkey.

To clarify, when starting the flow from an unauthenticated device, the user will be shown a `/activate#{code}` QR code and link.

This PR makes sure that this link is handled and redirected to the dashboard, after a user has first signed into the dashboard. Since we first have to sign in, the next param makes sure we're going back to the original intended location we tried to reach before being authenticated.

lmuntaner: Aaaahh, instead of adding an authentication flow there, you will redirect them there after requiring them to log in. Noted!

PR #3210, opened 2025-07-23
Continue from another device

sea-snake:
Continue from another device

# Changes

- Add "Continue from another device" button to `PickAuthenticationMethod` view.
- Render `RegisterAccessMethodWizard` in `AuthWizard` when the above is pressed.
- Add `pendingRegistrationId` to `/manage` page props.
- Use above prop to render dialog with `ConfirmAccessMethodWizard`.

# Tests

- Verified that the existing authentication flows are still working as intended.
- Verified that the `Continue from another device` flow works as intended (it doesn't directly log you in yet).

https://github.com/user-attachments/assets/71356b95-80f2-44e3-bb3f-989880fa2454

### DIFF
--- a/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
+++ b/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
@@ -25,7 +25,7 @@
     isAuthenticating = $bindable(),
     onSignIn,
     onSignUp,
-    onOtherDevice,
+    onOtherDevice = () => {},
     onError,
     withinDialog = false,
     children,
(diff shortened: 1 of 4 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/+page.svelte
```diff
@@ -29,3 +48,19 @@
     <AccessMethods />
   </div>
 </div>
+
+{#if $CONTINUE_FROM_ANOTHER_DEVICE && nonNullish(pendingRegistrationId)}
+  <Dialog onClose={() => (pendingRegistrationId = null)}>
```
lmuntaner, line 53: Should we keep onClose undefined until the end to avoid users closing it in the middle of the flow?

sea-snake: I wouldn't since a user should be able to cancel the flow if they intend to not continue with it.

lmuntaner: Ok, but we shuold show a warning in that case, or use a specific "Cancel" button, not the "X" button to close the modal or clicking outside the modal withuot wanting to

sea-snake: I think that's for another design iteration over this current implementation, currently the user could always restart the flow (scan qr / enter URL manually).

PR #3211, opened 2025-07-23
Andri/wave background in flow

LXIF:
# Motivation

We want to delight our users with flair.

# Changes

Add flair canvas component, add it to our login screen with animation on login/signup success

<video src="https://github.com/user-attachments/assets/96193667-cd36-4cdf-a906-69ffec8c456f" />

# Tests

- [x] Tried signup flow
- [x] Tried signin flow


src/frontend/src/routes/(new-styling)/authorize/+page.svelte
```diff
@@ -19,9 +21,29 @@
...
+  const preloadSignIn = () => {
+    void preloadCode("/authorize/account");
+    void preloadData("/authorize/account");
+  };
   const onSignIn = async (identityNumber: bigint) => {
     lastUsedIdentitiesStore.selectIdentity(identityNumber);
-    await goto("/authorize/account");
+    preloadSignIn();
+    if (triggerAnimation) {
+      triggerAnimation({
```
lmuntaner: Probably under your radar already, but I would create a set of predefined animations and use those instead of requiring all the params every time:

```
triggeraAnimation(DROP_WAVE);
```

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -62,17 +82,63 @@
...
+
+        motionType: "omni",
+        intensity: "light",
+        speed: 5,
+        nImpulses: "double",
+        size: "large",
+        impulseEasing: "cubicIn",
+      });
+    }
+    setTimeout(animateAuthenticating, 1200);
```
lmuntaner: Should we also have a constant for these waiting times:

```
setTimeout(animateAuthenticating, AUTHORIZING_DELAY_MILLIS);
```

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -52,9 +91,18 @@
     await authLastUsedFlow.authenticate(identity).catch(handleError);
     await onSignIn(identity.identityNumber);
   };
+
+  let triggerAnimation = $state<(opts: FlairAnimationOptions) => void>();
 </script>
 
 <div class="flex min-h-[100dvh] flex-col">
+  <FlairCanvas
```
lmuntaner: This `FlairCanvas` is used always with the same parameters now.

Do we need to expose them to the business related components?

I'm thinkin I'd rather have a separate layer handling that and the business components do not care about these parameters.

This way, if we want parameters to change randomly on every interaction, this other layer should take care of initializing with some different parameters.

src/frontend/src/routes/(new-styling)/authorize/+page.svelte
```diff
@@ -19,9 +21,29 @@
...
+        intensity: "light",
+        speed: 5,
+        nImpulses: "single",
+        size: "large",
+        impulseEasing: "cubicIn",
+      });
+    }
+    setTimeout(async () => {
+      await goto("/authorize/account");
+    }, 700);
```
lmuntaner, line 26: Ditto to convert to constant: `DELAY_ANIMATED_REDIRECT_MILLIS`

lmuntaner (commented):
Looking super good!

As a more architectural question, could we split where the component is rendered and where the animation is triggered?

It would be good that it works like the toaster component. It's always there and any component can trigger it.

Not necessarily in this PR, but as a comment for later.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
LXIF: This can of course be done, yes. It's a good idea especially if we want to have specific animations for specific states/actions.

src/frontend/src/routes/(new-styling)/+page.svelte
LXIF: The parameters are all optional. Personally, I am strongly against parameters changing randomly on every interaction - if so, this should be within the component for my taste. But ultimately either way is fine by me.

LXIF:
Adjusted parameters and animations as per UI designer guidance. The signin/authorize flows now look like this:

<video src="https://github.com/user-attachments/assets/2fb1b8fd-5918-4356-8eb7-ff08d6f5a4bd" />

<video src="https://github.com/user-attachments/assets/867091c9-b1ab-42da-8c97-15d62401fa34" />

Also, I have added a feature flag called `FLAIR` which enables/disables the FlairCanvas inside the WaveCanvas.

src/frontend/src/lib/components/backgrounds/WaveCanvas.svelte
```diff
@@ -0,0 +1,33 @@
...
+    enableRandomOpacity={false}
+    pointSizeNoiseScale="medium"
+    pointSizeNoiseMultiplier="medium"
+    springOrTween={{
+      type: "spring",
+      stiffness: "medium",
+      dampening: "medium",
+    }}
+    customColorMode="all"
+    customColor="#ECFF57"
```
lmuntaner: Is this color ok for dark and light mode?

src/frontend/src/lib/components/backgrounds/WaveCanvas.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+  import FlairCanvas from "./FlairCanvas.svelte";
+  import { FLAIR } from "$lib/state/featureFlags";
+
+  let { triggerAnimation = $bindable() } = $props();
+</script>
+
+{#if $FLAIR}
+  <FlairCanvas
```
lmuntaner, line 10: By default this doesn't render any dots? Only on animation?

Which is the parameter that sets that?

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
lmuntaner: Will you add a constant for this?

lmuntaner: Is there a recursive call here? Why?

src/frontend/src/routes/(new-styling)/authorize/+page.svelte
lmuntaner: Ditto

src/frontend/src/lib/components/backgrounds/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  target: ["motion"],
+  motionType: "omni",
+  intensity: "medium",
+  speed: "slow",
+  nImpulses: "single",
+  size: "large",
+  impulseEasing: "cubicIn",
+};
+
+export const WAVE_ANIMATION_DELAY_MILLIS = 1500;
```
lmuntaner: This doesn't seem enough time now... it feels short and it's not clear whether the animation finishes or not before being redirected.

Did Misha look into these parameters as well?

Next week we shuold work on converting the animation to an async function that we can await. Meanwhile, let's add something here that shows the whole animation.

That said, if we need to wait too long, maybe the animation should be shortened?

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
LXIF: Removed, it was for loading state but was decided to not be there.

src/frontend/src/lib/components/backgrounds/WaveCanvas.svelte
LXIF: This was in Misha's screenshots he sent me. Is changed now.

src/frontend/src/lib/components/backgrounds/WaveCanvas.svelte
LXIF: visibility set to maskwave means that it's only visible on animation.

PR #3212, opened 2025-07-24
Add more E2E to test different scenarios

lmuntaner:
# Motivation

Ensure we don't break any functionality in the future.

In this PR, new E2E tests from 1.0 have been ported to 2.0:
* Checking same principal is given by alternative origins.
* Setting the delegation expiration.
* Edge cases of handling post messages from the application.

# Changes

* New test on alternativeOrigins.spec.ts
* New test files delegationTtl.spec.ts and postMessage.spec.ts

### DIFF
--- a/src/frontend/tests/e2e-playwright/authorize/alternativeOrigins.spec.ts
+++ b/src/frontend/tests/e2e-playwright/authorize/alternativeOrigins.spec.ts
@@ -203,3 +203,79 @@ test("Should not follow redirect returned by /.well-known/ii-alternative-origins
   // Verify error message is displayed in II (redirect should not be followed)
   await expect(authPage.getByText("Unverified origin")).toBeVisible();
 });
+
+test("Should issue the same principal to nice url and canonical url", async ({
+  page,
+}) => {
+  // Create a new identity using dummy auth
+  const auth = dummyAuth();
+
+  // First authentication: Test app configured with canonical URL as derivation origin
+  await page.goto(TEST_APP_URL);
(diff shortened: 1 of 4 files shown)


src/frontend/tests/e2e-playwright/authorize/alternativeOrigins.spec.ts
```diff
@@ -203,3 +203,79 @@ test("Should not follow redirect returned by /.well-known/ii-alternative-origins
...
+
+  // Use existing passkey
+  auth(authPage2);
+  await authPage2.getByRole("button", { name: "Primary account" }).click();
+
+  // Wait for II window to close
+  await authPage2.waitForEvent("close");
+
+  // Get the second principal
+  const principal2 = await page.locator("#principal").textContent();
```
sea-snake, line 278: Are we sure we won't have a possible flaky issue here? I mean it could grab the text content just before it's updated instead of after right?

Maybe the following waits (with a timeout) till the text matches?

```ts
await page.locator("#principal").toHaveText(principal1);
```

sea-snake: Edit:

It seems like we're not doing this in the `authorize` util either.

The Playwright docs seems to explicitly mention the risk of flakiness: https://playwright.dev/docs/api/class-locator#locator-text-content

lmuntaner: I'll add `await expect(page.locator("#principal")).not.toBeEmpty();` before getting the content.

I also added it to the authorize util

sea-snake (approved):

PR #3213, opened 2025-07-24
Implement e2e test for "Continue from another device" flow.

sea-snake:
Implement e2e test for "Continue from another device" flow.

# Changes

- Add "Sign in from another device" to `index.spec.ts`.
- Add "Sign in from another device" to `authorize/index.spec.ts`.
- Fix issue found while adding the above: make sure to go back to authorize screen after creating passkey.

# Tests

- Visually verified the e2e test from the landing page behaved as expected and passed.
- Visually verified the e2e test from the authorize page behaved as expected and passed.

### DIFF
--- a/src/frontend/tests/e2e-playwright/utils.ts
+++ b/src/frontend/tests/e2e-playwright/utils.ts
@@ -3,6 +3,7 @@ import { Principal } from "@dfinity/principal";
 import { readCanisterId } from "@dfinity/internet-identity-vite-plugins/utils";
 
 const testAppCanisterId = readCanisterId({ canisterName: "test_app" });
+export const FEATURE_FLAGS = "?feature_flag_continue_from_another_device=true";
 export const II_URL = "https://id.ai";
 export const TEST_APP_URL = "https://nice-name.com";
 export const NOT_TEST_APP_URL = "https://very-nice-name.com";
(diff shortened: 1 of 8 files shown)


lmuntaner (commented):
Could you add a test from an appliction during the authentication flow?

sea-snake:
> Could you add a test from an appliction during the authentication flow?

Good idea, I'll add it.

src/frontend/tests/e2e-playwright/utils.ts
```diff
@@ -3,6 +3,7 @@ import { Principal } from "@dfinity/principal";
 import { readCanisterId } from "@dfinity/internet-identity-vite-plugins/utils";
 
 const testAppCanisterId = readCanisterId({ canisterName: "test_app" });
+export const FEATURE_FLAGS = "?feature_flag_continue_from_another_device=true";
```
lmuntaner, line 6: Should it be "CONTINUE_FROM_ANOTHER_DEVICE_FLAG"? and without the "?" in case we want to concatenate them?

lmuntaner (approved):
Thanks!

src/frontend/tests/e2e-playwright/utils.ts
sea-snake: This variable matches a possible value from the `search` param in `URL`, if we want to have multiple we could add & within this string value.

Overall, this is a temporary solution, when either the feature flag default becomes true or feature flags are moved to the canister config (means we can deploy the canister in the pipeline with true), we can remove this.

PR #3214, opened 2025-07-25
Change credential id param to list of credentials

lmuntaner:
# Motivation

We want to support multiple credentials being passed during the "Continue As" flow.

Therefore, we need to support passing a list of credentials instead of a single one in the helper functions.

This PR doesn't change any functionality, it just changes the parameter credential id from one to an array credentials. This way, we will be able to fetch the credentials based on the identity number and pass them for authentication.

# Changes

  - Updated `authenticateWithPasskey` to accept `credentialIds` (an array) instead of a single `credentialId`. This change is reflected in both the function signature and its internal logic.
  - Modified `DiscoverablePasskeyIdentity.useExisting` to handle `credentialIds` and updated the `credentialRequestOptions` accordingly. (`src/frontend/src/lib/utils/discoverablePasskeyIdentity.ts`
* **Credential Request Options**:
  - Updated `requestOptions` to map over `credentialIds` and create an array of `allowCredentials` if `credentialIds` are provided. This ensures compatibility with multiple passkeys.

* **Integration with Frontend Flow**:
  - Adjusted the `AuthLastUsedFlow` component to pass `credentialIds` to the authentication function, ensuring the updated API is utilized. (`src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts`

# Tests

No change in functionality.

### DIFF
--- a/src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
@@ -23,7 +23,7 @@ export class AuthLastUsedFlow {
         const { identity, identityNumber } = await authenticateWithPasskey({
           canisterId,
           session: get(sessionStore),
-          credentialId: lastUsedIdentity.authMethod.passkey.credentialId,
+          credentialIds: [lastUsedIdentity.authMethod.passkey.credentialId],
         });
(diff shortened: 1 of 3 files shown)


LXIF (approved):
LGTM!

PR #3216, opened 2025-07-25
PoC: Add the concept of sessions to the device registration state.

sea-snake:
Add the concept of sessions to the device registration state.

# Changes

### Add additional variants

- Add `SessionTentativelyAdded` and `SessionTentativelyConfirmed` to `RegistrationState`.

### Merge v1 and v2 methods

  - Merge `enter_device_registration_mode` and `enter_device_registration_mode_v2` since we can't differentiate between v1 and v2 either way in e.g. the device confirmation method.
  - Having a v2 concept for only part of the flow would be unclear.
- Refactor the above method to avoid id comparison (since id is optional), instead we simply check if an entry already exists.
- Merge `prune_expired_tentative_device_registrations` and `prune_expired_tentative_device_registrations_v2`, making sure the same data is pruned everywhere (the old method was used in some places that v2 should have been used).
    
### Fix existing method and add session variant
    
- Refactor `add_tentative_device` to handle additional states, explicitly only allow adding a tentative device when the current state is `DeviceRegistrationModeActive`.
- Add `add_tentative_session`, mirroring the above (not used yet).

### Fix device confirmation to also work with session

- Merge `verify_tentative_device` and `get_tentative_device` into new `confirm_tentative_device` method since this concept has changed with sessions.
- Refactor the above method to support sessions.
- Remove `IdentityUpdateError` variant from `VerifyTentativeDeviceError` since we no longer directly add the device directly, instead we return the confirmed device that should be added.

### Cleanup

- Add `Unauthorized(Principal)` to `AuthnMethodConfirmationError` canister error (v2 method error).
- Rename verification code to confirmation code for clarity in various places (while avoiding changing it in v1).

# Not implemented in this PR

- Add `authn_method_session` canister method to set `RegistrationState` to `SessionTentativelyAdded`.
- Update `authn_method_registration_mode_exit` to accept optional device data and move session to temp keys.

# Tests

TODO: Manually verify that existing device registration flow still works as expected.

### DIFF
--- a/src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
+++ b/src/internet_identity/tests/integration/v2_api/authn_method_registration.rs
@@ -882,7 +882,7 @@ fn should_reject_second_registration_with_different_id() -> Result<(), CallError
 }
 
 #[test]
-fn should_return_registration_with_same_id() -> Result<(), CallError> {
+fn should_reject_second_registration_with_same_id() -> Result<(), CallError> {
     let env = env();
     let canister_id = install_ii_with_archive(&env, None, None);
     let authn_method = test_authn_method();
@@ -890,7 +890,7 @@ fn should_return_registration_with_same_id() -> Result<(), CallError> {
     let first_registration_id = "0fZr4".to_string();
 
     // First call with the first registration ID should succeed
-    let result_1 = api_v2::authn_method_registration_mode_enter(
+    api_v2::authn_method_registration_mode_enter(
         &env,
         canister_id,
         authn_method.principal(),
@@ -899,16 +899,18 @@ fn should_return_registration_with_same_id() -> Result<(), CallError> {
     )?
     .expect("First authn_method_registration_mode_enter failed");
 
-    // Second call with a different registration ID should fail with InvalidRegistrationId
-    let result_2 = api_v2::authn_method_registration_mode_enter(
+    // Second call with a same registration ID should fail with AlreadyInProgress
+    let result = api_v2::authn_method_registration_mode_enter(
         &env,
         canister_id,
         authn_method.principal(),
         identity_number,
         Some(first_registration_id.clone()),
-    )?
-    .expect("Second authn_method_registration_mode_enter failed");
+    )?;
 
-    assert_eq!(result_1, result_2);
+    assert_eq!(
+        result,
+        Err(AuthnMethodRegistrationModeEnterError::AlreadyInProgress)
+    );
     Ok(())
 }


src/internet_identity/internet_identity.did
lmuntaner, line 1: This means new declaration files need to be created, right?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -156,81 +177,81 @@ pub enum VerifyTentativeDeviceError {
...
-/// Verifies the tentative device using the submitted `user_verification_code` and returns
-/// a result of [VerifyTentativeDeviceResponse]. [VerifyTentativeDeviceResponse] is used both as
-/// a success and an error type because it corresponds to the candid variant unifying success and
-/// error cases. See `authenticated_anchor_operation` for more details on how the [Result] is handled.
-pub fn verify_tentative_device(
-    anchor: &mut Anchor,
+/// Confirm the tentative device or session using the submitted `user_confirmation_code`.
+///
+/// # Returns
+/// [`DeviceData`] if a device has been confirmed.
```
lmuntaner, line 187: If a session is confirmed, does it return als a DeviceData?

src/internet_identity/src/main.rs
```diff
@@ -72,13 +73,22 @@ async fn init_salt() {
...
+                trap("Unreachable error")
+            }
+        })
 }
 
 #[update]
 fn exit_device_registration_mode(anchor_number: AnchorNumber) {
     check_authz_and_record_activity(anchor_number).unwrap_or_else(|err| trap(&format!("{err}")));
-    tentative_device_registration::exit_device_registration_mode(anchor_number)
+    tentative_device_registration::exit_device_registration_mode(anchor_number);
```
lmuntaner, line 91: Why is this needed here? Do we want to trap from `exit_device_registration_mode` as well?

src/internet_identity/src/main.rs
```diff
@@ -114,17 +124,23 @@ async fn add_tentative_device(
...
-    });
-    match result {
-        Ok(()) => VerifyTentativeDeviceResponse::Verified,
+    let (mut anchor, authorization_key) = check_authorization(anchor_number)
+        .unwrap_or_else(|err| trap(&format!("{} could not be authenticated.", err.principal)));
+    match tentative_device_registration::confirm_tentative_device(
+        anchor_number,
+        user_confirmation_code,
+    ) {
+        Ok(confirmed_device) => {
```
lmuntaner: Could we rename `Ok(maybe_confirmed_device)` to make it clear that there is an option within the `Ok()`.

Or mabye even `Ok(Some(confirmed_device))` and `Ok(None)` and you don't need the nested conditional.

src/internet_identity/src/main.rs
```diff
@@ -924,27 +930,38 @@ mod v2_api {
...
     ) -> Result<(), AuthnMethodConfirmationError> {
-        let response = verify_tentative_device(identity_number, confirmation_code);
-        match response {
-            VerifyTentativeDeviceResponse::Verified => Ok(()),
-            VerifyTentativeDeviceResponse::WrongCode { retries_left } => {
-                Err(AuthnMethodConfirmationError::WrongCode { retries_left })
+        let (mut anchor, authorization_key) = check_authorization(identity_number)
+            .map_err(|err| AuthnMethodConfirmationError::Unauthorized(err.principal))?;
+
+        let confirmed_device = tentative_device_registration::confirm_tentative_device(
```
lmuntaner: Ditto `maybe_confirmed_device`.

src/internet_identity/src/state.rs
```diff
@@ -76,13 +76,22 @@ impl TentativeDeviceRegistration {
 
 /// Registration state of new devices added using the two step device add flow
 #[derive(Clone)]
+#[allow(dead_code)]
 pub enum RegistrationState {
     DeviceRegistrationModeActive,
     DeviceTentativelyAdded {
         tentative_device: DeviceData,
-        verification_code: DeviceVerificationCode,
+        confirmation_code: DeviceConfirmationCode,
```
lmuntaner, line 84: Why renaming this?

lmuntaner (approved):
Thanks!!

For next time, leave refactors for another PR.

I agree with the refactor, but it would have made reviewing way easier on a separate PR.

lmuntaner (commented):
Why is this a PoC?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
sea-snake: No, it won't, I'll update the comment to clarify thx

src/internet_identity/src/main.rs
sea-snake: This resolves a clippy warning, no actual change.

sea-snake:
> Why is this a PoC?

Decided to split the PR into multiple PR given that it's a mix of refactors and actual implementation changes, using this PR to gather initial feedback and debug changes.

src/internet_identity/src/state.rs
sea-snake: To clarify it's a confirmation code (which matches naming in v2 endpoints too).

In v1 endpoints it's unfortunately called a verification code, which considering what it's actually used for, seems incorrect and confusing.

sea-snake:
Already merged in other PRs

PR #3217, opened 2025-07-28
Merge v1 and v2 device registration methods

sea-snake:
Merge v1 and v2 device registration methods, so we only need to maintain a single version.

# Changes

- Merge `enter_device_registration_mode` and `enter_device_registration_mode_v2`.
- Refactor the above to handle an optional registration id and make clear that only an error is returned when the registration ids don't match.
- Rename `InternalError` to `InternalCanisterError` in `AuthnMethodRegistrationModeEnterError` to match error variant in other methods.
- Merge `prune_expired_tentative_device_registrations` and `prune_expired_tentative_device_registrations_v2`.
- Refactor the above for clarity, it's now more explicit that registrations that are not expired will be retained.

# Tests

- Verified device registration still works as expected in both 1.0 and 2.0

### DIFF
--- a/src/frontend/src/lib/generated/internet_identity_idl.js
+++ b/src/frontend/src/lib/generated/internet_identity_idl.js
@@ -162,9 +162,9 @@ export const idlFactory = ({ IDL }) => {
   const RegistrationId = IDL.Text;
   const AuthnMethodRegistrationModeEnterError = IDL.Variant({
     'InvalidRegistrationId' : IDL.Text,
+    'InternalCanisterError' : IDL.Text,
     'AlreadyInProgress' : IDL.Null,
     'Unauthorized' : IDL.Principal,
-    'InternalError' : IDL.Text,
   });
   const AuthnMethodReplaceError = IDL.Variant({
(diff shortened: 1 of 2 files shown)


src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -125,29 +102,31 @@ pub async fn add_tentative_device(
     let now = time();
 
     state::tentative_device_registrations_mut(|registrations| {
-        prune_expired_tentative_device_registrations(registrations);
+        state::lookup_tentative_device_registration_mut(|lookup| {
+            prune_expired_tentative_device_registrations(registrations, lookup);
 
-        match registrations.get_mut(&anchor_number) {
```
sea-snake, line 130: Nothing significant changed in this method, the only change is an indent due to the additional `state::lookup_tentative_device_registration_mut(|lookup| {` wrapper to get the `lookup` that's now additionally passed into `prune_expired_tentative_device_registrations`.

lmuntaner (approved):
Thanks!

PR #3219, opened 2025-07-29
Add sessions to device registration state

sea-snake:
Add the concept of sessions to the device registration state.

# Changes

- Add `SessionTentativelyAdded` and `SessionTentativelyConfirmed` to `RegistrationState`.
- Refactor `add_tentative_device` to handle above additional states, explicitly only allow adding a tentative device when the current state is `DeviceRegistrationModeActive`.
- Merge `verify_tentative_device` and `get_tentative_device` into new `confirm_tentative_device_or_session` method since this concept has changed with additional session states.
- Refactor the above method to support sessions.
- Remove `IdentityUpdateError` variant from `VerifyTentativeDeviceError` since we no longer directly add the device directly in the above method, instead we return the confirmed device that should be added.
- Add `Unauthorized(Principal)` and `InternalCanisterError(text)` to `AuthnMethodConfirmationError` canister error (v2 did).
- Rename verification code to confirmation code for clarity in various places (v1 methods remain as is).

# Tests

- Manually verified that existing device registration flow still works as expected.

### DIFF
--- a/src/internet_identity/src/anchor_management/tentative_device_registration.rs
+++ b/src/internet_identity/src/anchor_management/tentative_device_registration.rs
@@ -1,18 +1,14 @@
-use crate::anchor_management::add_device;
 use crate::authz_utils::IdentityUpdateError;
 use crate::state::RegistrationState::{
     DeviceRegistrationModeActive, DeviceTentativelyAdded, SessionTentativelyAdded,
     SessionTentativelyConfirmed,
 };
 use crate::state::TentativeDeviceRegistration;
-use crate::storage::anchor::Anchor;
 use crate::{secs_to_nanos, state};
 use candid::{CandidType, Principal};
 use ic_cdk::api::time;
 use ic_cdk::{call, trap};
-use internet_identity_interface::archive::types::Operation;
 use internet_identity_interface::internet_identity::types::*;
-use std::collections::hash_map::Entry;
 use std::collections::{hash_map, HashMap};
 use TentativeDeviceRegistrationError::{AnotherDeviceTentativelyAdded, DeviceRegistrationModeOff};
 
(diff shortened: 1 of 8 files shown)


src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -135,82 +142,80 @@ pub enum VerifyTentativeDeviceError {
...
+                .get_mut(&anchor_number)
+                .ok_or(VerifyTentativeDeviceError::DeviceRegistrationModeOff)?;
+
+            let (should_remove, response) = match &mut registration.state {
+                DeviceRegistrationModeActive | SessionTentativelyConfirmed { .. } => {
+                    (false, Err(VerifyTentativeDeviceError::NoDeviceToVerify))
+                }
+                DeviceTentativelyAdded {
+                    failed_attempts,
+                    confirmation_code: verification_code,
```
lmuntaner: Why renaming it to `verification_code` here?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
```diff
@@ -96,35 +96,42 @@ impl From<IdentityUpdateError> for TentativeDeviceRegistrationError {
 
 pub async fn add_tentative_device(
```
lmuntaner, line 97: Will you add session functionality in another PR?

src/internet_identity/src/anchor_management/tentative_device_registration.rs
sea-snake: Seems like my IDE didn't rename the variable in all places when I renamed it in the state 😅

Good catch, will update it.

src/internet_identity/src/anchor_management/tentative_device_registration.rs
sea-snake: Yes, that will be in the next PR.

PR #3220, opened 2025-07-29
Hide system overlay on error.

sea-snake:
Hide system overlay on error.

# Changes

- Set `systemOverlay` to `false` in `finally` block.

# Tests

- Verified that that the system overlay now correctly hides on error e.g. cancellation.


lmuntaner (approved):
Thanks!

PR #3221, opened 2025-07-29
Move feature flag defaults to canister config

sea-snake:
Move feature flag defaults to canister config.

# Changes

- Add `initialize` method to feature flag stores so they can be initialized _after_ the canister config has been read.
- Use canister config as default for `feature_flag_continue_from_another_device` default value.
- Add `feature_flag_continue_from_another_device` to canister init config.
- Add `feature_flag_continue_from_another_device` to persistent config.

# Tests

- Verified that canister can be upgraded to WASM with feature flag.
   - Can be set to true/false or skipped.
- Verified that canister can be downgraded from WASM with feature flag to WASM without.
   - Flag is removed from persistent data which means that removing flags from implementation also cleans them up.
   - Old configuration will not be used again once upgraded to newer WASM with same flag name.


lmuntaner (approved):
Thanks!

We have a test file for each config field.

However, I guess that we don't need one for feature flags, it would be cumbersome to create and remove.

PR #3223, opened 2025-07-30
Fetch identity credentials

lmuntaner:
# Motivation

We want to enable new passkeys added from the same device as soon as they are added.

To do this, we fetch the identity's credentials before passing them for authentication using the `lookup` endpoint.

# Changes

* Fetch credentials with `lookup` in `init` in `AuthLastUsedFlow`.
* Trigger the init when creating `AuthLastUsedFlow` instance.

# Tests

* Tested locally by adding a new passkey from 1Password (initial one was iCloud) and then using immediately afterwards to log in. See video attached.
* I tested the flow from an application.
* I tested switching identity that the new passkey was requested as well.
* I teted that I could switch the identity from an application or dashboard.

### DIFF
--- a/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
+++ b/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
   import { nonNullish } from "@dfinity/utils";
   import { AuthFlow } from "$lib/flows/authFlow.svelte";
-  import { AuthLastUsedFlow } from "$lib/flows/authLastUsedFlow.svelte";
   import type { Snippet } from "svelte";
   import SolveCaptcha from "$lib/components/wizards/auth/views/SolveCaptcha.svelte";
   import PickAuthenticationMethod from "$lib/components/wizards/auth/views/PickAuthenticationMethod.svelte";
@@ -32,7 +31,6 @@
   }: Props = $props();
 
   const authFlow = new AuthFlow();
-  const authLastUsedFlow = new AuthLastUsedFlow();
 
   let isContinueFromAnotherDeviceVisible = $state(false);
 
@@ -109,6 +107,6 @@
   {/if}
 {/if}
 
-{#if authFlow.systemOverlay || authLastUsedFlow.systemOverlay}
+{#if authFlow.systemOverlay}
   <SystemOverlayBackdrop />
 {/if}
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -20,10 +21,30 @@ export class AuthLastUsedFlow {
...
+            .map(convertToValidCredentialData)
+            .filter(nonNullish);
+
+          if (validCredentials.length > 0) {
+            credentialIds = validCredentials.map(
+              (credential) => new Uint8Array(credential.credentialId),
+            );
+          }
+        } catch (error) {
+          console.warn("Error looking up identity credentials:", error);
```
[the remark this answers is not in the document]
lmuntaner: good idea

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -20,10 +21,30 @@ export class AuthLastUsedFlow {
...
+        }
```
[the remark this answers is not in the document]
lmuntaner: Good point, but weird proposed solution

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -11,7 +11,36 @@ import {
...
-import { isNullish } from "@dfinity/utils";
+import { isNullish, nonNullish } from "@dfinity/utils";
+import { convertToValidCredentialData } from "$lib/utils/credential-devices";
+
+const fetchIdentityCredentials = async (
+  identityNumber: bigint,
+): Promise<Uint8Array[] | undefined> => {
+  try {
+    const identityCredentials =
+      await get(sessionStore).actor.lookup(identityNumber);
```
sea-snake: You can use `anonymousActor` instead since this is public data.

lmuntaner: I forgot about that one, thanks!

lmuntaner:
Video:


https://github.com/user-attachments/assets/c19db1dd-447c-4b9f-b5e5-394d8eb74845

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
```diff
@@ -109,6 +107,6 @@
   {/if}
 {/if}
 
-{#if authFlow.systemOverlay || authLastUsedFlow.systemOverlay}
+{#if authFlow.systemOverlay}
```
lmuntaner, line 110: This wasn't necessary because this component doesn't use the `AuthLastUsedFlow` anymore.

sea-snake (approved):

PR #3225, opened 2025-07-30
Refactor `tentative_device_registration` to use v2 types instead.

sea-snake:
Refactor `tentative_device_registration.rs` to use v2 types instead. 

This moves the mapping between types from the v2 methods (v1 -> v2) to the v1 methods (v2 -> v1).

# Tests

- Verified that device registration flow still works as expected.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -110,9 +110,9 @@ async fn add_tentative_device(
                 AddTentativeDeviceResponse::AnotherDeviceTentativelyAdded
             }
             AuthnMethodRegisterError::InvalidMetadata(_) => {
-                // Not reachable since we don't convert from `AuthnMethodData` to `DeviceWithUsage`
+                // Unreachable since we don't convert from `AuthnMethodData` to `DeviceWithUsage`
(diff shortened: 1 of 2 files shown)


src/internet_identity/src/main.rs
```diff
@@ -98,23 +95,24 @@ async fn add_tentative_device(
...
+            AuthnMethodRegisterError::RegistrationAlreadyInProgress => {
                 AddTentativeDeviceResponse::AnotherDeviceTentativelyAdded
             }
-            TentativeDeviceRegistrationError::IdentityUpdateError(err) => {
-                // Legacy API traps instead of returning an error.
-                trap(String::from(err).as_str())
+            AuthnMethodRegisterError::InvalidMetadata(_) => {
+                // Not reachable since we don't convert from `AuthnMethodData` to `DeviceWithUsage`
+                // in this legacy method in comparison to the newer `authn_method_register` method.
+                AddTentativeDeviceResponse::DeviceRegistrationModeOff
```
lmuntaner: Why return this then and not trap?

sea-snake: Ah good idea! I'll update it.

PR #3229, opened 2025-07-31
Exit registration mode with device.

sea-snake:
Exit registration mode with device.

# Changes

- Update `authn_method_registration_mode_exit` to allow a confirmed session to exit by registering a authn method.
- Update above method to return new `AuthnMethodRegistrationModeExitError` error.
- Add `get_confirmed_session` method to get the confirmed session principal of a given anchor.

# Tests

- Verified exiting registration mode without device requires authentication and still works as expected.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -526,6 +526,13 @@ type AuthnMethodRegistrationModeEnterError = variant {
     InternalCanisterError : text;
 };
 
+type AuthnMethodRegistrationModeExitError = variant {
+    Unauthorized : principal;
+    InternalCanisterError : text;
+    RegistrationModeOff;
+    InvalidMetadata : text;
+};
+
 type LookupByRegistrationIdError = variant {
     InvalidRegistrationId : text;
 };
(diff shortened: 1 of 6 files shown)


src/internet_identity/internet_identity.did
```diff
@@ -860,7 +867,7 @@ service : (opt InternetIdentityInit) -> {
 
     // Exits the authentication method registration mode for the identity.
     // Requires authentication.
-    authn_method_registration_mode_exit : (IdentityNumber) -> (variant { Ok; Err });
+    authn_method_registration_mode_exit : (IdentityNumber) -> (variant { Ok; Err : AuthnMethodRegistrationModeExitError });
```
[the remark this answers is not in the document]
lmuntaner: @sea-snake they are right

src/internet_identity/src/main.rs
```diff
@@ -908,8 +909,51 @@ mod v2_api {
...
+                    AuthnMethodRegistrationModeExitError::InvalidMetadata(err.to_string())
+                })?;
+            // Add device to anchor with bookkeeping
+            let mut anchor = state::anchor(identity_number);
+            let operation = anchor_management::add_device(&mut anchor, device_data.clone());
+            state::storage_borrow_mut(|storage| storage.update(anchor)).map_err(|err| {
+                AuthnMethodRegistrationModeExitError::InternalCanisterError(err.to_string())
+            })?;
+            anchor_management::post_operation_bookkeeping(identity_number, operation);
+            // Add temp key for device
```
lmuntaner: Should we add why we do this in the comment?

sea-snake: Good idea, I'll clarify in the comment.

PR #3230, opened 2025-08-01
Return confirmed session in public identity_authn_info method.

sea-snake:
Return confirmed session in public identity_authn_info method.

# Changes

- Add additional optional `authn_method_session` field in `IdentityAuthnInfo` which returns a confirmed session principal.

# Tests

- Verified that None is returned when the flag is false or unset.


sea-snake:
Superseded by https://github.com/dfinity/internet-identity/pull/3230

Adding another method avoids making the session principal unnecessarily public and makes it possible to return additional required info like the identity name. Both could be added to `identity_authn_info` with a caller check, but this would add unnecessary complexity to an existing public method.

PR #3231, opened 2025-08-01
Add method to check if session is confirmed

sea-snake:
Add `authn_method_session_info` method to check if session is confirmed and get info (name).

# Changes

- Add `authn_method_session_info` method that returns info (identity name) if session is confirmed and caller matches session. 
- Rename `authn_method_session` method to `authn_method_session_register` for clarity.

# Tests

- Verified that new `authn_method_session_info` traps if flag is false or unset.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -974,7 +974,7 @@ mod v2_api {
     }
 
     #[update]
-    async fn authn_method_session(
+    async fn authn_method_session_register(
         identity_number: IdentityNumber,
     ) -> Result<AuthnMethodConfirmationCode, AuthnMethodRegisterError> {
         // Adding a session is behind a feature flag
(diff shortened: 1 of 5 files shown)


src/internet_identity/src/main.rs
```diff
@@ -989,6 +989,27 @@ mod v2_api {
...
+        if !persistent_state(|state| {
+            state
+                .feature_flag_continue_from_another_device
+                .unwrap_or(false)
+        }) {
+            trap("feature_flag_continue_from_another_device is disabled");
+        }
+
+        // Return session info if caller matches confirmed session
+        tentative_device_registration::get_confirmed_session(identity_number)
```
lmuntaner, line 1006: Is the session's principal the same as the identity?

This method will be used by the existing device to confirm that the flow finished, is that right?

sea-snake: No, this will be used by the new device to check if the tentative session has been confirmed.

Additionally it also returns info (in this case identity name) needed for creating the passkey that's used in the args to close the registration mode.

lmuntaner: Why do we need to confirm that it has been added?

lmuntaner (approved):

src/internet_identity/src/main.rs
sea-snake: Clarified and discussed in Slack.

PR #3232, opened 2025-08-01
Refactor anchor/identity_info to use v2 types and add session.

sea-snake:
Refactor anchor/identity_info to use v2 types and add session.

# Changes

- Refactor anchor/identity_info to use v2 types, mapping from v2 to v1 in legacy methods instead of other way around.
- Add `session` to `AuthnMethodRegistration`.

# Tests


sea-snake:
Abandoning this refactor, too many tests are failing due to metadata `Some({}) != None`.

PR #3233, opened 2025-08-04
Add tentative session in identity info.

sea-snake:
Add tentative session in identity info.

# Changes

- Add `get_tentative_registration` method in `tentative_device_registration.rs`
- Remove `to_info_if_still_valid` method in `state.rs`, registrations should be accessed through `tentative_device_registration.rs` instead of directly.
- Add `session` to `AuthnMethodRegistration`.
- Add `tentative_session` to `DeviceRegistrationInfo`.

# Tests


lmuntaner (approved):
Thanks!

PR #3236, opened 2025-08-05
Register passkey from another device through session to automatically sign in.

sea-snake:
Register passkey from another device through session to automatically sign in.

# Changes

- Updated register and confirm access method flows to use session instead of temp key (behind canister config flag).
- Added additional success toasts for visual feedback after completing the flow on either device.

# Tests

- Verified the session flow is behind a flag, temp key flow is fallback when flag is unset or false.

### DIFF
--- a/src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.ts
@@ -1,4 +1,4 @@
-import { anonymousActor } from "$lib/globals";
+import { anonymousActor, canisterConfig } from "$lib/globals";
 import { get } from "svelte/store";
 import { isNullish, nonNullish } from "@dfinity/utils";
 import { authenticatedStore } from "$lib/stores/authentication.store";
@@ -14,6 +14,7 @@ export class ConfirmAccessMethodFlow {
   >("continueOnNewDevice");
   #waitingForDevice?: AuthnMethodData;
   #newDeviceLink = $state<URL>();
+  #sessionExpiration = $state<bigint>();
 
   get view() {
     return this.#view;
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.ts
```diff
@@ -50,12 +51,24 @@ export class ConfirmAccessMethodFlow {
...
           break;
         }
-        // Show confirmation code view if we got a pending authn method
-        if (nonNullish(info.authn_method_registration[0]?.authn_method[0])) {
-          this.#waitingForDevice =
-            info.authn_method_registration[0]?.authn_method[0];
-          this.#view = "enterConfirmationCode";
-          break;
+        if (
+          canisterConfig.feature_flag_continue_from_another_device[0] === true
```
lmuntaner: Can we convert this into an instance variable? `continueFromAnotherDeviceIsEnabled` for example

src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
```diff
@@ -58,6 +59,12 @@ export class RegisterAccessMethodFlow {
         await anonymousActor.lookup_by_registration_mode_id(registrationId)
       )[0];
       if (nonNullish(identityNumber)) {
+        this.#identityNumber = identityNumber;
+        if (
+          canisterConfig.feature_flag_continue_from_another_device[0] === true
```
lmuntaner: Ditto variable

src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
```diff
@@ -141,22 +183,35 @@ export class RegisterAccessMethodFlow {
...
-        authnMethodData,
-      )
-      .then(throwCanisterError);
+    if (canisterConfig.feature_flag_continue_from_another_device[0] === true) {
+      await session.actor
+        .authn_method_registration_mode_exit(this.#identityNumber, [
+          authnMethodData,
+        ])
+        .then(throwCanisterError);
+      const identity = await authenticateWithSession({ session });
```
lmuntaner, line 186: Do you need to authenticate again? Why isn't it enough the one in `#registerSession`?

src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.ts
sea-snake: I skipped the variable since this condition and the temp key code would be removed once enabled and verified to be without issues on prod.

But sure, I'll make is a flow constructor param for clarity.

src/frontend/src/lib/flows/registerAccessMethodFlow.svelte.ts
sea-snake: In register session, we're only authenticating when the flag is _not_ true (flow with temp key instead of session). 

In the session flow, we can't authenticate yet until the registration mode has been closed with a passkey.

lmuntaner: > In register session, we're only authenticating when the flag is not true (flow with temp key instead of session).

This is when we call `#registerTempKey`. We don't call `#registerSession` if we don't do the session. So `!this.#sessionAvailable` will always be false, won't it?

```
if (this.#sessionAvailable) {
  return this.#registerSession(identityNumber);
}
```

PR #3237, opened 2025-08-05
Upgrade pocket-ic

lmuntaner:
# Motivation

We have random failing tests that we believe might be due to the old version of pocket-ic we are using.

In this PR, I upgrade pocket-ic. Both Rust and binary.

# Changes

* Upgrade cargo.toml
* Fix breaking changes (main an import of the error type of calls) 

# Tests

Only changes in tests.


sea-snake (approved):
LGTM!

PR #3240, opened 2025-08-06
Trigger of the migration flow

lmuntaner:
# Motivation

Offer the CTA to the user to migrate their legacy identities.

# Changes

* Added new footer to PickAuthenticationMethod as agreed with design.

# Tests

* Tested manually that it appears in all the flows where the modal is used and it doesn't look off.

### DIFF
--- a/src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
+++ b/src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
@@ -6,7 +6,10 @@
   import Alert from "$lib/components/ui/Alert.svelte";
   import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
   import { canisterConfig } from "$lib/globals";
-  import { CONTINUE_FROM_ANOTHER_DEVICE } from "$lib/state/featureFlags";
+  import {
+    CONTINUE_FROM_ANOTHER_DEVICE,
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
```diff
@@ -79,4 +82,10 @@
       </Button>
     {/if}
   </div>
+  {#if $ENABLE_MIGRATE_FLOW}
+    <div class="text-text-primary flex flex-row items-center justify-between">
```
sea-snake: Make sure to specify the text size e.g. `text-md`.

lmuntaner: Done!

lmuntaner: Why would that be necessary? Is it set differently from the parent sometimes?

sea-snake: Inheriting font size from the parent outside this svelte component means the it could unintentionally change due to a code change outside the component and no longer match design.

lmuntaner: Good point, we should never rely on the parent setting the font-size

PR #3241, opened 2025-08-07
Add Migration flow into AuthFlow

lmuntaner:
# Motivation

Enable users to migrate their identities either from II landing page or when coming from an app.

In this PR, the migration flow is embedded within the AuthFlow.

# Changes

* Added a new `"migrate"` view to the `AuthFlow` class and implemented a `migrate` method to switch to this view.
* Updated the authentication wizard (`AuthWizard.svelte`) to render the `MigrationWizard` component when the `"migrate"` view is active and handle migration success by signing in the user with their identity number.
* Modified the authentication method picker (`PickAuthenticationMethod.svelte`) to show an "Upgrade" button that triggers the migration flow using the new `migrate` prop and handler, replacing the previous link.
* Passed the `migrate` handler from the authentication flow to the authentication method picker and connected it through the wizard component tree.
* Updated the `MigrationWizard` component to accept an `onSuccess` callback that receives the `identityNumber` as a parameter and triggers it only when a valid identity number is present.

# Tests

Tested manually the following flows:
- Desktop: User from dapp and no cached identities.
- Desktop: User from dapp and cached identities.
- Desktop: User from landing page and cached identities.
- Desktop: User from landing page and no cached identities (see video below)
- Mobile simulator: User from dapp and cached identities
- Mobile simulator: User from dapp and no cached identities

### DIFF
--- a/src/frontend/src/lib/flows/authFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authFlow.svelte.ts
@@ -32,7 +32,7 @@ import { createGoogleRequestConfig, requestJWT } from "$lib/utils/openID";
 
 export class AuthFlow {
   #view = $state<
-    "chooseMethod" | "setupOrUseExistingPasskey" | "setupNewPasskey" | "migrate"
+    "chooseMethod" | "setupOrUseExistingPasskey" | "setupNewPasskey"
   >("chooseMethod");
   #captcha = $state<{
     image: string;
@@ -65,10 +65,6 @@ export class AuthFlow {
   chooseMethod = (): void => {
     authenticationV2Funnel.trigger(AuthenticationV2Events.SelectMethodScreen);
     this.#view = "chooseMethod";
-  };
-
-  migrate = (): void => {
-    this.#view = "migrate";
   };
 
   setupOrUseExistingPasskey = (): void => {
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -32,7 +32,7 @@ import { createGoogleRequestConfig, requestJWT } from "$lib/utils/openID";
 
 export class AuthFlow {
   #view = $state<
-    "chooseMethod" | "setupOrUseExistingPasskey" | "setupNewPasskey"
+    "chooseMethod" | "setupOrUseExistingPasskey" | "setupNewPasskey" | "migrate"
```
sea-snake: I would recommend against adding this state to `authFlow`, instead recommending a $state variable in `AuthWizard` similar to `isContinueFromAnotherDeviceVisible`.

Keeping a flow focused on the fewer relevant states keeps it more maintainable over time.

lmuntaner: I agree of not having it here, but I'm not convinced with having it as a state. This component is starting to feel too big.

Done for now.

sea-snake (approved):

PR #3244, opened 2025-08-07
Enable continue from another device flow with temp key.

sea-snake:
Enable continue from another device flow with temp key.

# Changes

- Set default value for `CONTINUE_FROM_ANOTHER_DEVICE ` frontend flag to true.
- Move "Continue from another device" button to passkey view.
- Fix e2e now that the button has moved to the passkey view.
- Remove feature flags from e2e since they are no longer needed (flags can be enabled in canister init instead).

# Tests

- Verified the continue from another device flow works as expected now that the button has moved.


lmuntaner (approved):
Thanks!

PR #3246, opened 2025-08-07
Implement integration tests for authn method registration through session.

sea-snake:
Implement integration tests for authn method registration through session.

# Changes

- Added optional `authn_method` param to `api_v2::authn_method_registration_mode_exit`.
- Added `authn_method_session_register` to `api::v2`.
- Added tests in `authn_method_session_registration.rs` to cover new session cases not covered in `authn_method_registration.rs`.

### DIFF
--- a/src/internet_identity/tests/integration/v2_api/authn_method_session_registration.rs
+++ b/src/internet_identity/tests/integration/v2_api/authn_method_session_registration.rs
@@ -1,0 +1,339 @@
+use crate::v2_api::authn_method_test_helpers::{
+    create_identity_with_authn_method, sample_webauthn_authn_method,
+};
+use canister_tests::api::internet_identity::api_v2;
+use canister_tests::framework::{env, install_ii_with_archive};
+use internet_identity_interface::internet_identity::types::{
+    AuthnMethodConfirmationError, AuthnMethodRegisterError, AuthnMethodRegistration,
+};
+use pocket_ic::RejectResponse;
+use std::time::Duration;
+
+#[test]
+fn should_register_authn_method_through_session() -> Result<(), RejectResponse> {
+    let env = env();
+    let canister_id = install_ii_with_archive(&env, None, None);
+    let existing_authn_method = sample_webauthn_authn_method(0);
+    let new_authn_method = sample_webauthn_authn_method(1);
+    let session = sample_webauthn_authn_method(2);
+    let identity_number =
+        create_identity_with_authn_method(&env, canister_id, &existing_authn_method);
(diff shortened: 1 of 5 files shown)


src/internet_identity/tests/integration/v2_api/authn_method_session_registration.rs
lmuntaner, line 1: How about adding the following tests:

* Two different users can add a device at the same time.
* The code from one user doesn't work with another user.
* Exiting mode in one user doesn't exist in the other.

Thanks!

sea-snake: These should already be covered in the other e2e tests file. But I'll add them given that sessions are involved and could have edge cases. Thanks!

lmuntaner: Ah, yeah, I guess just copy paste them. It should be easy to have them...

sea-snake: > Two different users can add a device at the same time.

Added `should_register_authn_methods_in_parallel` in `authn_method_session_registration.rs`

> The code from one user doesn't work with another user.

Added `should_reject_code_from_another_registration` in `authn_method_session_registration.rs`

> Exiting mode in one user doesn't exist in the other.

Added `should_exit_registrations_separately` in `authn_method_session_registration.rs`

src/internet_identity/tests/integration/v2_api/authn_method_session_registration.rs
```diff
@@ -0,0 +1,617 @@
...
+    )?
+    .expect("authn_method_registration_mode_exit failed for identity 1");
+    api_v2::authn_method_confirm(
+        &env,
+        canister_id,
+        authn_method2.principal(),
+        identity_number2,
+        &add_response1.confirmation_code,
+    )?
+    .expect("authn_method_confirm failed for identity 2");
```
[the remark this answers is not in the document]
lmuntaner: This seems right, it should fail as is now, right?

lmuntaner (commented):
One comment on the bot's comment as well

src/internet_identity/tests/integration/v2_api/authn_method_session_registration.rs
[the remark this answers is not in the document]
sea-snake: Yes, good catch by the bot.

lmuntaner (approved):

PR #3252, opened 2025-08-08
Limit test threads in canister tests

lmuntaner:
# Motivation

We see randome failures in our CI.

The problem might be that pocket-ic gets overloaded.

Limiting the number of test threads should fix the problem.

# Changes

* Pass `--test-threads=2` when running canister tests.

# Tests

See CI.


lmuntaner:
@sea-snake I don't know for sure whether this will fix it. Martin suggested to put it to two and I still got an error with two. However, with 1 I have run it three times and it was successful. But maybe it was just luck...

lmuntaner:
@sea-snake it failed in the merge-queue 😢 . But I'm merging it anyway, 1 out of 4 is better than what we have now I believe.

PR #3254, opened 2025-08-11
Success screen after successful migration on authorize flow

lmuntaner:
# Motivation

We want to show a success screen after a successful migration to the users that perform it during an authorize flow.

In this PR, I introduce the new flow without UI nor final UX (pending automatic redirect after 5 seconds)

# Changes

* Create a new route where the user will be redirected to the dapp on click of a button.
* Added an onMigration prop to AuthWizard and used it to redirect to the new route or manage page.
* Use the new onMigration prop accordingly.

# Tests

Tested locally with both use cases from a dapp (see video) and from the landing page.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/authorize/(authenticated)/upgrade-success/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/authorize/(authenticated)/upgrade-success/+page.svelte
@@ -1,0 +1,10 @@
+<script lang="ts">
+  import Button from "$lib/components/ui/Button.svelte";
+  import { authorizationStore } from "$lib/stores/authorization.store";
+
+  const handleRedirect = () => {
+    authorizationStore.authorize(undefined, 0);
+  };
+</script>
+
+<Button onclick={handleRedirect}>Go to app</Button>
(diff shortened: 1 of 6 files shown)


lmuntaner:
New UX:

https://github.com/user-attachments/assets/14f8080d-05fa-4322-83a0-e3de901f2482

src/frontend/src/routes/(new-styling)/authorize/(authenticated)/upgrade-success/+page.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+  import Button from "$lib/components/ui/Button.svelte";
+  import { authorizationStore } from "$lib/stores/authorization.store";
+
+  const handleRedirect = () => {
+    authorizationStore.authorize(undefined, 0);
```
sea-snake: You can omit the delay if it's 0.

```suggestion
    authorizationStore.authorize(undefined);
```

lmuntaner: done!

sea-snake (approved):

PR #3257, opened 2025-08-12
Share CreatePasskey component

lmuntaner:
# Motivation

Keep consistency and maintainability.

In this PR, we share the CreatePasskey screen from the migration and the registration flows. The UI is the one from the migration.

# Changes

* Move the UI of CreatePasskeyMigration.svelte to the CreatePasskey.svelte component.
* Use CreatePasskey.svelte in MigrationWizard.svelte
* Remove CreatePasskeyMigration.svelte
* I changed the height passed to the smiley illustration. "h-32" was on top of the margin set in "my-5"; therefore, I set it to "h-22" so that the full height is 128px as expected.

# Tests

Tested locally the registration and migration flows.


src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
```diff
@@ -73,5 +79,13 @@
         <span>Create Passkey</span>
       {/if}
     </Button>
+    {#if identityNumber !== undefined}
+      <p
+        class="text-text-primary flex items-center justify-center gap-2 text-xs"
+      >
+        <span>You are upgrading ID</span>
```
sea-snake: Nit: `text-text-primary` and `text-xs` should be on the span itself if it only needs to apply to the span in the first place. This also clarifies that the paragraph is a layout element (with layout css) and the span is a text element (with text css).

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
```diff
@@ -73,5 +79,13 @@
         <span>Create Passkey</span>
       {/if}
     </Button>
+    {#if identityNumber !== undefined}
+      <p
```
sea-snake: The `Badge` component (div) is not a valid element inside a paragraph. We should either make the `Badge` component a span or use a div instead instead of a paragraph.

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
```diff
@@ -32,16 +33,23 @@
...
-    </FeaturedIcon>
-    <h1 class="text-text-primary mb-3 text-2xl font-medium">
-      Name your identity
-    </h1>
-    <p class="text-md text-text-tertiary mb-6 font-medium">
-      This will label your passkey, and you can’t rename it later once set.
-    </p>
+  <div
+    class="text-text-primary mb-8 flex w-full flex-col items-center justify-center"
+    in:fly={{ duration: 200, x: 10 }}
```
sea-snake: We're only applying the horizontal slide transition on the illustration, title and description (not input)? Previously it was also applied to the input (not button).

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
lmuntaner: Making the Badge a "span" makes more sense to me.

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
lmuntaner: ah, good catch. Why wasn't it applied to the button?

It makes sense to move it to the parent `form` element.

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
lmuntaner: Why should it be in the span itself?

I can imaging that if there is more text after the Badge for example, I'll want it to also be small.

To me, it's the "p" tag setting the styles of this paragaraph.

I set the span to add the gap nicely between the badg and the sentence.

sea-snake: I mean, a paragraph is a block text element, not a layout element (like a div). Right now you're applying flex styling and alignment to it as if it was a layout element, removing the default behavior and purpose of the paragraph element (rendering one or more wrapping lines of text).

If you'd want spacing around the badge, I'd suggest to use margins instead, ideally the badge component itself should probably be a span instead div with inline-block style so it lines up within lines of text naturally.

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
sea-snake: I wasn't applied to the button, since it looked weird back then (big bright button sliding was a bit much). Feel free to include or exclude the button depending on what "feels" ok.

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
lmuntaner: Done!

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
lmuntaner: It feels right everything. The button is disabled, so it's not so impacting.

PR #3266, opened 2025-08-13
Add domain to passkeys in new dashboard

lmuntaner:
# Motivation

Help the user understand the status after migration.

In this PR, we introduce the origin of the passkey in the new dashboard. But only if the passkeys are from different origins.

# Changes

* Added a new utility function `getOrigin` to extract the origin from an authentication method's metadata, and a function `haveMultipleOrigins` to detect if multiple unique origins exist among the available authentication methods.
* Updated the `AccessMethod.svelte` component to optionally display the origin below the authentication method alias when the `showOrigin` prop is `true` and an origin exists.
* Modified the `AccessMethodsPanel.svelte` component to compute whether to show origins using the new `haveMultipleOrigins` utility and pass this information as a prop to each `AccessMethod` component.

# Tests

* Added unit tests for `getOrigin` and `haveMultipleOrigins` to verify correct behavior with various combinations of authentication methods and metadata.
* Tested in beta with an upgraded identity. See screenshot attached.
* I also checked that a user with only passkeys in one domain doesn't see them.

### DIFF
--- a/src/frontend/src/lib/components/ui/AccessMethod.svelte
+++ b/src/frontend/src/lib/components/ui/AccessMethod.svelte
@@ -73,28 +73,19 @@
       out:fade={{ duration: 30 }}
     >
       <div class="flex min-w-32 pr-3">
-        {#if showOrigin && getOrigin(accessMethod)}
-          <div class="flex flex-col justify-center">
-            <div class="flex items-center gap-2">
-              <span>{getAuthnMethodAlias(accessMethod)}</span>
-              {#if isCurrent}
-                <PulsatingCircleIcon />
-              {/if}
-            </div>
-            <div class="text-text-tertiary text-xs font-extralight">
-              {getOrigin(accessMethod)}
-            </div>
-          </div>
-        {:else}
-          <div class="flex items-center">
-            {getAuthnMethodAlias(accessMethod)}
+        <div class="flex flex-col justify-center">
+          <div class="flex items-center gap-2">
+            <span>{getAuthnMethodAlias(accessMethod)}</span>
             {#if isCurrent}
-              <span class="ml-2" aria-label="Current Passkey">
-                <PulsatingCircleIcon />
-              </span>
+              <PulsatingCircleIcon />
             {/if}
           </div>
-        {/if}
+          {#if showOrigin && getOrigin(accessMethod)}
+            <div class="text-text-tertiary font-extralight">
+              <Ellipsis text={getOrigin(accessMethod)!}></Ellipsis>
+            </div>
+          {/if}
+        </div>
       </div>
       {#if isCurrent}
         <div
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/ui/AccessMethod.svelte
```diff
@@ -10,18 +10,20 @@
   import Ellipsis from "../utils/Ellipsis.svelte";
   import PulsatingCircleIcon from "../icons/PulsatingCircleIcon.svelte";
   import { getAuthnMethodAlias } from "$lib/utils/webAuthn";
-  import { isLegacyAuthnMethod } from "$lib/utils/accessMethods";
```
lmuntaner, line 13: `isLegacyAuthnMethod` was not used...

src/frontend/src/lib/components/ui/AccessMethod.svelte
```diff
@@ -70,12 +72,28 @@
...
+          <div class="flex items-center">
+            {getAuthnMethodAlias(accessMethod)}
+            {#if isCurrent}
+              <span class="ml-2" aria-label="Current Passkey">
+                <PulsatingCircleIcon />
+              </span>
+            {/if}
+          </div>
         {/if}
       </div>
```
sea-snake: This could be simplified:

```suggestion
      <div class="flex min-w-32 pr-3">
        <div class="flex flex-col justify-center">
          <div class="flex items-center gap-2">
            <span>{getAuthnMethodAlias(accessMethod)}</span>
            {#if isCurrent}
              <PulsatingCircleIcon />
            {/if}
          </div>
          {#if showOrigin && getOrigin(accessMethod)}
            <div class="text-text-tertiary text-xs font-extralight">
              {getOrigin(accessMethod)}
            </div>
          {/if}
        </div>
      </div>
```

src/frontend/src/lib/utils/accessMethods.ts
```diff
@@ -53,14 +53,40 @@ export const getLastUsedAccessMethod = (
...
+ * Check if there are multiple unique origins across authentication methods
+ */
+export const haveMultipleOrigins = (
+  authnMethods: AuthnMethodData[],
+): boolean => {
+  const origins = new Set<string | undefined>();
+  for (const method of authnMethods) {
+    origins.add(getOrigin(method));
+  }
+  return origins.size > 1;
```
sea-snake: Nit, this could be simplified:

```suggestion
  return new Set(authnMethods.map(getOrigin)).size > 1;
```

src/frontend/src/lib/components/ui/AccessMethod.svelte
lmuntaner: Thanks!

src/frontend/src/lib/utils/accessMethods.ts
lmuntaner: of course!

src/frontend/src/lib/components/views/AccessMethodsPanel.svelte
```diff
@@ -164,6 +167,7 @@
           accessMethod={authnMethod}
           isDisabled={isLegacyAuthnMethod(authnMethod)}
           isCurrent={isCurrentAccessMethod(authnMethod)}
+          showOrigin={showPasskeyOrigin || true}
```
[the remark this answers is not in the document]
lmuntaner: good catch!

PR #3276, opened 2025-08-14
Lm remove https passkey origin

lmuntaner:
# Motivation

Make the origin shown to the user for passkeys more friendly by removing the protocol.

# Changes

- Updated `AccessMethod.svelte` to use the new `getRpId` function instead of `getOrigin`, so that only the hostname (RP ID) is displayed in the UI instead of the full origin URL. 
- Added a new function `getRpId` in `accessMethods.ts` that extracts the hostname from the origin metadata of an authentication method.

# Tests

* Added unit tests for `getRpId` in `accessMethods.test.ts`, covering cases with valid, missing, and invalid origin metadata.
* Checked locally, see screenshots.

### DIFF
--- a/src/frontend/src/lib/utils/accessMethods.ts
+++ b/src/frontend/src/lib/utils/accessMethods.ts
@@ -68,6 +68,17 @@ export const getOrigin = (
 };
 
 /**
+ * Extract the RP ID (origin without protocol) from an AuthnMethodData's metadata
+ */
+export const getRpId = (accessMethod: AuthnMethodData): string | undefined => {
+  const origin = getOrigin(accessMethod);
+  if (nonNullish(origin)) {
+    return new URL(origin).hostname;
+  }
+  return undefined;
+};
+
+/**
  * Check if there are multiple unique origins across authentication methods
  */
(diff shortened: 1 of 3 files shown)


lmuntaner:
New UI:

Screenshot 2025-08-14 at 16 14 02

src/frontend/src/lib/utils/accessMethods.ts
```diff
@@ -67,6 +67,21 @@ export const getOrigin = (
...
 
+/**
+ * Extract the RP ID (origin without protocol) from an AuthnMethodData's metadata
+ */
+export const getRpId = (accessMethod: AuthnMethodData): string | undefined => {
+  const origin = getOrigin(accessMethod);
+  if (nonNullish(origin)) {
+    try {
+      return new URL(origin).hostname;
+    } catch (_) {
```
sea-snake: Nit: you can leave out the error variable as a whole if you don't need it.

```suggestion
    } catch {
```

sea-snake (approved):
LGTM, added a small nit.

PR #3277, opened 2025-08-14
Remove "Migrated" reference from the user

lmuntaner:
# Motivation

We talk about "Upgrade" not "Migration". Therefore, there should be no reference to "migrate" words

# Changes

* Change "Identity already migrated" to "Identity already upgraded" and below as well

# Tests

Check locally that I get the right message.


sea-snake (approved):
Good catch!

PR #3279, opened 2025-08-15
Increase tooltip duration

sea-snake:
Increase tooltip duration

# Changes

- Increase duration to 4 seconds since this is helpful information, it blocking UI for 4 seconds is less of an issue compared to it disappearing too quickly.

# Tests

- Verified the tooltips show for a longer duration on the screen.

https://github.com/user-attachments/assets/e9c863ba-4e03-479e-8e62-e91e9348a6b7


sea-snake:
> @sea-snake can you add a video?

Sure

lmuntaner:
thanks for the video. Are you sure we don't want it a bit longer?

PR #3280, opened 2025-08-18
A/B Testing with Info Passkey Screen

lmuntaner:
# Motivation

There is a 22% drop-off in the Enter Name screen to create a successful passkey and 35% in the passkey registration in general.

Hypothesis: Users don't expect to use biometrics and are scared when they do.

Test: Show a new step to 20% of the users to explain to them that biometrics are coming.

# Changes

This pull request introduces an A/B test for the passkey creation flow, adding an informational step for a subset of users before creating a passkey. The changes include updates to the `AuthWizard` and related components to support the new flow, as well as modifications to the `AuthFlow` logic to handle the A/B test group assignment and transitions between views.

**A/B Test and Flow Logic Updates:**

* Added an `abTestGroup` property to `AuthFlow` that randomly assigns users to either the "infoPasskey" or "default" group, with 30% of users seeing the new informational step. The flow logic and event tracking are updated to include this group in all relevant analytics events.
* Refactored the passkey creation process: for users in the "infoPasskey" group, submitting a name transitions to a new "infoPasskey" view before passkey creation; for others, passkey creation proceeds immediately.
* Extended the `AuthFlow` view state to include the new "infoPasskey" step, ensuring seamless transitions between all flow stages.

**UI and Component Changes:**

* Added a new `InfoPasskey.svelte` component, which displays information about passkeys and a button to proceed with creation. This view is shown only for users in the "infoPasskey" group.
* Updated the `CreatePasskey.svelte` component to accept customizable button and loading labels, allowing the button to display "Continue" or "Create Passkey" depending on the flow.
* Modified `AuthWizard.svelte` to handle the new flow, including importing the `InfoPasskey` component, updating event handlers, and rendering the correct view based on the current step and A/B test group.

# Tests

Tested locally by setting the a/b testing to 1. See video attached.

### DIFF
--- a/src/frontend/src/lib/flows/authFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authFlow.svelte.ts
@@ -29,7 +29,6 @@ import {
   OpenIdDelegationError,
 } from "$lib/generated/internet_identity_types";
 import { createGoogleRequestConfig, requestJWT } from "$lib/utils/openID";
-import { isE2E } from "$lib/constants/environment.constants";
 
 export class AuthFlow {
   #view = $state<
@@ -66,8 +65,7 @@ export class AuthFlow {
 
   constructor() {
     this.chooseMethod();
-    // In E2E runs, disable the A/B test randomness to keep the UI deterministic for tests.
-    const GROUP_INFO_PERCENTAGE = isE2E ? 0 : 0.3;
+    const GROUP_INFO_PERCENTAGE = 0.2;
     this.abTestGroup =
       Math.random() < GROUP_INFO_PERCENTAGE ? "infoPasskey" : "default";
   }
--- a/src/frontend/tests/e2e-playwright/global.setup.ts
+++ b/src/frontend/tests/e2e-playwright/global.setup.ts
@@ -1,0 +1,13 @@
+import { test as setup } from "@playwright/test";
+
+setup("global setup", async ({ page }) => {
+  await page.addInitScript(() => {
+    // We override the function in the browser's context
+    // This script will run every time a page is created or navigated to
+    // We use Math.random() to determine the A/B test group
+    // In E2E runs, we force the A/B test group to be the same for all tests
+    Object.defineProperty(Math, "random", {
+      value: () => 1,
+    });
+  });
+});
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -36,6 +36,10 @@ export default defineConfig({
   /* Configure projects for major browsers */
   projects: [
     {
+      name: "setup",
+      testMatch: /global\.setup\.ts/,
+    },
+    {
       name: "desktop",
       use: {
         ...devices["Desktop Chrome"],
@@ -46,6 +50,7 @@ export default defineConfig({
           ],
         },
       },
+      dependencies: ["setup"],
     },
     {
       name: "mobile",
@@ -58,6 +63,7 @@ export default defineConfig({
           ],
         },
       },
+      dependencies: ["setup"],
     },
   ],
 });
--- a/src/frontend/src/lib/constants/environment.constants.ts
+++ b/src/frontend/src/lib/constants/environment.constants.ts
@@ -1,1 +1,0 @@
-export const isE2E = import.meta?.env?.MODE === "e2e";


src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -93,18 +106,36 @@ export class AuthFlow {
...
+    }
+  };
+
+  createPasskey = async (): Promise<bigint> => {
     authenticationV2Funnel.trigger(
       AuthenticationV2Events.StartWebauthnCreation,
+      { abTestGroup: this.abTestGroup },
     );
+    if (isNullish(this.#name)) {
+      throw new Error("Name is not set");
```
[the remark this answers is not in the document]
lmuntaner: This error should not happen, no need to improve the nessage.

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -60,22 +65,30 @@ export class AuthFlow {
 
   constructor() {
     this.chooseMethod();
+    const GROUP_INFO_PERCENTAGE = 0.2;
+    this.abTestGroup =
+      Math.random() < GROUP_INFO_PERCENTAGE ? "infoPasskey" : "default";
```
sea-snake, line 70: Are you sure that you'd want the same user to possibly end up in the other test group when they're revisiting? They're unauthenticated, so the best we can do is persist it in storage I suppose.

Maybe we can use a localstorage variable with a boolean? We could use the existing feature flag implementation and set a random true/false in `hooks.client.ts` (if unset).

src/frontend/tests/e2e-playwright/global.setup.ts
```diff
@@ -0,0 +1,13 @@
+import { test as setup } from "@playwright/test";
+
+setup("global setup", async ({ page }) => {
+  await page.addInitScript(() => {
+    // We override the function in the browser's context
+    // This script will run every time a page is created or navigated to
+    // We use Math.random() to determine the A/B test group
+    // In E2E runs, we force the A/B test group to be the same for all tests
+    Object.defineProperty(Math, "random", {
+      value: () => 1,
```
sea-snake, line 10: I'd rather not override the random method for all e2e tests, instead I would update the `abTestGroup` condition to additionally check if we're inside an e2e environment that needs a consistent auth flow UX. 

We could probably check if dummyAuth is true, and then always pick `default`, this also avoids the need for other devs to handle temporary auth flow UI changes in their e2e tests with II (assuming they're using dummy auth).

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -60,22 +65,30 @@ export class AuthFlow {
 
   constructor() {
     this.chooseMethod();
+    const GROUP_INFO_PERCENTAGE = 0.2;
```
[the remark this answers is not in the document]
lmuntaner: I changed the comment

src/frontend/src/lib/flows/authFlow.svelte.ts
lmuntaner: Mhm, I wanted to keep the code as simple as possible and I didn't think of it.

Why should we use the existing feature flag implementation instead of reading and reading directly inside the constructor. We will throw this code anyway.

src/frontend/tests/e2e-playwright/global.setup.ts
lmuntaner: I wanted to do that initially, but then I couldn't find any env var to use it and reading the docs came up with this idea which I kind of liked. It's similar to a before each in unit tests where we don't want randonmess. It made sense to me.

src/frontend/src/lib/flows/authFlow.svelte.ts
sea-snake: Fine with either, whichever is the most trivial to implement.

src/frontend/tests/e2e-playwright/global.setup.ts
sea-snake: The main issue I have with the A/B test in it's current form, is that it would make E2E tests from external devs that involve II flaky for developers who are unaware of the A/B test.

Maybe we can make the whole A/B test something behind a (init arg) feature flag (as in don't do the A/B test if feature flag is disabled or unset)? Then we can deploy with the feature flag enabled to do the A/B test.

For now I think it's probably fine as is, since we intend to remove it, but for the longer term and future A/B tests, I'd like to see something other than relying on monkey patching Math.random().

lmuntaner: That's a good idea. Let's add a new config for the next A/B testing.

src/frontend/src/lib/flows/authFlow.svelte.ts
lmuntaner: Let's keep it as is for now. This is our first A/B testing. We'll learn from it.

sea-snake (approved):

PR #3283, opened 2025-08-19
Make OpenID implementation generic

sea-snake:
Make OpenID implementation generic

# Changes

# Tests


src/internet_identity/internet_identity.did
```diff
@@ -361,6 +361,9 @@ type JWT = text;
 type Salt = blob;
 
 type OpenIdConfig = record {
+    name : text;
+    logo_uri : text;
```
lmuntaner: What is this logo_uri? Will we also need some of this for the icons shown in the buttons?

src/internet_identity/internet_identity.did
```diff
@@ -274,8 +274,8 @@ type InternetIdentityInit = record {
     // Configuration for New Origin Flows.
     // If present, list of origins using the new authentication flow.
     new_flow_origins : opt vec text;
-    // Configuration for OpenID Google client
-    openid_google : opt opt OpenIdConfig;
+    // Configurations for OpenID clients
+    openid_configs : opt vec OpenIdConfig;
```
lmuntaner, line 278: We will need a new field, right? This is stored in the stable memory. Or by using a new field we will already stop reading the old one?

src/internet_identity/src/openid/generic.rs
```diff
@@ -75,27 +83,72 @@ struct Claims {
...
+
+#[allow(dead_code)]
+pub struct SupportedProvider {
+    issuer: String,
+    keys: Vec<Jwk>,
+    updated_at: u64,
+}
+
+#[allow(dead_code)]
+pub enum ProviderStatus {
```
lmuntaner: Nice!!!

lmuntaner: Do we want to store this in stable memory? Otherwise they will get refetched on every initialization, right?

Maybe we can improve later though, not in this PR, but just thinkin ahead.

src/internet_identity/src/openid/generic.rs
```diff
@@ -347,7 +562,9 @@ fn verify_claims(
             claims.nonce
         )));
     }
-    if now > claims.iat * NANOSECONDS_PER_SECOND + MAX_VALIDITY_WINDOW {
+    if now > claims.iat * NANOSECONDS_PER_SECOND + MAX_VALIDITY_WINDOW
+        || now > claims.exp * NANOSECONDS_PER_SECOND
```
lmuntaner, line 344: Why is this `now > claims.exp * NANOSECONDS_PER_SECOND` added now?

src/internet_identity/tests/integration/openid.rs
```diff
@@ -359,16 +360,33 @@ impl Claims {
     }
 }
 
+#[derive(Serialize, Deserialize)]
```
lmuntaner, line 362: Is this a copy of the other type?

src/internet_identity/internet_identity.did
sea-snake: Yeah, it's for the icons in the buttons

lmuntaner: Can we rely on this URLs? Couldn't the inject something if turned malicious? I guess something for security to tell us

PR #3288, opened 2025-08-21
Fix bug where current passkey is wrongly set when removing it.

lmuntaner:
# Motivation

We had a bug that removing the current passkey didn't trigger a logout. Or the opposite, removing the non-current passkey triggered a logout.

The problem was that there were two diffrent methods to check which one was the current passkey and there was a bug in one of them (the one in identity-info) because it checked the last used method from the stored last used identity.

This was true until we started looking up the credentials and ignoring the cached credential in the identity.

# Changes

# Tests


lmuntaner:
Bug:

https://github.com/user-attachments/assets/b1e5633d-bd2a-40fa-abc2-d6105fd320c9

sea-snake (approved):
Good catch!

PR #3289, opened 2025-08-21
Implement generic OpenID provider

sea-snake:
Implement generic OpenID provider

# Changes

- Copy `google.rs` and rename to `generic.rs`.
- Add `exp` claim and verify it.
- Use `secs_to_nanos` fn instead of `NANOSECONDS_PER_SECOND` constant.
- Remove picture claim since it's not supported by all identity providers.

# Tests

- Copied and cleaned up all tests from `google.rs` into `generic.rs` .

### DIFF
--- a/src/internet_identity/src/openid/generic.rs
+++ b/src/internet_identity/src/openid/generic.rs
@@ -65,7 +65,7 @@ struct Claims {
     nonce: String,
     exp: u64,
     iat: u64,
-    // Optional Google specific claims
+    // Optional metadata claims
     email: Option<String>,
     name: Option<String>,
 }
@@ -176,8 +176,8 @@ fn schedule_fetch_certs(
     set_timer(Duration::from_secs(delay.unwrap_or(0)), move || {
         spawn(async move {
             let new_delay = match fetch_certs(jwks_uri.clone()).await {
-                Ok(google_certs) => {
-                    certs_reference.replace(google_certs);
+                Ok(certs) => {
+                    certs_reference.replace(certs);
                     FETCH_CERTS_INTERVAL
                 }
                 // Try again earlier with backoff if fetch failed, the HTTP outcall responses
@@ -284,8 +284,8 @@ fn create_rsa_public_key(jwk: &Jwk) -> Result<RsaPublicKey, String> {
 #[allow(clippy::needless_pass_by_value)]
 fn verify_signature(input: VerificationInput, jwk: &Jwk) -> Result<(), SignatureVerificationError> {
     // Ensure the algorithm specified in the JWT header matches the expected algorithm (RS256).
-    // JSON Web Keys (JWK) returned from Google API (v3) always use RSA with SHA-256.
     // If the algorithm does not match, return an UnsupportedAlg error.
+    // Additional algorithms can be implemented here if needed in the future.
     if input.alg != RS256 {
         return Err(SignatureVerificationErrorKind::UnsupportedAlg.into());
     }


src/internet_identity/src/openid/generic.rs
```diff
@@ -0,0 +1,627 @@
...
+) {
+    use ic_cdk::spawn;
+    use ic_cdk_timers::set_timer;
+    use std::cmp::min;
+    use std::time::Duration;
+
+    set_timer(Duration::from_secs(delay.unwrap_or(0)), move || {
+        spawn(async move {
+            let new_delay = match fetch_certs(jwks_uri.clone()).await {
+                Ok(google_certs) => {
```
lmuntaner: Rename `google_certs`

src/internet_identity/src/openid/generic.rs
```diff
@@ -0,0 +1,627 @@
...
+struct Claims {
+    iss: String,
+    sub: String,
+    aud: String,
+    nonce: String,
+    exp: u64,
+    iat: u64,
+    // Optional Google specific claims
+    email: Option<String>,
+    name: Option<String>,
```
lmuntaner, line 70: Should we remove it if we are not using it?

src/internet_identity/src/openid/generic.rs
```diff
@@ -0,0 +1,627 @@
...
+#[derive(Deserialize)]
+struct Claims {
+    iss: String,
+    sub: String,
+    aud: String,
+    nonce: String,
+    exp: u64,
+    iat: u64,
+    // Optional Google specific claims
+    email: Option<String>,
```
lmuntaner, line 69: Is `email` optional as well? If so, we should better pass it from the frontend, no?

src/internet_identity/src/openid/generic.rs
```diff
@@ -0,0 +1,627 @@
...
+/// Verifier implementation for `identity_jose` that verifies the signature of a JWT.
+///
+/// - `input`: A `VerificationInput` struct containing the JWT's algorithm (`alg`),
+///   the signing input (payload to be hashed and verified), and the decoded signature.
+/// - `jwk`: A reference to a `Jwk` (JSON Web Key) that contains the RSA public key
+///   parameters (`n` and `e`) used to verify the JWT signature.
+#[allow(clippy::needless_pass_by_value)]
+fn verify_signature(input: VerificationInput, jwk: &Jwk) -> Result<(), SignatureVerificationError> {
+    // Ensure the algorithm specified in the JWT header matches the expected algorithm (RS256).
+    // JSON Web Keys (JWK) returned from Google API (v3) always use RSA with SHA-256.
```
lmuntaner: Remove `Google` reference

lmuntaner (commented):
A few questions but looking good!

src/internet_identity/src/openid/generic.rs
sea-snake: Ah, we're actually using these (to render labels in the dashboard UI), I'll rename this comment above them though.

src/internet_identity/src/openid/generic.rs
sea-snake: No, since both these claims are to identify the linked account from the identity provider in the dashboard. While the name entered during registration is to name your identity. 

In practice this means that if you have e.g. an Apple account linked in the dashboard you'll see an email instead of your apple account name.

src/internet_identity/src/openid/generic.rs
sea-snake: Updated the comment.

src/internet_identity/src/openid/generic.rs
lmuntaner: The `email` claim is used to identify the linked account. So, why is it optional?

src/internet_identity/src/openid/generic.rs
lmuntaner: The `name` in this claim goes to the metadata of the openID credential which then will be used in the UI as a label in the access methods list, right?

src/internet_identity/src/openid/generic.rs
sea-snake: Since some OpenID provider might only have a name or email (vice versa). In the frontend we can in the worst case (neither is set) fallback to showing the sub claim. This allows for the most diverse support of open id providers.

src/internet_identity/src/openid/generic.rs
sea-snake: Yes exactly, same with the email.

src/internet_identity/src/openid/generic.rs
lmuntaner: Isn't the email checked to ensure that the same openID isn't added in two identities?

sea-snake: No, we check the iss + sub pair for that, both the name and email are only for identifying linked accounts in the dashboard.

PR #3296, opened 2025-08-22
Add canister flag to enable open id generic in frontend

lmuntaner:
# Motivation

Be able to enable open id generic feature without using a new wasm.

# Changes

Common changes for new config fields:

* Added `feature_flag_enable_generic_open_id_fe` to the `InternetIdentityInit` type in both the backend (`internet_identity.did`, `types.rs`) and frontend test configurations, allowing the flag to be set during initialization.
* Updated the persistent state struct and its default implementation to include the new feature flag.
* Extended the storage layer (`StorablePersistentState`) and its conversion implementations to persist and restore the new feature flag. 
* Modified the canister configuration logic to read and apply the new feature flag from install arguments and persistent state. 
* Updated integration and unit tests to include the new feature flag in relevant configuration objects, ensuring coverage and correct initialization in test scenarios.

# Tests

CI checks are enough.


sea-snake:
I don't see ENABLE_GENERIC_OPEN_ID being initialized with the canister config value?

PR #3298, opened 2025-08-22
feat: Proper Http request handling

aterga:
# Motivation

Clients want to get more meaningful responses for OPTIONS requests. 

# Changes

Remove http_request_update and handle OPTIONS, GET requests via query calls.

# Tests

Manual testing, cargo test

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -1130,15 +1130,15 @@ jobs:
             exit 1
           fi
 
-  interface-compatibility:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: ./.github/actions/setup-didc
-      - name: "Check canister interface compatibility"
-        run: |
-          curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity.did -o internet_identity_previous.did
-          didc check src/internet_identity/internet_identity.did internet_identity_previous.did
+  # interface-compatibility:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v4
+  #     - uses: ./.github/actions/setup-didc
+  #     - name: "Check canister interface compatibility"
+  #       run: |
+  #         curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity.did -o internet_identity_previous.did
+  #         didc check src/internet_identity/internet_identity.did internet_identity_previous.did
 
   sig-verifier-js:
     runs-on: ubuntu-latest
--- a/docs/ii-spec.mdx
+++ b/docs/ii-spec.mdx
@@ -616,7 +616,7 @@ See [here](vc-spec.md) for the specification of the verifiable credentials proto
 
 ### HTTP gateway protocol
 
-The methods `http_request` and `http_request_update` serve the front-end assets to browesers. See [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec) for additional details.
+The method `http_request` serve the front-end assets to browesers. See [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec) for additional details.
 
 ### Internal APIs
 
--- a/docs/ii-spec.mdx
+++ b/docs/ii-spec.mdx
@@ -616,7 +616,7 @@ See [here](vc-spec.md) for the specification of the verifiable credentials proto
 
 ### HTTP gateway protocol
 
-The method `http_request` serve the front-end assets to browesers. See [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec) for additional details.
+The method `http_request` serve the front-end assets to browsers. See [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec) for additional details.
 
 ### Internal APIs
 
--- a/src/frontend/src/hooks.ts
+++ b/src/frontend/src/hooks.ts
@@ -1,4 +1,4 @@
-import { Reroute, redirect } from "@sveltejs/kit";
+import { Reroute } from "@sveltejs/kit";
 import { WEBAUTHN_IFRAME_PATH } from "$lib/legacy/flows/iframeWebAuthn";
 import { getAddDeviceAnchor } from "$lib/utils/addDeviceLink";
 import { nonNullish } from "@dfinity/utils";
--- a/src/frontend/src/hooks.ts
+++ b/src/frontend/src/hooks.ts
@@ -1,4 +1,4 @@
-import { Reroute } from "@sveltejs/kit";
+import { Reroute, redirect } from "@sveltejs/kit";
 import { WEBAUTHN_IFRAME_PATH } from "$lib/legacy/flows/iframeWebAuthn";
 import { getAddDeviceAnchor } from "$lib/utils/addDeviceLink";
 import { nonNullish } from "@dfinity/utils";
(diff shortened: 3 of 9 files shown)


.github/workflows/canister-tests.yml
```diff
@@ -1130,15 +1130,15 @@ jobs:
...
-          didc check src/internet_identity/internet_identity.did internet_identity_previous.did
+  # interface-compatibility:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v4
+  #     - uses: ./.github/actions/setup-didc
+  #     - name: "Check canister interface compatibility"
+  #       run: |
+  #         curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity.did -o internet_identity_previous.did
+  #         didc check src/internet_identity/internet_identity.did internet_identity_previous.did
```
aterga, line 1141: Justification — we don't need the http_request_update function anymore since it was used back when v1 certification prevented redirects for the /faq endpoint. Without commenting this out, we were getting 

```
Run curl -sSL https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity.did -o internet_identity_previous.did
Error: Method http_request_update is only in the expected type
Error: Process completed with exit code 1.
```

See https://github.com/dfinity/internet-identity/actions/runs/17159360058/job/48684352155?pr=3298

docs/ii-spec.mdx
```diff
@@ -616,7 +616,7 @@ See [here](vc-spec.md) for the specification of the verifiable credentials proto
 
 ### HTTP gateway protocol
 
-The methods `http_request` and `http_request_update` serve the front-end assets to browesers. See [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec) for additional details.
+The method `http_request` serve the front-end assets to browesers. See [here](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec) for additional details.
```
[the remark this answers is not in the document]
lmuntaner: Bots are good at catching typos :)

src/internet_identity/src/http.rs
```diff
@@ -7,22 +7,79 @@ use serde_bytes::ByteBuf;
...
+            "Access-Control-Allow-Headers".to_string(),
+            "Content-Type".to_string(),
+        ),
+        ("Content-Length".to_string(), "0".to_string()),
+    ];
+
+    headers.append(&mut cors_safe_security_headers());
+
+    HttpResponse {
+        status_code: 204,
```
lmuntaner, line 16: Did yuo find out why is better to return 204? Shall we add it as a comment?

src/internet_identity/src/http.rs
```diff
@@ -7,22 +7,79 @@ use serde_bytes::ByteBuf;
...
-            upgrade: Some(true),
+            headers: vec![
+                (
+                    "Location".to_string(),
+                    "https://identitysupport.dfinity.org/hc/en-us".to_string(),
+                ),
+                ("Cache-Control".to_string(), "max-age=31536000".to_string()), // Cache for 1 year
+            ],
+            body: ByteBuf::from("Moved permanently to FAQ..."),
+            upgrade: None,
```
lmuntaner: Didn't we say we keep `/faq`? It's still used in II 1.0

aterga: The PR does keep the /faq handler, but now that we don't have http_request_update, we don't need to use the `upgrade` field. IIUC, the old behavior was 

user -> BN -> query http_request -> update http_request_update -> redirect to https://identitysupport.dfinity.org/hc/en-us

and now we can have just 

user -> BN -> query http_request -> redirect to https://identitysupport.dfinity.org/hc/en-us

Please correct me if this sounds wrong. Otherwise, I'm going to test this thoroughly before merging

src/internet_identity/src/http.rs
aterga: This makes sense as per https://www.rfc-editor.org/rfc/rfc9110#name-204-no-content

```suggestion
        // Indicates success without any additional content to be sent in the response content.
        status_code: 204,
```

src/internet_identity/src/http.rs
lmuntaner: I'm not sure it was like that but the solution sounds good. Please test it. I can add your principal to the staging canisters wallet so that you can deploy and test it there.

src/internet_identity/src/http.rs
lmuntaner: Thanks, yes, makes a lot of sense!

src/frontend/src/hooks.ts
```diff
@@ -1,4 +1,4 @@
-import { Reroute } from "@sveltejs/kit";
+import { Reroute, redirect } from "@sveltejs/kit";
```
aterga:
```suggestion
import { Reroute } from "@sveltejs/kit";
```

src/internet_identity/src/http.rs
```diff
@@ -7,24 +7,69 @@ use serde_bytes::ByteBuf;
 
 mod metrics;
 
-pub fn http_request(req: HttpRequest) -> HttpResponse {
-    let parts: Vec<&str> = req.url.split('?').collect();
```
aterga, line 11: The code form this function was moved to `http_get_request`

src/internet_identity/src/http.rs
```diff
@@ -89,19 +115,40 @@ pub fn security_headers(
         });
 
     vec![
+        // X-Frame-Options: DENY
```
aterga, line 118: The comments below were generated by Claude Sonnet 4, seem reasonable so I decided to keep them.

src/internet_identity/src/http.rs
aterga: Tested on https://fgte5-ciaaa-aaaad-aaatq-cai.ic0.app/faq - seems to work.

src/internet_identity/src/http.rs
```diff
@@ -7,24 +7,24 @@ use serde_bytes::ByteBuf;
...
+    }
+}
+
+fn http_get_request(url: String, certificate_version: Option<u16>) -> HttpResponse {
+    let parts: Vec<&str> = url.split('?').collect();
+
     match parts[0] {
-        // The FAQ used to live in '/faq' but we now use an external website. We redirect in order to not
-        // break existing links in the wild.
-        "/faq" => HttpResponse {
```
lmuntaner, line 15: So, instead of managing the `/faq` in the backend, you moved to redirecting from the frontend, is that right?

src/internet_identity/src/http.rs
lmuntaner: Thanks!

lmuntaner (commented):
A couple of questions but it looks ready!

sea-snake (approved):
As discussed in a call, PR keeps most implementation "as is" while adding the minimal changes required to support OPTIONS request.

src/internet_identity/src/http.rs
aterga: Yes, I think this is the best way to do it.

.github/workflows/canister-tests.yml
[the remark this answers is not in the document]
aterga: This was discussed with the team, we decided it's the right think to remove this function.

src/internet_identity/src/http.rs
lmuntaner: Sounds good!

aterga:
Converting to draft since it's supposed to be included in not the next but the second-next release

PR #3300, opened 2025-08-22
Change event in v2 authentication funnel

lmuntaner:
# Motivation

Safari sometimes blocks WebAuthn interactions if there is some async activity between the click and the webauthn call.

Therefore, we should move the analytics call to after the passkey creation.

# Changes

* Changed the event name from `start-webauthn-creation` to `successful-webauthn-creation` throughout the codebase for more accurate tracking.
* Moved the analytics trigger in `AuthFlow.createPasskey` to fire after passkey creation, which avoids issues with Safari blocking the event if triggered too early.

# Tests

No new functionality. CI checks suffice.

### DIFF
--- a/src/frontend/src/lib/flows/authFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authFlow.svelte.ts
@@ -135,10 +135,6 @@ export class AuthFlow {
   };
 
   createPasskey = async (): Promise<bigint> => {
-    authenticationV2Funnel.trigger(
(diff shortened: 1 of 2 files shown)


sea-snake:
> @sea-snake I looked at the data. Safari has a drop-off 67% in the registration, while Chrome has 72%. I don't think that it's because of the blocked WebAuthn, what do you think?

I agree, let's close this PR for now, we can always revisit later.

PR #3301, opened 2025-08-25
Allow users to upgrade again.

lmuntaner:
# Motivation

There are some cases where users need to upgrade a second time. For example, if they lose the passkey they created in id.ai they would be blocked even though they have access to their access methods from 1.0.

I also realized that we can know whether the user upgraded or not before authenticating, using the lookup of the authenticators. Therefore, I refactored to flow to do the check first. Which means that on upgrading again, the user need to authenticate because it didn't before. This is nicer because we about one biometric before telling the user that it's already upgraded.

# Changes

* Refactored the migration flow to detect already-migrated identities by checking if any device is associated with a new flow origin using the new `isNewOriginDevice` utility, instead of relying on authentication method types. When such a device is found, the wizard now shows an `alreadyMigrated` view.
* Added an `upgradeAgain` handler in the migration flow and UI, allowing users to retry the upgrade process from the `alreadyMigrated` screen.
* Updated the `AlreadyMigrated` component to display clearer instructions and provide an "Upgrade again" button, improving user guidance.
* Changed the help button style from tertiary to secondary for better visibility and make it more prominent than "Upgrade again"
* Removed the legacy `alreadyMigrated` logic from the migration wizard, shifting responsibility to the migration flow class.
* Added comprehensive unit tests for the new `isNewOriginDevice` utility to ensure correct detection of migrated devices based on origin and purpose.

These changes make the migration process more robust and user-friendly, especially for users who have already upgraded their identity.

# Tests

* Tested locally as seen in the video.
* Added tests for the new util `isNewOriginDevice`.

### DIFF
--- a/src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
+++ b/src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
@@ -8,7 +8,8 @@
     UpgradeIdentityEvents,
   } from "$lib/utils/analytics/upgradeIdentityFunnel";
 
-  let { name }: { name: string } = $props();
+  let { name, onUpgradeAgain }: { name: string; onUpgradeAgain: () => void } =
+    $props();
 
   onMount(() => {
     upgradeIdentityFunnel.trigger(UpgradeIdentityEvents.AlreadyMigratedScreen);
@@ -25,9 +26,14 @@
         Identity already upgraded
       </h1>
       <p
-        class="text-md text-text-tertiary font-medium text-balance sm:text-center"
+        class="text-md text-text-tertiary mb-2 font-medium text-balance sm:text-center"
       >
         {`${name} is already upgraded to the new experience.`}
+      </p>
+      <p
+        class="text-md text-text-tertiary font-medium text-balance sm:text-center"
+      >
+        Select "Use existing Passkey" in the login flow.
       </p>
     </div>
   </div>
@@ -36,10 +42,13 @@
       href={SUPPORT_URL}
       target="_blank"
       rel="noopener noreferrer"
-      variant="tertiary"
+      variant="secondary"
       size="lg"
     >
       Help & FAQ
+    </Button>
+    <Button onclick={onUpgradeAgain} variant="tertiary" size="lg">
+      Upgrade again
     </Button>
   </div>
 </div>
--- a/src/frontend/src/lib/flows/migrationFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/migrationFlow.svelte.ts
@@ -28,6 +28,7 @@ import {
 } from "$lib/utils/analytics/upgradeIdentityFunnel";
 import { features } from "$lib/legacy/features";
 import { DiscoverableDummyIdentity } from "$lib/utils/discoverableDummyIdentity";
+import { isNewOriginDevice } from "$lib/utils/accessMethods";
 
 export class WrongDomainError extends Error {
   constructor() {
@@ -40,9 +41,11 @@ export class WrongDomainError extends Error {
 }
 
 export class MigrationFlow {
-  view = $state<"enterNumber" | "enterName">("enterNumber");
+  view = $state<"enterNumber" | "enterName" | "alreadyMigrated">("enterNumber");
   identityNumber: UserNumber | undefined;
   #webAuthFlows: { flows: WebAuthnFlow[]; currentIndex: number } | undefined;
+  #attachElement?: HTMLElement;
+  #devices: Omit<DeviceData, "alias">[] = [];
 
   constructor() {
     this.identityNumber = undefined;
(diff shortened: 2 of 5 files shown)


src/frontend/src/lib/flows/migrationFlow.svelte.ts
```diff
@@ -53,85 +56,32 @@ export class MigrationFlow {
...
-          currentIndex: this.#webAuthFlows.currentIndex + 1,
-        };
-        throw new WrongDomainError();
-      }
-      throw error;
+  upgradeAgain = () => {
+    if (isNullish(this.identityNumber) || isNullish(this.#attachElement)) {
+      // The identityNumber and attachElement are set in authenticateWithIdentityNumber from "enterNumber" view.
+      this.view = "enterNumber";
+      throw new Error("Enter your identity number and try again.");
```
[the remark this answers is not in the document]
lmuntaner: Makes sense, we're not handling this error anyway.

lmuntaner:
Video of upgrading again:


https://github.com/user-attachments/assets/d0b6ecc3-b54c-4eaa-ad5d-15552d353e08


Video of the UI changes in the already migrated screen:

Screenshot 2025-08-25 at 10 00 17

sea-snake (changes requested):
See my other comment

src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
```diff
@@ -25,21 +26,28 @@
...
       </h1>
       <p
-        class="text-md text-text-tertiary font-medium text-balance sm:text-center"
+        class="text-md text-text-tertiary mb-2 font-medium text-balance sm:text-center"
       >
         {`${name} is already upgraded to the new experience.`}
       </p>
+      <!-- Removed "text-balance" because it breaks the button name `"Use existing Passkey"` into two lines -->
+      <p class="text-md text-text-tertiary font-medium sm:text-center">
+        Select "Use existing Passkey" in the login process.
```
sea-snake: Let's highlight with styling instead, similar to how we're doing it in other views.

```suggestion
        Select <b class="text-text-primary">Use existing Passkey</b> in the login process.
```

src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
```diff
@@ -25,21 +26,28 @@
         Identity already upgraded
       </h1>
       <p
-        class="text-md text-text-tertiary font-medium text-balance sm:text-center"
+        class="text-md text-text-tertiary mb-2 font-medium text-balance sm:text-center"
       >
         {`${name} is already upgraded to the new experience.`}
```
sea-snake: As mentioned in my other comment, `name` is undefined.

src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
```diff
@@ -25,21 +26,28 @@
...
       </p>
+      <!-- Removed "text-balance" because it breaks the button name `"Use existing Passkey"` into two lines -->
```
sea-snake: You can use the `whitespace-nowrap` class on the suggested `<b>` tag to prevent that from happening to it.

src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
lmuntaner: Yes, much better!

src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
lmuntaner: Ditto, and I'll remove the prop as well

src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
lmuntaner: Good thinking!

src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
```diff
@@ -24,10 +24,17 @@
...
+      >
+        This identity has already been upgraded to the new experience.
+      </p>
       <p
         class="text-md text-text-tertiary font-medium text-balance sm:text-center"
       >
-        {`${name} is already upgraded to the new experience.`}
+        Select <b class="text-text-primary whitespace-nowrap"
+          >Use existing Passkey</b
+        > in the login process.
```
sea-snake: Maybe a prefix like this would make the lines wrap better and clarify the intent behind this paragraph?

```suggestion
        Authenticate by selecting <b class="text-text-primary whitespace-nowrap"
          >Use existing Passkey</b
        > in the login process.
```

or 

```suggestion
        You can continue with <b class="text-text-primary whitespace-nowrap"
          >Use existing Passkey</b
        > in the login process.
```

PR #3303, opened 2025-08-25
OpenID and feature flag fixes

sea-snake:
OpenID and feature flag fixes.

# Changes

- Fix OpenID logo sizes, render at size-6 (24px).
- Fix loading state of OpenID buttons.
- Fix disabled state of OpenID and surrounding buttons.
- Use `auth_scope` config when requesting JWT through redirect.
- Fix feature flags, setting value in console now correctly overrides init value (e.g. canister value).
- Fix `isOpenIdCancelError`, also match when error is from a closed OpenID popup window.

# Tests

- Manually verified that feature flags work as expected.
- Manually verified that both existing Google auth and generic OpenID auth are working as expected.

### DIFF
--- a/src/frontend/src/lib/flows/authFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authFlow.svelte.ts
@@ -209,7 +209,6 @@ export class AuthFlow {
         const identityNumber = await this.#registerWithOpenId(jwt);
         return { identityNumber, type: "signUp" };
       }
-      this.#view = "chooseMethod";
       throw error;
     }
   };
@@ -225,6 +224,7 @@ export class AuthFlow {
     const requestConfig: RequestConfig = {
       clientId: config.client_id,
       authURL: config.auth_uri,
+      authScope: config.auth_scope.join(" "),
       configURL: config.fedcm_uri?.[0],
     };
     // Create two try-catch blocks to avoid double-triggering the analytics.
@@ -275,7 +275,6 @@ export class AuthFlow {
         const identityNumber = await this.#registerWithOpenId(jwt);
         return { identityNumber, type: "signUp" };
       }
-      this.#view = "chooseMethod";
       throw error;
     }
   };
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -209,7 +209,6 @@ export class AuthFlow {
         const identityNumber = await this.#registerWithOpenId(jwt);
         return { identityNumber, type: "signUp" };
       }
-      this.#view = "chooseMethod";
```
lmuntaner, line 212: Why remove resetting the view?

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -275,7 +275,6 @@ export class AuthFlow {
...
```
lmuntaner, line 278: ditto

src/frontend/src/lib/state/featureFlags.ts
```diff
@@ -54,8 +54,9 @@ const createFeatureFlagStore = (
     return initializedFeatureFlag;
   };
   const initialize = (): void => {
-    if (nonNullish(getInitValue)) {
-      initializedFeatureFlag.set(getInitValue() ?? defaultValue);
+    // Override feature flag with init method if not already set
```
lmuntaner, line 57: what does `with init method`? Do you mean with canister configuration value?

lmuntaner (commented):
A few comments!

src/frontend/src/lib/flows/authFlow.svelte.ts
sea-snake: It's no longer needed since error are handled with manual tooltips instead of closing the dialog and showing a toast.

src/frontend/src/lib/flows/authFlow.svelte.ts
sea-snake: See response on other comment.

src/frontend/src/lib/state/featureFlags.ts
sea-snake: Yes, but it's called "init method" since it could also be something other than canister configuration value.

lmuntaner: aahh, ok

PR #3304, opened 2025-08-25
User can link generic openID providers

lmuntaner:
# Motivation

If users can register and log in with the generic openID providers, they should also be able to link them from the dashboard.

In this PR, we add the openID provider buttons in the modal when the user clicks in "Add +" in the dashboard.

# Changes

* Added a new `linkOpenIdAccount` method to the `AddAccessMethodFlow` class, which handles the authentication flow and linking of generic OpenID accounts.
* Updated the `AddAccessMethod` and `AddAccessMethodWizard` components to accept and handle the new `linkOpenIdAccount` callback, enabling the UI to trigger the OpenID linking flow.
* Modified the UI to display a button for each configured OpenID provider when the generic OpenID feature flag is enabled, showing provider logos and handling authentication state per provider.
* Updated text in the wizard to refer to "identity provider" instead of just "Google account" for broader applicability.

# Tests

* Tested locally with Google and Microsoft.

### DIFF
--- a/src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
@@ -59,6 +59,7 @@ export class AddAccessMethodFlow {
           clientId: config.client_id,
           authURL: config.auth_uri,
           configURL: config.fedcm_uri?.[0],
+          authScope: config.auth_scope.join(" "),
         },
         {
           nonce,
--- a/src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
@@ -12,6 +12,7 @@ import { throwCanisterError } from "$lib/utils/utils";
 import type {
   AuthnMethodData,
   OpenIdCredential,
+  OpenIdConfig,
 } from "$lib/generated/internet_identity_types";
 import { features } from "$lib/legacy/features";
 import { DiscoverableDummyIdentity } from "$lib/utils/discoverableDummyIdentity";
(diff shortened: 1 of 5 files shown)


lmuntaner:
New UI:

Screenshot 2025-08-25 at 13 17 19

src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
```diff
@@ -42,6 +43,47 @@ export class AddAccessMethodFlow {
...
+          clientId: config.client_id,
+          authURL: config.auth_uri,
+          configURL: config.fedcm_uri?.[0],
+        },
+        {
+          nonce,
+          mediation: "required",
+        },
+      );
+      const { iss, sub, aud, name, email } = decodeJWTWithNameAndEmail(jwt);
```
sea-snake, line 69: Name and email aren't guaranteed, this method should be updated to make them optional.

After which below implementation should only add them to the metadata when non nullish.

src/frontend/src/lib/state/featureFlags.ts
```diff
@@ -98,7 +98,7 @@ export const ADD_ACCESS_METHOD = createFeatureFlagStore(
 );
 export const ENABLE_GENERIC_OPEN_ID = createFeatureFlagStore(
   "ENABLE_GENERIC_OPEN_ID",
-  false,
+  true,
```
sea-snake: This should stay on false :D

src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
lmuntaner: Done!

src/frontend/src/lib/state/featureFlags.ts
lmuntaner: Oopsie! Fixed

sea-snake (approved):

PR #3306, opened 2025-08-25
Implement effective OpenID issuer

sea-snake:
Implement effective OpenID issuer, replace placeholders in issuer with JWT claims. 

# Changes

- Implement `replace_issuer_placeholders` fn to handle OpenID providers that have multiple tenants e.g. Microsoft with https://login.microsoftonline.com/{tenantid}/v2.0.
- Use this method when matching provider so that generic provider can be found.
- Use this method to create effective issuer used when verifying JWT claims.
- Use the config issuer instead of claim issuer in storage since the latter could be different within the same OpenID provider as seen in the above tenants example.

# Tests

- Verified Google auth is still working as expected.
- Verified that multi tenant auth now works with a Microsoft OpenID config.

### DIFF
--- a/src/internet_identity/src/openid.rs
+++ b/src/internet_identity/src/openid.rs
@@ -199,7 +199,11 @@ where
     PROVIDERS.with_borrow(|providers| {
         providers
             .iter()
-            .find(|provider| provider.issuer() == claims.iss)
+            .find(|provider| {
+                let effective_issuer =
+                    replace_issuer_placeholders(&provider.issuer(), validation_item.claims());
+                effective_issuer == claims.iss
+            })
             .ok_or_else(|| {
                 OpenIDJWTVerificationError::GenericError(format!(
                     "Unsupported issuer: {}",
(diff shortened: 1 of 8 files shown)


src/internet_identity/src/openid.rs
```diff
@@ -210,6 +214,52 @@ where
...
+pub fn replace_issuer_placeholders(template: &str, claims_bytes: &[u8]) -> String {
+    let claims: serde_json::Value = match serde_json::from_slice(claims_bytes) {
+        Ok(val) => val,
+        Err(_) => return template.to_string(), // If claims cannot be decoded, return template as-is
+    };
+
+    let mut result = String::with_capacity(template.len());
+    let mut remaining = template;
+
+    while let Some(open_pos) = remaining.find('{') {
```
lmuntaner, line 229: Why not use regex?

src/internet_identity/src/openid/generic.rs
```diff
@@ -127,7 +128,12 @@ impl OpenIdProvider for Provider {
             metadata.insert("name".into(), MetadataEntryV2::String(name));
         }
         Ok(OpenIdCredential {
-            iss: claims.iss,
+            // Do NOT use claims.iss here since it could be different within the
+            // same OpenID provider as seen in Microsoft with multiple tenants.
+            //
+            // The issuer returned should therefore ALWAYS be the issuer from the config,
+            // so that credentials are always stored with a single issuer per provider.
```
lmuntaner, line 135: Can we add the example of Microsoft to make it clear what you mean?

lmuntaner (commented):
A couple of questions

src/internet_identity/src/openid.rs
sea-snake: Switched to Regex

src/internet_identity/src/openid/generic.rs
sea-snake: Added examples.

lmuntaner (approved):

src/internet_identity/src/openid.rs
```diff
@@ -210,6 +214,52 @@ where
...
+            }
+
+            // Move past '}'
+            remaining = &remaining[close_pos + 1..];
+        } else {
+            // No closing '}', treat '{' literally and append the rest
+            result.push('{');
+            result.push_str(remaining);
+            return result; // Stop processing
+        }
```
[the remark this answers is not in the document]
sea-snake: Good catch! We wouldn't want something like `https://example.com/{a/{b}` to partially have replacements, rather just have it abandon the replacement altogether in that case.

src/internet_identity/src/openid.rs
sea-snake: As discussed: switched back due to WASM size limit.

PR #3309, opened 2025-08-26
Enable Cached Identity with Generic OpenId Access method

lmuntaner:
# Motivation

Cached identities need to also adapt to use generic open ids accordingly.

# Changes

**OpenID Provider Identification and Display:**

- Added `lastUsedIdentityTypeName` utility to determine the display name of an identity provider, supporting both generic OpenID providers and Google, and integrated it into the UI for last-used identities.

**OpenID Configuration Handling:**

- Refactored OpenID authentication flow to use new `findConfig` and `isOpenIdConfig` utilities, ensuring the correct configuration is selected for the given issuer and improving error handling for unknown issuers.
- Added `findConfig` and `isOpenIdConfig` helpers to centralize logic for locating and type-checking OpenID configurations, supporting both Google and generic providers.

**Minor Improvements:**

- Simplified JWT request logic by removing unused login hint validation. The loginHint with Google wasn't matching because it was passing the sub as hint but returning the email.

# Tests

- Tested locally by creating identities with Google, Passkey and Microsoft and testing that they all work.
- Added comprehensive tests for `lastUsedIdentityTypeName`, covering various OpenID and passkey scenarios, and mocking global configuration for accurate coverage.

### DIFF
--- a/src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
@@ -53,7 +53,9 @@ export class AuthLastUsedFlow {
         const issuer = lastUsedIdentity.authMethod.openid.iss;
         const config = findConfig(issuer);
         if (isNullish(config)) {
-          throw new Error(`OpenID is not configured for ${issuer}`);
+          throw new Error(
+            "OpenID authentication is not available for this account.",
+          );
         }
         const requestConfig = isOpenIdConfig(config)
           ? {
--- a/src/frontend/src/lib/utils/lastUsedIdentity.ts
+++ b/src/frontend/src/lib/utils/lastUsedIdentity.ts
@@ -15,4 +15,7 @@ export const lastUsedIdentityTypeName = (
   if ("passkey" in lastUsedIdentity.authMethod) {
     return "Passkey";
   }
+  // Not possible, but in case type safety fails at runtime, we don't want to crash.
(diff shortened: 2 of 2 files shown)


lmuntaner:
UI:

Screenshot 2025-08-26 at 10 37 17

src/frontend/src/lib/utils/openID.ts
```diff
@@ -183,14 +210,11 @@ export const requestJWT = async (
   config: RequestConfig,
   options: RequestOptions,
 ): Promise<string> => {
+  console.log("in da request", config);
   const supportsFedCM = isFedCMSupported(navigator.userAgent, config);
   const jwt = supportsFedCM
     ? await requestWithCredentials(config, options)
     : await requestWithRedirect(config, options);
-  const { loginHint } = decodeJWT(jwt);
-  if (nonNullish(options.loginHint) && loginHint !== options.loginHint) {
```
lmuntaner, line 191: @sea-snake why was this necessary? It wasn't working with Google when using the sub as loginHint it was coming back as the email in the JWT.

sea-snake: This was previously comparing the loginHint with the sub directly, but was updated to use the stored loginHint instead (to support multiple providers). This indeed would break with existing local data as you've noticed, good catch!

Removing it seems fine with me, someone signing in with a different account than they selected from the last used list isn't an issue since we get the identity number afterwards from the JWT, not the storage anyway.

PR #3314, opened 2025-08-27
Prompt for identity name if missing in JWT

sea-snake:
Prompt for identity name if missing in JWT

# Changes

- Add `completeOpenIdRegistration` method to `authFlow`.
- Update `continueWithOpenId` method in `authFlow` to indicate if the above needs to be called and if we already have a name or not.
- Update `AuthWizard` to automatically call `completeOpenIdRegistration` if a name is returned, else it's done manually after a user has entered a name in the `setupNewIdentity` view.
- Merge `decodeJWT` and `decodeJWTWithNameAndEmail` in a single `decodeJWT` method.
- Remove claims from `authenticateWithJWT` method return value, instead callers should use the above `decodeJWT` themselves, this decouples the two methods.
- Add `name` argument to `openid_identity_registration_finish` canister method and use it instead of the JWT name claim.
- Remove `metadata_name` method from `OpenIdProvider` trait and its implementations since it's no longer used.

# Tests

- Verified that identity name screen is shown if JWT is missing name claim (updated `decodeJWT` to intentionally return undefined name since Apple doesn't work locally).
- Verified that identity name screen is skipped if JWT contains name claim.

https://github.com/user-attachments/assets/3bc3ff35-0d90-4aa8-b546-0cbfc2e62dcc

### DIFF
--- a/src/frontend/src/lib/utils/iiConnection.test.ts
+++ b/src/frontend/src/lib/utils/iiConnection.test.ts
@@ -784,6 +784,7 @@ describe("Connection.login", () => {
 const mockJwt = "mock.jwt.token";
 const mockSalt = [1, 2, 3];
 const mockNonce = "mock-nonce";
+const mockName = "John Doe";
 
 // Mock the openID utilities
 vi.mock("$lib/utils/openID", () => ({
@@ -791,7 +792,7 @@ vi.mock("$lib/utils/openID", () => ({
     Promise.resolve({ nonce: mockNonce, salt: mockSalt }),
   createGoogleRequestConfig: () => ({}),
   requestJWT: () => Promise.resolve(mockJwt),
-  decodeJWT: () => ({ name: "John Doe" }),
+  decodeJWT: () => ({ name: mockName }),
 }));
 
 vi.mock("$lib/templates/loader", () => ({
@@ -856,6 +857,7 @@ describe("openid_identity_registration_finish", () => {
     expect(mockActor.openid_identity_registration_finish).toHaveBeenCalledWith({
       jwt: mockJwt,
       salt: mockSalt,
+      name: mockName,
     });
   });
 


src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
```diff
@@ -112,9 +113,12 @@
...
     isAuthenticating = true;
     try {
-      const { identityNumber, type } =
-        await authFlow.continueWithOpenId(config);
-      (type === "signUp" ? onSignUp : onSignIn)(identityNumber);
+      const result = await authFlow.continueWithOpenId(config);
+      if (result.type === "signIn") {
+        onSignIn(result.identityNumber);
+      } else if (nonNullish(result.name)) {
+        onSignUp(await authFlow.completeOpenIdRegistration(result.name));
```
lmuntaner, line 120: Do we want the logic of completing the registration in the component? Why not continuing within  `continueWithOpenId` and that returns a type "signedIn", "signedUp", "enterName"?

I'd like to be able to test all the flow directly in the class at some point, so having the logic outside goes a bit against that idea.

lmuntaner (approved):
LGTM! Just a few comments but already approving.

I would appreciate a video as well :)

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
sea-snake: I was considering this, but opted to let the `authFlow` be the one in control of the flow (which views to show), not the wizard. So from that perspective, the `authFlow` return values shouldn't indicate what view to show, instead it should the the `view` property from the flow.

Overall, not really happy with the current flow approach, I still feel like there's a better approach but don't see it right now :/

lmuntaner: Yes, I understand your feeling. I have the same and even if we did this change, I wouldn't be super happy with the status anyway.

PR #3317, opened 2025-08-27
Adapt analytics to generic open id providers

lmuntaner:
# Motivation

This pull request updates the authentication analytics to generalize Google-specific events and properties to OpenID, improving support for multiple OpenID providers and making analytics tracking more flexible.

The changes include renaming analytics events and adding provider information to analytics properties, as well as updating related code to use these new conventions.

# Changes

**Analytics event generalization and property enhancements:**

* Renamed all Google-specific analytics events in `AuthenticationV2Events` and throughout the authentication flow to their OpenID equivalents (e.g., `ContinueWithGoogle` → `ContinueWithOpenID`, `RegisterWithGoogle` → `RegisterWithOpenID`, etc.), ensuring consistent terminology for multiple OpenID providers.
* Added `addProperties` method to the `Funnel` analytics utility, allowing dynamic addition of properties (such as provider name) to funnel events.
* Updated authentication flows to call `addProperties` with the provider name (e.g., "Google" or the OpenID config name) before triggering analytics events, so analytics data includes which provider was used.

**Frontend integration improvements:**

* Updated authentication flow handlers in `+page.svelte` files to add provider information for OpenID identities and trigger the analytics events that were missing in this flow.

# Tests

Tested locally by adding a console.log within the trigger function. That's how I discovered the missing trigger.


src/frontend/src/lib/utils/analytics/Funnel.ts
```diff
@@ -56,6 +56,13 @@ export class Funnel<T extends Record<string, string>> {
     }
   }
 
+  addProperties(additionalProperties: Record<string, string | number>): void {
+    this.#properties = {
+      ...(this.#properties || {}),
```
sea-snake: Why `||` instead of `??`?

sea-snake (approved):
Besides one nit, LGTM

src/frontend/src/lib/utils/analytics/Funnel.ts
lmuntaner: Autocomplete AI... I'll change it

PR #3318, opened 2025-08-27
Copy google integration tests for openid integration test

sea-snake:
Copy google integration tests for openid integration test, using `openid_configs` instead of `openid_google`.

# Changes

- Copied renamed `integration/openid.rs` to `google.rs` and created a copy `openid.rs`.
- Use `openid_configs` deploy arg for these copied integration tests
- Keep copied integration tests for now as is (we can later look into adding more test data and/or generated test data).
- Do the same as above for `should_report_active_google_authn_methods`.

# Tests

- No manual tests needed, new integration tests should pass in CI

### DIFF
--- a/src/internet_identity/tests/integration/activity_stats/authn_methods.rs
+++ b/src/internet_identity/tests/integration/activity_stats/authn_methods.rs
@@ -302,7 +302,7 @@ fn should_report_active_openid_authn_methods() {
 }
 
 /// Tests that active Google authn_methods are counted correctly.
-/// TODO: ID-155 Create OpenID mock data on-demand and use this instead in unit/integration tests
+/// TODO: Remove this once we remove Google specific OpenID implementation
 #[test]
 fn should_report_active_google_authn_methods() {
     // Create II instance that mocks Google certs


src/internet_identity/tests/integration/activity_stats/authn_methods.rs
```diff
@@ -301,6 +301,68 @@ fn should_report_active_openid_authn_methods() {
     );
 }
 
+/// Tests that active Google authn_methods are counted correctly.
+/// TODO: Remove this once we remove Google specific OpenID implementation
+#[test]
+fn should_report_active_google_authn_methods() {
```
lmuntaner, line 307: Why copy this just for Google?

sea-snake: Since I want to keep test coverage for both google and generic openid implementations. Once we remove the google specific implementation, this copy can be removed too (as mentioned in the TODO comment).

sea-snake: (The google one tests the Google specific OpenID implementation while the non google one tests the generic OpenID implementation)

This PR does intentionally not yet extend the integration tests with more test data (e.g. other OpenID providers or provider data generated on demand). This would be an exercise for another PR and first requires scoping (what do we test in integration vs e2e etc.

lmuntaner: Ok, noted!

PR #3320, opened 2025-08-28
Fix microsoft continue as flow

lmuntaner:
# Motivation

There was a bug that a user couldn't log in with a cached identity with Microsoft.

The problem was that the no config was because the issuer of the cached identity wasn't matching the config issuer because Microsoft's issuer uses a templating string.

I also found that Microsoft doesn't support "sub" as loginHint. So I removed it and only use the loginHint. We can improve in another PR how this is set.

# Changes

* Added the `issuerMatches` function to support matching issuer URLs with template placeholders (e.g., `{tid}`) in OpenID configuration, ensuring each placeholder matches exactly one path segment. This improves compatibility with providers like Microsoft.
* Updated and refactored the `findConfig` function to use `issuerMatches` for issuer lookup, and clarified the logic for selecting Google and generic OpenID configurations based on the `ENABLE_GENERIC_OPEN_ID` feature flag.
* Introduced a `GOOGLE_ISSUER` constant to avoid hardcoding the Google issuer string in multiple places.
* Simplified the assignment of the `loginHint` parameter in the `AuthLastUsedFlow` by removing the fallback to the `sub` property, now using only `loginHint`.

# Tests

* Tested manually that I can log in with cached identity Google and Microsoft locally.
* Added comprehensive unit tests for `issuerMatches` and `findConfig`, covering exact matches, template matches, edge cases, and feature flag behavior in `openID.test.ts`.


src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -69,9 +69,7 @@ export class AuthLastUsedFlow {
         const jwt = await requestJWT(requestConfig, {
           nonce: get(sessionStore).nonce,
           mediation: "optional",
-          loginHint:
-            lastUsedIdentity.authMethod.openid.loginHint ??
-            lastUsedIdentity.authMethod.openid.sub,
+          loginHint: lastUsedIdentity.authMethod.openid.loginHint,
```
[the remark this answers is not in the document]
lmuntaner: `loginHint` is optional

lmuntaner, line 75: The sub wasn't working with microsoft. It was putting that same sub in the email field, which was wrong and extremely confusing.

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -273,10 +286,13 @@ export class AuthFlow {
...
       const info =
         await get(authenticatedStore).actor.get_anchor_info(identityNumber);
+      const authnMethod = info.openid_credentials[0]?.find(
+        (method) => method.iss === iss,
+      );
       lastUsedIdentitiesStore.addLastUsedIdentity({
         identityNumber,
         name: info.name[0],
-        authMethod: { openid: { iss, sub } },
+        authMethod: { openid: { iss, sub, metadata: authnMethod?.metadata } },
```
sea-snake: `loginHint` is missing

src/frontend/src/lib/utils/openID.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    expect(
+      issuerMatches(
+        "https://accounts.google.com",
+        "https://accounts.google.com",
+        [],
+      ),
+    ).toBe(true);
+  });
+
+  it.only("returns false for half match", () => {
```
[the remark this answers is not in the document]
lmuntaner: done!

src/frontend/src/lib/flows/authFlow.svelte.ts
lmuntaner: Thanks! Done

sea-snake (approved):

PR #3323, opened 2025-08-28
Store issuer from claim in OpenID credential

sea-snake:
Store issuer from claim in OpenID credential and additionally store claims found as placeholders in issuer config as metadata.

# Changes

- Store iss claim found in JWT instead of config issuer in `OpenIdCredential`.
- Add claims found as placeholder in config issuer to credential metadata.
- Split `replace_issuer_placeholders` into multiple smaller functions so it parts of it can be used separately.
- Add `config_issuer` fn to `OpenIdCredential`, this method finds the issuer of the matching config.
- Use the above to add `ConfigIss` to `AuthorizationKey` so that metrics can keep track per issuer as defined in config instead of as defined in JWT.
  - This avoids a possibly large number of issuers in our metrics.
  - This avoids leaking which issuer tenants are used with II through our metrics.

# Tests

Updated the unit tests since `replace_issuer_placeholders` was split into multiple functions.

### DIFF
--- a/src/internet_identity/src/openid.rs
+++ b/src/internet_identity/src/openid.rs
@@ -140,6 +140,29 @@ impl OpenIdCredential {
             }
         })
     }
+
+    /// Find current config for stored credential
+    pub fn config_issuer(&self) -> Option<String> {
+        PROVIDERS.with_borrow(|providers| {
+            providers
+                .iter()
+                .find(|provider| {
+                    let issuer_placeholders = get_issuer_placeholders(&provider.issuer());
+                    let mut issuer_claims: Vec<(String, String)> = vec![];
+                    for key in issuer_placeholders {
+                        if let Some(MetadataEntryV2::String(value)) = self.metadata.get(&key) {
+                            issuer_claims.push((key, value.to_string()));
+                        } else {
+                            return false;
+                        }
+                    }
+                    let effective_issuer =
+                        replace_issuer_placeholders(&provider.issuer(), &issuer_claims);
+                    effective_issuer == self.iss
+                })
+                .map(|provider| provider.issuer())
+        })
+    }
 }
 
 pub trait OpenIdProvider {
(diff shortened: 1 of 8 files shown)


src/internet_identity/src/openid.rs
```diff
@@ -140,6 +140,29 @@ impl OpenIdCredential {
             }
         })
     }
+
+    /// Find current config for stored credential
+    pub fn config_issuer(&self) -> Option<String> {
```
lmuntaner, line 145: Could we add some tests for this? Or too much effort?

Maybe if we move the function inside "find" as a pure function and test that?

src/internet_identity/src/openid.rs
```diff
@@ -190,69 +213,75 @@ where
...
 /// As seen in <https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration>,
-/// the Microsoft issuer uri is dynamic based on the `tid` claim, this method makes sure to
-/// replace the placeholders like e.g. `tid` in the issuer uri with the corresponding claims.
-pub fn replace_issuer_placeholders(template: &str, claims_bytes: &[u8]) -> String {
-    let Ok(claims) = serde_json::from_slice::<serde_json::Value>(claims_bytes) else {
-        return template.to_string(); // If claims cannot be decoded, return template as-is
-    };
-
-    let mut result = String::with_capacity(template.len());
+/// the Microsoft issuer uri is dynamic based on placeholders, this method returns them e.g. `tid`.
```
lmuntaner: Update the comment to say it returns `["tid"]`, no?

src/internet_identity/src/openid.rs
```diff
@@ -190,69 +213,75 @@ where
...
         }
     }
 
-    // Append any remaining text
-    result.push_str(remaining);
+    keys
+}
+
+/// Either get all claims for the given keys or nothing.
+pub fn get_claims(claims_bytes: &[u8], keys: Vec<String>) -> Vec<(String, String)> {
```
lmuntaner: Shall we call it `get_all_claims` then?

lmuntaner (commented):
minor comments

src/internet_identity/src/openid.rs
sea-snake: Good idea, I'll do that.

src/internet_identity/src/openid.rs
sea-snake: Good idea

src/internet_identity/src/openid.rs
sea-snake: Ah yes.

lmuntaner (approved):
Thanks!

PR #3326, opened 2025-08-29
Authorize after registration

lmuntaner:
# Motivation

Users should not be required to select an account after registration.

The accounts page was being shown if the user created a new identity from the identity switcher.

# Changes

* Added a call to `triggerDropWaveAnimation` in the `authorize` handler of `+page.svelte` to provide visual feedback when authorization is triggered.
* Changed the authorization process in `+layout.svelte` to call `authorizationStore.authorize` directly after triggering the drop wave animation, and removed the call to `preloadAccounts` and `gotoAccounts`, simplifying the flow.

# Tests

Tested locally by registering from identity switcher and also with clean local storage.

### DIFF
--- a/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
+++ b/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
@@ -82,7 +82,7 @@
   const handleContinueCreatePasskey = async (): Promise<void | "cancelled"> => {
     isAuthenticating = true;
     try {
-      onSignIn(await authFlow.createPasskey());
+      onSignUp(await authFlow.createPasskey());
     } catch (error) {
       if (isWebAuthnCancelError(error)) {
         return "cancelled";
--- a/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
+++ b/src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
@@ -82,7 +82,7 @@
   const handleContinueCreatePasskey = async (): Promise<void | "cancelled"> => {
     isAuthenticating = true;
     try {
-      onSignUp(await authFlow.createPasskey());
+      onSignIn(await authFlow.createPasskey());
     } catch (error) {
       if (isWebAuthnCancelError(error)) {
         return "cancelled";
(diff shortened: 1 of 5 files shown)


src/frontend/src/routes/(new-styling)/authorize/(panel)/+page.svelte
```diff
@@ -37,6 +38,7 @@
       identityNumber,
       accountNumber: undefined,
     });
+    triggerDropWaveAnimation();
```
lmuntaner, line 41: This was missing, we don't want to leave users without our flair!

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -68,12 +68,9 @@
       closable: false,
     });
     lastUsedIdentitiesStore.selectIdentity(identityNumber);
-    preloadAccounts();
     triggerDropWaveAnimation();
-
-    await gotoAccounts();
-
     isAuthDialogOpen = false;
+    await authorizationStore.authorize(undefined, 4000);
```
lmuntaner, line 68: This was the issue...

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
```diff
@@ -82,7 +82,7 @@
   const handleContinueCreatePasskey = async (): Promise<void | "cancelled"> => {
     isAuthenticating = true;
     try {
-      onSignUp(await authFlow.createPasskey());
+      onSignIn(await authFlow.createPasskey());
```
lmuntaner: Probably it came because we saw that the accounts were not being used with continue and we expected them, but the problem was that it was calling the wrong prop

src/frontend/tests/e2e-playwright/authorize/continue.spec.ts
```diff
@@ -34,6 +35,16 @@ test("Authorize by switching to another identity", async ({ page }) => {
   expect(principal).not.toBe(otherPrincipal);
 });
 
+test("Authorize by creating a new identity", async ({ page }) => {
```
lmuntaner, line 38: One more test.

PR #3329, opened 2025-08-29
feat: Extend memory layout to support `lookup_application_with_origin` keyed by full SHA-256 sums

aterga:
# Motivation

`lookup_application_with_origin` is one of the stable lookup maps in the II backend canister. This map has keys that are 8-byte prefixes of a full sha-256 sum. This increases the chance that 8-byte hash collisions could be exploited to trigger surprising behavior. To reduce the risk to effectively zero, we would like to use the full sha256 sum as a key in these lookups. 

To migrate existing data and start relying on proper collision resilience of SHA-256, we need another map (keyed by full SHA-256 sum). That is being added in this PR. The new map is being updated along with the old one, but the source of truth will remain unchanged until the migration is complete.

# Changes

* Added a new stable memory index (19) for the new map.
* Extended `internet_identity::storage::Storage` with the new map keyed by full SHA-256 sum.
* Extended `internet_identity::storage::Storage::lookup_or_insert_application_number_with_origin` to insert new elements into the new map along side inserting them into the old one. 

# Tests

* Regression test was added (`should_handle_collision_prone_origins_with_sha256`)
* Source-of-truth test that demonstrates that the old map is still being read
* These and other unit tests were added to src/internet_identity/src/storage/tests.rs

| [Next PR](https://github.com/dfinity/internet-identity/pull/3337) >


src/internet_identity/src/storage.rs
```diff
@@ -289,9 +297,15 @@ pub struct Storage<M: Memory> {
     lookup_anchor_with_device_credential_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
     lookup_anchor_with_device_credential_memory:
         StableBTreeMap<StorableCredentialId, StorableAnchorNumber, ManagedMemory<M>>,
+
     lookup_application_with_origin_memory_wrapper: MemoryWrapper<ManagedMemory<M>>,
-    lookup_application_with_origin_memory:
+
+    // This field is deprecated in favor of `lookup_application_with_origin_memory`.
+    // TODO[ID-354]: Remove this after the production data is migrated to the new map.
+    lookup_application_with_origin_memory_old:
```
lmuntaner, line 305: Can we just change the name like this and be backwards compatible? Do I understand correctly, that this will read the part of the current memory, but it will be set in another field?

src/internet_identity/src/storage.rs
```diff
@@ -710,16 +735,27 @@ impl<M: Memory + Clone> Storage<M> {
...
 
-            self.lookup_application_with_origin_memory
+            // Update the source of truth.
+            // TODO[ID_353]: Remove this line.
+            self.lookup_application_with_origin_memory_old
                 .insert(origin_hash, new_number);
 
+            // Update the new map (will eventually become the source of truth).
+            self.lookup_application_with_origin_memory
+                .insert(origin_sha256, new_number);
```
lmuntaner, line 757: The new origin keeps the same identity numbers and if all goes well, after the migration checking the length or one will match the length of the other.

Should we add a check here with a log in case that's not the case? That could help us find a possible bug before fully moving to the new field.

src/internet_identity/src/storage/tests.rs
```diff
@@ -376,3 +376,336 @@ fn sample_persistent_state() -> PersistentState {
...
+        assert_application_lookup(&storage, &origin2, Some(1), Some(1));
+        assert_application_lookup(&storage, &origin3, Some(2), Some(2));
+    }
+
+    #[test]
+    fn should_handle_collision_prone_origins_with_sha256() {
+        let mut storage = Storage::new((10, 20), VectorMemory::default());
+
+        // These origins have hash prefix collisions with 8-byte hash but not with full SHA-256.
+        // Example taken from https://github.com/yugt/sha256-prefix-collision
```
lmuntaner, line 486: Cool!

lmuntaner (commented):
Thanks a lot! Looks good to me!

I added some minor comments.

Should we add a test in the "tests/integration/upgrade.rs" that a user can for example create an account and they are kept across upgrades?

src/internet_identity/src/storage/storable/application.rs
```diff
@@ -29,11 +32,65 @@ impl Storable for StorableApplication {
...
+    bounded[..copy_len].copy_from_slice(&slice[..copy_len]);
+    bounded
+}
+
+impl StorableOriginSha256 {
+    pub fn from_origin(origin: &FrontendHostname) -> Self {
+        let mut hasher = Sha256::new();
+        hasher.update(origin.as_bytes());
+        let sha256sum = hasher.finalize();
+        let hash = slice_to_bounded_32(&sha256sum);
```
sea-snake, line 60: Is this really needed? Since Sha256 by definition is always 256 bits which is always 32 bytes.

src/internet_identity/src/storage.rs
```diff
@@ -163,13 +164,15 @@ const REGISTRATION_CURRENT_RATE_MEMORY_INDEX: u8 = 6u8;
 const LOOKUP_ANCHOR_WITH_DEVICE_CREDENTIAL_MEMORY_INDEX: u8 = 9u8;
 const STABLE_ACCOUNT_MEMORY_INDEX: u8 = 10u8;
 const STABLE_APPLICATION_MEMORY_INDEX: u8 = 11u8;
-const LOOKUP_APPLICATION_WITH_ORIGIN_MEMORY_INDEX: u8 = 12u8;
+// TODO[ID-354]: Abandon this after the production data is migrated to the new map.
+const OLD_LOOKUP_APPLICATION_WITH_ORIGIN_MEMORY_INDEX: u8 = 12u8;
```
sea-snake: Nit: Let's either prefix or suffix everything consistently with `old` and `new`.

src/internet_identity/src/storage.rs
sea-snake: This field name should only be a pointer to `StableBTreeMap` instance that's initialized with a given `ManagedMemory` that's created with a given memory id.

TL;DR Yes, renaming this should be fine, only the memory id matters.

src/internet_identity/src/storage/storable/application.rs
```diff
@@ -29,11 +32,65 @@ impl Storable for StorableApplication {
...
+    }
+}
+
+impl Storable for StorableOriginSha256 {
+    fn to_bytes(&self) -> Cow<'_, [u8]> {
+        Cow::Owned(self.hash.to_vec())
+    }
+
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Self {
+        let hash = slice_to_bounded_32(bytes.as_ref());
```
sea-snake, line 71: I don't think this is needed either since below we mark this storable as fixed 32 bytes already.

src/internet_identity/src/storage/storable/application.rs
aterga: It's generally better to avoid explicit unwraps in production canisters. Admittedly, the benefits here are marginal. But since the type safe code is just a few lines, I suggest keeping it here to avoid having to debate if the 3-rd party library can always be trusted.

src/internet_identity/src/storage/storable/application.rs
aterga: Ideally, each function would be panic-safe in isolation (one shouldn't need to reason about other functions to prove it). I believe this is archived here.

src/internet_identity/src/storage.rs
aterga: +1

src/internet_identity/src/storage/tests.rs
```diff
@@ -376,3 +376,336 @@ fn sample_persistent_state() -> PersistentState {
...
+        // Should get the same application number, as we do not yet detect collisions (since the
+        // source of truth is still the old lookup map).
+        // TODO[ID-352]: This assertion should be changed to `assert_ne`.
+        assert_eq!(app_num1, app_num2);
+        assert_eq!(app_num1, 0);
+
+        // Both should be stored correctly in new SHA-256 map
+
+        assert_application_lookup(&storage, &origin1, Some(0), Some(0));
+        assert_application_lookup(&storage, &origin2, None, Some(0));
```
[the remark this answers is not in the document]
aterga: The insertion happens for both maps, however, it only happens if the old map does not already have the short hash as the key. Which it does after the first insertion due to the collision exhibited by this test. So there is really only one insertion, for origin1, and the reason an entry is found for origin2 is the very issue being addressed by this code (but not fully; this is just preparation for the migration that would resolve the issue). 

In other words, this test exhibits an issue we currently have, and it will be turned into a regression test in the near future.

src/internet_identity/src/storage/storable/application.rs
sea-snake: Wouldn't we want to catch that (panic) in CI instead of bounding it to an upper limit and possibly not catching and releasing it?

src/internet_identity/src/storage/storable/application.rs
sea-snake: Wouldn't we want to catch that (panic) in CI instead of bounding it to an upper limit and possibly not catching and releasing it?

aterga: Good question. 

For the CI to catch this, we would need to have test cases in which (1) this function's inputs are exactly the ones triggering the issue and (2) the tests's assertions containing direct or indirect postconditions to specify the correct behavior. So subtle bugs have a low chance to being caught by CI. Bottom line, we cannot guarantee that all bugs will be caught.

Therefore, we must choose how the production code behaves in a buggy situation. There are three options:
1. panic (typically the worst from security standpoint, especially for canister code)
2. propagate errors to the caller (typically the best from security standpoint; requires more complex code throughout the call stack)
3. fallback (a simpler way to avoid panics when justified)

I think in this case 3 is a good tradeoff for security and code complexity. To justify why it's sufficient, the producers and the consumers of our lookup indices will both go through the following route:

origin string -> key -> lookup value

where the key is produced by this (deterministic) function. So in this worst case, if this function has a subtle input that would produce fewer than 32 meaningful bytes from the SHA library (e.g., due to a bug in that library), our function would still produce the same 32-bytes outputs each time its called. So we still have to trust the library code for collision resistance, but not for the output format. 

Do you agree with my reasoning?

src/internet_identity/src/storage.rs
aterga: Note that before the migration, the lengths are not supposed to be equal for these two maps. So I'm not sure what useful property we can check here.

We will have a log to indicate that the lengths are equal after the migration, before switching the source of truth from the old to the new map.

src/internet_identity/src/storage/storable/application.rs
aterga: Let's continue the discussion [here](https://github.com/dfinity/internet-identity/pull/3329#discussion_r2313198477)

src/internet_identity/src/storage.rs
lmuntaner: I was thinking on confirming that the lengths are equal during the time we have both maps. But maybe it's too obvious that it's added to both...

lmuntaner:
@aterga I remove it from the queue because I see that there is an open discussion still with @sea-snake .

I approved it to mean that it was good to go on my side. But I'd rather wait to merge it until all discussions are finalized, even if we are just waiting for a simple "Yes, that makes sense".

lmuntaner:
Maybe I shouldn't have approved it? @aterga @sea-snake ?

aterga:
> @aterga I remove it from the queue because I see that there is an open discussion still with @sea-snake .
> 
> I approved it to mean that it was good to go on my side. But I'd rather wait to merge it until all discussions are finalized, even if we are just waiting for a simple "Yes, that makes sense".

Agree. Note that GitHub does not automatically merge until all discussions are resolved.

aterga:
> Maybe I shouldn't have approved it? @aterga @sea-snake ?

Why not?

src/internet_identity/src/storage.rs
aterga: Since you've approved the PR, I assume you don't have any follow ups for this thread. Otherwise, let me know and I'll address them in a follow-up PR (or this one, unless it's already merged)

src/internet_identity/src/storage/storable/application.rs
sea-snake: Reasoning makes sense, as you've mentioned there's two issues that could appear in the external lib:
1. incorrect hash
2. incorrect length

Since we can't really catch the first either way, it indeed makes sense to fallback for the second as you've mentioned so neither will result in an unexpected trap on production.

src/internet_identity/src/storage.rs
lmuntaner: That's correct, I don't have. But I saw that there was still an open discussion with @sea-snake

src/internet_identity/src/storage/storable/application.rs
aterga: Not taking any actions for the same reasons as in the linked thread.

PR #3330, opened 2025-08-29
Implement test OpenID provider for e2e tests

sea-snake:
Implement test OpenID provider for e2e tests and fix provider name in access methods list.

# Changes

- Added `test_openid_provider` in src with minimal implementation required to run an OpenID provider for e2e tests.
- Fix provider name in access methods list, not it correctly shows the name instead of Unknown when both email and name are missing.

# Tests

- Manually tested the OpenID provider to confirm it's working as expected (see video).
- Manually verified that the provider name is now shown correct in the access method list, additionally updated the unit tests.

https://github.com/user-attachments/assets/a7d8ce9f-a9fb-4890-9b19-32818bb5f4c5

### DIFF
--- a/src/frontend/src/lib/utils/accessMethods.test.ts
+++ b/src/frontend/src/lib/utils/accessMethods.test.ts
@@ -247,8 +247,17 @@ describe("getOpenIdTitles", () => {
     });
   });
 
-  it("returns Unknown account when neither name nor email is present", () => {
+  it("returns AcmeID account when neither name nor email is present", () => {
     const cred = makeOpenIdCredential("https://issuer.acme", {});
+    const res = getOpenIdTitles(cred);
+    expect(res).toEqual({
+      title: { ellipsis: false, text: "AcmeID Account" },
+      subtitle: undefined,
+    });
+  });
+
+  it("returns Unknown account when neither name nor email is present and config not found", () => {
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/utils/accessMethods.test.ts
```diff
@@ -247,9 +247,18 @@ describe("getOpenIdTitles", () => {
...
+  it("returns AcmeID account when neither name nor email is present", () => {
     const cred = makeOpenIdCredential("https://issuer.acme", {});
     const res = getOpenIdTitles(cred);
+    expect(res).toEqual({
+      title: { ellipsis: false, text: "AcmeID Account" },
+      subtitle: undefined,
+    });
+  });
+
+  it("returns Unknown account when neither name nor email is present and config not found", () => {
```
lmuntaner, line 259: `"returns Unknown account when config not found"`

lmuntaner (commented):
Nice!

lmuntaner (approved):

src/frontend/src/lib/utils/accessMethods.test.ts
sea-snake: That's already in a test above, this test is particularly the combination of all these things.

PR #3332, opened 2025-08-29
Only allow one account per provider.

sea-snake:
Only allow one account per provider, make sure to check with `issuerMatches` function.

# Changes

- User `issuerMatches` function in `hasCredential`.

# Tests

- Verified I can only add a single Microsoft account.

### DIFF
--- a/src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
+++ b/src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
@@ -6,6 +6,7 @@
   import Button from "$lib/components/ui/Button.svelte";
   import GoogleIcon from "$lib/components/icons/GoogleIcon.svelte";
   import Tooltip from "$lib/components/ui/Tooltip.svelte";
+  import { issuerMatches } from "$lib/utils/openID";
   import { nonNullish } from "@dfinity/utils";
   import { canisterConfig } from "$lib/globals";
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
```diff
@@ -60,9 +61,10 @@
     canisterConfig.openid_google?.[0]?.[0] && !$ENABLE_GENERIC_OPEN_ID;
   const openIdProviders = canisterConfig.openid_configs?.[0] ?? [];
 
-  const hasCredential = (issuer: string): boolean => {
-    return openIdCredentials.some((cred) => cred.iss === issuer);
-  };
+  const hasCredential = (configIssuer: string): boolean =>
+    openIdCredentials.some((cred) =>
+      issuerMatches(configIssuer, cred.iss, cred.metadata ?? []),
```
lmuntaner: Can `cred.metadata` be undefined? I thought it's always defined.

lmuntaner (approved):
One comment but already approving

src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
sea-snake: Ah yes, I saw you writing ?? [] in various places so I assumed, but this is the credential, not last used data indeed.

PR #3334, opened 2025-09-01
Fix popup blocked by Safari when linking openID accounts

lmuntaner:
# Motivation

Safari blocked the popup to link a Google account from the dashboard during testing.

We saw that we were doing an async call to get the nonce and salt for the account verification.

To avoid async calls after user interaction, I added the `salt` and `nonce` to the authenticated store similar to what we do for the sessionStore.

# Changes

* The `authenticationStore.set` method is now asynchronous and always generates and stores a `nonce` and `salt` using `createAnonymousNonce`, ensuring these values are present in the authentication context. The method signature and implementation have been updated accordingly. (`src/frontend/src/lib/stores/authentication.store.ts`)
* All usages of `authenticationStore.set` in authentication-related flows (`AuthFlow`, `AuthLastUsedFlow`, `MigrationFlow`, `RegisterAccessMethodFlow`) have been updated to use `await`, reflecting the new asynchronous nature of the method. This ensures that the authentication state is fully updated before proceeding.
* The `AddAccessMethodFlow` and `linkGoogleAccount` flows now retrieve `nonce` and `salt` directly from the authenticated store instead of generating them again, ensuring consistency and reducing redundant calls.

# Tests

Tested locally that the salt and nonce were set when linking an openID account.

### DIFF
--- a/src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
@@ -47,13 +47,16 @@ export class AddAccessMethodFlow {
   linkOpenIdAccount = async (
     config: OpenIdConfig,
   ): Promise<OpenIdCredential> => {
-    const { identity, actor, identityNumber } = get(authenticatedStore);
+    const { actor, identityNumber, identity, salt, nonce } =
+      get(authenticatedStore);
+
+    const openIDNonce: { nonce: string; salt: Uint8Array } =
+      nonNullish(salt) && nonNullish(nonce)
+        ? { nonce, salt }
+        : await createAnonymousNonce(identity.getPrincipal());
 
     try {
       this.#isSystemOverlayVisible = true;
-      const { nonce, salt } = await createAnonymousNonce(
-        identity.getPrincipal(),
-      );
       const jwt = await requestJWT(
         {
           clientId: config.client_id,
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/stores/authentication.store.ts
```diff
@@ -57,6 +62,20 @@ export const authenticationStore: AuthenticationStore = {
         throw new Error("Not initialized");
       }
       initialized.agent.replaceIdentity(authenticated.identity);
+      // Create the OpenID nonce using the identity's principal
+      void createAnonymousNonce(authenticated.identity.getPrincipal()).then(
```
lmuntaner: At the beginning I didn't want to make this async, but the more I think about it, the more I'm leaning towards making `set` async and setting salt and nonce as non-optional in the `Authenticated` type.

Wdyt @sea-snake ?

sea-snake: Yes, make it async, that should be fine. Without making it async we'd might end up with flakiness in tests 😅

src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
```diff
@@ -62,7 +60,7 @@ export class AddAccessMethodFlow {
           authScope: config.auth_scope.join(" "),
         },
         {
-          nonce,
+          nonce: nonce,
```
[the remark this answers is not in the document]
lmuntaner: good point

src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
```diff
@@ -47,13 +47,11 @@ export class AddAccessMethodFlow {
   linkOpenIdAccount = async (
     config: OpenIdConfig,
   ): Promise<OpenIdCredential> => {
-    const { identity, actor, identityNumber } = get(authenticatedStore);
+    const { actor, identityNumber, salt, nonce } = get(authenticatedStore);
+    console.log(actor, identityNumber, salt, nonce);
```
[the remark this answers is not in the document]
lmuntaner: oopsie

src/frontend/src/lib/stores/authentication.store.ts
lmuntaner: Ah, good point about the tests!

PR #3335, opened 2025-09-01
Add sign up with OIDC e2e tests.

sea-snake:
Add sign up with OIDC e2e tests.

# Changes

- Run test OpenID provider in e2e CI
- Set `feature_flag_enable_generic_open_id_fe` to true in e2e CI
- Add `openid_configs` with test OpenID provider in e2e CI
- Change test OpenID provider issuer to `http://localhost:11105`.
- Exclude localhost from being mapped to vite server in playwright chrome browser config
   - This makes the test OpenID provider reachable at localhost:11105
   - Everything we intend to actually map has a tld e.g. id.ai, identity.ic0.app and nice-name.com
- Add accessibility label to OpenID buttons.
- Implement "Sign up with OpenID" and "Sign up with OpenID (no name)" e2e tests.

# Tests

- Verified that the additional e2e pass and visually go through the correct flow, showing the correct screens.


.github/workflows/canister-tests.yml
```diff
@@ -632,7 +639,7 @@ jobs:
...
+          dfx canister install internet_identity --wasm internet_identity_test.wasm.gz --argument "(opt record { captcha_config = opt record { max_unsolved_captchas= 50:nat64; captcha_trigger = variant {Static = variant { CaptchaDisabled }}}; related_origins = opt vec { \"https://id.ai\"; \"https://identity.ic0.app\" }; new_flow_origins = opt vec { \"https://id.ai\" }; dummy_auth = opt opt record { prompt_for_index = true }; feature_flag_continue_from_another_device = opt true; feature_flag_enable_generic_open_id_fe = opt true; openid_configs = opt vec { record { name = \"Test OpenID\"; logo = \"<svg viewBox=\\\"0 0 24 24\\\"><path d=\\\"m14 2.8-2.6 1.8v16.6l2.6-1.4zm-3.3 5.4C1.9 9 2 14.7 2 14.7c0 5.6 8.7 6.5 8.7 6.5v-1.9c-6.3-1-5.5-4.5-5.5-4.5.3-4 5.5-4.3 5.5-4.3Zm4 0v2.1s1.6 0 3 1.2l-1.5.6 5.8 1.4V9l-2 1.1s-1.7-1.7-5.3-1.9z\\\" style=\\\"fill: currentColor;\\\"></path></svg>\"; issuer = \"http://localhost:11105\"; client_id = \"internet_identity\"; jwks_uri = \"http://localhost:11105/jwks\"; auth_uri = \"http://localhost:11105/auth\"; auth_scope = vec { \"openid\"; \"profile\"; \"email\" }; fedcm_uri = opt \"\"; } } })"
```
lmuntaner: Could we put `http://localhost:11105` in a variable with a comment? I would even create a variable for each of the configs to better understand where they come from.

playwright.config.ts
```diff
@@ -42,7 +42,7 @@ export default defineConfig({
         launchOptions: {
           args: [
             "--ignore-certificate-errors",
-            "--host-resolver-rules=MAP * localhost:5173",
+            "--host-resolver-rules=MAP * localhost:5173, EXCLUDE localhost",
```
lmuntaner, line 45: Do we still need `localhost:5173` then?

src/frontend/tests/e2e-playwright/index.spec.ts
```diff
@@ -125,6 +125,99 @@ test.describe("First visit", () => {
...
+
+    // Authenticate and authorize with OpenID
+    const authPage = await pagePromise;
+    await expect(
+      authPage.getByRole("heading", {
+        name: "Sign-in",
+      }),
+    ).toBeVisible();
+    await authPage.getByPlaceholder("Enter any login").fill(userId);
+    await authPage.getByPlaceholder("and password").fill("secret");
```
lmuntaner: Where does the password come from? maybe worth a comment?

lmuntaner (commented):
Amazing!!

Can you add one E2E tests in `authorize/index.spec.ts`? To make sure it also works from an application.

Thanks!

src/test_openid_provider/index.js
```diff
@@ -4,7 +4,7 @@ import express from "express";
 const app = express();
 const port = 11105; // "OpenID" (O = 111, I = 105)
 const accountClaims = new Map();
-const provider = new oidc.Provider("https://oidc.local", {
+const provider = new oidc.Provider("http://localhost:11105", {
```
aterga, line 7: Why was this not needed before, and why is it needed now?

src/test_openid_provider/index.js
```diff
@@ -4,7 +4,7 @@ import express from "express";
 const app = express();
 const port = 11105; // "OpenID" (O = 111, I = 105)
```
aterga, line 5: Not added in this PR, so no changes required (NCR)

The comment is confusing, what does this mapping mean (O = 111, I = 105)? How does that relate to the port?

src/test_openid_provider/index.js
sea-snake: The issuer value doesn't actually matter. It's changed here for consistency, `https://oidc.local` was a made up value in the previous PR but it's not very consistent and potentially confusing compared to the other values since it comes out of nowhere.

src/test_openid_provider/index.js
sea-snake: Ah, the port was something I've randomly picked, the comment just indicates why I picked 11105 for whoever is interested in the thoughts behind it. (O as in Open, I as in Id).

src/frontend/tests/e2e-playwright/index.spec.ts
sea-snake: I'll change the password to `any-password-works` since it doesn't actually have a password check.

playwright.config.ts
sea-snake: Yes `MAP * localhost:5173` points any domain (not localhost after this EXCLUDE is added) to localhost:5173 e.g. `nice-name.com` that is then proxied to the respective canister by vite.

src/frontend/tests/e2e-playwright/index.spec.ts
```diff
@@ -125,6 +125,99 @@ test.describe("First visit", () => {
...
+    const userId = crypto.randomUUID();
+
+    // Set name claim
+    await fetch(`http://localhost:11105/account/${userId}/claims`, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ name: DEFAULT_USER_NAME }),
+    });
```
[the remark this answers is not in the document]
sea-snake: We're not testing the test openid provider in this e2e test.

playwright.config.ts
lmuntaner: Noted, thanks!

src/frontend/tests/e2e-playwright/index.spec.ts
lmuntaner: Thanks!

sea-snake:
> Can you add one E2E tests in `authorize/index.spec.ts`? To make sure it also works from an application.

Done

.github/workflows/canister-tests.yml
sea-snake: Moved them all into an env variable and added a comment to explain where they come from.

lmuntaner (approved):

src/test_openid_provider/index.js
aterga: Then how about

```suggestion
const port = 111105; // "OpenID" (O = 111, I = 105)
```

aterga:
LGTM, and no changes required from my side (up to you if you want to apply my suggestion)

Note that I didn't carefully review each of the tests.

src/test_openid_provider/index.js
sea-snake: I wanted to do that but the port number would be too large (max is 65535).

PR #3337, opened 2025-09-01
chore: Implement data migration for `lookup_application_with_origin_map`

aterga:
# Motivation

This PR continues the work towards migrating the `lookup_application_with_origin` map's keys from 8-byte prefixes of a full sha-256 sum.

< [Previous PR](https://github.com/dfinity/internet-identity/pull/3329) |

# Changes

* Added a one-off call at the end of II's `post_upgrade`.

# Tests

## Integration
* `upgrade_and_rollback_with_realistic_data_migration` - Upgrading the II canister succeeds, migrating 2,000 application-specific user accounts. This test passed locally in 319.35s; if that's not acceptable for the CI, we can reduce the number of accounts in this test to 10, and re-run it locally to gain confidence in the migration.

## Unit
* `should_skip_migration_if_new_map_not_empty` - When the new map already contains entries, the migration leaves both maps unchanged.
* `should_migrate_single_entry_successfully` - Tests successful migration of one application from old hash map to new SHA-256 map, verifying correct return values and final state.
* `should_migrate_multiple_entries_successfully` - Tests successful migration of multiple applications, ensuring all entries are properly transferred with correct mappings.
* `should_handle_missing_application_gracefully` - Tests behavior when old map references non-existent applications, verifying missing apps are reported in return values and skipped during migration.
* `should_handle_empty_old_map` - Tests migration behavior with empty maps, ensuring function returns empty collections and maps remain empty.
* `should_handle_hash_collisions_correctly` - Tests collision detection where multiple applications exist in stable storage but only one is in the old map due to hash collision, verifying collision recovery mechanism.

### DIFF
--- a/src/internet_identity/src/migrations/lookup_application_with_origin.rs
+++ b/src/internet_identity/src/migrations/lookup_application_with_origin.rs
@@ -40,32 +40,10 @@ impl<M: Memory + Clone> Storage<M> {
                 self.lookup_application_with_origin_memory.len(),
             ));
             log_entries.push(format!(
-                "INFO: Migrated origins (origin:new hash):\n{}",
-                migrated_entries
-                    .iter()
-                    .map(|(k, v)| format!("{}:{}", k, v))
-                    .collect::<Vec<_>>()
-                    .join("\n")
-            ));
-            log_entries.push(format!(
-                "INFO: lookup_application_with_origin_memory_old:\n{}",
-                self.lookup_application_with_origin_memory_old
-                    .iter()
-                    .map(|(k, v)| format!("{}:{}", k, v))
-                    .collect::<Vec<_>>()
-                    .join("\n")
+                "INFO: Successfully migrated {} origins.",
+                migrated_entries.len()
             ));
         }
-
-        // Log all migrated entries (compactly)
-        log_entries.push(format!(
-            "INFO: lookup_application_with_origin_memory:\n{}",
-            self.lookup_application_with_origin_memory
-                .iter()
-                .map(|(k, v)| format!("{}:{}", k, v))
-                .collect::<Vec<_>>()
-                .join("\n")
-        ));
 
         if !app_collisions.is_empty() {
             log_entries.push(format!(
@@ -83,9 +61,10 @@ impl<M: Memory + Clone> Storage<M> {
         }
 
         if app_collisions.is_empty() && missing_apps.is_empty() {
-            log_entries.push(format!(
+            log_entries.push(
                 "INFO: Migration completed SUCCESSFULLY without app collisions or missing apps."
-            ));
+                    .to_string(),
+            );
         }
 
         for log_entry in &log_entries {
--- a/src/internet_identity/src/migrations/lookup_application_with_origin.rs
+++ b/src/internet_identity/src/migrations/lookup_application_with_origin.rs
@@ -61,10 +61,10 @@ impl<M: Memory + Clone> Storage<M> {
         }
 
         if app_collisions.is_empty() && missing_apps.is_empty() {
-            log_entries.push(
-                "INFO: Migration completed SUCCESSFULLY without app collisions or missing apps."
-                    .to_string(),
-            );
+            log_entries.push(format!(
+                "INFO: SUCCESSFULLY migrated {} entries without app collisions or missing apps.",
+                migrated_entries.len()
+            ));
         }
 
         for log_entry in &log_entries {
@@ -572,13 +572,7 @@ mod tests {
             storage.log_after_migration(&migrated_entries, &missing_apps, &app_collisions);
 
         let expected_log_entries = vec![
-            format!(
-                "INFO: lookup_application_with_origin_memory:\n\
-                 {}:102\n\
-                 {}:101",
-                new_key2, new_key1
-            ),
-            "INFO: Migration completed SUCCESSFULLY without app collisions or missing apps."
+            "INFO: SUCCESSFULLY migrated 2 entries without app collisions or missing apps."
                 .to_string(),
         ];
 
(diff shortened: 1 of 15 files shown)


src/internet_identity/src/migrations/lookup_application_with_origin.rs
```diff
@@ -0,0 +1,548 @@
...
+    /// being that `lookup_application_with_origin_memory` is populated with entries corresponding
+    /// to the existing applications).
+    ///
+    /// In the unlikely event that there already was a collision, this function populates
+    /// the abovementioned map also by directly scanning `stable_application_memory`.
+    ///
+    /// The return type is an optional tuple containing the migrated entries, missing apps, and app
+    /// collisions (where None indicates that the migration had already run before).
+    #[allow(clippy::type_complexity)]
+    pub fn maybe_apply_lookup_application_with_origin_map_data_migration(
```
lmuntaner, line 88: What's the limit in post_upgrade instructions and how many applications can we migrate with this function?

src/internet_identity/src/migrations/lookup_application_with_origin.rs
```diff
@@ -0,0 +1,548 @@
...
+                .is_some()
+            {
+                // No work to be done, key already exists in the new map
+                continue;
+            }
+
+            self.lookup_application_with_origin_memory
+                .insert(new_key.clone(), add_id);
+
+            app_collisions.insert(new_key, add_id);
```
lmuntaner: So, a collision would be one app that it's in the applications map but not in the origin map. Which means that it was first added there and then the collision substituted the entry in the origin map, right?

src/internet_identity/src/storage/storable/application.rs
```diff
@@ -34,7 +34,7 @@ impl Storable for StorableApplication {
 
 #[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct StorableOriginSha256 {
-    hash: [u8; 32],
+    pub(crate) hash: [u8; 32],
```
lmuntaner: Why is this needed?

src/internet_identity/src/storage/storable/application.rs
```diff
@@ -82,7 +82,7 @@ impl Storable for StorableOriginSha256 {
 // TODO[ID_353]: Remove this type.
 #[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct StorableOriginHash {
-    hash: [u8; 8],
+    pub(crate) hash: [u8; 8],
```
lmuntaner: Why is this needed?

lmuntaner (commented):
Looks great! Just a few questions before approving to better understand the changes.

src/internet_identity/src/migrations/lookup_application_with_origin.rs
aterga: Yes

src/internet_identity/src/storage/storable/application.rs
aterga: Thanks - this isn't needed, will remove

src/internet_identity/src/storage/storable/application.rs
aterga: Thanks - this isn't needed, will remove

src/internet_identity/src/migrations/lookup_application_with_origin.rs
aterga: Good question

We currently [have](https://rdmx6-jaaaa-aaaaa-aaadq-cai.raw.icp0.io/metrics) 766 apps. 5 days ago we had 649. So that's ~25 new apps per day. Assuming we roll this out in the next 2 weeks (14 days), we should expect 350 new apps, totaling 1,116. Rounding up, let's assume we're going to have 2,000 apps in the worst case.

With the help of AI (GPT-4.1 and independently Claude Sonnet 4) I got estimations for the migration function to take **below 5,000,000 instructions**. The most expensive operation within those loops being SHA-256.

Comparing that to the entire upgrade instruction limit of 300 Billion, that seems like a negligible fraction.

**Bottom line.** If I'm mistaken in my estimations, the post_upgrade function will trap and the upgrade will roll back; this would be a recoverable issue.

src/internet_identity/src/migrations/lookup_application_with_origin.rs
```diff
@@ -0,0 +1,608 @@
...
+
+        // Assert return values
+        let (_, expected_new_key1) = create_origin_keys(&origin1);
+        let (_, expected_new_key2) = create_origin_keys(&origin2);
+        assert_eq!(
+            migrated_entries,
+            BTreeMap::from([(origin1, expected_new_key1)])
+        );
+        assert!(missing_apps.is_empty());
+        assert_eq!(app_collisions, BTreeMap::from([(expected_new_key2, 401)]));
```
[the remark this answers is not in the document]
aterga: It's fine for the test to be fragile, as this code is implementing a data migration that is a temporary code path to begin with.

src/internet_identity/src/migrations/lookup_application_with_origin.rs
lmuntaner: it will roll back? I thought that if it trapped we were stuck forever, or is that the pre_upgrade?

lmuntaner (approved):
Thanks!

src/internet_identity/src/migrations/lookup_application_with_origin.rs
aterga: For context, a trap in the pre_upgrade prevents any subsequent upgrade (the solution there is to skip pre-upgrade, but that may involve data loss).

A trap in post-upgrade simply fails the current upgrade, so a subsequent upgrade to a different Wasm may still succeed (no risk of data loss).

aterga: Additionally, I've added a pocket-ic test, `upgrade_and_rollback_with_realistic_data_migration`, that demonstrates that the data migration with 2,000 accounts (each for a different app/origin) succeeds.

lmuntaner: Great, thanks!

lmuntaner (approved):
Looking great! Thanks!

PR #3338, opened 2025-09-02
Dialog Centralisation POC

danblackadder:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/dialogs/AllDialogs.svelte
+++ b/src/frontend/src/lib/components/dialogs/AllDialogs.svelte
@@ -1,0 +1,21 @@
+<script lang="ts">
+  import { onDestroy } from "svelte";
+  import { derived } from "svelte/store";
+  import TestDialog from "./TestDialog.svelte";
+  import { DialogId, dialogsStore } from "$lib/state/dialogs";
+  import TestDialog2 from "./TestDialog2.svelte";
+
+  const dialogs = derived(dialogsStore, ($store) => $store);
+
+  let activeDialogs: Partial<Record<DialogId, { isOpen: boolean }>> = {};
+
+  const unsubscribe = dialogs.subscribe((d) => (activeDialogs = d));
+  onDestroy(unsubscribe);
+</script>
+
+{#if activeDialogs[DialogId.Test]?.isOpen}
+  <TestDialog />
+{/if}
+{#if activeDialogs[DialogId.Test2]?.isOpen}
+  <TestDialog2 />
+{/if}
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/components/dialogs/AllDialogs.svelte
```diff
@@ -0,0 +1,21 @@
...
+  const dialogs = derived(dialogsStore, ($store) => $store);
+
+  let activeDialogs: Partial<Record<DialogId, { isOpen: boolean }>> = {};
+
+  const unsubscribe = dialogs.subscribe((d) => (activeDialogs = d));
+  onDestroy(unsubscribe);
+</script>
+
+{#if activeDialogs[DialogId.Test]?.isOpen}
+  <TestDialog />
```
lmuntaner, line 17: How do you pass the props type safe?

lmuntaner: can you also bind values like this?

danblackadder: props are passed through using the DialogDataMap

```

export interface DialogDataMap {
  [DialogId.Test]: { test: string };
  [DialogId.Test2]: { callback: () => void };
}
```

This could be cleaner, this is just for the POC

Effectively when you call `openDialog(ModalId, { params })` it checks against the DialogDataMap as to what types are expected.
image

The POC implementation was thrown together, so its not currently bindable, however its definitely possible.

danblackadder: The only question I would have on making a dialog state value binded, is would you expect the value to update from within the dialog from outside? If a dialog is updating, the arguable it should probably be from within the dialog handling itself?

lmuntaner: The idea of binding is tha it's two ways. I don't think we have any binding and it probalby doesn't make sense. But I was just thinking whether we'd lose some functionality like this.

danblackadder: I think we could proceed under the scope of "we add it when we need it", its definitely doable, and effectively we could update each Dialog individually and once one requires the binding then we implement it at that point?

lmuntaner: Yes, we can start implementing it for any new modal and then slowly move the others.

danblackadder: I'll close this PR for now to keep our open PR's clean, and keep it on the back burner for when I have free time in between more important tasks

PR #3339, opened 2025-09-02
Update generated types for ii and vc_issuer

danblackadder:
# Motivation

Auto generated types are not updated in the code base.

# Changes

- Updated internet_identity_types.d.ts
- Updated vc_issuer_types.ts

# Tests

Tested locally running linter.


danblackadder:
> And when yuo run `npm run generate` yuo get those types?
> 
> We just need to make sure that CI uses a consistent one. As I said, we have a check there to do that.

Yeah makes sense, its still happening, and yes I run `npm run generate` and still get these same changes.

We can do a quick screenshare when you get back to confirm things, as you say it should be picked up automatically, its strange that its not

danblackadder:
Closing this PR for now, issue still exists locally for me, but will assume it will eventually catch up and update automatically as stated

PR #3340, opened 2025-09-03
OpenID Integration tests

lmuntaner:
# Motivation

Keep openID functionality robust to future changes.

# Changes

This pull request extends OpenID authentication integration tests to support Microsoft Accounts in addition to Google, and refactors existing tests to use more specific test data helpers. The changes improve coverage for Microsoft OpenID flows, including linking, registration, and credential management, and add mock responses for Microsoft certificate endpoints.

* Added new tests for Microsoft account linking, registration, and credential management, including handling edge cases such as preventing the same Microsoft account from being linked to multiple identities and allowing linking accounts from different tenants. (`src/internet_identity/tests/integration/openid.rs`)
* Included Microsoft OpenID configuration in the test canister setup, enabling Microsoft authentication flows in integration tests. (`src/internet_identity/tests/integration/openid.rs`)
* Refactored existing Google account tests to use more specific helpers (`openid_google_test_data`, `second_openid_google_test_data`), replacing the generic `openid_test_data` and `second_openid_test_data` for clarity and maintainability. (`src/internet_identity/tests/integration/openid.rs`, `src/internet_identity/tests/integration/activity_stats/authn_methods.rs`)
* Added mock responses for Microsoft certificate endpoints to support JWT validation in Microsoft OpenID flows, in addition to the existing Google mock responses. (`src/internet_identity/tests/integration/openid.rs`)


src/internet_identity/tests/integration/activity_stats/authn_methods.rs
```diff
@@ -250,7 +250,7 @@ fn should_report_active_openid_authn_methods() {
 
     // OpenID test data and fetch Google certs (mock)
     let (jwt, salt, _claims, test_time, test_principal, _test_authn_method) =
-        openid::openid_test_data();
+        openid::openid_google_test_data();
```
aterga, line 253: Should there be a comment to explain why Google is picked for the test?

src/internet_identity/tests/integration/openid.rs
```diff
@@ -51,12 +52,164 @@ fn can_link_google_account() -> Result<(), RejectResponse> {
     Ok(())
 }
 
+/// Verifies that Microsoft Accounts can be added
+#[test]
+fn can_link_microsoft_account() -> Result<(), RejectResponse> {
```
aterga, line 61: What's the terminology here? 

Does "to link an account" mean the same as "to add an open-ai credential to an account"?

src/internet_identity/tests/integration/openid.rs
```diff
@@ -51,12 +52,164 @@ fn can_link_google_account() -> Result<(), RejectResponse> {
...
+        number_of_openid_credentials(&env, canister_id, test_principal, identity_number)?,
+        1
+    );
+
+    Ok(())
+}
+
+/// Verifies that the same Microsoft account cannot be linked to two different identities
+#[test]
+fn cannot_link_same_microsoft_account_two_identities() -> Result<(), RejectResponse> {
```
aterga:
```suggestion
fn cannot_link_same_microsoft_account_to_two_identities() -> Result<(), RejectResponse> {
```

src/internet_identity/tests/integration/openid.rs
```diff
@@ -51,12 +52,164 @@ fn can_link_google_account() -> Result<(), RejectResponse> {
...
+        &salt2,
+    )?;
+
+    assert_eq!(
+        result,
+        Err(OpenIdCredentialAddError::OpenIdCredentialAlreadyRegistered)
+    );
+    assert_eq!(
+        number_of_openid_credentials(&env, canister_id, test_principal2, identity_number2)?,
+        0
```
aterga, line 159: Should we also assert in the postcondition that `number_of_openid_credentials(&env, canister_id, test_principal, identity_number)?` is still 1?

Also, I would assert that in the precondition `number_of_openid_credentials(&env, canister_id, test_principal2, identity_number2)?` was zero to begin with.

So then the test spec would be, roughly

* Precondition: user_1 has 1, user_2 has 0
* Run code under test
* Postcondition: user_1 has 1, user_2 has 0

src/internet_identity/tests/integration/openid.rs
lmuntaner: Yes, it's the same.

src/internet_identity/tests/integration/activity_stats/authn_methods.rs
lmuntaner: Good point, I'll change the comment above

src/internet_identity/tests/integration/openid.rs
lmuntaner: Thanks!

src/internet_identity/tests/integration/openid.rs
lmuntaner: Good idea!

src/internet_identity/tests/integration/openid.rs
aterga: I would opt for a simpler terminology / not use synonyms where it's possible to avoid them (what if we later introduce the concept of linking two different II accounts)?

This is test code, so this discussion doesn't need to block the PR

src/internet_identity/tests/integration/openid.rs
```diff
@@ -51,12 +52,176 @@ fn can_link_google_account() -> Result<(), RejectResponse> {
...
+    let (jwt, salt, _claims, test_time, test_principal, test_authn_method) =
+        openid_microsoft_test_data();
+    let (jwt2, salt2, _claims2, test_time2, test_principal2, test_authn_method2) =
+        openid_microsoft_test_data_same_microsoft_different_user();
+
+    let identity_number = create_identity_with_authn_method(&env, canister_id, &test_authn_method);
+    let identity_number2 =
+        create_identity_with_authn_method(&env, canister_id, &test_authn_method2);
+
+    let time_to_advance = Duration::from_millis(test_time) - Duration::from_nanos(time(&env));
```
aterga: Nit. Seems like a common pattern, could be lifted to a helper function, e.g., `sync_test(&env, test_time)`

src/internet_identity/tests/integration/openid.rs
```diff
@@ -51,12 +52,176 @@ fn can_link_google_account() -> Result<(), RejectResponse> {
...
+        canister_id,
+        test_principal2,
+        identity_number2,
+        &jwt2,
+        &salt2,
+    )?;
+
+    assert_eq!(
+        result,
+        Err(OpenIdCredentialAddError::OpenIdCredentialAlreadyRegistered)
```
aterga, line 151: At this point, the system has the following information:
1. The user has access to this Microsoft credential.
2. The same Microsoft credential was already linked to another II.

Would it be possible (and would it be a good idea) to return more info about that second II in the error argument?

I imagine users who observe this error will be asking the question, e.g., "and how do I now log into the other II using this credential?"

lmuntaner: This is a good idea, but probably it's a job for the frontend and we can improve it in another PR. We could point them to use the OpenID button when authenticating.

Sharing the identity number that is linked is not necessary (they can't do anything with it) and is potentially problematic from a privacy point of view.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -51,12 +52,176 @@ fn can_link_google_account() -> Result<(), RejectResponse> {
...
+
+/// Verifies that the same Microsoft account cannot be linked to two different identities
+#[test]
+fn cannot_link_same_microsoft_account_to_two_identities() -> Result<(), RejectResponse> {
+    let env = env();
+    let canister_id = setup_canister(&env);
+    let (jwt, salt, _claims, test_time, test_principal, test_authn_method) =
+        openid_microsoft_test_data();
+    let (jwt2, salt2, _claims2, test_time2, test_principal2, test_authn_method2) =
+        openid_microsoft_test_data_same_microsoft_different_user();
```
aterga: This helper function currently needs to stay in sync with `openid_microsoft_test_data`.

I slightly better test spec would take the default test data, assign it to the `(jwt, salt, _claims, test_time, test_principal, test_authn_method)` tuple, then clone and modify the essential bits of the test data before assigning the second copy to `(jwt2, salt2, _claims2, test_time2, test_principal2, test_authn_method2)`.

Otherwise, to check that this test does what it claims, one needs to jump to the definition of `openid_microsoft_test_data_same_microsoft_different_user` and compare that with the definition of `openid_microsoft_test_data`.

I'll see if I can come up with a concrete suggestion.

src/internet_identity/tests/integration/openid.rs
lmuntaner: Good point, but I wonder what is the right way to call it, add openid credential or linking account. We use "linking" in the user interface.

src/internet_identity/tests/integration/openid.rs
lmuntaner: done!

src/internet_identity/tests/integration/openid.rs
```diff
@@ -557,6 +783,145 @@ fn second_openid_test_data() -> (String, [u8; 32], Claims, u64, Principal, Authn
...
+        jwt.into(),
+        salt,
+        claims,
+        test_time,
+        test_principal,
+        test_authn_method,
+    )
+}
+
+fn openid_microsoft_test_data_same_microsoft_different_user(
```
aterga: As mentioned [here](https://github.com/dfinity/internet-identity/pull/3340/files#r2324407628), I think this function is problematic, as it's hard to see how it relates to `openid_microsoft_test_data`. 

The general question I have after reading the function name is "same to what?"

I see that the following fields are not identical (comparing with `openid_microsoft_test_data`): 
1. `jwt`
2. `salt`
3. `validation_item` (since it's a function of `jwt`)
4. `claims` (since it's a function of `validation_item`)
5. `test_time`
6. `test_principal`
7. `test_pubkey`
8. `test_authn_method` — a function of a bunch of things. 

Is there maybe a typo somewhere?

src/internet_identity/tests/integration/openid.rs
lmuntaner: Yes, it's coupled in the sense that the same Microsoft user is used to build the JWT. However, we don't have a way to create JWTs on the fly.

Maybe I can rename it better though: `openid_microsoft_same_as_one_different_principal_test_data` and then we rename `openid_microsoft_test_data` to `one_openid_microsoft_test_data`?

src/internet_identity/tests/integration/openid.rs
lmuntaner: It uses the same microsoft account to create the JWT, but the rest is different. I proposed a better naming, maybe that can help.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -51,12 +52,176 @@ fn can_link_google_account() -> Result<(), RejectResponse> {
...
+    Ok(())
+}
+
+// Linking Microsoft accounts from different tenants to the same identity is not allowed in the frontend, but is allowed in the backend.
+// This test verifies that the backend permits this behaviour.
+#[test]
+fn can_link_microsoft_account_from_different_tenant() -> Result<(), RejectResponse> {
+    let env = env();
+    let canister_id = setup_canister(&env);
+    let (jwt, salt, _claims, test_time, test_principal, test_authn_method) =
```
aterga, line 171:
```suggestion
    // The tenant is part of `jwt`
    let (jwt, salt, _claims, test_time, test_principal, test_authn_method) =
```

lmuntaner: Added a comment, good point. Also in the other test.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -224,12 +410,12 @@ fn cannot_register_with_faulty_jwt() {
 
     let canister_id = setup_canister(&env);
 
-    let (_jwt, salt, _claims, test_time, test_principal, _test_authn_method) = openid_test_data();
+    let (_jwt, salt, _claims, test_time, test_principal, _test_authn_method) =
```
aterga, line 413: `#[should_panic]` should be used together with `expected`, e.g.

```
#[should_panic(expected = "missing field `signals_end`")]
``` 

Otherwise, the test isn't necessarily testing what it claims, as any panic in this code (even an undesirable one) would lead to the test succeeding.

No change required since it wasn't added in this PR, just wanted to mention.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -406,11 +603,24 @@ pub fn setup_canister(env: &PocketIc) -> Principal {
...
+    mock_certs_response(env, url, mock_certs);
+}
+
+fn mock_microsoft_certs_response(env: &PocketIc) {
```
aterga, line 619: I don't know how to review this, so just trusting that the mocks are fine.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -446,14 +653,42 @@ fn mock_google_certs_response(env: &PocketIc) {
...
+/**
+ * Explanation of the fields used for the test data:
+ * - JWT: the JWT received by the openID provider.
+ * - salt: the salt used to create the JWT from before.
+ * - test_time: the `iat` (issued at) field from the JWT
+ * - test_principal: the principal of the identity used the link the OpenID account.
+ * - test_pubkey: the public key of the credential used to sign in with the identity from `test_principal`.
+ *   Not the public key of the principal of the identity. You don't get it with connection.identity.getPublicKey().
+ *
+ * How to get the test data:
```
aterga, line 674: I wonder if we could automate the generation of this via `playwrite`? 

No changes required for this PR

src/internet_identity/tests/integration/openid.rs
aterga: Then let's always call it linking?

src/internet_identity/tests/integration/openid.rs
aterga: Let's add a comment to explain that one cannot see from the code that these were actually created via the same Microsoft account. That was the main source of confusion for me.

src/internet_identity/tests/integration/openid.rs
aterga: This, or move them to a common `mod`, e.g., 

```
mod openid_microsoft_test {
    lazy_static! {
        // Registered for abc@microsoft.com for principal X
        static ref ACCOUNT_0: (...) = {...};
        // Registered for abc@microsoft.com for principal Y
        // The fact that these data correspond to the same Microsoft account as `ACCOUNT_DATA`
        // cannot be observed at the level of this code. TODO: Explain how to validate this.
        static ref ACCOUNT_1: (...) = {...};
    }
}
```

aterga (approved):

src/internet_identity/tests/integration/openid.rs
lmuntaner: let's do that!

src/internet_identity/tests/integration/openid.rs
lmuntaner: What's the advantage? Just that it's coupled in the same `mod`? That might also confuse of why isn't the other microsoft there as well, no?

src/internet_identity/tests/integration/openid.rs
lmuntaner: Oh, that's a good point!

I can fix it in a following PR. What is `"missing field `signals_end`"`? Is that the exact string with how it should panic?

src/internet_identity/tests/integration/openid.rs
lmuntaner: We probably could but I wonder how worth is for the effort needed.

src/internet_identity/tests/integration/openid.rs
lmuntaner: I'll add a comment to explain the data.

src/internet_identity/tests/integration/openid.rs
lmuntaner: Done!

src/internet_identity/tests/integration/openid.rs
aterga: It's a substring of the panic message

PR #3343, opened 2025-09-04
first

aterga:
# Motivation

The original test was meant as a one-off experiment to validate the data migration. We don't plan to change the migration anymore until the next release, so the original test can now start using fewer accounts to speed up the CI.

# Changes

Use 10 instead of 2,000 accounts in `upgrade_and_rollback_with_realistic_data_migration`.

# Tests

The test being adjusted is still active.


lmuntaner (approved):
I agree. We won't change the migration now, only remove it. So it's safe.

PR #3345, opened 2025-09-05
Push id.ai credentials at the end for 1.0 RP ID calculation

lmuntaner:
# Motivation

We are seeing problems when users log in 1.0 after upgrading their identity.

Context:
- During the upgrade, a new passkey is created in id.ai.
- During 1.0 authentication, II needs to choose a RP ID to fetch the credentials. It uses the RP of the last used passkey.
- This means that the new passkey created during the upgrade will be used to authenticate in 1.0. This was initially considered a feature, not a bug.

The problem comes when users don't know where passkeys are stored. Therefore, the passkey that they use to authenticate in 1.0 is not in the same place where they upgraded. Which triggers a new UX for the user that confuses them because they were used to a different UX.

For example, we saw a user that had the passkey in 1.0 in iCloud. They were used to authenticate with the fingerprint right after clicking the identity number in 1.0. However, during the upgrade, they passkey was created in their Google Password Manager. Next time they went to authenticate in 1.0, the WebAuthn browser modal showed a modal to choose that passkey, instead of requesting the fingerprint immediately. That confused the user.

We have seen other similar reports in the forum of users having problems after upgrading.

# Changes

This pull request updates the logic for ordering WebAuthn flows in `findWebAuthnFlows` to ensure that devices registered on "new flow" origins are moved to the end of the list, while preserving their relative order. It also adds corresponding tests to verify this behavior.

* The `findWebAuthnFlows` function now sorts devices so those registered on origins listed in `canisterConfig.new_flow_origins` are placed at the end, preserving the relative order within each group. This is implemented using a custom sort function.
* Duplicates in the RP IDs list are removed after sorting with `reduce` instead of relying on `Set`, maintaining the order.

# Tests

* A new test verifies that RP IDs from `new_flow_origins` are pushed to the end of the flows, maintaining the relative order of devices.
* Mocks for `canisterConfig` are added in the test setup to simulate the presence of `new_flow_origins`.

### DIFF
--- a/src/frontend/src/lib/utils/findWebAuthnFlows.ts
+++ b/src/frontend/src/lib/utils/findWebAuthnFlows.ts
@@ -1,6 +1,8 @@
 import { II_LEGACY_ORIGIN } from "$lib/legacy/constants";
 import { isNullish, nonNullish } from "@dfinity/utils";
 import { CredentialData } from "./credential-devices";
+import { canisterConfig } from "$lib/globals";
+import { isSameOrigin } from "./urlUtils";
 
 export type WebAuthnFlow = {
   useIframe: boolean;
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/findWebAuthnFlows.ts
```diff
@@ -35,32 +43,58 @@ export const findWebAuthnFlows = ({
...
+    .reduce((rpIds: Array<string | undefined>, rpId: string | undefined) => {
+      if (rpIds.includes(rpId)) {
+        return rpIds;
+      }
+      return [...rpIds, rpId];
+    }, [])
+    .map((rpId) => ({
+      rpId,
+      useIframe: nonNullish(rpId) && rpId !== currentRpId,
+    }));
```
danblackadder, line 97:
```suggestion
const steps: WebAuthnFlow[] = [
  ...new Set(
    devices
      .map(d =>
        d.origin === currentOrigin || (currentOrigin === II_LEGACY_ORIGIN && isNullish(d.origin))
          ? undefined
          : new URL(d.origin ?? II_LEGACY_ORIGIN).hostname,
      )
      .filter(rpId => isNullish(rpId) || relatedRpIds.includes(rpId)),
  ),
].map(rpId => ({
  rpId,
  useIframe: nonNullish(rpId) && rpId !== currentRpId,
}));
```

Why not use a Set here to remove duplicates instead of reduce? 
Set has better performance, and automatically infers removing duplicates without the need to understand the logic of the reduce

lmuntaner: What I don't like about Set is that the order is not guaranteed to be maintained (most browsers do, but it's up to them, not part of the specs), and we care about the order. Moreover, there might be a maximum of 10 items in this list. Therefore, I don't think that performance is important. Instead, I want to make it clear that the order is important

src/frontend/src/lib/utils/findWebAuthnFlows.ts
```diff
@@ -35,32 +43,58 @@ export const findWebAuthnFlows = ({
...
 }: Parameters): WebAuthnFlow[] => {
+  // We need the helpers inside so that when `canisterConfig` is accessed, it already exists.
+  // The devices are expected to be ordered by recently used already
+  // Move devices registered on the new flow origins to the end using toSorted (preserving relative order within groups)
+  const newFlowOrigins = canisterConfig.new_flow_origins[0] ?? [];
+  const isInNewFlow = (credentialData: CredentialData): boolean => {
+    const origin = credentialData.origin ?? II_LEGACY_ORIGIN;
+    return newFlowOrigins.some((o) => isSameOrigin(o, origin));
+  };
+  const sortNewFlowOriginsToEnd = (
```
aterga, line 54:
```suggestion
  const compareCredentials = (
```

src/frontend/src/lib/utils/findWebAuthnFlows.ts
```diff
@@ -35,32 +43,58 @@ export const findWebAuthnFlows = ({
...
-    ),
-  ];
-
-  // Create steps from `deviceRpIds`, currently that's one step per RP ID
-  const steps: WebAuthnFlow[] = orderedDeviceRpIds.map((rpId) => ({
-    rpId,
-    useIframe: nonNullish(rpId) && rpId !== currentRpId,
-  }));
+  // Sort devices in place
+  devices.sort(sortNewFlowOriginsToEnd);
```
aterga, line 72:
```suggestion
  devices.sort(compareCredentials);
```

src/frontend/src/lib/utils/findWebAuthnFlows.ts
aterga: Let's keep the code, as the added complexity serves a purpose.

PR #3347, opened 2025-09-09
Add AuthMethod parameter to login page to allow for automatic authentication from external sources

danblackadder:
# Motivation

Caffeine.AI requires the ability to automatically initiate authentication directly from Caffeine through II by passing a URL parameter that then initiates the flow.

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/utils/authentication/jwt.ts
+++ b/src/frontend/src/lib/utils/authentication/jwt.ts
@@ -8,6 +8,7 @@ import {
   transformSignedDelegation,
 } from "$lib/utils/utils";
 import {
+  Delegation,
   DelegationChain,
   DelegationIdentity,
   ECDSAKeyIdentity,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/legacy/flows/redirect.ts
```diff
@@ -62,5 +62,26 @@ export const redirectInPopup = (url: string): Promise<string> => {
     "_blank",
     `width=${width},height=${height},left=${left},top=${top}`,
   );
+
+  // If the popup window is blocked for any reason by the browser
```
danblackadder, line 66: Although this is not relevant to this PR anymore as we have moved away from popups, this still feels like a good fix for the chance that the popup does get blocked for any reason to keep the user informed.

src/frontend/src/lib/utils/openID.ts
```diff
@@ -127,6 +130,13 @@ const requestWithRedirect = async (
     authURL.searchParams.set("login_hint", options.loginHint);
   }
 
+  if (nonNullish(options.useFullRedirect)) {
+    sessionStorage.setItem("openid_state", state);
+    sessionStorage.setItem("openid_nonce", options.nonce);
+    window.location.href = authURL.toString();
+    return new Promise(() => {}); // never resolves
```
danblackadder: this is purely to please TypeScript, ultimately the users url is changed, so this will never get hit anyway, unless for some weird strange reason window.location.href fails O.o

src/frontend/src/lib/legacy/flows/redirect.ts
lmuntaner: Yes, that's a good idea!

Did you test this?

lmuntaner (commented):
I did a PoC of this last week.

https://github.com/dfinity/internet-identity/pull/3349

The problem is that this isn't working on (at least) Safari because the popup is blocked.

Did you test your PR? I don't see how it's triggered... How where you planning on triggering it? Did you test it?

danblackadder:
> I did a PoC of this last week.
> 
> #3349
> 
> The problem is that this isn't working on (at least) Safari because the popup is blocked.
> 
> Did you test your PR? I don't see how it's triggered... How where you planning on triggering it? Did you test it?

I wasn't able to test on Safari, because Apple outright refuse to allow me to download it on Windows :)

However, we are not using a popup, so I see no reason why this would fail - as it is instead a redirect - it works on Chrome and Edge.

If you can test on Safari that would be appreciated :)

src/frontend/src/lib/legacy/flows/redirect.ts
danblackadder: Yes, I tested this in my initial route of popup display, and this was working on both Chrome and Edge when the popup was blocked. 

It was a nice little UI upgrade to catch certain situations

src/frontend/src/routes/(new-styling)/auth-redirect/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+        return "cancelled";
+      }
+      handleError(error);
+    }
+  };
+
+  if (provider) {
+    handleContinueWithOpenId(provider);
+  } else {
+    // if an incorrect provider is provided redirect the user to the login page
```
edenchazard:
```suggestion
    // If an incorrect provider is provided, redirect the user to the login page.
```

src/frontend/src/routes/(new-styling)/auth-redirect/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+    goto(`/login${data.next ? `next=${data.next}` : ""}`);
+  }
+</script>
+
+<div class="flex min-h-[100dvh] flex-col">
+  <div class="h-[env(safe-area-inset-top)]"></div>
+  <div class="flex h-dvh w-full flex-col items-center justify-center gap-2">
+    {#if provider}
+      <p class="text-text-primary">Redirecting you to {provider?.name}</p>
+    {/if}
```
edenchazard: In svelte, is the `#if provider` not enough to negate the need for the optional chaining on `provider?.name`?

PR #3363, opened 2025-09-24
feat: Support `AccountInfo.is_default` in II backend API

aterga:
# Motivation

This is the first step towards supporting the new concept of default accounts. The FE needs a way to distinguish default and non-default accounts associates with a (identity, application) pair.

# Changes

Add the `is_default` to the Candid API and propagate it through the Rust types. For now, we keep the value set to None in all places.

# Tests

Existing CI tests should suffice.

### DIFF
--- a/src/internet_identity/src/storage/account.rs
+++ b/src/internet_identity/src/storage/account.rs
@@ -146,7 +146,7 @@ impl Account {
             origin: self.origin.clone(),
             last_used: self.last_used,
             name: self.name.clone(),
-            is_default: self.is_default.clone(),
+            is_default: self.is_default,
         }
     }
 


src/internet_identity/internet_identity.did
```diff
@@ -775,6 +775,7 @@ type AccountInfo = record {
     last_used : opt Timestamp;
     // Configurable properties
     name : opt text;
+    is_default : opt bool;
```
aterga, line 778: @danblackadder, this is how the FE can tell which account is the default. 

In case this field is `null`, I suggest assuming that the default account = the primary account.

lmuntaner: Why do we need to set it as optional? I don't think we need to keep backwards compatibility.

src/internet_identity/src/storage/account.rs
```diff
@@ -65,6 +65,7 @@ pub struct Account {
     pub origin: FrontendHostname,
     pub last_used: Option<Timestamp>,
     pub name: Option<String>,
+    pub is_default: Option<bool>,
```
lmuntaner, line 68: Does it need to be an `Option`?

aterga: Adding all new record fields as optional is the recommended way to extend a Candid API. 

Since the II has been developed in a special way (with BE and FE changes being synchronized), I've explored also the alternative of adding the field as non-optional, see [here](https://github.com/dfinity/internet-identity/pull/3369)

We should discuss which approach to follow (I'm still leaning towards optional fields).

PR #3365, opened 2025-09-24
Dan/sign up create non functional updates

danblackadder:
# Motivation

We are updating the AuthFlow wording and providing additional information to the user.
Currently hidden behind a feature flag

# Changes

- Update popover to enable the user to click outside without closing modal beneath it
- Add backdrop to popover so that is front and centre on the UI so the user understands why clicking elements on the page outside of the popover won't work
- Handle AUTH_FOW_UPDATES feature flag text and tooltip updates across Creation and Existing auth views

# Tests

Tested locally on Chrome on Windows
e2e tests to be updated prior to go live and removal of feature flag


src/frontend/src/lib/components/ui/Popover.svelte
```diff
@@ -92,6 +94,12 @@
 <svelte:window bind:innerWidth={windowWidth} />
 
 {#if windowWidth >= MOBILE_BREAKPOINT}
+  <div
```
danblackadder, line 87: Due to modals, this is required to allow the user to click the backdrop with a popover open to then allow them to close, however our popovers do not have a close button associated with them?

Should I add an sr-only close button?

sea-snake: I think that's what ESC buttons are probably used for in an accessibility use case, but I'd recommend to have a look what's commonly recommended in this regard and make changes accordingly (W3C has various accessibility best practice examples around these sort of topics).

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
```diff
@@ -65,7 +71,12 @@
       <p
         class="text-md text-text-tertiary font-medium text-balance sm:text-center"
       >
-        This will label your passkey, and you can't rename it later once set.
+        {#if $AUTH_FLOW_UPDATES}
+          Internet Identity <b>does not</b> store your biometric data. It stays on
```
aterga, line 75: This text is hard coded more than once, would it be easy to factor it out to only have to edit the wording in one place?

danblackadder: To be honest this is quite pointless to do right now, as once we have internationalisation this will be taken care of anyway

src/frontend/src/lib/components/wizards/auth/views/SetupOrUseExistingPasskey.svelte
```diff
@@ -36,14 +44,24 @@
   <h1 class="text-text-primary mb-3 text-2xl font-medium sm:text-center">
     Simplify your sign-in
   </h1>
+
   <p class="text-md text-text-tertiary font-medium text-balance sm:text-center">
-    With passkeys, you can now use your fingerprint, face, or screen lock to
-    quickly and securely confirm it’s really you.
+    {#if $AUTH_FLOW_UPDATES}
+      Create an identity with a passkey, using biometrics or a security key.
```
aterga, line 50: Nit

In some places, Passkey is capitalized, but not here 

Let's make it consistent

aterga (approved):

src/frontend/src/lib/components/ui/Popover.svelte
danblackadder: Have updated for better accessibility

src/frontend/src/lib/components/wizards/auth/views/SetupOrUseExistingPasskey.svelte
danblackadder: As per design, agreed we are moving from `Passkeys` to `passkeys` going forward

src/frontend/src/lib/components/ui/Popover.svelte
```diff
@@ -77,34 +79,45 @@
...
     onintrostart={() => popoverRef?.showPopover()}
     onoutrostart={() => popoverRef?.hidePopover()}
-    class="popover fixed overflow-visible bg-transparent"
+    onfocusout={(e) => {
+      if (!popoverRef?.contains(e.relatedTarget as Node)) {
+        anchorRef?.querySelector("button")?.focus();
+        onClose?.();
+      }
+    }}
+    class="popover backdrop:bg-bg-overlay fixed overflow-visible bg-transparent backdrop:opacity-40"
```
sea-snake: Are you sure we need a visual backdrop? Since in comparison a popover isn't a modal thus from an accessibility perspective, the rest of the page is still interact-able, therefore I wouldn't cover it with a visible backdrop (neither do we have that in the designs).

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
```diff
@@ -111,5 +124,49 @@
...
+            />
+          </Button>
+        </div>
+      </div>
+      {#if showPrivacyPopover}
+        <Popover
+          anchor={popoverAnchorRef}
+          direction="up"
+          align="end"
+          distance="10px"
```
sea-snake: Let's use `rem` here to keep it consistent with other UI behavior when UI is scaled.

```suggestion
          distance="0.625rem"
```

Currently tooltips are `0.5rem` distance, so if we want this to have the same distance see below suggestion instead:

```suggestion
          distance="0.5rem"
```

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
```diff
@@ -130,15 +133,33 @@
...
-    >
-      Upgrade from legacy identity
-    </Button>
+    {#if $AUTH_FLOW_UPDATES}
+      <div class="border-border-tertiary border-t"></div>
+      <div class="flex flex-row items-center justify-between gap-4">
+        <p class="text-text-secondary text-sm">
+          Still have an identity number?
+        </p>
+        <Button
```
sea-snake: The focus outline will be weirdly aligned assuming we align the button text, maybe we could use a `button` element and underline the text on focus instead?

Fun fact, this "new" upgrade CTA design was actually already implemented before and then replaced with the current button at a later point 😅

Here's the previous time we implemented this UX with an underline on focus button so that the button text itself could be visually aligned correctly: https://github.com/dfinity/internet-identity/blob/227cfe0414e5c8292204dd3a9469b47c53cd8a7d/src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte#L84

src/frontend/src/lib/components/ui/Popover.svelte
danblackadder: Removed

src/frontend/src/lib/components/wizards/auth/views/CreatePasskey.svelte
danblackadder: Updated to 0.5rem

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
danblackadder: Change to <button>

PR #3371, opened 2025-09-25
Add ICRC-25 protocol to test app

sea-snake:
Add ICRC-25 protocol to test app.

# Changes

- Add signer-js dependencies to test app.
- Add checkbox to switch to ICRC-25 protocol.
- Use signer-js to request delegation when above checkbox is checked.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -112,27 +112,6 @@
         "vite-plugin-node-polyfills": "^0.23.0"
       }
     },
-    "demos/test-app/node_modules/@slide-computer/signer": {
-      "version": "3.18.0",
-      "resolved": "https://registry.npmjs.org/@slide-computer/signer/-/signer-3.18.0.tgz",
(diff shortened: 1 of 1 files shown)


demos/test-app/src/auth.ts
```diff
@@ -30,14 +32,30 @@ export const authWithII = async ({
...
   url: string;
   maxTimeToLive?: bigint;
   allowPinAuthentication?: boolean;
   derivationOrigin?: string;
   autoSelectionPrincipal?: string;
   sessionIdentity: SignIdentity;
+  useIcrc25?: boolean;
 }): Promise<{ identity: DelegationIdentity; authnMethod: string }> => {
+  // Authenticate with signer-js instead if we use the ICRC-25 protocol
+  if (useIcrc25) {
```
aterga, line 46: Nit, why not have this as a function on the same level as `authWithII`?

Could call it, e.g., `authWithSignerJs`

sea-snake: I considered that, but rather opted to keep it all in one place for now so there aren't two methods to maintain and keep in sync (parameters and return types).

I'll likely revisit and clean things up further once signer-js has been updated and moved to icp-sdk.

PR #3372, opened 2025-09-25
Add ICRC-25 RPC protocol to II

sea-snake:
Add ICRC-25 RPC protocol

# Changes

- Implement `rpcAuthenticationProtocol` with signer-js based on existing `authenticationProtocol`.
- Add `rpc` option to `authorizationStore.init()` to toggle between the two implementations.


src/frontend/src/lib/utils/postMessageInterface.ts
```diff
@@ -0,0 +1,225 @@
...
+> {
+  return new Promise((resolve) => {
+    authorizeClientFunnel.init();
+    onProgress("waiting");
+    new HeartbeatServer({
+      onDisconnect(): void {
+        resolve("closed");
+      },
+      onEstablish(origin: string, source: MessageEventSource): void {
+        window.addEventListener("message", async (event) => {
```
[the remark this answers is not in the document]
sea-snake: Connection is only established once, so this isn't an issue.

src/frontend/src/lib/utils/postMessageInterface.ts
```diff
@@ -0,0 +1,239 @@
...
+            ) {
+              authorizeClientFunnel.trigger(
+                AuthorizeClientEvents.AuthenticateError,
+                {
+                  origin: requestOrigin,
+                  failureReason: authenticateResult.text,
+                },
+              );
+              try {
+                (source as WindowProxy).postMessage(
```
lmuntaner: Why is casting necessary here and not before?

sea-snake: Ah yes, I can correct the typing in signer-js already to avoid the need to correct it here.

sea-snake: To clarify, by default the source of a message event is `type MessageEventSource = WindowProxy | MessagePort | ServiceWorker;` since TS doesn't know where the message event is originating from (another window, message broadcast channel or service worker).

I've updated signer-js to return WindowProxy instead since in this particular case it's coming from another window.

PR #3374, opened 2025-09-25
feat: Setting and getting default accounts

aterga:
# Motivation

This is the next step towards supporting the new concept of default accounts. The BE now allows authorized clients to set and get default accounts for each (anchor, origin).

# Changes

* Implemented `set_default_account` and `get_default_account`.
* `update_account_for_origin` initializes the default account in the `anchor_application_config` index while creating that account.

# Tests

## Coverage Summary
* Happy paths: Default retrieval, customization, reset, multi-origin
* Error cases: Non-existent anchor/origin/account
* Edge cases: Account state changes, operation sequencing 

### get_default_account_for_origin Tests

1. `should_get_default_account_for_origin` - Comprehensive integration test covering:
    * GET without SET (returns synthetic default)
    * SET to existing accounts + GET (customizes default)
    * SET to non-existent accounts + GET (SET fails, GET unchanged)
    * SET to None + GET (resets to synthetic default)
2. `can_get_default_before_update_account_for_origin` - Returns the synthetic account for uninitialized origins.
3. `should_get_updated_default_account_after_modification` - Returns updated default account with assigned account number after modification.
4. `should_fail_get_default_account_for_nonexistent_anchor` - Returns the synthetic account for non-existent anchors.
5. `should_get_default_account_for_different_origins` - Each origin has independent default accounts.

### set_default_account_for_origin Tests

Test scenarios:

✅ SET to existing accounts (succeeds)
✅ SET to non-existent accounts (fails with NoSuchAccount)
✅ SET to None (resets to synthetic default)
✅ State persistence across operations

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -653,7 +653,7 @@ type SetDefaultAccountError = variant {
     };
     NoSuchAccount : record {
         anchor_number : UserNumber;
-        account_number : opt AccountNumber;
+        origin : FrontendHostname;
     };
 };
 
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -1011,7 +1011,7 @@ service : (opt InternetIdentityInit) -> {
     get_default_account : (
         anchor_number : UserNumber,
         origin : FrontendHostname,
-    ) -> (variant { Ok : AccountInfo; Err: GetDefaultAccountError });
+    ) -> (variant { Ok : AccountInfo; Err: GetDefaultAccountError }) query;
 
     set_default_account : (
         anchor_number : UserNumber,
--- a/src/internet_identity/src/account_management.rs
+++ b/src/internet_identity/src/account_management.rs
@@ -44,28 +44,28 @@ pub fn get_accounts_for_origin(
     storage_borrow(|storage| storage.list_accounts(anchor_number, origin))
 }
 
+/// Helper function to read an account by application number, unlike `storage.read_account` this
+/// returns `Result` instead of `Option`.
+///
+/// This function is applicable only to numbered accounts; synthetic accounts are not currently
+/// stored in the memory and thus cannot be fetched from the storage.
 fn try_read_account_info(
     anchor_number: AnchorNumber,
-    origin: &FrontendHostname,
+    origin: FrontendHostname,
     application_number: ApplicationNumber,
-    account_number: Option<AccountNumber>,
+    account_number: AccountNumber,
 ) -> Result<AccountInfo, String> {
     let Some(account) = storage_borrow(|storage| {
         storage.read_account(ReadAccountParams {
-            account_number,
+            account_number: Some(account_number),
             anchor_number,
-            origin,
+            origin: &origin,
             known_app_num: Some(application_number),
         })
     }) else {
-        let account_str = account_number
-            .map(|num| format!("#{}", num))
-            // This can happen if update_account_for_origin was never called to init this account.
-            .unwrap_or_else(|| "without a number".to_string());
-
         let message = format!(
-            "Account {} does not exist for anchor {} and origin {}.",
-            account_str, anchor_number, origin
+            "Account #{} does not exist for anchor {} and origin {}.",
+            account_number, anchor_number, origin
         );
 
         return Err(message);
@@ -74,6 +74,9 @@ fn try_read_account_info(
     Ok(account.to_info())
 }
 
+/// Best effort to determin the default account for the given (anchor, origin).
+///
+/// An Err case would indicate internal inconsistency in the canister state.
 pub fn get_default_account_for_origin(
     anchor_number: AnchorNumber,
     origin: FrontendHostname,
@@ -81,38 +84,56 @@ pub fn get_default_account_for_origin(
     let Some(application_number) =
         storage_borrow(|storage| storage.lookup_application_number_with_origin(&origin))
     else {
-        return Err(GetDefaultAccountError::NoSuchOrigin { anchor_number });
+        return Ok(Account::synthetic(anchor_number, origin).to_info());
     };
 
     let AnchorApplicationConfig {
-        default_account_number: account_number,
+        default_account_number,
     } = storage_borrow(|storage| {
         storage.lookup_anchor_application_config(anchor_number, application_number)
     });
 
-    let account = try_read_account_info(anchor_number, &origin, application_number, account_number)
-        .map_err(GetDefaultAccountError::InternalCanisterError)?;
+    let Some(default_account_number) = default_account_number else {
+        return Ok(Account::synthetic(anchor_number, origin).to_info());
+    };
+
+    let account = try_read_account_info(
+        anchor_number,
+        origin,
+        application_number,
+        default_account_number,
+    )
+    .map_err(GetDefaultAccountError::InternalCanisterError)?;
 
     Ok(account)
 }
 
+/// Sets the default account for the given (anchor, origin) to the specified `account_number`.
+///
+/// If this is the first time an origin is seen for the anchor, a new application number is created.
 pub fn set_default_account_for_origin(
     anchor_number: AnchorNumber,
     origin: FrontendHostname,
     account_number: Option<AccountNumber>,
 ) -> Result<AccountInfo, SetDefaultAccountError> {
-    let Some(application_number) =
-        storage_borrow(|storage| storage.lookup_application_number_with_origin(&origin))
-    else {
-        return Err(SetDefaultAccountError::NoSuchOrigin { anchor_number });
-    };
+    let application_number = storage_borrow_mut(|storage| {
+        storage.lookup_or_insert_application_number_with_origin(&origin)
+    });
 
-    // Ensure the account exists for this (anchor, origin).
-    let account = try_read_account_info(anchor_number, &origin, application_number, account_number)
+    let account = if let Some(account_number) = account_number {
+        try_read_account_info(
+            anchor_number,
+            origin.clone(),
+            application_number,
+            account_number,
+        )
         .map_err(|_| SetDefaultAccountError::NoSuchAccount {
             anchor_number,
             origin,
-        })?;
+        })?
+    } else {
+        Account::synthetic(anchor_number, origin).to_info()
+    };
 
     let config = AnchorApplicationConfig {
         default_account_number: account_number,
@@ -443,7 +464,7 @@ fn should_get_accounts_for_origin() {
     assert_eq!(
         get_accounts_for_origin(anchor_number, &origin),
         vec![
-            Account::new(anchor_number, origin.clone(), None, None),
+            Account::synthetic(anchor_number, origin.clone()),
             Account::new_full(
                 anchor_number,
                 origin.clone(),
@@ -485,7 +506,7 @@ fn should_only_get_own_accounts_for_origin() {
     assert_eq!(
         get_accounts_for_origin(anchor_number, &origin),
         vec![
-            Account::new(anchor_number, origin.clone(), None, None),
+            Account::synthetic(anchor_number, origin.clone()),
             Account::new_full(
                 anchor_number,
                 origin.clone(),
@@ -500,7 +521,7 @@ fn should_only_get_own_accounts_for_origin() {
     assert_eq!(
         get_accounts_for_origin(anchor_number_two, &origin),
         vec![
-            Account::new(anchor_number_two, origin.clone(), None, None),
+            Account::synthetic(anchor_number_two, origin.clone()),
             Account::new_full(
                 anchor_number_two,
                 origin.clone(),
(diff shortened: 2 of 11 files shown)


src/internet_identity/src/account_management.rs
```diff
@@ -724,3 +810,284 @@ fn should_increment_discrepancy_counter() {
...
+            )),
+            Ok(AccountInfo {
+                account_number: Some(2),
+                origin: origin.clone(),
+                last_used: None,
+                name: Some("Bob".to_string()),
+            }),
+        ),
+        (
+            "SET to None (reset to default), then GET (can set default to None again)",
```
danblackadder, line 949: Is it possible to make this situation throw an error instead if None is passed as the account_number?

It doesn't make sense that I should be able to call `set` with `none` does it?

I know we are supporting `None` by default, as the default primary, but going forward should we continue to support this if we are trying to improve the situation is all I am thinking? From a UI perspective, there is never a case you would technically be able to call this anyway, other than if we either added a button to say "Remove all defaults" or the canister is directly called via the user outside of the UI

aterga: Good question.

I think currently setting to None is the only way to set the primary account as the default. 

If we wanted to migrate primary accounts to, e.g., `Some(0)` — which we might be able to do, we would need to implement some orthogonal changes on the BE

This approach assumes that we don't yet want to do that migration but already want default accounts to be customizable.

src/internet_identity/internet_identity.did
```diff
@@ -653,7 +653,7 @@ type SetDefaultAccountError = variant {
     };
     NoSuchAccount : record {
         anchor_number : UserNumber;
-        account_number : opt AccountNumber;
+        origin : FrontendHostname;
```
lmuntaner, line 656: This means you need to generate the frontend types again, right?

src/internet_identity/src/account_management.rs
```diff
@@ -39,6 +44,87 @@ pub fn get_accounts_for_origin(
...
+}
+
+pub fn get_default_account_for_origin(
+    anchor_number: AnchorNumber,
+    origin: FrontendHostname,
+) -> Result<AccountInfo, GetDefaultAccountError> {
+    let Some(application_number) =
+        storage_borrow(|storage| storage.lookup_application_number_with_origin(&origin))
+    else {
+        return Err(GetDefaultAccountError::NoSuchOrigin { anchor_number });
```
lmuntaner: We should not fail in this case. Applications are created only when accounts are created, but we might request the default account before creating them.

If no application exists, it means that no account was ever created, so it should send the "Primary Account". Btw, do we have a new name for this? We used to call them "Default account".

src/internet_identity/src/account_management.rs
```diff
@@ -39,6 +44,87 @@ pub fn get_accounts_for_origin(
     storage_borrow(|storage| storage.list_accounts(anchor_number, origin))
 }
 
+fn try_read_account_info(
```
lmuntaner, line 52: It would be good to have some comments to understand the intended functionality of this function.

src/internet_identity/src/account_management.rs
```diff
@@ -39,6 +44,87 @@ pub fn get_accounts_for_origin(
...
+            account_str, anchor_number, origin
+        );
+
+        return Err(message);
+    };
+
+    Ok(account.to_info())
+}
+
+pub fn get_default_account_for_origin(
```
lmuntaner, line 80: Ditto adding some comments to understand it.

src/internet_identity/src/account_management.rs
```diff
@@ -39,6 +44,87 @@ pub fn get_accounts_for_origin(
...
+    Ok(account)
+}
+
+pub fn set_default_account_for_origin(
+    anchor_number: AnchorNumber,
+    origin: FrontendHostname,
+    account_number: Option<AccountNumber>,
+) -> Result<AccountInfo, SetDefaultAccountError> {
+    let Some(application_number) =
+        storage_borrow(|storage| storage.lookup_application_number_with_origin(&origin))
```
lmuntaner: We should use lookup_or_insert here, shouldn't we? Applications are created when accounts are created, and here we might be creating an account by setting it as default.

src/internet_identity/src/account_management.rs
```diff
@@ -724,3 +810,284 @@ fn should_increment_discrepancy_counter() {
...
+
+        assert_eq!(
+            result, expected_get_result,
+            "{label}: unexpected GET result"
+        );
+    }
+}
+
+#[test]
+fn cannot_get_default_before_update_account_for_origin() {
```
lmuntaner: Why this functionality? Accounts are created when non-default data needs to be stored, therefore, getting accounts before creating them just returns the (old) default account ("Primary account").

See read_account logic, that might be helpful as well.

src/internet_identity/src/storage.rs
```diff
@@ -1232,6 +1278,17 @@ impl<M: Memory + Clone> Storage<M> {
         // Get or create an application number from the account's origin.
         let application_number =
             self.lookup_or_insert_application_number_with_origin(&params.origin);
+
+        // Update default account in the (anchor, origin) config.
```
lmuntaner, line 1282: This "Default account" is the "Primary Account", it doesn't have the same meaning as your "Default Account". We might want to rename this function...

Therefore, I don't think you should update the config here.

Default accounts are created for example updating the name of it. It might be that the user wants to rename from "Primary account" to another name. Those account don't necessarily exist in the storage, so we need to first create them before updating them.

See "update_account" for an example.

src/internet_identity/src/storage/account.rs
```diff
@@ -65,7 +65,7 @@ pub struct Account {
     pub origin: FrontendHostname,
     pub last_used: Option<Timestamp>,
     pub name: Option<String>,
-    seed_from_anchor: Option<AnchorNumber>,
+    pub(crate) seed_from_anchor: Option<AnchorNumber>,
```
lmuntaner: Why is this needed?

src/internet_identity/internet_identity.did
aterga: Yes, good catch

src/internet_identity/src/account_management.rs
aterga: Currently, `get_default_account_for_origin` is an update function, so we could do that,

but it should probably be changed to query. In which case we won't be able to modify the state.

src/internet_identity/src/account_management.rs
aterga: Yes, this is going to change. Thanks for confirming the intended behavior in this case.

src/internet_identity/src/storage.rs
aterga: Yes, we agreed with @sea-snake to do the renaming in a separate PR. 

For now, the terminology will be confusing for new readers, but at least we're all on the same page.

src/internet_identity/src/account_management.rs
lmuntaner: This is `set_default_account_for_origin`, it will always be an `update`, no?

In `get_default_account_for_origin` we shuold assume there are no accounts and return the "Primary Account" (shuold we start calling it that?)

aterga: > "Primary Account" (shuold we start calling it that?)

I prefer "synthetic" as it makes it clear it's synthesized and not stored. 

Starting to use the new convention in this PR already, let me know if you disagree.

src/internet_identity/src/account_management.rs
aterga: Changed this to `can_get_default_before_update_account_for_origin`

src/internet_identity/src/account_management.rs
lmuntaner: Sounds good!

src/internet_identity/src/storage/account.rs
aterga: Good catch, it isn't needed. Undoing the change.

src/internet_identity/internet_identity.did
```diff
@@ -1011,7 +1011,7 @@ service : (opt InternetIdentityInit) -> {
     get_default_account : (
         anchor_number : UserNumber,
         origin : FrontendHostname,
-    ) -> (variant { Ok : AccountInfo; Err: GetDefaultAccountError });
+    ) -> (variant { Ok : AccountInfo; Err: GetDefaultAccountError }) query;
```
lmuntaner, line 1014: Did you add this before or after the declarations? I'm not sure, but I think that the code generated changes also with query calls.

src/internet_identity/src/account_management.rs
```diff
@@ -39,6 +44,108 @@ pub fn get_accounts_for_origin(
...
+            account_number, anchor_number, origin
+        );
+
+        return Err(message);
+    };
+
+    Ok(account.to_info())
+}
+
+/// Best effort to determin the default account for the given (anchor, origin).
```
lmuntaner, line 77: Small typo "determin".

lmuntaner: I would add in the comment the scenarios:

* If application not found: synthetic account.
* If no default account stored: synthetic account.
* Else, read from storage, raise error if not found.

src/internet_identity/src/storage/storable/anchor_application_config.rs
```diff
@@ -0,0 +1,26 @@
...
+use ic_stable_structures::storable::Bound;
+use ic_stable_structures::Storable;
+use minicbor::{Decode, Encode};
+use std::borrow::Cow;
+
+#[derive(Encode, Decode, Default, Clone, Ord, Eq, PartialEq, PartialOrd)]
+#[cbor(map)]
+pub struct AnchorApplicationConfig {
+    #[n(0)]
+    pub default_account_number: Option<StorableAccountNumber>, // None is the unreserved default account
```
lmuntaner, line 11: Change "None is the unreserved default account" for "the unreserved synthetic account"?

src/internet_identity/src/account_management.rs
```diff
@@ -39,6 +44,108 @@ pub fn get_accounts_for_origin(
...
+        application_number,
+        default_account_number,
+    )
+    .map_err(GetDefaultAccountError::InternalCanisterError)?;
+
+    Ok(account)
+}
+
+/// Sets the default account for the given (anchor, origin) to the specified `account_number`.
+///
```
lmuntaner, line 112: I would add a comment that if the account_number is None, then the synthetic account is returned and also `None` is stored.

src/internet_identity/src/main.rs
```diff
@@ -388,31 +388,33 @@ fn update_account(
...
 
 #[update]
 fn set_default_account(
     anchor_number: AnchorNumber,
-    _origin: FrontendHostname,
-    _account_number: Option<AccountNumber>,
+    origin: FrontendHostname,
+    account_number: Option<AccountNumber>,
 ) -> Result<AccountInfo, SetDefaultAccountError> {
     check_authorization(anchor_number)
```
lmuntaner, line 411: I remembered now that we have a helper `anchor_operation_with_authz_check`. Worth doing in another PR though.

The idea is that all update operations are stored then in the archive canister.

Which also means that this operation or setting an account as default must be stored there as well.

src/internet_identity/src/storage.rs
lmuntaner: I would add in the comment of `create_default_account` that the account will be set as default as well.

Did you think of all the use cases for this function that it makes sense to set it as default account as well? Isn't there an edge case where another default account already exists and we don't want to substitute it?

Maybe setting the default should be done where this is called instead?

src/internet_identity/src/storage/account.rs
```diff
@@ -69,6 +69,18 @@ pub struct Account {
 }
 
 impl Account {
+    /// A default account that is not meant to be stored.
```
lmuntaner, line 72: I would not call it "default"... hehe.

src/internet_identity/src/storage/account.rs
```diff
@@ -69,6 +69,18 @@ pub struct Account {
...
+    pub fn synthetic(anchor_number: AnchorNumber, origin: FrontendHostname) -> Self {
```
lmuntaner, line 73: Great idea!!!

lmuntaner (approved):
Some more comments, but already approving!

src/internet_identity/internet_identity.did
aterga: Good catch, [done](https://github.com/dfinity/internet-identity/pull/3382)

src/internet_identity/src/main.rs
aterga: Why isn't this used in `update_account`?

aterga: Oh I see, it's just done directly via post_operation_bookkeeping

PR #3375, opened 2025-09-26
using-dev-build uses icp-js-auth instead of @dfinity/auth-client

lmuntaner:
# Motivation

Start using the new npm package icp-js-auth instead of the deprecated `@dfinity/auth-client`.

This will allow me to test changes in auth-client by linking the library to a local branch.

# Changes

* Install new package in `using-dev-build` project.
* Change import.

# Tests

Tested locally, and I can still authenticate.

### DIFF
--- a/demos/using-dev-build/tsconfig.json
+++ b/demos/using-dev-build/tsconfig.json
@@ -7,7 +7,7 @@
     "skipLibCheck": true,
 
     /* Bundler mode */
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "allowImportingTsExtensions": true,
     "resolveJsonModule": true,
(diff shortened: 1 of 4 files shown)


demos/using-dev-build/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "skipLibCheck": true,
 
     /* Bundler mode */
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
aterga, line 10: Why did this need to change?

lmuntaner: To be able to import from submodules `icp-js/auth/client`. The `/client` is a submodule.

PR #3381, opened 2025-09-30
Make session store persistent within tab

sea-snake:
Make session store persistent within tab.

# Changes

- Updated session store to keep persistent session identity (within a single tab, new tab results in a new session).
- Added check to ignore persisted sessions older than 5 min, a large enough time window for redirect flows.
- Updated switch identity and logout to reset session store, this makes sure sessions are not re-used across different identities.


src/frontend/src/lib/stores/session.store.ts
```diff
@@ -30,19 +31,75 @@ type SessionStore = Readable<Session> & {
...
+const read = async () => {
+  const item = sessionStorage.getItem(STORAGE_KEY);
+  if (isNullish(item)) {
+    return undefined;
+  }
+  const { privateJwk, publicJwk, nonce, salt, createdAt } = JSON.parse(item);
+  // Ignore persisted sessions older than 5 minutes
+  const expiry = new Date().getTime() + 5 * 60 * 1000;
+  if (parseInt(createdAt) > expiry) {
+    return undefined;
```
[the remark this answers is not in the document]
sea-snake: It's checking an older commit...

src/frontend/src/lib/stores/session.store.ts
```diff
@@ -30,19 +31,75 @@ type SessionStore = Readable<Session> & {
...
+    ["verify"],
+  );
+  return {
+    identity: await ECDSAKeyIdentity.fromKeyPair({ publicKey, privateKey }),
+    nonce,
+    salt: new Uint8Array(fromBase64(salt)),
+  };
+};
+
 export const sessionStore: SessionStore = {
```
lmuntaner, line 95: Could we add some tests for the new functionality?

Already approving.

lmuntaner (approved):

src/frontend/src/lib/stores/session.store.ts
sea-snake: Sure, will add some tests.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -79,6 +80,7 @@
   );
 
   const handleSwitchIdentity = async (identityNumber: bigint) => {
+    await sessionStore.reset();
```
aterga, line 83: It seems persistent session store is being reset in 3 different places in this PR.

What is the guiding principle for deciding when it needs to be reset?

sea-snake: Whenever the identity changes, it should be reset to avoid re-using session key pairs across different identities. 

This resolves the "shared device" edge case that came up earlier in discussions with prodsec, which due to the added persistency in this PR is of higher concern.

aterga: > Whenever the identity changes

Are we talking about a change w.r.t. the BE settings of a particular identity, or something else?

> it should be reset to avoid re-using session key pairs across different identities

Is there a thread that explains the worst case scenario if we were to forget to do this?

sea-snake: Discussed in DM

PR #3385, opened 2025-10-01
feat: Archive `set_default_account` operations

aterga:
# Motivation

All identity operations should be recorded in the archive, and the new `set_default_account` operations are not an exception.

# Changes

Wrap the internal function call into `anchor_operation_with_authz_check` which is the standard way to invoke identity operations that should be archived.

# Tests

Existing tests should suffice.


lmuntaner (approved):
Nice! Thanks!

PR #3389, opened 2025-10-03
Authorize default account in callback

sea-snake:
Authorize default account in callback.

# Changes

- Add Caffeine entry and logo to known dapps.
- Adjust `AuthorizeHeader` to render wider logos larger.
- Create `AuthorizeError` component so it can be re-used.
- Update `AuthFlow` to add `trackLastUsed` option, so we can disable this in the direct OpenID authentication flow.
- Update `continueWithOpenId` method in `AuthFlow` to add optional `existingJwt` param so the JWT from the direction OpenID authentication flow can be used.
- Update `AuthorizationStore` to make the `syncLastUsedAccounts` call optional, skipping it if the identity isn't stored in last used to begin with.
- Move callback page to `(new-styling)` and add progress bar and content according to direct OpenID authentication design.
- Add implementation to this page that either signs in or signs up with OpenID before automatically authorizing with the default account.


src/frontend/src/lib/legacy/flows/dappsExplorer/dapps.json
```diff
@@ -790,5 +790,11 @@
     "oneLiner": "Experience next-gen decentralized authentication.",
     "website": "https://try.id.ai",
     "logo": "try-id-ai.svg"
+  },
+  {
+    "name": "Caffeine",
+    "oneLiner": "Self-writing apps platform: the first complete tech stack designed for AI.",
+    "website": "https://caffeine.ai",
```
lmuntaner, line 797: Is the derivation origin used to find the known dapp or the alternative origin?

I believe that Caffeine uses app.caffeine.ai as derivation origin.

src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -113,16 +113,21 @@ export const authorizationStore: AuthorizationStore = {
...
-              .get_accounts(identityNumber, effectiveOrigin)
-              .then(throwCanisterError)
-              .then((accounts) =>
-                lastUsedIdentitiesStore.syncLastUsedAccounts(
-                  identityNumber,
-                  effectiveOrigin,
-                  accounts,
-                ),
-              );
+            // Only fetch and sync accounts if identity is actually stored
```
lmuntaner, line 116: This is just to speed up the process, right?

lmuntaner (approved):
LGTM!! Minor comments!

src/frontend/src/lib/legacy/flows/dappsExplorer/dapps.json
sea-snake: Good that you mention this, right now it seems that the known dapps implementation is using the request origin (so not derivation origin) to find the known dapp. Some dapps in the `dapps.json` have an additional `authOrigins` array to handle cases where a derivation origin is used.

In the long run when migrating away from storing dapps in II, we should probably use the derivation origin instead so there's a single origin used for getting the logo.

src/frontend/src/lib/stores/authorization.store.ts
sea-snake: Yes exactly, since this isn't needed in this flow, it's done to speed up the process.

src/frontend/src/lib/legacy/flows/dappsExplorer/dapps.json
sea-snake: I can confirm you're right, it's indeed using `app.caffeine.ai` as derivation origin, added `authOrigins` to Caffeine entry in `dapps.json`.

PR #3392, opened 2025-10-07
chore: Prepare id.ai for SEO

aterga:
# Motivation

Google currently does not properly index id.ai. This PR adds the missing assets and meta tags to aid search engine crawlers.

# Changes

* Add OpenGraph and Twitter meta-tags.
* Add robots.txt
* Add sitemap.txt

# Tests

No new tests needed

### DIFF
--- a/src/frontend/src/app.html
+++ b/src/frontend/src/app.html
@@ -1,17 +1,34 @@
 <!doctype html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <link rel="shortcut icon" href="%sveltekit.assets%/favicon.ico" />
-    <title>Internet Identity</title>
-    <meta
-      name="viewport"
-      content="width=device-width, initial-scale=1, viewport-fit=cover"
-    />
-    <link rel="canonical" href="https://identity.internetcomputer.org/" />
-    %sveltekit.head%
-  </head>
-  <body data-sveltekit-preload-data="hover">
-    <div style="display: contents">%sveltekit.body%</div>
-  </body>
+
+<head>
+  <meta charset="utf-8" />
+  <link rel="shortcut icon" href="%sveltekit.assets%/favicon.ico" />
+  <title>Internet Identity</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover" />
+  <meta name="description"
+    content="Internet Identity is the authentication service on ICP enabling real privacy, with quick, safe and hassle-free sign-in" />
+  <meta name="robots" content="index, follow">
+  <meta name="keywords"
+    content="Internet Identity, Internet Computer, ICP, DFINITY, authentication, digital identity, blockchain identity, decentralized identity, Web3 authentication, passwordless login, biometric authentication, secure login, privacy-focused identity, decentralized authentication, blockchain security, Web3 identity, cryptographic identity" />
+  <meta property="og:title" content="Internet Identity" />
+  <meta property="og:description"
+    content="Internet Identity is the authentication service on ICP enabling real privacy, with quick, safe and hassle-free sign-in" />
+  <meta property="og:type" content="website" />
+  <meta property="og:url" content="https://id.ai/" />
+  <!-- <meta property="og:image" context="" /> -->
+  <!-- <meta name="twitter:card" content="" /> -->
+  <meta name="twitter:title" content="Internet Identity" />
+  <meta name="twitter:description"
+    content="Internet Identity is the authentication service on ICP enabling real privacy, with quick, safe and hassle-free sign-in" />
+  <!-- <meta name="twitter:image" content="" /> -->
+  <meta name="twitter:creator" content="@dfinity" />
+  <link rel="canonical" href="https://identity.internetcomputer.org/" />
+  %sveltekit.head%
+</head>
+
+<body data-sveltekit-preload-data="hover">
+  <div style="display: contents">%sveltekit.body%</div>
+</body>
+
 </html>
--- a/src/frontend/static/robots.txt
+++ b/src/frontend/static/robots.txt
@@ -1,0 +1,4 @@
+User-agent: *
+Allow: /
+Sitemap: https://id.ai/sitemap.xml
+Host: https://id.ai
--- a/src/frontend/static/sitemap.xml
+++ b/src/frontend/static/sitemap.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<urlset
+        xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
+        xmlns:news="http://www.google.com/schemas/sitemap-news/0.9"
+        xmlns:xhtml="http://www.w3.org/1999/xhtml"
+        xmlns:image="http://www.google.com/schemas/sitemap-image/1.1"
+        xmlns:video="http://www.google.com/schemas/sitemap-video/1.1"
+>
+    <url>
+        <loc>https://id.ai/</loc>
+        <changefreq>weekly</changefreq>
+        <priority>1.0</priority>
+    </url>
+    <url>
+        <loc>https://https://id.ai/login/</loc>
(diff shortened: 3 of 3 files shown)


src/frontend/src/app.html
```diff
@@ -1,17 +1,34 @@
...
+  <meta property="og:type" content="website" />
+  <meta property="og:url" content="https://id.ai/" />
+  <!-- <meta property="og:image" context="" /> -->
+  <!-- <meta name="twitter:card" content="" /> -->
+  <meta name="twitter:title" content="Internet Identity" />
+  <meta name="twitter:description"
+    content="Internet Identity is the authentication service on ICP enabling real privacy, with quick, safe and hassle-free sign-in" />
+  <!-- <meta name="twitter:image" content="" /> -->
+  <meta name="twitter:creator" content="@dfinity" />
+  <link rel="canonical" href="https://identity.internetcomputer.org/" />
```
lmuntaner: Canonical should be id.ai, no?

src/frontend/static/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
+Allow: /
+Sitemap: https://id.ai/sitemap.xml
+Host: https://id.ai
```
lmuntaner: Why use the `Host`? What does it do? I read it's deprecated.

src/frontend/static/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
+Allow: /
+Sitemap: https://id.ai/sitemap.xml
```
lmuntaner, line 3: We support multiple domains; shouldn't this point to each of them?

Or should we ignore the alternative domains.

lmuntaner (commented):
Some comments

src/frontend/src/app.html
aterga: Yep, fixed it. Thanks

src/frontend/static/robots.txt
aterga: Good catch. I took this over from NNS dapp, but there seems to be no need for this. Removing Host.

src/frontend/static/robots.txt
aterga: Good question. Here's what I understand:

1. robots.txt is domain-specific 
2. We should pick the canonical domain for which we do SEO
3. Other domains could either point to the canonical Sitemap or not provide a robots.txt, either way should be fine.

lmuntaner: Ok, makes sense to me. I'm not an expert in SEO. Did you ask someone from comms? Someone there might have pointers or tools to improve it.

aterga: I'm also not an expert, but I've asked ChatGPT about this and will try to get feedback from a human expert. I suggest we don't block on this, as right now it seems that Google doesn't index id.ai at all.

src/frontend/static/sitemap.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
peterpeterparker, line 1: Sharing if interesting given that you use svelte: you can generate the file at build time as well

e.g. OISY https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/routes/sitemap.xml/%2Bserver.ts

danblackadder (approved):

src/frontend/src/app.html
```diff
@@ -8,9 +8,35 @@
       name="viewport"
       content="width=device-width, initial-scale=1, viewport-fit=cover"
     />
-    <link rel="canonical" href="https://identity.internetcomputer.org/" />
+    <meta
+      name="description"
+      content="Internet Identity is the authentication service on ICP enabling real privacy, with quick, safe and hassle-free sign-in"
+    />
+    <meta name="robots" content="index, follow" />
```
peterpeterparker, line 20: I wonder if this is not the default?

src/frontend/src/app.html
```diff
@@ -8,9 +8,35 @@
...
+    <meta property="og:image" content="https://id.ai/social-image.png" />
+    <meta name="twitter:card" content="summary_large_image" />
+    <meta name="twitter:image" content="https://id.ai/social-image.png" />
+    <meta name="twitter:title" content="Internet Identity" />
+    <meta
+      name="twitter:description"
+      content="Internet Identity is the authentication service on ICP enabling real privacy, with quick, safe and hassle-free sign-in"
+    />
+    <meta name="twitter:creator" content="@dfinity" />
+    <link rel="canonical" href="https://id.ai/" />
```
peterpeterparker, line 41: https://id.ai or https://id.ai/ - i.e. with or without ending slash? That's an explicit choice? Probably depends of your trailing slash strategy I guess. More asking in case useful, here too, feel free to ignore.

src/frontend/static/robots.txt
peterpeterparker: I'm not an expert in SEO, but if you're using a canonical URL, it's because you explicitly want search engines to index one and only one version of your content. From that perspective, having a single sitemap makes sense to me, it aligns with the goal of consolidating everything under the canonical domain but, again, more a personal gut feeling than anything backed by any technical knowledge.

Side note: I left a comment mentioning that you can also generate different sitemap.xml files at build time with Svelte in case each domain serves different assets.

src/frontend/src/app.html
peterpeterparker: Another side note: it probably doesn’t matter much since it’s a SPA and subpages are rendered with JavaScript, but a proper canonical URL should probably correspond to each page — for example, `https://id.ai` for the landing and `https://id.ai/login` for the login page.

I should probably improve this for Oisy, the NNS dapp, and Juno as well.

Just realized that now.

Update: I provided PR in those dapps (https://github.com/junobuild/juno/pull/2035, https://github.com/dfinity/oisy-wallet/pull/9455 and https://github.com/dfinity/nns-dapp/pull/7401)

src/frontend/src/app.html
aterga: True, but let's keep it to avoid confusion.

src/frontend/static/sitemap.xml
aterga: Thanks for pointing to that example. I'm not 100% sure what the advantage would be of generating this via Svelte, could you clarify, please? 

For now, I'm going to merge this to unblock google results ASAP, but I'm happy to iterate of the changes if needed.

src/frontend/src/app.html
aterga: Re id.ai/ vs id.ai — I took the same approach that was used in NNS dapp. This should work, right? I don't mind making this consistent, but I don't want to make the call and say "id.ai/ is wrong" or "id.ai is wring"

aterga: For II, we mostly care about SEO in the context of the main page (id.ai), although if robots want to crawl to id.ai/login, there's nothing wring with it. For simplicity, I would thus suggest that we keep this static and in once place, and them if there's demand for something more advanced, we will have a per-page SEO setup.

src/frontend/static/robots.txt
aterga: Thanks for your input, @peterpeterparker . Seems like this is safe to keep as-is for now.

src/frontend/static/sitemap.xml
peterpeterparker: It's maybe more idiomatic given that's the way Svelte [documents](https://svelte.dev/docs/kit/seo#Manual-setup-Sitemaps) it and makes its generation dynamic - i.e. provide more flexibility.

src/frontend/src/app.html
peterpeterparker: > id.ai/ vs id.ai

It depends of your trailing slash strategy but, as long as it's consistent (like using the same url for canonical and og:url), I'm guessing it does not matter that much. Plus as you said, you only care at the root.  But again just a guess - i.e. I guess it's fine.

PR #3393, opened 2025-10-07
Enable new Upgrade CTA

lmuntaner:
# Motivation

Enable the changes in the Upgrade from legacy identity CTA for all users.

# Changes

* Removed the conditional block that displayed either a legacy upgrade button or a new upgrade option based on the `$AUTH_FLOW_UPDATES` variable, ensuring a single upgrade UI is shown to all users.
* Streamlined the upgrade prompt and button presentation for users with an identity number, making the interface clearer and more consistent.

# Tests

Tested locally with mobile, desktop and light and dark mode.

### DIFF
--- a/src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
+++ b/src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
@@ -133,30 +133,15 @@
         </Button>
       </Tooltip>
     {/if}
-    {#if $AUTH_FLOW_UPDATES}
-      <div class="border-border-tertiary border-t"></div>
-      <div class="flex flex-row items-center justify-between gap-4">
-        <p class="text-text-secondary text-sm">
-          Still have an identity number?
-        </p>
-        <button
-          onclick={migrate}
-          class="text-text-primary text-sm font-semibold outline-0 hover:underline focus-visible:underline"
-        >
-          Upgrade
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
```diff
@@ -63,7 +60,7 @@
   const openIdProviders = canisterConfig.openid_configs?.[0] ?? [];
 </script>
 
-<div class="flex flex-col items-stretch gap-6">
+<div class="flex flex-col items-stretch gap-5">
```
lmuntaner, line 63: Figma uses 20px as separation

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
```diff
@@ -133,30 +130,15 @@
...
-        size="xl"
-        disabled={!supportsPasskeys ||
-          nonNullish(authenticatingProviderId) ||
-          isGoogleAuthenticating}
-      >
-        Upgrade from legacy identity
-      </Button>
-    {/if}
+  </div>
+  <div class="border-border-tertiary border-t"></div>
```
lmuntaner, line 134: Moved out of the div with the buttons because it uses different spacing than between buttons.

PR #3397, opened 2025-10-08
SSG (Static Site Generated) landing page

sea-snake:
SSG (Static Site Generated) landing page and make title shorter.

# Changes

- Make sure landing page is static generated with `ssr = true`.
- Remove JS from landing page, use css instead.
- Add reroute manually back to landing page since this hook is no longer called for pages that are SSG.
- Fix legacy add remote device flow (e2e was failing after above changes)
- Make title shorter (as suggested by @peterpeterparker).

# Tests

Verified that all routes are loading as expected.

# Note

When visiting legacy (non id.ai) pages, on first non cached load there might be a few ms that the new SSG id.ai landing page is shown before the correct page is shown since the JS that reroutes based on url hasn't loaded yet.


src/frontend/src/lib/icons/virtuaseal.webp
aterga, line 1: Are we adding this logo anew?

src/frontend/src/lib/legacy/flows/dappsExplorer/dapps.json
```diff
@@ -417,7 +417,7 @@
   {
     "name": "Virtuaseal",
     "website": "https://p3cwv-2aaaa-aaaap-abwba-cai.icp0.io/",
-    "logo": "virtuaseal.png",
+    "logo": "virtuaseal.webp",
```
aterga, line 420: Why is webp preferred over png?

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -171,8 +145,13 @@
         description="Forget about remembering complicated usernames and passwords. With passkeys, you simply pick your name to log in — quick, safe, and hassle-free."
       >
         <PasswordFreeIllustration
-          colors={illustrationColours}
-          class="max-w-[264px]"
+          class={[
+            "max-w-[264px]",
```
aterga, line 149: Were these parameters inlined from somewhere?

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -228,3 +212,18 @@
   <Footer />
   <div class="h-[env(safe-area-inset-bottom)]"></div>
 </div>
+
+<style>
```
aterga, line 216: Is this an acceptable way to add styled these days? Back in the day, all styles were to be added to the head section, IIRC

src/frontend/src/lib/icons/virtuaseal.webp
sea-snake: Swapping out the png with a webp (7kb reduction)

src/frontend/src/lib/legacy/flows/dappsExplorer/dapps.json
sea-snake: Cause I couldn't get the png smaller I tried webp as alternative which seemed to work.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: No idea, I kept the CSS "as is" from the landing page, only change I made was moving to tailwind css to color the SVG gradient color stops, avoiding the need for the current JS approach.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: Yes, style tags can be added anywhere these days, only priority styles that should be loaded before everything else should be in the head. On another note, svelte automatically transforms the class name to something unique during build time to scope the css to this svelte file.

src/frontend/src/lib/components/illustrations/landing/EasyAccessIllustration.svelte
```diff
@@ -8,10 +8,14 @@
   import type { SVGAttributes } from "svelte/elements";
 
   type Props = SVGAttributes<SVGSVGElement> & {
-    colors: { start: string; end: string };
+    colors?: { start: string; end: string };
```
lmuntaner, line 11: Why is this change necessary?

svelte.config.js
```diff
@@ -11,7 +11,7 @@ const config = {
     adapter: adapter({
       pages: "dist",
       assets: "dist",
-      fallback: "index.html",
+      fallback: "200.html",
```
lmuntaner, line 14: Why this change? What's this 200.html file?

src/frontend/src/routes/(new-styling)/+page.ts
```diff
@@ -1,5 +1,2 @@
-import type { PageLoad } from "./$types";
-
-export const load: PageLoad = ({ url }) => {
-  return { next: url.searchParams.get("next") };
```
lmuntaner, line 4: This is removed and now done in `manuallyReroute`, right?

lmuntaner (approved):
Minor comments and questions, already approving.

src/frontend/src/routes/(new-styling)/+page.ts
sea-snake: This was an inert leftover remnant when login was moved to the /login route. It's removed now since it broke the SSG render.

svelte.config.js
sea-snake: It's the html file that should be ideally be served for an unknown route by the backend. Right now it's not used yet, I've renamed it to prevent it from overwriting the SSG landing page html (there was a svelte kit warning).

The filename "200.html" was mentioned in the svelte kit docs, but we can name it however as long it doesn't overlap with an existing route.

src/frontend/src/lib/components/illustrations/landing/EasyAccessIllustration.svelte
sea-snake: I'm no longer using these params to color the svg gradient, instead a tailwind class is used directly. To avoid TS errors regarding a missing prop, I've made it optional with a default value. 

The illustrations can probably refactored further, but that's out of scope in this PR.

src/frontend/src/routes/(new-styling)/+page.ts
lmuntaner: Ah, of course, thanks for the explanation!

PR #3400, opened 2025-10-08
feat: Store and serve identity creation timestamps

aterga:
# Motivation

This PR implements serving identity creation timestamps in the response of `get_anchor_info`.

< [Previous PR](https://github.com/dfinity/internet-identity/pull/3399) |

# Changes

* `Storage.allocate_anchor` takes `now` and sets `Anchor.created_at` to it.
* The callers are adjusted to pass in the current timestamp. Note that this happens only in two places:
    * `identity_registration_finish` — supports the new flow for creating new identities.
    * `register` — a very old function that will most likely soon be deleted from the API.

# Tests

* The `should_register_new_identity_with_name` (now called `should_register_new_identity_with_name_timestamp`) integration test is extended with an assertion to check that `anchor_info.created_at` is set to the timestamp of the `identity_registration_finish` call.
* Existing unit tests are adjusted to pass in a mock timestamp.
* New unit test `should_set_timestamp` added to assert that calling `allocate_anchor(T)` indeed sets the `created_at` timestamp to `T`.

### DIFF
--- a/src/internet_identity/src/anchor_management.rs
+++ b/src/internet_identity/src/anchor_management.rs
@@ -326,3 +326,19 @@ fn should_set_name() {
     );
     assert_eq!(set_name(&mut anchor, None), Ok(Operation::RemoveName));
 }
+
+#[test]
+fn should_set_timestamp() {
+    use crate::state::{storage_borrow_mut, storage_replace};
+    use crate::storage::Storage;
+    use ic_stable_structures::VectorMemory;
+
+    // Prepare the world
+    storage_replace(Storage::new((0, 10000), VectorMemory::default()));
+
+    // Run code under test
+    let anchor = storage_borrow_mut(|storage| storage.allocate_anchor(123456789).unwrap());
+
+    // Assert postcondition
+    assert_eq!(anchor.created_at(), Some(123456789));
+}


src/frontend/src/lib/generated/internet_identity_idl.js
lmuntaner, line 296: Do you still need to rebase?

src/internet_identity/src/storage.rs
```diff
@@ -538,14 +538,15 @@ impl<M: Memory + Clone> Storage<M> {
     ///
     /// Returns None if the range of Identity Anchor assigned to this
     /// storage is exhausted.
-    pub fn allocate_anchor(&mut self) -> Option<Anchor> {
+    pub fn allocate_anchor(&mut self, now: Timestamp) -> Option<Anchor> {
```
lmuntaner, line 541: Aren't there unit tests for this function? I don't see them being updated.

src/internet_identity/src/storage/anchor/tests.rs
lmuntaner, line 1: Should we add a test here that passing a timestamp to `Anchor::new` adds it as `created_at`?

src/frontend/src/lib/generated/internet_identity_idl.js
aterga: There was a conflict, resolved it now

src/internet_identity/src/storage.rs
aterga: They were updated, passing 0 as a mock timestamp. 

Since the logic of the canister doesn't depend on the value of this new field, it doesn't add much value to have unit tests for it. Sure, I'll add one to check that calling `allocate_anchor` with a non-zero timestamp results in an Anchor that has that very timestamp set. 

But the only test that we really need is an integration test that demos that a timestamp is set (and can be obtained via the API) for a newly created identity. This is being tested in should_register_new_identity_with_name_timestamp.

src/internet_identity/src/storage/anchor/tests.rs
aterga: [Ditto](https://github.com/dfinity/internet-identity/pull/3400/#discussion_r2416209128)

PR #3401, opened 2025-10-09
Remove generic OpenID feature flag

sea-snake:
Remove generic OpenID feature flag.

# Changes

- Remove `feature_flag_enable_generic_open_id_fe` canister feature flag.
- Remove `ENABLE_GENERIC_OPEN_ID` canister feature flag.
- Remove frontend implementation behind disabled feature flag, no other frontend changes have been made.
- Remove tests specific for disabled feature flag, no other test changes have been made.

# Tests

Verified that canister builds and upgrades correctly, OpenID flows are covered by e2e and integration which should pass.


src/frontend/src/lib/generated/internet_identity_types.d.ts
lmuntaner, line 1: You might need to upgrade your `didc` to revert these changes

sea-snake: Ah that was it! upgrading dfx didn't help.

PR #3402, opened 2025-10-09
A/B Test Large Google Button

sea-snake:
A/B Test Large Google Button.

# Changes

- Update feature flag implementation to make it possible to set a default value based on a callback when unset.
- Add `LARGE_GOOGLE_BUTTON` feature flag with 50/50 user groups.
- Add `abTestGroup` property to `authenticationV2Funnel.init`.
- Update `PickAuthenticationMethod` to render a large Google button if feature flag is true.

# Tests

The authentication/registration e2e tests should be unaffected since from an accessibility perspective, the buttons haven't changed. 

image

### DIFF
--- a/src/frontend/src/routes/(new-styling)/login/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/login/+page.svelte
@@ -99,7 +99,7 @@
   onMount(() => {
     authenticationV2Funnel.init({
       origin: window.location.origin,
-      abTestGroup: get(LARGE_GOOGLE_BUTTON)
+      abTestGroup: $LARGE_GOOGLE_BUTTON
         ? "largeGoogleButton"
         : "smallGoogleButton",
     });


src/frontend/src/lib/state/featureFlags.ts
```diff
@@ -92,19 +92,33 @@ export const DISCOVERABLE_PASSKEY_FLOW = createFeatureFlagStore(
...
   "AUTH_FLOW_UPDATES",
   false,
 );
 
+export const LARGE_GOOGLE_BUTTON = createFeatureFlagStore(
+  "LARGE_GOOGLE_BUTTON",
+  true,
+  (featureFlag) => {
+    // A/B Test with 50/50 user groups
+    featureFlag.set(Math.random() > 0.5);
```
lmuntaner, line 112: cool idea!!!

PR #3406, opened 2025-10-10
Fix the flow to upgrade an upgraded legacy identity

lmuntaner:
# Motivation

Upgrading again isn't working. The problem was that the element used to attach the iframe doesn't exist anymore.

# Changes

* In `AlreadyMigrated.svelte`, the `onUpgradeAgain` callback now accepts an optional `attachElement` parameter, and the root `<div>` is bound to provide this reference. The "Upgrade again" button now calls a handler that passes this element. 
* In `migrationFlow.svelte.ts`, the `upgradeAgain` method is refactored to accept an optional `attachElement` argument directly, removing the need to store it as a class property. The method now uses the passed element for authentication, simplifying the flow.
* The assignment of `attachElement` as a class property is removed from the authentication method, reflecting the new direct passing approach.

# Tests

Tested in beta domains.

### DIFF
--- a/src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
+++ b/src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
@@ -14,10 +14,6 @@
 
   let attachElement = $state<HTMLElement>();
 
-  const handleUpgradeAgain = () => {
-    onUpgradeAgain(attachElement);
-  };
-
   onMount(() => {
     upgradeIdentityFunnel.trigger(UpgradeIdentityEvents.AlreadyMigratedScreen);
   });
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/wizards/migration/views/AlreadyMigrated.svelte
```diff
@@ -8,14 +8,22 @@
...
   } from "$lib/utils/analytics/upgradeIdentityFunnel";
 
-  let { onUpgradeAgain }: { onUpgradeAgain: () => void } = $props();
+  let {
+    onUpgradeAgain,
+  }: { onUpgradeAgain: (attachElement?: HTMLElement) => void } = $props();
+
+  let attachElement = $state<HTMLElement>();
+
+  const handleUpgradeAgain = () => {
```
aterga: Can this function be inlined?

lmuntaner: Yes, do you think that would be cleaner?

lmuntaner: Done!

aterga: Thanks - not sure, just trying to understand which function does what, it's a bit easier to do if there are fewer functions.

PR #3407, opened 2025-10-10
Add multi language lib

sea-snake:
Add [lingui](https://lingui.dev/) multi language lib.

Alternatives considered
- i18next: too large and (cli) incompatible with svelte
- formatjs: more complex and (cli) incompatible with svelte

Fortunately for lingui there was an up-to-date 3rd party lib [svelte-i18n-lingui](https://www.npmjs.com/package/svelte-i18n-lingui) that adds support for Svelte. After further inspection, the implementation was minimal enough to copy it over into II directly, keeping only lingui itself as dependency.

# Generated ids and extraction

Multi language text in svelte is defined as e.g. `` $t`Hello world` ``. Run `npm run extract` to update all `.po` locale files with latest strings found in the codebase. Changes to `.po` files are immediately picked up by a vite plugin.

# Changes

- Add lingui dependencies
- Copy over and cleanup minimal code needed to integrate with svelte
- Implement localeStore to keep track of locale state and switch locales

# Tests

Existing e2e tests should pass as is.

### DIFF
--- a/LOCALIZING.md
+++ b/LOCALIZING.md
@@ -10,7 +10,16 @@ Make sure Node.js has been installed and this project dependencies have been ins
 npm ci
 ```
 
-## Extracting
+## Adding languages
+
+To add another language, add another IETF BCP 47 language tag to the `availableLocales` variable in
+`src/frontend/src/lib/constants/locale.constants.ts`.
+
+After this, you can follow below [Updating translations](#Updating-translations) instructions.
+
+## Updating translations
+
+### Extracting
 
 First update the locale catalogs with the latest changes from the frontend implementation:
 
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/stores/locale.store.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (!availableLocales.includes(locale)) {
+      // Return if locale isn't available
+      return;
+    }
+    const { messages } = await import(`$lib/locales/${locale}.po`);
+    i18n.loadAndActivate({ locale: locale, messages });
+    internalStore.set(locale);
+  },
+  reset: () => {
+    // const { messages } = await import(
```
aterga: Should the comment be removed?

sea-snake: Yes, it has been removed.

package.json
```diff
@@ -59,8 +63,9 @@
     "astro": "^5.5.4",
     "devtools-protocol": "^0.0.1502260",
     "eslint": "^9.18.0",
-    "eslint-config-prettier": "^10.0.1",
+    "eslint-config-prettier": "^10.1.8",
```
sea-snake, line 66: Saw that this version had a vulnerability warning, so I've updated it to the latest.

src/frontend/src/lib/stores/locale.store.ts
lmuntaner, line 1: Could we add some tests for this?

It would be nice to make sure we don't break functionality in the future.

src/lingui-svelte/generateMessageId.ts
```diff
@@ -0,0 +1,17 @@
+import { sha256 } from "js-sha256";
+// noinspection ES6PreferShortImport
+import { toBase64 } from "../frontend/src/lib/utils/utils";
+
+const UNIT_SEPARATOR = "\u001F";
+
+// Generates a short, deterministic ID by hashing (message + context),
+// converting the SHA-256 hex digest to bytes, encoding as Base64, and
+// taking the first 6 chars for a compact identifier.
```
aterga: This reminds me of something. :-) 

Why truncate the SHA?

src/lingui-svelte/extractor.ts
aterga, line 1: Could you summarize why this file needed to be added in this PR, please?

src/frontend/src/lib/stores/locale.store.ts
sea-snake: Good idea, I'll add some.

src/frontend/src/lib/stores/locale.store.ts
```diff
@@ -0,0 +1,114 @@
...
+  defaultValue: null,
+  version: 1,
+});
+
+export const localeStore: LocaleStore = {
+  init: async () => {
+    const locale = get(internalStore) ?? availableBrowserLocale;
+    if (availableLocales[0] !== locale && availableLocales.includes(locale)) {
+      // Load locale if not default and it's available
+      const { messages } = await import(`$lib/locales/${locale}.po`);
```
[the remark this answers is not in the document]
sea-snake: This code runs in the browser and the input is coming from the browser language settings...

src/frontend/src/lib/stores/locale.store.ts
```diff
@@ -0,0 +1,114 @@
...
+  subscribe: derived(
+    [internalStore],
+    ([locale]) => locale ?? availableBrowserLocale,
+  ).subscribe,
+  set: async (locale: string) => {
+    if (!availableLocales.includes(locale)) {
+      // Return if locale isn't available
+      return;
+    }
+    const { messages } = await import(`$lib/locales/${locale}.po`);
```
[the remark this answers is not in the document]
sea-snake: See other reply

src/frontend/src/lib/stores/locale.store.ts
```diff
@@ -0,0 +1,114 @@
...
+    args.forEach((_arg, i) => {
+      message += `{${i}}` + descriptor[i + 1];
+    });
+    const id = generateMessageId(message);
+    const values = { ...args };
+
+    return i18n.t({ id, message, values });
+  }
+
+  throw new Error("Unknown descriptor");
```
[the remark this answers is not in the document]
sea-snake: This shouldn't be reachable to begin with if frontend code follows TS types.

src/lingui-svelte/generateMessageId.ts
```diff
@@ -0,0 +1,17 @@
...
+export const generateMessageId = (message: string, context = "") =>
+  toBase64(
+    new Uint8Array(
+      sha256(message + UNIT_SEPARATOR + context)
+        .match(/.{1,2}/g)!
```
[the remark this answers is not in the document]
sea-snake: It's safe since we're sure that the sha-256 won't be malformed.

src/lingui-svelte/generateMessageId.ts
sea-snake: To keep language filesize to a minimum the ids are truncated intentionally, but they're kept long enough to avoid conflicts. Compared to the canister hashes, these aren't based on user submitted data, so a collisions isn't feasible unless we intentionally add some weird text labels to our frontend.

vite.config.ts
```diff
@@ -5,7 +5,7 @@ import { sveltekit } from "@sveltejs/kit/vite";
 import basicSsl from "@vitejs/plugin-basic-ssl";
 import { type AliasOptions, type UserConfig, defineConfig } from "vite";
 import { nodePolyfills } from "vite-plugin-node-polyfills";
-import path from "path";
+import { lingui } from "@lingui/vite-plugin";
```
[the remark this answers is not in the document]
sea-snake: It was unused...

src/lingui-svelte/extractor.ts
sea-snake: https://lingui.dev/guides/message-extraction

Instead of defining explicit translation keys, we rely on message extraction to generate the translation catalogs based on the labels used within the implementation.  

Lingui does not offer a first party extractor implementation for svelte but there's a 3rd party implementation, this file is a copy of this implementation to avoid adding the 3rd party lib as dependency.

Without the extraction implementation, messages cannot be extracted, thus there isn't a `.po` file, without said file the frontend won't compile, so couldn't move it to a later PR.

src/frontend/src/lib/stores/locale.store.ts
sea-snake: Extensive tests have been added to all relevant parts.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -25,6 +25,7 @@
   import { page } from "$app/state";
   import { sessionStore } from "$lib/stores/session.store";
   import AuthorizeError from "$lib/components/views/AuthorizeError.svelte";
+  import { t } from "$lib/stores/locale.store";
```
lmuntaner: Is this import used?

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -19,6 +19,7 @@
   } from "$lib/config";
   import LandingHeader from "$lib/components/layout/LandingHeader.svelte";
   import { manuallyReroute } from "../../hooks";
+  import { t } from "$lib/stores/locale.store";
```
lmuntaner, line 22: The new component `Trans` shuold be used only to support nested tags, right?

package.json
```diff
@@ -34,14 +34,18 @@
     "lint:eslint": "eslint --max-warnings 0 --cache --cache-location node_modules/.cache/eslint 'src/frontend/**/*.ts*' 'src/showcase/**/*.ts'",
     "lint:lit": "lit-analyzer -- --rules.no-incompatible-type-binding off --rules.no-invalid-directive-binding off --rules.no-noncallable-event-binding off ./src/showcase ./src/frontend",
     "format": "prettier --write src/showcase src/frontend src/vc-api src/sig-verifier-js src/vite-plugins tsconfig.json tsconfig.all.json eslint.config.js vite.config.ts vitest.config.ts demos",
-    "format-check": "prettier --check src/showcase src/frontend src/vc-api src/sig-verifier-js src/vite-plugins tsconfig.json tsconfig.all.json eslint.config.js vite.config.ts vitest.config.ts demos"
+    "format-check": "prettier --check src/showcase src/frontend src/vc-api src/sig-verifier-js src/vite-plugins tsconfig.json tsconfig.all.json eslint.config.js vite.config.ts vitest.config.ts demos",
+    "extract": "lingui extract --clean",
+    "compile": "echo \"Does nothing, compilation is handled by the Vite plugin\""
```
lmuntaner, line 39: Do we need this script?

src/lingui-svelte/extractor.test.ts
lmuntaner, line 1: Shouldn't these files live inside `frontend`? They are tightly coupled with it, right? Maybe `frontend/scripts/lingui-svelte`?

LOCALIZING.md
lmuntaner, line 1: Should we add here that there is the `Trans` component and where to use it? Same about using `t`, where to import it and how to use it.

src/lingui-svelte/extractor.test.ts
sea-snake: Technically, it's only used in vite (preprocessor) and npm scripts (extractor), frontend isn't using any of these js files directly anymore.

package.json
sea-snake: No, added it since it would be confusing not to have this message when a developer sees the Lingui console message mentioning `npm run compile` after they just ran `npm run extract`.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: Yes, wherever possible, either`$t` or `$plural` should be the preferred choice.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
sea-snake: Good catch, it's not, removed it.

LOCALIZING.md
sea-snake: Good idea, I'll add these sections.

package.json
lmuntaner: Thanks for the explanation!

src/lingui-svelte/extractor.test.ts
lmuntaner: Kind of weird that frontend isn't using those directly, in that case ok.

LOCALIZING.md
sea-snake: @lmuntaner Added a rather extensive section that should cover most if not all things.

src/lingui-svelte/extractor.test.ts
sea-snake: It's the magic of the preprocessor :D

PR #3414, opened 2025-10-14
Add optional createdAt property to last used identities

lmuntaner:
# Motivation

We want to add when the identity was created in the identity switcher.

In this PR, I add a new property to the data stored in local storage that is then used in the identity switcher.

# Changes

* Added an optional `created_at` field to the `AuthnMethodSessionInfo` type in the backend interface and ensured it is populated from the anchor's creation time.
* Updated `AuthFlow` and `RegisterAccessMethodFlow` classes to extract, convert (using the new `nanosToMillis` utility), and pass `createdAtMillis` when handling identity and authentication method information.
* For legacy identities or flows where `created_at` is not available, set `createdAtMillis` to `undefined`.
* Extended the `LastUsedIdentity` type and the `lastUsedIdentitiesStore` to include the optional `createdAtMillis` field and updated the store's API accordingly. 
* Added a new `nanosToMillis` utility function for converting nanosecond timestamps to milliseconds.

# Tests

* Tested locally that the new property is added to the local storage with the right units. See screenshot attached.
* Updated tests for `lastUsedIdentitiesStore` and `lastUsedIdentityStore` to include checks for the new `createdAtMillis` property, ensuring correct behavior and coverage.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -1022,9 +1022,12 @@ mod v2_api {
         // Return session info if caller matches confirmed session
         tentative_device_registration::get_confirmed_session(identity_number)
             .is_some_and(|confirmed_session| confirmed_session == caller())
-            .then_some(AuthnMethodSessionInfo {
-                name: state::anchor(identity_number).name(),
-                created_at: state::anchor(identity_number).created_at(),
+            .then(|| {
+                let anchor = state::anchor(identity_number);
+                AuthnMethodSessionInfo {
+                    name: anchor.name(),
+                    created_at: anchor.created_at(),
+                }
             })
     }
 


src/internet_identity/src/main.rs
```diff
@@ -1024,6 +1024,7 @@ mod v2_api {
             .is_some_and(|confirmed_session| confirmed_session == caller())
             .then_some(AuthnMethodSessionInfo {
                 name: state::anchor(identity_number).name(),
+                created_at: state::anchor(identity_number).created_at(),
```
sea-snake: Not sure if calling state::anchor() twice results in reading the anchor twice unnecessarily.

If this is the case, probably should call it once and assign it to a variable first.

PR #3415, opened 2025-10-14
Hide SSG page

sea-snake:
Hide SSG page, compared to the previous PR, this implementation is within a single file (root layout) and also handles legacy `/#fragment` routes.

# Tests

The issue only happens though the SSG landing page, so need to be verified on a local production build instead of dev build.

### DIFF
--- a/src/frontend/src/routes/+layout.svelte
+++ b/src/frontend/src/routes/+layout.svelte
@@ -23,7 +23,7 @@
       }
 
       html[data-temp-hide-ssg="legacy"] body {
-        /* Make sure page background matches legacy */
+        /* Make sure page background matches legacy (--nvc-surface-dark in src/frontend/src/lib/legacy/styles/main.css) */
         background: oklch(0.15 0.0236 261.52) !important;
       }
     </style>


src/frontend/src/routes/+layout.svelte
```diff
@@ -2,16 +2,46 @@
...
 </script>
 
+<svelte:head>
+  {#if page.url.pathname === "/"}
+    <style>
+      html[data-temp-hide-ssg] body > div {
+        display: none !important;
+      }
+
+      html[data-temp-hide-ssg="legacy"] body {
```
lmuntaner, line 25: Nice idea!

src/frontend/src/routes/+layout.svelte
```diff
@@ -2,16 +2,46 @@
...
+        /* Make sure page background matches legacy */
+        background: oklch(0.15 0.0236 261.52) !important;
```
lmuntaner, line 27: Can we add a comment on how you obtained this color, in case we need to change it?

It's temporary, so you don't need to have it coupled, but the reference could be useful if more hacks are needed here.

sea-snake: Sure, updated the comment to refer to the color variable in legacy main.css

PR #3419, opened 2025-10-16
Remove continue from another device feature flag

sea-snake:
Remove continue from another device feature flag.

# Changes

- Remove feature flag in frontend.
- Remove feature flag in backend.
- Remove old frontend implementation (when flag was set to false).
- Remove backend checks that trapped (when flag was set to false).

# Tests

Verified that build completes, checked that the flow still works as expected, all e2e tests should pass.


lmuntaner (approved):
Thanks!

PR #3427, opened 2025-10-20
Merge authorize screens and implement enable multiple accounts

sea-snake:
Merge authorize screens and implement enable multiple accounts.

# Changes

- Remove `authorize/account` route.
- Update `authorize/continue` route with "Enable multiple account" flow.
- Update routing to navigate to `authorize/continue` after account creation (instead of directly authorizing).
- Update `--bg-quaternary` color.
- Add `--bg-quaternary_hover` color.
- Update `AuthorizeHeader` padding.
- Add label and hint props to `Checkbox` component.
- Implement `Toggle` component based on existing `Checkbox` component.
- Update `Input` component to also accept snippets as hint and error.
- Update `Tooltip` component to accept optional offset prop.
- Update `CreateAccount` component based on updated design.
- Add `EditAccount` component based on the above.
- Move analytics to `authLastUsedFlow`. 

# Tests

Updated all e2e tests to work with new flow and changed accounts e2e tests to cover cases in new flow.

image
image
image
image

### DIFF
--- a/demos/test-app/package.json
+++ b/demos/test-app/package.json
@@ -25,7 +25,7 @@
     "@types/react-dom": "^18.2.17",
     "@vitejs/plugin-react": "^4.2.0",
     "typescript": "5.2.2",
-    "vite": "^6.3.6",
+    "vite": "^6.4.1",
     "vite-plugin-node-polyfills": "^0.23.0"
   }
 }
--- a/demos/vc_issuer/package.json
+++ b/demos/vc_issuer/package.json
@@ -26,7 +26,7 @@
     "@types/react-dom": "^18.2.17",
     "@vitejs/plugin-react": "^4.2.0",
     "typescript": "5.2.2",
-    "vite": "^6.3.6",
+    "vite": "^6.4.1",
     "vite-plugin-node-polyfills": "^0.23.0"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "@eslint/js": "^9.18.0",
         "@lingui/cli": "^5.5.1",
         "@lingui/vite-plugin": "^5.5.1",
-        "@playwright/test": "^1.52.0",
+        "@playwright/test": "^1.56.1",
         "@sveltejs/adapter-static": "^3.0.8",
         "@sveltejs/kit": "^2.20.6",
         "@sveltejs/vite-plugin-svelte": "^5.0.0",
@@ -82,7 +82,7 @@
         "ts-node": "^10.9.2",
         "typescript": "5.8.2",
         "typescript-eslint": "^8.20.0",
-        "vite": "^6.3.6",
+        "vite": "^6.4.1",
         "vite-plugin-node-polyfills": "^0.23.0",
         "vitest": "^3.0.9",
         "webdriverio": "^8.43.0"
@@ -111,7 +111,7 @@
         "@types/react-dom": "^18.2.17",
         "@vitejs/plugin-react": "^4.2.0",
         "typescript": "5.2.2",
-        "vite": "^6.3.6",
+        "vite": "^6.4.1",
         "vite-plugin-node-polyfills": "^0.23.0"
       }
     },
@@ -221,7 +221,7 @@
         "@types/react-dom": "^18.2.17",
         "@vitejs/plugin-react": "^4.2.0",
         "typescript": "5.2.2",
-        "vite": "^6.3.6",
+        "vite": "^6.4.1",
         "vite-plugin-node-polyfills": "^0.23.0"
       }
     },
@@ -3076,13 +3076,13 @@
       }
     },
     "node_modules/@playwright/test": {
-      "version": "1.52.0",
-      "resolved": "https://registry.npmjs.org/@playwright/test/-/test-1.52.0.tgz",
-      "integrity": "sha512-uh6W7sb55hl7D6vsAeA+V2p5JnlAqzhqFyF0VcJkKZXkgnFcVG9PziERRHQfPLfNGx1C292a4JqbWzhR8L4R1g==",
+      "version": "1.56.1",
+      "resolved": "https://registry.npmjs.org/@playwright/test/-/test-1.56.1.tgz",
+      "integrity": "sha512-vSMYtL/zOcFpvJCW71Q/OEGQb7KYBPAdKh35WNSkaZA75JlAO8ED8UN6GUNTm3drWomcbcqRPFqQbLae8yBTdg==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
-        "playwright": "1.52.0"
+        "playwright": "1.56.1"
       },
       "bin": {
         "playwright": "cli.js"
@@ -13687,13 +13687,13 @@
       }
     },
     "node_modules/playwright": {
-      "version": "1.52.0",
-      "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.52.0.tgz",
-      "integrity": "sha512-JAwMNMBlxJ2oD1kce4KPtMkDeKGHQstdpFPcPH3maElAXon/QZeTvtsfXmTMRyO9TslfoYOXkSsvao2nE1ilTw==",
+      "version": "1.56.1",
+      "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.56.1.tgz",
+      "integrity": "sha512-aFi5B0WovBHTEvpM3DzXTUaeN6eN0qWnTkKx4NQaH4Wvcmc153PdaY2UBdSYKaGYw+UyWXSVyxDUg5DoPEttjw==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
-        "playwright-core": "1.52.0"
+        "playwright-core": "1.56.1"
       },
       "bin": {
         "playwright": "cli.js"
@@ -13706,9 +13706,9 @@
       }
     },
     "node_modules/playwright-core": {
-      "version": "1.52.0",
-      "resolved": "https://registry.npmjs.org/playwright-core/-/playwright-core-1.52.0.tgz",
-      "integrity": "sha512-l2osTgLXSMeuLZOML9qYODUQoPPnUsKsb5/P6LJ2e6uPKXUdPK5WYhN4z03G+YNbWmGDY4YENauNu4ZKczreHg==",
+      "version": "1.56.1",
+      "resolved": "https://registry.npmjs.org/playwright-core/-/playwright-core-1.56.1.tgz",
+      "integrity": "sha512-hutraynyn31F+Bifme+Ps9Vq59hKuUCz7H1kDOcBs+2oGguKkWTU50bBWrtz34OUWmIwpBTWDxaRPXrIXkgvmQ==",
       "dev": true,
       "license": "Apache-2.0",
       "bin": {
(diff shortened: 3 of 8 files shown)


src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,16 +1,17 @@
...
   import { nonNullish } from "@dfinity/utils";
+  import type { Snippet } from "svelte";
 
   type Size = "sm" | "md";
 
   type Props = Omit<HTMLInputAttributes, "size"> & {
     element?: HTMLInputElement;
     label?: string;
-    hint?: string;
+    hint?: string | Snippet;
```
lmuntaner, line 11: Can't a string be used as a snippet?

src/frontend/src/lib/components/views/CreateAccount.svelte
```diff
@@ -1,24 +1,34 @@
...
+  import { PlusIcon } from "@lucide/svelte";
   import Button from "$lib/components/ui/Button.svelte";
   import Input from "$lib/components/ui/Input.svelte";
+  import { t } from "$lib/stores/locale.store";
+  import Checkbox from "$lib/components/ui/Checkbox.svelte";
   import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
 
   interface Props {
-    create: (name: string) => void;
+    namesInUse: string[];
```
lmuntaner: It's a bit confusing, `namesInUse` and `isNameInUse`. How about: `existingNames`, `nameExists`?

src/frontend/src/lib/components/views/CreateAccount.svelte
```diff
@@ -29,28 +39,37 @@
 <form class="flex flex-1 flex-col">
```
lmuntaner: Further refactor to combine this form with the EditAccount one?

src/frontend/src/lib/components/views/EditAccount.svelte
```diff
@@ -0,0 +1,107 @@
...
+  import Button from "$lib/components/ui/Button.svelte";
+  import Input from "$lib/components/ui/Input.svelte";
+  import { t } from "$lib/stores/locale.store";
+  import Checkbox from "$lib/components/ui/Checkbox.svelte";
+  import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
+
+  interface Props {
+    name: string;
+    isDefaultSignIn: boolean;
+    namesInUse: string[];
```
lmuntaner: Ditto...

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -48,6 +52,9 @@ export class AuthLastUsedFlow {
         });
         await authenticationStore.set({ identity, identityNumber });
         lastUsedIdentitiesStore.addLastUsedIdentity(lastUsedIdentity);
+        authenticationV2Funnel.trigger(
+          AuthenticationV2Events.ContinueAsPasskey,
```
lmuntaner, line 56: Wasn't this triggered anywhere else?

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -82,6 +89,10 @@ export class AuthLastUsedFlow {
         });
         await authenticationStore.set({ identity, identityNumber });
         lastUsedIdentitiesStore.addLastUsedIdentity(lastUsedIdentity);
+        authenticationV2Funnel.addProperties({
```
lmuntaner, line 92: ditto, wasn't this triggered aywhere else?

src/frontend/src/lib/stores/last-used-identities.store.ts
```diff
@@ -129,10 +130,7 @@ export const initLastUsedIdentitiesStore = (): LastUsedIdentitiesStore => {
         return lastUsedIdentities;
       });
     },
-    // TODO: Update this method once we store usage timestamps in the canister,
-    //       additionally the tests should be updated to include this method.
     syncLastUsedAccounts: (identityNumber, origin, accounts) => {
```
lmuntaner, line 133: Weren't there tests for this store?

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
```diff
@@ -1,146 +1,401 @@
...
   </h1>
   <p class="text-text-secondary mb-6 self-start text-sm">
-    <span>for</span>
-    {#if nonNullish(dapp?.name)}
-      <span><b>{dapp.name}</b></span>
+    {$t`with your Internet Identity`}
+  </p>
+  <div class="grid">
+    {#if isMultipleAccountsEnabled}
+      <div class="col-start-1 row-start-1" out:fade={{ duration: 100 }}>
```
lmuntaner: Could we put this and the else into snippets to help readability?

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
```diff
@@ -1,146 +1,401 @@
...
+      const { effectiveOrigin } = $authorizationContextStore;
+      const values = await Promise.all([
+        actor
+          .get_accounts(identityNumber, effectiveOrigin)
+          .then(throwCanisterError),
+        actor
+          .get_default_account(identityNumber, effectiveOrigin)
+          .then(throwCanisterError),
+      ]);
+      accounts = values[0].sort((a, b) => {
```
lmuntaner, line 126: `sort` sorts the array in place, doesn't it? Does this work to assign it to the variable?

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
```diff
@@ -1,146 +1,401 @@
...
+    }
+  };
+  const handleCreateAccount = async (
+    name: string,
+    isDefaultSignIn: boolean,
+  ) => {
+    try {
+      const { identityNumber, actor } = $authenticationStore!;
+      const createdAccount = await actor
+        .create_account(
```
lmuntaner, line 146: Shouldn't we update the `create_account` to be able to create a default in one call?

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
```diff
@@ -1,146 +1,401 @@
...
-        authenticationV2Funnel.trigger(AuthenticationV2Events.ContinueAsOpenID);
+      if (!$isAuthenticatedStore) {
+        await authLastUsedFlow.authenticate($lastUsedIdentitiesStore.selected!);
       }
-      lastUsedIdentitiesStore.addLastUsedAccount(account);
-      await authorizationStore.authorize(account.accountNumber);
+      if (defaultAccountNumber === null) {
+        const { identityNumber, actor } = $authenticationStore!;
+        const { effectiveOrigin } = $authorizationContextStore;
+        defaultAccountNumber = (
```
lmuntaner, line 91: What would happen if the user changes the default in another device and then uses Continue in another where it had one account cached? Would it log in with the default account?

lmuntaner (commented):
Long but good PR!

Just one important questions and some nitpicks.

src/frontend/src/lib/components/ui/Input.svelte
sea-snake: In React you can indeed do this but unfortunately not in Svelte :(

src/frontend/src/lib/components/views/CreateAccount.svelte
sea-snake: Makes sense, will update.

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
sea-snake: Ah yes, it's trigger on `/login`, I'll remove it there.

src/frontend/src/lib/stores/last-used-identities.store.ts
sea-snake: Apparently not for this specific method, I'll add this method to the tests.

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
sea-snake: Good idea!

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
sea-snake: It does both, which makes it a common foot gun (modifies array in place unexpectedly).

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
sea-snake: Same applies to edit_account. For simplicities sake, we decided to keep default account methods (get/set) and accounts methods (get/set/delete) separate. 

Back when we discussed this, it was unclear how the edit_account method should behave when default account was set to false, which account becomes default? So instead we opted to isolate the concepts of updating individual accounts and information across accounts (e.g. default account).

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
sea-snake: We're not using the cached data anymore (only data directly from the canister), so it should always be the latest default account on both devices.

Nonetheless I've added a $effect to at least keep the local data in sync somewhat if we end up going back to using local data at a later point (so the local data doesn't completely become stale).

lmuntaner: Isn't that confusing then? the user might see on account in the screen and log in with another, can't they?

src/frontend/src/lib/components/views/CreateAccount.svelte
sea-snake: Done!

src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
sea-snake: As discussed, the login screen is no longer showing account names unless "Enable multiple accounts" is toggled. This means by default, it will only show a "Continue" button without any account label or other details.

I'll add screenshots.

src/frontend/src/routes/(new-styling)/login/+page.svelte
```diff
@@ -78,20 +78,6 @@
 
   const handleContinueAs = async (identity: LastUsedIdentity) => {
     await authLastUsedFlow.authenticate(identity).catch(handleError);
-    if ("passkey" in identity.authMethod) {
-      authenticationV2Funnel.trigger(AuthenticationV2Events.ContinueAsPasskey);
```
lmuntaner, line 82: I can't find where this is triggered now...

lmuntaner (approved):
Approving, but I added a comment about the analytics. Could you make sure they are triggered? Thanks!

src/frontend/src/routes/(new-styling)/login/+page.svelte
sea-snake: The line above it call the `authLastUsedFlow.authenticate` method:
```
await authLastUsedFlow.authenticate(identity).catch(handleError);
```

Within this flow, it's now triggering `AuthenticationV2Events.ContinueAsPasskey` instead. This brings it in line with other flows, where most events are triggered within the flow instead of the component/page.

PR #3430, opened 2025-10-21
Fix fetch certs interval

sea-snake:
As observed by @peterpeterparker,  earlier retry delays only applied to the first outcall error instead of any failed future outcalls.

Thanks @peterpeterparker for catching this one.

# Changes

- Switch around meaning of `delay ` argument in `schedule_fetch_certs`, `None` now default to `FETCH_CERTS_INTERVAL` instead of `0`.
- When outcall **succeeds**, `schedule_fetch_certs` is called with `None`:
   - Will default to `FETCH_CERTS_INTERVAL` with timer.
   - Will use `60` as retry delay.
- When outcall **fails**,  `schedule_fetch_certs` is called with `Some`:
   - Will use `delay` with timer.
   - Will use `min(FETCH_CERTS_INTERVAL, max(60, delay * 2))` as retry delay.

# Tests

Added `should_compute_next_certs_fetch_delay` test to make sure delay is computed correctly.

### DIFF
--- a/src/internet_identity/src/openid/generic.rs
+++ b/src/internet_identity/src/openid/generic.rs
@@ -4,7 +4,6 @@ use super::{
 };
 use crate::openid::OpenIdCredential;
 use crate::openid::OpenIdProvider;
-use crate::openid::MINUTE_NS;
 use crate::secs_to_nanos;
 use base64::prelude::BASE64_URL_SAFE_NO_PAD;
 use base64::Engine;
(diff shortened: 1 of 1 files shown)


src/internet_identity/src/openid/generic.rs
```diff
@@ -177,6 +176,29 @@ impl Provider {
     }
 }
 
+fn compute_next_certs_fetch_delay<T, E>(
+    result: &Result<T, E>,
+    current_delay: Option<u64>,
+) -> Option<u64> {
+    const MIN_DELAY: u64 = 60;
```
lmuntaner: Now that we are refactoring, could we add the units? `MIN_DELAY_SECONDS` and `FETCH_CERTS_INTERVAL_SECONDS`, so on...

src/internet_identity/src/openid/generic.rs
```diff
@@ -641,3 +661,40 @@ fn should_return_error_when_name_too_long() {
         ))
     );
 }
+#[test]
+fn should_compute_next_certs_fetch_delay() {
+    const MIN_DELAY: u64 = 60;
```
lmuntaner: Ditton adding units

src/internet_identity/src/openid/generic.rs
sea-snake: Good idea!

lmuntaner (approved):
Thanks!

PR #3432, opened 2025-10-21
Make remaining auth flow and related views localized

sea-snake:
Make remaining auth flow and related views localized.

# Changes

- Localize all auth flow and related views.
- Create `UpgradePasskey` view from `CreatePasskey`.
   - Separates the identity creation flows form the identity migration flows.
   - Avoids breaking (e2e) changes in the migration flow at a later point.
   - Allows for nuanced differences in localized labels e.g. "Upgrade identity" button.
- Updated identity creation labels based on latest design
- Nit: update `size="1.25rem"` icon syntax I came across in various places with `class="size-4"` for more clarity regarding units and icon sizes.
- Updated e2e tests with latest localized labels

# Tests

- Manually verified that all e2e tests are passing.

### DIFF
--- a/svelte.config.js
+++ b/svelte.config.js
@@ -26,9 +26,6 @@ const config = {
       },
     },
     version: { name: packagejson.version },
-    output: {
-      bundleStrategy: "single",
-    },
   },
 };
 
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -7,6 +7,7 @@ import { type AliasOptions, type UserConfig, defineConfig } from "vite";
 import { nodePolyfills } from "vite-plugin-node-polyfills";
 import { lingui } from "@lingui/vite-plugin";
 import { sveltePreprocessor } from "./src/lingui-svelte";
+import { basename, extname } from "path";
 
 export const aliasConfig: AliasOptions = {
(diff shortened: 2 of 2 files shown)


src/frontend/src/lib/components/wizards/auth/views/CreateIdentity.svelte
```diff
@@ -74,10 +79,19 @@
...
       {:else}
-        <span>Create identity</span>
+        <span>{$t`Create identity`}</span>
       {/if}
     </Button>
   </div>
 </form>
+
+<style>
+  @media (max-height: 700px) {
```
lmuntaner, line 91: Why is this needed?

lmuntaner (approved):
LGTM! A bit long but we know what the priorities are...

src/frontend/src/lib/components/wizards/auth/views/CreateIdentity.svelte
sea-snake: There's not enough vertical space on mobile when the keyboard is shown, thus we hide the illustration when we're about to run out of vertical space.

PR #3439, opened 2025-10-23
Fetch default account during authorization

sea-snake:
Fetch default account during authorization, this avoid the visible delay after clicking "Continue".

# Changes

- Update authorization store to accept `Promise<bigint | undefined>` besides `bigint | undefined` as account number.
- Use the above to fetch account number during authorization so that the "Redirecting to app..." screen can be shown immediately.

# Tests

Verified that authorization works as expected and there's no more visible delay.


lmuntaner (approved):
Nice!

PR #3441, opened 2025-10-23
Improve sign in speed and loading feedback

sea-snake:
Improve sign in speed and loading feedback.

# Changes

- Show loading indicator if loading accounts takes longer than 300ms.
- Remove `LazyHttpAgent.createLazy` since `HttpAgent.createSync` now implements the same (lazily fetch root key when needed) in latest agent-js versions.
- Add `void agent.fetchSubnetKeys(canisterId)` for all agent instances, this avoids having to fetch these later when making query calls on user interaction during sign in. This speeds up the sign-in flow by a few noticeable ms.

# Tests

- Verified that authentication still works as expected with default and additional accounts.

### DIFF
--- a/src/frontend/src/lib/stores/session.store.ts
+++ b/src/frontend/src/lib/stores/session.store.ts
@@ -13,7 +13,6 @@ import { Principal } from "@icp-sdk/core/principal";
 import { idlFactory as internet_identity_idl } from "$lib/generated/internet_identity_idl";
 import { fromBase64, toBase64 } from "$lib/utils/utils";
 import { isNullish } from "@dfinity/utils";
-import { anonymousAgent, canisterId } from "$lib/globals";
 
 export interface Session {
   identity: SignIdentity;


src/frontend/src/lib/stores/authentication.store.ts
```diff
@@ -40,7 +39,10 @@ const internalStore = writable<{
 
 export const authenticationStore: AuthenticationStore = {
   init: ({ canisterId, agentOptions }) => {
-    const agent = LazyHttpAgent.createLazy(agentOptions);
+    const agent = HttpAgent.createSync(agentOptions);
+    // Fetch subnet keys to speed up queries during authentication,
+    // this avoids having to fetch them later on user interaction.
+    void agent.fetchSubnetKeys(canisterId);
```
lmuntaner, line 45: I didn't know about this!

sea-snake: Noticed the extra calls when I was inspecting the waterfall in the network tab since I knew I was making query calls but still felt things were a bit slower than they should be.

PR #3446, opened 2025-10-27
Disable continue button and toggle on default account authentication.

sea-snake:
Disable continue button and toggle on default account authentication.

# Tests

Verified that button and toggle are disabled on authentication (passkey prompt or openid popup) and enabled again when e.g. passkey prompt is dismissed.

# Notes

No "authenticating" UI state was added for selecting an individual account since passkey/openid authentication already happened earlier when "enable multiple accounts" was toggled, so clicking these accounts already instantly shows the "Redirecting to the app" screen.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
@@ -396,7 +396,7 @@
         : handleEnableMultipleAccounts}
       label={$t`Enable multiple accounts`}
       size="sm"
-      disabled={true || isAuthenticatingDefault}
+      disabled={isAuthenticatingDefault}
     />
     <Tooltip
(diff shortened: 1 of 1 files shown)


src/frontend/src/routes/(new-styling)/authorize/(panel)/continue/+page.svelte
```diff
@@ -102,6 +104,7 @@
           : defaultAccountNumber,
       );
     } catch (error) {
+      isAuthenticatingDefault = false;
```
lmuntaner: Why not in the `finally`?? The page now redirects on success, but who knows in the future, we wouldn't want the state to stay `true` after authenticating...

sea-snake: Ah good catch! I'll make a change.

PR #3448, opened 2025-10-27
Edit account subtitle

sea-snake:
Edit account subtitle

# Changes

- Add subtitle variant for when "Set as default sign-in" is not shown.
- Update locale files.

# Tests

Verified that the correct subtitle is shown.

### DIFF
--- a/src/frontend/src/lib/locales/es.po
+++ b/src/frontend/src/lib/locales/es.po
@@ -558,7 +558,7 @@ msgstr "Renombra o establece como inicio de sesión predeterminado"
 
 #: src/frontend/src/lib/components/views/EditAccount.svelte:71
 msgid "Rename this account"
-msgstr "Renombrar esta cuenta"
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/locales/es.po
```diff
@@ -548,21 +552,25 @@ msgstr "Privacidad Verdadera"
...
 msgstr "Redirigiendo a la app..."
 
-#: src/frontend/src/lib/components/views/EditAccount.svelte:69
+#: src/frontend/src/lib/components/views/EditAccount.svelte:70
 msgid "Rename or make this your default sign-in"
 msgstr "Renombra o establece como inicio de sesión predeterminado"
 
+#: src/frontend/src/lib/components/views/EditAccount.svelte:71
+msgid "Rename this account"
+msgstr "Renombrar esta cuenta"
```
lmuntaner: "Renombra esta cuenta"

PR #3450, opened 2025-10-29
Implement responsive dashboard sidebar

sea-snake:
Implement responsive dashboard sidebar.

# Changes

- Add `NavItem` component for dashboard sidebar.
- Implement responsive sidebar in dashboard layout.
- Update dashboard homepage to render cards side by side on desktop.
- Add dashboard access methods page that renders existing access methods list implementation for now.

# Temporary changes 

_Will be replaced in follow up PRs_

- Update `AccessMethod` component so it fits on a smaller screen.
- Rename `AccessMethodsPanel` to `AccessMethods` and remove it's `Panel` container and header.
- Duplicate `IdentityPanel` as` AccessMethodsPanel` and update content with access methods text and data.
- Add `min-h-10` to `IdentityInfoPanel` and `AccessMethodsPanel` so they're both the same height.

# Tests

Updated dashboard e2e tests to handle additional required page navigation.

image
image
image
image
image
image
image

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
@@ -101,57 +101,6 @@
   });
 </script>
 
-<!-- Identity switcher & dialog -->
-{#snippet identitySwitcher()}
-  {#if isIdentityPopoverOpen}
-    <Popover
-      anchor={identityButtonRef}
-      onClose={() => (isIdentityPopoverOpen = false)}
-      direction="down"
-      align="end"
-      distance="0.75rem"
-    >
-      <IdentitySwitcher
-        selected={$authenticatedStore.identityNumber}
-        identities={lastUsedIdentities}
-        switchIdentity={handleSwitchIdentity}
-        useAnotherIdentity={() => {
-          isIdentityPopoverOpen = false;
-          isAuthDialogOpen = true;
-        }}
-        onClose={() => (isIdentityPopoverOpen = false)}
-        onLogout={identityInfo.logout}
-      />
-    </Popover>
-  {/if}
-  {#if isAuthDialogOpen}
-    <Dialog
-      onClose={() => (isAuthDialogOpen = false)}
-      showCloseButton={!isAuthenticating}
-      closeOnOutsideClick={!isAuthenticating}
-    >
-      <AuthWizard
-        bind:isAuthenticating
-        {onSignIn}
-        {onSignUp}
-        {onMigration}
-        onError={(error) => {
-          isAuthDialogOpen = false;
-          handleError(error);
-        }}
-        withinDialog
-      >
-        <h1 class="text-text-primary my-2 self-start text-2xl font-medium">
-          {$t`Use another identity`}
-        </h1>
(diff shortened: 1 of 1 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -86,182 +89,180 @@
...
+        useAnotherIdentity={() => {
+          isIdentityPopoverOpen = false;
+          isAuthDialogOpen = true;
+        }}
+        onClose={() => (isIdentityPopoverOpen = false)}
+        onLogout={identityInfo.logout}
+      />
+    </Popover>
+  {/if}
+  {#if isAuthDialogOpen}
```
lmuntaner: Does this need to go inside the `identitySwitcher` snippet?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -86,182 +89,180 @@
...
-      <LucideHome size="1rem" class="stroke-fg-quaternary" />
-      <h1>Home</h1>
-    </SideBarElement>
-    <SideBarElement class="text-text-primary" href="/manage/security">
-      <Shield size="1rem" class="stroke-fg-quaternary" />
-      <h1>Security</h1>
-    </SideBarElement>
-  </SideBarElementGroup>
+<!-- Identity switcher & dialog -->
+{#snippet identitySwitcher()}
```
lmuntaner: If the elements are rendered at the main level, I don't see the advantage of having it in a snippet, any reason to have it like this?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+page.svelte
```diff
@@ -45,14 +45,17 @@
...
+  <h2 class="text-text-tertiary mb-12 text-base">
+    Manage your identity and access methods.
   </h2>
 
-  <div class="flex flex-col gap-6">
-    <IdentityInfoPanel />
-
-    <AccessMethods />
+  <div class="flex flex-col gap-6 lg:flex-row">
+    <div class="flex-1">
```
lmuntaner, line 53: Why is the `flex-1` needed?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
sea-snake: Ah that's indeed no longer necessary.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
sea-snake: Used to make sense, now I can indeed just move it to the bottom of the component.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+page.svelte
sea-snake: To make both cards use 50% of the available width when they're next to each other. The cards don't take a class prop, so I had to add a wrapper div to avoid making changes to the components (since they're going to be replaced anyway).

PR #3451, opened 2025-10-29
Recovery phrase page setup

lmuntaner:
# Motivation

Users can gain access to their identity using the recovery phrase.

In this PR, setup the page that will start the recovery flow.

In other PRs, refactor the recovery phrase component so it can be reused in the dashboard.

# Changes

* New route `/recovery-phrase` with page and layout.

# Tests

Tested locally. See video and screenshot.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/recovery-phrase/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/recovery-phrase/+page.svelte
@@ -7,7 +7,7 @@
   let inputContainerRef: HTMLDivElement | null = null;
 
   const handleKeyDownInput = (event: KeyboardEvent, currentIndex: number) => {
-    if (event.key === "Enter" || event.key === " ") {
+    if (event.key === "Enter" || event.code === "Space") {
       event.preventDefault();
       const nextTabIndex = currentIndex + 1;
       const inputElements = inputContainerRef?.querySelectorAll(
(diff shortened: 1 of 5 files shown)


src/frontend/src/routes/(new-styling)/recovery-phrase/+page.svelte
```diff
@@ -0,0 +1,48 @@
...
+        <h1 class="text-text-primary text-2xl font-medium">
+          {$t`Secure recovery`}
+        </h1>
+        <p class="text-text-tertiary text-sm">
+          {$t`Enter your recovery phrases in each box, spelled correctly and in order.`}
+        </p>
+      </div>
+      <div class="flex flex-col gap-3">
+        <div class="grid grid-cols-3 gap-3">
+          {#each Array(24) as _, i}
```
sea-snake: I'm surprised this works, in a JS function you wouldn't be able to map it and instead need to write `Array.from( { length: 24 })`.

image

src/frontend/src/routes/(new-styling)/recovery-phrase/+page.svelte
```diff
@@ -0,0 +1,48 @@
...
+  class="flex flex-1 flex-row items-end justify-center sm:max-w-120 sm:items-center"
+>
+  <AuthPanel>
+    <div class="flex flex-col gap-6">
+      <div class="flex flex-col gap-3">
+        <h1 class="text-text-primary text-2xl font-medium">
+          {$t`Secure recovery`}
+        </h1>
+        <p class="text-text-tertiary text-sm">
+          {$t`Enter your recovery phrases in each box, spelled correctly and in order.`}
```
sea-snake: Probably should be singular instead of plural since there's one phrase with multiple words instead of multiple phrases.

```suggestion
          {$t`Enter your recovery phrase words in each box, spelled correctly and in order.`}
```

PR #3462, opened 2025-11-04
Implement Select component and use it to replace language selectors.

sea-snake:
Implement Select component and use it to replace language selectors.

# Tests

Verified that language selectors still work as expected.

### DIFF
--- a/src/frontend/src/lib/flows/authFlow.svelte.ts
+++ b/src/frontend/src/lib/flows/authFlow.svelte.ts
@@ -56,7 +56,6 @@ export class AuthFlow {
     solve: (solution: string) => void;
   }>();
   #systemOverlay = $state(false);
-  #confirmationCode = $state<string>();
   #name = $state<string>();
   #jwt = $state<string>();
   #configIssuer = $state<string>();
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/stores/locale.store.ts
```diff
@@ -105,3 +105,17 @@ export const formatNumber = derived(
   () => (value: number, format?: Intl.NumberFormatOptions) =>
     i18n.number(value, format),
 );
+
+export const localeOptions = derived(locales, (locales) =>
```
lmuntaner, line 177: Nice!

src/frontend/src/lib/components/ui/Select.svelte
```diff
@@ -0,0 +1,99 @@
+<script lang="ts" generics="T = number">
+  import type { Component } from "svelte";
+  import type { HTMLAttributes } from "svelte/elements";
+  import Popover from "$lib/components/ui/Popover.svelte";
+  import { isNullish, nonNullish } from "@dfinity/utils";
+  import Button from "$lib/components/ui/Button.svelte";
+
+  type Option<T> = {
+    value?: T;
```
lmuntaner, line 9: Why is the value optional?

sea-snake: It's optional since not all usages of select will require explicit values, some use cases might suffice with just a list of labels and click handlers while others use values and an onChange handler. The value default to an index if not specified.

PR #3464, opened 2025-11-04
Fetch identity info data in load function

sea-snake:
Fetch identity info data in load function.

# Changes

- Add new (temporary) route `/manage/access-new`.
- Add layout with load function to this route that fetches the identity info.
- Update login page to preload next route and keep showing the (existing) loading indicator while doing so.

# Tests

Verified that authentication still works as expected and data is indeed available in the new route.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/login/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/login/+page.svelte
@@ -85,7 +85,7 @@
 
   const handleContinueAs = async (identity: LastUsedIdentity) => {
     await authLastUsedFlow.authenticate(identity).catch(handleError);
-    await gotoNext(identity.identityNumber);
+    await onSignIn(identity.identityNumber);
   };
 
   onMount(() => {


src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/+page.svelte
lmuntaner, line 1: Will this eventually be moved to `manage/(authenticated)/access/+page.svelte`? How about the page in `manage/(authenticated)/+page.svelte`?

lmuntaner (commented):
one question

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/+page.svelte
sea-snake: The page added in this PR will replace the current `/manage/access/+page.svelte` eventually. 

The dashboard homepage you're referring to is currently out of scope, It'll need to be updated (instead of replaced) with minimal changes to use the load function instead of identity info store data when the identity info store is removed in a future PR.

Updating the dashboard homepage to the new design will be picked up once all PRs related to the access methods page have been merged.

lmuntaner: Thanks for the explanation!

PR #3469, opened 2025-11-04
Use class for custom icon dimensions

sea-snake:
Use class for custom icon dimensions instead of size prop, this allows for css size overrides without !important e.g. `!size-6` in tailwind.

# Tests

The size prop wasn't used yet, so no other changes were needed.


src/frontend/src/lib/components/icons/GoogleIcon.svelte
lmuntaner, line 1: Where is this icon used? Can't we remove it?

sea-snake: Yeah, we can probably remove it. I'll do a PR to cleanup this and other related unused things after the access methods (so it's just a single delete files PR).

PR #3470, opened 2025-11-04
New access methods page design.

sea-snake:
New access methods page design.

# Changes

- Implement new access methods design in `/manage/access-new`.
- Implement `PasskeyItem` and `OpenIdItem` components used by the above.
- Make sure II 1.0 access methods are still shown as disabled for now.

# Tests

Verified that the page works as expected by:
1. Visit `/manage/access-new` directly.
2. Sign in when redirected to login page.
3. End up on the new access methods page.

Manual tests:
- Can only register up to 8 passkeys in total.
- Old II 1.0 access methods are shown as disabled. 
- Cannot remove access method if there's only one remaining.
- Current access method is indicated with green dot and "right now" label for last used.
- Changes/additions/removals are reflected immediately after the dialogs are closed and in sync with data from load function after ~ 2 sec.

image
image

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/components/PasskeyItem.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/components/PasskeyItem.svelte
@@ -22,7 +22,7 @@
 
   const alias = $derived(getAuthnMethodAlias(passkey));
   const options = $derived([
-    ...(onRename
+    ...(onRename !== undefined
       ? [
           {
             label: $t`Rename`,
@@ -31,7 +31,7 @@
           },
         ]
       : []),
-    ...(onRemove
+    ...(onRemove !== undefined
       ? [
           {
             label: $t`Remove`,
@@ -89,7 +89,7 @@
           >
             <span>{$t`Right now`}</span>
           </Tooltip>
-        {:else if passkey.last_authentication[0]}
+        {:else if passkey.last_authentication[0] !== undefined}
           {@const date = new Date(
             nanosToMillis(passkey.last_authentication[0]),
           )}
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/components/OpenIdItem.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/components/OpenIdItem.svelte
@@ -33,7 +33,7 @@
 </script>
 
 <div class="mb-3 flex h-9 flex-row items-center">
-  {#if logo}
+  {#if logo !== undefined}
     <div class="text-fg-primary relative size-6">
       {@html logo}
       {#if isCurrentAccessMethod}
@@ -78,7 +78,7 @@
         >
           <span>{$t`Right now`}</span>
         </Tooltip>
-      {:else if openid.last_usage_timestamp[0]}
+      {:else if openid.last_usage_timestamp[0] !== undefined}
         {@const date = new Date(nanosToMillis(openid.last_usage_timestamp[0]))}
         <Tooltip
           label={$formatDate(date, {


src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/+page.svelte
```diff
@@ -1,3 +1,326 @@
...
+    | { openid: OpenIdCredential };
+
+  const { data }: PageProps = $props();
+
+  let isAddAccessMethodWizardOpen = $state(false);
+  let renamablePasskey = $state<AuthnMethodData>();
+  let removablePasskey = $state<AuthnMethodData>();
+  let unlinkableOpenIdCredential = $state<OpenIdCredential>();
+
+  const sortAccessMethods = (a: AccessMethod, b: AccessMethod) => {
```
lmuntaner: Could we move this to a separate file to make this component more readable? And it would also be good to have a comment on how the function sorts them. And if we move it to another file, we could also add tests? 😄

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/+page.svelte
```diff
@@ -1,3 +1,326 @@
...
+    "passkey" in $authenticatedStore.authMethod &&
+    "WebAuthn" in data.authn_method &&
+    bytesToHex($authenticatedStore.authMethod.passkey.credentialId) ===
+      bytesToHex(new Uint8Array(data.authn_method.WebAuthn.credential_id));
+  const openIdInUse = (credential: OpenIdCredential) =>
+    "openid" in $authenticatedStore.authMethod &&
+    $authenticatedStore.authMethod.openid.iss === credential.iss &&
+    $authenticatedStore.authMethod.openid.sub === credential.sub;
+
+  let accessMethods = $derived(
```
lmuntaner: Could we move these derived stores to `accessMethods.derived.svelte`?

It's nice because then we can easily add unit tests for the derived stores.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/+page.svelte
```diff
@@ -1,3 +1,326 @@
...
+  const logoutAndForgetIdentity = () => {
+    lastUsedIdentitiesStore.removeIdentity($authenticatedStore.identityNumber);
+    location.replace("/login");
+  };
 </script>
+
+<header class="flex flex-col gap-3">
+  <h1 class="text-text-primary text-3xl font-medium">{$t`Access methods`}</h1>
+  <p class="text-text-tertiary text-base">
+    <Trans>
```
lmuntaner, line 155: Why `Trans` and not `$t`?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/components/PasskeyItem.svelte
lmuntaner, line 1: Are there plans on sharing a common component between OpenIdItem and PasskeyItem?

sea-snake: I considered it but opted to keep two separate components with some code duplication since the data structure passed into them and how e.g. fallbacks are handled are so different.

Initially when I had a single component that accepted the `AccessMethod` type, it became a complicated component that was hard to read due to the "if openid else if passkey" logic everywhere.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/+page.svelte
sea-snake: I tend to consistently use `$t` for short titles and labels while using `Trans` for longer things like sentences. 

Doing so avoids (possibly) long sentences that are hard to read in the code (since `$t` doesn't wrap) also it allows for updating the sentence with e.g. <b> in the future without having to switch to `<Trans>` in the end anyway.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/components/PasskeyItem.svelte
lmuntaner: I agree to the two components. I was thinking more on both of them sharing a UI component that uses slots for the content and only serves as layout and some basic styles. Not super necessary though as it's only in two places duplicated.

sea-snake: Yeah maybe later I would use Snippets as props into a generic component, but for now that seems unnecessary.

src/frontend/src/lib/utils/accessMethods.ts
```diff
@@ -119,6 +119,7 @@ const hasSomeOrigin = (
  * @returns {boolean}
  */
 export const isLegacyAuthnMethod = (accessMethod: AuthnMethodData): boolean => {
+  return false;
```
lmuntaner: Probably a leftover and should be removed?

src/frontend/src/lib/components/views/AccessMethods.svelte
```diff
@@ -5,14 +5,14 @@
...
   import PasskeyIcon from "$lib/components/icons/PasskeyIcon.svelte";
-  import UnlinkOpenIdCredential from "$lib/components/views/UnlinkOpenIdCredential.svelte";
   import { invalidateAll } from "$app/navigation";
   import type {
     AuthnMethodData,
     OpenIdCredential,
   } from "$lib/generated/internet_identity_types";
-  import RemovePasskey from "$lib/components/views/RemovePasskey.svelte";
-  import RenamePasskey from "$lib/components/views/RenamePasskey.svelte";
+  import RemovePasskey from "../../../routes/(new-styling)/manage/(authenticated)/access-new/components/RemovePasskey.svelte";
```
lmuntaner, line 13: We don't care right now about this because this View will also be removed, right?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access-new/utils.test.ts
lmuntaner, line 1: nice!

lmuntaner (commented):
LGTM but not approving yet as there is a suspicious change

src/frontend/src/lib/components/views/AccessMethods.svelte
sea-snake: Exactly, so for now the relative import is ok.

src/frontend/src/lib/utils/accessMethods.ts
sea-snake: Ah, yes that was for local dev 😅

PR #3472, opened 2025-11-06
Replace old access route with new access route.

sea-snake:
Replace old access route with new access route.

# Changes

- Delete /routes/access
- Rename /routes/access-new to /routes/access
- Update e2e tests to handle updated UI correctly

# Tests

Additional e2e test was added that makes sure that an identity can only have up to 8 passkeys.


src/frontend/tests/e2e-playwright/dashboard/addPasskeys.spec.ts
```diff
@@ -234,3 +266,50 @@ test("User can add a new passkey and use it with cached identity without clearin
 
   await newPage.close();
 });
+
+test("User can log into the dashboard and up to 7 additional passkeys", async ({
```
lmuntaner: Typo, missing "and add up to ..."

PR #3475, opened 2025-11-07
Reset session store when logged out due to removing current access method.

sea-snake:
# Changes

- Make sure to call `sessionStore.reset()` on logout due to removing current access method.
- Remove unused imports.


lmuntaner (approved):
Thanks!

PR #3476, opened 2025-11-07
Cleanup unused and fix layout.

sea-snake:
Cleanup unused and fix layout.

# Changes

- Cleanup unused components.
- Cleanup unused derived.
- Cleanup unused stores.
- Use correct selected border color (secondary instead of primary) on `NavItem` component.
- Rename method from onX to handleX in layout.
- Add `handleLogout` method since `identityStore` has been removed.
- Replace `HomeIcon` with `HouseIcon` import, the prior is deprecated (same icon).

# Tests

All e2e tests should pass, no feature changes in this PR.

### DIFF
--- a/src/frontend/src/lib/components/ui/LandingCard.svelte
+++ b/src/frontend/src/lib/components/ui/LandingCard.svelte
@@ -8,8 +8,6 @@
     description: string;
   };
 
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Nice!

PR #3483, opened 2025-11-10
feat: Add new API function `lookup_caller_identity_by_recovery_phrase`

aterga:
# Motivation

Enable recovery phrases for id.ai

| [Next PR](https://github.com/dfinity/internet-identity/pull/3487) >

# Changes

Just adding the function to agree on the API shape.

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -1014,4 +1014,7 @@ service : (opt InternetIdentityInit) -> {
         origin : FrontendHostname,
         account_number : opt AccountNumber
     ) -> (variant { Ok : AccountInfo; Err: SetDefaultAccountError });
+
+    // Looks up identity number when called with a recovery phrase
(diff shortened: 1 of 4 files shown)


src/internet_identity/internet_identity.did
```diff
@@ -1014,4 +1014,6 @@ service : (opt InternetIdentityInit) -> {
         origin : FrontendHostname,
         account_number : opt AccountNumber
     ) -> (variant { Ok : AccountInfo; Err: SetDefaultAccountError });
+
+    lookup_caller_identity_by_recovery_phrase : () -> (opt IdentityNumber) query;
```
sea-snake, line 1019: Nit: I would add a comment above this method to explicitly indicate it's not part of the `// Multiple accounts` section e.g.
```suggestion
    // Looks up identity number when called with a recovery phrase
    lookup_caller_identity_by_recovery_phrase : () -> (opt IdentityNumber) query;
```

PR #3484, opened 2025-11-10
Fix switching identity in dashboard.

sea-snake:
Fix switching identity in dashboard.

# Changes

- Replace `gotoManage` method with `switchedToIdentity` method.
- Called in all methods that switched the identity (handle- switch, sign in, sign up and migration).
- Makes sure that load function data is invalidated when identity is switched.
- Add loading animation when identity is switched (since we're waiting for an update call).
- Nit: key in `$lastUsedIdentitiesStore.identities` is a `string`, so using `toString` instead in `handleSwitchIdentity`.

# Tests

Added e2e test to make sure page data is updated when identity is switched.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/recovery-phrase/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/recovery-phrase/+page.svelte
@@ -14,6 +14,7 @@
   import type { DelegationIdentity } from "@icp-sdk/core/identity";
   import { wordlists } from "bip39";
   import SuccessfulRecovery from "./components/SuccessfulRecovery.svelte";
+  import RecoveryError from "./components/RecoveryError.svelte";
   import { toaster } from "$lib/components/utils/toaster";
 
   type RecoveryWord = {
@@ -53,6 +54,9 @@
   >(undefined);
 
   let continueInProgress = $state(false);
+  let recoveryError = $state(false); // Recovery phrase fails checksum validation
+  // When this is `true`, the auto-submit is disabled, and the user has to manually submit the recovery phrase.
+  let manualSubmitRequired = $state(false);
 
   const submitEnabled = $derived(
     words.every((word) => word.value.trim().length > 0 && word.isValid),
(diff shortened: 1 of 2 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -34,36 +36,37 @@
...
     Object.values($lastUsedIdentitiesStore.identities)
       .sort((a, b) => b.lastUsedTimestampMillis - a.lastUsedTimestampMillis)
       .slice(0, 3),
   );
 
-  const gotoManage = () => goto("/manage", { replaceState: true });
-  const handleSignIn = async (identityNumber: bigint) => {
-    lastUsedIdentitiesStore.selectIdentity(identityNumber);
-    await gotoManage();
+  const switchedToIdentity = async (identityNumber: bigint) => {
```
lmuntaner: Why `switched` and not `switch`?

src/frontend/tests/e2e-playwright/dashboard/index.spec.ts
```diff
@@ -39,4 +39,46 @@ test.describe("Dashboard Navigation", () => {
     await expect(passkey).toBeVisible();
     await expect(passkey).toHaveCount(1);
   });
+
+  test("User can switch between identities", async ({ page }) => {
```
lmuntaner, line 43: nice!

lmuntaner (approved):
Thanks for the E2E test!

Minor comment only, already approving.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
sea-snake: Since this method is technically called after the user has already authenticated with another identity, thus switched identity.

For example it's called in `handleSwitchIdentity` after authentication with the next identity.

sea-snake: But since we're calling.selectIdentity() in this method, switch might actually read better than switched.

PR #3485, opened 2025-11-11
Show no identity found modal in recovery flow

lmuntaner:
# Motivation

Better error management in the recovery flow.

This PR adds a new modal when the recovery phrase is valid but no associated identity is found.

# Changes

* Added a new `IdentityNotFoundError` class in `recoverWithPhraseFlow.svelte.ts` to differentiate between an invalid phrase and a valid phrase with no associated identity. The recovery logic now throws this error when appropriate.
* Updated the recovery flow in `+page.svelte` to track and handle the new error state (`identityNotFoundError`), resetting it properly during retries and cancellation.
* Imported and integrated the new `IdentityNotFound` component in the recovery phrase page, displaying a dedicated dialog when the error occurs.
* Implemented the `IdentityNotFound.svelte` component to show an informative message and provide retry/cancel actions when no identity is found for the entered recovery phrase.

# Tests

Tested locally, see attached video.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/recovery-phrase/components/IdentityNotFound.svelte
+++ b/src/frontend/src/routes/(new-styling)/recovery-phrase/components/IdentityNotFound.svelte
@@ -3,6 +3,7 @@
   import Button from "$lib/components/ui/Button.svelte";
   import { UserXIcon } from "@lucide/svelte";
   import { t } from "$lib/stores/locale.store";
+  import Trans from "$lib/components/locale/Trans.svelte";
 
   type Props = {
     onRetry: () => void;
@@ -22,7 +23,10 @@
         {$t`No identity found`}
       </h2>
       <p class="text-text-tertiary text-sm">
-        {$t`Your recovery phrase is valid but not for internet identity. You may have used the wrong seed phrase for the wrong product.`}
+        <Trans>
+          Your recovery phrase is valid but not for internet identity. You may
+          have used the wrong seed phrase for the wrong product.
+        </Trans>
       </p>
     </div>
   </div>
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/flows/recoverWithPhraseFlow.svelte.ts
```diff
@@ -48,6 +48,12 @@ export class InvalidMnemonicError extends Error {
   }
 }
 
+export class IdentityNotFoundError extends Error {
+  constructor() {
+    super("Identity not found");
```
sea-snake, line 53: Was about to mention that `Object.setPrototypeOf(this, IdentityNotFoundError.prototype);` would need to be called for instanceof to work correctly, but apparently that's no longer relevant in modern browsers and the ES target in our tsconfig. (We have a few custom errors where we do this)

src/frontend/src/routes/(new-styling)/recovery-phrase/components/IdentityNotFound.svelte
```diff
@@ -0,0 +1,37 @@
...
+  <div class="mt-6 flex flex-col gap-4">
+    <FeaturedIcon size="md" variant="error">
+      <UserXIcon class="size-6" />
+    </FeaturedIcon>
+    <div class="flex flex-col gap-3">
+      <h2 class="text-text-primary text-xl">
+        {$t`No identity found`}
+      </h2>
+      <p class="text-text-tertiary text-sm">
+        {$t`Your recovery phrase is valid but not for internet identity. You may have used the wrong seed phrase for the wrong product.`}
```
sea-snake: Nit: I would use `<Trans>` here since it's a paragraph that could be longer and thus benefit from wrapping in the IDE.

lmuntaner: done!

PR #3486, opened 2025-11-11
Update homepage cards design

sea-snake:
Update homepage cards design.

# Changes

- Render cards with at least 24rem width (automatically adds more grid columns if space available).
- Make sure that cards are rendered smaller than 24 rem is less space available (mobile), apply this same fix to access methods grid.
- Co-locate components with dashboard home page
- Update locale files with translations
- Restructure home page header for accessibility (e.g. use <header> component)
- Move dashboard homepage route into (home) folder to clarify that the co-located components are only for this route.

# Tests

image
image
image
image
image
image

### DIFF
--- a/src/frontend/tests/e2e-playwright/dashboard/index.spec.ts
+++ b/src/frontend/tests/e2e-playwright/dashboard/index.spec.ts
@@ -23,9 +23,6 @@ test.describe("Dashboard Navigation", () => {
         name: new RegExp(`Welcome, ${TEST_USER_NAME}!`),
       }),
     ).toBeVisible();
-    await expect(
-      page.getByRole("heading", { level: 5, name: TEST_USER_NAME }),
-    ).toBeVisible();
 
     // Navigate to access methods
     const menuButton = page.getByRole("button", { name: "Open menu" });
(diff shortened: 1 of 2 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/(home)/+page.svelte
```diff
@@ -37,24 +40,24 @@
...
-      <IdentityInfoPanel {name} />
-    </div>
-    <div class="flex-1">
-      <AccessMethodsPanel {totalAccessMethods} />
-    </div>
-  </div>
+<div
+  class="mt-10 grid grid-cols-[repeat(auto-fill,minmax(min(100%,24rem),1fr))] gap-6"
+>
+  <Panel class="p-4">
```
lmuntaner: In my opnion the padding shuold be responsibility of the `Panel`. in the end, we want consistency in the panels, no? It's similar to allowing the parent to choose the border radius in my opinion.

sea-snake: Wasn't entirely sure, I was thinking that the Panel should decide it's styling but shouldn't enforce the padding since that might differ between places where it's used.

This largely depends on how the Panel is re-used, right now it's not re-used at all beyond this particular route (it's co-located). So I'll move the padding into the Panel component for now until we have more clarity on this.

PR #3487, opened 2025-11-11
feat: Add recovery phrase principal index and implement its synchronization with anchors

aterga:
# Motivation

This is the next step towards supporting recovery phrases in id.ai

< [Previous PR](https://github.com/dfinity/internet-identity/pull/3483) | [Next PR](https://github.com/dfinity/internet-identity/pull/3490) >

# Changes

* Added `lookup_anchor_with_recovery_phrase_principal_memory` index to map recovery phrase principals to anchor numbers.
* Implemented `sync_anchor_with_recovery_phrase_principal_index` function to maintain index consistency when devices are added/removed.

# Tests

1. `adds_new_seed_phrase_principals` (also checks that the index extensions are idempotent)
2. `removes_old_seed_phrase_principals`
3. `no_change_if_same_devices`
4. `handles_empty_current_and_previous`
5. `adds_and_removes_seed_phrase_principals_in_single_call`
6. `removes_seed_phrase_principals_only_for_specified_anchor`
7. `write_moves_recovery_phrase_principal_between_anchors`


src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -215,6 +215,16 @@ fn create_identity(arg: &CreateIdentityData) -> Result<IdentityNumber, IdRegFini
                 &AuthorizationKey::DeviceKey(device.pubkey.clone()),
             );
 
+            // Sync recovery phrase principal index (this is currently a no-op, as recovery phrases
+            // are not used as the initial device during registration).
+            state::storage_borrow_mut(|storage| {
+                storage.sync_anchor_with_recovery_phrase_principal_index(
```
sea-snake: I would recommend to implement the sync in the `storage.rs` write method instead, this way we guarantee that the anchors and lookup memories are always in sync.

This would also avoid the need to make sure that `sync_anchor_with_recovery_phrase_principal_index` is called in various business logic places. Instead, any change to an anchor is automatically reflected in the lookup memory.

aterga: Brilliant. Done

sea-snake (approved):
LGTM

But please double check the recovery phrase creation implementation (current and past implementations) that `KeyType::SeedPhrase = device.key_type` is correct and sufficient.

src/internet_identity/src/storage/tests.rs
```diff
@@ -805,3 +805,283 @@ mod storable_origin_sha256_tests {
...
+    fn pubkey(n: u8) -> PublicKey {
+        vec![n].into()
+    }
+
+    fn seed_phrase_device(pubkey: PublicKey) -> Device {
+        Device {
+            pubkey,
+            alias: "seed".to_string(),
+            credential_id: None,
+            purpose: Purpose::Authentication,
```
sea-snake: As mentioned in my other comment, should this be `Authentication` or something else?

aterga: Thanks for bringing this up!

While setting up a recovery device, we specify
* keyType = cross_platform
* purpose = recovery
While setting up a recovery phrase, we specify
* keyType = seed_phrase
* purpose = recovery

Added a proper definition to `impl Device` to avoid confusion.

PR #3489, opened 2025-11-12
Move passkey from another device from /manage to /manage/access route.

sea-snake:
Move passkey from another device from /manage to /manage/access route.

# Changes

- Update lucide from 0.508.0 to 0.553.0 to fix render issues in dev mode.
- Replace deprecated `MinusCircleIcon` with `CircleMinusIcon` (same icon, different import name).
- Move `ConfirmAccessMethodWizard` from `/manage` route to `/manage/access`.
- Update `/activate` route to redirect to `/manage/access` instead of `/manage`.


src/frontend/src/routes/(new-styling)/+page.ts
```diff
@@ -1,2 +1,2 @@
-export const ssr = true;
+export const ssr = false;
```
lmuntaner: Why this change? It doesn't look right...

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.ts
lmuntaner, line 1: Where was this file before?

src/frontend/src/routes/(new-styling)/+page.ts
sea-snake: Ah yes, disabled that when I was debugging the render issue I had on my local 😅

sea-snake: Reverted this change

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.ts
sea-snake: In `/manage/(authenticated)/(home)`

PR #3490, opened 2025-11-12
chore: Migrate recovery phrases to populate the new index

aterga:
# Motivation

This PR implements a data migration for synchronizing anchor indices. The immediate goal is to initialize the new recovery phrase index needed to enable using recovery phrases without asking users for their anchor numbers (which are, in id.ai, an internal implementation detail). For completeness, we also sync the other anchor index (`lookup_anchor_with_device_credential`), which is a consequence of the migration code that simply reads and write anchors using the latest, synchronization-preserving code. An added benefit of this migration approach is that it effectively tests that anchors can be migrated to unbounded memory, which has been a long-standing task for II.

< [Previous PR](https://github.com/dfinity/internet-identity/pull/3487) | [Next PR](https://github.com/dfinity/internet-identity/pull/3493) >

# Changes

* Add the migration based on `ic_cdk_timers` to dispatch 2_000 anchors per message with 1 second break-off (to minimize the chance that the migration causes a DoS).

# Tests

* Integration test `upgrade_and_rollback_with_realistic_data_migration` demonstrating that the migration succeeds on a small number of anchors (run manually for 2020 anchors; keeping just 20 for CI for performance reasons).

### DIFF
--- a/src/internet_identity/src/migrations/sync_anchor_indices.rs
+++ b/src/internet_identity/src/migrations/sync_anchor_indices.rs
@@ -207,9 +207,9 @@ mod sync_anchor_indices_tests {
         storage.sync_anchor_indices(BATCH_SIZE);
 
         // Check that recovery phrase principals are indexed for anchors 1 and 3
-        let principal_1 = Principal::self_authenticating(&pubkey(1));
-        let principal_2 = Principal::self_authenticating(&pubkey(2));
-        let principal_3 = Principal::self_authenticating(&pubkey(3));
+        let principal_1 = Principal::self_authenticating(pubkey(1));
+        let principal_2 = Principal::self_authenticating(pubkey(2));
+        let principal_3 = Principal::self_authenticating(pubkey(3));
 
         assert_eq!(
             storage
@@ -265,9 +265,9 @@ mod sync_anchor_indices_tests {
         storage.sync_anchor_indices(BATCH_SIZE);
 
         // Check that recovery phrase principals are indexed for anchors 1 and 3
-        let principal_1 = Principal::self_authenticating(&pubkey(1));
-        let principal_2 = Principal::self_authenticating(&pubkey(2));
-        let principal_3 = Principal::self_authenticating(&pubkey(3));
+        let principal_1 = Principal::self_authenticating(pubkey(1));
+        let principal_2 = Principal::self_authenticating(pubkey(2));
+        let principal_3 = Principal::self_authenticating(pubkey(3));
 
         assert_eq!(
             storage


src/internet_identity/src/storage.rs
```diff
@@ -766,43 +763,49 @@ impl<M: Memory + Clone> Storage<M> {
...
-                if !self
-                    .lookup_anchor_with_device_credential_memory
-                    .contains_key(&credential_id)
-                {
-                    self.lookup_anchor_with_device_credential_memory
-                        .insert(credential_id, anchor_number);
-                }
-            });
+
+        for credential_id in previous_credential_ids.difference(&current_credential_ids) {
```
aterga, line 776: This code is refactored to be consistent with the recovery phrase index migration code. 

The only behavior change is that we add not just `current_credential_ids.difference(previous_credential_ids)` but everything from `current_credential_ids` which is needed to ensure that the migration takes place for all elements.

src/internet_identity/src/main.rs
```diff
@@ -576,13 +620,46 @@ fn init(maybe_arg: Option<InternetIdentityInit>) {
...
+    if RECOVERY_PHRASE_MIGRATION_BATCH_ID.with(|id| *id.borrow()) == u64::MAX {
+        // Migration complete, clear timer.
+        TIMER_ID.with_borrow_mut(|saved_timer_id| {
+            if let Some(saved_timer_id) = *saved_timer_id {
+                ic_cdk_timers::clear_timer(saved_timer_id);
+            }
+            *saved_timer_id = None;
+        });
+    }
+}
```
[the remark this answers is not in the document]
aterga: Needs to be marked as async to be an argument of `ic_cdk::spawn`

src/internet_identity/src/storage.rs
```diff
@@ -716,20 +716,17 @@ impl<M: Memory + Clone> Storage<M> {
             Some(Principal::self_authenticating(&device.pubkey))
         };
 
-        let previous_devices = previous_devices
+        let previous_recovery_phrases = previous_devices
```
lmuntaner, line 719: Do the changes in this function change the functionality of the function or are they only to make it more readable?

src/internet_identity/src/storage.rs
lmuntaner: Does this function have a test?

src/internet_identity/src/migrations/lookup_anchor_with_recovery_phrase_principal.rs
```diff
@@ -0,0 +1,125 @@
+use ic_stable_structures::Memory;
+
+use crate::{
+    storage::Storage, RECOVERY_PHRASE_MIGRATION_BATCH_ID, RECOVERY_PHRASE_MIGRATION_BATCH_SIZE,
+    RECOVERY_PHRASE_MIGRATION_ERRORS, RECOVERY_PHRASE_MIGRATION_LAST_ANCHOR_ID,
+};
+
+impl<M: Memory + Clone> Storage<M> {
+    /// Migrates the next batch of anchors to synchronize the following indices:
+    /// `lookup_anchor_with_recovery_phrase_principal_memory_wrapper`
```
lmuntaner: Is it `lookup_anchor_with_recovery_phrase_principal_memory_wrapper` or just `lookup_anchor_with_recovery_phrase_principal`?

src/internet_identity/src/migrations/lookup_anchor_with_recovery_phrase_principal.rs
```diff
@@ -0,0 +1,125 @@
...
+            .saturating_add(RECOVERY_PHRASE_MIGRATION_BATCH_SIZE.saturating_mul(batch_id));
+
+        let end = begin.saturating_add(RECOVERY_PHRASE_MIGRATION_BATCH_SIZE);
+
+        // Stop *before* the next un-allocated anchor number.
+        let end = std::cmp::min(end, next_anchor_number);
+
+        let mut errors = vec![];
+
+        for anchor_number in begin..end {
```
lmuntaner: I don't understand the functionality of this loop. It reads and updates the anchor, is it because the storage syncs the indexes behind the scenes? It might make sense to add a comment maybe?

src/internet_identity/tests/integration/upgrade.rs
```diff
@@ -315,3 +317,105 @@ fn upgrade_and_rollback_keeps_accounts_intact() {
     assert_eq!(accounts_before, accounts_between);
     assert_eq!(accounts_between, accounts_after);
 }
+
+// TODO: Remove this test after the migration takes place.
+#[test]
+fn upgrade_and_rollback_with_realistic_data_migration() {
```
lmuntaner: 20 doesn't see "realistic" data, no? We have 2M anchors...

src/internet_identity/tests/integration/upgrade.rs
```diff
@@ -315,3 +317,105 @@ fn upgrade_and_rollback_keeps_accounts_intact() {
...
+            account_name,
+        )
+        .unwrap()
+        .unwrap();
+    }
+
+    upgrade_ii_canister(&env, canister_id, II_WASM.clone());
+
+    for _ in 0..NUM_ANCHORS {
+        env.tick();
```
lmuntaner, line 361: Why all these `tick`?

lmuntaner (commented):
Looking very good!

I wonder whether we want a test to verify that when there is an error the migration doesn't stop, or it stops if all succeed.

src/internet_identity/src/storage.rs
sea-snake: I added test coverage in both storage layer and integration test back then.

`src/internet_identity/src/storage/tests.rs`:
- `should_write_and_update_device_credential_lookup`
- `should_not_overwrite_device_credential_lookup`

`src/internet_identity/tests/integration/anchor_management/device_management.rs`:
- `should_lookup_device_key_with_credential_id`

src/internet_identity/tests/integration/upgrade.rs
aterga: We can't have it realistic, since a manual run with 2020 anchors takes about 10 min already. 

I think the fact that we're batching the migration means it's almost the same if we have 2 batches or 1000 batches.

src/internet_identity/tests/integration/upgrade.rs
aterga: Indeed, there's too many ticks here (we need 1 block per batch)

src/internet_identity/src/storage.rs
aterga: The only logical change is that `devices_to_be_added` is now `current_recovery_phrases` rather than `current_devices.difference(&previous_devices)`

src/internet_identity/src/migrations/lookup_anchor_with_recovery_phrase_principal.rs
aterga: > is it because the storage syncs the indexes behind the scenes

Yes — so we have the exact same code as if each user manually synched their anchor data 

> add a comment

Will do

src/internet_identity/src/migrations/sync_anchor_indices.rs
```diff
@@ -0,0 +1,326 @@
...
+            RECOVERY_PHRASE_MIGRATION_BATCH_ID
+                .with_borrow_mut(|id| *id = batch_id.saturating_add(1));
+        } else {
+            ic_cdk::println!("Recovery phrase migration COMPLETED.");
+            RECOVERY_PHRASE_MIGRATION_BATCH_ID.replace(u64::MAX);
+        }
+    }
+}
+
+#[cfg(test)]
```
lmuntaner, line 127: Thanks for the extra tests!

src/internet_identity/src/main.rs
```diff
@@ -62,6 +66,51 @@ const INTERNETCOMPUTER_ORG_ORIGIN: &str = "https://identity.internetcomputer.org
...
+pub(crate) const RECOVERY_PHRASE_MIGRATION_BATCH_BACKOFF_SECONDS: Duration = Duration::from_secs(1);
+
+thread_local! {
+    // TODO: Remove this state after the data migration is complete.
+    pub(crate) static RECOVERY_PHRASE_MIGRATION_BATCH_ID: RefCell<u64> = const { RefCell::new(0) };
+    pub(crate) static RECOVERY_PHRASE_MIGRATION_ERRORS: RefCell<Vec<String>> = const { RefCell::new(Vec::new()) };
+    pub(crate) static RECOVERY_PHRASE_MIGRATION_LAST_ANCHOR_ID: RefCell<Option<u64>> = const { RefCell::new(None) };
+
+    static TIMER_ID: RefCell<Option<TimerId>> = const { RefCell::new(None) };
+}
```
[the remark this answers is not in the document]
aterga: This is fine for us

PR #3493, opened 2025-11-13
feat: Implement `lookup_caller_identity_by_recovery_phrase`

aterga:
# Motivation

This PR implements the `lookup_caller_identity_by_recovery_phraseba` backend function that enables recovery phrases without needing users to provide their Identity anchor number, which is an internal implementation detail in id.ai.

< [Previous PR](https://github.com/dfinity/internet-identity/pull/3490) |

# Changes

Straightforward implementation for `lookup_caller_identity_by_recovery_phraseba`.

# Tests

Extended `test_recovery_phrase_migration` to demo how this function can be used.


src/internet_identity/src/main.rs
```diff
@@ -321,9 +321,7 @@ fn get_anchor_credentials(anchor_number: AnchorNumber) -> AnchorCredentials {
 
 #[query]
 fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
-    // TODO: Implement this function after the `lookup_anchor_with_recovery_phrase_principal_memory`
-    // TODO: index is initialized with existing user data.
-    None
+    state::storage_borrow(|storage| storage.lookup_anchor_with_recovery_phrase_principal(caller()))
```
sea-snake, line 324: Nit: we try to avoid calling the storage directly form the query/update methods, instead we call it though `anchor_management.rs` e.g. `lookup_anchor_with_openid_credential`.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/+page.ts
```diff
@@ -1,5 +0,0 @@
-import type { PageLoad } from "./$types";
```
sea-snake, line 1: No idea why this file is shown, but since this PR isn't pointing to `main` branch, it doesn't matter (as long the later PR pointing to `main` doesn't modify any unrelated files).

PR #3500, opened 2025-11-17
Implement recovery phrase wizard.

sea-snake:
Implement recovery phrase wizard.

# Changes

- Implement `CreateRecoveryPhraseWizard` component including views.
   - The wizard and view components are minimal, just enough to be able to complete the flow, they'll be updated in next PRs.
- Implement `UnverifiedRecoveryPhrase` component.
  - Based on  `ActiveRecoveryPhrase` component.
  - Last used will always be n/a since unverified status isn't (and can't be) stored in the canister, thus the recovery phrase will be active on e.g. full page reload.
- Tied above wizard into recovery page to have a working flow.

# Tests

Not yet in this PR, expect fn and e2e tests in next PRs.

### DIFF
--- a/src/internet_identity/tests/integration/upgrade.rs
+++ b/src/internet_identity/tests/integration/upgrade.rs
@@ -328,6 +328,7 @@ fn test_sync_anchor_indices_migration() {
     const NUM_ANCHORS: usize = 20;
 
     for i in 0..NUM_ANCHORS {
+        let name = Some(format!("Test User {}", i));
         let pubkey = format!("pub-key-{}", i);
         let origin = format!("https://www.app{}.org", i);
         let device_data = DeviceData {
@@ -339,22 +340,10 @@ fn test_sync_anchor_indices_migration() {
             key_type: KeyType::SeedPhrase,
             ..DeviceData::auth_test_device()
         };
+        let authn_method = AuthnMethodData::from(device_data);
 
-        let name = if i == 0 {
-            // A very long name should cause an error during identity creation. If this happens,
-            // the Identity anchor should not be allocated, or else the migration would crash while
-            // trying to read it.
-            Some("Test User".repeat(100))
-        } else {
-            Some(format!("Test User {}", i))
-        };
-
-        let _identity_number = create_identity_with_authn_method_and_name(
-            &env,
-            canister_id,
-            &AuthnMethodData::from(device_data),
-            name,
-        );
+        let _identity_number =
+            create_identity_with_authn_method_and_name(&env, canister_id, &authn_method, name);
     }
 
     upgrade_ii_canister(&env, canister_id, II_WASM.clone());
@@ -418,8 +407,7 @@ fn test_sync_anchor_indices_migration() {
 
         let count_recovery_phrases: u64 = candid::decode_one(&data).unwrap();
 
-        // One of the anchors did not actually exist, so only NUM_ANCHORS - 1 recovery phrases.
-        assert_eq!(count_recovery_phrases, NUM_ANCHORS as u64 - 1);
+        assert_eq!(count_recovery_phrases, NUM_ANCHORS as u64);
     }
 
     // smoke test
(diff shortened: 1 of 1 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -6,10 +6,24 @@
...
+  import UnverifiedRecoveryPhrase from "./components/UnverifiedRecoveryPhrase.svelte";
+  import { recoveryAuthnMethodData } from "$lib/utils/authnMethodData";
+  import type { AuthnMethodData } from "$lib/generated/internet_identity_types";
+  import { authnMethodEqual } from "$lib/utils/webAuthn";
 
   const { data }: PageProps = $props();
 
-  const recoveryPhrase = $derived(
+  let showRecoveryPhraseSetup = $state<"activate" | "reset" | "verify">();
+  let unverifiedRecoveryPhrase = $state<{
```
lmuntaner, line 23: Unverified means only in that session in the browser, right? On refresh it will disappear.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -33,14 +98,21 @@
...
       <ActiveRecoveryPhrase
-        onReset={() => {}}
-        {recoveryPhrase}
+        onReset={() => (showRecoveryPhraseSetup = "reset")}
+        recoveryPhrase={recoveryPhraseData}
         {isCurrentAccessMethod}
       />
     {:else}
-      <InactiveRecoveryPhrase onActivate={() => {}} />
+      <InactiveRecoveryPhrase
```
lmuntaner, line 118: What does it mean "inactive"? Is it that it doesn't exist yet?

src/frontend/src/lib/components/wizards/createRecoveryPhrase/CreateRecoveryPhraseWizard.svelte
lmuntaner, line 1: Shuold it be called "CreateRecoveryPhraseWizard" when it doesn't only create? How about "ManageRecoveryPhraseWizard"? Or too generic?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
sea-snake: Unverified means it's been already registered with the identity but hasn't been verified yet by the user. This "is unverified" state is indeed only retained in page state, reloading the page will result in showing the phrase as "active".

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
sea-snake: Yes, this is when the recovery phrase feature isn't activated yet aka a recovery phrase has been registered with the identity.

I'll add some code comments to clarify this and the above.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/CreateRecoveryPhraseWizard.svelte
sea-snake: Technically it does only "create" but also allows one to "continue" where they left off (verify if it's correct). 

Given that this latter part is more of an edge case, I think it's probably still the clearest to name it "create" since it e.g. doesn't do manage stuff like "rename" or "delete".

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
lmuntaner: Is this a new functionality?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
lmuntaner: "a recovery phrase has NOT been registered with the identity.". It was missing the "not", right?

PR #3502, opened 2025-11-18
fix: No more dangling identity anchors and make data migration work with non-existent anchors

aterga:
# Motivation

This PR fixes a bug in the identity anchor creation flow that caused some anchor numbers to remain allocated while the identity creation failed. The anchor index data migration is adjusted that occurs when encountering non-existent anchor records within the allocated anchor range. The changes enable the migration to handle missing anchors gracefully by creating dummy placeholder entries instead of crashing.

# Changes

* Changed the logic in `registration_flow_v2::create_identity` to allocate anchor numbers only upon their successful creation.
* Modified the storage layer to detect non-existent anchors through zero-byte checks and return AnchorNotFound errors.
* Updated the migration logic to create dummy anchors with <DUMMY ANCHOR> names when missing entries are encountered.

# Tests

* Changed test helpers to return Result types for proper error propagation in test scenarios.
* Enhanced test coverage for migration edge cases involving failed identity registrations.
* Regression test for `registration_flow_v2::create_identity`.
* Unit tests for the underlying new function `allocate_anchor_safe`.

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -590,6 +590,8 @@ impl<M: Memory + Clone> Storage<M> {
 
         self.header.num_anchors = self.header.num_anchors.saturating_add(1);
 
+        self.flush();
+
         Ok(result)
     }
 


src/internet_identity/src/storage.rs
```diff
@@ -658,6 +662,11 @@ impl<M: Memory + Clone> Storage<M> {
 
         reader.read_exact(&mut buf).expect("failed to read memory");
 
+        // Check if this anchor exists
```
sea-snake: This comment isn't sufficient for what it does, it probably should mention something like:
> anchors that are allocated but have never been written to, due to a prior missing allocation cleanup implementation, are handled as if they don't exist.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/+page.ts
```diff
@@ -1,5 +0,0 @@
-import type { PageLoad } from "./$types";
```
sea-snake, line 1: Why is this file deleted? Is the branch up to date with main?

aterga: I can't build II while this file exists

aterga: I undeleted it

src/internet_identity/src/storage.rs
```diff
@@ -567,6 +567,32 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// The new anchor is an optional argument to `f`, which is `None` if no anchor could be allocated.
+    ///
+    /// Otherwise, no state change is made, and the error from `f` is returned.
+    pub fn allocate_anchor_safe<F, T, E>(&mut self, now: Timestamp, f: F) -> Result<T, E>
+    where
+        F: FnOnce(Option<Anchor>) -> Result<T, E>,
+    {
+        let num_anchors = u64::from(self.header.num_anchors);
+
+        let anchor_number = self.header.id_range_lo.saturating_add(num_anchors);
```
sea-snake, line 581: Are you sure this gives the same anchor number as in `allocate_anchor`? The unsafe method has a +1 which I can't find here.

aterga: In practice an overflow should not occur here, since we only deep anchors useful if they are between two u64 values `id_range_lo` and `id_range_hi`. 

Avoiding unsafe arithmetic seems like a good thing anyway, since otherwise the reader would have to think about the consequences of a potential panic (which requires a non-local code analysis) / analyze its likelihood.

aterga: Addressed the main concern (should be +1 rather than -1)

PR #3504, opened 2025-11-18
refactor: Even safer anchor allocation

aterga:
# Motivation

This PR addresses comments brought up by Copilot in https://github.com/dfinity/internet-identity/pull/3502. This is generally improving code quality in storage.rs and registration_flow_v2.

# Changes

* Improved the `allocate_anchor_safe` method that only commits anchor allocation after successful validation and creation.
* Refactored `create_identity` to validate all data (not just key preconditions) before allocation and handle errors without leaving dangling anchors.
* Adjusted anchor creation to work with legacy flows (first, reserve the anchor number, then write to memory) as well as the new flow (first, ensure the new anchor is successfully written to memory, then reserve the anchor number).

# Tests

### DIFF
--- a/src/frontend/src/lib/components/utils/error.ts
+++ b/src/frontend/src/lib/components/utils/error.ts
@@ -43,6 +43,10 @@ export const handleError = (error: unknown) => {
   ) {
     switch (error.type) {
       case "RateLimitExceeded":
+        toaster.error({
+          title: "It seems like registration is unavailable at this moment",
+        });
+        break;
       case "InvalidCaller":
       case "UnexpectedCall":
       case "NoRegistrationFlow":


src/internet_identity_interface/src/internet_identity/types/api_v2.rs
```diff
@@ -244,7 +244,6 @@ pub struct IdRegFinishResult {
 
 #[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
 pub enum IdRegFinishError {
-    IdentityLimitReached,
```
sea-snake, line 247: If you remove this enum (unused?) you'll also need to remove it from the did file and update bindings afterwards (npm run generate).

aterga: Discussed over zoom, decider to ditch this case

PR #3510, opened 2025-11-20
Update create recovery phrase views

sea-snake:
Update create recovery phrase happy path views and their functionality to match design. The error and reset views will be updated in a follow-up PR. 

# Changes

- Updated `Acknowledge`, `Write` and `Verify` views to match design and rewrote it's copy.
- Implement ordered selection logic in `Verify` view.
- Implement `Steps` component to indicate the user journey through the recovery phrase set-up.
- Implement `StepsProgress` component that's used as an animated replacement of the above to indicate loading/waiting.
- Remove erroneous classname from `Checkbox` component (it's already on the outer wrapper).

# Tests

The e2e tests will be added in a later follow-up PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,7 @@
         "@wdio/globals": "^8.43.0",
         "@zag-js/svelte": "^1.12.0",
         "@zag-js/toast": "^1.12.0",
-        "astro": "^5.5.4",
+        "astro": "^5.15.9",
         "devtools-protocol": "^0.0.1502260",
         "eslint": "^9.18.0",
         "eslint-config-prettier": "^10.1.8",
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/wizards/createRecoveryPhrase/components/ProgressSteps.svelte
lmuntaner, line 1: I find the name a bit confusing, initially I thought it was the steps in the top of the wizard. I didn't understand it as a loading or spinner. Maybe `LoadingSteps` would be more accurate?

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Write.svelte
```diff
@@ -1,19 +1,73 @@
...
+    </span>
+    <span class="text-text-primary text-sm text-balance">
+      <Trans>
+        Make sure no one can see what's on your screen or what you are writing.
+      </Trans>
+    </span>
+  </button>
 </div>
-<Button onclick={onWritten}>{$t`I have written it down`}</Button>
+<Button onclick={onWritten} disabled={!isRevealed} size="lg">
```
lmuntaner, line 71: No button to copy all the words at once?

lmuntaner (approved):
LGTM! Some minor questions, but already approving.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Write.svelte
sea-snake: It's not in the design, neither would that align with what the interface is telling the user (store offline). Nonetheless, I made sure that you can still reliably select all words and copy to clipboard for those few power users that choose to ignore those warnings.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/components/ProgressSteps.svelte
sea-snake: I think I'll name it `StepsProgressBar` (in another PR).

PR #3511, opened 2025-11-21
Add remaining retry and reset views to recovery phrase creation wizard.

sea-snake:
Add remaining retry and reset views to recovery phrase creation wizard.

# Changes

- Implement `Retry` view.
- Move state above methods that mutate it in `Acknowledge`.
- Update `Reset` view according to design.
- Improve rendering of long words and accessibility in `Write` and `Verify` views.
- Show toast on recovery phrase activation.
- Update auth store if we just replaced the auth method currently in use.
- Warn user if they're leaving in the middle of a recovery phrase set-up.

# Tests

The e2e tests will be in a later follow-up PR.


src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -64,6 +70,19 @@
...
       .then(throwCanisterError);
     unverifiedRecoveryPhrase = { words, data };
+
+    // Update auth store if we just replaced the auth method currently in use
+    if (isCurrentAccessMethod) {
+      const identity = await fromMnemonicWithoutValidation(
+        words.join(" "),
+        IC_DERIVATION_PATH,
+      );
+      await authenticateWithSession({
```
lmuntaner, line 80: Why is this necessary?

sea-snake: It's an edge case:

If a user recovers with a recovery phrase A and directly resets it (replaces it with phrase B), any future actions on the dashboard would throw an error unless the authentication store state is updated from A to B.

aterga: Great that you thought about this corner case.

In my view, we should not allow deleting the current auth device at all. 

Let's discuss on Monday (NCR for this PR)

PR #3517, opened 2025-11-24
Recovery phrase set-up e2e tests.

sea-snake:
Recovery phrase set-up e2e tests.

# Changes

- Make dialog close button accessible.
- Make recovery phrase verification step buttons accessible (wrap in unordered list and aria-pressable attribute).
- Add `readClipboard` to e2e utils that avoids CSP errors.

# Tests

Implemented e2e tests to cover all currently possible flows on the new recovery phrase setup page.


src/frontend/src/lib/components/ui/Dialog.svelte
```diff
@@ -133,8 +134,9 @@
         type="button"
         class="absolute top-2 right-2 z-2 !rounded-full"
         onclick={onClose}
+        aria-label={$t`Close`}
       >
-        <XIcon class="size-5" />
+        <XIcon class="size-5" aria-hidden="true" />
```
aterga, line 139: what is aria?

sea-snake: They're standardized additional HTML accessibility attributes for e.g. screen readers.

In our case, the e2e tests rely on the accessibility tree instead of the DOM tree, this makes them less prone to break due to e.g. html class/id changes and also guarantees that human interactions can navigate through the screens (e.g. buttons have the correct label and are visible).

https://web.dev/learn/accessibility/aria-html

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
```diff
@@ -54,55 +54,58 @@
...
-        isCheckingOrder && "!bg-bg-disabled !border-border-disabled_subtle",
-      ]}
-      disabled={(isSelected && !isLastSelected) || isCheckingOrder}
-    >
-      <span
+    <li class="contents">
+      <button
+        onclick={() =>
+          (selectedIndexes = isLastSelected
+            ? selectedIndexes.slice(0, -1)
```
aterga: what does this do? all but the last index?

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
```diff
@@ -54,55 +54,58 @@
     <Trans>Select each word in the correct order</Trans>
   {/if}
 </p>
-<div class={["mb-8 grid grid-cols-3 gap-3"]}>
+<ul class={["mb-8 grid grid-cols-3 gap-3"]}>
```
sea-snake, line 70: Only changes in this file:
- `div` -> `ul`.
- `li` wrapper around button.
- `aria-pressed` attribute on button.

These changes make the view more accessible, which was also a requirement for the e2e test so it can find and click the correct buttons.

There are no other changes on this page, other line changes are coming from prettier formatting.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
```diff
@@ -54,55 +54,58 @@
...
-      ]}
-      disabled={(isSelected && !isLastSelected) || isCheckingOrder}
-    >
-      <span
+    <li class="contents">
+      <button
+        onclick={() =>
+          (selectedIndexes = isLastSelected
+            ? selectedIndexes.slice(0, -1)
+            : [...selectedIndexes, index])}
```
aterga: recursive expression; maybe add a comment to explain the idea?

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,255 @@
+import { expect, Locator, Page, test } from "@playwright/test";
+import {
+  createNewIdentityInII,
+  dummyAuth,
+  II_URL,
+  readClipboard,
+} from "../utils";
+
+class RecoveryPhraseWizard {
```
lmuntaner, line 9: Why a class instead of helpers that accept the dialog as parameter?

lmuntaner (commented):
Minor comment

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
sea-snake: Yes, I'll add a comment.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
sea-snake: I'll mention it in the same comment.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
```diff
@@ -54,55 +54,58 @@
...
+            ? `${selectedPosition + 1}`.padStart(2, "0")
+            : "＿"}
+        </span>
+        <span
+          class={[
+            "border-border-secondary mx-1 h-full border-r",
+            isSelected && "!border-fg-primary",
+            isCheckingOrder && "!border-border-disabled_subtle",
+          ]}
+        ></span>
```
aterga, line 105: Why do we need an empty span?

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: To group the relevant helpers together, having helpers for both the page and wizard was becoming hard to differentiate between them.

Nonetheless, the class is not intended to become something stateful, all functions within are still stateless helpers.

src/frontend/src/lib/components/ui/Dialog.svelte
aterga: Thanks for the clarification. Nice

src/frontend/tests/e2e-playwright/utils.ts
```diff
@@ -354,3 +354,22 @@ export const addCredentialToVirtualAuthenticator = async (
     credential,
   });
 };
+
+/**
+ * Read current text value from clipboard, this is an alternative to
+ * `navigator.clipboard.readText()` since that is blocked by CSP.
+ */
+export const readClipboard = async (page: Page): Promise<string> => {
```
aterga, line 362: I wonder why we need this, and if this is okay given that there's probably a good reason why `navigator.clipboard.readText` is not okay to be used by a website.

src/frontend/tests/e2e-playwright/utils.ts
```diff
@@ -354,3 +354,22 @@ export const addCredentialToVirtualAuthenticator = async (
...
+    const el = document.createElement("textarea");
+    // Append element to dialog if present so it can always be focused
+    const dialog = document.body.querySelector("dialog");
+    (dialog ?? document.body).appendChild(el);
+    el.focus();
+    return el;
+  });
+  await page.keyboard.press("Meta+v");
+  const value = await page.evaluate((el) => el.value, textarea);
+  await page.evaluate((el) => el.remove(), textarea);
```
aterga: Not entirely clear what this line does. Are we replacing whatever `el` is with `textarea`? Why isn't `textarea` already an element of `page`?

src/frontend/tests/e2e-playwright/utils.ts
aterga: Ahhh. This is a test-only function, so it's of course fine.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+        .getByRole("button", { name: word, disabled: false, pressed: false })
+        .first()
+        .click();
+    }
+  }
+
+  async reset(): Promise<void> {
+    await expect(
+      this.#dialog.getByRole("heading", {
+        name: "Reset your recovery phrase?",
```
aterga: Nit. This doesn't sound like a grammatically complete English sentence, maybe we should pick something like "Shall we reset your recovery phrase?" or at least not make it a question, e.g., "Confirm resetting your recovery phrase"

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      }),
+    ).toBeVisible();
+    await this.#dialog.getByRole("button", { name: "Reset" }).click();
```
aterga, line 74: Why no final assertion after clicking the button? 

Ideally, we should have a postcondition here, e.g., "resetting worked fine"

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  async writeDown(): Promise<string[]> {
+    await expect(
+      this.#dialog.getByRole("heading", { name: "Save your recovery phrase" }),
+    ).toBeVisible();
+    await this.#dialog.getByRole("button", { name: "Click to reveal" }).click();
+    // We use selection and copying to the clipboard here to make sure that it
+    // works correctly for power-users that don't physically write things down.
+    await this.#dialog.getByRole("list").selectText();
+    await this.#dialog.press("Meta+c");
```
aterga: Is this cross-platform?

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    const clipboard = await readClipboard(this.#dialog.page());
+    await this.#dialog
+      .getByRole("button", { name: "I have written it down" })
+      .click();
+    return clipboard.trim().split("\n");
```
aterga, line 44: I don't get why this function involves the clipboard

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+class RecoveryPhrasePage {
+  #page: Page;
+
+  constructor(page: Page) {
+    this.#page = page;
+  }
+
+  async activate(options?: { skipVerification?: boolean }): Promise<string[]> {
```
aterga, line 86: Nit - this function is called "activate," but shouldn't it be called, e.g., `check_activation_succeeds`? It seems to be starting with an expectation that the page in question has already been activated.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
sea-snake: Normally I would use a `div` to just draw a vertical line, but since we're inside a `button` that's not a valid child element, so we're using a `span` instead.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
aterga: Never mind, I think I misread the code.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: Yes, Meta is the cross platform key to indicate to playwright we want it work across platforms.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: As mentioned in the code comment above:

> We use selection and copying to the clipboard here to make sure that it
    works correctly for power-users that don't physically write things down.

We could have just grabbed the text content from the html tree to get the words, but that wouldn't have tested that actual (power) users that select the words and copy them, would get the correct clipboard content.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      await dialog.getByRole("button", { name: "Close" }).click();
+    } else {
+      await wizard.verify(words);
+    }
+
+    await expect(dialog).toBeHidden();
+    await expect(
+      this.#page.getByRole("heading", {
+        name:
+          options?.skipVerification === true
```
aterga, line 147: Shouldn't the branches be the other way around? 

If `skipVerification` holds, then `Recovery phrase activated` already here, otherwise `Recovery phrase not verified`

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: It's a dialog title, so it's intentionally short to be scannable by the user as is typical with UI labels.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: Ah yes, we should assert that the `Save your recovery phrase` step is shown. I'll add an assertion.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+test.describe("Recovery phrase", () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(II_URL + "/manage/recovery");
+    await createNewIdentityInII(page, "Test User", dummyAuth());
+    await page.waitForURL(II_URL + "/manage/recovery");
+  });
+
+  test("Activate", async ({ page }) => {
```
aterga:
```suggestion
  test(" can be activated", async ({ page }) => {
```

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    const recoveryPhrasePage = new RecoveryPhrasePage(page);
+    await recoveryPhrasePage.activate();
+  });
+
+  test.describe("Verify", () => {
```
aterga:
```suggestion
  test.describe(" verification: ", () => {
```

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: Edit: added assertion that the next step is shown (thus resetting was successful).

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    test("Unverified (skipped during activation)", async ({ page }) => {
```
aterga:
```suggestion
    test("can be verified after initial activation", async ({ page }) => {
```

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
aterga: Not sure this nesting is needed. Wdyt?

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: This is a post operation assertion, so when verification is skipped we'd expect the UI to indicate that verification is still required.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      const recoveryPhrasePage = new RecoveryPhrasePage(page);
+      const words = await recoveryPhrasePage.activate({
+        skipVerification: true,
+      });
+      await recoveryPhrasePage.verify(words);
+    });
+
+    test("Unverified (skipped during reset)", async ({ page }) => {
```
aterga: I don't understand what this test is covering; could you explain, please?

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: I'll update the tests description to match that style.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      const recoveryPhrasePage = new RecoveryPhrasePage(page);
+      await recoveryPhrasePage.activate();
+      const words = await recoveryPhrasePage.reset({
+        skipVerification: true,
+      });
+      await recoveryPhrasePage.verify(words);
+    });
+
+    test("Retry (incorrect word order)", async ({ page }) => {
```
aterga:
```suggestion
    test("requires correct word order", async ({ page }) => {
```

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      await wizard.verify(words);
+      await expect(dialog).toBeHidden();
+      await expect(
+        page.getByRole("heading", { name: "Recovery phrase activated" }),
+      ).toBeVisible();
+    });
+  });
+
+  test.describe("Reset", () => {
+    test("Activated", async ({ page }) => {
```
aterga:
```suggestion
    test("can be reset after activation", async ({ page }) => {
```

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      const recoveryPhrasePage = new RecoveryPhrasePage(page);
+      await recoveryPhrasePage.activate();
+      await recoveryPhrasePage.reset();
+    });
+
+    test("Unverified (skipped during activation)", async ({ page }) => {
```
aterga:
```suggestion
    test("can be reset if activation was skipped", async ({ page }) => {
```

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      const recoveryPhrasePage = new RecoveryPhrasePage(page);
+      await recoveryPhrasePage.activate({
+        skipVerification: true,
+      });
+      await recoveryPhrasePage.reset({ isUnverified: true });
+    });
+
+    test("Unverified (skipped during reset)", async ({ page }) => {
```
aterga:
```suggestion
    test("can be reset twice in a row", async ({ page }) => {
```

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Verify.svelte
aterga: Okay

src/frontend/tests/e2e-playwright/utils.ts
sea-snake: We insert a temporary textarea element so we can paste the clipboard content into it and then read it's value. It's a workaround for `navigator.clipboard.readText()` since that is blocked by CSP.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
aterga: Okay

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
aterga: Similar questions for 
* `verify`
* `acknowledge`
* `writeDown`

src/frontend/tests/e2e-playwright/utils.ts
sea-snake: I've added code comments to clarify.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
aterga: Okay

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: These functions should only assert if they're seeing the correct view and if they're able to interact with it as expected.

Asserting what happens after within these functions would leak implementation details of the other views into these functions.

Instead this is already handled by the caller that calls these functions one after another, asserting that the next view is visible.

As an alternative approach, I'll add post assertions that verify that the current view title is no longer visible, making sure the view didn't possibly throw an error and got stuck.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: Should be more clear now with updated titles.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: To clarify `skipVerification` means don't verify, exit early instead, resulting in an unverified recovery phrase.

It doesn't mean that the recovery phrase is already verified and the verification can be skipped.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
aterga: > As an alternative approach, I'll add post assertions that verify that the current view title is no longer visible, making sure the view didn't possibly throw an error and got stuck.

Better than nothing, but this condition is still a bit weak. 

In general, we need two kinds of helpers to write these tests: 
* operations
* assertions

Every test should be composed of a sequence like this:
1. assert precondition (optional smoke test)
2. operation 1
3. assert intermediate state 1 (optional)
4. operation 2
5. ...
6. operation N
7. assert postcondition (non-optional; this checks that the test completed in a state that satisfies the success criteria of the test)

For convenience, it is fine to include some intermediate assertions into the helpers (e.g., the helper implementing "operation 1" could also assert intermediate state 1). 

However, I think we still need the postcondition assertions. 

For instance, what does it mean for a recovery phrase to be reset? 

I suggest we implement these assertions as precisely as possible (but not to the point of making the tests fragile, of course). And write them directly as the last statement of each test case. Unless it makes sense to factor them out into their own helpers.

Wdyt?

sea-snake:
> Approving, but let's revisit the postconditions in all test cases. Are they actually ensuring that the test scenario succeeded, or should they be strengthened?

The `RecoveryPhrasePage` methods currently solely rely on the page content to test if a scenario succeeded (basically check if the recovery phrase card has changed).

The inner `RecoveryPhraseWizard` methods rely on a pre and post check to see if the view has changed, towards which particular view is handled by having these methods called in order from within `RecoveryPhrasePage`.

One big missing part in these e2e tests is actually checking if the identity can be recovered with the (activated/reset) recovery phrase. This will be added in a follow-up PR using the lost access flow.

src/frontend/tests/e2e-playwright/dashboard/recoveryPhrase.spec.ts
sea-snake: I think overall, the pre and post conditions should be together with the operations themselves instead of in isolation with e2e tests.

In comparison to integration or unit tests, these operations aren't existing methods being tested and asserted afterwards but should be considered individual "tests" on their own that are orchestrated in a given scenario.

PR #3519, opened 2025-11-27
Recovery phrase unverified status based on `last_authentication`.

sea-snake:
Recovery phrase unverified status based on `last_authentication`.

# Changes

- Renamed `Verify` component to `VerifySelecting`.
- Implemented `RecoveryPhraseInput` component.
- Use above component to implement `VerifyTyping`.
- Update `RecoveryPhraseWizard` to use either `VerifySelecting` or `VerifyTyping` depending on the availability of the valid recovery phrase in memory (e.g. user signed out and in -> not available).
- Update `/manage/recovery` page to use `last_authentication` to decide if a recovery phrase is unverified (null = not used yet).

# Tests

No tests have been updated in this PR, existing e2e tests should pass in the CI/CD pipeline.

# Notes

- The `RecoveryPhraseInput` component will be used on the `use recovery phrase screen` in a later PR.
- Additional e2e tests that cover `VerifyTyping` will be added in a later PR.

### DIFF
--- a/src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
+++ b/src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
@@ -6,13 +6,13 @@
 
   interface Props {
     value: string[];
-    showAll?: boolean;
+    showValues?: boolean;
     disabled?: boolean;
   }
 
   let {
     value = $bindable(),
-    showAll = false,
+    showValues = false,
     disabled = false,
   }: Props = $props();
 
@@ -126,7 +126,7 @@
           "border-border-primary rounded-full",
           "focus:not-disabled:border-fg-primary",
           "not-focus:user-invalid:!border-border-error not-focus:user-invalid:!bg-bg-error-primary/30 not-focus:user-invalid:!pe-7",
-          !showAll &&
+          !showValues &&
             "not-focus:valid:!text-transparent disabled:!text-transparent",
           "disabled:!text-text-disabled disabled:!bg-bg-disabled disabled:!border-border-disabled_subtle",
           word.length > 7 && "tracking-tight",
@@ -156,7 +156,7 @@
           "text-text-primary bg-transparent text-base tracking-tight",
           "peer-disabled:!text-text-disabled",
           word.length > 0 &&
-            !showAll &&
+            !showValues &&
             "peer-not-focus:peer-valid:!block peer-disabled:!block",
         ]}
         aria-hidden="true"
--- a/src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/VerifyTyping.svelte
+++ b/src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/VerifyTyping.svelte
@@ -16,7 +16,7 @@
   const { onCompleted, recoveryPhrase }: Props = $props();
 
   let value = $state<string[]>(recoveryPhrase ?? EMPTY_PHRASE);
-  let showAll = $state(recoveryPhrase !== undefined);
+  let showValues = $state(recoveryPhrase !== undefined);
   let isCheckingPhrase = $state(false);
 
   const phraseValid = $derived(value.every((word) => word.length > 0));
@@ -61,15 +61,15 @@
     <Trans>Enter each word in the correct order:</Trans>
   {/if}
 </p>
-<RecoveryPhraseInput bind:value {showAll} disabled={isCheckingPhrase} />
+<RecoveryPhraseInput bind:value {showValues} disabled={isCheckingPhrase} />
 <div class="more-limited-height mt-5 flex flex-row">
   <Button
-    onclick={() => (showAll = !showAll)}
+    onclick={() => (showValues = !showValues)}
     variant="tertiary"
     disabled={isCheckingPhrase}
     class="flex-1"
   >
-    {showAll ? $t`Hide all` : $t`Show all`}
+    {showValues ? $t`Hide all` : $t`Show all`}
   </Button>
   <Button
     onclick={() => (value = EMPTY_PHRASE)}


src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -92,22 +96,54 @@
...
+      ) {
+        // Authenticated call failed due to incorrect recovery phrase
+        return false;
+      } else {
+        throw error;
+      }
     }
+
+    const data = await recoveryAuthnMethodData(recoveryPhrase);
+    const nowMillis = BigInt(Date.now()) * BigInt(1_000_000);
```
aterga: Why do we need to multiply by `1_000_000`? Isn't `Date.now()` already returning a timestamp in milliseconds?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -92,22 +96,54 @@
...
+    const agent = await HttpAgent.from(anonymousAgent);
+    agent.replaceIdentity(identity);
+    try {
+      await Promise.all([
+        // Make authenticated call with recovery phrase to mark it as used (verified)
+        anonymousActor.identity_info
+          .withOptions({ agent })($authenticatedStore.identityNumber)
+          .then(throwCanisterError),
+        // Artificial delay to improve UX, wait at least 2 seconds even if network is faster.
+        waitFor(2000),
```
aterga, line 118: Could we instead wait for max((2 seconds) - (how long has actually passed), 2 seconds)?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -92,22 +96,54 @@
...
+      ]);
+    } catch (error) {
+      if (
+        isCanisterError<IdentityInfoError>(error) &&
+        error.type == "Unauthorized"
```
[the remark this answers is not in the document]
sea-snake: Good catch!

src/frontend/src/lib/components/wizards/createRecoveryPhrase/CreateRecoveryPhraseWizard.svelte
```diff
@@ -1,63 +1,81 @@
...
+      incorrectRecoveryPhrase = isIncorrect ? entered : undefined;
+    } catch (error) {
+      onError(error);
     }
-    onVerified();
   };
   const retryVerification = () => {
     isWritten = false;
     isIncorrect = false;
   };
```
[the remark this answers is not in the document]
sea-snake: The `Write` component is explicitly only shown if both `isWritten === true` and `recoveryPhrase !== undefined` conditions are true, to avoid this from being a problem. Also both these conditions are needed to make sure the `recoveryPhrase` is narrowed to a type that isn't nullable since the prop is non optional on the `Write` component.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
...
+          !showAll &&
+            "not-focus:valid:!text-transparent disabled:!text-transparent",
+          "disabled:!text-text-disabled disabled:!bg-bg-disabled disabled:!border-border-disabled_subtle",
+          word.length > 7 && "tracking-tight",
+        ]}
+        data-lpignore="true"
+        data-1p-ignore="true"
+        data-bwignore="true"
+        data-form-type="other"
+      />
```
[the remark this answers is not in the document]
sea-snake: Good idea, I'll pick this up though in another PR when I re-use this component in the recovery flow and write e2e tests that rely on this (missing) accessibility label.

Not entirely sure if it's needed though, since the inputs are already within an ordered list (ol -> li). I'll double check what the recommended approach is here from an accessibility perspective.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
+<script lang="ts">
+  import Tooltip from "$lib/components/ui/Tooltip.svelte";
+  import { InfoIcon } from "@lucide/svelte";
+  import { t } from "$lib//stores/locale.store";
```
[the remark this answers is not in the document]
sea-snake: Woops, good catch!

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
...
+        bind:value={
+          () => word,
+          (v) => {
+            words[index] = v.toLowerCase().replace(/[^a-z]/g, "");
+            words = [...words];
+          }
+        }
+        onkeydown={(event) => handleKeyDown(event, index)}
+        onpaste={(event) => handlePaste(event, index)}
+        pattern={dictionary?.join("|")}
```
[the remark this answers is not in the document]
sea-snake: Datalist would result in auto completed suggestions, which is currently not intended by design. Good idea to memoize the variable, I'll create a re-usable $derived variable for it.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
...
+      .trim()
+      .split(/\s/)
+      .map((word) => word.toLowerCase().replace(/[^a-z]/g, ""));
+    if (clipboard.length > 0) {
+      clipboard.forEach((word, i) => (words[index + i] = word));
+      words = [...words];
+      event.preventDefault();
+      wrapperRef
+        ?.querySelectorAll("input")
+        [index + clipboard.length - 1]?.focus();
```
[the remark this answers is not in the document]
sea-snake: Good catch! I'll check bounds before assignment.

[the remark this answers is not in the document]
sea-snake: This is already mentioned in above comment with better fixes than this one (that would unintentionally overwrite words).

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/VerifyTyping.svelte
```diff
@@ -0,0 +1,102 @@
...
+  let value = $state<string[]>(recoveryPhrase ?? EMPTY_PHRASE);
+  let showAll = $state(recoveryPhrase !== undefined);
+  let isCheckingPhrase = $state(false);
+
+  const phraseValid = $derived(value.every((word) => word.length > 0));
+  const autoSubmit = $derived(recoveryPhrase === undefined);
+
+  const handleSubmit = () => {
+    isCheckingPhrase = true;
+    onCompleted(value);
```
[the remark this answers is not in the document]
sea-snake: Ah yes, it should await indeed.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
sea-snake: Yes but we need nanoseconds, seems I called the variable incorrectly by mistake 😅

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
sea-snake: That's what `Promise.all` does here, it waits intentionally at least 2 seconds and possibly longer if the network call takes longer.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
aterga: It's good that you included the intended units in the variable name, though. We should do this more often.

sea-snake: Now I just need to do it correctly 😂

sea-snake: Renamed to `nowNanos`

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -92,22 +96,54 @@
...
+      } else {
+        throw error;
+      }
     }
+
+    const data = await recoveryAuthnMethodData(recoveryPhrase);
+    const nowMillis = BigInt(Date.now()) * BigInt(1_000_000);
+    recoveryPhraseData = {
+      ...data,
+      last_authentication: [nowMillis],
```
aterga: I'm not getting why we have to define this attribute in the FE. Doesn't the BE take care of that?

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
...
+    if (event.code === "Backspace" && words[index].length === 0) {
+      wrapperRef?.querySelectorAll("input")[index - 1]?.focus();
+      event.preventDefault();
+    }
+    if (event.code === "Space" || event.code === "Enter") {
+      wrapperRef?.querySelectorAll("input")[index + 1]?.focus();
+      event.preventDefault();
+    }
+  };
+  const handlePaste = (event: ClipboardEvent, index: number) => {
```
aterga, line 58: Could this function be unit tested?

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
...
+  onMount(() => {
+    // Lazy load dictionary so this component doesn't include it in the bundle eagerly
+    import("bip39").then((bip39) => (dictionary = bip39.wordlists.english));
+    // Focus on first empty input
+    wrapperRef
+      ?.querySelectorAll("input")
+      [words.findIndex((word) => word.length === 0)]?.focus();
+  });
+
+  $effect(() => {
```
aterga: I can't easily understand this function's intention. 

It feels like we're adding a lot of inline functions that do clever things, and have zero tests and zero doc strings for their intended behavior. Am I missing something?

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
...
+  import { onMount } from "svelte";
+
+  interface Props {
+    value: string[];
+    showAll?: boolean;
+    disabled?: boolean;
+  }
+
+  let {
+    value = $bindable(),
```
aterga, line 14: This variable name seems too generic. What value is this?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
sea-snake: This is an optimistic UI update, we're basically creating the data locally and updating the state before the backend does it after invalidateAll().

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -0,0 +1,160 @@
...
+    showAll = false,
```
aterga: Ditto for the other two - show all what? disable what?

aterga: Maybe in Svelte, it's common to assume that everything inside a component file is an attribute of that component. 

So maybe I should think of these local variables as 

* (RecoveryPhraseInput)_value (makes sense)
* (RecoveryPhraseInput)_showAll (still doesn't make sense)
* (RecoveryPhraseInput)_disabled (makes sense)

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: The value passed into `RecoveryPhraseInput`, the html `<input>` also uses a prop named `value`. 

This prop is `$bindable` which means it's value is both ways, it's set from the outside and the outside is updated from the inside.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: Both props are regarding the whole component, the whole component hides or shows the content of all input fields and disabled/enables all input fields inside this component.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: I'll move these two if conditions into variables so it's more clear when the effect doesn't run. The effect itself doesn't do much besides setting a `$bindable` variable, so don't really see how this could be tested.

If we want to test component behavior we should probably look into component tests and/or e2e tests. Regarding docs, I'll move more things into variables and/or add comments where needed.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
aterga: If you have a property called "showAll" to refer to all the components (words), why not call the other two "allValue**s**" and "allDisabled" for consistency? Alternatively, I would rename `showAll` as simply `show`.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
aterga: Seems like this comes with some pros and cons.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: I'm updating it to `showValues` to indicate that the values aren't shown when this is true (the component itself is still visible).

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: Not now.

PR #3521, opened 2025-11-28
Redirect to landing page when unsupported page is visited directly.

sea-snake:
Redirect to landing page when this page is visited directly. Reloading the unsupported page either manually or automatically due to inactivity, does not result in a redirect.

# Changes

- Updated unsupported page with `beforeNavigate` and `afterNavigate` hooks.

# Tests

Manually verified the page behaves as expected, visiting `/#authorize` still redirects to `/unsupported` and is out of scope for this PR.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/unsupported/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/unsupported/+page.svelte
@@ -4,7 +4,7 @@
   import { afterNavigate, beforeNavigate, goto } from "$app/navigation";
   import type { AfterNavigate, BeforeNavigate } from "@sveltejs/kit";
 
-  const SESSION_KEY = "ii-unsupported-page-reloaded";
+  const PAGE_RELOADED_KEY = "ii-unsupported-page-reloaded";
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/routes/(new-styling)/unsupported/+page.svelte
```diff
@@ -1,8 +1,45 @@
 <script lang="ts">
   import XBrowser from "./components/views/XBrowser.svelte";
   import Generic from "./components/views/Generic.svelte";
+  import { afterNavigate, beforeNavigate, goto } from "$app/navigation";
+  import type { AfterNavigate, BeforeNavigate } from "@sveltejs/kit";
+
+  const SESSION_KEY = "ii-unsupported-page-reloaded";
```
aterga: Nit. The variable name implies that this is a session-specific key, whereas we have a constant key for this page. Consider renaming this, e.g.

```suggestion
  const PAGE_RELOAD_KEY = "ii-unsupported-page-reloaded";
```

aterga (approved):

PR #3523, opened 2025-12-01
Refactor recovery phrase e2e and add more test cases.

sea-snake:
Refactor recovery phrase e2e and add more test cases.

# Changes

- Create recovery phrase fixture
- Create identity fixture
- Export both merged fixtures as a test instance
- Use this test instance to refactor recovery phrase tests, directly using the fixtures so we can focus on the logic instead of set-up.
- Reorganize and add recovery phrase tests, making sure all possible flows are tested.
- Add aria-label to recovery phrase input boxes so tests can find and fill these boxes.

### DIFF
--- a/src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
+++ b/src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
@@ -4,6 +4,8 @@ import { expect } from "@playwright/test";
 /**
  * Swap the first word around with the next different word found,
  * compared to random shuffle, this guarantees a different phrase.
+ *
+ * @example swapWordsAround(["abc", "abc", "xyz", "xyz"]) // ["xyz", "abc", "abc", "xyz"]
  */
 const swapWordsAround = (words: string[]) => {
   const incorrectOrder = [...words];
@@ -13,7 +15,7 @@ const swapWordsAround = (words: string[]) => {
   );
   incorrectOrder[0] = incorrectOrder[differentWordIndex];
   incorrectOrder[differentWordIndex] = firstWord;
-  expect(incorrectOrder).not.toEqual(words);
+  expect(incorrectOrder).not.toEqual(words); // Smoke test
   return incorrectOrder;
 };
 
@@ -125,6 +127,23 @@ test.describe("Recovery phrase", () => {
           await wizard.retry();
           await wizard.verifyTyping(words.current!);
         });
+      });
+    });
+
+    test("when signed in (incorrectly) and then when coming back (correctly)", async ({
+      manageRecoveryPage,
+      identity,
+      words,
+    }) => {
+      await manageRecoveryPage.verify(async (wizard) => {
+        await wizard.verifySelecting(swapWordsAround(words.current!));
+        await wizard.close();
+      });
+      await identity.signOut();
+      await manageRecoveryPage.goto();
+      await identity.signIn();
+      await manageRecoveryPage.verify(async (wizard) => {
+        await wizard.verifyTyping(words.current!);
       });
     });
   });


src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -0,0 +1,323 @@
+import { test as base } from "../fixtures";
+import { expect } from "@playwright/test";
+
+/**
+ * Swap the first word around with the next different word found,
+ * compared to random shuffle, this guarantees a different phrase.
+ */
+const swapWordsAround = (words: string[]) => {
+  const incorrectOrder = [...words];
```
aterga, line 11: Same as `words.slice()`?

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -0,0 +1,323 @@
+import { test as base } from "../fixtures";
+import { expect } from "@playwright/test";
+
+/**
+ * Swap the first word around with the next different word found,
+ * compared to random shuffle, this guarantees a different phrase.
```
aterga, line 6: Maybe an example could help in the comment, 

Say the phrase is "abc def abc xyz"

Is the function going to return "def abc abc xyz"?

The fact that the function both checks something and returns something makes it harder to understand its purpose IMHO.

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+ */
+const swapWordsAround = (words: string[]) => {
+  const incorrectOrder = [...words];
+  const firstWord = incorrectOrder[0];
+  const differentWordIndex = incorrectOrder.findIndex(
+    (word) => word !== firstWord,
+  );
+  incorrectOrder[0] = incorrectOrder[differentWordIndex];
+  incorrectOrder[differentWordIndex] = firstWord;
+  expect(incorrectOrder).not.toEqual(words);
```
aterga: Just a smoke test?

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  test.describe("can be activated", () => {
+    test.afterEach(async ({ manageRecoveryPage, identity }) => {
+      await manageRecoveryPage.assertActivated();
+      // Assert it's still activated after signing back in
+      await identity.signOut();
+      await manageRecoveryPage.goto();
+      await identity.signIn();
+      await manageRecoveryPage.assertActivated();
+      // TODO: Verify we can recover using `words.current`
```
aterga, line 48: Please remind me, is this a todo for this PR?

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+        await wizard.verifySelecting(words.current);
+      });
+    });
+
+    test("on retry", async ({ manageRecoveryPage, words }) => {
+      await manageRecoveryPage.activate(async (wizard) => {
+        await wizard.acknowledge();
+        words.current = await wizard.writeDown();
+        await wizard.verifySelecting(swapWordsAround(words.current));
+        await wizard.retry();
```
aterga, line 64: So this tests that after the first recovery phrase verification fails, the user can create a new phrase and verify it, right? 

We could also have a test that shows that the same phrase could be verified after the verification dialogue / the whole website is closed during the first verification attempt. (Reading further)

aterga (approved):
Very thorough (and much clearer) testing. Thanks!

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: Same thing indeed (both create a copy of the array), this is to make sure I'm not modifying the input array.

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: Ah yes, the check is there as a smoke test, so we're sure it actually changed to different one, I'll add a comment.

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: Yup

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: Ah good idea, I'll add it.

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: No, it's a TODO for another PR once I'm diving into the "recover with recovery phrase" flows.

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: Added `when signed in (incorrectly) and then when coming back (correctly)`.

PR #3526, opened 2025-12-02
chore: Fallback for passkeys without credential IDs

aterga:
# Motivation

It is unclear if passkeys without credential IDs could have even been created in production. Normally, a passkey must specify its credential ID to be used.

However, since we cannot rule this case out (due to the legacy `Device` type representing all kinds of credentials), we opt for keeping those passkeys while initializing the missing credential IDs with a fixed set of bytes that could be identified later on if needed.

Additionally, this PR refactors some storage code to enhance readability and improve consistency. 

# Changes

* Set hypothetical missing passkey credential IDs to `0xdeadbeef`. 
* Refactor `Storage.write` to prepare for the switch of the source of truth from `StorableFixedAnchor` to `StorableAnchor`. 

# Tests

* Adjusted `should_handle_devices_without_credential_id`.

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -696,7 +696,7 @@ impl<M: Memory + Clone> Storage<M> {
         // Second, deconstruct the previous anchor, obtaining the previous credentials and recovery keys.
         let (previous_openid_credentials, _previous_passkey_credentials, _previous_recovery_keys) =
             if let Some(StorableAnchor {
-                // The following fileds need to be merged with the previous anchor.
+                // The following fields need to be compared with the previous anchor
                 openid_credentials,
                 passkey_credentials,
                 recovery_keys,


src/internet_identity/src/storage.rs
```diff
@@ -657,53 +658,85 @@ impl<M: Memory + Clone> Storage<M> {
...
-            .map(|anchor| anchor.openid_credentials)
-            .unwrap_or_default();
-        let current_openid_credentials = stable_anchor.openid_credentials;
-        self.update_lookup_anchors_with_openid_credential(
+            .insert(anchor_number, storable_anchor.clone());
+
+        // Second, deconstruct the previous anchor, obtaining the previous credentials and recovery keys.
+        let (previous_openid_credentials, _previous_passkey_credentials, _previous_recovery_keys) =
+            if let Some(StorableAnchor {
+                // The following fileds need to be merged with the previous anchor.
```
sea-snake: Merged is slightly confusing here since we're comparing the fields, not combining them.
```suggestion
                // The following fields need to be compared with the previous anchor
```

PR #3527, opened 2025-12-02
Support for resetting locked recovery phrases.

sea-snake:
Support for resetting locked recovery phrases.

# Changes

- Fix typo in `RecoveryPhraseInput` that caused pasted phrases to be only entered partially.
- Update `CreateRecoveryPhraseWizard` to differentiate between protected and unprotected recovery phrases.
- Add additional `Unlock` view for recovery phrase entry for unlocking it.
- Update recovery phrase utils to use `globalThis`, so they can be used in e2e (Node).
- Update recovery phrase page to handle unlocking and replacing a locked recovery phrase.
- Remove em tags from "How to stay secure" section.
- Update `ActiveRecoveryPhrase` to show "Unlock and reset" button if the recovery phrase is locked. 
- Update `identity` fixture to additionally return method to get an authenticated actor instance.
- Use this actor in e2e tests to create a locked recovery phrase.

# Tests

Added e2e tests for resetting a locked recovery phrase.

https://github.com/user-attachments/assets/229e9649-e28a-4c41-8899-c2a4e539edf5

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
@@ -45,9 +45,6 @@
         getMetadataString(m.metadata, "usage") === "recovery_phrase",
     ),
   );
-  const isCurrentAccessMethod = $derived(
-    "recoveryPhrase" in $authenticatedStore.authMethod,
-  );
   const recoveryPhraseType = $derived(
     recoveryPhraseData !== undefined
       ? "Protected" in recoveryPhraseData.security_settings.protection
@@ -55,9 +52,14 @@
         : "unprotected"
       : undefined,
   );
+  const isCurrentAccessMethod = $derived(
+    "recoveryPhrase" in $authenticatedStore.authMethod,
+  );
   const isUnverified = $derived(
     recoveryPhraseData !== undefined &&
       recoveryPhraseData.last_authentication[0] === undefined &&
+      // A locked recovery phrase is always considered verified
+      // since locking required re-typing the recovery phrase.
       recoveryPhraseType === "unprotected",
   );
 
--- a/src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
+++ b/src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
@@ -260,7 +260,7 @@ test.describe("Recovery phrase", () => {
         await manageRecoveryPage.assertLocked();
       });
 
-      test("on first attempt ", async ({ manageRecoveryPage, words }) => {
+      test("on first attempt", async ({ manageRecoveryPage, words }) => {
         words.current = await manageRecoveryPage.unlockAndReset(
           async (wizard) => {
             await wizard.unlock(words.current!);


src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -201,35 +268,34 @@
...
     </article>
     <article>
       <h3>{$t`Replace it anytime`}</h3>
       <p>
         <Trans>
-          You can <em>reset</em> your recovery phrase at any time. This keeps
-          your identity <em>secure</em> if it is ever exposed.
+          You can reset your recovery phrase at any time. This keeps your
+          identity secure if it is ever exposed.
         </Trans>
```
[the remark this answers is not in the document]
sea-snake: Already fixed.

src/frontend/src/lib/components/wizards/createRecoveryPhrase/CreateRecoveryPhraseWizard.svelte
```diff
@@ -43,39 +47,57 @@
...
       isIncorrect = !(await onVerify(entered));
       incorrectRecoveryPhrase = isIncorrect ? entered : undefined;
     } catch (error) {
       onError(error);
     }
   };
-  const retryVerification = () => {
+  const handleRetry = () => {
     isWritten = false;
     isIncorrect = false;
```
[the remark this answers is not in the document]
sea-snake: This is intentional, once unlocked, it's not necessary to unlock it again when making a mistake in the verification step.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -62,7 +62,7 @@
       .map((word) => word.toLowerCase().replace(/[^a-z]/g, ""));
     if (clipboard.length > 0) {
       clipboard.forEach((word, i) => {
-        if (index + i >= word.length) {
+        if (index + i >= words.length) {
```
aterga, line 65: 🤦 good catch

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
aterga, line 58: the doc string should explain the `index` arg

src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Unlock.svelte
```diff
@@ -0,0 +1,104 @@
...
+    variant="primary"
+    size="xl"
+    disabled={isCheckingPhrase || !phraseValid}
+    class="mt-5"
+  >
+    {$t`Submit`}
+  </Button>
+{/if}
+
+<style>
```
aterga, line 93: NCR

I keep wondering if it's a good idea that the business logic and the styling are so much entangled in our FE code. 

For example, to explain how recovery works for the end user, it shouldn't be important to know why we choose 640px and 570px. 

I wonder if we could organize the code to separate the following concerns: 

1. Structure (what is the hierarchy among all the elements?)
2. Styles (what hints do we give the web browser to decide how to render this structure?)
3. Event handlers (how can the user interact with various elements on this page?)

Here's an attempt to illustrate the concrete issues I have understanding the code:

Example A. Concerns 1 and 2 are entangled, e.g., `<div class="more-limited-height mt-5 flex flex-row">` is just a block that has style-related properties based on its 4 classes; whether it serves any other purpose is unclear. Ideally, we would not even need to add an element just for the sake of style, but if we do, at least one would hope to express it in plain English simply as, e.g., "the div that contains the following button" rather than "the div that has these 4 classes; go figure why."

Example B. 

```
  <Button
    onclick={() => (value = EMPTY_PHRASE)}
    variant="tertiary"
    disabled={isCheckingPhrase}
    class="flex-1"
  >
```

* `onclick` is an event handler (Concern 3).
* `variant` defines the prominence of this element (style-related ==> Concern 2).
* `disabled` is an event handler (what happens upon initialization; Concern 3).
* `class` could mean anything, but in this case I'm guessing it's about the style / rendering of the element (Concern 2).

So, while reading the code and trying to focus on the business logic, the reader stumbles upon this salad of concerns. Bottom line, it's hard to develop a 100% precise understanding of intent just by reading a local code snippet. And analyzing an unbounded chunk of code to gain precise answers to all the potential questions is quite inefficient. 

For reference, one could express this button in two plain English statements: 
* Here is a button that clears the entire recovery phrase input.
* That button is tertiary and has style attributes A, B, C.

Lets' discuss this in a call when we get a chance.

sea-snake: That's basically how tailwind is intended to be used, inline css classes. It's an alternative approach to defining custom classes with properties separately, it avoids the need for a reader to jump back and forth between the markup and stylesheet to understand what is rendered and how it is laid out.

In prior svelte versions, it was more trivial to differentiate between handlers/events `on:click` and data props (for e.g. styling with `variant`). Now svelte has moved in the direction of React, everything is a prop. This does mean that the distinction between these props is no longer clear unless you're aware what each prop does.

As for `div` being used for the sake of style, that's their sole purpose of existence since their early days 😅

sea-snake: Oh and as for the `limited-height` and `more-limited-height` classes, that's unfortunately a workaround since tailwind does not support height based media queries by default.

I could configure height based media queries as [custom variants](https://tailwindcss.com/docs/hover-focus-and-other-states#registering-a-custom-variant), but I'd rather avoid changing the tailwind default configuration more than necessary since that's always an issue when you need to upgrade at some point.

sea-snake: I would argue though that the actual `Button` component might be more redundant and confusing than having a plain html `button` with classes `btn-primary btn-xl` instead of props `variant="primary" size="xl"`. This would make sure components are only used for actual larger separate views, avoiding the unnecessary use of components for the sake of re-usable styling.

Back when I had this discussion last year, the decisions was made to go the other way, create components with props similar to those in the Figma designs (e.g. `size="xl"`). Personally, I don't really agree with that approach, I'd rather avoid props for styling, solely keeping the `class` prop for that, but ultimately they're both valid approaches towards the same goal with pros and cons.

I'd be happy to discuss this in a call :D

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -221,6 +226,68 @@ test.describe("Recovery phrase", () => {
...
+        // Use an actor to create a locked recovery phrase
+        // since this functionality is no longer available.
+        const { actor, identityNumber } = await identity.createActor();
+        words.current = generateMnemonic();
+        const recoveryIdentity = await fromMnemonicWithoutValidation(
+          words.current.join(" "),
+          IC_DERIVATION_PATH,
+        );
+        await actor.authn_method_add(identityNumber, {
+          metadata: [
```
aterga, line 241: is this actually needed?

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -221,6 +226,68 @@ test.describe("Recovery phrase", () => {
...
+      });
+
+      test("on first attempt", async ({ manageRecoveryPage, words }) => {
+        words.current = await manageRecoveryPage.unlockAndReset(
+          async (wizard) => {
+            await wizard.unlock(words.current!);
+            await wizard.confirmReset();
+            const newWords = await wizard.writeDown();
+            expect(newWords).not.toEqual(words.current);
+            await wizard.verifySelecting(newWords);
```
aterga, line 270: Could we assert here that `newWords` can be used to sign in?

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
```diff
@@ -221,6 +226,68 @@ test.describe("Recovery phrase", () => {
...
+            return newWords;
+          },
+        );
+      });
+
+      test("on retry", async ({ manageRecoveryPage, words }) => {
+        words.current = await manageRecoveryPage.unlockAndReset(
+          async (wizard) => {
+            await wizard.unlock(swapWordsAround(words.current!));
+            await wizard.retry();
```
aterga, line 280: Should there be an assertion here to ensure the phrase is still locked after the first attempt?

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -201,35 +270,34 @@
...
       </p>
     </article>
     <article>
       <h3>{$t`Store it safely`}</h3>
       <p>
         <Trans>
-          Write down all <em>24 words</em> correctly and keep them
-          <em>offline</em>. Do <em>not</em> store them in the cloud or on devices.
+          Write down all 24 words correctly and keep them offline. Do not store
```
aterga, line 290: Thank you!

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -145,6 +171,49 @@
...
+      recoveryPhrase.join(" "),
+      IC_DERIVATION_PATH,
+    );
+    const sessionIdentity = await ECDSAKeyIdentity.generate();
+    const delegation = await DelegationChain.create(
+      identity,
+      sessionIdentity.getPublicKey(),
+      // This gives 10 minutes to click the reset button in the dialog,
+      // an error will be thrown if the button is clicked after that.
+      new Date(Date.now() + 10 * 60 * 1000),
```
aterga, line 185: Ideally, we'd have a test to demo what happens when this timeout is reached.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
```diff
@@ -145,6 +171,49 @@
     });
     return true;
   };
+  const handleUnlock = async (recoveryPhrase: string[]): Promise<boolean> => {
```
aterga, line 174: It's not clear from this function's name what is does. 

How about


```suggestion
  const isRecoveryPhraseCorrect = async (recoveryPhrase: string[]): Promise<boolean> => {
```

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
sea-snake: Testing this would be non trivial thing since the expiration error is coming from dfx, so triggering it would require the e2e test code to change the local dfx instance time, which isn't trivial. Changing the browser time in the other direction might work, but that will likely affect all calls besides this one (ingress expiry).

Overall, session expiry isn't properly handled yet in II 2.0 and should be properly picked up, it'll likely need design too (what's the UX if a session is or about to be expired?).

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: All tests in the `can be reset` scope have an `test.afterEach` hook that will test if the phrase has indeed been reset and can be used to sign in (this latter part is still a TODO comment till the recovery phrase flow is done in another PR).

This is why the `unlockAndReset` return value is assigned to `words.current`, this enables the `afterEach` hook to access these words to do these checks.

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: Good idea, right now the implementation doesn't actually "unlock" it, it just stores the phrase in memory to use it later when resetting. But indeed, making sure in this test it's not actually unlocked is a good idea to make sure future changes don't break this.

I'll add it to this test (including an additional logout & login check).

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: No idea, I added the metadata to be consistent with the metadata in II 1.0, which is what matches the scenario of locked recovery phrases in production.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/recovery/+page.svelte
sea-snake: It's not just returning the recovery phrase correct true/false, more importantly it's setting the `lockedRecoveryPhraseIdentity` so it can be used later to replace the locked recovery phrase, since the replace method needs to be called with the locked recovery phrase identity.

Also I'd rather avoid using method names that might be confused with boolean variables that are often used with `$state` 😅

I'll add a comment above the method to better explain what it does and an inline comment above `lockedRecoveryPhraseIdentity` to explain why it's set and what it will be used for.

src/frontend/tests/e2e-playwright/dashboard/recovery.spec.ts
sea-snake: Will do in another PR

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: Will do in another PR

sea-snake (commented):
(Had to comment to close the review that was triggered by respond to a comment...)

PR #3533, opened 2025-12-10
Mobile friendly recovery phrase input

sea-snake:
Mobile friendly recovery phrase input

# Changes

- Move show/hide all, clear all and submit buttons into `RecoveryPhraseInput` component.
- Update the component to no longer use a two way value binding, instead there's a `onSubmit` callback prop.
- Change `Button` into `button` wherever I came across the former.
- Remove `limited-height` classes and use tailwind classes instead e.g. `[@media(max-height:640px)]:hidden`.

# Tests

Existing e2e tests should cover these changes and should pass without any changes to the tests.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -4,6 +4,7 @@ use crate::assets::init_assets;
 use crate::authz_utils::IdentityUpdateError;
 use crate::state::persistent_state;
 use crate::stats::event_stats::all_aggregations_top_n;
+use crate::storage::storable::special_device_migration::SpecialDeviceMigration;
 use anchor_management::registration;
 use authz_utils::{
     anchor_operation_with_authz_check, check_authorization, check_authz_and_record_activity,
@@ -27,7 +28,7 @@ use internet_identity_interface::internet_identity::types::vc_mvp::{
 use internet_identity_interface::internet_identity::types::*;
 use serde_bytes::ByteBuf;
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::time::Duration;
 use storage::account::{AccountDelegationError, PrepareAccountDelegation};
 use storage::{Salt, Storage};
@@ -77,8 +78,20 @@ thread_local! {
     pub(crate) static RECOVERY_PHRASE_MIGRATION_BATCH_ID: RefCell<u64> = const { RefCell::new(0) };
     pub(crate) static RECOVERY_PHRASE_MIGRATION_ERRORS: RefCell<Vec<String>> = const { RefCell::new(Vec::new()) };
     pub(crate) static RECOVERY_PHRASE_MIGRATION_LAST_ANCHOR_ID: RefCell<Option<u64>> = const { RefCell::new(None) };
+    pub(crate) static ANCHOR_MIGRATION_SPECIAL_CASES: RefCell<BTreeMap<AnchorNumber, Vec<SpecialDeviceMigration>>> = const { RefCell::new(BTreeMap::new()) };
 
     static TIMER_ID: RefCell<Option<TimerId>> = const { RefCell::new(None) };
+}
+
+#[query(hidden = true)]
+fn get_anchor_migration_special_cases(anchor_number: AnchorNumber) -> Vec<SpecialDeviceMigration> {
+    ANCHOR_MIGRATION_SPECIAL_CASES
+        .with_borrow(|cases| cases.get(&anchor_number).cloned().unwrap_or_default())
+}
+
+#[query(hidden = true)]
+fn get_anchor_migration_special_cases_keys() -> Vec<AnchorNumber> {
+    ANCHOR_MIGRATION_SPECIAL_CASES.with_borrow(|cases| cases.keys().cloned().collect())
 }
 
 /// Temporary function to list migration errors.
--- a/src/internet_identity/src/storage/anchor.rs
+++ b/src/internet_identity/src/storage/anchor.rs
@@ -4,11 +4,15 @@ use crate::storage::storable::anchor::StorableAnchor;
 use crate::storage::storable::fixed_anchor::StorableFixedAnchor;
 use crate::storage::storable::passkey_credential::StorablePasskeyCredential;
 use crate::storage::storable::recovery_key::StorableRecoveryKey;
-use crate::{IC0_APP_ORIGIN, ID_AI_ORIGIN, INTERNETCOMPUTER_ORG_ORIGIN};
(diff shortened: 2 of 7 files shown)


src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -62,129 +80,211 @@
...
+          }}
+          disabled={isEmpty}
+          class={[
+            "btn btn-tertiary relative",
+            "max-sm:btn-icon  max-sm:after:absolute max-sm:after:-inset-2.5",
+            "sm:flex-1",
+          ]}
+        >
+          <ListXIcon />
+          <span>{$t`Clear all`}</span>
```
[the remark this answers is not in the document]
sea-snake: The `btn-icon` class already takes care of this.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -62,129 +80,211 @@
...
+          disabled={isEmpty}
+          class={[
+            "btn btn-tertiary relative",
+            "max-sm:btn-icon max-sm:after:absolute max-sm:after:-inset-2.5",
+            "sm:flex-1",
+          ]}
+        >
+          {#if showValues}
+            <EyeOffIcon />
+            <span>{$t`Hide all`}</span>
```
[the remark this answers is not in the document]
sea-snake: The `btn-icon` class already takes care of this.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -62,129 +80,211 @@
...
+          {:else}
+            <EyeIcon />
+            <span>{$t`Show all`}</span>
```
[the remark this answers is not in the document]
sea-snake: The `btn-icon` class already takes care of this.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -1,59 +1,77 @@
...
 
-  /** Focus on input (if it can be found for given index) */
-  const focusInput = (index: number) =>
+  /**
+   * Focus on an input field, noop if not found
+   *
+   * @param index of the input field to focus
+   */
+  const focusInput = (index: number) => {
     wrapperRef?.querySelectorAll("input")[index]?.focus();
```
aterga, line 43: could you clarify if it's always safe to call `.focus` on whatever is returned by `wrapperRef?.querySelectorAll("input")[index]?`?

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -1,59 +1,77 @@
...
-  /** Change value into lower case and filter out disallowed characters */
+  };
+  /**
+   * Mask input values to keep only allowed characters
+   *
+   * @param value with disallowed characters
+   * @return lower case alpha string
+   */
   const maskInput = (value: string) =>
     value.toLowerCase().replace(/[^a-z]/g, "");
```
aterga, line 52: This function seems to be removing latin letters, not masking them

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -1,59 +1,77 @@
...
+   *
+   * @param event to identify which key was pressed
+   * @param index of the currently focused input field
+   */
   const handleKeyDown = (event: KeyboardEvent, index: number) => {
-    if (event.code === "Backspace" && words[index].length === 0) {
-      focusInput(index - 1);
+    if (event.key === "Backspace" && value[index].length === 0) {
       event.preventDefault();
+      focusInput(index - 1);
```
aterga, line 62: Is thus supposed to focus the input cursor on the previous input field if there's nothing in the current field and the user presses the Backspace button? 

Doesn't seem to work this way on Mac / Chrome / https://beta.id.ai/recovery

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
```diff
@@ -62,129 +80,211 @@
...
-        aria-hidden="true"
-      >
-        <Tooltip
-          label={$t`Incorrect spelling`}
-          direction="up"
-          align="end"
-          distance="0.5rem"
+              "not-focus:valid:!text-transparent disabled:!text-transparent",
+            // Value too long
+            word.length > 7 && "tracking-tight",
```
aterga, line 166: FYI: Current iOS / Safari UI only fits 3 chars in the input before there's a horizontal overflow

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: That would be the case with `replace(/[a-z]/g, "")` but here it's `replace(/[^a-z]/g, "")`, which means all characters except `a-z` are replaced by `""` (empty string). 

As for other non latin characters (e.g. Cyrillic), the BIP-39 dictionary only contains english words with latin characters, so filtering non latin characters is intentional.

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: This PR doesn't update the recovery page input yet, it's only updating the inputs used in the dashboard (verify typing and unlock).

For the recovery page input, there's this draft PR: https://github.com/dfinity/internet-identity/pull/3535

aterga: Thanks. Then please disregard the comment I left in this PR in the context of https://beta.id.ai/recovery page's appearance

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: See: https://github.com/dfinity/internet-identity/pull/3533#discussion_r2606899125

src/frontend/src/lib/components/views/RecoveryPhraseInput.svelte
sea-snake: A not found input is handled by the optional chaining (`?.`), it would short circuit (noop).

As for calling focus on a html element returned by the query selector, since it's contained within `wrapperRef`, it would never return an input outside the `RecoveryPhraseInput` component. This means that the focus behavior is contained to the component, it won't unexpectedly focus outside of this component.

From a security perspective, focusing isn't a risk in comparison to clicking on behalf of the user (particularly, if this were a click to an element queried from the page).

PR #3535, opened 2025-12-10
Refactor recovery flow according to design

sea-snake:
Refactor recovery flow according to updated design.

# Changes

- Move business logic to recovery page itself and remove recovery phrase inputs and e.g. their focus handlers.
- The user interactions of the recovery flow are now within an isolated wizard component that manages the various views shown based on the state of the flow and takes callback props from the page that make the actual canister calls.

# Tests

In depth e2e tests will be in a follow up PR.

### DIFF
--- a/src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/VerifyTyping.svelte
+++ b/src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/VerifyTyping.svelte
@@ -24,7 +24,7 @@
   };
 </script>
 
-<div class={["my-10", "[@media(max-height:640px)]:hidden"]}>
+<div class="my-10 max-sm:hidden">
   {#if isCheckingPhrase}
     <StepsProgressBar total={3} />
   {:else}
--- a/src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Write.svelte
+++ b/src/frontend/src/lib/components/wizards/createRecoveryPhrase/views/Write.svelte
@@ -14,7 +14,7 @@
   let isRevealed = $state(false);
 </script>
 
-<div class={["my-10", "[@media(max-height:700px)]:hidden"]}>
+<div class="my-10 max-sm:hidden">
   <Steps total={3} current={2} />
 </div>
 <h2 class="text-text-primary text-2xl font-medium">


src/frontend/src/routes/(new-styling)/recovery/+page.svelte
```diff
@@ -1,482 +1,141 @@
...
+      await authenticationStore.set({
+        identity: await authenticateWithSession({ session: { identity } }),
+        identityNumber,
+        authMethod: {
+          recoveryPhrase: { principal: identity.getPrincipal() },
+        },
+      });
+      // Set name if identity didn't have a name yet (upgrading from II 1.0)
+      if (newName !== undefined) {
+        await $authenticationStore?.actor
```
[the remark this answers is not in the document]
sea-snake: This isn't true (`authenticationStore !== $authenticationStore), the latter is the store value (from a svelte subscription with $), while the former is a direct reference to the store instance.

src/frontend/src/lib/components/wizards/recoverIdentity/views/MissingName.svelte
```diff
@@ -0,0 +1,95 @@
...
+      continueInProgress = false;
+    }
+  };
+
+  onMount(() => {
+    inputRef?.focus();
+  });
+</script>
+
+<form class="flex flex-col">
```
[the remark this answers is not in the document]
sea-snake: This isn't relevant in svelte kit, the handler on the button with type `submit` will be called on e.g. pressing enter in the input box.

src/frontend/src/lib/components/wizards/recoverIdentity/views/IdentityNotFound.svelte
```diff
@@ -0,0 +1,32 @@
...
+
+<FeaturedIcon variant="error" size="lg" class="mb-4">
+  <UserXIcon class="size-6" />
+</FeaturedIcon>
+<h2 class="text-text-primary mb-3 text-2xl font-medium">
+  {$t`No identity found`}
+</h2>
+<p class="text-text-tertiary mb-8 text-base font-medium text-pretty">
+  <Trans>
+    Your recovery phrase is valid but not for Internet Identity. You may have
```
aterga:
```suggestion
    This is not a valid Internet Identity recovery phrase. Could it be from a different product?
```

src/frontend/src/lib/components/wizards/recoverIdentity/views/EnterRecoveryPhrase.svelte
```diff
@@ -0,0 +1,35 @@
...
+</script>
+
+<h2 class="text-text-primary mb-3 text-2xl font-medium">
+  {$t`Enter recovery phrase`}
+</h2>
+<p class="text-text-tertiary mb-6 text-base font-medium">
+  {#if isSubmitting}
+    <Trans>This may take a few seconds</Trans>
+  {:else}
+    <Trans>Type each word in the correct order:</Trans>
```
aterga:
```suggestion
    <Trans>Please type each word in order:</Trans>
```

src/frontend/src/lib/components/wizards/recoverIdentity/views/InvalidChecksum.svelte
```diff
@@ -0,0 +1,29 @@
...
+</script>
+
+<FeaturedIcon variant="error" size="lg" class="mb-4">
+  <MessageSquareXIcon class="size-6" />
+</FeaturedIcon>
+<h2 class="text-text-primary mb-3 text-2xl font-medium">
+  {$t`Something is wrong!`}
+</h2>
+<p class="text-text-tertiary mb-8 text-base font-medium">
+  <Trans>Incorrect word order. Review and try again.</Trans>
```
aterga:
```suggestion
  <Trans>Incorrect word order. Please review and try again.</Trans>
```

src/frontend/src/lib/components/wizards/recoverIdentity/views/InvalidChecksum.svelte
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  const { onRetry, onCancel }: Props = $props();
+</script>
+
+<FeaturedIcon variant="error" size="lg" class="mb-4">
+  <MessageSquareXIcon class="size-6" />
+</FeaturedIcon>
+<h2 class="text-text-primary mb-3 text-2xl font-medium">
+  {$t`Something is wrong!`}
```
aterga: Is there nothing better we could tell the user in this case?

src/frontend/src/lib/components/wizards/recoverIdentity/views/MissingName.svelte
```diff
@@ -0,0 +1,95 @@
...
+  <FeaturedIcon variant="success" size="lg" class="mb-4">
+    <UserCheckIcon class="size-6" />
+  </FeaturedIcon>
+  <h2 class="text-text-primary mb-3 text-2xl font-medium">
+    {$t`We have found your identity`}
```
aterga:
```suggestion
    {$t`We found your identity`}
```

src/frontend/src/lib/components/wizards/recoverIdentity/views/MissingName.svelte
```diff
@@ -0,0 +1,95 @@
...
+  </h2>
+  <p class="text-text-primary mb-6 text-base font-bold">
+    {identityNumber.toString()}
+  </p>
+  <p class="text-text-tertiary mb-4 text-base font-medium">
+    <Trans>
+      Name your identity to upgrade to II 2.0 and gain access. You can't rename
```
aterga:
```suggestion
      Name your identity to upgrade. Warning: the name cannot be changed later.
```

sea-snake: I'd prefer to avoid "Warning" here, I've changed it to below, wdyt?
> Enter a name for your identity to complete the upgrade. Once set, it cannot be changed later.

src/frontend/src/lib/components/wizards/recoverIdentity/views/MissingName.svelte
sea-snake: Updated to be even shorter (also avoids we) since it's supposed to be a short label title wdyt:
> Identity found

src/frontend/src/lib/components/wizards/recoverIdentity/views/InvalidChecksum.svelte
sea-snake: Actually "review" is incorrect here since we can't show the correct phrase in comparison to the flow where we activate and verify a recovery phrase (retry shows the correct phrase for review there).

Since there's a retry button, asking to try again is probably redundant, combining a more helpful description and clearer title results in this, wdyt?

image

src/frontend/src/lib/components/wizards/recoverIdentity/views/InvalidChecksum.svelte
sea-snake: See discussion in other comment regarding description of this view.

src/frontend/src/lib/components/wizards/recoverIdentity/views/EnterRecoveryPhrase.svelte
sea-snake: Not sure if "please" is a better trade off than "correct", the former doesn't add anything for the user besides politeness. The latter make it clear that words need to be in the correct order, particularly relevant when a user is retrying after making a mistake earlier.

src/frontend/src/lib/components/wizards/recoverIdentity/views/IdentityNotFound.svelte
sea-snake: > Could it be from a different product?
I would avoid making this a question since the implementation is rather absolute, it's either an II recovery phrase or not.

Making this a question would add uncertainty for the user if they did something wrong or if II has an issue. It seems the current phrasing was also not as absolute as it should have been (may have).

Also to clarify the "Your recovery phrase is valid" part was there to make clear it's not a word order/typo thing, but a wrong phrase for the wrong product thing.

What about the following, this is absolute, short and clear that it's entered correctly but for the wrong product:
> The phrase was entered correctly but belongs to a product other than Internet Identity.

PR #3544, opened 2025-12-16
Show users where their passkey is stored

sea-snake:
Show users where their passkey is stored.

# Changes

- Update passkey creation flows to include aaguid and omit alias in AuthnMethodData sent to the canister.
- Update `PasskeyItem` component to indicate where a passkey is stored (e.g. Apple Passwords).
- Update `RenamePasskey` component to allow for an empty input field if the passkey is a known provider (to remove a custom name).

# Tests

- The e2e tests have been updated to now work with the fallback passkey label "Unknown" instead of "Chrome". Previously passkeys were given an alias inferred from the browser but this PR omits the alias, since the AAGUID will suffice.

### DIFF
--- a/src/frontend/src/lib/assets/aaguid/index.ts
+++ b/src/frontend/src/lib/assets/aaguid/index.ts
@@ -14,8 +14,8 @@ import knownSecurityKeys from "./security_keys.json";
 /**
  * e.g. Stored in your Google account and synced across your Android devices.
  */
-interface StoredInAccount {
-  type: "account";
+interface StoredInCloud {
+  type: "cloud";
   account: string;
   platform?: string;
 }
@@ -23,8 +23,8 @@ interface StoredInAccount {
 /**
  * e.g. Stored and usable only on the Windows device it was created on.
  */
-interface StoredOnPlatform {
-  type: "platform";
+interface StoredInOperatingSystem {
+  type: "os";
   platform: string;
 }
 
@@ -45,8 +45,8 @@ interface StoredInBrowser {
 }
 
 export type Provider = { name: string } & (
-  | StoredInAccount
-  | StoredOnPlatform
+  | StoredInCloud
+  | StoredInOperatingSystem
   | StoredOnDevice
   | StoredInBrowser
 );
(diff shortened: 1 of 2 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/components/RenamePasskey.svelte
```diff
@@ -5,86 +5,107 @@
...
-    <Button
-      onclick={props.onCancel}
-      variant="tertiary"
-      size="lg"
-      disabled={isSubmitting}
-    >
-      {$t`Cancel`}
-    </Button>
-  </div>
+  <fieldset class="flex flex-1 flex-col" disabled={isSubmitting}>
```
sea-snake, line 64: The html structure has two changes (which this diff doesn't clearly indicate):
1. Wrap everything in fieldset so everything can be disabled in a single location while submitting.
2. Disable submit button with an empty input field only when there's no provider name (existing passkeys without AAGUID).

src/frontend/src/lib/utils/webAuthn.test.ts
```diff
@@ -28,8 +28,10 @@ describe("webauthn", () => {
     ]);
 
     it("should return valid AAGUID", () => {
-      expect(extractAAGUID(authData)).toBe(
-        "fbfc3007-154e-4ecc-8c0b-6e020557d7bd", // iCloud Keychain
+      expect(extractAAGUID(authData)).toEqual(
+        new Uint8Array([
```
aterga, line 33: Why not keep the comment? iCloud Keychain

src/frontend/src/lib/utils/webAuthn.ts
```diff
@@ -46,18 +46,28 @@ export const constructIdentity = async ({
  * @param authData from which to extract AAGUID
  * @returns AAGUID or undefined
  */
-export const extractAAGUID = (authData: Uint8Array): string | undefined => {
+export const extractAAGUID = (authData: Uint8Array): Uint8Array | undefined => {
   if (authData.byteLength < 53) {
     return;
   }
-  const aaguid = [...authData.slice(37, 53)]
+  const aaguid = authData.slice(37, 53);
```
aterga, line 53: Can we explain where the indices come from in a comment?

aterga: NCR since this was like so before

PR #3547, opened 2025-12-16
fix: Adjust anchor migration to correctly handle key type unknown

aterga:
# Motivation

We discovered that passkeys with key type unknown are in fact a legitimate case (not inconsistent data), originating from devices that for some reason do not follow the recommendation to specify either the platform or cross-platform key type. This PR thus adjusts the device migration logic to reflect this.

# Changes

* Make `KeyType::Unknown` indicate legitimate passkeys.

# Tests

* Added `should_handle_unknown_key_type_with_credential_id` to cover migrating normal and recovery passkeys with unknown key type.


sea-snake (approved):
LGTM

PR #3557, opened 2025-12-23
Redesign sign-in on landing page.

sea-snake:
Redesign sign-in on landing page.

# Changes

- Remove `LandingHeader` component,  it's only used in one place (landing page) and can be replaced with a single div and some classes. 

- `IdentitySwitcher` component changes:
   - Rename `onLogout` prop to `onSignOut`.
   - Rename `switchIdentity` prop to `onSwitchIdentity`.
   - Rename `useAnotherIdentity ` prop to `onUseAnotherIdentity`.
   - Add `onError` prop, to align with error handling in other components.
   - Add `onManageIdentity` prop, to allow for custom behavior.
   - Make `onClose` prop required, since it's always used and needed to begin with.
   - Make all `on` props  functions that return a promise
   - Add handler functions and rendering spinners around these `on` functions.

- `Popover` component changes:
  -  Remove `closeOnOutsideClick` prop, it was never set to false (true by default).
  - Change popover behavior from `auto | manual` to `manual`.
  - This enables the `onClose` prop to be conditional, e.g. it's now used to _not_ change the popover shown state in the parent component if the user is currently signing in.

- `AuthWizard` component changes:
   - Rename `onMigration ` prop to `onUpgrade`.
   - Make all `on` props  functions that return a promise
   - Update handler functions and around these `on` functions to prevent closing a dialog during authentication.
   
- Landing page changes:
  - Remove language switcher from header following latest design.
  - Add navigation handler to open sign-in dialog if redirected from `/login`.
  - Move all authentication logic to `handleSignIn`, having other `handle` methods call it and e.g. additionally show a toast message.
  - Move authentication funnel calls out of these methods into their own separate navigate/effect functions.
  - Move `For developers` button from header to below hero content.
  - Simplify header, make it mostly look the same on desktop and mobile. Only difference remaining is the logo with/without `Internet Identity` label on desktop/mobile.
  -Replace `Manage identity` button with either `Sign in` button or identity switcher, depending on first visit or not.

- Replace login page with a redirect to landing page that includes a `login` state indicator to trigger the sign-in dialog.
  
# Tests

All existing tests should pass, some tests have been updated to align with the new sign in flow (it's now on the landing page instead of it's own login page).

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
@@ -43,6 +43,7 @@
   import Avatar from "$lib/components/ui/Avatar.svelte";
   import { Trans } from "$lib/components/locale";
   import { getMetadataString } from "$lib/utils/openID";
+  import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
 
   const { children, data }: LayoutProps = $props();
 


src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
```diff
@@ -17,106 +16,168 @@
...
+    } catch (error) {
+      onError(error);
+    } finally {
+      switchingToIdentity = undefined;
+    }
+  };
+
+  const handleManageIdentity = async () => {
+    if (onManageIdentity === undefined) {
+      window.open(manageIdentityUrl, "_blank");
```
aterga, line 63: Why do we need this?

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
```diff
@@ -17,106 +16,168 @@
...
+      isNavigatingToManage = true;
+      await onManageIdentity();
+    } catch (error) {
+      onError(error);
+    } finally {
+      isNavigatingToManage = false;
+    }
+  };
+
+  const handleSignOut = async () => {
```
aterga, line 76: Seems like a common pattern, NCR though

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
```diff
@@ -38,14 +36,15 @@
...
 
   const handleContinueWithExistingPasskey = async (): Promise<
     void | "cancelled"
   > => {
-    isAuthenticating = true;
     try {
-      onSignIn(await authFlow.continueWithExistingPasskey());
+      isAuthenticating = true;
+      await onSignIn(await authFlow.continueWithExistingPasskey());
     } catch (error) {
```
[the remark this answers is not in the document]
aterga: Should be covered by the finally block.

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -23,7 +22,74 @@
...
+    Object.values($lastUsedIdentitiesStore.identities)
+      .sort((a, b) => b.lastUsedTimestampMillis - a.lastUsedTimestampMillis)
+      .slice(0, 3),
+  );
+  const selectedIdentity = $derived($lastUsedIdentitiesStore.selected);
+
+  const handleSignIn = async (identityNumber: bigint) => {
+    isAuthenticating = true;
+    if ($authenticationStore?.identityNumber !== identityNumber) {
+      // Switch sign in if not authenticated with this identity yet
```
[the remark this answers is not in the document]
sea-snake: No actually this isn't a typo, if the identity number doesn't match it could also indicate the user isn't signed in with the correct identity number hence switch.

src/frontend/src/lib/components/wizards/auth/AuthWizard.svelte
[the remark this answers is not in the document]
sea-snake: Its reset in the `finally` clause...

src/frontend/src/routes/(new-styling)/login/+page.svelte
```diff
@@ -1,197 +1,11 @@
...
-  };
-
   onMount(() => {
-    authenticationV2Funnel.init({
-      origin: window.location.origin,
-      abTestGroup: $LARGE_GOOGLE_BUTTON
-        ? "largeGoogleButton"
-        : "smallGoogleButton",
-    });
+    goto("/", { state: { login: data.next }, replaceState: true });
```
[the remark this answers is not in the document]
sea-snake: The `data.next` state is used later on with `goto` that doesn't support anything other than relative paths by design. But indeed, it can't hurt to sanitize this data early one so no mistakes can be made with it later. I'll directly sanitize it in the load function to do so.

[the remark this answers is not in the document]
sea-snake: See: https://github.com/dfinity/internet-identity/pull/3557/changes#diff-9e77cc3ad37f288708f933cc9508fcef208bbfa27c138b069fea035aff48910fR4

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
```diff
@@ -17,106 +16,168 @@
...
-  <Button href="/manage" target="_blank" variant="tertiary">
-    <Settings class="size-5" />
-    <span>{$t`Manage Identity`}</span>
-  </Button>
-{/if}
+            {#if selected === identity.identityNumber}
+              <Checkbox
+                checked
+                size="md"
+                class="pointer-events-none mr-1 ml-auto !rounded-full"
```
aterga: `pointer-events-none` is another class responsible for more than just visualization, right?

src/frontend/src/routes/(new-styling)/login/+page.svelte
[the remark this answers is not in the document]
aterga: Would be nice to use the type system to track where you already have sanitized data.

Here it should then be, e.g., 

```
    goto("/", { state: { login: valid_data.next }, replaceState: true });
```

src/frontend/src/routes/(new-styling)/login/+page.ts
```diff
@@ -1,5 +1,19 @@
 import type { PageLoad } from "./$types";
 
 export const load: PageLoad = ({ url }) => {
-  return { next: url.searchParams.get("next") };
+  // Sanitize the next search param, only accept relative paths
```
aterga, line 4: (repeating an idea fro above) 

The best sanitization is also a (unique) constructor of an object of type `ValidT` for your daat of type `T`.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -205,71 +202,70 @@
...
-      useAnotherIdentity={() => {
+      onSwitchIdentity={handleSignIn}
+      onUseAnotherIdentity={() => {
         isIdentityPopoverOpen = false;
         isAuthDialogOpen = true;
       }}
+      onError={(error) => {
+        isIdentityPopoverOpen = false;
+        isAuthenticating = false;
+        handleError(error);
```
aterga, line 401: This seems like a suboptimal way to write the error handler. 

How does the reader supposed to remember what happens _here_ vs. `handleError`? 

Each UI component should clearly define its state machine, and resetting it should be clearly defined in one place (as returning to the initial state, not an arbitrary sequence of low-level operations, because this is prone to errors). 

Since this isn't immediately actionable, NCR for this PR.

src/frontend/tests/e2e-playwright/fixtures/identity.ts
```diff
@@ -71,14 +65,31 @@ class IdentityWizard {
...
-      this.#page.getByRole("button", { name: "Manage identity" }),
-      this.#page.getByRole("button", { name: "Switch identity" }),
-      this.#page.getByRole("button", { name: "Use another identity" }),
+      "Sign in",
+      "Switch identity",
+      "Use another identity",
+      "Continue with passkey",
     ];
+    await this.#page
+      .getByRole("button", { name: new RegExp(buttons.join("|")) })
```
aterga, line 76: Don't really need a regex to find the first word that is in given set. NCR

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -23,7 +22,74 @@
...
+  let next = $state("/manage");
+  let isAuthDialogOpen = $state(false);
+  let isIdentityPopoverOpen = $state(false);
+  let isAuthenticating = $state(false);
+  let identityButtonRef = $state<HTMLButtonElement>();
+
+  const lastUsedIdentities = $derived(
+    Object.values($lastUsedIdentitiesStore.identities)
+      .sort((a, b) => b.lastUsedTimestampMillis - a.lastUsedTimestampMillis)
+      .slice(0, 3),
```
aterga: Maybe add a comment to justify why you take the first 3 items

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -23,7 +22,74 @@
...
+    await goto(next, { replaceState: true });
+    isIdentityPopoverOpen = false;
+    isAuthDialogOpen = false;
+    isAuthenticating = false;
+  };
+  const handleUpgrade = async (identityNumber: bigint) => {
+    await handleSignIn(identityNumber);
+    toaster.success({
+      title: $t`Upgrade completed successfully`,
+      duration: 4000,
```
aterga, line 83: How did you pick this value?

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -23,7 +22,74 @@
...
+    });
+  };
+  const handleSignUp = async (identityNumber: bigint) => {
+    await handleSignIn(identityNumber);
+    toaster.success({
+      title: $t`You're all set. Your identity has been created.`,
+      duration: 2000,
```
aterga, line 90: Why is this 2x shorter than the above?

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -40,6 +106,43 @@
...
+      origin: window.location.origin,
+    });
+  });
+
+  // Close authentication funnel once completed
+  beforeNavigate((navigation) => {
+    if (!navigation.to?.url.pathname.startsWith("/manage")) {
+      return;
+    }
+    authenticationV2Funnel.trigger(AuthenticationV2Events.GoToDashboard);
```
aterga, line 143: Why do we need this?

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -459,6 +569,68 @@
...
+    anchor={identityButtonRef}
+    onClose={() => {
+      if (isAuthenticating) {
+        return;
+      }
+      isIdentityPopoverOpen = false;
+    }}
+    direction="down"
+    align="end"
+    distance="0.75rem"
```
aterga, line 613: Shouldn't this be in the CSS?

aterga (approved):

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
sea-snake: The `IdentitySwitcher` has `onManageIdentity` as optional prop, so if the prop is not passed in, we need a fallback behavior.

src/frontend/tests/e2e-playwright/fixtures/identity.ts
sea-snake: You do unfortunately with the `getByRole` function in PlayWright since this code isn't actually executed "as is" but transformed into more complex HTML node matching code down the line in the internal PlayWright implementation.

I could use multiple `getByRole` calls instead for each button label, but then not all of those (as expected) would actually match, and thus some timeout, resulting in a failed test.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: Ah yes, I'll add a comment.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: It's an arbitrary value, back when the first toasts were implemented we decided upon 2 or 4 seconds for most toasts, where the latter is only used for toasts that are less common but important for users to see.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: See above comment.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: This event completes the last step in the auth funnel in Plausible.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: The `Popover` component uses JS to calculate it's position based on it's component props. Ideally, I'd like to avoid the need for a `Popover` component altogether but the modern CSS only alternative [CSS anchor positioning](https://web.dev/learn/css/anchor-positioning) isn't supported in all browsers yet.

Thankfully, the one remaining browser without support, Firefox should soon support it directly (no longer behind a dev flag). Looking forward to phase out the `Popover` component with simple css classes once that's the case.

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
sea-snake: It's a class that make the element only visual (non interactive) while still displaying it, the respective CSS rule is `pointer-events: none;`.

src/frontend/src/routes/(new-styling)/login/+page.ts
sea-snake: Since TS types are structural/duck typed, there's no way to prevent another type with the same shape from being used, in comparison to Rust where every struct is actually unique.

You might hack around that with e.g. JS symbols but that's a bit too much over engineering for the problem at hand here.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
sea-snake: To clarify, `handleError` method naming is suboptimal, it basically is a function that means "take any error and display it to the user in a toast, ideally in human readable language".

As for the multiple state variables being reset, yes there's probably a cleaner way of doing it.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: Added a comment

PR #3559, opened 2025-12-24
Language selector redesign.

sea-snake:
Language selector redesign.

# Changes

- Update `Radiobox` component implementation to match `Checkbox`.
- Implement `ChooseLanguage` view that lists all available languages as radio boxes in their native and translated format.
- Use this view in dialogs on the landing and dashboard pages, adding a new sidebar item in the dashboard.
- Remove language selector from header on landing page.
- Add `setOrReset` method to locale store, switches back to locale auto-detection if locale matches browser.
- Remove `localeOptions` from locale store since it's no longer used.

# Tests

Verified that language can be now be changed on both the landing page and dashboard. Also verified that the `Intl.DisplayNames` browser API is supported in all browsers.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
@@ -24,13 +24,24 @@
   import Dialog from "$lib/components/ui/Dialog.svelte";
   import AuthWizard from "$lib/components/wizards/auth/AuthWizard.svelte";
   import { sessionStore } from "$lib/stores/session.store";
-  import { formatDate, t } from "$lib/stores/locale.store";
+  import {
+    formatDate,
+    locales,
+    localeStore,
+    t,
+  } from "$lib/stores/locale.store";
   import Logo from "$lib/components/ui/Logo.svelte";
   import NavItem from "$lib/components/ui/NavItem.svelte";
   import { SOURCE_CODE_URL, SUPPORT_URL } from "$lib/config";
   import type { LayoutProps } from "./$types";
   import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
   import ChooseLanguage from "$lib/components/views/ChooseLanguage.svelte";
+  import { nanosToMillis } from "$lib/utils/time";
+  import { lastUsedIdentityTypeName } from "$lib/utils/lastUsedIdentity";
+  import ButtonCard from "$lib/components/ui/ButtonCard.svelte";
+  import Avatar from "$lib/components/ui/Avatar.svelte";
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/views/ChooseLanguage.svelte
```diff
@@ -0,0 +1,53 @@
...
+
+  interface Props {
+    locales: string[];
+    value: string;
+    onChange: (value: string) => void;
+  }
+
+  const { locales, value, onChange }: Props = $props();
+
+  let selected = $derived(value);
```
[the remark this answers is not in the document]
sea-snake: This is untrue, `$derived` can be temporarily overridden until it's inner value is updated, ideal for optimistic UI updates or in this case state that's synced from outside the component through props.

src/frontend/src/lib/components/views/ChooseLanguage.svelte
```diff
@@ -0,0 +1,53 @@
...
+    }
+    onChange(selected);
+  });
+</script>
+
+<h2 class="text-text-primary mb-3 text-2xl">
+  {$t`Language`}
+</h2>
+<p class="text-text-tertiary mb-6 text-base">
+  <Trans>choose language to continue</Trans>
```
aterga, line 28: This itself isn't translated to Spanish right now.

aterga: Also (nit)


```suggestion
  <Trans>Select a language to continue</Trans>
```

src/frontend/src/lib/components/views/ChooseLanguage.svelte
```diff
@@ -0,0 +1,53 @@
...
+  $effect(() => {
+    if (value === selected) {
+      return;
+    }
+    onChange(selected);
+  });
+</script>
+
+<h2 class="text-text-primary mb-3 text-2xl">
+  {$t`Language`}
```
aterga, line 25: This label should maybe also be translated?

src/frontend/src/lib/components/views/ChooseLanguage.svelte
sea-snake: Instruction subtitles in these dialogs are lower case (it's a design choice that's also applied in other places).

As for missing translations, I've stopped updating these for now since that would push us past the 2MB limit.

src/frontend/src/lib/components/views/ChooseLanguage.svelte
sea-snake: It's translated, but as mentioned above, the translation files are not currently updated.

PR #3587, opened 2026-01-20
chore: Gracefully report device add errors

aterga:
# Motivation

Avoid confusing users who attempt to add too many devices.

# Changes

High-level: Don't panic, just error out / handle it gracefully in the FE. 

The function `authn_method_add` now distinguishes between the following error cases: 

* InvalidMetadata
* Unauthorized
* StorageError
* InternalCanisterError
* AuthnMethodLimitReached

# Tests

Manual, pending.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -1397,21 +1397,22 @@ candid::export_service!();
 
 #[cfg(test)]
 mod test {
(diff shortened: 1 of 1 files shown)


aterga:
Switching from traps to errors changes the state-unroll behavior, so we need more work to enable graceful errors.

PR #3600, opened 2026-01-28
feat: OpenID attribute certification

aterga:
# Motivation

Support OpenID attribute certification.

# Changes

* Added new interface types and validation logic for attribute preparation/retrieval requests and responses.
* Implemented canister-side attribute certification/signature preparation and exposed new prepare_attributes / get_attributes Candid methods.
* Added supporting plumbing (account lookup helper + error type, exported OpenID session duration constant, new dependencies, updated .did).

# Tests

* Extensive unit tests for stateless functions.
* PocketIC integration test.

### DIFF
--- a/src/internet_identity/tests/integration/attributes.rs
+++ b/src/internet_identity/tests/integration/attributes.rs
@@ -12,6 +12,7 @@ use std::time::Duration;
 #[test]
 fn should_get_certified_attributes() {
     let env = env();
+    #[allow(unused_variables)]
     let (jwt, salt, claims, test_time, test_principal, test_authn_method) =
         crate::openid::openid_google_test_data();
 


src/internet_identity/src/attributes.rs
```diff
@@ -0,0 +1,227 @@
...
+fn expiration_timestamp_ns(issued_at_timestamp_ns: Timestamp) -> Timestamp {
+    issued_at_timestamp_ns.saturating_add(ATTRIBUTES_CERTIFICATION_SESSION_DURATION_NS)
+}
+
+impl Anchor {
+    pub fn get_attributes(
+        &self,
+        mut attributes: BTreeMap<Option<AttributeScope>, BTreeSet<Attribute>>,
+        account: Account,
+        issued_at_timestamp_ns: Timestamp,
```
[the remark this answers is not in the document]
aterga: `CertificateNotFound` is redundant, removing it.

src/internet_identity/src/attributes.rs
```diff
@@ -0,0 +1,227 @@
...
+    ) -> BTreeMap<OpenIdCredentialKey, Vec<Attribute>> {
+        self.openid_credentials
+            .iter()
+            .filter_map(|openid_credential| {
+                // E.g., `openid:google.com`
+                let scope = Some(AttributeScope::OpenId {
+                    issuer: openid_credential.iss.clone(),
+                });
+                // E.g., {`email`, `name`}
+                let attribute_names = requested_attributes.remove(&scope)?;
```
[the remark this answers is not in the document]
aterga: My explanation of why the code is correct:

Each certification request is performed for a particular anchor. For each anchor, there is a unique Subject Identifier (`sub`).

Thus, the requested attributes (for a given anchor) can be keyed just by the Issuer Identifier (`iss`).

src/internet_identity/src/attributes.rs
```diff
@@ -0,0 +1,227 @@
...
+                            let inputs = CanisterSigInputs {
+                                domain,
+                                seed: &seed,
+                                message: &message,
+                            };
+
+                            let signature = sigs
+                                .get_signature_as_cbor(&inputs, Some(certified_assets.root_hash()))
+                                .ok()?;
+
```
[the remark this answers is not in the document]
aterga: To figure out for which attributes there has not been a signature found in the II backend, the client can compare the list of requested attributes to the list of certified attributes. Thus, `CertificateNotFound` is superfluous, and I'm dropping it.

src/internet_identity_interface/src/internet_identity/types/attributes.rs
```diff
@@ -0,0 +1,821 @@
...
+    pub account_number: Option<AccountNumber>,
+    pub issued_at_timestamp_ns: Timestamp,
+    pub attributes: BTreeMap<Option<AttributeScope>, BTreeSet<Attribute>>,
+}
+
+impl TryFrom<GetAttributesRequest> for ValidatedGetAttributesRequest {
+    type Error = GetAttributesError;
+
+    fn try_from(value: GetAttributesRequest) -> Result<Self, Self::Error> {
+        let GetAttributesRequest {
```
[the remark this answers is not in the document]
aterga: Will add more tests.

src/internet_identity/src/account_management.rs
```diff
@@ -71,7 +71,46 @@ fn try_read_account_info(
...
+
+pub fn get_account_for_origin(
+    anchor_number: AnchorNumber,
+    origin: FrontendHostname,
+    account_number: Option<AccountNumber>,
+) -> Result<Account, GetAccountError> {
+    // If the account_number is not specific, there is a *synthetic account* to return for any
+    // (anchor, origin) pair.
+    let Some(account_number) = account_number else {
+        return Ok(Account::synthetic(anchor_number, origin));
```
sea-snake, line 95: There's no need for this, the storage layer (`storage.read_account`) already returns synthetic accounts (abstracts away the default/numbered account logic).

src/internet_identity/src/attributes.rs
```diff
@@ -0,0 +1,672 @@
...
+        CertifiedAttributes {
+            certified_attributes: openid_certified_attributes,
+            expires_at_timestamp_ns,
+        }
+    }
+
+    /// Processes `requested_attributes` for all attribute scopes, prepares signatures for
+    /// the (attribute_key, attribute_value) pairs that can be fulfilled for this (anchor, account).
+    ///
+    /// Returns the list of attribute keys certified with expiry `now_timestamp_ns + OPENID_SESSION_DURATION_NS`.
```
sea-snake, line 87: Don't re-use `OPENID_SESSION_DURATION_NS` here, it has a different purpose (lifetime of JWT delegation) than these attributes will have. Even if we want the same value, giving this it's own constant separates the concerns.

src/internet_identity/src/attributes.rs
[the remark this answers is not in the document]
sea-snake: An anchor can only have a single `iss` linked once currently, the storage layer allows for multiple but the implementation has been restricted to one-to-one relationship. So this implementation is correct, though I would add a code comment to mention this explicitly.

src/internet_identity/src/attributes.rs
```diff
@@ -0,0 +1,672 @@
...
+        &self,
+        account: &Account,
+        attributes: &Vec<Attribute>,
+        now_timestamp_ns: Timestamp,
+    ) {
+        let expiration_timestamp_ns = now_timestamp_ns.saturating_add(OPENID_SESSION_DURATION_NS);
+
+        let seed = account.calculate_seed();
+
+        state::signature_map_mut(|sigs| {
```
sea-snake, line 195: Reading from state here is a bit unexpected, maybe pass in a mutable sigs as argument instead?

src/internet_identity/src/attributes.rs
```diff
@@ -0,0 +1,672 @@
...
+                add_attribute_signature(sigs, &seed, &message);
+            }
+        });
+    }
+}
+
+// TODO: This should eventually be moved to a library function.
+fn attribute_signature_msg(attribute: &Attribute, expiration_timestamp_ns: u64) -> Vec<u8> {
+    let m: Vec<(String, Value)> = vec![
+        ("expiration".into(), Value::Number(expiration_timestamp_ns)),
```
sea-snake, line 207: Nit: If the key here is `expiration` shouldn't we change `expiration_timestamp_ns` to match?

src/internet_identity/src/main.rs
```diff
@@ -1210,8 +1214,92 @@ mod openid_api {
...
+        let (anchor, _) =
+            check_authorization(identity_number).map_err(|AuthorizationError { principal }| {
+                PrepareAttributeError::AuthorizationError(principal)
+            })?;
+
+        let account = get_account_for_origin(anchor.anchor_number(), origin, account_number)
+            .map_err(PrepareAttributeError::GetAccountError)?;
+
+        // This is the only async operation, so we do it first, call operations that depend on
+        // the time. TODO: refactor to avoid asynchronicity here.
```
sea-snake, line 1250: Nit: For now moving this to the top of the function would make sure authorization happens together with attribute preparation.

src/internet_identity/src/main.rs
```diff
@@ -1210,8 +1214,92 @@ mod openid_api {
...
+            // Arguments for computing the seed
+            identity_number,
+            origin,
+            account_number,
+
+            // Which attributes to prepare
+            attributes,
+
+            // When were the attribute certificates issued
+            issued_at_timestamp_ns,
```
sea-snake, line 1283: Didn't we only have expiration? Wheres does issued_at_timestamp_ns come from?

Edit: Ah I see, we're returning the issued at in the preparation response, probably makes more sense to return the expiration there instead.

src/internet_identity_interface/src/internet_identity/types/attributes.rs
```diff
@@ -0,0 +1,964 @@
...
+                (
+                    "openid",
+                    "openid:google.com",
+                    Ok(AttributeScope::OpenId {
+                        issuer: "google.com".to_string(),
+                    }),
+                ),
+                (
+                    "openid complex issuer",
+                    "openid:accounts.google.com",
```
sea-snake, line 397: Do we want to test here with the protocol included (`https://`)?

src/internet_identity/src/main.rs
aterga: This way, we can use the exact same code to prepare and fetch the signatures (i.e., issues_at is just the ultimate input arg)

src/internet_identity/src/attributes.rs
aterga: I like `expiration_timestamp_ns` better, but I decided to keep it `expiration` here to match what we have for delegations. Let's discuss / agree on the format

src/internet_identity/src/attributes.rs
aterga: There's a separate constant, will adjust the comment.

src/internet_identity/tests/integration/attributes.rs
```diff
@@ -0,0 +1,128 @@
+//! Tests related to prepare_delegation, get_delegation and get_principal II canister calls.
```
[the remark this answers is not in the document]
aterga: Will address in a follow-up

src/internet_identity/src/main.rs
```diff
@@ -1210,8 +1214,92 @@ mod openid_api {
...
+            .map_err(GetAttributesError::GetAccountError)?;
+
+        let certified_attributes =
+            anchor.get_attributes(attributes, account, issued_at_timestamp_ns);
+
+        Ok(certified_attributes)
+    }
+}
+
 /// API for the attribute sharing mvp
```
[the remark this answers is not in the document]
aterga: Will address in a follow-up

src/internet_identity/src/attributes.rs
aterga: Will address in a follow-up

src/internet_identity/src/attributes.rs
aterga: Why is this an issue?

src/internet_identity/src/main.rs
aterga: # prepare
msg = f(attrs, g(issued_at_timestamp_ns))

response = msg

# get(msg = response)
msg = f(attrs, g(issued_at_timestamp_ns))

ultimate response still has expires_at_timestamp_ns

Agreed to keep if for now in a call.

src/internet_identity/src/attributes.rs
aterga: Agreed to keep this for now for consistency.

src/internet_identity/src/main.rs
aterga: Will address this later.

src/internet_identity/src/attributes.rs
aterga: So this is about state access that isn't seen from the function signature. 

We should address this systematically after the legacy product is removed form the code.

src/internet_identity/src/account_management.rs
aterga: We agreed to keep this for now.

src/internet_identity/src/attributes.rs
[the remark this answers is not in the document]
aterga: Will add a comment in a follow-up.

src/internet_identity_interface/src/internet_identity/types/attributes.rs
aterga: Will do this in a follow-up.

PR #3601, opened 2026-01-30
Update upgrade modal footer to contain two links.

sea-snake:
Update upgrade modal footer to contain two links.

# Changes

- Rename `Help` link to `FAQ & Support`.
- Replace `Forgot your identity number?` question with `Lookup identity` link.
- Pass parent window origin into iframe so it can be used for the lookup link.
- Use either the above passed in origin, the current origin (if not id.ai) or fallback to `identity.ic0.app`.

# Tests

Manually verified on beta the links work as expected.

### DIFF
--- a/src/frontend/src/lib/assets/aaguid/security_keys.json
+++ b/src/frontend/src/lib/assets/aaguid/security_keys.json
@@ -124,6 +124,7 @@
   "59f85fe7-faa5-4c92-9f52-697b9d4d5473": "RSA Authenticator 4 for Android",
   "d41f5a69-b817-4144-a13c-9ebd6d9254d6": "ATKey.Card CTAP2.0",
   "2c0df832-92de-4be1-8412-88a8f074df4a": "Feitian FIDO Smart Card",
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/wizards/migration/views/EnterIdentityNumber.svelte
```diff
@@ -17,6 +17,14 @@
     ) => Promise<void | "cancelled" | "wrongDomain">;
   }
 
+  const primaryOrigin = getPrimaryOrigin();
+  const selfServiceOrigin =
+    parentIFrameOrigin !== undefined
+      ? parentIFrameOrigin
+      : window.location.origin !== primaryOrigin
```
aterga, line 24: I think this line should be replaced with


```suggestion
      : window.location.origin;
```

aterga: On second thoughts, this is okay

PR #3615, opened 2026-02-09
Implement certified attributes request handler.

sea-snake:
Implement certified attributes request handler.

# Changes

- Update generated bindings.
- Implement attribute request handler with implicit consent for openid attributes based on issuer.
- Add optional `certified_attributes` boolean config to dapps to enable per-dapp early opt-in to the new attributes feature.

# Tests

Added e2e tests that check whether expected attributes from implicit consent are returned while others are not returned.

### DIFF
--- a/src/frontend/src/lib/legacy/flows/dappsExplorer/dapps.json
+++ b/src/frontend/src/lib/legacy/flows/dappsExplorer/dapps.json
@@ -637,7 +637,11 @@
     "name": "Caffeine",
     "oneLiner": "Self-writing apps platform: the first complete tech stack designed for AI.",
     "website": "https://caffeine.ai",
-    "authOrigins": ["https://caffeine.ai", "https://app.caffeine.ai", "https://dev.caffeine.ai"],
+    "authOrigins": [
+      "https://caffeine.ai",
+      "https://app.caffeine.ai",
+      "https://dev.caffeine.ai"
+    ],
     "logo": "caffeine_logo.svg",
     "logoDark": "caffeine_logo_dark.svg",
     "certified_attributes": true


src/frontend/src/routes/(new-styling)/(resuming-channel)/resume-openid-authorize/+page.svelte
```diff
@@ -20,6 +28,80 @@
...
+          jsonrpc: "2.0",
+          id: request.id,
+          error: {
+            code: INVALID_PARAMS_ERROR_CODE,
+            message: z.prettifyError(paramsResult.error),
+          },
+        });
+        return;
+      }
+      const implicitConsentAttributeKeys = paramsResult.data.attributes.filter(
```
aterga, line 49: Please add a comment above this line to explain what is going on here. E.g.:

> Implicit consent applies to the special case when `scope === "openid:${issuer}"` where `issuer` is the OpenID issuer ID (e.g., Google) that matches the authentication method user by the user in this flow. In this case, consent stems from the fact that the user chose this OpenID issuer directly on the app website, and then consented to share the requested data with id.ai (so it is not confusing that this data is being shared back with the app).

src/frontend/tests/e2e-playwright/authorize/index.spec.ts
```diff
@@ -223,45 +223,6 @@ test("Authorize with ICRC-29", async ({ page }) => {
   );
 });
 
-test("Authorize with ICRC-29 (directly through OpenID)", async ({ page }) => {
```
aterga, line 226: Why do we no longer need this test?

src/frontend/src/lib/components/utils/AuthorizationChannel.svelte
```diff
@@ -15,8 +16,6 @@
   import { type Snippet } from "svelte";
   import { goto } from "$app/navigation";
 
-  const INVALID_PARAMS_ERROR_CODE = -32602;
```
aterga, line 18: What happened here?

demos/test-app/src/auth.ts
```diff
@@ -41,20 +48,69 @@ export const authWithII = async ({
...
+                attributes: requestAttributes,
+              },
+            })
+            .then((response) => {
+              if (
+                !("result" in response) ||
+                typeof response.result !== "object" ||
+                response.result === null ||
+                !("attributes" in response.result)
+              ) {
```
aterga, line 79: Nit. It is surprising to me that in TypeScript we need to do these things to assert that an instance is of a particular shape... Couldn't we assert that it's an instance of some interface, for example?

aterga: At the very least, could we hide these sanity checks in some helper functions?

demos/test-app/src/auth.ts
```diff
@@ -41,20 +48,69 @@ export const authWithII = async ({
...
+      delegationPromise,
+      attributesPromise,
+    ]);
     return {
-      identity: DelegationIdentity.fromDelegation(sessionIdentity, delegation),
+      identity: DelegationIdentity.fromDelegation(
+        sessionIdentity,
+        // We need to cast the delegation from signer-js to avoid a TS issue because one type is imported from cjs and another esm:
+        // Types of property 'delegations' are incompatible.
+        delegation as unknown as DelegationChain,
```
aterga, line 110: Nit. This could probably be done more elegantly in a helper function.

src/frontend/src/lib/utils/transport/utils.ts
```diff
@@ -3,6 +3,8 @@ import { Delegation, DelegationChain } from "@icp-sdk/core/identity";
 import { type Signature } from "@icp-sdk/core/agent";
 import { Principal } from "@icp-sdk/core/principal";
 
+export const INVALID_PARAMS_ERROR_CODE = -32602;
```
aterga, line 7: Where does this value come from?

src/frontend/src/lib/utils/transport/utils.ts
```diff
@@ -352,3 +360,7 @@ export const AuthResponseCodec = z.codec(
 );
 
 export type AuthResponse = z.output<typeof AuthResponseCodec>;
+
+export const AttributesParamsSchema = z.object({
```
aterga, line 368: A doc string wouldn't hurt here (I'm not 100% sure why we need this)

src/frontend/tests/e2e-playwright/fixtures/openid.ts
```diff
@@ -0,0 +1,49 @@
+import { test as base, expect, type Page } from "@playwright/test";
+
+export type OpenIdClaims = Record<string, unknown>;
+export type OpenIdUser = {
+  id: string;
+  claims: OpenIdClaims;
+  signIn: (page?: Page) => Promise<void>;
+};
+
+export const test = base.extend<{
```
aterga, line 10: Nit. A doc string could help me better understand the intended tests where this fixture is applicable, since it's making some simplifications.

src/frontend/src/lib/utils/transport/utils.ts
sea-snake: Added comment with reference to json-rpc spec

src/frontend/src/lib/utils/transport/utils.ts
sea-snake: This is a schema that parses the attributes request params to be a given shape. I've added a comment.

src/frontend/tests/e2e-playwright/authorize/index.spec.ts
sea-snake: It was moved to `openid.spec.ts`.

src/frontend/src/lib/components/utils/AuthorizationChannel.svelte
sea-snake: Variable was moved to `utils.ts` so it could be re-used in other files.

demos/test-app/src/auth.ts
sea-snake: In the II code itself we're using the Zod library to validate schemas, but there's no Zod or related library in the test-app and adding it (or any large changes to the test app) is out of the scope of this PR.

demos/test-app/src/auth.ts
sea-snake: More elegantly would be fixing the dependencies in the test app to avoid this type conflict to begin with, but that's outside the scope of this PR. To clarify, this comment was already there a few lines up, I've only moved it down.

src/frontend/tests/e2e-playwright/fixtures/openid.ts
aterga: Not urgent enough to block the PR

src/frontend/src/routes/(new-styling)/(resuming-channel)/resume-openid-authorize/+page.svelte
aterga: Not urgent enough to block the PR

PR #3620, opened 2026-02-16
Fix ICRC-29 heartbeat in Safari

sea-snake:
Fix ICRC-29 heartbeat in Safari and remove signer-js deps from II.

# Changes

- Move `HeartbeatServer` implementation from signer-js into II.
- Add `"wake-up-client"` message workaround for Safari to it's implementation.
- Remove other usages of signer-js so the dependency can be removed from II.
- Update test-app to latest signer-js version.

# Tests

Manually verified that direct OpenID flow now works as expected in Safari.

### DIFF
--- a/demos/test-app/package.json
+++ b/demos/test-app/package.json
@@ -21,7 +21,7 @@
   "devDependencies": {
     "@dfinity/internet-identity-vc-api": "*",
     "@dfinity/internet-identity-vite-plugins": "*",
-    "@icp-sdk/core": "^5.0.0",
+    "@icp-sdk/core": "^4.2.3",
     "@types/react": "^18.2.38",
     "@types/react-dom": "^18.2.17",
     "@vitejs/plugin-react": "^4.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -105,7 +105,7 @@
       "devDependencies": {
         "@dfinity/internet-identity-vc-api": "*",
         "@dfinity/internet-identity-vite-plugins": "*",
-        "@icp-sdk/core": "^5.0.0",
+        "@icp-sdk/core": "^4.2.3",
         "@types/react": "^18.2.38",
         "@types/react-dom": "^18.2.17",
         "@vitejs/plugin-react": "^4.2.0",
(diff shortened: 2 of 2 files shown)


demos/test-app/src/auth.ts
```diff
@@ -6,7 +6,7 @@ import {
   SignedDelegation,
 } from "@icp-sdk/core/identity";
 import { Principal } from "@icp-sdk/core/principal";
-import { fromBase64, Signer } from "@slide-computer/signer";
+import { Signer } from "@slide-computer/signer";
```
aterga, line 9: The PR says "remove signer-js deps", yet we still see them here.

sea-snake: It's removed from II, the test app (this file) still relies upon it. I've updated the PR description to clarify.

demos/test-app/src/auth.ts
```diff
@@ -85,8 +85,10 @@ export const authWithII = async ({
                     [
                       key,
                       {
-                        value: new Uint8Array(fromBase64(value)),
-                        signature: new Uint8Array(fromBase64(signature)),
+                        // @ts-ignore Not known in TS types yet but supported in all browsers
+                        value: Uint8Array.fromBase64(value),
+                        // @ts-ignore Not known in TS types yet but supported in all browsers
+                        signature: Uint8Array.fromBase64(signature),
                         expiration: BigInt(expiration),
```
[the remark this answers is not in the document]
sea-snake: As the comment indicates, it _is_ available in all browsers. But the type isn't available yet in the current Typescript version. Since this is an internal test/debugging app, this is fine.

src/frontend/src/lib/utils/transport/heartBeat.ts
```diff
@@ -0,0 +1,159 @@
+import { JsonRequestSchema } from "./utils";
```
aterga, line 1: IIUC, this code comes from the signer-js library. 

Could you link to the file where this code is taken from, please?

sea-snake: This is not from the signer-js lib but refers to a local `utils.ts` file in the same folder, implemented in a prior PR (any imports starting with either `$lib` or a path (`/`, `./` etc) are local).

sea-snake: (Additionally, the updated `package.json` in this PR indicates the signer-js dependency was removed).

src/frontend/src/lib/utils/transport/heartBeat.ts
```diff
@@ -0,0 +1,159 @@
...
+    //
+    // This resumes the client page from sleep in the case
+    // of browsers like Safari that unload background pages
+    // after a certain idle time and requiring an event to
+    // wake them up again (e.g. by sending a message to them).
+    if (this.#options.allowedOrigin !== null) {
+      this.#sendStatusResponse(
+        "wake-up-client",
+        this.#options.allowedOrigin,
+        window.opener,
```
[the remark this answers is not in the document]
sea-snake: Good idea, I've added the check.

src/frontend/src/lib/utils/transport/heartBeat.ts
aterga: Oh, so the code in this file is completely new? 

I was just surprised since the PR title says "fix ICRC-29 heartbeat in Safari" as opposed to, e.g., "implement ICRC-29 heartbeat server"

sea-snake: Ah yes that's what I intended to communicate with `Move HeartbeatServer implementation from signer-js into II.` in the PR description 😅   

Previously the contents of `heartBeat.ts` were imported from signer-js. Since this is the sole real usage of signer-js, it made more sense to just move it into II and remove the dependency.

aterga: What you did makes sense. 

But then my original question still stands: 
> Could you link to the file where this code is taken from, please?

Maybe you could provide the exact revision of signer-js from which this code comes from, which would allow me to review this code quicker than if I had to assume it's totally new.

sea-snake: Ah here's the `HeartBeatServer` file in signer-js: https://github.com/slide-computer/signer-js/blob/7b45e5fb904b3f9a4bf8291e5083062df53681d0/packages/signer-web/src/icrc29/heartbeat/server.ts

src/frontend/src/lib/utils/transport/heartBeat.ts
```diff
@@ -0,0 +1,162 @@
...
+  readonly #options: Required<HeartbeatServerOptions>;
+
+  constructor(options: HeartbeatServerOptions) {
+    this.#options = {
+      status: "ready",
+      establishTimeout: 10000,
+      disconnectTimeout: 2000,
+      window: globalThis.window,
+      ...options,
+      allowedOrigin: options.allowedOrigin ?? null,
```
aterga, line 52: Change No 1

src/frontend/src/lib/utils/transport/heartBeat.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  changeStatus(status: "pending" | "ready"): void {
+    this.#options.status = status;
+  }
+
+  #establish(): void {
+    // Establish communication channel if a request is received
+    const listener = this.#receiveStatusRequest((request) => {
+      if (request.source === null || request.data.id === undefined) {
```
aterga, line 65: Change No 2

src/frontend/src/lib/utils/transport/heartBeat.ts
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    // Send initial status response to kickstart the process
+    // in case the client is already waiting for responses.
+    //
+    // This resumes the client page from sleep in the case
+    // of browsers like Safari that unload background pages
+    // after a certain idle time and requiring an event to
+    // wake them up again (e.g. by sending a message to them).
+    if (
```
aterga, line 83: Change No 3

src/frontend/src/lib/utils/transport/heartBeat.ts
```diff
@@ -0,0 +1,162 @@
...
+          request.origin,
+          request.source,
+        );
+      }
+    });
+  }
+
+  #receiveStatusRequest(handler: (event: MessageEvent) => void): () => void {
+    const listener = (event: MessageEvent) => {
+      const result = JsonRequestSchema.safeParse(event.data);
```
aterga, line 137: Change No 4

src/frontend/src/lib/utils/transport/heartBeat.ts
```diff
@@ -0,0 +1,162 @@
...
+      if (
+        !result.success ||
+        result.data.method !== "icrc29_status" ||
+        (this.#options.allowedOrigin !== null &&
+          event.origin !== this.#options.allowedOrigin)
+      ) {
+        return;
+      }
+      handler(event);
```
aterga, line 146: Shouldn't it now be

```
handler(result);
```

?

sea-snake: That would only be the event data instead of the whole event, we need the whole event so we can reply to it's source.

To clarify `event.data` and `result.data` should be exact equals, Zod validation doesn't transform the data here.

PR #3621, opened 2026-02-16
Fix dark dapp logo in direct OpenID flow

sea-snake:
Fix dark dapp logo in direct OpenID flow

# Changes

- Apply same fix as previously seen in `AuthorizeHeader` component to direct OpenID page.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/(resuming-channel)/resume-openid-authorize/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/(resuming-channel)/resume-openid-authorize/+page.svelte
@@ -164,10 +164,9 @@
 
 <div class="flex min-h-[100dvh] flex-col items-center justify-center px-8">
   {#if dapp?.logoSrc !== undefined}
-    {@const name = dapp?.name ?? $authorizationContextStore.requestOrigin}
     <img
       src={dapp.logoSrc}
-      alt={$t`${name} logo`}
+      alt={$t`${dapp.name} logo`}
       class={[
         "mb-10 h-16 max-w-50 object-contain",
         dapp.logoDarkSrc !== undefined && "dark:hidden",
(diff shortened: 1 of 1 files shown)


src/frontend/src/routes/(new-styling)/(resuming-channel)/resume-openid-authorize/+page.svelte
```diff
@@ -164,12 +164,22 @@
...
     <img
-      src={dapp?.logoSrc}
-      alt={$t`${name} logo`}
-      class="mb-10 h-16 max-w-50 object-contain"
+      src={dapp.logoSrc}
+      alt={$t`${dapp.name} logo`}
+      class={[
+        "mb-10 h-16 max-w-50 object-contain",
+        dapp.logoDarkSrc !== undefined && "dark:hidden",
+      ]}
```
aterga, line 173: Could you clarify the semantics of this syntax, please? 

```
class={[
    "concreteClassName",
    condition && "anotherConcreetClassName"
]}
```

If this is adding `anotherConcreetClassName` based on the value of `condition`, then why not branch on the same level as you're branching right below this code:

```
{#if dapp.logoDarkSrc !== undefined}
    ...
{/if}
```

sea-snake: Because that branch points to `dapp.logoDarkSrc` instead of `dapp.logoSrc`.

The general logic here:

- if there's a light mode logo variant:
   render light mode logo
   - if there's a dark mode logo variant:
      add CSS class that hides light mode logo when browser/OS is in dark mode
      render dark mode logo variant
      - hides unless browser/OS is in dark mode

Basically, only a single logo image is rendered unless there's an explicit dark mode variant of a logo, then it renders both the light and dark mode images with css classes to hide one or the other based on the browser/OS being in dark mode.

PR #3629, opened 2026-02-23
chore: Add e2e test for split canister setup

sea-snake:
Add e2e test for split canister setup.

# Changes

- Increase e2e test sharding from 3 to 6 (reduces waiting times on PR)
- Add single/split canister property to e2e matrix to run tests on both the single and split canister setup
- Add frontend canister test flavor build step so it can be deployed for e2e tests where canister setup is split
- Update root layout to correctly show page after is has been loaded (there was a mismatched attribute name)
- Update http certification to only certify paths coming from `get_static_assets()`, avoiding unintentional overriding of html assets.
- Remove usage of `new_flow_origins` to check if passkey is legacy in dashboard.
- Remove usage of `new_flow_origins` to sort passkeys when creating web authn flows.
- Remove usage of `new_flow_origins` in new discoverable passkey UX feature flag by cleaning up this unused code.

### DIFF
--- a/src/frontend/src/lib/utils/findWebAuthnFlows.ts
+++ b/src/frontend/src/lib/utils/findWebAuthnFlows.ts
@@ -1,7 +1,7 @@
 import { II_LEGACY_ORIGIN } from "$lib/legacy/constants";
 import { isNullish, nonNullish } from "@dfinity/utils";
 import { CredentialData } from "./credential-devices";
-import { canisterConfig, getPrimaryOrigin } from "$lib/globals";
+import { getPrimaryOrigin } from "$lib/globals";
 import { isSameOrigin } from "./urlUtils";
 
 export type WebAuthnFlow = {


src/frontend/src/lib/flows/migrationFlow.svelte.ts
```diff
@@ -55,8 +54,11 @@ export class MigrationFlow {
   ): Promise<void> => {
     this.identityNumber = identityNumber;
     this.#devices = await this.#lookupAuthenticators(identityNumber);
-    const alreadyMigrated = this.#devices.some(isNewOriginDevice);
-    if (alreadyMigrated) {
+    const primaryOrigin = getPrimaryOrigin();
+    if (
+      primaryOrigin !== undefined &&
+      this.#devices.some((device) => device.origin[0] === primaryOrigin)
```
aterga, line 60: Is `device.origin` guaranteed to be non-empty?

sea-snake: Yes, for non legacy passkeys it's non empty. In case it's empty (legacy passkey), this condition would be false as intended.

PR #3636, opened 2026-02-26
Production frontend build

sea-snake:
Add production frontend canister build to CI pipeline.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -88,7 +88,7 @@ ARG II_SPLIT_FRONTEND=
 RUN touch src/*/src/lib.rs
 RUN npm ci
 
-RUN if [ -n "$II_SPLIT_FRONTEND" ]; then \
+RUN if [ "$II_SPLIT_FRONTEND" = "1" ]; then \
     ./scripts/build --frontend && \
     mv /internet_identity_frontend.wasm.gz /internet_identity.wasm.gz; \
     else \


Dockerfile
```diff
@@ -83,11 +83,17 @@ ARG II_FETCH_ROOT_KEY=
...
 ARG II_DEV_CSP=
+ARG II_SPLIT_FRONTEND=
 
 RUN touch src/*/src/lib.rs
 RUN npm ci
 
-RUN ./scripts/build
+RUN if [ -n "$II_SPLIT_FRONTEND" ]; then \
+    ./scripts/build --frontend && \
+    mv /internet_identity_frontend.wasm.gz /internet_identity.wasm.gz; \
```
aterga, line 93: Why do we rename `internet_identity_frontend` as `internet_identity`?

sea-snake: The build_script outputs a different filename when `--frontend` is passed. I wanted to keep the changes to a minimum, so renaming it here avoids having to make larger changes to the docker build pipeline.

aterga: This can be very confusing and error prone. Where is this file going to be consumed?

sea-snake: It's consumed through [here](https://github.com/dfinity/internet-identity/pull/3636/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R109) and then [here](https://github.com/dfinity/internet-identity/pull/3636/changes#diff-94c653b63027516068a92d1f6fdaf2e9b86999c0a679ac8ba2f7835f907fe672R111).

PR #3641, opened 2026-02-28
chore: Fix candid encoding for II upgrade args

aterga:
# Motivation

This PR ensures that the Candid encoding of II install arguments takes API type definitions into account. Candid encoding should use the correct type definitions, not automatically inferred types, otherwise distinct variant cases may get unexpectedly collapsed into one.

# Changes

Provide options `-d ./src/internet_identity/internet_identity.did -t '(opt InternetIdentityInit)'` to all `candid` commands in the upgrade script.

# Tests

Tested manually.

### DIFF
--- a/scripts/make-upgrade-proposal
+++ b/scripts/make-upgrade-proposal
@@ -297,7 +297,7 @@ deploy_release_candidate() {
       --mode upgrade \
       --wasm ./internet_identity.wasm.gz \
       --argument-type raw \
-      --argument-file "$(cat ./dfx_arg.bin)"
+      --argument "$(cat ./dfx_arg.bin)"
 }
 
 test_release_candidate() {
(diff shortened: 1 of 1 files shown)


scripts/make-upgrade-proposal
```diff
@@ -290,7 +290,14 @@ request_verification_command() {
...
   echo "Deploying II release candidate with same arguments as prod"
-  dfx canister --network ic --wallet cvthj-wyaaa-aaaad-aaaaq-cai install fgte5-ciaaa-aaaad-aaatq-cai --mode=upgrade --wasm ./internet_identity.wasm.gz --argument-file ./args.txt
+  dfx canister \
+    --network ic \
+    --wallet cvthj-wyaaa-aaaad-aaaaq-cai \
+    install fgte5-ciaaa-aaaad-aaatq-cai \
+      --mode upgrade \
+      --wasm ./internet_identity.wasm.gz \
+      --argument-type raw \
+      --argument-file "$(cat ./dfx_arg.bin)"
```
sea-snake: Is this supposed to be `--argument-file`? Given that we pass the output of `cat` into it instead of a file path.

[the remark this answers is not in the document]
aterga: Good catch, but `--argument-file` doesn't work with raw args, so has to be changes to `--argument`

aterga: Good catch, it should be --argument

PR #3645, opened 2026-03-03
test: Refactor manage access e2e tests

sea-snake:
Refactor manage access e2e tests, based on a [prior PR](https://github.com/dfinity/internet-identity/pull/3561), this PR implements a fixture for the manage access methods page and rewrites and consolidates the tests accordingly.

# Changes

- Wrap all e2e tests in `routes` folder next to `fixtures`.
- Rename `dashboard` to `manage` to match route naming
- Implement `manageAccessPage` fixture based on prior PR.
- Update identity fixture to add method to replace dummy auth index of an identity.
- Rewrites and consolidates passkey e2e tests into a single `access` tests file.

### DIFF
--- a/src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
+++ b/src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
@@ -28,14 +28,14 @@ test.describe("Access methods", () => {
     await managePage.signOut();
     await manageAccessPage.goto();
     await signInWithIdentity(page, identities[0].identityNumber);
-    await managePage.assertVisible();
+    await manageAccessPage.assertVisible();
 
     // Verify we can now also sign in with the new passkey
     await managePage.signOut();
     replaceAuthForIdentity(identities[0].identityNumber, authIndex);
     await manageAccessPage.goto();
     await signInWithIdentity(page, identities[0].identityNumber);
-    await managePage.assertVisible();
+    await manageAccessPage.assertVisible();
   });
 
   test.describe("can rename a passkey", () => {


src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    await manageAccessPage.assertPasskeyCount(1);
+    for (let i = 0; i < 15; i++) {
+      await manageAccessPage.add((dialog) => dialog.passkey(dummyAuth()));
+    }
+    await manageAccessPage.assertPasskeyCount(16);
+    await manageAccessPage.add(async (dialog) => {
+      await dialog.assertPasskeyUnavailable();
+      await dialog.close();
+    });
+  });
```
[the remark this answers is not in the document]
sea-snake: This has already been adjusted.

src/frontend/tests/e2e-playwright/fixtures/manageAccessPage.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  findPasskey(name: string): PasskeyItem {
+    const item = this.#page
+      .getByRole("main")
+      .getByRole("listitem")
+      .filter({ hasText: "Passkey" });
+    return new PasskeyItem(item, name);
+  }
+
```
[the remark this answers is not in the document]
sea-snake: Already fixed too.

src/frontend/tests/e2e-playwright/fixtures/manageAccessPage.ts
```diff
@@ -0,0 +1,213 @@
...
+    await this.locator
+      .getByRole("textbox", { name: "Passkey name" })
+      .fill(value);
+  }
+
+  async submit(): Promise<void> {
+    await this.locator.getByRole("button", { name: "Save changes" }).click();
+    const value = await this.locator
+      .getByRole("textbox", { name: "Passkey name" })
+      .inputValue();
```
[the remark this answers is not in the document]
sea-snake: Good catch!

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    const authIndex = getRandomIndex();
+    const auth = dummyAuth(authIndex);
+    await manageAccessPage.assertPasskeyCount(1);
+    await manageAccessPage.add((dialog) => dialog.passkey(auth));
+    await manageAccessPage.assertPasskeyCount(2);
+
+    // Verify we can still sign in with the existing passkey
+    await managePage.signOut();
+    await manageAccessPage.goto();
+    await signInWithIdentity(page, identities[0].identityNumber);
```
aterga, line 30: What is the postcondition here? Just that the URL starts with /manage from `IdentityWizard.signIn`?

Is that sufficient?

sea-snake: Yes, ideally I'd like to avoid repeating common assertion across various tests. When a fixture is used within a test, the test shouldn't have to evaluate if the fixture is working correctly.

The `signInWithIdentity` fixture isn't specific to the manage page sign-in flow but also used for example during authorization, so adding additional assertions within that fixture seems incorrect. Instead I'll add an assertion method `assertVisible` to the `managePage` fixture that can be used to assert that the manage page has been loaded where explicitly needed (for example in this test).

aterga: > I'll add an assertion method assertVisible to the managePage fixture that can be used to assert that the manage page has been loaded where explicitly needed

Sounds like a good idea.

Additionally, I think a more general success check could be added, e.g., "assertIsAuthorized", which would check if an authenticated call can be made using the expected agent. Wdyt?

sea-snake: Added explicit post assertion `managePage.assertVisible()` here and to other relevant tests.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+        .findPasskey(DEFAULT_PASSKEY_NAME)
+        .rename(async (dialog) => {
+          await dialog.fill("Dell XPS");
+          await dialog.submit();
+        });
+      await manageAccessPage.assertPasskeyCount(1);
+      await manageAccessPage.assertPasskeyExists("Dell XPS");
+    });
+
+    test("unless it's an empty string", async ({ manageAccessPage }) => {
```
aterga, line 56: Should we test if the same name can be set to two distinct passkeys?

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      await manageAccessPage.add((dialog) => dialog.passkey(auth));
+      await manageAccessPage
+        .findPasskey(DEFAULT_PASSKEY_NAME)
+        .rename(async (dialog) => {
+          await dialog.fill("additional-passkey");
+          await dialog.submit();
+        });
+    });
+
+    test("which has been added previously", async ({ manageAccessPage }) => {
```
aterga: test name seems moot — obviously something that wasn't added cannot be removed

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      await page.waitForURL(II_URL); // Expect to be signed out
+      await manageAccessPage.goto(); // Go back to the manage page
+
+      // Sign in with the new passkey and assert it's the only passkey
+      replaceAuthForIdentity(
+        identities[0].identityNumber,
+        identities[0].dummyAuthIndex + BigInt(1),
+      );
+      await signInWithIdentity(page, identities[0].identityNumber);
+      await manageAccessPage.assertPasskeyCount(1);
```
aterga, line 128: We should assert `await manageAccessPage.assertPasskeyCount(1);` also in the precondition of this test, right?

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      await manageAccessPage.add((dialog) => dialog.passkey(dummyAuth()));
+    }
+    await manageAccessPage.assertPasskeyCount(16);
+    await manageAccessPage.add(async (dialog) => {
+      await dialog.assertPasskeyUnavailable();
+      await dialog.close();
+    });
+  });
+
+  test.describe("can be cancelled", () => {
```
aterga, line 154:
```suggestion
  test.describe("ech closable dialogue can be successfully closed", () => {
```

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: I think there are various other test cases to add besides the current test cases. This PR intended to refactor and consolidate the test cases without going for a broader scope of extending the current test coverage.

I'd like to revisit the exact test cases and their coverage later on in combination with the active device protection efforts given that these will affect most of these tests to begin with.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+  test("cannot have more than 16 passkeys", async ({ manageAccessPage }) => {
+    await manageAccessPage.assertPasskeyCount(1);
+    for (let i = 0; i < 15; i++) {
+      await manageAccessPage.add((dialog) => dialog.passkey(dummyAuth()));
+    }
+    await manageAccessPage.assertPasskeyCount(16);
+    await manageAccessPage.add(async (dialog) => {
+      await dialog.assertPasskeyUnavailable();
+      await dialog.close();
+    });
```
aterga, line 150:
```suggestion
    });
    await manageAccessPage.assertPasskeyCount(16);
```

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      await manageAccessPage.assertPasskeyCount(1);
+      await manageAccessPage.assertPasskeyExists("additional-passkey");
+    });
+  });
+
+  test("cannot remove a single passkey", async ({ manageAccessPage }) => {
+    await manageAccessPage.assertPasskeyCount(1);
+    await manageAccessPage
+      .findPasskey(DEFAULT_PASSKEY_NAME)
+      .assertUnremovable();
```
aterga, line 137: Nit. Would be better to express this as, e.g.:


```suggestion
      await manageAccessPage.assertPasskeyCount(1);
      await manageAccessPage.tryRemovePasskey(DEFAULT_PASSKEY_NAME);
      await manageAccessPage.assertPasskeyCount(1);
```

aterga: Basically, the code under test and the postcondition assertions should not be inside the same helper function, ideally.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: I intended to communicate the opposite of `which is currently in use` but ended up indeed writing something vague. I'll rename this to be more clear to `which is currently not in use`.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+    test("which is currently in use", async ({
+      page,
+      manageAccessPage,
+      identities,
+      signInWithIdentity,
+      replaceAuthForIdentity,
+    }) => {
+      // Remove currently in use passkey
+      await manageAccessPage
+        .findPasskey("in-use-passkey")
```
aterga, line 113: Who's defining which passkey is "in-use-passkey"?

aterga (changes requested):
Looks good so far

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: There are two passkeys (one in use and the other not in use). So indeed the precondition is missing, but it should be `2`.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: I understand the idea of wanting to be more explicit, but the actual purpose of these tests is not to verify if dialogs can be closed, instead we're verifying if a user action can be cancelled (e.g. you select rename but want to cancel/abort midway without any changes).

So renaming these tests to "dialog can be closed" would hide the actual underlying UX concept of cancelling a user flow while making sure things remain unchanged. That the act of cancelling is either closing a dialog, clicking a cancel button or something else, is irrelevant.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: See the `beforeEach` at line 71 where passkeys are renamed before every test to explicitly assign meaning to them.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: Good idea to add a post assertion here!

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: Added `await manageAccessPage.assertPasskeyCount(1);` post assertion.

I kept the `assertUnremovable` as-is since `findPasskey` and it's underlying methods are separate implementations on a larger grid view and inner menu view, making this a single method `tryRemovePasskey` would just obfuscate this fact in the fixture without any actual benefit.

To clarify `assertUnremovable` asserts that the remove button isn't presents in the passkey's dropdown menu. 

An alternative approach (as I assume you meant with `tryRemovePasskey`) would be to just run the code that tries to remove it (as if there was an option in the dropdown menu) and then assert it didn't happen. 

But this alternative approach could result in a false positive, where due to some other circumstances the removal didn't happen, but in reality it will happen in production.

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
aterga: But are the tests actually asserting that the state is unchanged after the dialogue is closed?

sea-snake: Yes there's an `afterEach` condition on all `can be cancelled` tests that:
- Verifies the number of passkeys hasn't changed (1)
- Verifies we can still sign in with the existing passkey

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
aterga: The `tryRemovePasskey` function should ideally return an error that can be used to distinguish the reason it failed. 

> To clarify assertUnremovable asserts that the remove button isn't presents in the passkey's dropdown menu.

This sees fine for now. It just makes it harder to understand the test without jumping to the fixture definitions, since it's not immediately clear what "unremovable" means and how you ensure that.

NCR

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
aterga: What I don't get is why do we rely on the name to determine if it's in use... How would this work in production?

sea-snake: In production we have this barely visible green dot to indicate it's in-use, something I can't reliably match on within these tests without having to use id/class selectors.

In the updated protected access methods design, we have more explicit "Active" labels, which should enable us to match on those instead of relying on passkey names.

aterga: Okay, so you're saying this will soon change. SGTM

src/frontend/tests/e2e-playwright/routes/manage/access.spec.ts
sea-snake: > The tryRemovePasskey function should ideally return an error that can be used to distinguish the reason it failed.

Expecting an error in playwright is kind of an anti pattern since it relies on waiting for a click attempt on the remove button to timeout, slowing down the test case to the timeout duration and making it flaky if it doesn't show the button for a given duration (and thus times out) for another reasons.

Alternatively the `remove` method could use playwright visible/hidden checks and throw an error if it's hidden, but doing such a check on the DOM without any timeout would also be an anti pattern and make the tests flaky.

The general idea in non flaky playwright tests is that you write code towards the intended path (button visible, missing, disabled etc) and in case this path doesn't pass, your test will fail (after a timeout). Inverting this concept, waiting on timeouts for a test to pass, slows down CI significantly and is thus one of the main anti patterns.

PR #3650, opened 2026-03-05
chore: Support `--end <front|back>` in PR-to-staging deployment script

aterga:
# Motivation

This is needed due to the recent II canister split. For now, only Staging-A comes with a frontend staging environment, but we could easily add more as needed.

# Changes

* Generalize `scripts/deploy-pr-to-beta` to support the new option `--end` (mandatory)

# Tests

* Tested manually using `./scripts/deploy-pr-to-beta --end front --staging-a 3647`

### DIFF
--- a/scripts/deploy-pr-to-beta
+++ b/scripts/deploy-pr-to-beta
@@ -1,25 +1,31 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Usage: ./deploy-pr-to-beta [--staging-a|-sa | --staging-b|-sb | --staging-c|-sc | --staging <CANISTER_ID>] <PR_NUMBER>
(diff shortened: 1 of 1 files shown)


scripts/deploy-pr-to-beta
```diff
@@ -83,25 +119,153 @@ if [ -z "$STAGING_CANISTER_ID" ]; then
...
+        exit 1
+    fi
+
+    echo ""
+    echo "Current frontend config:"
+    echo "$raw_config"
+    echo ""
+    echo "Configure install arguments (press Enter to keep each current value):"
+
+    # Parse individual fields from the Candid text output.
```
sea-snake, line 185: Nit: Can't we parse and merge this with didc instead?

PR #3658, opened 2026-03-09
Fix CSP, allow related_origins to embed one another.

sea-snake:
Fix CSP, allow related_origins to embed one another.

# Changes

- Pass `related_origins` into `get_content_security_policy` to set `frame-src` and `frame-ancestors`.

# Tests

To be tested on beta.

### DIFF
--- a/src/internet_identity_frontend/src/main.rs
+++ b/src/internet_identity_frontend/src/main.rs
@@ -146,7 +146,7 @@ fn get_asset_headers(
     related_origins: Option<&Vec<String>>,
     additional_headers: Vec<HeaderField>,
 ) -> Vec<HeaderField> {
-    let cross_origin_credentials_allowlist = if let Some(related_origins) = related_origins {
+    let credentials_allowlist = if let Some(related_origins) = related_origins {
         if related_origins.is_empty() {
             "self".to_string()
         } else {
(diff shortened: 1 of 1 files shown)


src/internet_identity_frontend/src/main.rs
```diff
@@ -137,8 +143,23 @@ fn certify_all_assets(args: InternetIdentityFrontendArgs) {
...
 fn get_asset_headers(
     integrity_hashes: Vec<String>,
+    related_origins: Option<&Vec<String>>,
     additional_headers: Vec<HeaderField>,
 ) -> Vec<HeaderField> {
+    let credentials_allowlist = if let Some(related_origins) = related_origins {
+        if related_origins.is_empty() {
+            "self".to_string()
+        } else {
+            let quoted: Vec<String> = related_origins
```
aterga, line 153: So here you don't include `self` but you do include it on https://github.com/dfinity/internet-identity/pull/3658/changes#diff-b0f359add31706a73b515f10b8356a12b4afebd0cd6f2293481d3805ad270f5fR307

could you clarify why this is the case?

sea-snake: I'm including self, it's on line 157. Keep in mind the format is different here (quotes) from the embedding allowlist.

PR #3668, opened 2026-03-12
chore: improve German translations

marc0olo:
Improves German translations.


src/frontend/src/lib/locales/de.po
```diff
@@ -315,7 +315,7 @@ msgid "Full control"
 msgstr "Volle Kontrolle"
 
 msgid "Full Ownership"
-msgstr "Volle Eigentümerschaft"
+msgstr "Volle Datenhoheit"
```
marc0olo, line 318: this makes most sense in that context IMO, even though it is not a 100% accurate translation.

PR #3760, opened 2026-04-05
[DO NOT MERGE] feat: email reception with Postbox UI for identity owners

aterga:
## Summary
- Identity owners can now receive emails at `<anchor_number>@id.ai` and view them in a new **Postbox** tab in the management dashboard.
- An SMTP gateway accepts incoming emails, validates the recipient is a valid anchor number, and stores up to 10 emails per identity.

### Backend
- `smtp_request` (update) and `smtp_request_validate` (query) canister endpoints implementing the SMTP Gateway Protocol.
- `get_postbox` (query) endpoint returning stored emails for an anchor number (newest first).
- Persistent email storage in a new `smtp_postbox` stable BTreeMap (memory ID 23).
- Validation: recipient must be `<valid_u64>@id.ai`, body ≤ 5 KB, ≤ 10 headers.

### Frontend
- New **Postbox** nav tab (with mail icon), visible only when the user has emails.
- Two-column Postbox page: email list on the left, selected email content on the right.
- Postbox data loaded alongside `identity_info` in the authenticated layout.
- Regenerated IDL and TypeScript types from updated `.did` file.


src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,580 @@
...
+        // In tests, we don't do HTTP outcalls — return Unverified
+        let _ = &sig;
+        return DkimVerificationStatus::Unverified {
+            reason: "DKIM key fetch not available in tests".into(),
+        };
+    };
+
+    let signing_input = build_signing_input(headers, &dkim_header.value, &sig);
+
+    match verify_rsa_sha256(&signing_input, &sig.signature, &public_key) {
```
aterga, line 625: PTAL @sea-snake

src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,702 @@
...
+// --- Verification (only used in non-test builds via the async orchestrator) ---
+
+#[cfg(not(test))]
+fn verify_body_hash(raw_body: &[u8], sig: &DkimSignature) -> Result<(), String> {
+    let canonicalized = match sig.body_canon {
+        Canon::Simple => canonicalize_body_simple(raw_body),
+        Canon::Relaxed => canonicalize_body_relaxed(raw_body),
+    };
+
+    let body_to_hash = match sig.body_length {
```
sea-snake, line 200: If we intend to handle the body length in DKIM per spec, we should make sure to **only** retain that "trusted" part of the body in future processing.

src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,702 @@
...
+
+/// Remove the value of the b= tag from the DKIM-Signature header value,
+/// keeping the tag name and delimiters.
+fn strip_b_value(header_value: &str) -> String {
+    let mut result = String::new();
+    let mut remaining = header_value;
+
+    // Find "b=" that is not "bh=" — it follows a semicolon or starts the string
+    loop {
+        if let Some(pos) = remaining.find("b=") {
```
sea-snake, line 264: DKIM headers:
- Allow folding (line breaks + whitespace)
- Allow arbitrary whitespace around = and ;
- Are case-insensitive
- Can contain multiple b=-like substrings inside values

DKIM headers should not be parsed manually like this.

Instead:

Parse into structured tags first (b, bh, h, etc.)
Reconstruct the header canonically with b= empty

or:

Reuse a well-tested DKIM parsing library

src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,702 @@
...
+// --- DKIM DNS public key parsing ---
+
+/// Parse a DKIM DNS TXT record value (e.g., "v=DKIM1; k=rsa; p=MIGfMA0G...") into an RSA public key.
+pub fn parse_dkim_public_key(txt_record: &str) -> Result<RsaPublicKey, String> {
+    let mut p_value: Option<String> = None;
+
+    for part in txt_record.split(';') {
+        let part = part.trim();
+        if let Some((key, val)) = part.split_once('=') {
+            if key.trim() == "p" {
```
sea-snake, line 325: This is not fully DKIM-compliant parsing.

DKIM DNS records:

- Allow folding across multiple strings (DNS TXT chunks)
- Allow whitespace almost anywhere
- Are case-insensitive
- Can contain unexpected formatting

Example fix:
```
if key.trim().eq_ignore_ascii_case("p")
```

src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,702 @@
...
+        }
+    }
+
+    let p_base64 = p_value.ok_or("Missing p= tag in DKIM DNS record")?;
+    let der_bytes = BASE64
+        .decode(&p_base64)
+        .map_err(|e| format!("Invalid base64 in DKIM public key: {e}"))?;
+
+    use rsa::pkcs8::DecodePublicKey;
+    RsaPublicKey::from_public_key_der(&der_bytes)
```
sea-snake, line 337: DKIM allows small keys in theory, but in practice modern policy recommends:
- ≥ 1024 bits (minimum)
- Preferably 2048 bits

src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,702 @@
...
+            .iter()
+            .enumerate()
+            .rev()
+            .find(|(i, h)| !used_indices[*i] && h.name.eq_ignore_ascii_case(signed_name));
+
+        if let Some((idx, header)) = found {
+            used_indices[idx] = true;
+            let line = match sig.header_canon {
+                Canon::Relaxed => canonicalize_header_relaxed(&header.name, &header.value),
+                Canon::Simple => format!("{}: {}", header.name, header.value),
```
sea-snake, line 237: This violates RFC 6376, simple canonicalization requires exact original bytes, extra whitespace is added here after `:`.

```suggestion
                Canon::Simple => format!("{}:{}", header.name, header.value),
```

sea-snake: To summarize the issue with current headers:

> DKIM verification must operate on the **exact original header bytes as received on the wire**, but your implementation reconstructs headers from parsed `name` and `value` fields using formatting (and partial normalization). This loses critical details like original whitespace, folding (`CRLF` continuations), and exact spacing, which are part of the signed data even in “simple” canonicalization. As a result, your reconstructed headers can differ from what was actually signed, causing valid DKIM signatures to fail verification or behave inconsistently with real-world mail servers.

src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,702 @@
...
+    pub selector: String,
+    pub signed_headers: Vec<String>,
+    pub body_hash: Vec<u8>,
+    pub signature: Vec<u8>,
+    pub header_canon: Canon,
+    pub body_canon: Canon,
+    pub body_length: Option<usize>,
+}
+
+pub fn parse_dkim_signature(value: &str) -> Result<DkimSignature, String> {
```
sea-snake, line 35: We need some more checks:

> The implementation ignores several important DKIM signature and DNS policy tags beyond the cryptographic fields, such as `i=` (identity alignment), `t=` (signature timestamp), `x=` (expiration time), and DNS-side constraints like `k=` and `v=`. These fields are not just metadata—they define whether a signature is *valid and acceptable* under DKIM policy rules, not just whether it can be cryptographically verified. By ignoring them, your verifier may incorrectly accept replayed, expired, misaligned, or malformed signatures, meaning it performs only raw signature validation rather than full DKIM authentication semantics.

src/internet_identity/src/dkim.rs
```diff
@@ -0,0 +1,702 @@
...
+}
+
+#[cfg(not(test))]
+pub async fn verify_email_dkim(headers: &[SmtpHeader], raw_body: &[u8]) -> DkimVerificationStatus {
+    let mut checks: Vec<DkimCheck> = Vec::new();
+
+    // 1. DKIM-Signature header present
+    let dkim_header = match headers
+        .iter()
+        .find(|h| h.name.eq_ignore_ascii_case("dkim-signature"))
```
sea-snake, line 492: The implementation only verifies the first DKIM-Signature header it finds, even though an email can legitimately contain multiple DKIM signatures from different domains (e.g., original sender and mailing list forwarders). By not iterating over all signatures and accepting the email if any signature is valid, you lose DKIM’s intended robustness against forwarding and multi-hop email processing, which can lead to unnecessary verification failures and reduced interoperability with real-world mail systems.

sea-snake:
Another note, we also need to check this on top of DKIM:
> DMARC adds a policy and identity-check layer on top of DKIM by tying the visible From: domain (Y) to the authenticated signing domain (X). After DKIM verifies that a domain X signed the message, DMARC checks whether X is aligned with Y, meaning either they match directly or are considered related under DNS-based alignment rules (typically the same organizational domain or a permitted subdomain relationship defined by DMARC policy). Importantly, this alignment can allow a different domain X to pass for Y only if Y’s DNS explicitly permits it through its DMARC configuration and domain structure rules. If alignment succeeds and Y’s policy allows it, the message is accepted; otherwise, even a valid DKIM signature is treated as spoofing.

src/internet_identity_interface/src/internet_identity/types/smtp.rs
```diff
@@ -0,0 +1,231 @@
...
+    Ok(())
+}
+
+fn extract_subject(headers: &[SmtpHeader]) -> String {
+    headers
+        .iter()
+        .find(|h| h.name.eq_ignore_ascii_case("subject"))
+        .map(|h| {
+            let mut s = h.value.clone();
+            s.truncate(MAX_SUBJECT_BYTES);
```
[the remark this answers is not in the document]
aterga: Fixed —  now calls a new `truncate_at_char_boundary` helper that clamps to the previous UTF-8 char boundary instead of calling `String::truncate` on a byte offset. A multi-byte subject at exactly `MAX_SUBJECT_BYTES` can no longer trap. Covered by `multibyte_subject_truncates_on_char_boundary` in the new `smtp.rs` test module.

src/internet_identity_interface/src/internet_identity/types/smtp.rs
```diff
@@ -0,0 +1,231 @@
...
+        validate_message(message)?;
+
+        let body = String::from_utf8_lossy(&message.body).into_owned();
+
+        Ok(ValidatedSmtpRequest {
+            sender: format_address(&envelope.from),
+            recipient: format_address(&envelope.to),
+            subject: extract_subject(&message.headers),
+            body,
+        })
```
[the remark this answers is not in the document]
aterga: Fixed — after `from_utf8_lossy`, we now clamp the resulting string to `MAX_BODY_BYTES` on a char boundary via the same `truncate_at_char_boundary` helper used for the subject. So even a body of all-invalid UTF-8 bytes (each replaced by the 3-byte U+FFFD) cannot overflow the storage bound. Covered by the new `non_utf8_body_does_not_panic` test, which feeds `MAX_BODY_BYTES` copies of `0xFF` and asserts the validated string stays within bounds without trapping.

src/internet_identity_interface/src/internet_identity/types/smtp.rs
```diff
@@ -0,0 +1,231 @@
...
+    let user_lower = envelope.to.user.to_lowercase();
+    if !ACCEPTED_USERS.contains(&user_lower.as_str()) {
+        return Err(smtp_err(
+            SMTP_ERR_MAILBOX_UNAVAILABLE,
+            "Mailbox unavailable",
+        ));
+    }
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
aterga: Added a unit-test module at the bottom of `smtp.rs` covering:

- missing envelope / missing message → `SMTP_ERR_SYNTAX_ERROR`
- non-numeric recipient user → `SMTP_ERR_MAILBOX_UNAVAILABLE`
- too many headers, oversize header name, oversize header value, oversize body
- non-UTF-8 body does not trap and stays under `MAX_BODY_BYTES`
- multi-byte (Cyrillic) subject truncates on a char boundary without panicking
- `format_address` lowercases both parts
- happy-path round-trip
- `validate_envelope_only` accepts envelope-only requests and rejects missing envelope

13 tests total, all passing via `cargo test -p internet_identity_interface --lib`.

src/internet_identity/src/main.rs
```diff
@@ -29,6 +29,7 @@ use internet_identity_interface::internet_identity::types::openid::{
     OpenIdCredentialAddError, OpenIdCredentialRemoveError, OpenIdDelegationError,
     OpenIdPrepareDelegationResponse,
 };
+use internet_identity_interface::internet_identity::types::smtp::{SmtpRequest, SmtpResponse};
```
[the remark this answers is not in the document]
aterga: Fixed — the duplicate `use ...types::smtp::{...}` inside `mod smtp_gateway` has been removed. The outer import in `main.rs` is the one we have to keep: `candid::export_service!()` expands at top-level scope and needs `PostboxEmail`/`SmtpRequest`/`SmtpResponse` visible there. With the inner duplicate gone, clippy is clean under `-D warnings`.

src/internet_identity_interface/src/internet_identity/types/smtp.rs
```diff
@@ -0,0 +1,231 @@
...
+
+fn smtp_err(code: u64, message: impl Into<String>) -> SmtpResponse {
+    SmtpResponse::Err(SmtpRequestError {
+        code,
+        message: message.into(),
+    })
+}
+
+fn format_address(addr: &SmtpAddress) -> String {
+    format!("{}@{}", addr.user, addr.domain)
```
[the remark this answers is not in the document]
aterga: Fixed — `format_address` now lowercases both the user and the domain before returning, so the stable-map key is canonical regardless of envelope casing. New unit test `format_address_lowercases_both_parts` pins the behavior.

src/internet_identity/src/dkim.rs
aterga: Good catch. Today we only honor `l=` when computing the body hash — we still retain the full untrusted body in the postbox. Fixing this properly means either truncating the stored body to the signed prefix, or carrying the trust boundary through storage so the UI can render only the verified portion. I'm leaving this as a follow-up in the DKIM hardening track rather than addressing it in this PoC PR; noted in the commit message under 'known remaining DKIM gaps'.

src/internet_identity/src/dkim.rs
aterga: Agreed. Full DKIM-Signature tag parsing (folding, whitespace tolerance, multiple `b=`-like substrings inside values) isn't there yet. Our current parser is a simple split-on-`;` and split-on-`=` pass that handles the shapes real senders emit today but would misread a hand-crafted malicious header. Replacing it with either a dedicated DKIM parser crate or a small state machine is tracked for the DKIM hardening follow-up — not fixed in this PoC.

src/internet_identity/src/dkim.rs
aterga: Fixed for the most concrete part — `p=` lookup is now `eq_ignore_ascii_case` so a TXT record using `P=` won't drop the key on the floor. Broader DNS TXT parsing tolerance (folding across chunks, arbitrary whitespace) is tracked with the other DKIM parsing hardening. Resolving this one since the specific case you called out is addressed.

src/internet_identity/src/dkim.rs
aterga: Fixed — after decoding the `p=` tag we now check `RsaPublicKey::n().bits() >= DKIM_MIN_RSA_KEY_BITS` (1024) and reject otherwise. 2048 is the preferred target but we keep 1024 as the floor so senders still on 1024-bit keys continue to work. Implemented in `parse_dkim_public_key`.

src/internet_identity/src/dkim.rs
aterga: Partially addressed. The concrete bug — the extra space after the colon under *simple* header canonicalization — is fixed: we now build `"{name}:{value}"` without any whitespace insertion. The broader point (full byte-exactness over the original wire bytes including folding) isn't addressed in this PR, because our SMTP gateway hands us pre-parsed `(name, value)` pairs rather than the raw header bytes — we'd need a gateway contract change to thread the originals through. Leaving the thread open to track that larger item.

src/internet_identity/src/dkim.rs
aterga: Partially addressed in this round:

- `v=` — already enforced in `parse_dkim_signature` (rejects anything other than `1`).
- `x=` (expiration) — now parsed and enforced: an expired signature fails the `SignatureValid` check with `Signature expired (x=..., now=...)` as the detail. Check is against `ic_cdk::api::time() / 1e9`.
- `t=` (timestamp) — parsed for completeness but not enforced (a future-dated signature is still accepted today).
- `i=` (identity alignment) and DNS-side `k=` — not implemented yet.

Leaving the thread open to track `i=`, `k=`, and the `t=` future-dated check.

src/internet_identity/src/dkim.rs
aterga: Fixed — `verify_email_dkim` now collects *all* `DKIM-Signature` headers (`headers.iter().filter(eq_ignore_ascii_case)`) and runs the full check pipeline against each via a new `verify_single_signature` helper. Returns `Verified` as soon as any signature passes; if all fail, returns `Unverified` with the last attempt's per-step checks so the UI still has something to show. This should fix mailing-list forwarding and multi-hop authentication paths.

src/internet_identity/src/dkim.rs
aterga: Thanks @sea-snake — all your review comments have been addressed or explicitly deferred in replies above. Resolving the ping.

aterga:
Thanks @sea-snake — agreed, DMARC is the right next layer.

Tracking this alongside the existing deferred DKIM hardening items rather than addressing it in this PoC PR. A correct DMARC alignment check needs:

1. Resolve the `From:` header domain (Y) — currently we only consume the SMTP envelope's `MAIL FROM`, which DMARC explicitly does not use for alignment. We'd need to parse the `From:` header from the message itself.
2. DoH lookup of `_dmarc.<Y>` and parse the policy (`p=`, `sp=`, `adkim=`, `aspf=`, `pct=`, …).
3. Alignment check against each verified DKIM `d=` (X):
   - `adkim=s` → exact match required.
   - `adkim=r` (default) → organizational-domain match. This needs a Public Suffix List, which is non-trivial to ship inside the canister (~190 KB compressed list, must be kept reasonably current). Cheaper interim approach: treat `adkim=r` as "X must equal Y or be a subdomain of Y" without PSL — strictly more permissive than the spec, but at least stops obvious spoofing without a full PSL implementation.
4. Policy enforcement: `p=reject` / `p=quarantine` / `p=none` mapping to our verification status. We probably want a new `DkimVerificationStatus::SpoofingSuspected` (or extend `Unverified`) with a DMARC-failure detail.
5. Per-failure record (`rua`/`ruf` reporting) — out of scope for this canister.

I've added DMARC to the open follow-up list so it doesn't get lost. This PoC PR keeps the threads referenced in earlier replies open: trusted-body retention, full DKIM-Signature parser, full byte-exact canonicalization, policy tags `i=`/`k=` + future-dated `t=`, and now DMARC alignment.

aterga:
Closing this PoC as planned: the production email-recovery stack it explored (DKIM/DMARC/DNSSEC verification, the `email_recovery` API, and the frontend wizards) has landed on `main`, and the design doc (#3836) anticipated closing this PR once that happened. The Postbox storage UI itself was deemed out of scope.

---

PR #3768, opened 2026-04-10
feat(fe): support ii-icrc3-attributes JSON-RPC method

sea-snake:
Apps using the new `AuthClient.requestAttributes()` method send an `ii-icrc3-attributes` JSON-RPC request with a different message format than the existing `ii_attributes` method. Internet Identity needs to handle this new method while still supporting the legacy flow.

# Changes

**Frontend**
- Added `Icrc3AttributesParamsSchema` for validating the new `{ keys, nonce }` params format.
- Added `handleIcrc3Attributes` handler that calls the ICRC-3 canister methods and responds with `{ data, signature }` as base64.
- Regenerated candid bindings to include `prepare_icrc3_attributes` and `get_icrc3_attributes`.

**Test app**
- Added ICRC-3 attribute request support with a "Use ICRC-3 attributes" checkbox toggle.

# Tests

- Split OpenID e2e tests into `openid.spec.ts` (ICRC-3) and `openid-legacy.spec.ts` (legacy `ii_attributes`).
- ICRC-3 spec covers: authenticate-only, name+email, unavailable attributes, and verified_email.

---
<div align="right">Next: dfinity/internet-identity#3770 →</div>


src/internet_identity/src/attributes.rs
```diff
@@ -267,78 +267,51 @@ impl Anchor {
...
-        if !problems.is_empty() {
-            return Err(PrepareIcrc3AttributeError::AttributeMismatch { problems });
+            // Compute the certified key.
+            let certified_key = if spec.omit_scope {
+                spec.key.attribute_name.to_string()
+            } else {
+                spec.key.to_string()
+            };
+
+            // Skip duplicates silently.
```
aterga, line 309: So is this the executive summary of the BE changes in this PR?

sea-snake: As discussed, BE changes are reverted from this PR.

aterga (approved):

PR #3771, opened 2026-04-10
feat(fe): ICRC-3 attribute consent UI with store-driven rendering

sea-snake:
When a dapp requests ICRC-3 attributes, users previously had no way to review or choose which attributes to share — everything was handled via implicit consent. This PR adds a consent UI that lets users review requested attributes, pick from available options (e.g., which provider's email to share), and deny sharing entirely.

# Changes

- **Attribute consent store.** Context (as a promise) and consent result, following the same pattern as the authorization store. The handler sets the context promise immediately so the UI can show a loading skeleton while attributes resolve.
- **AttributeConsentView.** Renders after authorization when consent is needed. Shows attribute groups with checkboxes, a "Deny All" option, and a "Continue" button. Merges email/verified_email into a single "Email" picker — prefers the intersection of verified + plain emails, falls back to plain if no verified emails exist.
- **AttributePicker.** Inline expandable dropdown for attributes with multiple options (e.g., email from different providers). Shows the selected provider's logo + value, with a "Change" button that expands the option list.
- **Implicit consent detection.** Scoped OpenID attributes (name, email, verified_email) matching the user's provider are auto-consented without showing the UI. Uses `pendingOpenIdIssuerStore` set early in the OpenID resume flow for synchronous detection.
- **Skeleton CSS utility.** Reusable `.skeleton` class with shimmer animation using `color-mix` for theme-adaptive appearance.
- **Promise-based context.** The consent view `{#await}`s the context promise — shows skeleton while resolving, then the real content.
- **Accessible markup.** `<fieldset>`, `role="listbox"`/`role="option"`, `aria-expanded`, `aria-selected`, `aria-busy` on loading state.

---

<div align="left">← Previous: dfinity/internet-identity#3781</div>


sea-snake:
Superseded by #3777

PR #3777, opened 2026-04-14
refactor(fe): decouple channel handlers from UI through store-driven architecture

sea-snake:
The authorize flow previously had request handling intertwined with UI components — handlers would trigger UI changes directly, and UI components would drive handler logic. This made the flow hard to follow and brittle to change.

This PR introduces a clean separation: **handlers ← stores → UI**. Channel handlers write to stores and observe stores (via `waitForStore`). The UI reads stores to decide what to render, and writes to stores when the user takes action. Neither side knows about the other.

# Changes

- **Extract channel handlers into dedicated files.** `channelHandlers/icrc25.ts` (standards, permissions), `channelHandlers/delegation.ts` (ICRC-34), and `channelHandlers/attributes.ts` (legacy + ICRC-3 attributes) — each self-contained with their own imports.
- **Split authorization store into context and outcome.** `authorizationStore` holds the request context (`effectiveOrigin`), `authorizedStore` holds the outcome (`accountNumber`). Handlers `waitForStore(authorizedStore)` instead of polling a combined state.
- **Add `waitForStore` utility.** Generic helper that resolves when a Svelte store value satisfies a condition — eliminates repeated subscribe/unsubscribe boilerplate across handlers.
- **Gate attribute handlers on dapp opt-in.** Attributes are only served to known dapps with `certifiedAttributes: true`, or in dev environments (`fetch_root_key`). Unknown origins get an error response.
- **Channel error component.** Extracted `ChannelError.svelte` with per-error-type messages, shared across channel, pending-channel, and resuming-channel layouts.
- **Fix pending/resuming channel layout guards.** Pending-channel now renders errors. Resuming-channel waits for channel establishment before rendering children, shows errors when appropriate.
- **Remove `AuthorizationChannel.svelte`.** Layouts call `channelStore.establish()` directly and gate rendering on store state.

---

<div align="left">← Previous: dfinity/internet-identity#3791</div>
<div align="right">Next: dfinity/internet-identity#3781 →</div>


src/frontend/src/lib/stores/channelStore.ts
```diff
@@ -48,71 +82,488 @@ const supportedStandards = [
...
+    const configIssuer = getConfigIssuer(authenticated);
+    if (configIssuer === undefined) {
+      return;
+    }
+
+    // Use the request's derivation origin if provided, else channel origin,
+    // remapped to legacy domain for backwards compatibility.
+    const origin = remapToLegacyDomain(
+      paramsResult.data.icrc95DerivationOrigin ?? channel.origin,
     );
```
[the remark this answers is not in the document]
sea-snake: Added `validateDerivationOrigin()` call before computing the origin in `handleLegacyAttributes`, matching the delegation handler's validation pattern. Returns a JSON-RPC error to the caller if invalid.

src/frontend/src/lib/stores/channelStore.ts
```diff
@@ -48,71 +82,488 @@ const supportedStandards = [
...
+    const configIssuer = getConfigIssuer(authenticated);
+    if (configIssuer === undefined) {
+      return;
+    }
+
+    // Use the request's derivation origin if provided, else channel origin,
+    // remapped to legacy domain for backwards compatibility.
+    const origin = remapToLegacyDomain(
+      paramsResult.data.icrc95DerivationOrigin ?? channel.origin,
+    );
```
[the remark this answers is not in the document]
sea-snake: Same fix as above — added `validateDerivationOrigin()` in `handleIcrc3Attributes` before using the derivation origin.

src/frontend/src/lib/stores/channelStore.ts
```diff
@@ -48,71 +82,488 @@ const supportedStandards = [
...
+      const effectiveOrigin = remapToLegacyDomain(
+        params.icrc95DerivationOrigin ?? channel.origin,
+      );
+      authorizationStore.setContext(effectiveOrigin);
+
+      // Authorization is the commit point — the user may switch identities
+      // freely before this. Once authorized, the UI is no longer needed.
+      const { accountNumber } = await waitForAuthorization();
+      channelIdleStore.set(true);
+
```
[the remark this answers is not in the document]
sea-snake: Added `channelIdleStore.set(false)` at the start of `handleDelegationRequest` so it resets when a new delegation request comes in. Attribute requests don't independently reset idle since they piggyback on the delegation flow's authorization — they call `waitForAuthorization()` which blocks until the delegation has been authorized.

src/frontend/src/lib/stores/channelStore.ts
```diff
@@ -48,71 +82,488 @@ const supportedStandards = [
...
+        channelErrorStore.set("connection-closed"),
+      );
+    })().catch((error) => {
+      console.error(error);
+      if (error instanceof PostMessageUnsupportedError) {
+        void goto("/unsupported");
+        return;
+      }
+      channelErrorStore.set("unable-to-connect");
+    });
```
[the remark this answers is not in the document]
sea-snake: Unwrap `AggregateError.errors` before checking for `PostMessageUnsupportedError`. Also moved the navigation to `/unsupported` out of the store entirely — it now sets `channelErrorStore` to `"unsupported-browser"` and the layouts handle routing, same as all other errors.

src/frontend/src/lib/stores/channelHandlers/attributes.ts
```diff
@@ -0,0 +1,267 @@
...
+                ),
+                expiration: expires_at_timestamp_ns.toString(),
+              },
+            ]),
+          ),
+        },
+      });
+    } catch (error) {
+      console.error(error);
+    }
```
timothyaterton, line 152: **Non-blocking (hardening):** Both attribute handlers (`handleLegacyAttributes` here, and `handleIcrc3Attributes` at line 264) silently swallow canister errors — `console.error(error)` without calling `onError()` or sending an error response to the channel.

In contrast, the delegation handler at `delegation.ts:115-118` correctly calls `onError("delegation-failed")`, which sets `channelErrorStore` and shows the user an error banner.

The consequence here is that if `prepare_attributes` or `get_attributes` fails (network error, canister reject, etc.), the requesting dapp gets no response at all (silent timeout), and the II UI shows no error state. This could mask legitimate failures during attribute certification.

Consider calling `onError()` here (perhaps a new `"attributes-failed"` error type) for parity with the delegation handler.

src/frontend/src/lib/stores/authentication.store.ts
```diff
@@ -20,7 +21,7 @@ export interface Authenticated {
   actor: ActorSubclass<_SERVICE>;
   authMethod:
     | { passkey: { credentialId: Uint8Array } }
-    | { openid: { iss: string; sub: string } }
+    | { openid: { iss: string; sub: string; metadata: MetadataMapV2 } }
```
timothyaterton: **Non-blocking (observation):** The new `metadata: MetadataMapV2` field expands the data surface of `authenticatedStore`. Previously, only `iss` and `sub` were stored for OpenID auth methods. Now all JWT claims (except `iss`, `sub`, `loginHint`) are captured via `Object.entries(restClaims)` in `authFlow.svelte.ts`.

Since `authenticatedStore` is a `Readable` store, any subscribing component gets access to these claims (which can include `email`, `name`, `picture`, `aud`, `exp`, `iat`, and any provider-specific custom claims).

This is needed for the attribute certification flow (issuer matching), so the addition is justified. Just noting that any future subscriber to `authenticatedStore` should be aware it now carries more sensitive data than before.

timothyaterton (commented):
## Security review — channel refactor (store-driven architecture)

Reviewed this PR's security-sensitive paths: origin/source validation on incoming channel messages, delegation handling, ICRC-25 handlers, attribute certification, store-level data exposure, Rust backend type changes, and test coverage.

### Verdict: **No blocking security concerns found.**

The core security properties are properly maintained through the refactor:

- **Origin validation**: `PostMessageChannel` correctly validates both `event.origin` and `event.source` before processing any request (postMessage.ts). Once a channel is established, the origin/source pair is locked — no other window can hijack it.
- **Derivation origin validation**: `validateDerivationOrigin()` is correctly called in both the delegation handler and both attribute handlers before any canister call uses the origin. The `.well-known/ii-alternative-origins` certificate check is intact.
- **PostMessage targeting**: All `channel.send()` calls route through `this.#source.postMessage(response, this.#origin)` — responses are targeted to the established origin, never using `"*"`.
- **User authorization gate**: The delegation handler correctly calls `waitForStore(authorizedStore)` before issuing any delegation. No delegation can be created without explicit user authorization.
- **Implicit consent filtering**: `filterImplicitConsentKeys()` correctly restricts attribute responses to only `name`, `email`, `verified_email` scoped to the config issuer. The dapp allowlist check (`dapp?.certifiedAttributes`) is enforced.
- **Rust type improvements**: The change from raw bytes to typed `Icrc3Value` (Text for origin, Nat for timestamp, Blob for nonce) is strictly safer — it eliminates ambiguous encoding and the tests correctly validate the new types.
- **Test coverage**: No security-relevant tests were removed. New integration test `should_return_error_for_unavailable_icrc3_attributes` strengthens coverage. E2E tests updated to validate correct ICRC-3 value types.

Two non-blocking observations are flagged inline.

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -198,7 +201,7 @@ export class AuthFlow {
       await authenticationStore.set({
         identity,
         identityNumber,
-        authMethod: { openid: { iss, sub } },
+        authMethod: { openid: { iss, sub, metadata: jwtMetadata } },
```
aterga: Let's remove this for now

aterga: (the metadata, here and below)

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -79,7 +80,10 @@ export class AuthLastUsedFlow {
           mediation: "optional",
           loginHint: lastUsedIdentity.authMethod.openid.loginHint,
         });
-        const { iss, sub } = decodeJWT(jwt);
+        const { iss, sub, ...restClaims } = decodeJWT(jwt);
+        const metadata: MetadataMapV2 = Object.entries(restClaims)
```
aterga: ditto

src/frontend/src/lib/stores/channelHandlers/attributes.ts
```diff
@@ -0,0 +1,267 @@
...
+): string[] =>
+  keys.filter(
+    (key) =>
+      key === `openid:${configIssuer}:name` ||
+      key === `openid:${configIssuer}:email` ||
+      key === `openid:${configIssuer}:verified_email`,
+  );
+
+/**
+ * Handle legacy `ii_attributes` requests.
```
aterga: what does legacy mean in this context?

src/frontend/src/lib/stores/channelHandlers/attributes.ts
```diff
@@ -0,0 +1,267 @@
...
+              attributes: [implicitKeys],
+            })
+            .then(throwCanisterError)
+        : [];
+    const availableKeys = new Set(available.map(([key]) => key));
+    const attributeKeys = implicitKeys.filter((key) => availableKeys.has(key));
+
+    try {
+      const { message } = await authenticated.actor
+        .prepare_icrc3_attributes({
```
aterga, line 234: Graceful error support opportunity

src/frontend/src/lib/stores/channelHandlers/attributes.ts
```diff
@@ -0,0 +1,267 @@
...
+            value: [],
+            omit_scope: false,
+          })),
+          nonce: z.util.base64ToUint8Array(paramsResult.data.nonce),
+        })
+        .then(throwCanisterError);
+
+      const { signature } = await retryFor(5, () =>
+        authenticated.actor
+          .get_icrc3_attributes({
```
aterga, line 249: Graceful error support opportunity

src/frontend/src/lib/stores/channelHandlers/attributes.ts
```diff
@@ -0,0 +1,267 @@
...
+      await channel.send({
+        jsonrpc: "2.0",
+        id: request.id,
+        result: {
+          data: z.util.uint8ArrayToBase64(new Uint8Array(message)),
+          signature: z.util.uint8ArrayToBase64(new Uint8Array(signature)),
+        },
+      });
+    } catch (error) {
+      console.error(error);
```
aterga, line 267: Let's handle errors gracefully, e.g., like for the delegations in src/frontend/src/lib/stores/channelHandlers/delegation.ts

src/frontend/src/lib/stores/authentication.store.ts
aterga: ditto - remove metadata (since it's not useful for this PR)

src/frontend/src/lib/stores/channelStore.ts
```diff
@@ -1,25 +1,57 @@
...
+// ---------------------------------------------------------------------------
+// Transports
+// ---------------------------------------------------------------------------
+
 const getTransports = (): Transport[] => {
   const primaryOrigin = getPrimaryOrigin();
   return [
     new PostMessageTransport(),
     new LegacyTransport(
-      // Redirect requests and responses between related origins and primary origin
```
aterga, line 22: why was this removed?

src/frontend/src/routes/(new-styling)/(channel)/authorize/(panel)/continue/+page.svelte
```diff
@@ -108,44 +96,47 @@
...
+                identityNumber,
+                $authorizationContextStore.effectiveOrigin,
+              )
               .then(throwCanisterError)
               .then((account) => account.account_number[0])
-          : defaultAccountNumber,
-      );
+          : defaultAccountNumber;
+      void triggerDropWaveAnimation();
+      authorize(accountNumber);
```
aterga, line 110: please double check that this doesn't need an await

aterga (approved):
Let's consider what are out options for having more graceful degradation in the error handing code. Otherwise, LGTM

src/frontend/src/lib/stores/channelHandlers/attributes.ts
sea-snake: Added `onError("attributes-failed")` to the catch blocks in both `handleLegacyAttributes` and `handleIcrc3Attributes`, along with a new `"attributes-failed"` `ChannelError` variant and a matching error message in `ChannelError.svelte`.

src/frontend/src/lib/stores/authentication.store.ts
sea-snake: Removed the `metadata` field from `authenticationStore`'s OpenID auth method — it wasn't needed for this PR. OpenID issuer detection for the ICRC-3 attribute handler now uses `pendingOpenIdIssuerStore` (from #3790) instead of auth-state metadata. `lastUsedIdentitiesStore` still carries metadata since `findConfig` needs it; that's a separate concept.

src/frontend/src/lib/flows/authFlow.svelte.ts
sea-snake: Done — removed metadata here.

sea-snake: Removed all metadata construction/usage in `authFlow` and `authLastUsedFlow`.

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
sea-snake: Done — removed metadata here too.

src/frontend/src/lib/stores/channelHandlers/attributes.ts
sea-snake: Done — both handlers now call `onError("attributes-failed")` mirroring `delegation.ts`, with a matching `ChannelError` variant and a localized message in `ChannelError.svelte`.

src/frontend/src/lib/stores/authentication.store.ts
sea-snake: Done — reverted the `authentication.store` type back to `{ iss: string; sub: string }`.

src/frontend/src/lib/stores/channelHandlers/attributes.ts
sea-snake: Clarified the doc comment — "legacy" here means the pre-ICRC-3 attribute sharing method (`ii_attributes`) that produces per-attribute signatures. Changed the text to say that explicitly.

src/frontend/src/routes/(new-styling)/(channel)/authorize/(panel)/continue/+page.svelte
sea-snake: Double-checked — `onAuthorize` takes a `Promise<bigint | undefined>`, and in `ContinueView` we pass `Promise.resolve(accountNumber)`, so no `await` is needed here.

src/frontend/src/lib/stores/channelHandlers/attributes.ts
sea-snake: Addressed — added `onError("attributes-failed")` in both handlers' catch blocks. See the main reply at comment 3118392683.

src/frontend/src/lib/stores/channelHandlers/attributes.ts
sea-snake: Same — `onError("attributes-failed")` is now called from the ICRC-3 handler's catch too.

src/frontend/src/lib/stores/channelStore.ts
sea-snake: Restored the `// Redirect requests and responses between related origins and primary origin` comment above `LegacyTransport` — it got lost in the refactor.

PR #3778, opened 2026-04-15
feat(be): add DiscoverableOidcConfig type with OIDC discovery support

timothyaterton:
## Summary

- Introduces `DiscoverableOidcConfig` type and `oidc_configs` init field (mutually exclusive with existing `openid_configs`) that relies on OIDC discovery (`.well-known/openid-configuration`) instead of requiring all provider details in the static config
- Adds `DiscoverableProvider` that periodically fetches discovery metadata to obtain `issuer` and `jwks_uri` for JWT verification
- Adds `discovered_oidc_configs` query endpoint returning `OidcConfig` with resolved provider state
- Validates that discovered `issuer` domain matches the `discovery_url` domain (prevents impersonation)
- When both `openid_configs` and `oidc_configs` are provided, falls back to `openid_configs` as the proven path

## Test plan

- [x] 7 new integration tests in `config/oidc_configs.rs` (init, upgrade, retain, XOR, query)
- [x] Existing `openid_configs` tests pass unchanged (backward compat)
- [ ] Manual E2E with deployed canister using `oidc_configs`

---
[Next PR >](https://github.com/dfinity/internet-identity/pull/3784)

### DIFF
--- a/src/internet_identity/Cargo.toml
+++ b/src/internet_identity/Cargo.toml
@@ -30,6 +30,7 @@ captcha = { git = "https://github.com/dfinity/captcha", rev = "9c0d2dd9bf519e255
 
 # OpenID deps
 identity_jose = { git = "https://github.com/dfinity/identity.rs.git", rev = "aa510ef7f441848d6c78058fe51ad4ad1d9bd5d8", default-features = false }
+url = { version = "2.5", default-features = false }
 
 # All IC deps
 candid.workspace = true
--- a/src/internet_identity/Cargo.toml
+++ b/src/internet_identity/Cargo.toml
@@ -30,7 +30,7 @@ captcha = { git = "https://github.com/dfinity/captcha", rev = "9c0d2dd9bf519e255
 
 # OpenID deps
 identity_jose = { git = "https://github.com/dfinity/identity.rs.git", rev = "aa510ef7f441848d6c78058fe51ad4ad1d9bd5d8", default-features = false }
-url = { version = "2.5", default-features = false }
+url = { version = "2.5", default-features = false, features = ["std"] }
 
 # All IC deps
 candid.workspace = true
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -385,24 +385,21 @@ type OpenIdConfig = record {
     email_verification : opt OpenIdEmailVerification;
 };
 
-// Simplified OIDC provider config that relies on discovery.
-// client_id must be set. discovery_url is the standard OIDC .well-known endpoint.
+// SSO provider config that uses two-hop discovery.
+// The backend fetches https://{discovery_domain}/.well-known/ii-openid-configuration
+// for { client_id, openid_configuration } and then fetches the standard OIDC
+// discovery at openid_configuration for { issuer, jwks_uri }.
 type DiscoverableOidcConfig = record {
-    name : text;
-    logo : text;
-    discovery_url : text;
-    client_id : opt text;
-    email_verification : opt OpenIdEmailVerification;
+    discovery_domain : text;
 };
 
-// Resolved OIDC config with discovered fields.
+// Resolved SSO provider state.
+// All fields other than discovery_domain are None until discovery completes.
 type OidcConfig = record {
-    name : text;
-    logo : text;
-    discovery_url : text;
+    discovery_domain : text;
     client_id : opt text;
+    openid_configuration : opt text;
     issuer : opt text;
-    email_verification : opt OpenIdEmailVerification;
 };
 
 type OpenIdCredentialKey = record { Iss; Sub };
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -280,6 +280,8 @@ type InternetIdentityInit = record {
     new_flow_origins : opt vec text;
     // Configurations for OpenID clients
     openid_configs : opt vec OpenIdConfig;
+    // Simplified OIDC configs using discovery. Mutually exclusive with openid_configs.
+    oidc_configs : opt vec OidcConfig;
     // Configuration for Web Analytics
     analytics_config : opt opt AnalyticsConfig;
     // Configuration to show dapps explorer or not
@@ -380,6 +382,27 @@ type OpenIdConfig = record {
     auth_uri : text;
     auth_scope : vec text;
     fedcm_uri : opt text;
+    email_verification : opt OpenIdEmailVerification;
+};
+
+// Simplified OIDC provider config that relies on discovery.
+// For dedicated providers: client_id is set, discovery_url is standard OIDC endpoint.
+// For SSO providers: client_id is null, discovery_url is custom ii-openid-configuration.
+type OidcConfig = record {
+    name : text;
+    logo : text;
+    discovery_url : text;
+    client_id : opt text;
+    email_verification : opt OpenIdEmailVerification;
+};
+
+// Resolved OIDC config with discovered fields.
+type ActiveOidcConfig = record {
+    name : text;
+    logo : text;
+    discovery_url : text;
+    client_id : opt text;
+    issuer : opt text;
     email_verification : opt OpenIdEmailVerification;
 };
 
@@ -1174,6 +1197,10 @@ service : (opt InternetIdentityInit) -> {
     // HTTP Gateway protocol
     // =====================
     http_request : (request : HttpRequest) -> (HttpResponse) query;
+
+    // OIDC Discovery
+    // ===============
+    active_oidc_configs : () -> (vec ActiveOidcConfig) query;
 
     // Internal Methods
     // ================
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -386,8 +386,7 @@ type OpenIdConfig = record {
 };
 
 // Simplified OIDC provider config that relies on discovery.
-// For dedicated providers: client_id is set, discovery_url is standard OIDC endpoint.
-// For SSO providers: client_id is null, discovery_url is custom ii-openid-configuration.
+// client_id must be set. discovery_url is the standard OIDC .well-known endpoint.
 type DiscoverableOidcConfig = record {
     name : text;
     logo : text;
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -397,7 +397,7 @@ type OidcConfig = record {
 };
 
 // Resolved OIDC config with discovered fields.
-type ActiveOidcConfig = record {
+type DiscoveredOidcConfig = record {
     name : text;
     logo : text;
     discovery_url : text;
@@ -1200,7 +1200,7 @@ service : (opt InternetIdentityInit) -> {
 
     // OIDC Discovery
     // ===============
-    active_oidc_configs : () -> (vec ActiveOidcConfig) query;
+    discovered_oidc_configs : () -> (vec DiscoveredOidcConfig) query;
 
     // Internal Methods
     // ================
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -280,7 +280,8 @@ type InternetIdentityInit = record {
     new_flow_origins : opt vec text;
     // Configurations for OpenID clients
     openid_configs : opt vec OpenIdConfig;
-    // Simplified OIDC configs using discovery. Mutually exclusive with openid_configs.
+    // Deprecated in init args: use add_discoverable_oidc_config update call instead.
+    // Kept for backward compatibility and config() query output.
     oidc_configs : opt vec DiscoverableOidcConfig;
     // Configuration for Web Analytics
     analytics_config : opt opt AnalyticsConfig;
@@ -1197,6 +1198,7 @@ service : (opt InternetIdentityInit) -> {
     // OIDC Discovery
     // ===============
     discovered_oidc_configs : () -> (vec OidcConfig) query;
+    add_discoverable_oidc_config : (DiscoverableOidcConfig) -> ();
 
     // Internal Methods
     // ================
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -281,7 +281,7 @@ type InternetIdentityInit = record {
     // Configurations for OpenID clients
     openid_configs : opt vec OpenIdConfig;
     // Simplified OIDC configs using discovery. Mutually exclusive with openid_configs.
-    oidc_configs : opt vec OidcConfig;
+    oidc_configs : opt vec DiscoverableOidcConfig;
     // Configuration for Web Analytics
     analytics_config : opt opt AnalyticsConfig;
     // Configuration to show dapps explorer or not
@@ -388,7 +388,7 @@ type OpenIdConfig = record {
 // Simplified OIDC provider config that relies on discovery.
 // For dedicated providers: client_id is set, discovery_url is standard OIDC endpoint.
 // For SSO providers: client_id is null, discovery_url is custom ii-openid-configuration.
-type OidcConfig = record {
+type DiscoverableOidcConfig = record {
     name : text;
     logo : text;
     discovery_url : text;
@@ -397,7 +397,7 @@ type OidcConfig = record {
 };
 
 // Resolved OIDC config with discovered fields.
-type DiscoveredOidcConfig = record {
+type OidcConfig = record {
     name : text;
     logo : text;
     discovery_url : text;
@@ -1200,7 +1200,7 @@ service : (opt InternetIdentityInit) -> {
 
     // OIDC Discovery
     // ===============
-    discovered_oidc_configs : () -> (vec DiscoveredOidcConfig) query;
+    discovered_oidc_configs : () -> (vec OidcConfig) query;
 
     // Internal Methods
     // ================
--- a/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
+++ b/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
@@ -202,7 +202,12 @@ fn create_openid_credential_and_config(
     let OpenIDRegFinishArg { jwt, salt, name: _ } = openid_registration_data;
 
     let (openid_credential, openid_config_iss) = openid::with_provider(jwt, |provider| {
-        Ok((provider.verify(jwt, salt)?, provider.issuer()))
+        // `with_provider` already skipped providers with pending discovery, so
+        // `provider.issuer()` is guaranteed to be Some here.
+        Ok((
+            provider.verify(jwt, salt)?,
+            provider.issuer().unwrap_or_default(),
+        ))
     })?;
 
     Ok((openid_credential, openid_config_iss))
--- a/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
+++ b/src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
@@ -202,12 +202,14 @@ fn create_openid_credential_and_config(
     let OpenIDRegFinishArg { jwt, salt, name: _ } = openid_registration_data;
 
     let (openid_credential, openid_config_iss) = openid::with_provider(jwt, |provider| {
-        // `with_provider` already skipped providers with pending discovery, so
-        // `provider.issuer()` is guaranteed to be Some here.
-        Ok((
-            provider.verify(jwt, salt)?,
-            provider.issuer().unwrap_or_default(),
-        ))
+        // `with_provider` only yields providers whose discovery has completed, so
+        // `issuer()` is always Some — surface an explicit error if that ever breaks.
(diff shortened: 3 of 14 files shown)


aterga (dismissed):

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -262,8 +264,12 @@ pub struct InternetIdentityInit {
     pub register_rate_limit: Option<RateLimitConfig>,
     pub captcha_config: Option<CaptchaConfig>,
     pub related_origins: Option<Vec<String>>,
+    /// Deprecated: origins that use the new authentication flow. Mutually exclusive with `oidc_configs`.
     pub new_flow_origins: Option<Vec<String>>,
+    /// Deprecated: use `oidc_configs` instead. Mutually exclusive with `oidc_configs`.
     pub openid_configs: Option<Vec<OpenIdConfig>>,
+    /// Simplified OIDC configs that use discovery. Mutually exclusive with `openid_configs`.
```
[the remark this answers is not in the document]
timothyaterton: Addressed: `apply_install_arg` now enforces the XOR. When `oidc_configs` is supplied, `new_flow_origins` is cleared (and vice versa). When `openid_configs` is supplied alongside `oidc_configs`, a warning is logged and `openid_configs` takes precedence, clearing `oidc_configs`.

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -380,6 +386,36 @@ pub struct OpenIdConfig {
...
 
+/// Simplified OIDC provider configuration that relies on OIDC discovery
+/// instead of requiring all provider details in the init config.
+///
+/// For dedicated providers (Google, Apple, Microsoft): `client_id` is set,
+/// `discovery_url` points to the standard `.well-known/openid-configuration` endpoint.
+///
+/// For SSO providers: `client_id` is `None`, `discovery_url` points to a custom
+/// `ii-openid-configuration` endpoint that provides `client_id` and a reference
+/// to the standard OIDC discovery endpoint.
```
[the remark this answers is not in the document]
timothyaterton: Addressed: the doc comment on `DiscoverableOidcConfig` has been updated to clarify that `client_id` must be set and that `discovery_url` points to the standard `.well-known/openid-configuration` endpoint. The ambiguous SSO-without-client-id description was removed to match the actual implementation.

src/internet_identity/src/openid/generic.rs
```diff
@@ -215,6 +223,290 @@ impl Provider {
...
+    certs: Rc<RefCell<Vec<Jwk>>>,
+    email_verification: Option<OpenIdEmailVerificationScheme>,
+}
+
+impl OpenIdProvider for DiscoverableProvider {
+    fn issuer(&self) -> String {
+        self.discovered_issuer
+            .borrow()
+            .clone()
+            .unwrap_or_default()
```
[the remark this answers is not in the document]
timothyaterton: `with_provider` in `openid.rs` now detects when any registered provider has an empty issuer (discovery still pending) and returns a dedicated error — `"OIDC provider discovery is still in progress, please try again shortly"` — instead of the generic `Unsupported issuer` message.

src/internet_identity/src/openid/generic.rs
```diff
@@ -215,6 +223,290 @@ impl Provider {
...
+            if let Err(err) = validate_issuer_domain(&task.discovery_url, &doc.issuer) {
+                ic_cdk::println!("Discovery issuer validation failed: {err}");
+                continue;
+            }
+            task.issuer_ref.replace(Some(doc.issuer));
+
+            if !*task.certs_started.borrow() {
+                *task.certs_started.borrow_mut() = true;
+                schedule_fetch_certs(doc.jwks_uri, Rc::clone(&task.certs_ref), Some(0));
+            }
```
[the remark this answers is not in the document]
timothyaterton: Addressed: `DiscoveryState` now includes a `last_jwks_uri` field. In `run_discovery_tasks`, after each successful discovery fetch, the new `jwks_uri` is compared against the last seen value — if changed, `schedule_fetch_certs` is called immediately with the updated URI.

src/internet_identity/Cargo.toml
```diff
@@ -30,6 +30,7 @@ captcha = { git = "https://github.com/dfinity/captcha", rev = "9c0d2dd9bf519e255
 
 # OpenID deps
 identity_jose = { git = "https://github.com/dfinity/identity.rs.git", rev = "aa510ef7f441848d6c78058fe51ad4ad1d9bd5d8", default-features = false }
+url = { version = "2.5", default-features = false }
```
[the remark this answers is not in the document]
timothyaterton: Addressed: `url` in `Cargo.toml` now explicitly sets `features = ["std"]` even with `default-features = false`, ensuring `Url::parse` has the std runtime it needs on the wasm target.

src/internet_identity/src/openid/generic.rs
```diff
@@ -215,6 +224,303 @@ impl Provider {
...
+/// Used with the new `OidcConfig` type.
+pub struct DiscoverableProvider {
+    client_id: Option<String>,
+    discovered_issuer: Rc<RefCell<Option<String>>>,
+    certs: Rc<RefCell<Vec<Jwk>>>,
+    email_verification: Option<OpenIdEmailVerificationScheme>,
+}
+
+impl OpenIdProvider for DiscoverableProvider {
+    fn issuer(&self) -> String {
```
sea-snake: As mentioned by Copilot this should at least be an Option<String> since the issuer might not be discovered yet. Those providers should be skipped when looking for a match with a JWT.

Also providers should be matched to a JWT by iss+aud(= client_id) pairs, just matching on iss is **not** sufficient.

src/internet_identity/src/openid/generic.rs
```diff
@@ -215,6 +224,303 @@ impl Provider {
...
+    static DISCOVERY_TASKS: RefCell<Vec<DiscoveryState>> = const { RefCell::new(vec![]) };
+}
+
+impl DiscoverableProvider {
+    pub fn create(config: DiscoverableOidcConfig) -> DiscoverableProvider {
+        let certs: Rc<RefCell<Vec<Jwk>>> = Rc::new(RefCell::new(vec![]));
+        let discovered_issuer: Rc<RefCell<Option<String>>> = Rc::new(RefCell::new(None));
+
+        DISCOVERY_TASKS.with_borrow_mut(|tasks| {
+            tasks.push(DiscoveryState {
```
sea-snake, line 373: There's no bound to tasks, as long as we allow list to only `dfinity.org` as first SSO this is fine. But long term we should look into how things scale with many SSO providers.

src/internet_identity/tests/integration/config/oidc_configs.rs
```diff
@@ -0,0 +1,162 @@
...
+};
+use internet_identity_interface::internet_identity::types::{
+    InternetIdentityInit, DiscoverableOidcConfig, OpenIdConfig,
+};
+
+fn example_oidc_config() -> DiscoverableOidcConfig {
+    DiscoverableOidcConfig {
+        name: "Google".into(),
+        logo: String::new(),
+        discovery_url: "https://accounts.google.com/.well-known/openid-configuration".into(),
```
sea-snake: Discovery config should be based on domain only e.g. 

```suggestion
        discovery_domain "accounts.google.com".into(),
```

This should be concatinated with `/.well-known/ii-openid-configuration`. This in turn should be a json file with a open id configuration key pointing to the actual .well-known/openid-configuration URL (could be any domain) and a client id.

src/internet_identity/tests/integration/config/oidc_configs.rs
```diff
@@ -0,0 +1,162 @@
...
+    env, install_ii_canister_with_arg, upgrade_ii_canister_with_arg, II_WASM,
+};
+use internet_identity_interface::internet_identity::types::{
+    InternetIdentityInit, DiscoverableOidcConfig, OpenIdConfig,
+};
+
+fn example_oidc_config() -> DiscoverableOidcConfig {
+    DiscoverableOidcConfig {
+        name: "Google".into(),
+        logo: String::new(),
```
sea-snake: SSO configs don't need a name and logo.

src/internet_identity/tests/integration/config/oidc_configs.rs
```diff
@@ -0,0 +1,162 @@
...
+        discovery_url: "https://accounts.google.com/.well-known/openid-configuration".into(),
+        client_id: Some("test-client-id".into()),
+        email_verification: None,
```
sea-snake: Neither does it have email verification.

src/internet_identity/src/openid/generic.rs
timothyaterton: Addressed in b6828827.

- `OpenIdProvider::issuer()` and a new `client_id()` both now return `Option<String>`. `Provider` (for static `OpenIdConfig`) always returns `Some(...)`; `DiscoverableProvider` returns whatever the two-hop discovery has populated so far.
- Providers with `issuer() == None || client_id() == None` are skipped during matching in both `OpenIdCredential::with_provider` and the JWT-side `openid::with_provider`. If every registered provider is pending, callers get the "OIDC provider discovery is still in progress" error instead of "Unsupported issuer".
- Matching is now done on the `(iss, aud)` pair. `PartialClaims` gained an `aud` field (accepting either `String` or `Vec<String>` per RFC 7519 — Microsoft sometimes ships `aud` as an array).
- New unit tests: `should_skip_provider_when_discovery_pending`, `should_disambiguate_providers_by_aud`.

src/internet_identity/src/openid/generic.rs
timothyaterton: Captured this as a TODO comment above the `DISCOVERY_TASKS` `thread_local` in [`generic.rs`](https://github.com/timothyaterton/internet-identity/blob/oidc-discovery-config/src/internet_identity/src/openid/generic.rs) — left it as-is for now since the canary allowlist limits `discovery_domain` to `dfinity.org` (see reply on the tests file). The TODO links back to this comment so it gets revisited when the allowlist widens.

src/internet_identity/tests/integration/config/oidc_configs.rs
timothyaterton: Addressed in b6828827.

- `DiscoverableOidcConfig` is now `{ discovery_domain: String }`. The backend builds `https://{discovery_domain}/.well-known/ii-openid-configuration` and fetches it for `{ client_id, openid_configuration }`, then fetches `openid_configuration` (which can be any domain) for the standard OIDC discovery document `{ issuer, jwks_uri }`.
- `client_id` is now discovered from hop 1, not declared in config.
- The issuer-host validation now compares against the `openid_configuration` URL host (hop 2) — the indirection from `discovery_domain` to `openid_configuration` is intentional and legitimate for an SSO setup hosting its IdP on a different domain (e.g., `dfinity.org` pointing at `dfinity.okta.com`).
- Additionally, canary allowlist added at setup time: `setup_oidc` traps if `discovery_domain` isn't on `ALLOWED_DISCOVERY_DOMAINS = &["dfinity.org"]`. Comment on the constant explains the canary phase and the plan to widen/configurable it later.

src/internet_identity/tests/integration/config/oidc_configs.rs
timothyaterton: Addressed in b6828827. `name` and `logo` are gone from `DiscoverableOidcConfig` and `OidcConfig`. The frontend will render SSO provider labels from the domain input (and from the hop-1 response if we decide to surface that there later), not from backend-stored metadata.

src/internet_identity/tests/integration/config/oidc_configs.rs
timothyaterton: Addressed in b6828827. `email_verification` removed from `DiscoverableOidcConfig` and `OidcConfig`. `DiscoverableProvider::email_verification_scheme()` now returns `None` unconditionally — email verification for SSO is an operator trust decision when the domain is configured, not an II-side per-provider heuristic like the existing Google/Microsoft schemes.

src/internet_identity/src/main.rs
```diff
@@ -657,14 +666,31 @@ fn apply_install_arg(maybe_arg: Option<InternetIdentityInit>) {
                 persistent_state.related_origins = Some(related_origins);
             })
         }
-        if let Some(new_flow_origins) = arg.new_flow_origins {
+        // oidc_configs, openid_configs, and new_flow_origins are mutually exclusive.
+        // When openid_configs is provided alongside oidc_configs, openid_configs takes
+        // precedence as the proven path. When oidc_configs is set, new_flow_origins is
+        // cleared (and vice versa) since oidc_configs subsumes new_flow_origins.
```
sea-snake: `new_flow_origins` is completely irrelevant to SSO, no idea why this was picked up as relevant. For context, this was used previously as a flag to toggle between new and older interface depending on the domain the user was visiting.

src/internet_identity/src/main.rs
```diff
@@ -657,14 +666,31 @@ fn apply_install_arg(maybe_arg: Option<InternetIdentityInit>) {
                 persistent_state.related_origins = Some(related_origins);
             })
         }
-        if let Some(new_flow_origins) = arg.new_flow_origins {
+        // oidc_configs, openid_configs, and new_flow_origins are mutually exclusive.
```
sea-snake: Also, a canister can have both oidc_configs and openid_configs, they are _not_ exclusive.

src/internet_identity/src/main.rs
```diff
@@ -568,6 +573,7 @@ fn config() -> InternetIdentityInit {
         related_origins: persistent_state.related_origins.clone(),
         new_flow_origins: persistent_state.new_flow_origins.clone(),
         openid_configs: persistent_state.openid_configs.clone(),
+        oidc_configs: persistent_state.oidc_configs.clone(),
```
sea-snake, line 581: Discoverable OpenID configs shouldn't be defined on canister deployment, instead they should be added at any moment in time with an update call. (So any SSO provider can be used without explicit need to update the canister, currently allow listed to only dfinity.com).

src/internet_identity/internet_identity.did
```diff
@@ -280,6 +280,8 @@ type InternetIdentityInit = record {
     new_flow_origins : opt vec text;
     // Configurations for OpenID clients
     openid_configs : opt vec OpenIdConfig;
+    // Simplified OIDC configs using discovery. Mutually exclusive with openid_configs.
+    oidc_configs : opt vec DiscoverableOidcConfig;
```
sea-snake, line 285: See other comment, this shouldn't be a config, instead rely on an update call.

src/internet_identity/src/main.rs
timothyaterton: Addressed in 2468699. Removed all XOR logic between `new_flow_origins` and `oidc_configs` — they are now completely independent. `new_flow_origins` is no longer touched when `oidc_configs` changes (and vice versa).

src/internet_identity/src/main.rs
timothyaterton: Addressed in 2468699. Removed the mutual exclusivity between `oidc_configs` and `openid_configs`. They now coexist — setting one no longer clears the other. Added a test (`should_coexist_with_openid_configs`) that verifies both can be present simultaneously.

src/internet_identity/src/main.rs
timothyaterton: Addressed in 2468699. Added `add_discoverable_oidc_config(DiscoverableOidcConfig) -> ()` update call so SSO providers can be added at any time without a canister upgrade. The `oidc_configs` field is kept in init args only for backward compatibility and `config()` query output — `apply_install_arg` no longer processes it. The update call validates against the canary allowlist, deduplicates, persists to state (survives upgrades), and starts discovery immediately.

src/internet_identity/internet_identity.did
timothyaterton: Addressed in 2468699. Added `add_discoverable_oidc_config : (DiscoverableOidcConfig) -> ()` to the Candid interface. The `oidc_configs` field in `InternetIdentityInit` is now documented as deprecated for init args (kept for backward compat and `config()` output only).

sea-snake (dismissed):

src/internet_identity/src/openid/generic.rs
```diff
@@ -215,6 +228,410 @@ impl Provider {
...
+
+        // Start or restart cert fetching when jwks_uri changes
+        let jwks_changed = {
+            let last = task.last_jwks_uri.borrow();
+            last.as_deref() != Some(&doc.jwks_uri)
+        };
+        if jwks_changed {
+            task.last_jwks_uri.replace(Some(doc.jwks_uri.clone()));
+            schedule_fetch_certs(doc.jwks_uri, Rc::clone(&task.certs_ref), Some(0));
+        }
```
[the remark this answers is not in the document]
aterga: Addressed in 29d0590. Added `validate_https_url` in `generic.rs` — before scheduling cert fetches in `run_discovery_tasks`, `doc.jwks_uri` is now parsed and required to use the `https` scheme. Malformed or non-HTTPS values are logged and the task is skipped instead of looping indefinitely on failing outcalls.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -202,7 +202,12 @@ fn create_openid_credential_and_config(
     let OpenIDRegFinishArg { jwt, salt, name: _ } = openid_registration_data;
 
     let (openid_credential, openid_config_iss) = openid::with_provider(jwt, |provider| {
-        Ok((provider.verify(jwt, salt)?, provider.issuer()))
+        // `with_provider` already skipped providers with pending discovery, so
+        // `provider.issuer()` is guaranteed to be Some here.
+        Ok((
+            provider.verify(jwt, salt)?,
+            provider.issuer().unwrap_or_default(),
```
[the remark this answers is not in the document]
aterga: Addressed in 29d0590. Replaced `provider.issuer().unwrap_or_default()` with an explicit `ok_or_else` that surfaces an `OpenIDJWTVerificationError::GenericError` if the invariant ever breaks. We no longer silently record an empty `openid_config_iss`.

src/internet_identity/src/openid/generic.rs
```diff
@@ -215,6 +228,410 @@ impl Provider {
...
+            .ok_or(OpenIDJWTVerificationError::GenericError(
+                "JWT is missing kid".to_string(),
+            ))?;
+        let certs = self.certs.borrow();
+        let cert = certs
+            .iter()
+            .find(|cert| cert.kid().is_some_and(|v| v == kid))
+            .ok_or(OpenIDJWTVerificationError::GenericError(format!(
+                "Certificate not found for {kid}"
+            )))?;
```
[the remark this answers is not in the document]
aterga: Addressed in 29d0590. `DiscoverableProvider::verify` now checks `certs.is_empty()` before the kid lookup and returns a dedicated `"OIDC provider JWKS not yet fetched, please try again shortly"` error in that window. A real kid mismatch still surfaces as "Certificate not found for {kid}", so the frontend can distinguish the retryable case from the fatal one.

src/internet_identity/src/openid.rs
```diff
@@ -251,10 +265,31 @@ pub trait OpenIdProvider {
...
+    Single(String),
+    Multiple(Vec<String>),
+}
+
+impl AudClaim {
+    fn as_str(&self) -> Option<&str> {
+        match self {
+            AudClaim::Single(s) => Some(s.as_str()),
+            AudClaim::Multiple(v) => v.first().map(String::as_str),
+        }
```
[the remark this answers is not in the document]
aterga: Addressed in 29d0590. `AudClaim` now exposes `matches(expected)` — `Single` does direct equality, `Multiple` returns true if any element equals `expected`. `with_provider` and `verify_claims` both switched to `aud.matches(client_id)`. JWTs whose `aud` array contains the configured `client_id` at any position are now accepted. Added unit tests `should_accept_client_id_in_aud_array` and `should_reject_aud_array_without_client_id`.

src/internet_identity/src/openid/generic.rs
```diff
@@ -215,6 +228,410 @@ impl Provider {
...
+            })?;
+        let claims: Claims = serde_json::from_slice(validation_item.claims()).map_err(|_| {
+            OpenIDJWTVerificationError::GenericError(
+                "Unable to decode claims or expected claims are missing".to_string(),
+            )
+        })?;
+
+        // Verify claims against discovered issuer
+        verify_claims(&issuer, &client_id, &claims, salt)?;
+
```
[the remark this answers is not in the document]
aterga: Addressed in 29d0590. `Claims.aud` is now `AudClaim` (shared with `openid.rs`) and therefore parses both string and array forms. `verify_claims` uses `claims.aud.matches(client_id)`. Credentials now record the canonical `client_id` value rather than the raw claim, which is what we actually verified against.

sea-snake (approved):

PR #3784, opened 2026-04-16
feat(be): OpenIdCredentialKey with data migration — add aud for SSO security

timothyaterton:
## Summary

Add the `aud` (audience / client_id) field to `OpenIdCredentialKey`, changing it from `(iss, sub)` to `(iss, sub, aud)`. This is a security prerequisite for SSO: since SSO allows anyone to provide a `client_id` via their `ii-openid-configuration` endpoint, without `aud` in the key two different OIDC clients at the same provider with the same user `sub` would collide, enabling impersonation.

## Changes

- **Type update**: `OpenIdCredentialKey` type alias changed from `(Iss, Sub)` to `(Iss, Sub, Aud)` in both `internet_identity_interface` and the `openid` module
- **CBOR encoding**: `StorableOpenIdCredentialKey` rewritten with manual `Encode`/`Decode` impls — new entries use CBOR map format `{0:iss, 1:sub, 2:aud}`; the decoder also handles legacy CBOR array format `[iss, sub]` for backward compatibility
- **Migration**: `post_upgrade` drains the credential key index via `pop_first`, resolves `aud` from each anchor's `StorableOpenIdCredential` (which already stores `aud` at CBOR index `#[n(2)]`), and re-inserts with the complete `(iss, sub, aud)` key. Unresolvable entries are preserved with empty `aud` for retry on next upgrade.
- **Key construction**: Updated `OpenIdCredential::key()`, `StorableOpenIdCredential::key()`, `calculate_delegation_seed()`, and all call sites
- **Candid interface**: Updated `.did` file and generated JS/TS declarations
- **Frontend**: Updated credential removal call to pass `aud`
- **Tests**: Added unit tests for new CBOR map encoding, legacy array decoding, and round-trip serialization. Updated existing test assertions to use 3-tuple keys.

## Delegation seed backward compatibility

The `calculate_delegation_seed` function already receives `client_id` (which equals `aud`) as a separate parameter. The seed calculation is unchanged — `aud` from the key tuple is ignored (`_aud`) in the destructuring, preserving identical `Principal` derivation for existing credentials.

## Migration safety

- Uses `pop_first()` to drain the BTreeMap, avoiding byte-level encoding mismatches between legacy array-encoded keys and new map-encoded keys
- Resolves `aud` from the anchor's stored `StorableOpenIdCredential` which already has `aud` at CBOR index 2
- Falls back to re-inserting with empty `aud` if resolution fails, with a logged warning — the entry is preserved for retry on next upgrade
- Idempotent: safe to run on every upgrade; entries already in the new format are preserved unchanged

## Test plan

- [x] All 209 unit tests pass (including Candid interface compatibility)
- [ ] Integration tests (require canister WASM build — pass in CI)
- [ ] Deploy to testnet and verify migration of existing credentials
- [ ] Verify credential lookup works after migration
- [ ] Verify new credential registration includes `aud` in key

---
[< Previous PR](https://github.com/dfinity/internet-identity/pull/3778) | [Next PR >](https://github.com/dfinity/internet-identity/pull/3785)

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/utils.ts
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/utils.ts
@@ -55,7 +55,8 @@ export const toAccessMethods = (
 
 /**
  * Returns a unique string key for an access method.
- * Uses credential ID for passkeys and issuer+subject for OpenID.
+ * Uses credential ID for passkeys and a JSON-encoded (iss, sub, aud) triple
+ * for OpenID so components can't collide via concatenation.
  */
 export const toKey = (accessMethod: AccessMethod): string => {
   if (
@@ -67,11 +68,11 @@ export const toKey = (accessMethod: AccessMethod): string => {
     );
   }
   if ("openid" in accessMethod) {
-    return (
-      accessMethod.openid.iss +
-      accessMethod.openid.sub +
-      accessMethod.openid.aud
-    );
+    return JSON.stringify([
+      accessMethod.openid.iss,
+      accessMethod.openid.sub,
+      accessMethod.openid.aud,
+    ]);
   }
   throw new Error("Unknown access method type");
 };
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/utils.ts
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/utils.ts
@@ -67,7 +67,7 @@ export const toKey = (accessMethod: AccessMethod): string => {
     );
   }
   if ("openid" in accessMethod) {
-    return accessMethod.openid.iss + accessMethod.openid.sub;
+    return accessMethod.openid.iss + accessMethod.openid.sub + accessMethod.openid.aud;
   }
   throw new Error("Unknown access method type");
 };
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/utils.ts
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/utils.ts
@@ -67,7 +67,11 @@ export const toKey = (accessMethod: AccessMethod): string => {
     );
   }
   if ("openid" in accessMethod) {
-    return accessMethod.openid.iss + accessMethod.openid.sub;
+    return (
+      accessMethod.openid.iss +
+      accessMethod.openid.sub +
+      accessMethod.openid.aud
+    );
   }
   throw new Error("Unknown access method type");
 };
--- a/src/internet_identity/src/openid.rs
+++ b/src/internet_identity/src/openid.rs
@@ -12,7 +12,7 @@ use internet_identity_interface::internet_identity::types::openid::{
     OpenIdCredentialAddError, OpenIdDelegationError,
 };
 use internet_identity_interface::internet_identity::types::{
-    AnchorNumber, Delegation, DiscoveredOidcConfig, IdRegFinishError, MetadataEntryV2, OidcConfig,
+    AnchorNumber, Delegation, DiscoverableOidcConfig, IdRegFinishError, MetadataEntryV2, OidcConfig,
     OpenIdConfig, OpenIdEmailVerificationScheme, PublicKey, SessionKey, SignedDelegation, Timestamp,
     UserKey,
 };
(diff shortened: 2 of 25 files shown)


src/internet_identity/src/openid.rs
```diff
@@ -485,12 +485,12 @@ pub fn replace_issuer_placeholders(template: &str, claims: &Vec<(String, String)
 /// # Arguments
 ///
 /// * `client_id`: The client id for which the `OpenIdCredential` was created
-/// * `(iss, sub)`: The key of the `OpenIdCredential` to create a `Hash` from
+/// * `(iss, sub, _aud)`: The key of the `OpenIdCredential` to create a `Hash` from
 /// * `anchor_number`: The anchor number the credential is assigned to
 #[allow(clippy::cast_possible_truncation)]
 fn calculate_delegation_seed(
```
sea-snake, line 494: This seed should also be based on aud too!

src/internet_identity/src/storage/storable/openid_credential_key.rs
```diff
@@ -1,14 +1,89 @@
...
+                Ok(StorableOpenIdCredentialKey {
+                    iss: iss.ok_or_else(|| minicbor::decode::Error::message("missing iss"))?,
+                    sub: sub.ok_or_else(|| minicbor::decode::Error::message("missing sub"))?,
+                    aud: aud.unwrap_or_default(),
+                })
+            }
+            other => Err(minicbor::decode::Error::type_mismatch(other)),
+        }
+    }
+}
```
sea-snake: We're using minicbor_derive which should also support something along these lines without the manual implementation.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/utils.ts
```diff
@@ -67,7 +67,11 @@ export const toKey = (accessMethod: AccessMethod): string => {
     );
   }
   if ("openid" in accessMethod) {
-    return accessMethod.openid.iss + accessMethod.openid.sub;
+    return (
+      accessMethod.openid.iss +
+      accessMethod.openid.sub +
+      accessMethod.openid.aud
+    );
```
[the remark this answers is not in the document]
aterga: Addressed in 4024338. `toKey` for OpenID access methods now returns `JSON.stringify([iss, sub, aud])` — a collision-safe encoding where component boundaries are explicitly represented (string escaping + brackets). No two distinct triples can produce the same string.

src/internet_identity/src/openid.rs
aterga: Addressed in 4024338. `calculate_delegation_seed` now takes `aud` directly from the key tuple — the redundant `client_id` parameter is gone. All prior callers passed `self.aud` as `client_id`, so the seed bytes are bitwise-identical to before and existing Principals are preserved. The dependency of the seed on `aud` is now explicit in the signature, and documented as the property that makes SSO safe (same `(iss, sub)` at different OIDC clients now derives distinct principals).

src/internet_identity/src/storage/storable/openid_credential_key.rs
```diff
@@ -1,14 +1,89 @@
...
+                    }
+                }
+                Ok(StorableOpenIdCredentialKey {
+                    iss: iss.ok_or_else(|| minicbor::decode::Error::message("missing iss"))?,
+                    sub: sub.ok_or_else(|| minicbor::decode::Error::message("missing sub"))?,
+                    aud: aud.unwrap_or_default(),
+                })
+            }
+            other => Err(minicbor::decode::Error::type_mismatch(other)),
+        }
```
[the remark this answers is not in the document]
aterga: Addressed in 4024338. Dropped `MapIndef` from the match — we only ever emit definite-length maps (`e.map(3)`), so indefinite-form input is a bug anyway. `d.map()?.ok_or_else(...)` now returns an explicit `"indefinite-length map is not supported"` error instead of silently yielding an empty loop body.

src/internet_identity/src/storage/storable/openid_credential_key.rs
aterga: Addressed in 2d07f80. Refactored to use `#[derive(Encode, Decode)]` with `#[cbor(map)]` — the manual encode/decode impls are gone. Legacy decode falls back to a separate `LegacyStorableOpenIdCredentialKey` with `#[cbor(array)]` via a top-level CBOR-type peek in `from_bytes`. ~70 lines of hand-written map parsing removed.

sea-snake (approved):

PR #3785, opened 2026-04-16
feat(be,fe): SSO sign-in via two-hop OIDC discovery

timothyaterton:
# Problem

Internet Identity users can sign in with personal Google/Apple/Microsoft accounts, but there's no way for an organization to let their employees sign in with *their own* SSO domain (e.g. `alice@dfinity.org`) while still having II act as the identity provider. Each new organization would otherwise require a code change on II to add their issuer, client id, and OAuth endpoints.

# Solution

A user types their organization's domain on the SSO screen; the frontend calls `add_discoverable_oidc_config` to register the domain (gated by the backend's canary allowlist), then runs a two-hop discovery chain to resolve the provider's OAuth endpoint, then redirects them to sign in.

Registration is **user-initiated**: the SSO screen itself drives the canister update call. An II admin still has to land a new domain in the backend's canary allowlist (`openid::generic::allowed_discovery_domains()`) via canister upgrade — what's new is that, once that's done, individual users can register their own org's domain via the SSO screen instead of the config living inline in II's init args.

The allowlist is gated on the deployment's `is_production` init flag so the two mainnet canisters don't share a domain: on `id.ai` only `dfinity.org` is accepted, on `beta.id.ai` (and everywhere else — staging, local, CI) only `beta.dfinity.org`. Keeping them disjoint means a DNS takeover of the beta test domain can't backdoor production, and we can stage new IdP wiring on `beta.dfinity.org` without risking the prod issuer.

Replaces #3786, which is closed.

# Changes

### Removed: `oidc_configs` from init args and synchronized config

`InternetIdentityInit` previously carried `oidc_configs`, but applying it had never been wired through `apply_install_arg` — it was silently dropped on install/upgrade. Removed the field entirely from `InternetIdentityInit`, `InternetIdentitySynchronizedConfig`, the Candid interface, `config()`, and the `From<&InternetIdentityInit> for InternetIdentitySynchronizedConfig` impl. Registration goes exclusively through the `add_discoverable_oidc_config` update call from here on, and `/.config.did.bin` only carries `openid_configs` (the direct Google/Apple/Microsoft configs).

### `add_discoverable_oidc_config`

Traps for any domain not in `ALLOWED_DISCOVERY_DOMAINS`, otherwise inserts the domain into persistent state and kicks off a backend-side two-hop discovery (to populate JWKS for signature verification on subsequent sign-ins). The canister also exposes `discovered_oidc_configs` for querying resolved SSO provider state.

## Frontend

**Type alignment with backend.** `DiscoverableOidcConfig` is `{ discovery_domain: string }`.

**Two-hop discovery (`ssoDiscovery.ts`, new).**
1. `GET https://{domain}/.well-known/ii-openid-configuration` returns `{ client_id, openid_configuration }`. The domain owner publishes this at their DNS-backed origin.
2. `GET {openid_configuration}` is the provider's standard OIDC discovery, yielding `authorization_endpoint` and `scopes_supported`.

Both hops run from the browser. (The backend has its own copy in `openid/generic.rs`; keeping the implementations separate for now minimizes BE↔FE synchronization.)

**SSO flow UI.**
- `SignInWithSso.svelte` (new): domain input screen. Framed icon chip, title `"Sign In With SSO"`, subtitle `"Enter your company domain"`, placeholder `"company.domain.com"`. Input triggers a debounced (200ms) lookup: `anonymousActor.add_discoverable_oidc_config` (idempotent; traps if the domain isn't on the backend allowlist) → `discoverSsoConfig`. The Continue button enables only once the lookup succeeds; clicking it opens the OAuth popup synchronously from the user gesture (critical for Safari). Canary-allowlist traps are mapped to `"SSO is not available for \"<domain>\" yet."` inline.
- `SsoIcon.svelte` (new): key icon.
- `PickAuthenticationMethod.svelte`: "Continue with passkey" is the top button; OIDC provider icons + SSO key icon render in the row below at equal width. **The SSO icon is always rendered** — users always see SSO as an option, regardless of whether any domain is registered.
- `authFlow.svelte.ts`: new `signInWithSso` view state and `continueWithSso()` method.
- Subtitle text across all three sign-in entry points updated from `"Choose method to continue"` → `"Choose an authentication method to continue"`.

**SSO vs direct-provider credential disambiguation.**

An SSO credential can arrive via the same underlying IdP (e.g. Google) as a direct-Google credential. They differ in `aud` (the `client_id`), but two things broke simple issuer-only matching:
1. Labeling an SSO-via-Google credential as "Google account" (it's not — it's the user's org account).
2. Mis-labeling a legacy direct-Google credential as SSO after `client_id` rotation.

Resolved via three-tier lookup: strict `(iss, aud)` match → localStorage SSO-domain map (`{iss, sub, aud} → domain`) → issuer-only fallback. Credentials linked via SSO remember the typed domain in localStorage so they can be surfaced by domain in the access-methods UI on the same device. Cross-device labeling needs backend support and is tracked in #3795.

**Safari popup.** The OAuth popup must open synchronously from the click event or Safari blocks it. The SSO lookup (`add_discoverable_oidc_config` + two-hop) is done in a debounced input handler, stashed in local state, and consumed by the click handler with no intervening awaits before `window.open`.

**Already-linked disambiguation.** If the backend returns `OpenIdCredentialAlreadyRegistered`, the FE queries `get_anchor_info` and distinguishes "already linked to **this** identity" (specialized `OpenIdCredentialAlreadyLinkedHereError`) from "already linked to another identity" (the generic error).

**Cleanup.** `oidc_configs` removed from the frontend's `BackendCanisterConfig` decode schema (no remaining FE consumer after the refactor). Candid is forward-compatible, so the backend continues to accept old init-arg shapes that may have included `oidc_configs`.

**Security.**
- Domain input DNS-format validated (length, label length, forbidden characters).
- Backend canary allowlist (`allowed_discovery_domains()`, gated on `is_production`) is the sole source of truth for which domains can register. The frontend does **not** carry its own copy — the gate lives on the canister where a compromised device can't bypass it.
- All three URLs (`/.well-known/ii-openid-configuration`, provider discovery, authorization endpoint) must be HTTPS.
- Issuer and `authorization_endpoint` hostnames must match the `openid_configuration` hostname *exactly* or as a true subdomain (prevents a tampered provider-discovery doc from bouncing auth off-host after we've committed to a provider). `endsWith` alone would accept look-alikes like `evildfinity.okta.com`.
- Second-hop trust is inherited from the first-hop canary allowlist: an attacker who can tamper with an II-approved domain's `.well-known` has already breached something more fundamental than any FE-side allowlist would catch, and the org knows its own IdP better than II does.
- Per-domain rate limit (1 attempt per 10 min), max 2 concurrent discoveries, 4-hour cache per hop, exponential backoff, timeouts with `clearTimeout` in `finally`.

# Tests

- `config/oidc_configs.rs` covers both branches of the `is_production`-gated allowlist: default install rejects `dfinity.org` (and accepts `beta.dfinity.org`), `is_production: Some(true)` install rejects `beta.dfinity.org` (and accepts `dfinity.org`). `should_coexist_with_openid_configs` now queries SSO state via `discovered_oidc_configs` instead of the removed `config().oidc_configs`.
- 23 unit tests in `ssoDiscovery.test.ts`.
- 22 unit tests in `openID.test.ts`, including 4 for `selectAuthScopes` and the `(iss, aud)` strict-then-fallback resolution for direct-vs-SSO credentials.
- `cargo test -p internet_identity --bin internet_identity`: **219/219 pass**.
- `cargo clippy --all-targets -D warnings`: clean.
- `npm run lint` + `svelte-check`: clean.

---
[< Previous PR](https://github.com/dfinity/internet-identity/pull/3784)


src/frontend/src/lib/utils/oidcDiscovery.ts
```diff
@@ -0,0 +1,230 @@
...
+      const response = await fetch(url, { signal: controller.signal });
+      clearTimeout(timeoutId);
+
+      if (!response.ok) {
+        throw new Error(
+          `Discovery fetch failed: ${response.status} ${response.statusText}`,
+        );
+      }
+      const data: unknown = await response.json();
+      return validateDocument(data, expectedHostname);
```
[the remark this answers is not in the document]
timothyaterton: Fixed in 0ebd6e2e — moved `clearTimeout(timeoutId)` into a `finally` on the per-attempt try block so the timer is always cleared regardless of whether `fetch`, `response.json()`, or the validation step throws.

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -238,6 +240,62 @@ export class AuthFlow {
...
+    // otherwise default to "openid profile email"
+    const authScope =
+      discovery.scopes_supported !== undefined &&
+      discovery.scopes_supported.length > 0
+        ? discovery.scopes_supported
+            .filter((s) =>
+              ["openid", "profile", "email"].includes(s),
+            )
+            .join(" ")
+        : "openid profile email";
```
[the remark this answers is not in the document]
timothyaterton: Fixed in 0ebd6e2e — extracted `selectAuthScopes(scopes_supported)` in `openID.ts` which falls back to `["openid", "profile", "email"]` when the filtered intersection is empty (or when `scopes_supported` is missing/empty), and reused it here instead of the inline `filter(...).join(" ")`.

src/frontend/src/lib/flows/authFlow.svelte.ts
```diff
@@ -238,6 +240,62 @@ export class AuthFlow {
...
+
+    // Build RequestConfig from static client_id + discovered endpoint
+    const requestConfig: RequestConfig = {
+      clientId,
+      authURL: discovery.authorization_endpoint,
+      authScope,
+    };
```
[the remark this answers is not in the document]
timothyaterton: Fixed in 0ebd6e2e — removed the unused `requestConfig`. `continueWithOpenId` re-derives its own `RequestConfig` from the synthetic `OpenIdConfig`, so the manual build was dead code. This was also what was failing ESLint (`max-warnings 0`) in CI.

src/frontend/src/lib/utils/openID.ts
```diff
@@ -267,10 +269,48 @@ export const issuerMatches = (
...
+    jwks_uri: "",
+    logo: oidcConfig.logo,
+    name: oidcConfig.name,
+    fedcm_uri: [],
+    email_verification: oidcConfig.email_verification,
+    issuer: cached.document.issuer,
+    auth_scope: cached.document.scopes_supported?.filter((s) =>
+      ["openid", "profile", "email"].includes(s),
+    ) ?? ["openid", "profile", "email"],
+    client_id: oidcConfig.client_id[0],
```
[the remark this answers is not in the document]
timothyaterton: Fixed in 0ebd6e2e — now uses `selectAuthScopes(cached.document.scopes_supported)`, which falls back to `["openid", "profile", "email"]` when the intersection is empty. Tests cover all four branches (undefined, non-empty intersection, empty after filter, empty input).

src/frontend/src/lib/utils/oidcDiscovery.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  // Validate issuer matches expected provider domain
+  const issuerUrl = new URL(doc.issuer);
+  if (issuerUrl.protocol !== "https:") {
+    throw new Error(`Issuer must use HTTPS: ${doc.issuer}`);
+  }
+  if (!issuerUrl.hostname.endsWith(expectedHostname)) {
+    throw new Error(
+      `Issuer hostname mismatch: expected *${expectedHostname}, got ${issuerUrl.hostname}`,
```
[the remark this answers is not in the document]
timothyaterton: Fixed in 0ebd6e2e — added a `hostnameMatchesAllowed(host, expected)` helper that requires either `host === expected` or `host.endsWith("." + expected)`. Both issuer and `authorization_endpoint` hostnames now go through it, and there's a regression test (`rejects look-alike issuer hostname (no leading dot separator)`) that asserts `evilaccounts.google.com` is rejected against expected `accounts.google.com`.

src/frontend/src/lib/utils/oidcDiscovery.ts
```diff
@@ -0,0 +1,230 @@
...
+    );
+  }
+
+  // Validate authorization_endpoint uses HTTPS
+  const authUrl = new URL(doc.authorization_endpoint);
+  if (authUrl.protocol !== "https:") {
+    throw new Error(
+      `Authorization endpoint must use HTTPS: ${doc.authorization_endpoint}`,
+    );
+  }
```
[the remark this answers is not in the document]
timothyaterton: Fixed in 0ebd6e2e — `authorization_endpoint` hostname is now also checked against the same trusted provider domain via `hostnameMatchesAllowed`, not just for HTTPS. A new test `rejects authorization_endpoint on a different host` guards against a tampered discovery response pointing to an off-host auth URL.

src/frontend/src/lib/utils/oidcDiscovery.ts
```diff
@@ -0,0 +1,242 @@
...
+      // Don't retry abort errors or validation errors
+      if (
+        error instanceof Error &&
+        (error.name === "AbortError" ||
+          error.message.startsWith("Discovery document") ||
+          error.message.startsWith("Issuer") ||
+          error.message.startsWith("Authorization endpoint"))
+      ) {
+        throw error;
+      }
```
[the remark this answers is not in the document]
timothyaterton: This targets `oidcDiscovery.ts`, which was deleted in 3d587a18 as part of consolidating #3785 + #3786 into one SSO flow. The replacement `ssoDiscovery.ts` doesn't have this pattern: validation (including `new URL(...)` parsing, which throws `TypeError`) runs **after** `fetchWithRetry` returns, outside the retry loop. The only retry-abort condition in the new `fetchWithRetry` is `error.name === 'AbortError'` — no brittle string matching, and `TypeError` from URL parsing propagates directly without retry because it happens after the fetch succeeds.

Resolving as obsolete.

src/frontend/src/lib/utils/openID.ts
```diff
@@ -267,10 +269,61 @@ export const issuerMatches = (
...
+  // Fall back to oidc_configs using cached discovery data
+  const oidcConfigs = backendCanisterConfig.oidc_configs?.[0];
+  if (oidcConfigs === undefined) {
+    return undefined;
+  }
+
+  const cached = findCachedDiscoveryByIssuer(issuer);
+  if (cached === undefined) {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
timothyaterton: The cache-based `oidc_configs` extension in `findConfig` was removed in 3d587a18. `findConfig` now only looks in `openid_configs` (pre-configured providers with static `name`/`logo`/`issuer`); SSO identities synthesize their `OpenIdConfig` at sign-in time from two-hop discovery results and never hit `findConfig`, so the "cache empty after reload" failure mode doesn't apply to the current code.

Your deeper point is still valid though: after a page reload, a last-used SSO identity will not render a "continue as" shortcut because we've lost the `(issuer → discovery_domain)` mapping. Fixing that properly needs the `discovery_domain` to be persisted on the `LastUsedIdentity` entry so we can re-run discovery on reload — a larger change that touches the last-used-identity store and the "continue as" rendering code. That's follow-up work tracked for a subsequent PR; the SSO entry point in this PR always starts fresh from the domain-input screen, which is the current intended UX.

Resolving this thread since the code you cited is gone; will open a follow-up issue for the reload path.

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
```diff
@@ -69,6 +91,32 @@
...
+        {@const providerId = provider.client_id[0] ?? provider.discovery_url}
+        <Tooltip
+          label={$t`Interaction canceled. Please try again.`}
+          hidden={cancelledProviderId !== providerId}
+          manual
+        >
+          <Button
+            onclick={() => handleContinueWithOidc(provider)}
+            variant="secondary"
+            disabled={authenticatingProviderId !== undefined}
```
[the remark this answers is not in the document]
timothyaterton: Per-provider OIDC buttons were removed in 3d587a18 — under the consolidated design there's a single "Sign in with SSO" button that leads to a domain-input screen, not one button per `oidc_configs` entry. The new `DiscoverableOidcConfig` type doesn't even have a `client_id` field anymore (just `discovery_domain`); `client_id` is resolved at sign-in time via the two-hop discovery in `ssoDiscovery.ts`. So the "button that always errors on click" failure mode can't occur.

Resolving as obsolete.

src/frontend/src/lib/utils/oidcDiscovery.ts
```diff
@@ -0,0 +1,242 @@
...
+ */
+export const findCachedDiscoveryByIssuer = (
+  issuer: string,
+): { discoveryUrl: string; document: OidcDiscoveryDocument } | undefined => {
+  for (const [discoveryUrl, entry] of cache) {
+    if (entry.document.issuer === issuer) {
+      return { discoveryUrl, document: entry.document };
+    }
+  }
+  return undefined;
```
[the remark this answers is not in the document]
timothyaterton: `findCachedDiscoveryByIssuer` was deleted in 3d587a18 along with the rest of `oidcDiscovery.ts` — it had no remaining callers after the `findConfig` extension was removed. The TTL-consistency concern is moot because the function no longer exists.

Resolving as obsolete.

src/frontend/src/lib/utils/openID.ts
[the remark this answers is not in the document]
aterga: Opened #3795 to track the reload-path fix (persist `discovery_domain` on `LastUsedIdentity` so "continue as" shortcuts can re-run discovery without the user re-typing their domain).

src/frontend/src/lib/utils/openID.test.ts
```diff
@@ -1,15 +1,17 @@
...
   extractIssuerTemplateClaims,
+  selectAuthScopes,
 } from "./openID";
 import { OpenIdConfig } from "$lib/generated/internet_identity_types";
 import { backendCanisterConfig } from "$lib/globals";
 
 vi.mock("$lib/globals", () => ({
   backendCanisterConfig: {
     openid_configs: [],
+    oidc_configs: [],
```
sea-snake: Nit: this should have been removed

src/frontend/src/lib/utils/openID.ts
```diff
@@ -272,6 +273,21 @@ export const findConfig = (
     issuerMatches(config.issuer, issuer, metadata),
   );
 
+/**
+ * Pick the subset of OIDC scopes we actually request from what the provider
+ * advertises. If nothing advertised (or the intersection is empty after
+ * filtering), fall back to the canonical defaults — sending an empty `scope`
+ * to the authorize endpoint would cause the request to fail.
+ */
+export const selectAuthScopes = (scopesSupported?: string[]): string[] => {
```
sea-snake, line 387: The openid scope is a requirement, nothing works without at least the openid scope.

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -0,0 +1,476 @@
...
+ * `ii-openid-configuration` is tampered with, we refuse to redirect users
+ * into an auth flow on a host we've never heard of. Entries must be either
+ * mainstream OIDC providers or known SaaS identity platforms.
+ */
+const TRUSTED_PROVIDER_DOMAINS = new Set([
+  "accounts.google.com",
+  "appleid.apple.com",
+  "login.microsoftonline.com",
+  "dfinity.okta.com",
+  "login.dfinity.org",
```
sea-snake: Only `dfinity.okta.com` is correct here, others are either not SSO (but just registered openid) or hallucinations (login.dfinity.org).

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -0,0 +1,476 @@
...
+    }
+  }
+  if (!trusted) {
+    throw new Error(`Provider domain not trusted: ${parsed.hostname}`);
+  }
+  return parsed;
+};
+
+/** Validate the `ii-openid-configuration` response structure. */
+const validateIIConfig = (data: unknown): IIOpenIdConfiguration => {
```
sea-snake, line 203: Nit: Please use zod for validation of unknown data.

src/frontend/src/lib/globals.ts
```diff
@@ -29,19 +35,20 @@ const backendCanisterConfigIDL = IDL.Record({
...
       }),
     ),
   ),
+  // `oidc_configs` is deliberately omitted from the frontend decode — the
+  // backend still serves it in `InternetIdentitySynchronizedConfig` for any
+  // future consumer, but the current SSO flow is user-initiated: the SSO
+  // screen calls `add_discoverable_oidc_config` directly on submit and
+  // relies on the backend's canary allowlist as the gate, not a pre-baked
+  // FE-visible list. Candid is forward-compatible, so the extra wire field
+  // is ignored here.
```
sea-snake: > the extra wire field is ignored here

There shouldn't be a wire field like oidc_config to begin with, this is probably a hallucination based on earlier implementation.

This comment can be simply removed.

src/frontend/src/routes/(new-styling)/(channel)/authorize/+layout.svelte
```diff
@@ -240,7 +240,7 @@
               {$t`Sign in`}
             </h1>
             <p class="text-text-secondary mb-6 self-start text-sm">
-              {$t`Choose method to continue`}
+              {$t`Choose an authentication method to continue`}
```
sea-snake: This wording change is out of scope, please revert it.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -549,7 +549,7 @@
         {$t`Sign in`}
       </h1>
       <p class="text-text-secondary mb-6 self-start text-sm">
-        {$t`Choose method to continue`}
+        {$t`Choose an authentication method to continue`}
```
sea-snake: Ditto, this wording change is out of scope, please revert it.

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -646,7 +646,7 @@
         {$t`Sign in`}
       </h1>
       <p class="text-text-secondary mb-6 self-start text-sm">
-        {$t`Choose method to continue`}
+        {$t`Choose an authentication method to continue`}
```
sea-snake: Ditto, this wording change is out of scope, please revert it.

src/internet_identity/src/assets.rs
```diff
@@ -21,10 +21,17 @@ pub fn init_assets(config: &InternetIdentityInit) {
...
 pub fn get_static_assets(config: &InternetIdentityInit) -> Vec<Asset> {
     // Required to make the synchronized config available to the II frontend canister.
-    // The synchronized config is the part of `InternetIdentityInit` that is needed for both the II backend and frontend.
+    // The synchronized config is the subset of state the frontend needs on page
+    // load. `openid_configs` comes from init args (they are set that way);
+    // `oidc_configs` is not in init args at all — SSO providers are managed via
+    // `add_discoverable_oidc_config` and live in persistent state.
+    let synchronized = InternetIdentitySynchronizedConfig {
+        openid_configs: config.openid_configs.clone(),
+        oidc_configs: state::persistent_state(|s| s.oidc_configs.clone()),
```
sea-snake: Frontend doesn't need oidc configs, so don't send them.

src/frontend/src/lib/utils/openID.test.ts
aterga: Fixed in 379b5e2d — removed `oidc_configs: []` from the `backendCanisterConfig` mock in `openID.test.ts` (it was stale after `oidc_configs` was dropped from `BackendCanisterConfig`).

src/frontend/src/lib/utils/openID.ts
aterga: Fixed in 379b5e2d — `selectAuthScopes` now always prepends `"openid"` to the result, regardless of what the provider lists in `scopes_supported`. Test coverage updated to assert the guarantee even when the provider's `scopes_supported` omits `openid`.

src/frontend/src/lib/utils/ssoDiscovery.ts
aterga: Fixed in 379b5e2d — trimmed `TRUSTED_PROVIDER_DOMAINS` to `{ "dfinity.okta.com" }`. Google/Apple/Microsoft entries were conflating `openid_configs` with SSO, and `login.dfinity.org` was a hallucination; dropped all four.

src/frontend/src/lib/utils/ssoDiscovery.ts
aterga: Fixed in 379b5e2d — both validators (`validateIIConfig` and `validateProviderDiscovery`) now use zod schemas (`IIOpenIdConfigurationSchema`, `OidcDiscoveryDocumentSchema`). The exported TS types are `z.infer`'d from the schemas so there's a single source of truth. A small `parseOrThrow` helper surfaces field-level error paths.

src/frontend/src/lib/globals.ts
aterga: Fixed in 379b5e2d — dropped both the IDL entry and the explanatory comment block. Paired with thread #9's change: the backend no longer serves `oidc_configs` in `InternetIdentitySynchronizedConfig` at all, so the "forward-compatibility" rationale is moot.

src/frontend/src/routes/(new-styling)/(channel)/authorize/+layout.svelte
aterga: Reverted in 379b5e2d — restored `"Choose method to continue"` in `authorize/+layout.svelte`.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
aterga: Reverted in 379b5e2d — restored `"Choose method to continue"` in `manage/(authenticated)/+layout.svelte`.

src/frontend/src/routes/(new-styling)/+page.svelte
aterga: Reverted in 379b5e2d — restored `"Choose method to continue"` in `(new-styling)/+page.svelte`. Locale catalogs re-extracted via `npm run extract` so the corresponding msgids are gone from all 11 `.po` files.

src/internet_identity/src/assets.rs
aterga: Fixed in 379b5e2d — removed `oidc_configs` from `InternetIdentitySynchronizedConfig` entirely (both the Rust struct and the Candid type). `get_static_assets` no longer reads from `persistent_state.oidc_configs` for the certified asset. The SSO flow calls `add_discoverable_oidc_config` directly on user submit, so there's no remaining need to ship registered domains over the wire.

src/frontend/src/lib/utils/openID.test.ts
aterga: Done in 379b5e2d.

src/frontend/src/lib/utils/openID.ts
aterga: Right, fixed in 379b5e2d. `selectAuthScopes` now always includes `openid` in the returned list regardless of whether the provider advertises it in `scopes_supported` — providers that omit it from their discovery doc (against spec) still work. Tests updated to match.

src/frontend/src/lib/utils/ssoDiscovery.ts
aterga: Fixed in 379b5e2d — `TRUSTED_PROVIDER_DOMAINS` trimmed to just `dfinity.okta.com`. Google/Apple/Microsoft are served via `openid_configs` (the non-SSO path) and don't belong in the second-hop allowlist; `login.dfinity.org` was a hallucination.

src/frontend/src/lib/utils/ssoDiscovery.ts
aterga: Done in 379b5e2d. Wire responses are now validated via zod schemas (`IIOpenIdConfigurationSchema`, `OidcDiscoveryDocumentSchema`) — the exported TS types are derived from the schemas via `z.infer<...>` so any consumer of the module can't skip validation. Scope-filtering on `scopes_supported` got simpler too since zod already guarantees `string[] | undefined`.

src/frontend/src/lib/globals.ts
aterga: Right, the whole premise of that comment was wrong — there shouldn't be an `oidc_configs` wire field at all. Fixed in 379b5e2d by removing `oidc_configs` from `InternetIdentitySynchronizedConfig` entirely (Rust struct + Candid type + frontend decode). The comment is gone along with the field.

src/frontend/src/routes/(new-styling)/(channel)/authorize/+layout.svelte
aterga: Reverted in 379b5e2d.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
aterga: Reverted in 379b5e2d.

src/frontend/src/routes/(new-styling)/+page.svelte
aterga: Reverted in 379b5e2d.

src/internet_identity/src/assets.rs
aterga: Fixed in 379b5e2d. `oidc_configs` removed from `InternetIdentitySynchronizedConfig` (both the Rust struct and the served `/.config.did.bin` blob). `get_static_assets` no longer reads from persistent state for it. Also dropped the now-unused `refresh_config_asset` hook in `openid::add_oidc_config` since the FE no longer has any asset state that depends on SSO registrations — runtime `add_discoverable_oidc_config` calls just persist.

src/frontend/src/lib/locales/id.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Pilih metode"
 
 msgid "Choose method to continue"
-msgstr "Pilih metode untuk melanjutkan"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/fr.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Choisissez une méthode"
 
 msgid "Choose method to continue"
-msgstr "Choisissez une méthode pour continuer"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/ur.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "طریقہ منتخب کریں"
 
 msgid "Choose method to continue"
-msgstr "جاری رکھنے کے لیے طریقہ منتخب کریں"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/uk.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Виберіть метод"
 
 msgid "Choose method to continue"
-msgstr "Виберіть метод для продовження"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/pl.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Wybierz metodę"
 
 msgid "Choose method to continue"
-msgstr "Wybierz metodę, aby kontynuować"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/nl.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Kies methode"
 
 msgid "Choose method to continue"
-msgstr "Kies methode om door te gaan"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/de.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Methode auswählen"
 
 msgid "Choose method to continue"
-msgstr "Methode wählen, um fortzufahren"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/it.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Scegli il metodo"
 
 msgid "Choose method to continue"
-msgstr "Scegli il metodo per continuare"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/es.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Elige un método"
 
 msgid "Choose method to continue"
-msgstr "Elige un método para continuar"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/locales/ru.po
```diff
@@ -164,7 +179,7 @@ msgid "Choose method"
 msgstr "Выберите способ"
 
 msgid "Choose method to continue"
-msgstr "Выберите способ входа"
+msgstr ""
```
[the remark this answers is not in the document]
aterga: Restored in commit d2b23e1d (`fix(fe): restore translations wiped by auto-extract; sync new msgids`). The auto-extract had emptied this msgstr when the SSO strings were added; reverted it to the `origin/main` translation so the Svelte UI does not fall back to English at runtime.

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
```diff
@@ -69,16 +84,23 @@
           </Button>
         </Tooltip>
       {/each}
+      <!--
+        SSO entry is always rendered. If no domains are registered on the
+        deployment, the SignInWithSso screen will reject every input with
+        "This domain is not registered as an OIDC provider." — we still
+        surface the option so users know the mechanism exists.
```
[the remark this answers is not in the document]
aterga: Good catch — the gating lives on the canister now (`ALLOWED_DISCOVERY_DOMAINS` trap inside `add_discoverable_oidc_config`), not in the `SignInWithSso` screen. Rewrote the comment in ec0d3304 to describe the actual behavior: unregistered domains surface as an error from inside `SignInWithSso`, and this entry is kept visible so users know the mechanism exists.

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -0,0 +1,478 @@
...
+ */
+const validateProviderUrl = (url: string): URL => {
+  const parsed = new URL(url);
+  if (parsed.protocol !== "https:") {
+    throw new Error(
+      `Provider URL must use HTTPS: ${parsed.protocol}//${parsed.hostname}`,
+    );
+  }
+  return parsed;
+};
```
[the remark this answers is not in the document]
aterga: The PR description and module header were updated together with commit 9b730f65 (`fix(fe): inherit SSO second-hop trust from first-hop canary allowlist`). `validateProviderUrl` now intentionally only enforces HTTPS because the first hop is already gated by `ALLOWED_DISCOVERY_DOMAINS` on the canister — an attacker who can tamper with an II-approved domain`s `.well-known/ii-openid-configuration` has breached something more fundamental than any second-hop FE allowlist would catch, and the org knows its own IdP better than we do. The file header was further clarified in ec0d3304.

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -0,0 +1,478 @@
...
+ *
+ * 1. Fetch `https://{domain}/.well-known/ii-openid-configuration`
+ *    → returns `{ client_id, openid_configuration }`
+ * 2. Fetch the standard OIDC discovery document from the `openid_configuration`
+ *    URL → returns `authorization_endpoint`, `scopes_supported`, etc.
+ *
+ * The caller is responsible for checking that `domain` is in the backend's
+ * `oidc_configs` list before calling `discoverSsoConfig`. Unknown domains
+ * should be rejected at the UI layer; this module does not carry its own
+ * domain allowlist.
```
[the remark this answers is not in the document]
aterga: Correct — the file header was stale. Rewrote it in ec0d3304 to describe the real flow: the caller must call `add_discoverable_oidc_config({ discovery_domain })` on the canister before invoking `discoverSsoConfig`; that call traps on `ALLOWED_DISCOVERY_DOMAINS`, so by the time this module runs the domain has already been blessed. The FE carries no allowlist of its own — it lives on the canister where the user`s device cant bypass it.

src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
```diff
@@ -0,0 +1,192 @@
...
+      {error}
+      aria-label={$t`Company domain`}
+    />
+    <Button
+      onclick={handleSubmit}
+      variant="primary"
+      size="lg"
+      type="submit"
+      disabled={domain.trim().length === 0 || isSubmitting}
+    >
```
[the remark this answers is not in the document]
aterga: Good catch — both `onclick` on the button and `onsubmit` on the form fired `handleSubmit` on click. Dropped the buttons `onclick` in ec0d3304; the forms `onsubmit={(e) => { e.preventDefault(); handleSubmit(); }}` already covers both Enter-in-input and button-click, and keeps the synchronous chain from the user gesture into `continueWithSso → requestJWT → requestWithPopup → window.open` that Safari needs for the popup to be trusted.

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -237,16 +237,6 @@ pub struct InternetIdentityFrontendArgs {
...
-}
-
-impl From<&InternetIdentityInit> for InternetIdentitySynchronizedConfig {
-    fn from(value: &InternetIdentityInit) -> Self {
-        Self {
-            openid_configs: value.openid_configs.clone(),
-            oidc_configs: value.oidc_configs.clone(),
-        }
-    }
 }
```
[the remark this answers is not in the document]
aterga: Outdated in the best sense — the FE stopped consuming `oidc_configs` from the synchronized config (per @sea-snakes thread 9 in the earlier review), so `oidc_configs` was dropped from `InternetIdentitySynchronizedConfig` entirely. The PR description was updated to match in one of the earlier pushes; the type now only carries `openid_configs`.

src/internet_identity/src/openid/generic.rs
```diff
@@ -229,10 +229,16 @@ impl Provider {
...
+/// CANARY: SSO via two-hop discovery is a proof-of-concept. Only these
+/// domains are accepted as a `discovery_domain` for now. Widen this list
+/// (or move it to canister config) once the feature exits canary.
+pub const ALLOWED_DISCOVERY_DOMAINS: &[&str] = &[
+    "dfinity.org",
+    // DO NOT MERGE: temporary entry for testing the two-hop flow end-to-end
+    // on staging-C from a test canister that hosts `/.well-known/ii-openid-
+    // configuration`. Must be removed before this PR lands on main.
+    "s4i6f-riaaa-aaaad-agnna-cai.icp0.io",
+];
```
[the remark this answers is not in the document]
aterga: Acknowledged — the `s4i6f-…icp0.io` entry is a staging/QA hook for the test SSO server we use to exercise the two-hop flow end-to-end. Its deliberately tagged `DO NOT MERGE` in-code and will be stripped in the pre-merge cleanup (the PR is still `BLOCKED` on review, so it wont slip in). Leaving this thread open as the pre-merge checklist item.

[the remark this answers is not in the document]
aterga: Resolved properly in 77ed8d4a (`feat(be): gate SSO allowlist on is_production (id.ai vs beta.id.ai)`). The staging `s4i6f-…icp0.io` entry is gone; the allowlist is now returned by `allowed_discovery_domains()`, which branches on the `is_production` init flag — `dfinity.org` on `id.ai`, `beta.dfinity.org` everywhere else (beta.id.ai, staging, local, CI). Integration tests cover both branches.

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
```diff
@@ -69,16 +84,25 @@
...
+        here — we keep this option visible so users know the mechanism
+        exists.
+      -->
+      <Button
+        onclick={signInWithSso}
+        variant="secondary"
+        disabled={authenticatingProviderId !== undefined}
+        size="xl"
+        class="flex-1"
+        aria-label={$t`Sign in with SSO`}
```
sea-snake, line 101: Nit: should be `Continue with SSO` since it's either sign-up or sign-in.

src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
```diff
@@ -0,0 +1,263 @@
...
+   * Result of the most recent successful discovery for the currently-typed
+   * `domain`. Non-undefined means the Continue button can invoke
+   * `continueWithSso` directly from the click handler — critical for
+   * Safari, which blocks `window.open` calls that follow an `await`.
+   */
+  let preparedResult = $state<SsoDiscoveryResult>();
+
+  let debounceTimer: ReturnType<typeof setTimeout> | undefined;
+
+  const mapSubmitError = (e: unknown, domainInput: string): string => {
```
sea-snake, line 49: Nit: this function is excessive, it should focus on the actual useful error messages for the user while just logging the raw error to the console for devs.

src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
```diff
@@ -47,9 +57,30 @@ export class AddAccessMethodFlow {
...
       this.#isSystemOverlayVisible = false;
-      await actor
-        .openid_credential_add(identityNumber, jwt, salt)
-        .then(throwCanisterError);
+      try {
+        await actor
+          .openid_credential_add(identityNumber, jwt, salt)
+          .then(throwCanisterError);
+      } catch (err) {
+        // Specialize the generic "already registered to another identity"
```
sea-snake, line 65: Nit: this isn't reachable for regular OpenID since an already linked provider button is disabled. For SSO it can similarly disable the continue button if it detects the SSO domain is already linked to the identity.

src/frontend/src/lib/utils/ssoDomainStorage.ts
```diff
@@ -0,0 +1,81 @@
+/**
```
sea-snake, line 1: This is a rather big unreliable FE approach hack for something that should just be fixed in the BE (expose SSO details for openid credentials linked through SSO).

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/components/OpenIdItem.svelte
```diff
@@ -16,9 +17,17 @@
...
+  );
   const email = $derived(getMetadataString(openid.metadata, "email"));
-  const logo = $derived(openIdLogo(openid.iss, openid.metadata));
+  const logo = $derived(
+    openIdLogo(openid.iss, openid.sub, openid.aud, openid.metadata),
+  );
+  // Credential didn't match any entry in `openid_configs` on (iss, aud).
+  // Treat it as SSO and render the generic SSO icon as a fallback.
+  const isSso = $derived(logo === undefined);
+  const displayName = $derived(name ?? $t`SSO`);
```
sea-snake, line 30: Nit: As discussed, if `/.well-known/ii-openid-configuration` contains a name entry we can render that name e.g. `DFINITY`, else fallback to maybe the domain e.g. `dfinity.org`.

sea-snake (approved):
Mostly nit comments, one non-nit comment that can be handled in a separate PR.

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
timothyaterton: Fixed in f8ff1de — changed `aria-label` from "Sign in with SSO" to "Continue with SSO" on both the auth screen (`PickAuthenticationMethod`) and the add-access-method screen (`AddAccessMethod`).

src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
timothyaterton: Simplified in f8ff1de — `mapSubmitError` now `console.error`s the raw error for devs and returns concise user-facing messages for the key cases (domain not configured, OAuth provider errors, canary-allowlist rejection, rate limiting). All the pattern-matching on internal error messages is gone; anything unexpected falls through to a generic "SSO sign-in failed" message.

src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
timothyaterton: Fixed in f8ff1de — `SignInWithSso` now accepts an optional `existingCredentials` prop (passed from `AddAccessMethodWizard`). After discovery succeeds, if any existing credential matches `(iss, aud)` from the discovery result, the continue button stays disabled and the input shows "This SSO domain is already linked to your identity." The auth-flow (initial sign-in) doesn't pass credentials, so the check only applies in the add-access-method context.

src/frontend/src/lib/utils/ssoDomainStorage.ts
timothyaterton: Agreed — the FE-only localStorage approach is inherently unreliable (doesn't survive across devices or storage clears). The right fix is for the backend to expose the SSO discovery domain on the credential metadata so any device can render it correctly. Tracked as #3795 (persist `discovery_domain` on `LastUsedIdentity` / credential metadata). Will address in a follow-up PR.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/components/OpenIdItem.svelte
timothyaterton: Implemented in f8ff1de — added an optional `name` field to the `IIOpenIdConfigurationSchema` (the zod schema for `/.well-known/ii-openid-configuration`). The `SsoDiscoveryResult` now carries `orgName`, which flows through `ssoDomainStorage` and into `openIdName`. When a discovery document includes `{ "name": "DFINITY" }`, the access-methods UI renders "DFINITY" instead of the domain; if `name` is absent, the domain (`dfinity.org`) remains the fallback. Once the backend solution (comment above) lands, the name can come directly from the credential metadata without the localStorage bridge.

src/frontend/src/lib/components/wizards/auth/views/PickAuthenticationMethod.svelte
aterga: Addressed in follow-up PR #3803 — commit `fix(fe): use "Continue with SSO" instead of "Sign in with SSO"` updates the `aria-label` in `PickAuthenticationMethod.svelte` + `AddAccessMethod.svelte` and the h1 in `SignInWithSso.svelte`. https://github.com/dfinity/internet-identity/pull/3803

src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
aterga: Addressed in follow-up PR #3803. `mapSubmitError` now only keeps branches the user (or their SSO admin) can act on — domain-not-configured, canary allowlist, OAuth-provider errors surfaced from the callback fragment, and rate-limits. Everything else falls back to a generic `"SSO sign-in for <domain> failed"` message, and all thrown errors are `console.error`'d unconditionally so the stack is still there for devs. https://github.com/dfinity/internet-identity/pull/3803

src/frontend/src/lib/flows/addAccessMethodFlow.svelte.ts
aterga: Addressed in follow-up PR #3803. `SignInWithSso` now takes an optional `openIdCredentials` prop; in the add-access-method flow, `AddAccessMethodWizard` passes the identity's credentials through, and a `$derived` check disables the Continue button with an inline hint once two-hop discovery reveals the SSO's `(iss, aud)` matches an already-linked credential — mirroring the direct-provider button-disable in `AddAccessMethod.svelte`. The `OpenIdCredentialAlreadyLinkedHereError` specialization in `linkOpenIdAccount` and the corresponding `handleError` branch are removed as unreachable. https://github.com/dfinity/internet-identity/pull/3803

src/frontend/src/lib/utils/ssoDomainStorage.ts
aterga: Addressed in follow-up PR #3803 — `ssoDomainStorage.ts` and both its callers are deleted. The canister now stamps `sso_domain` (and optional `sso_name`) onto the credential metadata inside `DiscoverableProvider::verify`. Each `DiscoverableProvider` carries its `discovery_domain` and a `discovered_name` ref that hop-1 populates from `{domain}/.well-known/ii-openid-configuration`. `openIdName` / `openIdLogo` on the FE read the metadata keys directly, so labels survive across devices and reloads. https://github.com/dfinity/internet-identity/pull/3803

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/components/OpenIdItem.svelte
aterga: Addressed in follow-up PR #3803. The hop-1 schema (`IIOpenIdConfiguration` on the canister) gained an optional `name: Option<String>` with `#[serde(default)]`; when present, the canister stamps it as `sso_name` on the credential metadata (alongside the always-present `sso_domain`). `openIdName` resolves `sso_name` → `sso_domain` → `findConfig(iss, aud).name`, so the access-methods list reads `"DFINITY account"` when the domain publishes `name: "DFINITY"`, and falls back to `"dfinity.org account"` when it doesn't. https://github.com/dfinity/internet-identity/pull/3803

PR #3786, opened 2026-04-16
feat(fe): SSO domain input and two-hop discovery flow

timothyaterton:
Users whose organization uses SSO don't want to know which OIDC provider their IT team chose — they just want to type `dfinity.org` (or their own domain) and sign in. This PR adds a dedicated "Sign in with SSO" entry point in the auth picker: the user types a domain, the frontend discovers the provider for that domain in two hops, and falls through into the existing OpenID sign-in flow.

Built on top of #3785.

# Changes

New two-hop discovery chain in `ssoDiscovery.ts`:
1. `GET https://{domain}/.well-known/ii-openid-configuration` — the domain's DNS owner publishes their `client_id` plus a pointer to the real OIDC provider's discovery URL.
2. `GET {openid_configuration}` — standard OIDC discovery from the provider, reusing the same validation + caching as the direct-provider path.
3. Synthesize an `OpenIdConfig` and hand off to `continueWithOpenId`.

Files:
- `ssoDiscovery.ts` (new): two-hop discovery with domain-format validation, PoC allowlist, trusted-provider-domain check, per-domain rate limit, caching, exponential backoff.
- `SignInWithSso.svelte` (new): domain input with debounced (750ms) discovery fetch and inline status feedback.
- `SsoIcon.svelte` (new): key icon for the SSO button.
- `authFlow.svelte.ts`: new `signInWithSso` view state and `continueWithSso()` method. Uses the shared `selectAuthScopes()` helper from #3785 so empty advertised-scope lists fall back to canonical defaults.
- `PickAuthenticationMethod.svelte`: shows the SSO button when `oidc_configs` has any entry with `client_id = null` (the signal that this provider is SSO-gated).
- `AuthWizard.svelte`: wires the SSO view and handler.

Security (layered on top of the OIDC discovery hardening from #3785):
- Domain input validated for DNS format, length, and forbidden characters.
- **PoC allowlist — only `dfinity.org` is accepted.** This will be replaced with a configurable allowlist before this ships to production.
- `openid_configuration` URL must be HTTPS and on the trusted-provider-domain allowlist (Google, Apple, Microsoft, Okta).
- Rate limiting: 1 request per domain per 10 minutes, max 2 concurrent SSO discoveries, exponential backoff on failure.
- Caching: 4-hour TTL for both hops.
- Timeouts: 5s for the `ii-openid-configuration` hop, 10s for the provider discovery hop.

# Tests

- 24 unit tests in `ssoDiscovery.test.ts` (domain validation, allowlist, two-hop chain, failure/retry, caching, trusted-provider check).
- No regressions in the 45 tests from #3785.
- `npm run lint` clean, svelte-check clean on touched files.

---
[< Previous PR](https://github.com/dfinity/internet-identity/pull/3785)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2189,6 +2189,7 @@ dependencies = [
  "serde_cbor",
  "serde_json",
  "sha2 0.10.9",
+ "url",
 ]
 
 [[package]]
--- a/src/canister_tests/src/api/internet_identity.rs
+++ b/src/canister_tests/src/api/internet_identity.rs
@@ -368,7 +368,7 @@ pub fn config(
 pub fn discovered_oidc_configs(
     env: &PocketIc,
     canister_id: CanisterId,
-) -> Result<Vec<types::DiscoveredOidcConfig>, RejectResponse> {
+) -> Result<Vec<types::OidcConfig>, RejectResponse> {
     call_candid(
         env,
         canister_id,
--- a/src/canister_tests/src/api/internet_identity.rs
+++ b/src/canister_tests/src/api/internet_identity.rs
@@ -365,15 +365,15 @@ pub fn config(
     call_candid(env, canister_id, RawEffectivePrincipal::None, "config", ()).map(|(x,)| x)
 }
 
-pub fn active_oidc_configs(
+pub fn discovered_oidc_configs(
     env: &PocketIc,
     canister_id: CanisterId,
-) -> Result<Vec<types::ActiveOidcConfig>, RejectResponse> {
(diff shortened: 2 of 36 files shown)


timothyaterton:
Folded into #3785 after aligning on the design. The `sso-ui` branch's two-hop discovery + domain-input UI now live on `frontend-oidc-discovery` as a single PR, with the allowlist sourced from the backend's `oidc_configs` instead of the hardcoded `dfinity.org` from this branch.

See [3d587a18](https://github.com/timothyaterton/internet-identity/commit/3d587a18) on #3785 for the consolidated change.

PR #3798, opened 2026-04-21
feat(fe): decode ICRC-3 attributes in the test app

aterga:
The test app surfaces ICRC-3 attributes as a base64-encoded Candid blob, which is not human-readable during manual testing. This PR decodes the blob client-side so the attribute key/value pairs are visible directly in the UI.

# Changes

- Add `demos/test-app/src/icrc3.ts`: a recursive `IDL` type for ICRC-3 `Value` plus a small formatter, using `IDL.decode` from `@icp-sdk/core/candid` (already a project dependency).
- Render the decoded attribute map in a new `ICRC-3 Attributes (decoded)` section of the test app, above the existing raw base64 block.


demos/test-app/src/icrc3.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+  const argRef = Number(reader.sleb());
+  const decoded = decodeValue(reader, resolve(argRef, table), table) as unknown;
+
+  return raw2Icrc3Value(decoded);
+}
+
+// Turn the intermediate decoder output (records keyed by field hash) into the
+// nicely-typed Icrc3Value form. Map records use Candid tuple-field hashes 0/1.
+function raw2Icrc3Value(v: unknown): Icrc3Value {
```
sea-snake: The test app could use the decoder from `@icp-sdk/core/principal` (already a dependency) instead of reimplementing this.

See snippet from e2e tests below:
```
import { IDL } from "@icp-sdk/core/candid";

type Icrc3Value =
  | { Nat: bigint }
  | { Int: bigint }
  | { Blob: number[] }
  | { Text: string }
  | { Array: Icrc3Value[] }
  | { Map: [string, Icrc3Value][] };

const Icrc3Value = IDL.Rec();
Icrc3Value.fill(
  IDL.Variant({
    Nat: IDL.Nat,
    Int: IDL.Int,
    Blob: IDL.Vec(IDL.Nat8),
    Text: IDL.Text,
    Array: IDL.Vec(Icrc3Value),
    Map: IDL.Vec(IDL.Tuple(IDL.Text, Icrc3Value)),
  }),
);

function decodeIcrc3Map(base64Data: string): Record<string, Icrc3Value> {
  const dataBytes = fromBase64(base64Data);
  const { Map: map } = IDL.decode([Icrc3Value], dataBytes)[0] as {
    Map: [string, Icrc3Value][];
  };
  return Object.fromEntries(map);
}
```

aterga: Good call — switched to `IDL.decode` in 2ae41300. Dropped the hand-rolled decoder entirely; module is now ~60 LOC.

sea-snake (approved):

PR #3803, opened 2026-04-24
feat(be,fe): address SSO review feedback on #3785 (follow-up)

aterga:
Follow-up to #3785 addressing [@sea-snake's review](https://github.com/dfinity/internet-identity/pull/3785#pullrequestreview-2710015497). All five unresolved threads there are addressed here, and will be marked resolved on #3785 with pointers back to this PR.

Stacked on top of #3785: this branch includes all commits from #3785 plus three new commits that cover the review fixes.

# Changes

## 1. `Sign in with SSO` → `Continue with SSO`

SSO button `aria-label` in `PickAuthenticationMethod` and `AddAccessMethod`, and the h1 inside the SSO screen, now read "Continue with SSO" — the downstream flow can end in either a sign-in or a sign-up, so "Sign in" was misleading.

## 2. Slim `mapSubmitError` to user-actionable copy

`SignInWithSso.svelte#mapSubmitError` used to enumerate every shape of provider misconfiguration (hostname mismatch, non-HTTPS endpoints, malformed discovery document, ...). Those read as implementation-detail leakage. Keep only the branches the user or their SSO admin can act on — domain not configured, canary allowlist, OAuth provider errors from the callback fragment, rate limits — and fall back to a generic `"SSO sign-in for <domain> failed"` for everything else. Every thrown error is `console.error`'d unconditionally so engineers still have the stack.

## 3. Disable Continue when the SSO is already linked to this identity

`SignInWithSso` takes a new optional `openIdCredentials` prop. In the add-access-method flow, `AddAccessMethodWizard` passes the identity's current credentials through; once two-hop discovery reveals the SSO's `(iss, aud)`, a `$derived` check disables the Continue button with an inline hint if that SSO is already linked. Mirrors how `AddAccessMethod.svelte` already disables direct-provider buttons for already-linked providers. Left `undefined` in the sign-in flow, where reusing an existing credential is the point.

This makes reaching `OpenIdCredentialAlreadyRegistered` for *this* identity impossible, so the `OpenIdCredentialAlreadyLinkedHereError` specialization in `linkOpenIdAccount` and the corresponding toaster branch in `handleError` are dropped as unreachable.

## 4 + 5. Move SSO domain + name from FE localStorage to canister-stamped credential metadata

`ssoDomainStorage.ts` (a per-device localStorage map of `(iss, sub, aud) → domain`) is removed. The canister now stamps two new metadata keys on any credential verified by a `DiscoverableProvider`:

- `sso_domain` — the `discovery_domain` the user entered at sign-up. The canonical SSO label; always present for SSO credentials.
- `sso_name` — optional human-readable name from `{domain}/.well-known/ii-openid-configuration`. When the domain publishes `name: "DFINITY"`, the access-methods list reads "DFINITY account" instead of "dfinity.org account".

BE:
- `IIOpenIdConfiguration` hop-1 schema gets `name: Option<String>` (with `#[serde(default)]` so older deployments still parse).
- `DiscoverableProvider` and `DiscoveryState` carry `discovery_domain` and a `discovered_name` ref that hop-1 populates each refresh.
- `DiscoverableProvider::verify()` inserts the two keys before returning the credential.

FE:
- `openIdName` now resolves `sso_name` → `sso_domain` → `findConfig(iss, aud, metadata).name`, so an SSO-via-Google credential reads as "DFINITY" (or "dfinity.org" if the domain doesn't publish a name), not "Google".
- `openIdLogo` returns `undefined` when `sso_domain` is present (generic SSO icon).
- `OpenIdItem.svelte` detects SSO via `sso_domain` directly, instead of the previous "no logo found" heuristic that was brittle for direct-provider credentials whose issuer didn't match any `openid_configs` entry.
- `authFlow#continueWithSso` and `addAccessMethodFlow#linkSsoAccount` both simplify — no pre-decoding the JWT or remembering the domain locally — because the canister handles the labeling end-to-end, so the mapping survives reloads and crosses devices.

# Tests

- `cargo test -p internet_identity --bin internet_identity`: **219/219 pass**.
- `cargo clippy --all-targets -D warnings`: clean.
- `cargo fmt --check`: clean.
- `npm run check` (tsc + svelte-check): 0 errors (19 pre-existing warnings unchanged).
- `npm run lint` + `prettier --check`: clean.

---
[< Previous PR](https://github.com/dfinity/internet-identity/pull/3785)


src/frontend/src/lib/components/utils/error.ts
```diff
@@ -28,6 +28,25 @@ export const handleError = (error: unknown) => {
...
+  // this branch covers callers (direct-OpenID entry points) that route
+  // through `handleError` instead.
+  if (error instanceof OAuthProviderError) {
+    toaster.error({
+      title: `SSO provider returned "${error.error}"`,
+      description:
+        error.errorDescription !== undefined &&
+        error.errorDescription.length > 0
+          ? error.errorDescription
+          : "Ask your SSO admin to check the app's OAuth configuration.",
```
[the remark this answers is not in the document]
aterga: Good catch — broadened in `addbb671`. The toast title now reads `OAuth provider returned "<code>"` (covers both SSO and direct-provider errors), and the description fallback is generic: `"Ask your administrator to check the provider's OAuth configuration."` The `SignInWithSso`-specific guidance (e.g. the `unsupported_response_type` hint about `Implicit (hybrid)`) still lives in `mapSubmitError` where the caller knows it's SSO.

src/frontend/src/lib/utils/openID.test.ts
```diff
@@ -230,7 +240,189 @@ describe("findConfig", () => {
...
     ).toBeUndefined();
   });
+
+  it("falls back to issuer-only match when aud doesn't match any config", () => {
+    // Legacy/migrated credentials can have an `aud` that doesn't line up
+    // with the current `openid_configs` entry (e.g. client_id rotation).
+    // We prefer returning SOMETHING so the UI can label these direct-
+    // provider credentials correctly, and rely on the localStorage SSO
+    // map in `openIdName`/`openIdLogo` to short-circuit the SSO case
+    // before `findConfig` is consulted.
```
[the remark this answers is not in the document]
aterga: Fixed in `addbb671`. The test comment now describes the current mechanism: SSO credentials are disambiguated earlier via the canister-stamped `sso_domain` / `sso_name` metadata, and `openIdName` / `openIdLogo` short-circuit on those before ever consulting `findConfig`, so the issuer-only fallback can't mis-attribute an SSO credential.

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -0,0 +1,477 @@
...
+ *    → returns `{ client_id, openid_configuration }`
+ * 2. Fetch the standard OIDC discovery document from the `openid_configuration`
+ *    URL → returns `authorization_endpoint`, `scopes_supported`, etc.
+ *
+ * Security — trust model:
+ * - The caller must call `add_discoverable_oidc_config({ discovery_domain })`
+ *   on the backend BEFORE invoking `discoverSsoConfig`. That call traps on
+ *   the backend's canary allowlist (`ALLOWED_DISCOVERY_DOMAINS`) for any
+ *   domain an II admin hasn't approved, so by the time this module runs,
+ *   the domain has been explicitly blessed by II.
```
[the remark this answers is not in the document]
aterga: Fixed in `addbb671`. The header now references `openid::generic::allowed_discovery_domains()`.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/(access-and-recovery)/access/components/OpenIdItem.svelte
```diff
@@ -33,16 +47,18 @@
...
+    {:else}
       {@html logo}
-      {#if isCurrentAccessMethod}
-        <div
-          class="bg-bg-success-secondary border-bg-primary absolute -top-0.25 -right-0.5 size-2.5 rounded-full border-2"
-        ></div>
-      {/if}
-    </div>
-  {/if}
+    {/if}
```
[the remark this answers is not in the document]
aterga: Nice catch — fixed in `addbb671`. The guard is now `{#if isSso || logo === undefined}` so the generic `SsoIcon` renders as a fallback for both SSO credentials (the intended case) and direct-provider credentials whose issuer doesn't match any `openid_configs` entry (e.g. after a client_id rotation — exactly the case Copilot flagged).

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -0,0 +1,477 @@
...
+        throw error;
+      }
+    } finally {
+      // Clear the abort timer on every exit path so a late abort can't fire
+      // after the attempt is already done.
+      clearTimeout(timeoutId);
+    }
+  }
+  throw lastError;
+};
```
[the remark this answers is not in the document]
aterga: Addressed in `addbb671`. Introduced an `isTransientHttpStatus` predicate that returns true only for 408 / 429 / 5xx; `fetchWithRetry` classifies each attempt and breaks out of the loop immediately on deterministic 4xx (404, 400, 401, 403, ...). Network errors (`TypeError` from `fetch`) and unknowns still retry. Added a test that asserts `fetch` is called exactly once for a 404 (so future regressions would show up as 3 calls with 14s of backoff), and a companion test for the 503 path that still goes through the full retry budget.

src/internet_identity/src/openid/generic.rs
```diff
@@ -352,6 +364,20 @@ impl OpenIdProvider for DiscoverableProvider {
...
+        // SSO credentials correctly across devices — without this, the
+        // per-device localStorage hack the previous iteration carried
+        // couldn't survive reloads, let alone a different browser.
+        metadata.insert(
+            "sso_domain".into(),
+            MetadataEntryV2::String(self.discovery_domain.clone()),
+        );
+        if let Some(sso_name) = self.discovered_name.borrow().clone() {
+            metadata.insert("sso_name".into(), MetadataEntryV2::String(sso_name));
+        }
```
sea-snake: This shouldn't be metadata, instead it should lookup the SSO domain and email (oidc cofnig or such) by the credential iss+aud pair and additionally return those in the credential when requested (optional field(s)).

src/internet_identity/src/openid/generic.rs
```diff
@@ -490,6 +527,9 @@ async fn run_discovery_tasks() {
         task.openid_configuration_ref
             .replace(Some(ii_config.openid_configuration));
         task.issuer_ref.replace(Some(doc.issuer));
+        // `name` may be `None` (field absent / older deployment) — in that
+        // case the FE falls back to the `discovery_domain` for the label.
+        task.name_ref.replace(ii_config.name);
```
sea-snake: Fallback should be done where needed in the frontend, not backend. This enables the frontend to be aware if it actually has a name to render or only a domain, making it possible in the future to render those two cases differently where needed.

src/internet_identity/src/openid/generic.rs
aterga: Refactored in `fbc35927`. `DiscoverableProvider::verify` no longer stamps `sso_domain` / `sso_name` into the credential metadata. Instead:
- `OpenIdCredentialData` gains two optional fields (`sso_domain`, `sso_name`) on the Candid side.
- New helper `openid::generic::sso_fields_for(iss, aud)` consults current `DISCOVERY_TASKS` state to look up both.
- The `OpenIdCredential → OpenIdCredentialData` conversion in `storage/anchor.rs` calls that helper on every credential shaped for API output.

Computing on demand means if an SSO admin updates the `name` in `/.well-known/ii-openid-configuration` later, existing credentials pick it up on the next `get_anchor_info` with no stored-metadata migration; and if an SSO is de-registered from the canary allowlist, the old credentials correctly surface `sso_domain = None` (we don't know anything about that SSO anymore). Metadata stays the place for per-credential JWT claims (`email`, `name`, `tid`, ...), not for cross-credential SSO state.

src/internet_identity/src/openid/generic.rs
aterga: Fixed in `fbc35927`. Both the backend and the Candid type carry `sso_name` and `sso_domain` as independent `Option<>` / `[] | [string]` — the BE never collapses one into the other. `openIdName` on the FE resolves `ssoName → ssoDomain → findConfig(iss, aud).name`, so future FE changes that want to render "has a name" differently from "only a domain" have both values available. Also updated the hop-1 write comment and the `IIOpenIdConfiguration` docstring to reflect the no-fallback-in-BE stance.

src/internet_identity/src/storage/anchor.rs
```diff
@@ -127,18 +127,30 @@ impl From<Device> for DeviceDataWithoutAlias {
...
+        let (sso_domain, sso_name) =
+            crate::openid::sso_fields_for(&openid_credential.iss, &openid_credential.aud);
         Self {
             iss: openid_credential.iss,
             sub: openid_credential.sub,
             aud: openid_credential.aud,
             last_usage_timestamp: openid_credential.last_usage_timestamp,
             metadata: openid_credential.metadata,
+            sso_domain,
+            sso_name,
```
sea-snake: These two fields should be together in a single struct assigned to a field called `sso_configuration`.

sea-snake (dismissed):

src/internet_identity/src/storage/anchor.rs
aterga: Done in `93f5fabc`. Introduced a Candid `SsoConfiguration { domain : text; name : opt text }` record and replaced the two separate optional fields on `OpenIdCredentialData` with a single `sso_configuration : opt SsoConfiguration`. `sso_configuration : null` is the single "not an SSO credential" check; when present, `domain` is always populated and `name` stays optional so the FE can still render "has a name" vs. "only a domain" differently (fallback lives on the FE side per the earlier thread). The canister-side lookup is now `openid::generic::sso_configuration_for(iss, aud) -> Option<SsoConfiguration>`.

PR #3805, opened 2026-04-24
feat(be): add `sso:<domain>` attribute scope alongside `openid:<issuer>`

aterga:
## Summary

Adds a new `AttributeScope::Sso { domain }` variant that lets relying parties request certified `name` / `email` attributes under an SSO domain key (e.g. `sso:dfinity.org:email`) rather than only under the upstream OIDC issuer URL.

**Scope exclusivity**: a stored `OpenIdCredential` is addressable via exactly one scope.

- Credentials matched to a `DiscoverableProvider` (two-hop SSO discovery, `DiscoverableOidcConfig`) are reachable only via `sso:<domain>`.
- Credentials from hardcoded providers (Google, Microsoft, generic OIDC) remain reachable only via `openid:<issuer>`.

`verified_email` is intentionally not supported under `sso:` yet — it's silently dropped in the `prepare`/`list` flows and surfaces as `AttributeMismatch` in the ICRC-3 flow (consistent with that flow's existing explicit-presence semantics).

## What changed

- **Types** (`internet_identity_interface`):
  - New `AttributeScope::Sso { domain }` variant + `Display` impl.
  - `validate_sso_domain` — format-only (non-empty, no `:`, no whitespace, ≤ 253 bytes). Intentionally decoupled from `allowed_discovery_domains()` so the parser stays pure; the canary allowlist still gates `add_oidc_config` admission.
  - `AttributeScope::try_from` handles the new `"sso"` prefix.
- **OpenID plumbing** (`src/internet_identity/src/openid.rs`, `openid/generic.rs`):
  - New `OpenIdProvider::discovery_domain() -> Option<String>` trait method with default `None`.
  - `DiscoverableProvider` stores and returns its `discovery_domain`.
  - `OpenIdCredential::discovery_domain()` delegates via the existing `with_provider` matching.
- **Attribute flows** (`src/internet_identity/src/attributes.rs`):
  - New `Anchor::prepare_sso_attributes` mirroring `prepare_openid_attributes`, restricted to `Email` / `Name`.
  - Exclusivity filters added to `prepare_openid_attributes`, `get_attributes`, `list_available_attributes`, and `prepare_icrc3_attributes`.
  - `prepare_attributes` now runs both preparers; their per-credential output maps are disjoint, so a simple `or_else` per credential is sufficient.
- **Candid** (`internet_identity.did`): doc-only updates on `PrepareAttributeRequest.attribute_keys`, `AttributeSpec.key`, and `ListAvailableAttributesRequest.attributes`. No new request/response types — keys remain `text`.

## Test plan

- [x] Unit tests (`internet_identity_interface` / `internet_identity`, passing locally): scope parsing, key round-trip, sso extraction, exclusivity (openid skips sso credentials, sso skips plain credentials), `verified_email` silent-drop, `list_available_attributes` only lists the chosen scope, ICRC-3 returns `AttributeMismatch` for `sso:*:verified_email` and for `openid:` on an SSO-sourced credential.
- [x] Integration test `sso_scope_request_is_accepted_and_silently_drops_when_no_sso_credential` — verifies the request validator accepts the new scope and the mixed openid+sso request silently drops the sso: part when no SSO credential exists. Deeper SSO flows cannot be exercised from PocketIC because two-hop discovery requires HTTP outcalls; they are covered by unit tests using a `#[cfg(test)]` helper that populates the discovery state directly.
- [x] `cargo check` / `cargo clippy` clean on both crates (only pre-existing warnings remain).
- [ ] Full CI suite on this PR.

https://claude.ai/code/session_01HFm4Dj2FdnT8nEBzQj2zPm

---
_Generated by [Claude Code](https://claude.ai/code/session_01HFm4Dj2FdnT8nEBzQj2zPm)_


src/internet_identity/src/attributes.rs
aterga, line 121: This is the legacy function that should fail if the new SSO-scoped attributes are requested.

aterga: Please indicate in the doc string that this function (and the whole legacy attribute sharing flow) are deprecated and should not be changed in the future. Instead, the code for the ICRC-3 attributes should be used and evolved.

sea-snake (changes requested):
Need to modify prepare_icrc3_attributes and the corresponding ICRC-3 flow, and not change the legacy attribute sharing flow (non icrc3).

sea-snake (approved):

PR #3836, opened 2026-05-04
docs(ongoing): design doc for verifiable email recovery

sea-snake:
📄 **Read the design doc directly:** [`docs/ongoing/email-recovery.md` on the fork](https://github.com/sea-snake/internet-identity/blob/design/email-recovery/docs/ongoing/email-recovery.md)

---

## Summary

Adds `docs/ongoing/email-recovery.md` — a design doc for the production approach that should supersede the DKIM postbox PoC in #3760.

The design covers four threads in one architecture:

1. **DKIM hardening** — replace the PoC's hand-rolled parser with a vetted library (`mail-auth`), implement trusted-body retention (`l=`), enforce missing tag checks (`i=`, `k=`, future-dated `t=`), and change the gateway contract to deliver raw header bytes so byte-exact canonicalization is possible.
2. **DMARC alignment** — parse the message `From:` header (not the SMTP envelope), check DKIM `d=` against the From: domain under `adkim=s`/`adkim=r` using a bundled Public Suffix List, and surface the outcome in a renamed `EmailVerificationStatus`.
3. **DNSSEC arguments instead of HTTP outcalls** — callers (browser) fetch DKIM/DMARC TXT records along with the full DNSSEC chain to root and pass them to the canister as a `DnsProofBundle`. The canister verifies the chain against an IANA root anchor delivered as a deploy arg. No DoH outcall during recovery; deterministic verification without consensus tricks.
4. **Email recovery as an authn method** — alongside the recovery phrase, the user enters their email, submits the DNSSEC records, gets back a one-line magic token, sends one email containing the token, and is registered (setup) or signed in (recovery). §8 walks through both flows step by step with mermaid diagrams and ASCII UX mockups.

The doc also includes a threat model, a test-corpus plan, a phased rollout, and the open questions (DNSSEC root key management, alias handling, IDN, enumeration mitigations).

Per request, this PR is the design doc only — no code changes. The PoC PR (#3760) is expected to close once Phase 0 of the rollout in §10 lands as a fresh PR series.

## Reading order

- §1–§4: motivation and threat model.
- §5–§7: DKIM, DMARC, and DNSSEC components.
- §8: the email recovery flow itself — start with §8.1 for the user-visible shape, §8.4–§8.5 for the call sequences, §8.6 for screen mockups.
- §10: rollout / how this lands.

cc @aterga


docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+**Targets:** A new production-grade PR series, not a follow-up to #3760.
+
+---
+
+## 1. Background
+
+Internet Identity currently offers two recovery options when a user can no longer authenticate with any of their registered passkeys:
+
+1. **Recovery phrase** — a BIP-39-style seed phrase the user is responsible for storing offline.
+2. **Recovery device** — a separate WebAuthn key bound to the same anchor.
```
sea-snake: Recovery devices are no longer a thing, this is a legacy thing that has since no longer been made available to end-users in the frontend.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+Both require the user to have *prepared* a recovery method before losing access, and both require the user to retain *something* (paper, hardware) outside the device that is now unusable. We hear from users that both fall through in practice — phrases get lost, recovery devices live next to the primary device and are lost together.
+
+Email is the recovery channel almost every user already has and almost every user can reach from any browser. The PoC PR [#3760](https://github.com/dfinity/internet-identity/pull/3760) added enough plumbing to *receive* DKIM-signed emails inside the canister and view them in a "Postbox" tab, but it deliberately did not wire emails into the recovery surface, and several DKIM/DMARC correctness items were deferred.
+
+This doc proposes the production design that supersedes the PoC. The PoC PR will be closed; the work below should land as a fresh PR series against `main`.
+
+### What the PoC got right
+
+- Candid surface for the SMTP gateway (`smtp_request`, `smtp_request_validate`).
+- Stable storage layout for received messages (`smtp_postbox`, memory ID 23).
```
sea-snake: Given it's a recovery feature, we'd like to avoid storing incoming emails if possible. If DNSSEC records are received priorly, there's no need to store the emails, they can be used immediately to either recover or set-up recovery.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+## 2. Goals & non-goals
+
+**Goals**
+
+- A user can register an email address as a recovery method, and use it to regain access if they lose every other authn method.
+- DKIM verification is RFC-6376-compliant against the corpus of mainstream senders (Gmail, iCloud, Outlook, Fastmail, Proton, ProtonMail, Tutanota, AWS SES, SendGrid, Postmark, Mailgun).
+- DMARC alignment is checked and enforced according to the sender's published policy.
+- The canister's DKIM/DMARC verification is **fully deterministic** and does not depend on HTTPS outcalls during the recovery flow.
+- A registered email holder can prove control of that address with a single signed email; the canister does not have to trust the SMTP gateway for *anything other than message delivery*.
+- The Postbox feature can ride on the same primitives but is gated independently.
```
sea-snake: Postbox feature was a PoC, this is out of scope for email recovery.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+This buys us:
+
+- **Determinism without consensus tricks.** No HTTP transform, no `max_response_bytes`, no boundary-node trust.
+- **No cycles spent on outcalls** during recovery, which is the latency-sensitive path.
+- **Full transparency.** The DNS records the canister verified are part of the audit trail; the gateway/user can replay the call against any node.
+- **Recovery works even if all DoH providers go down.** The user's browser fetches DNS via the OS resolver; we just need a DNSSEC-aware client (a small WASM library, see §7.4).
+
+We pay for it in:
+
+- **Caller complexity.** The user's browser (or the SMTP gateway, for Postbox) has to assemble the DNSSEC chain. We ship a TS library that wraps this.
```
sea-snake: Yet again you're metnioning post box, which was a PoC that's out of scope.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+```
+
+The key architectural shift versus the PoC is the path through which the canister learns the sender's DKIM/DMARC DNS records. The PoC fetched them via an HTTPS outcall to `dns.google` with a transform function for replica consensus. The new design has the *caller* fetch DNS, fetch the DNSSEC proof chain, and pass both to the canister, which validates the chain cryptographically.
+
+This buys us:
+
+- **Determinism without consensus tricks.** No HTTP transform, no `max_response_bytes`, no boundary-node trust.
+- **No cycles spent on outcalls** during recovery, which is the latency-sensitive path.
+- **Full transparency.** The DNS records the canister verified are part of the audit trail; the gateway/user can replay the call against any node.
+- **Recovery works even if all DoH providers go down.** The user's browser fetches DNS via the OS resolver; we just need a DNSSEC-aware client (a small WASM library, see §7.4).
```
sea-snake: Why does this need to be a library and why wasm? I think both can be avoided.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+
+### 8.8 Removed: PoC web push notification path
+
+The PoC's `web_push.rs` and service-worker `postMessage` handler are *not* part of the email-recovery delivery. Push notifications are nice-to-have for Postbox but introduce the CodeQL alert (#127) and 700+ lines of code. Push lands in a separate PR; recovery does not depend on it.
+
+---
+
+## 9. Postbox revisited
+
+The Postbox feature does still have value (anchor-as-mailbox for app-to-user comms, future ICRC-style notifications), but it stops being a load-bearing first deliverable.
```
sea-snake: Yet again, you're mentioning functionality that's an out of scope PoC

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+
+### 8.7 Frontend changes
+
+- **Manage page**: a new "Email recovery" row alongside "Recovery phrase" and "Recovery device". Standard add/remove buttons.
+- **`promptRecovery` screen** (see `FLOWS.mdx`): a third option, "Recover with email", sits next to "Recover with phrase" and "Recover with device".
+- **`recoverWithEmail` screen**: address entry → instructions to send → spinner that polls the gateway → success.
+
+### 8.8 Removed: PoC web push notification path
+
+The PoC's `web_push.rs` and service-worker `postMessage` handler are *not* part of the email-recovery delivery. Push notifications are nice-to-have for Postbox but introduce the CodeQL alert (#127) and 700+ lines of code. Push lands in a separate PR; recovery does not depend on it.
```
sea-snake: Also here, out of scope stuff.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+| `email_recovery_prepare_register` per anchor | 5 / hour | UX cost of sending an email is asymmetric — we don't need fast retries |
+| `email_recovery_prepare_delegation` per address_hash | 5 / 15 min | Slow attacker who controls an inbox |
+| `email_recovery_register` per anchor | max 3 active addresses | Bounded storage |
+| Total stored emails per anchor | unchanged from PoC (10) | Postbox |
+
+Limits are enforced in-memory in `state` with stable-memory backed counters keyed by `(anchor, hour_bucket)`; the existing rate-limiter module is reused.
+
+### 8.7 Frontend changes
+
+- **Manage page**: a new "Email recovery" row alongside "Recovery phrase" and "Recovery device". Standard add/remove buttons.
```
sea-snake: Let's change the Recovery phrase row into a "Recovery methods" row, then that page can have a recovery phase and email recovery card.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+- **`promptRecovery` screen** (see `FLOWS.mdx`): a third option, "Recover with email", sits next to "Recover with phrase" and "Recover with device".
```
sea-snake: As I mentioned earlier, there's no recover with device anymore. Please ignore FLOWS.mdx that's very old stuff. Only trust the svelte code in the repo regarding the frontend routes and their flows.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+12. Remove the beta label; remove the feature flag.
+13. Land the Postbox UI (now just a consumer of the same primitives).
+
+We do **not** keep PoC PR #3760's WASM in any release. Its branch closes when Phase 0 lands.
+
+---
+
+## 12. Open questions
+
+- **DNSSEC root key in WASM vs. governance.** Bundling the root anchor in the WASM means any rollover requires a canister upgrade. An alternative is to store the trust anchor in the canister's persistent state, mutable only through a governance proposal. This is more flexible but introduces another governable parameter; ii doesn't have many today. *Default: bundle in WASM; revisit if rollover proves painful.*
```
sea-snake: Make it a deploy arg, we're already deploying at least once a week any way. Please mention how often the DNSSEC root key rolls over.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+- **Email aliases.** Should `alice+ii@gmail.com` and `alice@gmail.com` collapse to one `address_hash`? Gmail treats them as the same mailbox; Outlook does not. *Default: do not collapse — store the address the user typed, verbatim and lowercased; let the user pick.*
```
sea-snake: Do not collapse!

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+- **Lost mailbox.** If a user's email account is gone (provider closed, domain expired), recovery is unrecoverable through this channel. Surface clearly; document that email recovery is *one* of multiple parallel recovery options.
```
sea-snake: That's a good point, but not really an issue in the current scope. Let's consider it out of scope for now.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+- **Privacy: enumeration.** A naïve `email_recovery_lookup_hint` lets an attacker test arbitrary addresses. We mitigate with rate limiting and salted hashing; should we also gate the lookup behind a captcha for unauthenticated callers? *Default: rate limit only, capture metrics, add captcha if abuse appears.*
```
sea-snake: The lookup is only used for email recovery, something you could only do if you sent a validly DKIM signed email from the correct address anyway. If you sent it from an incorrect address, the UI will simply keep waiting for the valid email and then ultimately time out. (Better ideas are welcome).

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+- **Multi-anchor per address.** Do we allow the same email on two anchors? OpenID currently does not. *Default: same — `EmailRecoveryRegisterError::AddressAlreadyRegistered`.*
```
sea-snake: We should not, else we wouldn't know which identity to recover for a given email (users don't know their anchor number).

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+- **Internationalised domains (IDN).** A-label vs U-label canonicalization matters for both the local-part and domain. *Default: ASCII lowercase the A-label form everywhere; reject U-label inputs at the Candid boundary.*
```
sea-snake: Seems fine for now.

docs/ongoing/email-recovery.md
sea-snake: Addressed — removed all recovery-device references. §1 now says: `(Earlier versions exposed a "Recovery device" flow as well; that surface is no longer offered to end users.)` and the rest of the doc only mentions recovery phrase.

docs/ongoing/email-recovery.md
sea-snake: Addressed — the canister never stores inbound email. Verification is in-flight only: the signed email + DNSSEC bundle are passed as call arguments, verified synchronously, and discarded. §2 non-goals and §4 make this explicit. §8.1 storage model only has the registered address + pending challenge nonces.

docs/ongoing/email-recovery.md
sea-snake: Addressed — all Postbox content removed. Old §9 ("Postbox revisited") is gone entirely. Remaining mentions are explicitly "out of scope" markers in §1 and §2.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+### 8.6 Rate limits
+
+| Surface | Limit | Reason |
+|---|---|---|
+| `email_recovery_prepare_register` per anchor | 5 / hour | UX cost of sending an email is asymmetric — we don't need fast retries |
+| `email_recovery_prepare_delegation` per address_hash | 5 / 15 min | Slow attacker who controls an inbox |
+| `email_recovery_register` per anchor | max 3 active addresses | Bounded storage |
+| Total stored emails per anchor | unchanged from PoC (10) | Postbox |
+
+Limits are enforced in-memory in `state` with stable-memory backed counters keyed by `(anchor, hour_bucket)`; the existing rate-limiter module is reused.
```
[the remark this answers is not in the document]
sea-snake: Checked — the tables already use single-pipe `|` format, not `||`. This appears to be a false positive from the diff context rendering. No change needed.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+
+```candid
+type EmailRecoveryRegisterArg = record {
+    // The challenge was issued via `email_recovery_prepare_register`.
+    challenge_id : blob;
+    // The signed email from the user's address proving control.
+    email_message : SmtpMessage;
+    // DNSSEC-validated DNS records used to verify DKIM/DMARC.
+    dns_proof : DnsProofBundle;
+};
```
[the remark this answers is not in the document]
sea-snake: Addressed — renamed to `EmailRecoveryProof` which is shared between registration and recovery. §8.2 has the updated Candid with a note explaining the rename.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+This also closes a subtle storage-bound issue: the PoC's `MAX_BODY_BYTES` was checked against the raw body, but `from_utf8_lossy` could expand U+FFFD past the bound. Operating on the canonicalized signed prefix sidesteps that entirely; aterga's `truncate_at_char_boundary` helper from the PoC remains the safety net at storage time.
+
+### 5.4 Tag enforcement
+
+Beyond the cryptographic check, the verifier rejects:
+
+- `v != 1`.
+- `a` outside the supported algorithm set: `rsa-sha256`, `ed25519-sha256`. (PoC supported `rsa-sha256` only.)
+- `t > now + skew_window` — future-dated signatures (PoC parsed but did not enforce).
+- `x < now` — expired signatures (PoC enforces this since `3137585324`).
```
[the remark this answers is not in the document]
sea-snake: Addressed — replaced with a link to the actual PoC PR comment thread: `[aterga's reply](https://github.com/dfinity/internet-identity/pull/3760#discussion_r3137585324)`.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+    root_keys = bundle.root_dnskey.unwrap_or(BUNDLED_ROOT_DNSKEY)
+    verify_rrsig(root_keys.rrsig, root_keys.rdata, IANA_ROOT_KSK)
+
+    # 2. walk down the chain
+    parent_keys = root_keys
+    for link in bundle.chain:
+        verify_rrsig(link.child_ds.rrsig, link.child_ds.rdata, parent_keys)
+        verify_ds_matches_dnskey(link.child_ds, link.child_dnskey)
+        verify_rrsig(link.child_dnskey.rrsig, link.child_dnskey.rdata, link.child_dnskey)
+        parent_keys = link.child_dnskey
```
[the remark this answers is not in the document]
sea-snake: Addressed — §7.3 now explicitly states the trust anchor is a DS digest (not a DNSKEY), and the verification steps are rewritten: (1) pick the DNSKEY whose KSK digest matches the trust anchor DS, (2) verify the root DNSKEY RRset RRSIG with that KSK, (3) walk the chain using DS→DNSKEY matching at each delegation.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+### What the PoC explicitly deferred
+
+From the PR review thread (sea-snake's comments and aterga's replies) the following were left for a follow-up:
+
+| Area | Status in PoC | Spec gap |
+|---|---|---|
+| Trusted body retention (`l=`) | Stores full body, hashes only signed prefix | Storage may include bytes not covered by the signature |
+| DKIM-Signature parser | Naive split on `;` and `=` | RFC 6376 §3.5 allows folding, arbitrary whitespace, multiple `b=`-like substrings inside other tag values |
+| DNS TXT record parser | Tolerant of `P=` casing only | RFC 6376 §3.6.2.2 allows folding across multiple TXT chunks, arbitrary whitespace |
+| Header canonicalization | "simple" rebuilt from parsed `(name, value)` | RFC 6376 §3.4.1 requires the *exact original* bytes; gateway contract change required |
```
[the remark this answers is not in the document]
sea-snake: Same as above — the tables use single-pipe format in the source. No `||` present.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+
+### 5.3 Trusted-body retention (`l=`)
+
+When a DKIM signature includes `l=N`, only the first N bytes of the canonicalized body are signed. Anything past byte N is unauthenticated and could have been appended by a forwarder or an attacker.
+
+**Storage rule:** `StorableEmail::body` is the *signed prefix* only, decanonicalized back to displayable text. The unsigned tail is dropped. If multiple signatures verify, we keep the prefix covered by the longest of them. We also persist `body_signed_len: Option<u32>` so the UI can show "(message truncated; remaining bytes were not signed)".
+
+This also closes a subtle storage-bound issue: the PoC's `MAX_BODY_BYTES` was checked against the raw body, but `from_utf8_lossy` could expand U+FFFD past the bound. Operating on the canonicalized signed prefix sidesteps that entirely; aterga's `truncate_at_char_boundary` helper from the PoC remains the safety net at storage time.
```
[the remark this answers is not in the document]
sea-snake: Addressed — §5.3 is rewritten. No storage at all now (emails aren't persisted). The nonce search operates on the canonicalized signed-prefix bytes directly; we explicitly note that relaxed canonicalization is not byte-reversible. No decanonicalization attempted.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+### 7.4 Caller-side helper
+
+We ship a TypeScript library `@dfinity/dnssec-bundle` that:
+
+- Takes a domain name and a record type.
+- Resolves via DoH (`dns.google` or `cloudflare-dns.com`, picked at random) requesting `do=1` for DNSSEC RRSIGs.
+- Walks the delegation chain by querying the DS at each parent.
+- Returns a `DnsProofBundle` ready to pass to the canister.
+
+This library is identical regardless of whether the caller is the SMTP gateway (assembling Postbox bundles) or the user's browser (assembling recovery bundles).
```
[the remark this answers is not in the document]
sea-snake: Addressed — §7.4 now notes that browsers don't expose the OS resolver, so the practical fallback is between multiple DoH endpoints (dns.google, cloudflare-dns.com), not to the OS. The earlier §4 claim about "works even if all DoH providers go down" is adjusted.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+
+The salt is per-anchor (already stored as `anchor.salt`), so two anchors with the same email produce different `address_hash` values. This frustrates trivial enumeration.
+
+A new stable map indexes `address_hash → AnchorNumber` to support the recovery lookup. Memory ID 24 (next free).
+
+### 8.2 Candid surface
+
+```candid
+type EmailRecoveryRegisterArg = record {
+    // The challenge was issued via `email_recovery_prepare_register`.
```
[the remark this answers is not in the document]
sea-snake: Addressed — the salted hash scheme is dropped entirely. The index is keyed by `lowercase(address)` directly (no salt). §3.1 explains why this is safe: the lookup is gated by DKIM verification, so enumeration requires mailbox control. No `anchor.salt` reference remains.

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+
+    // Step 2: submit the signed reply.
+    email_recovery_register : (IdentityNumber, EmailRecoveryRegisterArg)
+        -> (variant { Ok; Err : EmailRecoveryRegisterError });
+
+    // Recovery (no caller identity required).
+    email_recovery_prepare_delegation : (EmailRecoveryRegisterArg, SessionKey)
+        -> (variant { Ok : OpenIdPrepareDelegationResponse;
+                       Err : EmailRecoveryDelegationError });
+    email_recovery_get_delegation : (record { ... }) -> (...) query;
```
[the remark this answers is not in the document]
sea-snake: Addressed — §8.2 Candid surface now includes `email_recovery_prepare_recovery` and `email_recovery_get_delegation`. The old `email_recovery_lookup_hint` is removed (no pre-lookup needed; §8.4 explains why).

docs/ongoing/email-recovery.md
```diff
@@ -0,0 +1,659 @@
...
+| Tag | Meaning | Default |
+|---|---|---|
+| `v=DMARC1` | Required | — |
+| `p=` | Policy: `none` / `quarantine` / `reject` | required |
+| `sp=` | Subdomain policy | inherits `p=` |
+| `adkim=` | DKIM alignment mode: `s` strict, `r` relaxed | `r` |
+| `aspf=` | SPF alignment | (we don't check SPF — see §6.4) |
+| `pct=` | Percentage of failing mail to apply policy | `100` |
+| `fo=`, `rua=`, `ruf=`, `rf=` | Reporting | ignored |
+
```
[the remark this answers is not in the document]
sea-snake: Same as the other table comments — single-pipe in source, no change needed.

docs/ongoing/email-recovery.md
sea-snake: Addressed — all Postbox references removed or marked out-of-scope.

docs/ongoing/email-recovery.md
sea-snake: Agreed — dropped. §7.4 now describes inline TS (~300 lines in `src/frontend/src/lib/utils/dnssec/`) using DoH directly. No library, no WASM.

docs/ongoing/email-recovery.md
sea-snake: Addressed — section removed.

docs/ongoing/email-recovery.md
sea-snake: Addressed — section removed.

docs/ongoing/email-recovery.md
sea-snake: Addressed — §8.7 now says: rename the heading to **"Recovery methods"** and split into two cards (existing recovery phrase card + new email recovery card).

docs/ongoing/email-recovery.md
sea-snake: Addressed — FLOWS.mdx is explicitly noted as stale. §8.7 references only the svelte routes: `(access-and-recovery)/recovery/+page.svelte` for manage and `(new-styling)/recovery/+page.svelte` for recovery sign-in. No device recovery anywhere.

docs/ongoing/email-recovery.md
sea-snake: Addressed — §7.5 is rewritten: trust anchor is a deploy/upgrade arg (`EmailRecoveryConfig.dnssec_root_anchors`), refreshed alongside weekly deploys. Explicit note: "In practice the IANA root KSK rolls *very rarely*: once in DNSSEC's history, in October 2018. No further rollover has happened or is publicly scheduled."

docs/ongoing/email-recovery.md
sea-snake: Settled — §11 says: distinct addresses, not collapsed. The user registers exactly the address they typed, lowercased.

docs/ongoing/email-recovery.md
sea-snake: Marked as out of scope in §11.

docs/ongoing/email-recovery.md
sea-snake: Agreed — this is the design in §3.1 and §8.4 now. No pre-lookup; the FE just shows a "waiting…" spinner while polling the gateway. If the user typed the wrong address, it times out. The canister only sees the email after DKIM verification via `email_recovery_prepare_delegation`.

docs/ongoing/email-recovery.md
sea-snake: Agreed — §8.4 explicitly says: one anchor per address, `AddressAlreadyRegistered` if taken. Users don't know their anchor number, so the mapping must be functional.

docs/ongoing/email-recovery.md
sea-snake: Kept as-is in §11. ASCII-lowercase the A-label form; reject U-label inputs at the boundary.

timothyaterton (approved):
✅ LGTM — ship it; follow-up suggestions below.

This is one of the most thorough design docs I've reviewed. The two-phase DNSSEC architecture, the threat model, the DoH fallback, and the bounded-state-over-rate-limits reasoning are all well-grounded. A few observations and suggestions:

## Security

**[FOLLOW-UP] `email_recovery_status` and `email_recovery_get_delegation` are both query calls (§8.3).** A malicious boundary node could serve fabricated responses to both — e.g. returning a fake `RecoveryReady` followed by a fake `SignedDelegation`. The resulting delegation would fail verification at any relying party that checks II's certified root key, so the damage is limited to UX confusion (the user thinks they're signed in but can't actually do anything). This mirrors the existing `openid_get_delegation` pattern, so it's a known trust posture, not a regression. Worth a one-liner in the threat model noting the boundary-node assumption on the query path — purely for completeness.

**[FOLLOW-UP] §3.2 recovery phishing — outbound confirmation as upgrade path.** The nonce-rotation + Subject-visibility stack is the practical ceiling for an inbound-only flow, and the doc is honest about that. §11 mentions outbound email confirmation as the natural next step. One refinement worth noting for that future work: when the outbound confirmation step lands, the `email_recovery_prepare_delegation` endpoint should probably also start returning a short-lived opaque token that the confirmation link carries back, rather than reusing the same nonce for both directions — mixing inbound and outbound tokens in a single nonce namespace would make replay analysis harder. Not blocking now; just a design breadcrumb for the follow-up.

**[FOLLOW-UP] §6.1 — From: header parser rejects quoted local-parts.** The doc calls out that `"alice"@example.com` and obs-route forms are rejected as `MalformedFromHeader`. This is pragmatic for v1 since every mainstream provider emits bare `addr-spec`, but RFC 5321 §4.1.2 does allow quoted strings as local-parts. If user reports surface this, the fix is straightforward (extend the parser), but it might be worth a one-line note in §11 (Open questions) or the error-reporting UX (§8.7) so the "try a different email" copy is clear about *why*.

## Scalability

**[FOLLOW-UP] DoH fallback heap cache eviction under sustained load (§7.6).** The 1h default TTL with 24h cap is reasonable, but the doc doesn't specify a maximum cache size or eviction policy for the DoH heap cache itself. DKIM selectors are per-provider (not per-user) so the keyspace is small in practice, but a cap + LRU eviction would formalize the bound. This is an implementation detail for the DoH PR (#3841), not a design gap here — just flagging it for that review.

The pending-challenge fill-rate analysis in §8.9 is thorough and the nonce-only keying argument against targeted eviction is well-reasoned.

## Maintainability

The doc is exemplary in structure: the glossary, the reading-order guide in the PR description, the cross-section references (e.g. §5.4 ↔ §8.4 on Subject-in-h=), and the ASCII UX mockups that stay in sync with the actual Svelte routes. The "what the PoC got right / explicitly deferred" tables in §1 are a clean handoff from the prior work.

The `live.com` example in §7.6 (apex signed but DKIM CNAMEs into unsigned `protection.outlook.com`) is an excellent real-world catch — the kind of subtlety that would bite an implementation that naively checked only the apex DS record.

---

**Merge recommendation:** Approve. No blocking concerns. The follow-up items above are all refinements that can land in subsequent PRs or as addenda to this doc without affecting the implementation stack that's already in review.

timothyaterton (approved):
test review

aterga (approved):

PR #3838, opened 2026-05-04
feat(dnssec): scaffold verifier module + canister-side trust-anchor wiring

sea-snake:
## Summary

First PR in the email-recovery stack (`docs/ongoing/email-recovery.md` §10 Phase 0). Lands a working RFC-4035-compliant DNSSEC verifier for caller-supplied DNS proof bundles, plus the trust-anchor wiring that drives it. PR 2 (DKIM verifier) and PRs 4–9 (storage + recovery methods) build on this.

### Verifier core
- New `dnssec/` module under `src/internet_identity/src/`:
  - `types.rs` — `DnsProofBundle`, `SignedRRset`, `DelegationLink`, `Rrsig`, `DnsName`, `DnssecError`, `VerifiedRecord`.
  - `canonical.rs` — owner-name canonicalization, RR canonical form, RRSIG signed-data construction (RFC 4034 §3.1.8.1, §6.2, §6.3), DS digest input.
  - `signature.rs` — algorithm dispatch + DS digest matching (SHA-256).
  - `verify.rs` — four-step algorithm (root anchor match → chain walk → leaf RRSIG → freshness).
- Algorithm coverage (RFC 8624 MUST set):
  - **alg 8** — RSA-SHA256 (RFC 5702): root, com., most legacy zones.
  - **alg 13** — ECDSA-P256-SHA256 (RFC 6605): most TLDs, Cloudflare, modern zones.
  - **alg 15** — Ed25519 (RFC 8080): rare in production but mandatory.
  - Anything else returns `UnsupportedAlgorithm`.

### Wiring
- New `DnssecConfig` and `DnssecRootAnchor` types in `internet_identity_interface`, exposed at the top of `InternetIdentityInit` as `dnssec_config: opt opt DnssecConfig` (set/clear semantics matching `analytics_config` and `dummy_auth`).
- Trust-anchor list plumbed through `init`/`post_upgrade` into `PersistentState.dnssec_config` (and `StorablePersistentState` for cross-upgrade persistence).
- `internet_identity.did` updated.

### Tests
13 unit tests in `dnssec/` covering:
- Real cloudflare.com chain verifies end-to-end (exercises alg 8 at root and alg 13 at com → cloudflare.com → leaf).
- Empty trust-anchor list rejected with `NoTrustAnchors`.
- Wrong trust anchor rejected with `RootAnchorMismatch`.
- Flipped byte in root DNSKEY → `RootAnchorMismatch` or `BadSignature`.
[...]

PR #3879, opened 2026-05-12
feat(doh): DoH fallback for DKIM/DMARC TXT records on unsigned domains

sea-snake:
## Summary

PR 4 of the email-recovery stack. Adds the DoH (DNS-over-HTTPS) fallback path for fetching DKIM/DMARC TXT records when a domain isn't DNSSEC-signed (Gmail, Outlook, iCloud, etc.). Stacks on #3878 (DMARC alignment).

- **Allowlist-gated outcalls.** `DohConfig.allowed_domains` (deploy/upgrade arg, persisted in `PersistentState`). Domains not on the list return `DomainNotAllowed` without touching the network.
- **3-of-5 quorum across 4 jurisdictions.** Cloudflare 🇺🇸 + Google 🇺🇸 + Quad9 🇨🇭 + CIRA 🇨🇦 + IIJ 🇯🇵. Strict majority; no single jurisdiction can reach quorum on its own (asserted in unit tests).
- **In-flight dedup.** Hand-rolled `oneshot`-style `Waker` primitive collapses concurrent fetches for the same FQDN to one outcall fan-out.
- **IC-correct outcall path.** `http_request_with_closure` with a `transform` query function that reduces the wire DNS response to its TXT RDATA (drops TTL/transaction-ID drift across replicas, so consensus succeeds). `max_response_bytes = 4 KiB`, cycle budget mirrors the OIDC discovery outcall.
- **Cache hygiene.** Stale-pending TTL (`PENDING_STALE_AFTER_SECS = 120`) recovers from the IC-specific hazard where a trapped post-yield continuation leaves an InFlight entry stuck forever (state across `.await` is committed, not transactional). Lazy eviction on lookup + opportunistic sweep on publish bound the value-cache size to {unique FQDNs queried in `max_cache_age_secs`}.

Public API: `pub async fn fetch_txt(name, registered_domain) -> Result<Vec<u8>, DohError>`. Allowlist match is case-insensitive on the registered domain. Steady-state outcall load for a continuous Gmail flow at default 1h TTL: ~10 outcalls/hr per domain (2 fan-outs × 5 providers), independent of email volume.

PR 5+6 (#3880) wires this into `smtp_request` and is where the production async loop is first exercised end-to-end — see the testing note below. The DKIM/DMARC verifiers from PRs 2/3 and the DNSSEC chain from PR 1 are wired alongside it there.

## Test plan

- [x] doh unit tests: parser (round-trip, malformed, NXDOMAIN), `decide_quorum` bucketing (3-of-5/2-of-5 split-bucket/all-fail/all-malformed), dedup wakeups, stale-pending eviction wakes orphan waiters with AllProvidersFailed, late-publisher no-op after stale takeover, lazy eviction on lookup, sweep on publish, end-to-end `fetch_txt` with NotConfigured / DomainNotAllowed / case-insensitive allowlist / cache hit-then-refetch / TTL cap / quorum-failure-not-cached.
- [x] Full unit test suite green.
- [x] `cargo check --target wasm32-unknown-unknown` clean.
- [x] `cargo clippy --tests -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [x] Async outcall loop coverage lives downstream in PR 5+6 (#3880), in `src/internet_identity/tests/integration/email_recovery.rs::full_setup_flow_binds_credential_to_anchor`: drives PocketIC, fulfils all 5 DoH provider outcalls with synthesized wire-format DNS responses, asserts quorum success + status flip. Negative paths (allowlist rejection, DNSSEC-takes-precedence-so-zero-outcalls) are covered there too. This matches the convention used by the openid module (production outcall code is `cfg(not(test))`-gated; coverage of the async loop is integration-only).
- [ ] Smoke test on a real local replica: `fetch_txt` against a real allowlisted domain, verify provider fan-out + transform + consensus against live DoH servers (deferred until gateway plumbing lands).
- [ ] Verify IIJ DoH endpoint URL (`public.dns.iij.jp`) with a real outcall before deploy.

### Note on the quorum module

The original revision exposed a synchronous `run_quorum` wrapper plus a `DohFetcher` trait so unit tests could mock provider responses. That duplicated the async fan-out and gave misleading coverage (tests exercised a parallel sync code path, not the one that actually ships). Dropped in the last commit — `quorum.rs` now exposes only the pure `decide_quorum` step, and the eight quorum tests now construct `Outcome` vectors directly. End-to-end coverage of the async loop is delegated to PR 5+6 as noted above.

## PR Stack
| # | PR | Description | Status |
|---|---|---|---|
| 0 | [#3836](https://github.com/dfinity/internet-identity/pull/3836) | Design doc | Open |
| 1 | [#3838](https://github.com/dfinity/internet-identity/pull/3838) | DNSSEC verifier scaffold | Open |
| 2 | [#3877](https://github.com/dfinity/internet-identity/pull/3877) | DKIM verifier (RFC 6376) | Open |
| 3 | [#3878](https://github.com/dfinity/internet-identity/pull/3878) | DMARC alignment (RFC 7489) | Open |
| 4 | [#3879](https://github.com/dfinity/internet-identity/pull/3879) | DoH fallback | Open |
| 5+6 | [#3880](https://github.com/dfinity/internet-identity/pull/3880) | Setup flow (storage + smtp_request) | Open |
| 7 | [#3881](https://github.com/dfinity/internet-identity/pull/3881) | Recovery flow (delegation) | Open |
| 8 | [#3882](https://github.com/dfinity/internet-identity/pull/3882) | Frontend + feature flag | Open |
| 9 | [#3883](https://github.com/dfinity/internet-identity/pull/3883) | Deploy/upgrade scripts: dnssec_config + doh_config | Open |
| 10 | [#3884](https://github.com/dfinity/internet-identity/pull/3884) | Email-recovery UX overhaul | Open |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2212,6 +2212,7 @@ dependencies = [
  "captcha",
  "ed25519-dalek",
  "flate2",
+ "futures",
  "getrandom 0.2.16",
  "hex",
  "ic-canister-sig-creation",
--- a/src/internet_identity/Cargo.toml
+++ b/src/internet_identity/Cargo.toml
@@ -36,6 +36,12 @@ captcha = { git = "https://github.com/dfinity/captcha", rev = "9c0d2dd9bf519e255
 identity_jose = { git = "https://github.com/dfinity/identity.rs.git", rev = "aa510ef7f441848d6c78058fe51ad4ad1d9bd5d8", default-features = false }
 url = { version = "2.5", default-features = false, features = ["std"] }
 
+# DoH fallback fan-out: pulled in only for `future::join_all`, which
+# lets us await all 5 provider outcalls concurrently rather than
+# serially. `default-features = false` keeps the wasm32 build lean
+# (no executor / threading code).
+futures = { version = "0.3", default-features = false, features = ["alloc"] }
+
 # All IC deps
 candid.workspace = true
 ic-cdk = { workspace = true, features = ["transform-closure"] }
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -309,6 +309,11 @@ type InternetIdentityInit = record {
     // stored value across an upgrade, `opt null` clears it, `opt opt c`
     // sets it to `c`.
     dnssec_config : opt opt DnssecConfig;
+    // DoH (DNS-over-HTTPS) fallback configuration. Allowlists the
+    // domains for which the canister may fetch DKIM/DMARC TXT records
+    // via HTTP outcalls when no DNSSEC chain is available — see
+    // `docs/ongoing/email-recovery.md` §7.6. Same set/clear pattern.
+    doh_config : opt opt DohConfig;
 };
 
 // DNSSEC trust-anchor list. Any feature that needs DNSSEC-verified DNS
@@ -328,6 +333,17 @@ type DnssecRootAnchor = record {
     algorithm : nat8;
     digest_type : nat8;
     digest : blob;
+};
+
+// DoH (DNS-over-HTTPS) fallback configuration.
+//
+// The canister will only fetch DKIM/DMARC TXT records via HTTP outcalls
+// for an FQDN whose registered domain is in `allowed_domains`. Cache
+// entries are populated on demand and re-used until `max_cache_age_secs`
+// elapses (default 3600s when null, capped at 24h).
+type DohConfig = record {
+    allowed_domains : vec text;
+    max_cache_age_secs : opt nat64;
 };
 
 type ChallengeKey = text;
--- a/src/internet_identity/src/doh/cache.rs
+++ b/src/internet_identity/src/doh/cache.rs
@@ -1,0 +1,548 @@
+//! In-memory DoH cache with concurrent-fetch deduplication.
+//!
+//! Two responsibilities:
+//!
+//! 1. **Cache** — keep a fetched TXT record around for as long as the
+//!    deploy-arg `max_cache_age_secs` (default 1 h, capped at 24 h) so
+//!    a flurry of mail from the same domain doesn't re-fetch on every
+//!    message. The cache is heap-only — losing it on upgrade is fine,
+//!    we just refetch the next time mail arrives.
+//!
+//! 2. **Dedup** — when several concurrent `smtp_request` calls arrive
+//!    for the same domain and the cache is cold, all of them share a
+//!    single underlying outcall fan-out. The first arrival fires the
+//!    outcalls and writes the result; subsequent arrivals register a
+//!    `Waker` against the in-flight entry and `await` the publication.
+//!
+//! The dedup mechanism is a hand-rolled `oneshot`-style primitive
+//! (see [`Pending`]) — `ic-cdk` doesn't ship with the tokio sync
+//! primitives we'd otherwise reach for, but the standard Rust `Waker`
+//! machinery is enough since the canister is single-threaded inside
+//! one wasm execution.
+
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::future::Future;
+use std::pin::Pin;
+use std::rc::Rc;
+use std::task::{Context, Poll, Waker};
+
+use super::types::DohError;
+
+/// How long a `Pending` entry can sit in the cache before a new
+/// arrival treats it as abandoned.
+///
+/// The IC doesn't roll back state across `.await` points: state
+/// changes that happened *before* the yield are committed even if the
+/// resumed-message continuation traps. So if the post-outcall code
+/// path traps, the `Pending` entry inserted by the first arrival
+/// stays in the map and every subsequent caller would `Wait` on a
+/// `PendingState::InFlight` that nothing will ever resolve — until
+/// canister upgrade.
+///
+/// 120 s is comfortably longer than any plausible HTTP-outcall round
+/// trip (the IC caps individual outcalls at ~5 minutes, but successful
+/// ones complete in seconds; a 2-minute floor on "looks abandoned" is
+/// safe for live fetches and quick enough that genuinely stuck entries
+/// clear within a couple of email retries).
+pub const PENDING_STALE_AFTER_SECS: u64 = 120;
+
+/// One cache entry — bytes and the wall-clock time at which they go
+/// stale.
+#[derive(Clone, Debug)]
+pub struct CacheEntry {
+    pub bytes: Vec<u8>,
+    pub expires_at_secs: u64,
+}
+
+/// Shared state used by the dedup primitive. `Rc<RefCell<...>>` because
+/// the canister's wasm execution is single-threaded; Send/Sync don't
+/// apply.
+type SharedPending = Rc<RefCell<PendingState>>;
+
+#[derive(Debug)]
+enum PendingState {
+    /// The first arrival is mid-fetch. Each subsequent arrival has
+    /// registered its `Waker` here and will be re-polled when the
+    /// fetch publishes a result.
+    InFlight { wakers: Vec<Waker> },
+    /// The fetch has completed. Late arrivals see this and pick up
+    /// the result without waiting.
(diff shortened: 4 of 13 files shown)


aterga:
**PR Overview (pragmatic lens)**  
This is a clean, well-scoped addition: a DNS-over-HTTPS (DoH) fallback path for fetching DKIM/DMARC TXT records on domains that lack DNSSEC (the vast majority of consumer email providers). It is gated, resilient, IC-native, and low-overhead. The design choices (allowlist, 3-of-5 cross-jurisdiction quorum, transform-based deterministic outcalls, in-flight deduplication, and cache-with-stale-pending hygiene) directly address the real risks of outcalls on the Internet Computer.  

43 new unit tests + clean `cargo check`/`clippy`/`fmt` + explicit test plan for local-replica smoke test in PR 8 make this low-risk to merge once the small items below are addressed. No breaking changes; everything is additive and configuration-gated.

### Blockers (must-fix before merge)
None identified. The security posture is already stronger than many production outcall patterns in the IC ecosystem.

### Major Security Gaps
None.  
- **Outcall surface** is strictly allowlist-gated on the *registered domain* (case-insensitive) before any network activity → `DomainNotAllowed` is returned with zero cycles spent.  
- **Quorum** (3-of-5 across Cloudflare 🇺🇸, Google 🇺🇸, Quad9 🇨🇭, CIRA 🇨🇦, IIJ 🇯🇵) plus the unit-test assertion that “no single jurisdiction can reach quorum on its own” eliminates single-provider or single-country poisoning.  
- **Response canonicalization** via `http_request_with_closure` + transform query function (drops TTL/transaction-ID variance) guarantees deterministic consensus.  
- **Cache hygiene** explicitly mitigates the IC-specific “trapped continuation after yield” hazard (state is committed, not rolled back).  
- **Parsing/validation** is defensive (invalid DNS names rejected, label-anchored suffix match between queried name and registered domain).  

This is materially safer than a naïve `http_request` to arbitrary DoH endpoints.

### Easy-to-Change Improvements (quick wins, <30 min each)
1. **Expose the provider list and DoH URLs as constants with comments** (in `doh/mod.rs` or `types.rs`).  
   - Action: Add a `const PROVIDERS: &[(&str, &str)]` (name + URL) and a short comment explaining the jurisdiction distribution and why two US providers are acceptable given the quorum rule.  
   - Why: Reviewers/deployers will want to know exactly which endpoints are called; makes future rotation trivial.

2. **Make `PENDING_STALE_AFTER_SECS` (120 s) a named const at the top of `cache.rs` with a one-line comment referencing the IC yield hazard**.  
   - Also add a matching integration-test case that forces a continuation trap (or mocks it) to prove eviction works.

3. **Document the exact steady-state outcall cost** in the PR description / module docs.  
   - You already have it internally (“~10 outcalls/hr per domain at 1 h TTL”). Surface it as a comment on `fetch_txt` so operators know the cycle budget impact before enabling email recovery at scale.

4. **Add one explicit test for the allowlist edge case** where the registered domain matches but the queried name does *not* suffix-match (should be rejected before outcall).  
   - The defensive validation is mentioned; just make the test visible so future refactors can’t regress it.

5. **Tiny Cargo.toml hygiene** (optional but nice): the new deps (`futures` alloc-only, `url`, `identity_jose`) are all justified by the stack, but add a one-line comment next to each explaining why it is pulled in by *this* PR (avoids “why is this here?” questions later).

### Strategic Recommendations
1. **Rate-limit / back-pressure story** (post-merge, before PR 8 wiring).  
   Even though per-domain load is tiny, a popular anchor tenant using Gmail/Outlook could still generate bursts. Consider a simple global token-bucket (or just a `last_outcall` timestamp + cooldown) in a follow-up PR so one misbehaving recovery flow cannot starve the rest of the canister.

2. **Observability hooks**.  
   Expose a `DohMetrics` struct (or just increment counters via `ic_cdk::api::call::call_perform` counters if available) for:  
   - cache hit/miss  
   - quorum success/failure (and which providers disagreed)  
   - allowlist denials  
   This will be gold when email recovery goes live.

3. **Configuration defaults & upgrade safety**.  
   The `DohConfig` fields are optional in both upgrade arg and persistent state — perfect. In the next config PR (or in `main.rs` init), consider a sensible default allowlist (`vec!["gmail.com".to_string(), "outlook.com".to_string(), ...]`) or at least document the empty-list behaviour clearly so mainnet upgrades don’t silently disable the fallback.

4. **Long-term: consider a hybrid DNSSEC + DoH path** (already planned in the stack).  
   Once DNSSEC path (PR 1) is wired, the DoH path becomes the “unsigned” branch. Document the decision matrix (DNSSEC → DoH → fail) in one place so the full email-recovery verification story is easy to audit.

5. **Testing strategy for the full stack**.  
   When PR 8 wires this in, add at least one end-to-end test that uses a real Gmail/Outlook test account (with known DKIM/DMARC records) against the local replica. The current unit-test coverage is excellent for the DoH layer; the integration surface is where subtle bugs hide.

**Overall verdict**: Ship after the five easy items above (mostly documentation + one test). This is a high-quality, defensively-written piece of infrastructure that materially unblocks email recovery for the most common real-world domains. Great work.

aterga (approved):
Follow-up suggestion:

Please avoid referring to PRs in the code docs, e.g., here: https://github.com/dfinity/internet-identity/pull/3879/changes#diff-d6fcf251670a60f497657a6fade596d1d18ca71a7882709cf8476c8fa2a406bfR3

sea-snake:
Thanks for the careful review @aterga. Quick map of the items vs what I'm doing:

**Easy-to-Change Improvements**

1. *Expose provider list and URLs as constants* — already the case: `PROVIDERS: &[DohProvider]` lives in `src/internet_identity/src/doh/types.rs` with the jurisdiction breakdown documented alongside it.
2. *Make `PENDING_STALE_AFTER_SECS` an explicit named const + add an integration test for the trapped-continuation eviction path* — declining. The constant is already named at the top of `cache.rs` (the value is exposed; the comment references the IC yield hazard inline). I'd rather not add a forced-trap test purely for clarity — the eviction path is already covered by unit tests against the cache module directly, and integration coverage of the broader flow lands in the next PR.
3. *Document steady-state outcall cost in the module docs* — fair, the number already in the PR description can move into a `fetch_txt` doc comment in a follow-up if it becomes a recurring question.
4. *Test for allowlist-matched-but-name-mismatched edge* — the defense-in-depth check (`name_within_domain`) is unit-tested in `mod.rs`; happy to add an explicit case in a follow-up if reviewers want it called out by name.
5. *Per-dep comments in `Cargo.toml`* — the new deps (`futures`, `url`, `identity_jose`) are already justified by surrounding context (futures: outcall fan-out; url: candidate URL validation; identity_jose: shared with the DKIM/DMARC stack). Skipping the inline comments — those tend to rot faster than the deps themselves.

**Strategic Recommendations** — declining the lot as out of scope for this PR:

1. *Rate-limit / back-pressure* — explicitly not adding. The canister has no per-tenant data to rate-limit on (the outcalls are keyed on an allowlist controlled by the deploy arg, not on caller-supplied state), and per-anchor rate limiting belongs at the gateway / smtp_request layer, not in the DoH module. The 3-of-5 fan-out + cache TTL already bounds steady-state outcall volume.
2. *Observability / `DohMetrics`* — not in scope; if we want canister-side metrics, that's a separate, cross-cutting piece of work.
3. *Default allowlist baked in* — deliberately not baking a default. Operators set the allowlist in the install arg; an empty allowlist returning `DomainNotAllowed` is the safe default. Documented in PR 9 (#3883).
4. *Hybrid DNSSEC + DoH decision-matrix doc* — covered by the design doc (#3836) and the wiring in PR 5+6 (#3880); won't duplicate it here.
5. *Real Gmail/Outlook end-to-end tests* — orthogonal; live-domain tests don't belong on the DoH unit, they belong on the full email-recovery flow, where they're already planned.

**Follow-up suggestion** (no PR refs in code docs): noted, but treating as out of scope for this PR — happy to address as a standalone cleanup later if it builds up.

PR #3895, opened 2026-05-16
build(ci): cache rustc invocations across builds via sccache

aterga:
## Summary

Wraps `rustc` with [sccache](https://github.com/mozilla/sccache) inside the Docker build, persisted across CI runs via BuildKit cache mounts and GHA cache. Unchanged compilation units are served from cache instead of being recompiled.

## Why

Today, every push to a PR recompiles the entire `internet_identity` crate from scratch in `docker-build-internet_identity`. With recent feature work that adds material new code to the backend (e.g. the DKIM verifier in #3877, ~2,800 LOC), the backend Docker build is the longest pole — frequently 25–35+ min — and downstream `canister-tests` jobs wait on its artifact before they can start.

sccache is a transparent `rustc` wrapper: it stores rustc outputs in a content-addressed cache keyed by source + flags + rustc version. When inputs match, it returns the cached object bytes; otherwise it falls through to real rustc.

## Reproducibility

This is intentionally a no-op for the build output:

- `-j1` is preserved.
- `--remap-path-prefix`, pinned rustc, pinned Docker image, `lto = true`, `opt-level = 'z'`, `gzip --no-name` — all preserved.
- sccache returns byte-identical rustc outputs for matching inputs (Mozilla, ChromeOS, and many others rely on this property in production).
- Crates with non-deterministic build scripts are silently skipped by sccache (it falls through to real rustc), so output is always correct even on misses.

**Verification before un-drafting / merging**: compare the `sha256sum internet_identity.wasm.gz` printed in this PR's CI to the most recent `main` build of the same source commit. If hashes match, sccache preserves reproducibility on the II codepath as designed. If they don't, this PR should not merge.

## What changed

- **`Dockerfile`**:
  - Install pinned sccache v0.15.0 (musl prebuilt, no compile required).
  - Set `RUSTC_WRAPPER=sccache`, `SCCACHE_DIR=/sccache-cache`.
  - Add a shared cache mount (`id=ii-sccache`) to every RUN that compiles Rust: deps pre-build, `build_internet_identity`, `build_archive`, `build_internet_identity_frontend`. Print `sccache --show-stats` after each for visibility.
- **`.github/workflows/canister-tests.yml`**:
  - Each of the three downstream `docker-build-*` jobs gets its own GHA cache scope (`ii-build-backend`, `ii-build-archive`, `ii-build-frontend`) with `mode=max` (required to export BuildKit cache mounts). Each pulls from both its own scope and the shared `cached-stage` base.

## Expected impact

| Scenario | Today | With sccache |
| --- | --- | --- |
| 1st build after Cargo.lock change | ~21 min deps + ~30 min backend | Similar deps build (caches cold), similar 1st backend |
| 2nd+ push on same PR, few files changed | Full backend recompile (~30 min) | Backend drops to a few minutes (only changed CGUs recompile) |
| Steady-state across PRs | Cold each time | Warm — most CGUs hit |

## Costs / caveats

- GHA Actions cache is 10 GB/repo with LRU eviction. We add three new scopes; under cache pressure, eviction churn may reduce hit rate. Cache hit rate is visible in `sccache --show-stats` output in each build's logs — worth watching for the first few PRs after merge.
- BuildKit cache mounts and `cache-to: mode=max` only work with BuildKit ≥ 0.10 (provided by `docker/setup-buildx-action@v3`, already in use).
- `-j1` is still the bottleneck on cold compilation. This PR doesn't touch that — happy to follow up with a separate experiment validating whether modern rustc produces bit-identical output under `-j$(nproc)` before relaxing it.

## Test plan

- [ ] CI green on this PR.
- [ ] Compare `internet_identity_backend.wasm.gz` sha256 against latest `main` build of the same commit → confirm bit-identical.
- [ ] Push a trivial 1-line change to this branch → confirm backend build time drops materially on the 2nd run vs the 1st.
- [ ] Inspect `sccache --show-stats` output in the backend job's log to confirm meaningful hit rate.

### DIFF
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -25,7 +25,38 @@ jobs:
       # more information, see
       # https://github.com/docker/build-push-action/issues/539
       - name: Set up docker buildx
+        id: setup-buildx
         uses: docker/setup-buildx-action@v3
+
+      # The GHA cache backend (type=gha on docker/build-push-action) does NOT
+      # persist BuildKit cache-mount contents across runs — only image layers.
+      # To persist the sccache directory (where compiled rustc outputs live),
+      # we use actions/cache for a host directory and buildkit-cache-dance to
+      # inject it into BuildKit's cache mount before the build / extract it
+      # back after. Done here on the base job because that's where the deps
+      # compile happens (~335 rustc invocations); downstream jobs only compile
+      # their canister crate on top of the pre-built deps layer and don't need
+      # their own persisted sccache.
+      - name: Restore sccache directory
+        id: sccache-cache
+        uses: actions/cache@v4
+        with:
+          path: sccache-cache
+          key: ii-sccache-base-${{ runner.os }}-${{ github.run_id }}-${{ github.run_attempt }}
+          restore-keys: |
+            ii-sccache-base-${{ runner.os }}-
+
+      - name: Inject sccache into BuildKit cache mount
+        uses: reproducible-containers/buildkit-cache-dance@v3
+        with:
+          builder: ${{ steps.setup-buildx.outputs.name }}
+          cache-map: |
+            {
+              "sccache-cache": {
+                "target": "/sccache-cache",
+                "id": "ii-sccache"
+              }
+            }
 
       - name: Build base Docker image
         uses: docker/build-push-action@v5
@@ -64,13 +95,12 @@ jobs:
           file: Dockerfile
           build-args: |
             II_VERSION=${{ steps.version.outputs.version }}
-          # Pull deps-stage layers from the shared base cache, and this job's
-          # own sccache cache mount from prior runs. Push only the sccache
-          # additions back so we don't pollute the base scope.
-          cache-from: |
-            type=gha,scope=cached-stage
-            type=gha,scope=ii-build-backend
-          cache-to: type=gha,scope=ii-build-backend,mode=max
+          # Reuses the deps layer from the base build (compiled deps live in
+          # /cargo_target). The canister-specific compilation here is small;
+          # adding a persisted sccache scope for it didn't pay for itself
+          # (mount contents aren't preserved by type=gha — see comment on
+          # docker-build-base for the persistence mechanism).
+          cache-from: type=gha,scope=cached-stage
           # Exports the artefacts from the final stage
           outputs: ./out
           target: scratch_internet_identity
--- a/.github/workflows/canister-tests.yml
+++ b/.github/workflows/canister-tests.yml
@@ -64,7 +64,13 @@ jobs:
           file: Dockerfile
           build-args: |
             II_VERSION=${{ steps.version.outputs.version }}
-          cache-from: type=gha,scope=cached-stage
+          # Pull deps-stage layers from the shared base cache, and this job's
+          # own sccache cache mount from prior runs. Push only the sccache
+          # additions back so we don't pollute the base scope.
+          cache-from: |
+            type=gha,scope=cached-stage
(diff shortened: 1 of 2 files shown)


aterga:
## Closing — full writeup of findings

Closing this PR. The approach proved fundamentally limited in this repo's CI architecture; documenting findings here so the work isn't lost.

### What we tried

**Iteration 1** ([d56484f](https://github.com/dfinity/internet-identity/pull/3895/commits/d56484f)): sccache via Dockerfile + BuildKit cache mounts + GHA cache backend, with per-job cache scopes (`ii-build-backend`, etc.).

**Iteration 2** ([1170f53](https://github.com/dfinity/internet-identity/pull/3895/commits/1170f53)): dropped per-job scopes from archive/frontend after observing they caused BuildKit to re-execute the deps stage instead of reusing it from `cached-stage`.

**Iteration 3** ([94e62cc](https://github.com/dfinity/internet-identity/pull/3895/commits/94e62cc)): switched to `actions/cache` + [`reproducible-containers/buildkit-cache-dance`](https://github.com/reproducible-containers/buildkit-cache-dance) for the base job to actually persist the sccache directory across runs (the GHA cache backend on `docker/build-push-action` exports image layers but **not** BuildKit cache-mount contents — that's a known limitation, hence cache-dance).

### What worked

Reproducibility was preserved at every step — every run produced a wasm.gz with sha256 `32ae2365b692e8648731cc0d621773904f191f407a6fdaccad9885e2fe317942`, bit-identical to `main`'s build of the same commit.

The cache-dance integration itself works as designed. Warm-cache run on `docker-build-base` (run [25968749018](https://github.com/dfinity/internet-identity/actions/runs/25968749018)):

```
Compile requests:    442
Cache hits:          335   <-- 100% hit rate on rust compilations
Cache misses:          0
#34 DONE 47.8s   (was ~155s cold)
```

### What didn't work

The `--mount=type=cache` on the deps prebuild RUN makes that layer **flaky-cacheable across concurrent downstream jobs**. In every run, one or two of `docker-build-{archive,internet_identity,internet_identity_frontend}` would re-execute the full ~150s deps compile instead of reusing the layer that `docker-build-base` just built. Which job won the race varied non-deterministically between runs (archive cached in run 3 but re-executed in run 4; backend was uncached in both). This is a BuildKit/GHA-cache interaction issue that's independent of sccache and that I couldn't find a clean fix for.

Net effect on wall-clock CI time vs `main`: somewhere between break-even and a ~2 min regression depending on which downstream jobs win the race. The ~107s base-job saving from warm sccache gets eaten by 1–2× ~150s downstream re-executions.

### Why sccache isn't the right tool here anyway

For the original concern that triggered this work — PR #3877's slow backend build after adding ~2,800 LOC of DKIM code — sccache wouldn't have helped much. Cargo compiles the `internet_identity` crate in a single `rustc` invocation. sccache hashes by source+flags+rustc, so any change to the crate's source = full miss. sccache helps cross-PR iteration on **unchanged code**, not the case where you're actively adding code to a single crate.

The bigger CI-time wins are likely elsewhere:

1. **`-j1` → `-j$(nproc)` in `scripts/build`** with a reproducibility-assertion CI job to verify modern rustc produces deterministic output under `-jN`. If it does, this could 2–3x cold compilation throughput.
2. **`lto = true` → `lto = "thin"`** — typically near-identical wasm size, 2–4x faster link step. Drop-in Cargo.toml change with a wasm-size guard.
3. **Frontend asset build dedup** — `docker-build-internet_identity` and `docker-build-internet_identity_frontend` each run `npm ci && npm run build` independently. Could share via a fan-out stage.

I'll open follow-ups for these as separate, smaller PRs.

PR #3903, opened 2026-05-17
fix(be): distinguish DoH 'no record' from transient outage

aterga:
## Problem

The DMARC verifier is supposed to fall back to strict alignment when a domain publishes no `_dmarc` TXT (RFC 7489 §6.3 — a valid DNS state meaning "no policy"). But every DMARC DoH fetch error was being collapsed to `None` via `.ok()`, which silently turned a transient DoH outage into "no DMARC policy" — quietly tightening the check and breaking otherwise-legitimate setup flows.

The verifier had no way to express "the quorum reported authoritatively that no record exists" as a distinct outcome from "we couldn't reach a verdict at all". That distinction is what the DMARC fallback needs.

## Changes

Add `DohError::NoAnswer` so the DoH stack reports "no record exists" distinctly from the existing `AllProvidersFailed` / `QuorumFailed`:

- **Parser**: RCODE=3 (NXDOMAIN) and RCODE=0 empty-answer (NODATA) both surface as `ParseError::NoAnswer`. Other RCODEs (SERVFAIL, REFUSED) stay distinct since they're transient or operator failures, not authoritative answers.
- **Transform**: emits HTTP status 204 ("No Content") for NoAnswer, mirroring the existing 422-for-parse-error design that signals failure modes out-of-band of the TXT body (an in-band sentinel could collide with a legitimate TXT).
- **Outcall + quorum**: status 204 → `Outcome::NoAnswer`. The quorum step counts a NoAnswer bucket alongside TXT buckets; 3-of-5 agreement on "no record" yields `DohError::NoAnswer`. TXT majorities beat NoAnswer minorities and vice-versa.
- **Email-recovery**: `verify_setup_email_doh` only takes the strict-alignment fallback for `DohError::NoAnswer`; every other DMARC fetch error propagates as `DohFetchFailed`. A DKIM-side NoAnswer (the signed selector's public key is gone — key rotation or forged `s=`) surfaces as `EmailVerificationFailed`.

NoAnswer is **not** cached — a record can be published at any moment, so the next setup attempt re-fetches.

## Tests

- 6 new quorum tests cover NoAnswer scenarios (3-of-5 NoAnswer, 5-of-5 NoAnswer, NoAnswer minority, NoAnswer majority over TXT minority, split, etc.).
- New parser test pins NXDOMAIN → NoAnswer and SERVFAIL/REFUSED → `BadResponseCode`.
- 2 new `fetch_txt` end-to-end tests through the mock fetcher (quorum NoAnswer + cache-not-poisoned-by-NoAnswer).


src/internet_identity/src/doh/mod.rs
```diff
@@ -259,12 +262,16 @@ mod prod {
...
+                    Outcome::Txt(response.body)
+                } else if status_u16 == HTTP_STATUS_TRANSFORM_NO_ANSWER {
+                    Outcome::NoAnswer
+                } else {
+                    Outcome::FetchError(format!("HTTP {status_u16}"))
                 }
-                Ok(response.body)
             }
-            Err((_, err)) => Err(err),
+            Err((_, err)) => Outcome::FetchError(err),
```
sea-snake: Why was the u16 mapping added here?

sea-snake (approved):

src/internet_identity/src/doh/mod.rs
```diff
@@ -215,6 +209,15 @@ mod prod {
...
     pub(super) const HTTP_STATUS_OK: u8 = 200;
 
+    /// HTTP status the transform returns when the wire-format DNS
+    /// response was a valid "no such record" answer (NXDOMAIN or an
+    /// RCODE=0 empty answer section). 204 ("No Content") is a
+    /// defensible bucket for "the request succeeded but there's
+    /// nothing to return", and [`outcall`] maps it to
+    /// [`Outcome::NoAnswer`] so the quorum can count "no record"
+    /// agreement separately from transient outages.
+    pub(super) const HTTP_STATUS_TRANSFORM_NO_ANSWER: u16 = 204;
```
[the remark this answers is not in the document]
aterga: Real concern — fixed in a2fe4f27. The transform now collapses every upstream non-200 (including a counterfeit 204) onto a separate `HTTP 502` sentinel via a pure `classify_upstream` helper, so `Outcome::NoAnswer` can only originate from our own transform's NoAnswer branch. Added a regression test (`classify_upstream_spoofed_204_is_remapped_to_upstream_error`) plus full classifier branch coverage.

src/internet_identity/src/doh/mod.rs
aterga: Good question — the conversion was there because I was matching against multiple typed sentinels (some `u8`, some `u16`) and wanted uniform comparisons. Dropped it in a2fe4f27: the per-call mapping is gone; the candid::Nat ↔ u16 conversion now lives once inside `transform_doh` / `outcall`, right where the IC management-canister boundary is, and the downstream classification is on plain `u16`.

PR #3907, opened 2026-05-18
docs(ongoing): design doc for production-ready OpenID & SSO

sea-snake:
📄 **Read the design doc directly:** [`docs/ongoing/openid-sso-prod-readiness.md` on the fork](https://github.com/sea-snake/internet-identity/blob/design/openid-sso-prod/docs/ongoing/openid-sso-prod-readiness.md)

---

## Summary

Adds `docs/ongoing/openid-sso-prod-readiness.md` — a design doc for putting the OpenID/SSO stack on a production footing so the canary allowlist can be lifted.

The doc is problem-driven: it identifies the five concrete things the allowlist masks today, and proposes a solution for each:

1. **Stateful provider registry** (§4) — replace `Vec<Box<dyn OpenIdProvider>>` + parallel `DISCOVERY_TASKS` with a data-only `CONFIG_REGISTRY` and a free-function `verify_jwt`. Trait dispatch goes away; per-provider differences become enum-tagged config fields.
2. **Timer-driven outcalls** (§5) — replace `init_discovery_timers` and `schedule_fetch_certs` with an on-demand LRU cache (1000 entries per cache, 1 h TTL). Concurrent-fetch dedup uses the same `Pending`/`Waker` primitive as the DoH cache in #3841, copy-pasted rather than lifted.
3. **HTTP outcall safety** (§6) — `max_response_bytes` caps (8 KB hop-1, 64 KB hop-2, 64 KB JWKS), transform validation before parse, tightened per-call cycle ceilings.
4. **Fragment callback** (§7) — switch `response_mode=fragment` → `response_mode=form_post`. New canister POST handler (update mode for response certification) parses the form body, returns certified HTML that delivers `{id_token, state}` via BroadcastChannel (popup) or sessionStorage (same-tab). Apple Sign In returning `name`/`email` only under form_post is the operational driver. Discussion in §7.4 of why the canister can't consume the JWT itself.
5. **Allowlist removal** (§8) — drop `add_discoverable_oidc_config` and the persisted `OIDC_CONFIGS` list. Add `discover_sso(domain) -> DiscoverySsoResult`, anonymous, cache-backed. Thread `discovery_domain: opt text` through the four JWT-consuming API methods. Stamp `sso_domain` on `OpenIdCredential` at creation; lazy backfill for existing credentials.

The salt + nonce + `caller()` binding that secures the JWT against compromised transports stays end-to-end identical. §3 walks through why several otherwise-attractive simplifications (consume the JWT in the POST, 302-translate to fragment, etc.) were considered and rejected.

§9 lays out the rollout: tracks A (form_post), B (outcall safety), C1→C2→C3→C4 (cache stack) can run in parallel; only within Track C is there a stack.

## Reading order

- §1: the five problems framed.
- §2-§3: scope + threat model.
- §4-§8: one section per problem, each with what's broken today + what we're changing.
- §9: dependency layout and suggested PR sequence.
- §10: open questions.

Per request, this PR is the design doc only — no code changes.

cc @aterga


docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+2. The `salt` — never on the wire except inside signed ingress messages.
+3. The session_sk corresponding to `session_principal` — never leaves the FE; required for the ingress signature so the canister sees `caller() == session_principal`.
+
+Without all three, no actor (boundary node, MITM, replicating outcall transport, …) can redeem a JWT against an anchor that is not their own. This is the load-bearing security property.
+
+### 3.2 What changes under this design
+
+| Change | Effect on trust model |
+| ------ | --------------------- |
+| Stateless verifier + cache (§4, §5) | None. The verifier still runs `SHA256(salt | caller()) == jwt.nonce` against `caller()` from a signed ingress message. The cache holds JWKS + discovery results, not JWTs or salts. |
```
aterga: MD rendering is broken on this line: 

Image

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+4. **Cache-miss flood from anonymous `discover_sso` calls.** Bounded by (a) the per-outcall cycle cost cap, (b) LRU eviction past the cache size limit (older entries fall out under sustained pressure, sustaining repeat work for the attacker), and (c) standard replicated-outcall pricing on the IC. The doc does not propose additional per-caller rate limiting; if that becomes necessary, it is added at the gating layer, not the architecture layer.
+5. **Replay of a stolen `(jwt, salt)` pair from a different session.** Already defended today: the JWT's `nonce` was computed with the original session's principal, so `caller()` from a different session won't match.
+
+### 3.4 Attacks explicitly **not** defended
+
+- Compromise of the IdP's signing key. Every credential issued by that IdP is at risk; that is the unavoidable trust assumption of any OIDC integration.
+- Compromise of the user's device / browser that holds the session_sk. Standard endpoint compromise — out of scope.
+- Phishing the user into completing an OAuth flow on a malicious site impersonating Internet Identity. This is the standard "website phishing" problem; we don't try to solve it in this layer.
```
aterga: Isn't this mitigated by the fact that the SSO provider needs to add a well-known configuration fiel specifically for II?

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+- whether `discovery_domain` is `Some` or `None`.
+
+These are data, not behavior. The `verify()` body is essentially the same in both `Provider::verify` and `DiscoverableProvider::verify` — same JWT decode, same claims parse, same nonce check, same JWKS lookup, same signature verify, same credential build. The "trait" is paying for differences that are at most a few `match` arms.
+
+### 4.3 Solution: data-only registry, free-function verify
+
+Collapse the three collections into one:
+
+```rust
+pub struct DirectProviderConfig {
```
aterga, line 168:
```suggestion
pub struct DirectSsoProviderConfig {
```

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+    issuer: String,                              // may contain {placeholders}
+    client_id: String,
+    jwks_uri: String,
+    email_verification: Option<OpenIdEmailVerificationScheme>,
+}
+
+thread_local! {
+    static CONFIG_REGISTRY: RefCell<Vec<DirectProviderConfig>> = RefCell::new(vec![]);
```
aterga, line 176:
```suggestion
    static SSO_CONFIG_REGISTRY: RefCell<Vec<DirectSsoProviderConfig>> = RefCell::new(vec![]);
```

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+
+Verification becomes a free function over the registry + the cache:
+
+```rust
+pub fn verify_jwt(
+    jwt: &str,
+    salt: &[u8; 32],
+    discovery_domain: Option<&str>,
+    cache: &mut OidcCache,
+) -> Result<OpenIdCredential, OpenIDJWTVerificationError> {
```
aterga, line 190: The planned caching design isn't clear from these comments; e.g., how many caches do we have? 

I recommend outlining the lifecycle for a particular SSO <> II integration (first time it's a cache miss, second time it's a hit, third time it's a miss due to cache eviction, etc.)

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+pub struct BoundedLruDedupCache<K, V> {
+    entries:  HashMap<K, CacheEntry<V>>,
+    lru:      VecDeque<K>,        // most-recently-used at the back
+    capacity: usize,              // 1000 (§5.4)
+}
+
+enum CacheEntry<V> {
+    /// Fetched and currently usable.
+    Done { value: V, expires_at_secs: u64 },
+    /// First caller is fetching; concurrent arrivals register their `Waker`
```
aterga: Can we omit this new terminology?

At the very least, please clearly define all the terms, e.g., "waker"

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+1000 entries per cache. Each entry covers one SSO domain (or one `jwks_uri`) across every user of that SSO — the cache is global, not per-anchor — so 1000 is the simultaneous-hot-SSO cap, not the user count. At ~5 KB per entry that's ~10 MB across both caches, vs. the canister's 3 GB heap budget. LRU eviction takes care of the long tail of rarely-used SSOs.
+
+```rust
+pub const DISCOVERY_CACHE_CAPACITY: usize = 1000;
+pub const JWKS_CACHE_CAPACITY:      usize = 1000;
+```
+
+Entry TTL: 1 hour for both. Matches the upstream OIDC discovery doc cache headers we've observed, and is well under the JWKS rotation window of every mainstream IdP (Google rotates ~every 4–6 h; Microsoft daily; Apple monthly). On TTL lapse, the next verifier call observes the entry as expired and re-fetches inline — same code path as a cold start.
+
+Numbers tunable; they live in two consts in the cache module:
```
aterga: nit. a bit repetitive?

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+
+```rust
+pub const DISCOVERY_CACHE_CAPACITY: usize = 1000;
+pub const JWKS_CACHE_CAPACITY:      usize = 1000;
+```
+
+Entry TTL: 1 hour for both. Matches the upstream OIDC discovery doc cache headers we've observed, and matches the JWKS rotation window of every mainstream IdP (Google rotates ~every 4–6 h; Microsoft daily; Apple monthly). On TTL lapse, the next verifier call observes a cache miss and re-fetches — same code path as a cold start.
```
aterga: Refer to the section where it's explained how the TTL is implemented.

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+- The "discovery still pending" branches in `Provider::verify` / `DiscoverableProvider::verify` — replaced by the cache's `Wait` arm, which an async verifier `await`s through.
+- `Rc<RefCell<Option<String>>>` plumbing on `DiscoverableProvider` — replaced by ordinary cache reads.
+
+### 5.6 Cycle-budget implications
+
+Today's worst case: ~30 Gcycles per outcall × ~6 outcalls/h × N providers = ~180N Gcycles/h, ongoing forever.
+
+Proposed worst case at steady state with a hot cache: ~0 cycles/h (no scheduled outcalls). Per-sign-in cost: ~30 Gcycles on cache miss (cold), ~0 on cache hit (warm). Multiplied by sign-in volume rather than provider count.
+
+For DoS analysis (anonymous `discover_sso` flood): N attacker calls × ~60 Gcycles per cold path (hop-1 + hop-2 + JWKS fetch) = ~60N Gcycles. Bounded by per-call cycle accounting (§6) and the IC's existing reject-on-low-cycles behaviour.
```
aterga: > reject-on-low-cycles behaviour

I don't think this is applicable to II, let's make it clearer what DoS risks really are on a system subnet. 

Hint: We could rely on the possibility to rate limit via boundary nodes, if necessary.

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+**Per-call cycle ceiling.** Today, `CERTS_CALL_CYCLES = DISCOVERY_CALL_CYCLES = 30 Gcycles`, attached to each `http_request_with_closure` call. With caps in place, the actual ceiling is bounded by `(max_response_bytes × replication_factor × bytes_per_cycle)`. Concretely:
+- Hop 1: 8 KB × 13 × ~400 cycles/byte ≈ 42 Mcycles. Allocate 1 Gcycle (large safety margin).
+- Hop 2: 64 KB × 13 × ~400 ≈ 330 Mcycles. Allocate 5 Gcycles.
+- JWKS: same as hop 2 — 5 Gcycles.
+
+This is a tightening of today's blanket 30 G allocation, so a cache-miss flood costs the attacker more cycles per attempt while reducing what the canister consumes per outcall.
+
+### 6.3 What's deliberately not done
+
+- **TLS certificate pinning.** The IC's HTTPS outcall layer terminates TLS at the boundary node; we trust the gateway's certificate chain. Pinning per-SSO certs would require an out-of-band trust anchor distribution, which is the kind of governance question (§2) we're explicitly punting on.
```
aterga: > we trust the gateway's certificate chain

double check this

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+- **Response-header inspection.** We strip headers in the transform; we don't use `Cache-Control` to drive our TTL (we hardcode 1 h, see §5.4). This mirrors today's behaviour and avoids a malicious IdP serving `Cache-Control: max-age=100000000` to pin a poisoned key indefinitely.
+- **HEAD-before-GET to size-check.** Two outcalls instead of one would double the steady-state cost. The `max_response_bytes` cap is the cleaner defense.
+
+---
+
+## 7. Problem 4 — Fragment callback
```
aterga, line 470: This section wasn't easy enough for me to understand. 

Could you ELI5 what the problem and the proposed solutions are, please?

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+
+### 8.6 Migration
+
+Existing credentials in stable storage have no `sso_domain` stamp. There are two upgrade-time approaches:
+
+- **Backfill from `OIDC_CONFIGS`.** The upgrade hook reads the about-to-be-deleted `persistent_state.oidc_configs`. For each registered domain, it runs the discovery cache lookup (cold, so triggers outcalls — this could be deferred until the cache is warm post-upgrade). Then iterates credentials, looking for any whose `iss + aud` matches a discovered provider, and stamps `sso_domain` on each. One-shot, runs in `post_upgrade`. Risk: discovery outcalls from an upgrade hook are unusual; if the upgrade fails the canister rolls back, which is the safe failure mode.
+- **Lazy backfill.** Existing credentials keep `sso_domain = None`. The FE supplies `discovery_domain` on the next use of each credential. The canister, on successful JWT verification with `Some(domain)`, stamps `sso_domain = domain` on the matching stored credential. Slower convergence; no upgrade-time outcalls; works for credentials that are actually used.
```
aterga: This is the right approach.

aterga: Note that outcalls aren't possible on upgrade anyway, we'd need a batched migration (if we didn't have this approach)

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+- **`src/frontend/src/lib/utils/ssoDiscovery.ts` is deleted.** All ~560 lines of two-hop fetch, per-hop cache, exponential-backoff retry, abort plumbing, hostname-match validation, HTTPS enforcement, and `DomainNotConfiguredError` classification go away. The canister now does the fetching, caching, and validation; the FE just calls `discover_sso` and reads the result. Removes the only FE-side direct fetch to arbitrary SSO domains (and with it, the CORS, IPv6, and timeout edge cases that module currently handles).
+
+The FE-side simplification has its own cost: `discover_sso` is an update call (~2-4 s warm, ~5-8 s cold) whereas the FE's direct fetches today are ~200-500 ms when the SSO host is fast. For the input-debounce UX in `SignInWithSso.svelte` we accept this tradeoff — the user only ever runs discovery once per SSO before signing in, and the canister-cache means subsequent runs (same domain or other users on the same domain) hit the warm path.
```
aterga, line 832: We could improve the UX significantly by also having a query, e.g. `is_sso_discovered`, that would provide data in the warmed-up cache case; only if that returns `null` would the FE need to call the update function.

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+Day 0+N:
+  PR-C2 (cache wiring + provider refactor)    — opens after C1 lands.
+
+Day 0+2N:
+  PR-C3 (API reshape, allowlist removal)      — opens after C2 lands.
+                                                Stacked FE PR opens at the same time
+                                                or in the same PR.
+
+Day 0+3N:
+  PR-C4 (lazy backfill)                       — opens after C3 lands.
```
aterga: Intuitively, this shouldn't be the last step. 

Is is safe to release the feature without this PR? If not, then it should take place before, so that we always have releasable code on the main branch.

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+```
+
+A and B can land any time, in any order, mid-stack. They don't gate C.
+
+### 9.3 What changes when this all lands
+
+- Heap memory from registered SSOs: O(cache_size) ≤ ~20 MB.
+- Background outcall fanout: zero. Discovery and JWKS fetching are sign-in-driven.
+- Apple Sign In returns name + email. Okta / Auth0 / Keycloak / Authelia / generic OIDC become viable without per-IdP workarounds for fragment-mode quirks.
```
aterga, line 908: Is this true? I certainly hope so! 

NCR

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+- The salt + nonce + `caller()` JWT redemption binding.
+- FedCM as the preferred path where the browser supports it.
+- `OpenIdCredential` `(iss, sub, aud)` storage keying. The schema gains an optional `sso_domain` field; existing data is upgrade-safe.
+- The shape of `openid_credential_remove`, the attribute-sharing layer, the wizard flows.
+
+---
+
+## 10. Open questions
+
+- **Cache size revisit.** 1000 entries per cache is a guess based on "1000 simultaneously-hot SSOs is more than we expect" and "10 MB total heap is comfortably under budget." Once we have telemetry on real-world miss rates we should revisit. If miss rate is non-trivial we either raise the cap or shorten the TTL — both have cycle-budget implications.
```
aterga: I would start more conservatively at 100

aterga: Also, please add a section to describe how we add observability:
1. metrics to record the number of cached SSO configs
2. histogram metrics for how many credentials we have for each SSO domain (bonus points)
3. plausible funnels (if they need any changes)

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+- **TTL on `Pending` entries.** `PENDING_STALE_AFTER_SECS = 120 s` is comfortable for IC HTTP outcalls under normal conditions; if we see verifier latency spikes above that floor we'd revisit.
+- **Lazy backfill convergence.** Credentials that are never used again never get `sso_domain` stamped. The metadata layer (`sso_fields_for` callers) gracefully handles `None` today (renders as the bare domain or falls through to the direct-provider config), so this is not a functional problem — but a future cleanup pass could backfill in bulk if telemetry shows long-tailed credentials hampering attribute-sharing UX.
+- **`response_mode=form_post` and FedCM coexistence.** Today's `requestJWT` falls back to popup when FedCM isn't available. Form_post applies only to the popup path. If a future FedCM upgrade adds a redirect fallback we'd revisit; for now the two paths are independent.
+- **Direct provider JWKS cache behaviour.** Hardcoded providers' JWKS today is fetched eagerly at canister bring-up via `Provider::create`. Under §5 it becomes lazy on first sign-in attempt. First-after-deploy sign-in pays one cache-miss outcall. Acceptable, but worth flagging in the rollout notes.
```
aterga: We should ideally provide visual feedback to the user to indicate that while signing them in, II is fetching a fresh SSO domain config (better progress indication ==> less frustration)

docs/ongoing/openid-sso-prod-readiness.md
```diff
@@ -0,0 +1,756 @@
...
+- The shape of `openid_credential_remove`, the attribute-sharing layer, the wizard flows.
+
+---
+
+## 10. Open questions
+
+- **Cache size revisit.** 1000 entries per cache is a guess based on "1000 simultaneously-hot SSOs is more than we expect" and "10 MB total heap is comfortably under budget." Once we have telemetry on real-world miss rates we should revisit. If miss rate is non-trivial we either raise the cap or shorten the TTL — both have cycle-budget implications.
+- **TTL on `Pending` entries.** `PENDING_STALE_AFTER_SECS = 120 s` is comfortable for IC HTTP outcalls under normal conditions; if we see verifier latency spikes above that floor we'd revisit.
+- **Lazy backfill convergence.** Credentials that are never used again never get `sso_domain` stamped. The metadata layer (`sso_fields_for` callers) gracefully handles `None` today (renders as the bare domain or falls through to the direct-provider config), so this is not a functional problem — but a future cleanup pass could backfill in bulk if telemetry shows long-tailed credentials hampering attribute-sharing UX.
```
aterga: which exact telemetry would provide this data?

aterga (approved):

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: Yes this is indeed mitigated by that, seems like this was a bad hallucination.

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: I've manually confirmed a while back that Apple returns names and emails in this case.

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: See:

> `CONFIG_REGISTRY` holds direct providers (Google, Microsoft, Apple) from the `open_id_configs` init arg and is fixed-size at boot. SSO domains are not stored anywhere on the canister — neither in a registry nor in persistent state. The cache (§5) is the only place SSO discovery data exists, and even there it's transient.

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: See:

> `CONFIG_REGISTRY` holds direct providers (Google, Microsoft, Apple) from the `open_id_configs` init arg and is fixed-size at boot. SSO domains are not stored anywhere on the canister — neither in a registry nor in persistent state. The cache (§5) is the only place SSO discovery data exists, and even there it's transient.

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: Yes, probably a clear diagram doesn't hurt here either.

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: Indeed section 5.4 has unintended repeated code block and paragraph.

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: I think data migration should be consider as early as possible indeed, instead of being an after thought.

docs/ongoing/openid-sso-prod-readiness.md
sea-snake: We wouldn't need outcalls since we could migrate with the data from the sso configuration (if not removed yet at this stage).

PR #3909, opened 2026-05-19
feat(be,fe): smtp_request envelope.to is vec, document SmtpRequestError code 550

aterga:
The off-chain SMTP gateway forwards every inbound message to `smtp_request(SmtpRequest)`. The Candid type modelled `Envelope.to` as a single `SmtpAddress`, but SMTP allows a `MAIL FROM` to be followed by multiple `RCPT TO`s and the gateway has no good way to surface that to the canister today. Bring the wire type in line with the protocol, and tighten the canister's recipient policy so the new flexibility doesn't open a phishing door.

# Changes

- `Envelope.to` is now `vec SmtpAddress` (Candid + Rust + regenerated FE bindings).
- `smtp_request` and `smtp_request_validate` accept an envelope iff it carries **exactly one** recipient and that recipient is `register@<d>` or `recover@<d>` for some `d` in `related_origins`. Multi-recipient (and empty) envelopes are rejected with **550** — a legitimate recovery email never targets a CC/BCC alongside our reserved mailbox, so accepting one would let an attacker exfiltrate the user's canister-signed challenge nonce.
- `validate_envelope` caps the recipient list at `MAX_RECIPIENTS = 100` (RFC 5321 §4.5.3.1.10) and rejects over-cap envelopes with **555**, so an open-update caller can't pad the list out to the ingress size limit.
- `SmtpRequestError` now documents what each `code` means: **550** (mailbox unavailable) is "No such user here"; **555** (syntax error) flags a malformed request shape including over-cap recipient lists.

# Tests

- `validate_rejects_multi_recipient_envelope_even_when_one_is_known` / `validate_rejects_multi_recipient_even_when_all_unknown` / `validate_rejects_empty_recipient_list` — both methods reject anything but a single recipient with 550.
- `validate_envelope_rejects_any_oversized_recipient` — bounds-check every recipient, not just the first.
- `validate_envelope_rejects_too_many_recipients` / `validate_envelope_accepts_max_recipients` — `MAX_RECIPIENTS` cap, with the inclusive upper bound covered.
- `validate_envelope_accepts_empty_to_vec` — empty `to` is shape-valid at the bounds layer; recipient-dispatch handles the 550 separately.

### DIFF
--- a/src/frontend/src/lib/generated/internet_identity_types.d.ts
+++ b/src/frontend/src/lib/generated/internet_identity_types.d.ts
@@ -1358,7 +1358,16 @@ export interface SignedRRset {
   'rtype' : number,
 }
 export interface SmtpAddress { 'domain' : string, 'user' : string }
-export interface SmtpEnvelope { 'to' : SmtpAddress, 'from' : SmtpAddress }
+export interface SmtpEnvelope {
+  /**
+   * SMTP allows multiple `RCPT TO` recipients per envelope, so this
+   * is a vec. The canister picks the first recognised recipient
+   * (`register@<domain>` or `recover@<domain>`) to decide which flow
+   * to run; unknown recipients in the same vec are ignored.
+   */
+  'to' : Array<SmtpAddress>,
+  'from' : SmtpAddress,
+}
 /**
  * SMTP gateway types — see `internet_identity_interface::smtp`. Carried
  * forward from PoC #3760 surface (the existing gateway can target this
@@ -1374,6 +1383,18 @@ export interface SmtpRequest {
   'message' : [] | [SmtpMessage],
   'gateway_flags' : [] | [Array<string>],
 }
+/**
+ * Error returned by `smtp_request` / `smtp_request_validate`.
+ * 
+ * `code` mirrors the SMTP reply codes the off-chain gateway should
+ * emit upstream:
+ * - `550` (mailbox unavailable) — "No such user here". Returned when
+ * none of the envelope recipients match a mailbox this canister
+ * handles (i.e. neither `register@<domain>` nor `recover@<domain>`,
+ * for any `<domain>` in the deploy's `related_origins`).
+ * - `555` (syntax error) — the request shape itself is malformed
+ * (e.g. missing envelope, oversize address/header/body).
+ */
 export interface SmtpRequestError { 'code' : bigint, 'message' : string }
 export type SmtpResponse = { 'Ok' : {} } |
   { 'Err' : SmtpRequestError };
--- a/src/frontend/src/lib/generated/internet_identity_types.d.ts
+++ b/src/frontend/src/lib/generated/internet_identity_types.d.ts
@@ -1361,11 +1361,16 @@ export interface SmtpAddress { 'domain' : string, 'user' : string }
 export interface SmtpEnvelope {
   /**
    * SMTP allows multiple `RCPT TO` recipients per envelope, so this
-   * is a vec. The canister picks the first recognised recipient
-   * (`register@<domain>` or `recover@<domain>`) to decide which flow
-   * to run; unknown recipients in the same vec are ignored. Capped
-   * at 100 recipients (RFC 5321 §4.5.3.1.10); envelopes exceeding
-   * the cap are rejected with code 555.
+   * is a vec at the wire level. For the recovery flows this canister
+   * serves, however, we require *exactly one* recipient and it must
+   * be `register@<domain>` or `recover@<domain>` — a legitimate
+   * recovery email never targets a CC/BCC alongside us, so any
+   * additional recipient can only come from a phishy forwarder
+   * trying to exfiltrate the user's canister-signed challenge.
+   * Multi-recipient envelopes (and empty ones) are rejected with
+   * code 550. The vec is also capped at 100 entries (RFC 5321
+   * §4.5.3.1.10); envelopes exceeding the cap are rejected with
+   * code 555.
    */
   'to' : Array<SmtpAddress>,
   'from' : SmtpAddress,
--- a/src/frontend/src/lib/generated/internet_identity_types.d.ts
+++ b/src/frontend/src/lib/generated/internet_identity_types.d.ts
@@ -1368,9 +1368,11 @@ export interface SmtpEnvelope {
    * additional recipient can only come from a phishy forwarder
    * trying to exfiltrate the user's canister-signed challenge.
    * Multi-recipient envelopes (and empty ones) are rejected with
-   * code 550. The vec is also capped at 100 entries (RFC 5321
-   * §4.5.3.1.10); envelopes exceeding the cap are rejected with
-   * code 555.
+   * code 551 ("User not local"); single-recipient envelopes whose
+   * recipient isn't one of our reserved mailboxes get 550 ("No
+   * such user here"). The vec is also capped at 100 entries (RFC
+   * 5321 §4.5.3.1.10); envelopes exceeding the cap are rejected
+   * with code 555.
    */
   'to' : Array<SmtpAddress>,
   'from' : SmtpAddress,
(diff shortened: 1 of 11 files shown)


src/internet_identity/src/email_recovery/smtp.rs
```diff
@@ -112,9 +112,17 @@ pub fn handle_smtp_request_validate(request: SmtpRequest) -> SmtpResponse {
...
     };
-    let to = &envelope.to;
-    if recipient_matches(to, SETUP_RECIPIENT_USER) || recipient_matches(to, RECOVERY_RECIPIENT_USER)
-    {
+    // The gateway may batch multiple `RCPT TO` recipients into a
+    // single envelope. Accept the transaction if at least one
+    // recipient names a mailbox this canister handles — the
+    // dispatcher in `handle_smtp_request` ignores the rest. If none
+    // match, answer 550 ("No such user here") so the gateway bounces
+    // upstream rather than wasting bandwidth pulling the body.
```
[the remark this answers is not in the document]
aterga: Reworded the rustdoc on handle_smtp_request_validate to state the new rule (accept if at least one recipient is known; 550 only when none). Pushed in 040e30da.

src/internet_identity_interface/src/internet_identity/types/smtp.rs
```diff
@@ -141,7 +148,9 @@ pub fn validate_address_bounds(addr: &SmtpAddress, label: &str) -> Result<(), Sm
 
 pub fn validate_envelope(envelope: &SmtpEnvelope) -> Result<(), SmtpResponse> {
     validate_address_bounds(&envelope.from, "Sender")?;
-    validate_address_bounds(&envelope.to, "Recipient")?;
+    for to in &envelope.to {
+        validate_address_bounds(to, "Recipient")?;
+    }
```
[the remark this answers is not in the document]
aterga: Added MAX_RECIPIENTS = 100 (RFC 5321 §4.5.3.1.10) to validate_envelope — envelopes exceeding the cap get rejected with code 555. Documented on the Candid type and covered by validate_envelope_rejects_too_many_recipients / validate_envelope_accepts_max_recipients. Pushed in 040e30da.

src/internet_identity/src/email_recovery/smtp.rs
```diff
@@ -143,18 +154,31 @@ pub async fn handle_smtp_request(request: SmtpRequest) -> SmtpResponse {
...
+        } else {
+            None
+        }
+    });
+    let recipient_flow = match recipient_flow {
+        Some(flow) => flow,
+        None => {
+            // Drop with Ok — we don't emit a per-recipient signal back
+            // to the gateway.
+            return SmtpResponse::Ok {};
```
blind-oracle: Is it intentional? Even if II does not recognize the user - we just accept & drop the message?

aterga: Good point!

aterga: Good catch, the silent-drop is wrong here — `register@<d>` / `recover@<d>` are in the public Candid surface so there was never a secret to hide, and skipping validate left a no-signal hole on the update path. Fix is out as a separate PR: https://github.com/dfinity/internet-identity/pull/3912 (didn't roll it into this one to keep the wire change and the behavioural change reviewable independently).

src/frontend/src/lib/generated/internet_identity_types.d.ts
```diff
@@ -1358,7 +1358,18 @@ export interface SignedRRset {
   'rtype' : number,
 }
 export interface SmtpAddress { 'domain' : string, 'user' : string }
-export interface SmtpEnvelope { 'to' : SmtpAddress, 'from' : SmtpAddress }
+export interface SmtpEnvelope {
+  /**
+   * SMTP allows multiple `RCPT TO` recipients per envelope, so this
+   * is a vec. The canister picks the first recognised recipient
+   * (`register@<domain>` or `recover@<domain>`) to decide which flow
+   * to run; unknown recipients in the same vec are ignored. Capped
```
sea-snake: > The canister picks the first recognised recipient (`register@<domain>` or `recover@<domain>`) to decide which flow to run; unknown recipients in the same vec are ignored.

We should reject emails with multiple recipients for the recovery flows, this only opens up an unnecessary door for phishing.

src/internet_identity/src/email_recovery/smtp.rs
```diff
@@ -112,9 +115,17 @@ pub fn handle_smtp_request_validate(request: SmtpRequest) -> SmtpResponse {
...
     };
-    let to = &envelope.to;
-    if recipient_matches(to, SETUP_RECIPIENT_USER) || recipient_matches(to, RECOVERY_RECIPIENT_USER)
-    {
+    // The gateway may batch multiple `RCPT TO` recipients into a
+    // single envelope. Accept the transaction if at least one
+    // recipient names a mailbox this canister handles — the
+    // dispatcher in `handle_smtp_request` ignores the rest. If none
+    // match, answer 550 ("No such user here") so the gateway bounces
+    // upstream rather than wasting bandwidth pulling the body.
```
sea-snake: Not sure how a gateway would batch multiple envelopes into a single envelope without breaking DKIM signatures, is this a hallucination?

src/internet_identity/src/email_recovery/smtp.rs
```diff
@@ -143,18 +154,31 @@ pub async fn handle_smtp_request(request: SmtpRequest) -> SmtpResponse {
     // address (user + domain), case-insensitively, so a direct caller
     // can't bypass dispatch by spoofing just the user-part with a
     // different domain — `smtp_request` is an open update.
+    //
+    // An SMTP envelope may carry multiple `RCPT TO` recipients (e.g.
+    // the gateway batched a multi-recipient transaction). We pick the
```
sea-snake: Same as above.

src/internet_identity/src/email_recovery/smtp.rs
```diff
@@ -1090,4 +1117,38 @@ mod tests {
         };
         assert_smtp_err_code(handle_smtp_request_validate(req), SMTP_ERR_SYNTAX_ERROR);
     }
+
+    #[test]
+    fn validate_accepts_when_one_of_many_recipients_known() {
+        // The gateway may batch multiple RCPT TOs into one envelope.
```
sea-snake: Same

src/frontend/src/lib/generated/internet_identity_types.d.ts
aterga: Good call — agreed. Pushed 14c03abf: both `smtp_request` and `smtp_request_validate` now require `envelope.to.len() == 1` and bounce multi-recipient (or empty) envelopes with 550. Updated the Candid docstring + tests; the previously-permissive `validate_accepts_when_one_of_many_recipients_known` is now `validate_rejects_multi_recipient_envelope_even_when_one_is_known`.

src/internet_identity/src/email_recovery/smtp.rs
aterga: Fair point — the 'gateway batches' framing was sloppy. The gateway doesn't decide to batch; it forwards what an upstream MTA delivered, and DKIM doesn't sign the envelope (RCPT TO sits at the SMTP transport layer), so a multi-RCPT envelope is technically representable but practically only ever arises from something we don't want to support. That entire comment block was rewritten in 14c03abf when I switched to the stricter `envelope.to.len() == 1` rule (per your other comment), so the misleading text is gone.

src/internet_identity/src/email_recovery/smtp.rs
aterga: Fair point — the 'gateway batches' framing was sloppy. The gateway doesn't decide to batch; it forwards what an upstream MTA delivered, and DKIM doesn't sign the envelope (RCPT TO sits at the SMTP transport layer), so a multi-RCPT envelope is technically representable but practically only ever arises from something we don't want to support. That entire comment block was rewritten in 14c03abf when I switched to the stricter `envelope.to.len() == 1` rule (per your other comment), so the misleading text is gone.

src/internet_identity/src/email_recovery/smtp.rs
aterga: Fair point — the 'gateway batches' framing was sloppy. The gateway doesn't decide to batch; it forwards what an upstream MTA delivered, and DKIM doesn't sign the envelope (RCPT TO sits at the SMTP transport layer), so a multi-RCPT envelope is technically representable but practically only ever arises from something we don't want to support. That entire comment block was rewritten in 14c03abf when I switched to the stricter `envelope.to.len() == 1` rule (per your other comment), so the misleading text is gone.

sea-snake (dismissed):

src/frontend/src/lib/generated/internet_identity_types.d.ts
aterga: Please use error code 551 in this case (and document it in the code)

aterga: Done in c1d36afd. Added `SMTP_ERR_USER_NOT_LOCAL = 551` in the interface crate (with a doc-comment laying out when each of 550 / 551 / 555 applies); switched the multi-recipient and empty-`to` rejection paths in both `handle_smtp_request` and `handle_smtp_request_validate` to use 551, and kept 550 only for the single-recipient-but-unknown-mailbox case. Candid `SmtpRequestError` docstring + the 3 affected unit tests updated to match.

sea-snake (approved):

PR #3921, opened 2026-05-20
feat(fe): access method controls — disable, switch, remove/unlink

MRmarioruci:
## Problem

The access method cards on the manage page had no way to switch between methods or safely remove/unlink them. The only options were rename and remove, with no guards against locking yourself out (e.g. removing the only method while signed in with a passkey, or removing the currently active method without switching first).

The recovery phrase card also had no visible "Active" indicator when the current session was established via recovery, making it unclear to users which method they were currently signed in with.

## Changes

**Disable remove/unlink on active method**
- Remove/Unlink options in the dropdown are visible but disabled when the user is signed in with that method
- Passkey tooltip: "Switch to another method before removing"
- OpenID tooltip: "Switch to another method before unlinking"
- `isCurrentAccessMethod` prop added to `PasskeyItem` and `OpenIdItem`, defaulting to `false`

**Switch option in dropdown**
- Switch menu item added to every access method card dropdown (passkeys and OpenID)
- Disabled with "Already signed-in" tooltip when that method is already the active session

**Switch modal**
- `SwitchAccessMethod.svelte` confirmation dialog re-authenticates with the selected method
- Passkeys use WebAuthn assertion via `authenticateWithPasskey`
- OpenID / SSO uses FedCM / OAuth popup via `requestJWT` + `authenticateWithJWT`
- On success updates the active session store and last-used identity record
- Throws if the returned `identityNumber` doesn't match the managed identity (guards against shared authenticators and OAuth-popup account-swapping)

**Remove/Unlink modals**
- Consolidates `RemovePasskey.svelte` and `RemoveOpenIdCredential.svelte` into `RemoveAccessMethod.svelte`
- `Props` is a discriminated union: `{ type: "passkey" }` vs `{ type: "openid"; providerName: string }` — TypeScript enforces `providerName` for OpenID
- Last-method removal gated on an active recovery session — Remove/Unlink disabled with tooltip "Sign in with a recovery method to remove/unlink" until signed in via recovery phrase or email recovery
- `isSignedInWithRecovery` derived as `true` when none of the access methods match the current session (i.e. signed in via recovery); defaults to `false` on the item components to fail closed
- Last-method copy says "recovery method" (not "recovery phrase") to cover both phrase and email recovery users

**Recovery phrase Active badge**
- `ActiveRecoveryPhrase.svelte` now shows a green "Active" badge next to the shield icon when the current session was established via the recovery phrase
- Replaces the small status dot used previously; matches the wording shown elsewhere in the manage UI

## Tests

- Unit: fix `toKey` assertion for OpenID credentials; `isCurrentAccessMethod` returns `false` for `recoveryPhrase` and `emailRecovery` auth methods
- E2E fixture: updated button labels, `assertUnremovable` changed to check `toBeDisabled()`, new helpers `assertRemoveDisabledWithTooltip`, `assertSwitchDisabled`, `SwitchAccessMethodDialog`
- E2E: "currently in use passkey is disabled when other methods exist" with tooltip assertion
- E2E: "cannot remove a single passkey without a recovery method" with tooltip assertion
- E2E: "switch is disabled when already signed in with that passkey"
- E2E: "can be cancelled — when removing a passkey" updated to cancel removal of the non-active passkey (active passkey is no longer removable by design)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
```diff
@@ -169,6 +189,101 @@
...
+          identity,
+          identityNumber,
+          authMethod: { passkey: { credentialId: authedId } },
+        });
+        lastUsedIdentitiesStore.addLastUsedIdentity({
+          identityNumber: data.identityNumber,
+          name: data.identityInfo.name[0],
+          createdAtMillis,
+          authMethod: { passkey: { credentialId: authedId } },
+        });
```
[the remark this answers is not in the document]
MRmarioruci: Good catch. Fixed in 288923e5: both the passkey and OpenID switch paths now throw when the returned `identityNumber` doesn't equal `data.identityNumber`, and `lastUsedIdentitiesStore` is updated using the verified `identityNumber` rather than `data.identityNumber`. The OAuth-popup case (different IdP account) is the higher-risk path and is covered by the same guard.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/components/RemoveAccessMethod.svelte
```diff
@@ -0,0 +1,151 @@
...
+  import { Trans } from "$lib/components/locale";
+
+  interface Props {
+    onRemove: () => Promise<void>;
+    onCancel: () => void;
+    type: "passkey" | "openid";
+    providerName?: string;
+    isCurrentAccessMethod?: boolean;
+    isLastAccessMethod?: boolean;
+  }
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in 288923e5. `Props` is now a discriminated union: `{ type: "openid"; providerName: string } | { type: "passkey" }`. The call site is split into two `<RemoveAccessMethod>` invocations so TypeScript enforces that `providerName` is supplied for the OpenID variant.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/components/PasskeyItem.svelte
```diff
@@ -18,15 +23,21 @@
...
   const {
     passkey,
     recoveryPhraseStatus,
     onRename,
     onRemove,
-    isCurrentAccessMethod,
+    onSwitch,
+    isCurrentAccessMethod = false,
+    isLastAccessMethod = false,
+    isSignedInWithRecovery = true,
```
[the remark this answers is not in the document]
MRmarioruci: Agreed — fail closed. Default flipped to `false` in 288923e5. Future callers that forget the prop now get the safer (blocking) behavior.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/components/OpenIdItem.svelte
```diff
@@ -12,10 +16,20 @@
...
   }
 
-  const { openid, onUnlink, isCurrentAccessMethod }: Props = $props();
+  const {
+    openid,
+    onUnlink,
+    onSwitch,
+    isCurrentAccessMethod = false,
+    isLastAccessMethod = false,
+    isSignedInWithRecovery = true,
```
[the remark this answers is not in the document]
MRmarioruci: Same fix as the PasskeyItem comment — default flipped to `false` in 288923e5.

MRmarioruci:
All four inline findings addressed in 288923e5, replies on each thread.

The two suppressed/low-confidence comments in the review body are also addressed by the same commit:

- **OpenID switch — JWT/account validation (+page.svelte:275):** Same root cause as the inline passkey-switch comment. Both paths now `throw` when the returned `identityNumber` doesn't equal `data.identityNumber`, so a user picking a different IdP account in the popup is rejected rather than silently writing the wrong record into `lastUsedIdentitiesStore`.
- **"Recovery phrase" copy when last method (RemoveAccessMethod.svelte:65):** Valid — the removal gate accepts either recovery phrase or email recovery, so the copy was misleading for email-recovery users. Reworded to "Your recovery method will become your sole way of regaining access to this identity."

src/frontend/src/lib/components/ui/Select.svelte
```diff
@@ -77,7 +77,7 @@
         <Tooltip
           label={option.tooltip ?? ""}
           hidden={option.tooltip === undefined ? true : undefined}
-          direction="left"
+          direction="right"
```
sea-snake: Why was the direction changed from left to right here?

src/frontend/src/lib/components/ui/Tooltip.svelte
```diff
@@ -72,39 +84,105 @@
...
+          margin,
+          Math.min(bestLeft, window.innerWidth - tooltipRect.width - margin),
+        );
+        bestTop = Math.max(
+          margin,
+          Math.min(bestTop, window.innerHeight - tooltipRect.height - margin),
+        );
+
+        tooltipRef.style.top = `${bestTop}px`;
+        tooltipRef.style.left = `${bestLeft}px`;
```
sea-snake, line 187: I don't see any early return, doesn't this mean it'll always clamp even if it isn't needed?

src/frontend/src/lib/components/ui/Tooltip.svelte
```diff
@@ -57,6 +57,18 @@
...
+    const n = parseFloat(value);
+    if (Number.isNaN(n)) return 0;
+    if (value.endsWith("rem") || value.endsWith("em")) {
+      const rootSize = getComputedStyle(document.documentElement).fontSize;
+      return n * (rootSize !== "" ? parseFloat(rootSize) : 16);
+    }
+    return n;
+  };
+
   $effect(() => {
```
sea-snake, line 72: Now we finally have [CSS anchor positioning with baseline support](https://caniuse.com/css-anchor-positioning), this whole component can probably be simplified to some css. But that's outside the scope of this PR and would likely still need a polyfill for older browsers for the time being :/

(You can ignore this comment in this PR)

MRmarioruci: how far back do we usually support? Or is it not a guarantee on our end?

src/frontend/src/lib/components/ui/Tooltip.svelte
MRmarioruci: You're right, but it's a no-op mathematically (the Math.max/Math.min returns the same value when the position is already in-bounds). No need for the useless calculation though, so will wrap it.

src/frontend/src/lib/components/ui/Select.svelte
MRmarioruci: reverted in 7003be91. Probably hallucinated..

src/frontend/src/lib/components/ui/Tooltip.svelte
MRmarioruci: fixed in 7003be91.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
```diff
@@ -53,6 +67,12 @@
   const removingAccessMethod = $derived(
     accessMethods.find((m) => removingAccessMethodKey === toKey(m)),
   );
+  const switchingAccessMethod = $derived(
+    accessMethods.find((m) => switchingAccessMethodKey === toKey(m)),
+  );
+  const isSignedInWithRecovery = $derived(
+    !accessMethods.some((m) => isCurrentAccessMethod($authenticatedStore, m)),
```
sea-snake: Nit: assuming the user is signed in with a recovery method if the none of the access methods are active is a potential footgun. It's probably better to check the method type in the authentication store to see if it's a recovery phrase or email.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
```diff
@@ -169,6 +189,118 @@
...
+              "OpenID authentication is not available for this account.",
+            );
+          jwt = await requestJWT(
+            {
+              clientId: config.client_id,
+              configURL: config.fedcm_uri[0],
+              authURL: config.auth_uri,
+              authScope: config.auth_scope.join(" "),
+            },
+            { nonce: session.nonce, mediation: "optional" },
```
sea-snake: Add a loginHint here so selecting the account can be skipped for most OpenID providers.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
```diff
@@ -169,6 +189,118 @@
...
+        if (ssoDomain !== undefined) {
+          jwt = await requestWithPopup(
+            discoverSsoConfig(ssoDomain).then((ssoResult) => ({
+              clientId: ssoResult.clientId,
+              authURL: ssoResult.discovery.authorization_endpoint,
+              authScope: selectAuthScopes(
+                ssoResult.discovery.scopes_supported,
+              ).join(" "),
+            })),
+            { nonce: session.nonce, mediation: "optional" },
```
sea-snake: Add a loginHint here so selecting the account can be skipped for most OpenID providers.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
MRmarioruci: Good point — fixed in bd55b6b9. Now derives directly from `authMethod`:

```ts
const isSignedInWithRecovery = $derived(
  "recoveryPhrase" in $authenticatedStore.authMethod ||
    "emailRecovery" in $authenticatedStore.authMethod,
);
```

This sidesteps the implicit "no access method matched ⇒ recovery" assumption and won't drift if a new auth method variant is added later.

sea-snake (approved):
Pre-approving, please see the comments for pending changes.

src/frontend/src/lib/components/ui/Tooltip.svelte
sea-snake: > how far back do we usually support? Or is it not a guarantee on our end?

It depends on the browser, for example Chrome and Firefox have reliable auto updating mechanisms that enable us to basically focus only on supporting their latest release in practice. But Safari is frustratingly tied to to the macOS/iOS version, in practice that means we have quite a few users with outdated Safari browsers as can be seen here: https://plausible.io/identity.internetcomputer.org?f=is,browser,Safari

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
MRmarioruci: fixed 1abbb9cd08c44d6024d43b0c6109004f3029bb7d

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
MRmarioruci: 1abbb9cd08c44d6024d43b0c6109004f3029bb7d

aterga (approved):

sea-snake (changes requested):
See my latest comment regarding login hint (blocker)

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
```diff
@@ -169,6 +190,123 @@
...
+          authMethod: { passkey: { credentialId: authedId } },
+        });
+      } else if ("openid" in switchingAccessMethod) {
+        const { iss, aud, metadata, sso_domain } = switchingAccessMethod.openid;
+        const ssoDomain = sso_domain[0];
+        // Pre-select the same account the user is switching to. Most OpenID
+        // providers honor `login_hint` and skip the account picker; this also
+        // reduces the chance of landing on a different anchor (caught by the
+        // identityNumber guard below).
+        const loginHint = getMetadataString(metadata, "email");
```
sea-snake: Blocking: login hint is a literal property on last used storage, don't use the email metadata since that's not valid for all openid providers.

MRmarioruci: thanks, fetching it from storage now 39c35cfce3601808c3c2743eccd52d8860f95f6a

sea-snake (approved):

PR #3929, opened 2026-05-21
fix(fe): hand off auth to manage tab via postMessage from authorize popover

MRmarioruci:
Clicking "Manage your Internet Identity" from the identity popover on the authorize page used to open a new tab that bounced back to the welcome page instead of showing the manage screen. The popover button was visible before the user had finished signing in on the authorize page, so the new tab had no authentication and the manage route's auth guard redirected it away.

## Changes

- New `src/frontend/src/lib/utils/auth-handoff.ts` exposes a same-origin postMessage handshake: `receiveAuthFromOpener` (in the new tab) generates a fresh non-extractable ECDSA key and shares only its public key with the opener via an `ii-handoff:ready` message. `sendAuthToOpenedTab` (in the opener) mints a short-lived delegation scoped to the II canister to that public key and returns the augmented delegation chain. The new tab constructs a `DelegationIdentity` from the chain and its own key. Both sides verify `event.source`, `event.origin`, and a per-handoff nonce shared via the new tab's URL fragment.
- The authorize-page popover authenticates with the cached identity if needed, generates a 128-bit random nonce, and opens \`/manage#h=<nonce>\` via \`window.open()\`. If \`window.open()\` returns \`null\` (popup blocked), it falls back to same-tab \`goto(\"/manage\")\` — the in-memory session survives the navigation.
- The manage-route layout load calls \`receiveAuthFromOpener\` before the redirect-to-\`/login\` path; on a successful handshake it hydrates the auth store and proceeds normally. An \`afterNavigate\` hook in the manage layout also strips the handoff fragment from the URL in case SvelteKit's router re-syncs it after the eager \`history.replaceState\` inside \`receiveAuthFromOpener\`.

## Tests

Unit tests in \`auth-handoff.test.ts\` cover the round-trip for all four auth-method variants (passkey, OpenID, recovery phrase, email recovery), principal preservation through the delegation extension, the \`targets\` field scoped to the II canister, a 30-minute expiry bound, expired-delegation rejection, the race where an auth message arrives before the receiver's key has finished generating, the cancel-before-ready path on the opener, and source/origin/nonce filtering on both sides.

Playwright spec \`manage-identity.spec.ts\` exercises the full popover → handshake → manage-page flow via the canonical test-app-driven second-popup pattern.


MRmarioruci:
btw I would expect zeropath to at least flag the Security concern I raised..

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,279 @@
...
+    return null;
+  }
+}
+
+export function sendAuthToOpenedTab(
+  target: Window,
+  auth: Omit<Authenticated, "agent" | "actor" | "salt" | "nonce">,
+  expectedNonce: string,
+  timeoutMs = 2000,
+): { cancel: () => void } {
```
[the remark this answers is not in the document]
MRmarioruci: Addressed in a01dbf00 — `.catch(() => undefined)` on the promise + `undefined` guard in the listener.

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,279 @@
...
+    chainJson,
+    authMethod: serializeAuthMethod(auth.authMethod),
+  };
+}
+
+export async function deserializeAuth(
+  payload: AuthHandoffPayload,
+): Promise<Omit<Authenticated, "agent" | "actor" | "salt" | "nonce"> | null> {
+  try {
+    const privateKey = await crypto.subtle.importKey(
```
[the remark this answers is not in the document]
MRmarioruci: Addressed in a01dbf00 — `privateKey` imported as non-extractable. `publicKey` stays extractable: the SDK derives the principal via `exportKey("spki", publicKey)` in `ECDSAKeyIdentity.fromKeyPair`.

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,284 @@
...
+  | { kind: "emailRecovery"; principal: string };
+
+// Nonce in the new tab's URL fragment; opener requires a match before sending auth.
+export const HANDOFF_HASH_KEY = "h";
+
+export function generateHandoffNonce(): string {
+  return crypto.randomUUID();
+}
+
+function isAuthHandoffReady(data: unknown): data is AuthHandoffReady {
```
sea-snake: Serializaition and validation can now be done with `zod` since we've moved to relying on that in the authorization flows recently. This should simplify below code significantly :D

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,284 @@
...
+  auth: Omit<Authenticated, "agent" | "actor" | "salt" | "nonce">,
+  expectedNonce: string,
+  timeoutMs = 2000,
+): { cancel: () => void } {
+  let cancelled = false;
+
+  const listener = async (event: MessageEvent) => {
+    if (
+      event.source !== target ||
+      event.origin !== location.origin ||
```
sea-snake: When reading global variables, particularly around security sensitive code, let's always make sure to use `globalThis` e.g . `globalThis.location.origin`, this avoids reading from a possible shadowed variable.

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,284 @@
...
+      ) as DerEncodedPublicKey;
+      const receiverPublicKey = { toDer: () => receiverDer };
+
+      const newChain = await DelegationChain.create(
+        auth.identity,
+        receiverPublicKey,
+        new Date(Date.now() + 30 * 60 * 1000),
+        {
+          previous: auth.identity.getDelegation(),
+          targets: [canisterId],
```
sea-snake: Why is a target specified here? The default value we're currently using for II FE delegations is `undefined`.

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,284 @@
...
+    emailRecovery: { principal: Principal.fromText(method.principal) },
+  };
+}
+
+export function sendAuthToOpenedTab(
+  target: Window,
+  auth: Omit<Authenticated, "agent" | "actor" | "salt" | "nonce">,
+  expectedNonce: string,
+  timeoutMs = 2000,
+): { cancel: () => void } {
```
sea-snake, line 107: cancel methods are a great opportunity to use the abort signal API.

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,284 @@
...
+import type { Authenticated } from "$lib/stores/authentication.store";
+import { canisterId } from "$lib/globals";
+import { fromBase64, toBase64 } from "$lib/utils/utils";
+
+const MSG_READY = "ii-handoff:ready";
+const MSG_AUTH = "ii-handoff:auth";
+
+interface AuthHandoffReady {
+  type: typeof MSG_READY;
+  nonce: string;
```
sea-snake: What's the nonce for?

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -317,10 +336,55 @@
                   isIdentityPopoverOpen = false;
                   isAuthDialogOpen = true;
                 }}
-                onManageIdentity={(): Promise<void> => {
+                onManageIdentity={async (): Promise<void> => {
```
sea-snake: This inline function should be extracted into a `handleManageIdentity` variable.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -317,10 +336,55 @@
...
+                      // blocked. Surface the confirmation dialog and let
+                      // its own button click drive window.open with fresh
+                      // activation. When the user was already signed in we
+                      // never awaited anything, so the popup goes straight
+                      // through.
+                      pendingManageOpen = { nonce, auth };
+                      return;
+                    }
+                    const w = window.open(
+                      `/manage#${HANDOFF_HASH_KEY}=${encodeURIComponent(nonce)}`,
```
sea-snake: What's the handoff hash key and nonce for?

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -429,6 +493,46 @@
...
+            const w = window.open(
+              `/manage#${HANDOFF_HASH_KEY}=${encodeURIComponent(nonce)}`,
+              "_blank",
+            );
+            if (w === null) {
+              pendingManageOpen = undefined;
+              void goto("/manage");
+              return;
+            }
+            pendingHandoff?.cancel();
```
sea-snake: I assume this cancel is to cleanup any prior attempts, but ideally those should have been cleaned up on closing the dialog already, to avoid the need to cleanup before sending in the first place.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -429,6 +493,46 @@
...
+      <div class="flex flex-col">
+        <h2 class="text-text-primary my-2 self-start text-2xl font-medium">
+          {$t`You're signed in`}
+        </h2>
+        <p class="text-text-secondary mb-6 self-start text-sm">
+          {$t`Open Internet Identity in a new tab to manage your access methods and recovery options.`}
+        </p>
+        <button
+          class="btn btn-primary btn-lg w-full gap-2"
+          onclick={() => {
```
sea-snake: Nit: This method should also be extracted into e.g. `handleOpenManageTab`.

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,284 @@
...
+  if (opener === null || opener.closed) {
+    return Promise.resolve(null);
+  }
+
+  const nonce = new URLSearchParams(window.location.hash.slice(1)).get(
+    HANDOFF_HASH_KEY,
+  );
+  if (nonce === null) {
+    return Promise.resolve(null);
+  }
```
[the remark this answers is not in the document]
MRmarioruci: Done in 820d080e: strip via globalThis.history.replaceState eagerly inside receiveAuthFromOpener. afterNavigate stays as the backstop.

src/frontend/src/lib/utils/auth-handoff.ts
MRmarioruci: Done in 820d080e: zod schemas for both message shapes, types via z.infer.

src/frontend/src/lib/utils/auth-handoff.ts
MRmarioruci: Done in 820d080e: location.origin, location.hash, and history.replaceState all go through globalThis. Left window.addEventListener and window.opener as-is since they're not the shadowing risk and the tests stub them there.

src/frontend/src/lib/utils/auth-handoff.ts
MRmarioruci: Defense in depth. If this chain ever leaks (postMessage logging, devtools, etc.), targets keeps it usable only against II. Happy to drop it if you prefer.

src/frontend/src/lib/utils/auth-handoff.ts
MRmarioruci: Done in 820d080e: AbortController owns both the listener registration and the timer. cancel() calls controller.abort().

src/frontend/src/lib/utils/auth-handoff.ts
MRmarioruci: Opener generates a fresh nonce per click, writes it into the new tab's URL fragment, and only sends auth after seeing a ready message that echoes the same nonce back. Stops any other tab on the origin from spoofing the handshake.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
MRmarioruci: Done in 04b1dffa.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
MRmarioruci: Same nonce as in auth-handoff.ts. Opener writes it into the new tab's URL fragment, the new tab echoes it in its ready message, opener checks they match before posting the chain. The hash key is just the URL param name we look for (h=...).

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
MRmarioruci: Done in 04b1dffa. Cancel runs at the start of handleManageIdentity and in the confirmation dialog's onClose. The pre-send cancel is gone.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
MRmarioruci: Done in 04b1dffa.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
MRmarioruci: basically stops other tabs on the origin from spoofing the handshake

src/frontend/src/lib/utils/auth-handoff.ts
sea-snake: Please drop it, it's not scoped to a single canister in other places either.

MRmarioruci: Dropped in 91fb9db5.

src/frontend/src/lib/utils/auth-handoff.ts
MRmarioruci: Dropped in 91fb9db5. Source + origin checks already bind the handshake to the specific opened tab, so the nonce was belt and suspenders. /manage now uses a fixed #handoff marker.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
MRmarioruci: Nonce dropped in 91fb9db5. Hash key kept as a fixed `#handoff` marker so the receiver knows to ping the opener instead of running the normal sign-in flow (avoids unnecessary key generation on cold /manage tabs).

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -134,6 +142,15 @@
...
   let isAuthenticating = $state(false);
   let isManageIdentitiesDialogOpen = $state(false);
+  let pendingHandoff: { cancel: () => void } | undefined;
+  // Set when the user clicked manage but had to authenticate first — the
+  // passkey/IdP prompt consumes the click's transient activation on Safari
+  // (and strict Firefox), so a follow-up window.open() would be silently
+  // blocked. The confirmation dialog's own button click provides fresh
+  // activation and drives window.open from there.
+  let pendingManageOpen = $state<{
+    auth: Omit<Authenticated, "agent" | "actor" | "salt" | "nonce">;
```
aterga: Where does this set come from?

* `agent`
* `actor`
* `salt`
* `nonce`

MRmarioruci: Mirrors the set signature in authentication.store.ts:33-35: agent/actor are created once in init() and shared by the store, and salt/nonce are derived inside set() from identity.getPrincipal() via createAnonymousNonce. Truth be told this could be deduped a bit cause it's duplicated in other places as well

MRmarioruci: 88be0e2d

src/frontend/src/lib/utils/auth-handoff.ts
```diff
@@ -0,0 +1,258 @@
...
+  if ("recoveryPhrase" in method) {
+    return {
+      kind: "recoveryPhrase",
+      principal: method.recoveryPhrase.principal.toText(),
+    };
+  }
+  return {
+    kind: "emailRecovery",
+    principal: method.emailRecovery.principal.toText(),
+  };
```
sea-snake:
```suggestion
  if ("emailRecovery" in method) {
    return {
      kind: "emailRecovery",
     principal: method.emailRecovery.principal.toText(),
    };
  }
  
  method satisfies never;
```

sea-snake (approved):

src/frontend/src/lib/utils/auth-handoff.ts
MRmarioruci: Applied in 7c6c9c04. Added a `throw` after `method satisfies never` — `satisfies` is type-only, so TS still wanted an ending return/throw.

PR #3934, opened 2026-05-22
DO NOT MERGE: add email_recovery_logs debug query

sea-snake:
Adds a debug-only `email_recovery_logs` query that returns the canister-side view of recent recovery-flow steps. Helpful when investigating why an in-flight recovery is stuck (DNSSEC verification, DKIM/DMARC failure, gateway misdelivery, etc.) without redeploying the canister with ad-hoc `ic_cdk::println!` instrumentation.

Each log entry is `[<unix_secs>] <message>` and covers `prepare_add`, `prepare_delegation`, `smtp_request`, `submit_dkim_leaf`, and `credential_remove`. Nonces are truncated (verbose prefix + 4 suffix chars) and `session_pk` is logged only as a byte count so the per-challenge secret isn't recoverable from the buffer.

The log lives in a heap `Vec<String>` keyed off thread-local state, bounded by `RECOVERY_LOG_MAX_ENTRIES` (1024 entries) with oldest-first eviction, and is not persisted across canister upgrades.

## Changes

- New `state::recovery_logs` field + `push_recovery_log` helper with bounded growth.
- New `email_recovery::log` / `nonce_for_log` helpers used from `prepare`, `smtp`, `submit_leaf`, and the `credential_remove` canister method.
- New `email_recovery_logs : () -> (vec text) query` in the candid surface.

<div align="left">Previous: #3939</div>

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -1485,6 +1485,14 @@ service : (opt InternetIdentityInit) -> {
     email_recovery_get_delegation : (EmailRecoveryGetDelegationArgs) -> (variant { Ok : SignedDelegation; Err : EmailRecoveryError }) query;
     email_recovery_credential_remove : (IdentityNumber, text) -> (variant { Ok; Err : EmailRecoveryError });
 
+    // Debug-only. Returns the heap-resident log of recovery-flow
+    // steps (prepare_add, prepare_delegation, smtp_request,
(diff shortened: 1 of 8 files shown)


aterga:
Closing: this was a temporary debugging aid for the stuck-recovery investigation, which concluded when #3989 (email recovery hangs for Outlook addresses) merged.

---

PR #3946, opened 2026-05-25
refactor(be): typestate pipeline for SMTP request validation

aterga:
## Summary

- The off-chain SMTP gateway is untrusted, but until now the canister threaded the wire-shape `SmtpRequest` through DKIM and DMARC verifiers and trusted each layer to re-extract what it needed; nothing in the type system distinguished a just-arrived request from one that had passed RFC structural checks, DKIM-certification, or DMARC alignment. This refactor introduces a three-stage typestate (`UnverifiedSmtpRequest` → `Vec<SignedSmtpRequestProjection>` → `VerifiedSmtpRequest`) so the trust level is encoded in the type. Downstream code now reads message content via `header()` / `body()` on a verified value only — no path can hand a partially-checked request to logic that assumes verification ran.
- Stage 1 enforces RFC 5322 §3.6 single-occurrence on the message headers (exactly-once `From`/`Date`/`Subject`, at-most-once `Sender`/`Reply-To`/`To`/`Cc`/`Bcc`/`Message-ID`/`In-Reply-To`/`References`, at-least-once `DKIM-Signature` when a message is present). Wire-shape problems now surface at the canister entry as SMTP 555 before any cryptographic work runs.
- Bottom-up signature iteration (RFC 6376 §5.4) replaces the previous top-down "first-only" assumption: forwarders prepend signatures, so the originating-domain signer sits towards the bottom; both the DoH and DNSSEC paths now pick the bottom-most signature whose `d=` matches the relevant domain.
- Legacy `dkim::verify::verify` and `dmarc::verify::verify_email` (plus the `DkimVerifyResult` / `EmailVerificationStatus` result enums) are deleted; their per-signature DKIM math and DMARC-alignment helper live in the typestate's `verify_one_signature` and `compute_dmarc_outcome` — single source of truth, no duplication.

## Tests

- 17 new unit tests in `email_recovery/typestate.rs` cover stage-1 RFC checks (exactly-once / at-most-once / at-least-once enforcement, case-insensitive name match, envelope-only acceptance, bounds propagation), stage-2 (per-signature `Vec`, aggregated diagnostics on all-fail, public surface limited to `signature_domain`/`selector`/`algorithm`), stage-3 (per-signature error aggregation, bottom-up iteration order), and the `compute_dmarc_outcome` alignment matrix lifted out of the deleted dmarc/verify.rs.
- `dkim/test_vectors.rs` and `dmarc/test_vectors.rs` migrated to drive the typestate end-to-end against the existing synth-RSA `.eml` fixtures.
- `dkim/verify.rs`'s 7 inline tests rewritten to assert on `VerificationError.last_reason`; the 7 `blank_b_tag_value` unit tests stay alongside the primitive itself.
- `cargo test -p internet_identity --bin internet_identity` → 549 passed.
- `cargo clippy --all-targets -D warnings` clean (no `#[allow(clippy::...)]` decorators added).
- `cargo build --release --target wasm32-unknown-unknown` succeeds.


src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1345 @@
...
+            RfcError::HeaderCount {
+                header,
+                found,
+                expected,
+            } => smtp_err(
+                SMTP_ERR_SYNTAX_ERROR,
+                format!(
+                    "RFC 5322 §3.6 violation: header '{header}' appears {found} times, \
+                     expected {expected:?}"
+                ),
```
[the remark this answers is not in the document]
aterga: Fixed — `HeaderCount` now has a `Display` impl, so the 555 message reads `expected exactly once` instead of leaking the `HeaderCount { min: 1, max: Some(1) }` debug shape.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1345 @@
...
+        // when message present", so an absent message here is the
+        // envelope-only case — which doesn't have signatures to scope
+        // and shouldn't have reached stage 2 in the first place. We
+        // surface this as an aggregated empty diagnostic rather than
+        // panicking; callers that legitimately want to stop after
+        // stage 1 (the `smtp_request_validate` query) don't call
+        // through to stage 2 at all.
+        let Some(message) = unverified.request.message.as_ref() else {
+            return Err(DkimScopeError {
+                per_signature: vec!["stage 2 reached without a message body — caller bug".into()],
```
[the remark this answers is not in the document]
aterga: Fixed — replaced with `missing message body`; same level of detail, and the wording is safe to surface back to the gateway as SMTP 555.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1345 @@
...
+    /// times the DKIM math picks the latest matching header bottom-up
+    /// (RFC 6376 §5.4); for the recovery flow's headers (`From`,
+    /// `Subject`, `Date`, ...) stage 1 already enforces single-
+    /// occurrence so there is at most one match here.
+    pub fn header(&self, name: &str) -> Option<&str> {
+        // `signed_headers` already contains only the headers covered by
+        // h=, so a name miss here means "not signed."
+        self.signed_headers
+            .iter()
+            .rev()
```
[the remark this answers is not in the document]
aterga: Fixed — `build_verified` now sorts the selected indices ascending so `signed_headers` is emitted in message source order; `header()`s reverse scan is genuinely bottom-up over the actual message bytes, matching what DKIM consumed for the first `h=` entry. Same edit covers the O(n²) concern in #3298363867.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1345 @@
...
+        }
+        if let Some(idx) = chosen {
+            consumed.push(idx);
+            signed_headers.push(message.headers[idx].clone());
+        }
+        // h= name absent from the message contributes nothing — DKIM
+        // signed the empty string for it; from the caller's perspective
+        // `header()` will return None, which matches "DKIM does not
+        // vouch for this header."
+    }
```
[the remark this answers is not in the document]
aterga: Fixed — switched to a `Vec<bool>` consumed-mask (O(|h|·|headers|), no `Vec::contains` linear scan) and sort selected indices ascending so `signed_headers` is in true source order. Same edit as #3298363847.

src/internet_identity/src/email_recovery/smtp.rs
```diff
@@ -476,34 +489,43 @@ fn find_nonce_in(haystack: &str) -> Option<String> {
...
-    // first by every signer we've observed. If the first header is
-    // unparseable or doesn't match the registered domain, we reject
-    // rather than walking the rest of the list: that policy keeps the
-    // attack surface small for an authentication-critical path and
-    // matches what `crate::dkim::verify` does on its main entry.
-    let dkim_header = message
-        .headers
+    let zone = &snapshot.registered_domain;
+    let projections: Vec<SignedSmtpRequestProjection> =
+        unverified.clone().try_into().map_err(map_scope_error)?;
```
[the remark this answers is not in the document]
aterga: Fixed — `prepare_partial_verification` now takes `UnverifiedSmtpRequest` by value; `handle_smtp_request` scopes its envelope/nonce borrows so ownership flows through cleanly without a clone.

src/internet_identity/src/email_recovery/smtp.rs
```diff
@@ -604,45 +595,65 @@ fn prepare_partial_verification(
...
         .rsplit_once('@')
         .map(|(_, d)| d.to_string())
         .ok_or_else(|| {
             EmailRecoveryError::InternalCanisterError("stored claimed address has no '@'".into())
         })?;
-    let dkim_fqdn = format!("{}._domainkey.{}", sig.s, domain);
+
+    let projections: Vec<SignedSmtpRequestProjection> =
+        unverified.clone().try_into().map_err(map_scope_error)?;
+
```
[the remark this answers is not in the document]
aterga: Fixed — same shape as #3298363896; `verify_setup_email_doh` takes `UnverifiedSmtpRequest` by value, exactly one of the DNSSEC/DoH branches in `handle_smtp_request` consumes it.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1345 @@
...
+        assert_eq!(
+            err.per_signature.len(),
+            2,
+            "expected both signatures attempted; got {err:?}"
+        );
+        // First attempt is index 0 in the err's per_signature list;
+        // its (attempt_idx, _, _) field is also 0.
+        assert_eq!(
+            err.per_signature[0].0, 0,
+            "first attempt must be attempt index 0 (the bottom signature)"
```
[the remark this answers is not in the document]
aterga: Fixed — the two test signatures now fail with distinct reasons (bottom=`BodyHashMismatch`, top=`UnsupportedCanonicalization` via `c=simple/simple`). `per_signature[0].1` and `last_reason` pin iteration direction; a top-down regression would surface `UnsupportedCanonicalization` at index 0 and fail the assertion.

src/internet_identity/src/email_recovery/typestate.rs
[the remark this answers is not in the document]
aterga: Fixed in 3029cfa7: added `Display for HeaderCount` rendering "exactly once" / "at most once" / "at least once", used in `RfcError → SmtpResponse` so the 555 message no longer leaks the `{min, max}` Debug shape.

src/internet_identity/src/email_recovery/typestate.rs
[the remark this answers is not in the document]
aterga: Fixed in 3029cfa7: changed the message to `"missing message body"`. Encoding the precondition in the type would force two `TryFrom<UnverifiedSmtpRequest>` variants (one for envelope-only, one for message-present) just to forbid a caller path that already does the right thing — the neutral string keeps the API single-shape while staying safe to surface as SMTP 555.

src/internet_identity/src/email_recovery/typestate.rs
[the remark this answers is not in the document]
aterga: Fixed in 3029cfa7: `build_verified` now `selected.sort_unstable()`s the selected indices before materialising, so `signed_headers` is in true message source order. `header()`'s `.iter().rev()` now genuinely walks bottom-up over the actual message bytes — and for an `h=` that lists the same name multiple times it returns the bottom-most surviving instance (the one DKIM consumed for the first `h=` entry).

src/internet_identity/src/email_recovery/typestate.rs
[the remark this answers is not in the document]
aterga: Fixed in 3029cfa7: `consumed` is now `vec![false; message.headers.len()]` (O(1) per lookup), and the selected indices are `sort_unstable()`-ed ascending before mapping to `signed_headers` so the result is in message source order rather than `h=` order.

src/internet_identity/src/email_recovery/smtp.rs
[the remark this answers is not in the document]
aterga: Fixed in 3029cfa7: `prepare_partial_verification` now takes `unverified: UnverifiedSmtpRequest` by value (no clone). `handle_smtp_request` scopes its envelope/nonce borrows so ownership flows through cleanly.

src/internet_identity/src/email_recovery/smtp.rs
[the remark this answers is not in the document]
aterga: Fixed in 3029cfa7: `verify_setup_email_doh` now takes `unverified: UnverifiedSmtpRequest` by value (no clone on the DoH hot path either).

src/internet_identity/src/email_recovery/typestate.rs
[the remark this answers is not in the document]
aterga: Fixed in 3029cfa7: the two signatures now fail with distinct reasons — top (`s=forwarder`, `c=simple/simple`) fails as `UnsupportedCanonicalization`, bottom (`s=mail`, valid canon but a synthetic mismatched `bh=`) fails as `BodyHashMismatch`. The test asserts `per_signature[0].1 == BodyHashMismatch`, which only holds under bottom-up iteration; a top-down regression would surface `UnsupportedCanonicalization` first and trip the assertion.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1403 @@
...
+            ] {
+                check_count(&message.headers, name, HeaderCount::AT_MOST_ONE)?;
+            }
+
+            check_count(
+                &message.headers,
+                "DKIM-Signature",
+                HeaderCount::AT_LEAST_ONE,
+            )?;
+        }
```
sea-snake: These checks should probably be combined, not sure why these are separate from `validate_smtp_request`.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1403 @@
...
+    /// Crate-private: the validated message body the signature claims
+    /// to cover. Same hazard as `parsed_signature()` — the body hash
+    /// hasn't been verified yet.
+    pub(crate) fn message(
+        &self,
+    ) -> &internet_identity_interface::internet_identity::types::smtp::SmtpMessage {
+        self.request
+            .message
+            .as_ref()
+            .expect("stage 2 always carries a message")
```
sea-snake: Throwing should be avoided.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1403 @@
...
+    /// Raw body bytes — the same bytes whose canonical form the winning
+    /// signature's `bh=` validated. Returned by `body()` so callers can
+    /// parse the message body content (e.g. extract nonces, render
+    /// rich content).
+    body: Vec<u8>,
+    /// The winning signature's `d=`. `pub(crate)` so the
+    /// `dkim`/`dmarc` test-vector modules can assert which signer was
+    /// admitted. Not exposed publicly — downstream code outside the
+    /// crate reads message content via `header()` and `body()` only.
+    pub(crate) winning_dkim_domain: String,
```
sea-snake: Not sure why some of these fields are prefixed with `winning_` and others are not. I'd personally omit the prefix, the fact that there were multiple signatures is irrelevant at this point.

src/internet_identity/src/email_recovery/typestate.rs
aterga: Fixed in f1fad4b3: `validate_smtp_request` now runs bounds + RFC 5322 §3.6 uniqueness in one pass in the wire-types crate. Stage 1's `try_from` collapses to a thin marker over that contract; `RfcError` and `HeaderCount` are gone from the canister side, and the stage-1 unit tests followed the validation logic into the interface crate.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1403 @@
...
+            last_reason,
+            combined_checks,
+        })
+    }
+}
+
+/// Verify one specific signature: tag contract, DNS record parse, DNS
+/// tag contract, body-hash, signed-headers canonicalisation, signature
+/// math, From-header extraction. Returns the From-domain on success.
+fn verify_one_signature(
```
sea-snake: Not sure why this method needs to explicit have `_one_` in it's name, it should already be obvious based on the fact it takes a single projection.

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,1403 @@
...
+        dmarc_outcome,
+        winning_checks,
+    }
+}
+
+/// Compute the DMARC alignment outcome given the DKIM `d=`, the From-
+/// header domain, and the optional published DMARC record. `pub(crate)`
+/// so the DMARC unit tests can pin its behaviour without going through
+/// the cryptographic stage-3 round-trip.
+pub(crate) fn compute_dmarc_outcome(
```
sea-snake: Shouldn't dkim and dmarc specific methods live in those respective crates? The typestate implementation should only orchestrate functions from various crates, not implement them itself.

src/internet_identity/src/dmarc/verify.rs
```diff
@@ -1,254 +0,0 @@
...
-            _ => VerificationFailReason::DmarcMisaligned,
-        };
-        EmailVerificationStatus::Unverified {
-            reason,
-            checks: dkim_checks,
-        }
-    }
-}
-
-fn compute_outcome(dkim_domain: &str, from_domain: &str, dmarc_txt: Option<&str>) -> DmarcOutcome {
```
sea-snake, line 115: Why was this moved to the typestate file? Seems to me that it would make sense to keep the dmarc outcome computation in the dmarc crate.

src/internet_identity/src/email_recovery/typestate.rs
aterga: Addressed in f1fad4b3 — RFC §3.6 multiplicity logic now lives inside `validate_smtp_request` (`validate_message_header_uniqueness` helper). Stage 1 collapses to a thin `validate_smtp_request(&request)?` wrapper, and the per-rule unit tests live alongside the validator in the wire-types crate. The split that prompted the question is gone.

src/internet_identity/src/email_recovery/typestate.rs
aterga: Fixed in a72b836f — `SignedSmtpRequestProjection` now stores `Rc<SmtpMessage>` instead of `Rc<SmtpRequest>`. Stage 1 → stage 2 destructure-let-else takes ownership of the message, so the "stage 2 always carries a message" invariant is encoded in the type. All three `.expect()` sites collapsed to a trivial `&self.message` getter / `proj.message()` call.

src/internet_identity/src/email_recovery/typestate.rs
aterga: Fixed in a72b836f — `winning_dkim_domain` → `dkim_domain` and `winning_checks` → `checks`. Once a caller holds a `Verified` value the multi-signature concept is over; the prefix was noise.

src/internet_identity/src/email_recovery/typestate.rs
aterga: Fixed in a72b836f — renamed to `verify_signature`. The dkim-primitive `verify_signature` (RSA/Ed25519 byte-check) is now referenced via fully-qualified `crate::dkim::verify_signature` at the one call site, dropping it from the `use` list to avoid the name clash.

src/internet_identity/src/dmarc/verify.rs
aterga: Addressed in 88a11636 — `compute_outcome` now lives in `dmarc/mod.rs` (with its unit tests alongside). Stage 3 calls `crate::dmarc::compute_outcome(...)` directly.

src/internet_identity/src/email_recovery/typestate.rs
aterga: Addressed in 88a11636 — DKIM per-signature math moved to `dkim::run_signature_check`, taking primitive args (sig, headers, body, dkim_header_value, dkim_txt, now). The typestate `verify_signature` collapsed to ~20 lines of orchestration: call dkim, then `dmarc::extract_from_domain` for the From-domain DMARC alignment needs.

src/internet_identity_interface/src/internet_identity/types/smtp.rs
```diff
@@ -239,10 +249,93 @@ pub fn validate_smtp_request(request: &SmtpRequest) -> Result<(), SmtpResponse>
     validate_envelope(envelope)?;
     if let Some(message) = &request.message {
         validate_message(message)?;
+        validate_message_header_uniqueness(message)?;
```
sea-snake: Why are these still two separate methods?

src/internet_identity/src/email_recovery/typestate.rs
```diff
@@ -0,0 +1,859 @@
...
+    /// Raw body bytes — the same bytes whose canonical form the winning
+    /// signature's `bh=` validated. Returned by `body()` so callers can
+    /// parse the message body content (e.g. extract nonces, render
+    /// rich content).
+    body: Vec<u8>,
+    /// DKIM signing domain (`d=`). `pub(crate)` so the `dkim`/`dmarc`
+    /// test-vector modules can assert which signer was admitted. Not
+    /// exposed publicly — downstream code outside the crate reads
+    /// message content via `header()` and `body()` only.
+    pub(crate) dkim_domain: String,
```
sea-snake: Can we avoid pub(crate) fields?

src/internet_identity_interface/src/internet_identity/types/smtp.rs
aterga: Addressed in 59e8ee58 (amends the WIP). `validate_message_header_uniqueness` is gone from the wire-types crate entirely; §3.6 enforcement moved into the typestate as type-system invariants on `unverified::SmtpMessage` (required `String` fields for exactly-once headers, `Option<String>` for at-most-once, `Vec<String>` for `DKIM-Signature`). The wire-types crate keeps only `validate_message` for bounds — single method, single concern.

src/internet_identity/src/email_recovery/typestate.rs
aterga: Addressed in 59e8ee58. `VerifiedSmtpRequest` fields are now private; the four crate-private accessors (`dkim_domain()`, `mail_from_domain()`, `dmarc_outcome()`, `checks()`) expose read-only views for the test-vector modules without letting any caller poke at internal state. (`mail_from_domain` rather than `from_domain` to dodge clippy `wrong_self_convention` on `from_*` names.)

aterga:
Merged `main` (incl. #3948) with a deliberate inversion of direction: rather than keep the new wire-types `validate_header_occurrences`, this merge **reverts** main's §3.6 additions and consolidates the enforcement inside the canister-side typestate. The two crates land on opposite sides of the API-type vs inner-type split — wire-types stays bounds-only (`validate_envelope`, `validate_message`, `validate_smtp_request`); the typestate owns §3.6 well-formedness through `unverified::SmtpMessageBuilder`.

Each stage's inner type now reflects what got checked:

- `unverified::SmtpRequest` carries an optional `SmtpMessage` whose §3.6 single-occurrence headers occupy typed slots (`From`/`Date` required, the rest `Option<String>`). Uniqueness/missing checks are type-system invariants. Everything else, including `DKIM-Signature`, lives in `other_headers` in source order — DKIM isn't §3.6, so stage 1 doesn't route it specially.
- `signed::SmtpRequest` reflects "≥1 parsed signature" via a `first + rest` shape rather than a wrapped `Vec`.
- `verified::SmtpRequest` moves into its own submodule for symmetry.

Errors are Rust enums (`SmtpError`/`EnvelopeError`/`MessageError` for stage 1, `DkimScopeError` for stage 2) with `From<…> for SmtpResponse` adapters at the canister boundary. Duplicate-§3.6 detection aggregates every offender into a single `DuplicateUniqueHeaders(Vec<&'static str>)` so one SMTP 555 response can name them all.

aterga:
Merged `main` (incl. #3992 and #3995). This was the deepest reconciliation so far because #3992 restructured the same flow this PR refactors. Resolution direction: **main's architecture wins on flow, the typestate wins on how stage 1 is expressed.**

- `handle_smtp_request` is sync again: both paths stash a `PartialVerification` and return, with status `NeedDkimLeaf { selector }` (DNSSEC) or `ResolvingDoh` (DoH). The branch's `verify_setup_email_doh` / `complete_doh_path` / `finalize_doh_success` are deleted — superseded by the partial-based completion in `submit_leaf::resolve_via_doh`.
- `prepare_partial_verification` keeps the typestate signature (consumes `UnverifiedSmtpRequest`, picks the bottom-most signature anchored in the registered zone) and now serves both paths, matching main's unified shape.
- The stage-1-first entry points survive: `UnverifiedSmtpRequest::try_from` is still the first statement of both handlers and subsumes `validate_smtp_request`.
- Kept `parse_from_address(&str)` over main's `extract_from_address(&SmtpMessage)` — the typed message exposes the single-occurrence From value directly; nothing else referenced the wire-shape variant.

One consequence worth noting: stage 3 (`VerifiedSmtpRequest`) is no longer on the production path — completion runs off the stashed partial in `submit_leaf`. It remains exercised by the dkim/dmarc test-vector suites. If we want production to keep a stage-3 consumer, that's a follow-up discussion, not something this merge changes.

PR #3951, opened 2026-05-27
fix(fe): restore Start over in cross-device passkey pairing

MRmarioruci:
When a user reaches the "Authorize new device" screen via the sign-in fallback ("Can't find your identity?" → `/activate#<id>` → `/manage/access?activate=<id>`) and clicks **Start over**, nothing useful happens — the wizard reuses the original `registrationId` instead of issuing a fresh one, so no new `/pair#<id>` link is generated and the dialog gets stuck on the "Waiting for new device" spinner with no escape.

## Changes

- The confirm access-method wizard now always generates a fresh `registrationId` (and a new `/pair#<id>` link) when the user clicks **Start over**, regardless of how the wizard was opened. Previously it captured the `registrationId` prop and re-passed it to `enterRegistrationMode`, which short-circuited the new-link generation.
- `/manage/access` drops the `?activate=<id>` URL parameter on Start over so a page refresh doesn't reopen the stale pending session.
- Unrelated: centers the `ConfirmThisDevice` view content (bundled with this branch).

## Tests

- Vitest unit tests on `ConfirmAccessMethodFlow.enterRegistrationMode` covering:
  - calling with no arg generates a `/pair#<id>` link
  - calling with an existing `registrationId` does not regenerate the link
  - a subsequent no-arg call after a previous arg call produces a fresh link with a different id (the Start over path)

### DIFF
--- a/src/frontend/src/lib/components/wizards/confirmAccessMethod/ConfirmAccessMethodWizard.svelte
+++ b/src/frontend/src/lib/components/wizards/confirmAccessMethod/ConfirmAccessMethodWizard.svelte
@@ -11,16 +11,29 @@
   interface Props {
     registrationId?: string;
     onConfirm: () => void;
+    onRestart?: () => void;
     onError: (error: unknown) => void;
   }
 
-  const { registrationId, onConfirm, onError }: Props = $props();
+  const { registrationId, onConfirm, onRestart, onError }: Props = $props();
 
   const confirmAccessMethodFlow = new ConfirmAccessMethodFlow();
 
   const handleEnterRegistrationMode = async () => {
     try {
       await confirmAccessMethodFlow.enterRegistrationMode(registrationId);
+    } catch (error) {
+      onError(error);
+    }
+  };
+  // Always generate a fresh registrationId on user-initiated restart,
+  // even if the wizard was mounted with one (e.g. via /activate#<id>).
+  // Reusing the same id leaves #newDeviceLink unset and traps the dialog
+  // in the WaitingForNewDevice view.
+  const handleRestart = async () => {
+    onRestart?.();
+    try {
+      await confirmAccessMethodFlow.enterRegistrationMode();
     } catch (error) {
       onError(error);
     }
@@ -56,7 +69,7 @@
 {:else if confirmAccessMethodFlow.view === "enterConfirmationCode"}
   <EnterConfirmationCode
     confirm={handleConfirmDevice}
-    restart={handleEnterRegistrationMode}
+    restart={handleRestart}
   />
 {:else if confirmAccessMethodFlow.view === "finishOnNewDevice"}
   <FinishOnNewDevice />
--- a/src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.test.ts
+++ b/src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.test.ts
@@ -1,0 +1,78 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+vi.mock("$app/environment", () => ({ browser: true }));
+
+// Hand-rolled Svelte-store contract so vi.hoisted can construct it
+// synchronously before the mock factory below runs.
+const authStore = vi.hoisted(() => {
+  let value: unknown = undefined;
+  const subs = new Set<(v: unknown) => void>();
+  return {
+    subscribe: (fn: (v: unknown) => void) => {
+      subs.add(fn);
+      fn(value);
+      return () => {
+        subs.delete(fn);
+      };
+    },
+    set: (v: unknown) => {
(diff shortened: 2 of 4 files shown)


src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.test.ts
```diff
@@ -0,0 +1,79 @@
...
+});
+
+vi.mock("$lib/stores/authentication.store", () => ({
+  authenticatedStore: authStore,
+}));
+
+import { ConfirmAccessMethodFlow } from "./confirmAccessMethodFlow.svelte";
+
+const makeActor = () => ({
+  authn_method_registration_mode_exit: vi.fn(async () => ({ Ok: null })),
```
sea-snake: For this particular lint error where you need an async function but don't have any actual `await` inside it, to justify it, you can have a non async function that returns e.g. `Promise.resolve(x)` as value.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
```diff
@@ -64,6 +64,9 @@
   let switchingAccessMethodKey = $state<string>();
   let accessMethods = $derived(toAccessMethods(data.identityInfo));
   let pendingRegistrationId = $state(data.pendingRegistrationId);
+  let isConfirmingPendingRegistration = $state(
```
sea-snake: Nit: I would name this `showRegistrationDialog` for clarity that it solely controls the visibility of the dialog.

src/frontend/src/lib/components/wizards/confirmAccessMethod/views/WaitingForNewDevice.svelte
```diff
@@ -12,10 +19,16 @@
...
     <Trans>
       The <b class="text-text-primary">new device</b> is preparing to continue.
     </Trans>
   </p>
+  {#if restart}
+    <button class="btn btn-secondary btn-xl" onclick={restart}>
+      <RotateCcwIcon class="size-5" />
+      <span>{$t`Start over`}</span>
+    </button>
+  {/if}
```
sea-snake: I wouldn't add a restart button on the waiting step, one could always just close and re-open the dialog if needed. The restart button mostly makes sense on the code entry screen, since here the user might want to see a clear one-click action to restart if they have security concerns.

MRmarioruci:
Addressed review feedback (force-pushed):

- **WaitingForNewDevice restart button** — reverted per @sea-snake's comment. Agreed that the code-entry screen is the right place for the explicit one-click bail-out; users on the waiting view can just close the dialog.
- **`isConfirmingPendingRegistration` → `showRegistrationDialog`** — renamed for clarity.
- **`require-await` lint error in the test** — switched the actor mocks from `async () => ({...})` to `() => Promise.resolve({...})`. This was also the cause of the failing `frontend-checks` CI job.

src/frontend/src/lib/flows/confirmAccessMethodFlow.svelte.test.ts
MRmarioruci: Done — switched to `Promise.resolve`.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
MRmarioruci: Renamed.

src/frontend/src/lib/components/wizards/confirmAccessMethod/views/WaitingForNewDevice.svelte
MRmarioruci: Reverted.

sea-snake (approved):

src/frontend/src/lib/components/wizards/confirmAccessMethod/ConfirmAccessMethodWizard.svelte
```diff
@@ -25,6 +26,18 @@
       onError(error);
     }
   };
+  // Always generate a fresh registrationId on user-initiated restart,
+  // even if the wizard was mounted with one (e.g. via /activate#<id>).
+  // Reusing the same id leaves #newDeviceLink unset and traps the dialog
+  // in the WaitingForNewDevice view.
+  const handleRestart = async () => {
+    onRestart?.();
```
aterga, line 34: What does this `.` do? (Unfamiliar syntax)

MRmarioruci: Optional chaining on the function call. Like saying if (onRestart != null) {}

PR #3955, opened 2026-05-27
fix(ci): unbreak review-request Slack ping for private-org-member PRs

aterga:
# Changes

The `pr-review-requested` workflow gated its only job on
`pull_request.author_association in (MEMBER, OWNER, COLLABORATOR)`. GitHub's
webhook payload, though, derives `author_association` from *publicly
visible* org relationships only. dfinity org members whose membership is
private (e.g. @MRmarioruci, @sea-snake) come through as `CONTRIBUTOR`, so
the job was silently skipped and no Slack ping fired when a reviewer was
added to their PRs.

Concrete impact on #3929: when @sea-snake was requested as a reviewer the
runner evaluated `'CONTRIBUTOR' == 'MEMBER' || == 'OWNER' || == 'COLLABORATOR'`
→ false and skipped before reaching the Slack step. Looking at history,
every `review_requested` event on a private-org-member-authored PR has
skipped this way — 0/5 successes for @MRmarioruci, 0/36+ for @sea-snake.

Dropping the outer guard. The step-level `requested_reviewer.login`
allowlist (now @aterga/@sea-snake/@MRmarioruci) is what actually restricts
pings, and the workflow checks out at `pull_request.base.sha` so no
PR-supplied code is executed.

Also refreshed the reviewer allowlist to match the current Identity team:
dropped @danblackadder and @lmuntaner (no longer on the team), added
@MRmarioruci.


.github/workflows/pr-review-requested.yml
sea-snake: This list should be updated too, right now it's not mentioning @MRmarioruci  whiile mentioning ex-members of the identity team.

aterga: Good catch — refreshed the allowlist in e1f1a134: dropped @danblackadder and @lmuntaner, added @MRmarioruci.

MRmarioruci (approved):
An alternative way of doing this would be utilizing the CODEOWNERS file and adding/removing people from that, but this lgtm

PR #3956, opened 2026-05-28
feat(fe): add /cli authorize route, per-device CLI access setting, and loopback form-POST delivery

sea-snake:
Moves the CLI sign-in flow currently hosted at cli.id.ai into Internet Identity itself at id.ai/cli, and adds a per-identity, device-local toggle on the new manage Settings page that gates app-mode CLI sign-in.

## Changes

**New `/cli` route** — `routes/(new-styling)/cli/`
- `+page.svelte` orchestrates the sign-in phases: `wizard` (new user / no last-used identity), `authorize` (consent), `close` (success), `cli-disabled` (app mode with CLI access off), plus terminal `invalid` (malformed request) and `error` (the CLI rejected the delegation) screens.
- A returning user with a recently-used identity opens directly on the `authorize`/Continue screen for that identity — mirroring `/authorize` — instead of being shown the sign-in method wizard. An `identity-mismatch` redirect stays on the Continue screen too (with a toast), so the user picks the right identity from the header switcher rather than being dropped back into the method wizard.
- `+page.ts` validates URL params with native APIs (`new URL`, `fromBase64URL`) — no regex. Accepts `public_key` (base64url DER), `callback` (loopback only — 127.0.0.1 or ::1, per RFC 8252), `nonce` (the CLI's single-use secret), optional `domain` (the delegation domain to get an identity for), optional `ttl` (minutes; default 480 to match cli.id.ai), and parses the `status` the loopback server redirects back with.
- `utils.ts` (`cliAuthorize`) calls `prepare_account_delegation` / `get_account_delegation` to build a two-hop chain, then delivers it to the loopback callback via a **top-level form POST** rather than `fetch`. A `fetch` from a public origin to a loopback address triggers Chrome's Local Network Access permission prompt; a top-level navigation does not. The form echoes back the `nonce` from the URL fragment so the loopback server can tell this page's POST from a stray or forged local request. The derivation origin is `https://<domain>` when `domain` is supplied; with no `domain` it falls back to `https://cli.id.ai`, so existing CLI principals stay valid after the migration.
- The loopback server 303-redirects the browser back to `/cli#status=…`, so id.ai keeps ownership of the outcome UI:
  - `success` → close-window screen.
  - `identity-mismatch` (CLI `login` re-auth) → a toast plus the Continue screen again, so the user can retry with the correct identity in place.
  - `error` → a generic error screen pointing the user to the CLI (the loopback server has shut down, so there is nothing to retry).
- The header identity switcher only *selects* an identity; the Continue button is the single sign-in point (it authenticates the selected identity, then builds and submits the delegation). The switcher is shown only while signing in (wizard, authorize) and hidden on the terminal screens.
- `components/` — `CliHeader`, `TerminalBlock`. `views/` — `CliAuthorizeView`, `CliCloseWindowView`, `CliErrorView`. Reuses the production `AuthWizard` for new users. The terminal block matches the design (the "Requested by" badge straddles the top edge; the `$` is separated from the command; rows align), and shows `icp identity link ii [--app <domain>]` — the `<name>` positional is omitted because the page can't know it and a placeholder would read as literal. The progress line shown while the delegation is in flight reads "Linking Internet Identity" to match the CLI's spinner (dfinity/icp-cli#575).

**CSP** — `internet_identity_frontend`
- The app-wide Content-Security-Policy set `form-action 'none'`, which blocked the form POST above. Relaxed to `form-action 'self' http://127.0.0.1:* http://localhost:* http://[::1]:*` — same origin plus http loopback only, so a form can never post to a remote origin.

**Manage Settings page** — `routes/(new-styling)/manage/(authenticated)/settings/`
- `+page.svelte` with a CLI access toggle card.
- `components/CliAccessSection`, `CliConfirmDialog`, `VerifyPanel`, `CopyCommandRow`. Confirm dialog when enabling: warning + install command (`npm install -g @icp-sdk/icp-cli`, click-to-copy with a "Copied to clipboard" tooltip matching the existing pattern) + checkbox acknowledgement.
- New `Settings` entry in the manage sidebar.

**Storage** — `ii-cli-access` localStorage key, per identity number. Purely client-side — no backend involvement. The threat model is the user's own opt-in per device.

**Analytics** — `lib/utils/analytics/cliAuthorizeFunnel.ts`
- A Plausible funnel tracks the `/cli` flow end to end: `start-cli-authorize` on entry, `request-received` / `request-invalid`, `confirmed` (Continue / Allow access), the `cli-disabled` gate, and the `success` / `error` / `identity-mismatch` outcomes the loopback server redirects back with. Because the flow navigates away to the loopback server and back, the outcome events fire on a fresh page load and Plausible correlates them with the start by visitor session.

## Tests

- [x] e2e: invalid params show the invalid-request screen.
- [x] e2e: non-loopback callback is rejected.
- [x] e2e: generic mode posts a delegation chain to the loopback callback and shows the close-window page.
- [x] e2e: a returning user with a last-used identity opens on the Continue screen, not the method wizard.
- [x] e2e: app mode without CLI access lands on the gated error screen.
- [x] e2e: enabling CLI access in /manage/settings unlocks app-mode sign-in.
- [x] e2e: an identity mismatch keeps the Continue screen (with a toast) and lets the user retry in place.
- [x] e2e: `--app`/`domain` derives a different root principal than generic mode for the same identity (proves the delegation domain drives the derivation origin).

The `cli` fixture stands up an actual 127.0.0.1 HTTP server per test (OS-picked port, so concurrent fixtures never collide) to receive the form-POSTed delegation and 303-redirect back to /cli with a status. It generates an Ed25519 keypair to stand in for the CLI's session key, supplies the `nonce` in the fragment, rejects a POST that doesn't echo it back (mirroring the real CLI's gate), and collects every received delegation so a test can compare derivation origins.

Pairs with the CLI side in dfinity/icp-cli#575 (form-POST callback + status redirect + principal check + nonce gate + `--auth`/`--app`/`domain` contract). cli.id.ai (which today does the postMessage middleman) can be retired once CLI binaries point at id.ai/cli.


src/frontend/src/routes/(new-styling)/cli/+page.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (url.protocol !== "http:" && url.protocol !== "https:") {
+    return false;
+  }
+  if (url.username !== "" || url.password !== "") {
+    return false;
+  }
+  return url.hostname === "127.0.0.1" || url.hostname === "::1";
+};
+
+const isValidPublicKey = (raw: string): boolean => /^[0-9a-fA-F]+$/.test(raw);
```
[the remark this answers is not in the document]
sea-snake: Resolved in 3cdc005 — the regex was replaced with `parseHex()`, which uses `fromHex()` to validate. `fromHex` rejects odd-length strings, so `public_key=a` now lands on the invalid-request screen.

src/frontend/tests/e2e-playwright/fixtures/cli.ts
```diff
@@ -0,0 +1,58 @@
...
+        try {
+          resolveDelegation(JSON.parse(raw));
+        } catch {
+          resolveDelegation(raw);
+        }
+        res.statusCode = 200;
+        res.setHeader("Access-Control-Allow-Origin", "*");
+        res.end("ok");
+      });
+    });
```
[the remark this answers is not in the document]
sea-snake: Good catch — handled in f87ad91. The fixture now replies to OPTIONS with `Access-Control-Allow-Methods: POST, OPTIONS` and `Access-Control-Allow-Headers: Content-Type` (204, no body), and only the subsequent POST resolves `receivedDelegation`.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/CliAccessSection.svelte
```diff
@@ -0,0 +1,86 @@
...
+        </Trans>
+      {/if}
+    </p>
+  </div>
+
+  <div class="shrink-0">
+    <Toggle
+      checked={enabled}
+      onchange={(e) => handleToggle((e.target as HTMLInputElement).checked)}
+    />
```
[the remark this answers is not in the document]
sea-snake: Fixed in f87ad91 — added `aria-labelledby="cli-access-section-title"` on the `<Toggle>` and gave the section `<h3>` that id. Also dropped the `as HTMLInputElement` cast in the change handler in favour of an `instanceof HTMLInputElement` guard.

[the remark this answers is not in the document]
sea-snake: Switched to Svelte 5's `$props.id()` rune in 18221fa rather than a hand-rolled static id — same labelledby wiring, no risk of id collisions if the section ever renders twice.

src/internet_identity_frontend/src/main.rs
```diff
@@ -327,7 +331,7 @@ fn get_content_security_policy(
          img-src 'self' data: https://*.googleusercontent.com;\
          script-src {script_src};\
          base-uri 'none';\
-         form-action 'none';\
+         form-action 'self' http://127.0.0.1:* http://localhost:* http://[::1]:*;\
```
MRmarioruci: nit: the FE parser only accepts `127.0.0.1` and `::1` per RFC 8252 (`localhost` can resolve to non-loopback in odd configs, so it's intentionally rejected). `http://localhost:*` here is dead allowance — no parser-approved callback can ever use it — but it still broadens `form-action` for any other form on the origin. could drop it to keep CSP and parser in sync.

src/frontend/src/routes/(new-styling)/cli/+page.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (raw === null || raw === "") {
+    return undefined;
+  }
+  let url: URL;
+  try {
+    url = new URL(raw);
+  } catch {
+    return undefined;
+  }
+  if (url.protocol !== "http:" && url.protocol !== "https:") {
```
MRmarioruci: nit: `https://127.0.0.1:8443/cb` gets through here, but the CSP only allows `http://` loopback, so the form-POST would die at submit time with a CSP error and no UI feedback. RFC 8252 also wants http for loopback (the listener can't realistically have a CA-trusted cert). could drop `https:` from this check to fail fast.

src/internet_identity_frontend/src/main.rs
sea-snake: Good call — dropped `http://localhost:*` from `form-action` in 94024060, so the CSP now matches what the `/cli` callback parser actually accepts (`127.0.0.1`/`::1` only). Kept `http://127.0.0.1:*` and `http://[::1]:*`.

src/frontend/src/routes/(new-styling)/cli/+page.ts
sea-snake: Fixed in 94024060 — `parseLoopbackCallback` now accepts `http:` only. An `https://127.0.0.1` callback used to pass validation then silently die on the `form-action` CSP at submit; it now lands on the invalid-request screen up front. (RFC 8252 wants http for loopback anyway.)

src/frontend/src/routes/(new-styling)/cli/+page.ts
```diff
@@ -0,0 +1,148 @@
+import type { PageLoad } from "./$types";
+import { fromBase64URL } from "$lib/utils/utils";
+
+/** Default delegation lifetime in minutes. */
+const DEFAULT_TTL_MINUTES = 480;
+
+export type CliParams =
+  | {
+      kind: "valid";
```
aterga, line 17: what is this for?

src/frontend/src/routes/(new-styling)/cli/+page.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  | { kind: "invalid" };
+
+/**
+ * Outcome the loopback server redirects back with after receiving the
+ * delegation. `success` and `error` arrive on their own; `identity-mismatch`
+ * arrives alongside `public_key`/`callback` so the authorize screen can be
+ * re-shown for an in-place retry.
+ */
+export type CliStatus = "success" | "identity-mismatch" | "error";
```
aterga, line 44: the doc string should provide an example scenario for when we expect to observe `identity-mismatch`

src/internet_identity_frontend/src/main.rs
```diff
@@ -327,7 +333,7 @@ fn get_content_security_policy(
          img-src 'self' data: https://*.googleusercontent.com;\
          script-src {script_src};\
          base-uri 'none';\
-         form-action 'none';\
+         form-action 'self' http://127.0.0.1:* http://[::1]:*;\
```
aterga, line 336: Why are we changing this now?

src/frontend/tests/e2e-playwright/routes/cli.spec.ts
```diff
@@ -0,0 +1,437 @@
...
+
+  // Same identity, two derivation origins → two distinct root principals. If
+  // `domain` were ignored, these would collide.
+  expect(cli.receivedDelegations.length).toBe(2);
+  expect(rootPublicKey(cli.receivedDelegations[1])).not.toBe(
+    rootPublicKey(cli.receivedDelegations[0]),
+  );
+});
+
+test("`--app` links the same principal that /authorize gives for that app", async ({
```
aterga, line 391: ❤️

aterga (approved):

src/frontend/src/routes/(new-styling)/cli/+page.ts
sea-snake: Added a docstring on CliParams in 0e40c649 — it is the parsed /cli request from the URL fragment: the valid variant carries the validated fields (session public key, loopback callback, nonce, ttl, optional delegation domain) and the invalid variant drives the invalid-request screen.

src/frontend/src/routes/(new-styling)/cli/+page.ts
sea-snake: Done in 0e40c649 — the CliStatus docstring now spells out the scenario: the `icp identity login` re-auth path, where the user signs in with a different II identity than the one originally linked, so the delegation resolves to a different principal and the CLI redirects back here to retry with the correct identity.

src/internet_identity_frontend/src/main.rs
sea-snake: It is required for the /cli delegation delivery in this PR: the page submits the delegation to the loopback callback via a top-level form POST (chosen over fetch to avoid Chrome Local Network Access prompts), and the app-wide `form-action none` blocked all form submissions. This PR is the first thing to POST a form on the II origin, so the allowance is added here and scoped to self plus http loopback (127.0.0.1 / ::1) only — already tightened to drop the dead localhost entry per the earlier nit.

PR #3957, opened 2026-05-29
fix(be): Microsoft sign-in intermittently fails to verify the token

aterga:
Signing in with a Microsoft account intermittently fails with **"Authorization invalid — There was an error verifying your account."** The OAuth flow completes and Microsoft returns a valid token, but the II canister rejects it during JWT signature verification because the token's signing key (`kid`) isn't in the canister's cached JWKS.

We fetch Microsoft's signing keys from `https://login.microsoftonline.com/common/discovery/v2.0/keys` every ~15 minutes and **replace** the cache wholesale on each successful fetch. During a Microsoft key rollover, `/common/` returns an inconsistent subset of keys (observed flapping between 6 and 9 keys across requests, and it can even differ across subnet nodes within a single HTTPS outcall). When a thin response omits a key that is still signing live tokens, the wholesale replace drops it, and every token signed with that key then fails verification until a later fetch happens to include it again. Personal Microsoft accounts were hit hardest because their consumer signing keys are the ones `/common/` tends to drop. The bug stayed dormant for months because it only surfaces during a rollover — in steady state `/common/` returns a complete, stable set.

## Changes
- Merge freshly fetched JWKS into the cache keyed by `kid` (refresh known keys, append new ones) instead of replacing the whole set. The cache now converges to the union of every key seen, so a partial `/common/` response can no longer evict a key that is still in use.

Note: merged keys are never evicted, so the cache also retains keys Microsoft has rotated out. In practice that is a handful of harmless keys (matched by `kid`); a follow-up could bound staleness or validate against the tenant-specific issuer metadata to sidestep `/common/`'s inconsistency entirely.

## Tests
- Added a unit test asserting that a later thin fetch (missing a previously cached `kid`) does not evict it and still adds new keys.

### DIFF
--- a/src/internet_identity/src/openid/generic.rs
+++ b/src/internet_identity/src/openid/generic.rs
@@ -871,6 +871,29 @@ fn compute_next_certs_fetch_delay<T, E>(
     }
 }
 
+/// Merge freshly fetched JWKS into the cache keyed by `kid`: refresh known keys
+/// and append new ones, but never drop a key we've already seen.
+///
+/// Microsoft's `/common/` JWKS endpoint serves an inconsistent subset of its
+/// signing keys during key rollovers (and can even disagree across subnet nodes
+/// within a single outcall). Replacing the cache wholesale lets such a thin
+/// response evict a key that is still signing live tokens, which breaks sign-in
+/// for the affected accounts until a later fetch happens to include it again.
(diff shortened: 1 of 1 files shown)


sea-snake:
Superseded by #3959

PR #3959, opened 2026-05-29
feat(be): seed OpenID JWK cache from install args and persist it across upgrades

aterga:
## What

Adds `seed_jwks : opt vec vec record { text; text }` to `OpenIdConfig` in the II backend install args, and makes each provider's JWK cache persistent across canister upgrades using a dedicated stable structure.

## Why

Today each hardcoded OpenID provider (`Provider` in `openid/generic.rs`) keeps its JWK cache as an in-memory `Rc<RefCell<Vec<Jwk>>>` that is **lost on every upgrade** and only repopulated by the periodic `jwks_uri` HTTP fetch. This leaves a window after install/upgrade where JWT verification fails until the first fetch completes. `seed_jwks` lets the cache be pre-populated at install time, and persisting the cache removes the post-upgrade gap (and avoids needing to mock the certs HTTP fetch in tests).

## Seed format

`seed_jwks` is an **optional set of JWKs**: the outer vec is the set, and each inner vec is **one JWK** represented as the list of its JSON `(field, value)` pairs (e.g. `("kty","RSA")`, `("kid","…")`, `("n","…")`, `("e","AQAB")`). Each inner list is assembled into a JSON object and parsed with the same `serde` path used for fetched certs, so a seeded key is byte-for-byte identical to a fetched one. Invalid entries are logged and skipped.

## Persistence & precedence

A new issuer-keyed `StableBTreeMap<String, StorableJwks>` (memory id `24`) holds each provider's JWK cache. `initial_certs` runs on `init`/`post_upgrade`:

- **If the config supplies any valid `seed_jwks`, they are authoritative** — they're persisted to stable memory and used as the cache.
- **Otherwise**, the cache falls back to the JWKs already persisted for that issuer (from an earlier seed or fetch), so verification keeps working across upgrades before the first post-upgrade fetch.

Every successful periodic re-fetch writes its keys through to the same map, so for providers without a seed the latest fetched keys survive upgrades. Discoverable SSO providers are unchanged — they re-derive keys from discovery after each upgrade and pass `None` for the persistence key.

## Frontend

`OpenIdConfig` is part of the config served to the frontend, so the committed Candid bindings are regenerated (`didc bind`) and the hand-written `globals.ts` mirror is kept structurally identical to the generated type (so `findConfig`'s cross-type return still type-checks).

## Testing

- Unit: `build_seed_jwk` (pairs → JWK, identical to fetched, rejects missing `kty`), `build_seed_jwks` (multiple keys + skip invalid), `StorableJwks` round-trip, and `storage::should_persist_openid_jwks_across_reload`.
- Candid: `check_candid_interface_compatibility` passes.
- Integration: `can_link_google_account_with_seeded_jwks` seeds **two** Google keys (a decoy + the signing key) and links a Google account using only the seeded cache — **no** certs HTTP response mocked — exercising the multi-key path and by-`kid` matching.
- Full backend unit suite (544 tests), clippy, and the full `frontend-checks` set (`tsc`, `svelte-check`, `eslint`, `prettier`) all pass.

## Compatibility

Backward compatible: the new memory id is allocated lazily on upgrade, and the new optional `seed_jwks` field decodes to `None` for existing persisted state.

https://claude.ai/code/session_01Rwvp6fpwc4iBvRCa7h8QNo

### DIFF
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -432,6 +432,13 @@ type OpenIdConfig = record {
     auth_scope : vec text;
     fedcm_uri : opt text;
     email_verification : opt OpenIdEmailVerification;
+    // Optional initial set of JWKs used to seed this provider's JWK cache on
+    // install, so JWT verification works before the first jwks_uri fetch and
+    // across upgrades (the cache is persisted in stable memory). Each entry is
+    // one JWK represented as the list of its JSON (field, value) pairs, e.g.
+    // vec { record { "kty"; "RSA" }; record { "kid"; "..." };
+    //       record { "n"; "..." }; record { "e"; "AQAB" } }.
+    seed_jwks : opt vec record { text; text };
 };
 
 // SSO provider config that uses two-hop discovery.
--- a/src/internet_identity/internet_identity.did
+++ b/src/internet_identity/internet_identity.did
@@ -434,11 +434,12 @@ type OpenIdConfig = record {
     email_verification : opt OpenIdEmailVerification;
     // Optional initial set of JWKs used to seed this provider's JWK cache on
     // install, so JWT verification works before the first jwks_uri fetch and
-    // across upgrades (the cache is persisted in stable memory). Each entry is
-    // one JWK represented as the list of its JSON (field, value) pairs, e.g.
-    // vec { record { "kty"; "RSA" }; record { "kid"; "..." };
-    //       record { "n"; "..." }; record { "e"; "AQAB" } }.
-    seed_jwks : opt vec record { text; text };
+    // across upgrades (the cache is persisted in stable memory). The outer vec
+    // is the set of JWKs; each inner vec is one JWK, given as the list of its
+    // JSON (field, value) pairs, e.g. a single RSA key is
+    // vec { vec { record { "kty"; "RSA" }; record { "kid"; "..." };
+    //             record { "n"; "..." }; record { "e"; "AQAB" } } }.
+    seed_jwks : opt vec vec record { text; text };
 };
 
 // SSO provider config that uses two-hop discovery.
--- a/src/internet_identity/src/openid/generic.rs
+++ b/src/internet_identity/src/openid/generic.rs
@@ -216,16 +216,73 @@ impl Provider {
         let certs = Rc::new(RefCell::new(TEST_CERTS.take()));
 
         #[cfg(not(test))]
-        let certs: Rc<RefCell<Vec<Jwk>>> = Rc::new(RefCell::new(vec![]));
+        let certs: Rc<RefCell<Vec<Jwk>>> = Rc::new(RefCell::new(initial_certs(&config)));
 
+        // Persist fetched keys under the provider's `issuer` so the cache
+        // survives upgrades (see `initial_certs`).
         #[cfg(not(test))]
-        schedule_fetch_certs(config.jwks_uri, Rc::clone(&certs), Some(0));
+        schedule_fetch_certs(
+            Some(config.issuer.clone()),
+            config.jwks_uri,
+            Rc::clone(&certs),
+            Some(0),
+        );
 
         Provider {
             client_id: config.client_id,
             issuer: config.issuer,
             certs,
             email_verification: config.email_verification,
+        }
+    }
+}
+
+/// Build a single [`Jwk`] from the `(field, value)` pairs supplied via
+/// [`OpenIdConfig::seed_jwks`]. Each pair is one field of the JWK's JSON object
+/// (e.g. `("kty","RSA")`, `("kid","...")`, `("n","...")`, `("e","AQAB")`). The
+/// pairs are assembled into a JSON object and parsed with the same `serde` path
+/// used for fetched certs, so a seeded key is byte-for-byte identical to a
+/// fetched one.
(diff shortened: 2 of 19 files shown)


src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -404,6 +404,19 @@ pub struct OpenIdConfig {
...
+    /// backend's storage layer).
+    ///
+    /// Each entry is one JWK represented as the list of its JSON
+    /// `(field, value)` pairs, e.g.
+    /// `vec { record { "kty"; "RSA" }; record { "kid"; "..." };
+    ///        record { "n"; "..." }; record { "e"; "AQAB" } }`.
+    /// The pairs are assembled into a JSON object and parsed with the same
+    /// path used for fetched certs, so all string-valued JWK fields are
+    /// supported.
+    pub seed_jwks: Option<Vec<(String, String)>>,
```
sea-snake: Looking at the type, this seems to only support a single JWK value it's field, not multiple JWK values (which we need).

src/internet_identity/src/openid/generic.rs
```diff
@@ -230,6 +237,56 @@ impl Provider {
...
+///    immediately after an upgrade, before the first post-upgrade fetch.
+/// 2. Otherwise, if `seed_jwks` is provided, bootstrap from it and persist the
+///    result so it too survives the next upgrade.
+///
+/// `seed_jwks` only takes effect when nothing is persisted yet; once the cache
+/// holds keys (seeded or fetched) they are never clobbered by a stale seed on
+/// a later upgrade. Use a reinstall to reset.
+#[cfg(not(test))]
+fn initial_certs(config: &OpenIdConfig) -> Vec<Jwk> {
+    if let Some(persisted) = state::storage_borrow(|s| s.read_openid_jwks(&config.issuer)) {
```
aterga: The logic is wring, 

if we have any JWKs in `config`, those should be stored in the persistent cache and returned from this function.

Otherwise, we should return the priorly cached JWKs.

src/internet_identity/internet_identity.did
```diff
@@ -432,6 +432,13 @@ type OpenIdConfig = record {
     auth_scope : vec text;
     fedcm_uri : opt text;
     email_verification : opt OpenIdEmailVerification;
+    // Optional initial set of JWKs used to seed this provider's JWK cache on
+    // install, so JWT verification works before the first jwks_uri fetch and
+    // across upgrades (the cache is persisted in stable memory). Each entry is
+    // one JWK represented as the list of its JSON (field, value) pairs, e.g.
```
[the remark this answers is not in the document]
aterga: Fixed in 011bf25. Changed the type to `opt vec vec record { text; text }` — the outer vec is the set of JWKs and each inner vec is one JWK's JSON `(field, value)` pairs — and updated the Candid comment to match. So the shape now lines up with the "each entry is a JWK" wording and supports multiple seeded keys.

---

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -404,6 +404,19 @@ pub struct OpenIdConfig {
     pub auth_scope: Vec<String>,
     pub fedcm_uri: Option<String>,
     pub email_verification: Option<OpenIdEmailVerificationScheme>,
+    /// Optional initial set of JWKs used to seed this provider's JWK cache on
+    /// install, so JWT verification works before the first `jwks_uri` fetch
+    /// completes (and across upgrades — see the persistent JWK cache in the
+    /// backend's storage layer).
+    ///
+    /// Each entry is one JWK represented as the list of its JSON
+    /// `(field, value)` pairs, e.g.
```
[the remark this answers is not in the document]
aterga: Fixed in 011bf25. `seed_jwks` is now `Option<Vec<Vec<(String, String)>>>` and the Rustdoc describes it accordingly: the outer vec is the set of JWKs, each inner vec is one JWK's `(field, value)` pairs. `build_seed_jwk` builds one `Jwk` from a single entry, and `build_seed_jwks` maps over all entries (skipping + logging any that fail to parse).

---

src/internet_identity_interface/src/internet_identity/types.rs
aterga: Good catch — addressed in 011bf25. `seed_jwks` is now `opt vec vec record { text; text }` (`Option<Vec<Vec<(String, String)>>>`): the outer vec is the set of keys and each inner vec is a single JWK's `(field, value)` pairs, so multiple keys can be seeded. The integration test now seeds two Google keys (a decoy + the signing key) to exercise the multi-key path and the by-`kid` match during verification.

---

sea-snake (approved):

src/internet_identity/src/openid/generic.rs
aterga: Fixed in 011bf25. `initial_certs` now does exactly that: if the config supplies any valid seed JWKs, they're persisted to the stable cache and returned (authoritative); otherwise it returns the priorly cached JWKs.

One consequence worth flagging: since `config` here is the *persisted* `openid_configs`, a seed left in `openid_configs` is re-applied on every upgrade (overwriting fetched keys until the next fetch refreshes them). That's fine for the bootstrap/test use case, but if you'd prefer the seed to apply only when the persisted cache is empty (or only on a fresh install), I'm happy to adjust.

---

PR #3969, opened 2026-06-02
ci: Slack notifier for PRs awaiting CI approval

timothyaterton:
## Summary

Adds a cron-triggered workflow that detects fork-PR workflow runs stuck in `action_required` state and posts a notification to the **`#eng-identity-request-reviews`** Slack channel. This addresses the common failure mode where external-contributor PRs (including Tim's fork PRs) wait silently for CI approval because no push notification exists.

## What it does

- **Cron schedule**: Runs every 15 minutes (also supports `workflow_dispatch` for manual testing).
- **Detection**: Queries `GET /repos/{repo}/actions/runs?status=action_required` and groups by the originating PR's **source-branch HEAD commit** (not the merge commit) to avoid spurious re-notifications when the base branch advances.
- **De-duplication**: Tracks notified state via an HTML-comment marker on the PR (`<!-- ci-approval-slack: head=<sha> -->`). New head commit → fresh notification; same head, still pending → silent. Same idiom used by `nudge-codeowners`.
- **Slack payload**: PR title, PR link, author login, head SHA (short), pending-run count, and a direct link to the Actions approval UI.
- **Safe degradation**: If `SLACK_PRIVATE_IDENTITY_WEBHOOK_URL` secret is unset, the workflow logs a message and exits 0 — fork builds and first-run-after-merge are clean no-ops, never failures.
- **Reuses existing secret**: Uses `SLACK_PRIVATE_IDENTITY_WEBHOOK_URL`, same secret as `pr-review-requested.yml` — no new secret setup required.

## Files changed

| File | Description |
|------|-------------|
| `.github/workflows/ci-approval-slack-notifier.yml` | Workflow definition (cron + dispatch) |
| `.github/scripts/notify-ci-approval.sh` | Script that queries pending runs, de-duplicates, and posts to Slack |

## Test plan

- [ ] Merge this PR (workflow runs as no-op if no `action_required` runs exist)
- [ ] Run `workflow_dispatch` on `main` → verify clean exit in logs
- [ ] Wait for (or trigger) a fork PR with `action_required` runs → verify Slack message appears
- [ ] Verify de-duplication: subsequent cron ticks with the same head SHA should log "already notified, skipping"

### DIFF
--- a/.github/scripts/notify-ci-approval.sh
+++ b/.github/scripts/notify-ci-approval.sh
@@ -1,0 +1,130 @@
+#!/usr/bin/env bash
+# Notify Slack when fork-PR workflow runs are stuck in action_required.
+#
+# Requires: gh (authenticated), curl, jq
+# Env:      SLACK_CI_APPROVAL_WEBHOOK_URL (optional — exits cleanly if unset)
+#           GITHUB_REPOSITORY (e.g. dfinity/internet-identity)
+set -euo pipefail
+
+REPO="${GITHUB_REPOSITORY:?GITHUB_REPOSITORY must be set}"
+MARKER_PREFIX="<!-- ci-approval-slack:"
+
+if [[ -z "${SLACK_CI_APPROVAL_WEBHOOK_URL:-}" ]]; then
+  echo "SLACK_CI_APPROVAL_WEBHOOK_URL is not set — skipping notification."
+  exit 0
+fi
+
+# Fetch all action_required workflow runs.
+runs=$(gh api "repos/${REPO}/actions/runs?status=action_required&per_page=100" \
+  --jq '.workflow_runs')
+
+if [[ -z "$runs" || "$runs" == "null" || "$runs" == "[]" ]]; then
+  echo "No action_required runs found."
+  exit 0
+fi
+
+# Group runs by the originating PR's head SHA (source branch HEAD, not
+# the merge commit). We key on (pr_number, head_sha) to avoid spurious
+# re-notifications when the base branch advances.
+#
+# Each run has .pull_requests[] — we take the first one. Runs without a
+# PR association (e.g. branch pushes) are skipped.
+#
+# Output: one JSON object per unique (pr_number, head_sha) with fields:
+#   pr_number, head_sha, head_short, run_count, pr_url, author, title, approve_url
+grouped=$(echo "$runs" | jq -r '
+  [ .[]
+    | select(.pull_requests | length > 0)
+    | {
+        pr_number: .pull_requests[0].number,
+        head_sha:  .head_sha,
+        head_short: (.head_sha[:7]),
+        repo_url:  (.repository.html_url // ""),
+        run_html:  .html_url,
+        actor:     (.actor.login // "unknown"),
+        pr_url:    (.pull_requests[0].url // "")
+      }
+  ]
+  | group_by(.pr_number + ":" + .head_sha)
+  | [ .[] | {
+        pr_number:  .[0].pr_number,
+        head_sha:   .[0].head_sha,
+        head_short: .[0].head_short,
+        run_count:  length,
+        actor:      .[0].actor,
+        repo_url:   .[0].repo_url,
+        run_html:   .[0].run_html,
+        pr_url:     .[0].pr_url
+      }
+    ]
+')
+
+if [[ -z "$grouped" || "$grouped" == "null" || "$grouped" == "[]" ]]; then
+  echo "No action_required runs linked to open PRs."
+  exit 0
+fi
+
+# For each unique (PR, head SHA), check dedup marker and notify if new.
+notified=0
+echo "$grouped" | jq -c '.[]' | while IFS= read -r entry; do
+  pr_number=$(echo "$entry" | jq -r '.pr_number')
+  head_sha=$(echo "$entry"  | jq -r '.head_sha')
+  head_short=$(echo "$entry" | jq -r '.head_short')
+  run_count=$(echo "$entry"  | jq -r '.run_count')
+  actor=$(echo "$entry"      | jq -r '.actor')
+  run_html=$(echo "$entry"   | jq -r '.run_html')
+  pr_rest_url=$(echo "$entry" | jq -r '.pr_url')
+
+  # Fetch PR details (title, html_url) from the REST API.
+  pr_data=$(gh api "repos/${REPO}/pulls/${pr_number}" --jq '{title: .title, html_url: .html_url, state: .state}' 2>/dev/null || echo '{}')
+  pr_title=$(echo "$pr_data" | jq -r '.title // "PR #'"${pr_number}"'"')
+  pr_html=$(echo "$pr_data"  | jq -r '.html_url // "https://github.com/'"${REPO}"'/pull/'"${pr_number}"'"')
+  pr_state=$(echo "$pr_data" | jq -r '.state // "unknown"')
+
+  # Skip closed PRs — their runs are stale.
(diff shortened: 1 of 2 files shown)


.github/scripts/notify-ci-approval.sh
```diff
@@ -0,0 +1,130 @@
...
+# re-notifications when the base branch advances.
+#
+# Each run has .pull_requests[] — we take the first one. Runs without a
+# PR association (e.g. branch pushes) are skipped.
+#
+# Output: one JSON object per unique (pr_number, head_sha) with fields:
+#   pr_number, head_sha, head_short, run_count, pr_url, author, title, approve_url
+grouped=$(echo "$runs" | jq -r '
+  [ .[]
+    | select(.pull_requests | length > 0)
```
MRmarioruci: Fork-PR runs in `action_required` commonly arrive with `pull_requests: []` — the base repo can't fully populate that until the workflow is approved. So this `length > 0` filter likely drops every run the notifier exists to catch. Worth matching by `head_sha` against open PRs (`gh api repos/$REPO/pulls?state=open`) or falling back to `repos/$REPO/commits/$head_sha/pulls`. Run it once against a real fork-PR pending run before merging.

.github/scripts/notify-ci-approval.sh
```diff
@@ -0,0 +1,130 @@
...
+  # Skip closed PRs — their runs are stale.
+  if [[ "$pr_state" != "open" ]]; then
+    echo "PR #${pr_number} is ${pr_state}, skipping."
+    continue
+  fi
+
+  # Check dedup: look for our marker comment on this PR with this head SHA.
+  marker="${MARKER_PREFIX} head=${head_sha} -->"
+  existing=$(gh api "repos/${REPO}/issues/${pr_number}/comments?per_page=100" \
+    --jq '[.[] | select(.body | contains("'"${MARKER_PREFIX}"'")) | .body] | last // ""' 2>/dev/null || echo "")
```
MRmarioruci: Two issues here: `2>/dev/null` swallows 403/404/rate-limit errors silently (the dedup check then sees an empty string and re-sends every 15 min), and `?per_page=100` doesn't paginate, so PRs with >100 comments lose the marker. Drop the stderr redirect and add `--paginate`.

.github/scripts/notify-ci-approval.sh
```diff
@@ -0,0 +1,130 @@
...
+  # Build the approval-UI link (Actions tab filtered to the PR).
+  approve_url="https://github.com/${REPO}/actions?query=event%3Apull_request_target+is%3Aaction_required"
+
+  # Post Slack notification.
+  slack_text=":warning: *CI approval needed* for <${pr_html}|${pr_title}> by \`${actor}\`\n:point_right: Head: \`${head_short}\` · ${run_count} pending run(s)\n<${approve_url}|Approve workflow runs>"
+
+  response=$(curl -s -o /dev/null -w "%{http_code}" -X POST \
+    -H 'Content-Type: application/json' \
+    --data "$(jq -n --arg text "$slack_text" '{text: $text}')" \
+    "$SLACK_CI_APPROVAL_WEBHOOK_URL")
```
MRmarioruci: `.github/actions/slack/action.yml` already wraps this — `pr-review-requested.yml` calls it. Worth invoking the composite action here too instead of inlining curl + jq, so Slack-formatting lives in one place.

.github/scripts/notify-ci-approval.sh
timothyaterton: Good catch — you're right that fork-PR runs in `action_required` arrive with `pull_requests: []`, so the filter was dropping exactly the runs we need.

Fixed in 2d8fe20: the script now fetches all open PRs up front and builds a `head_sha → PR` lookup map. For each run, it tries `pull_requests[0].number` first, then falls back to the sha map. This also avoids per-PR API calls for metadata since we already have title/html_url from the open-PR list.

.github/scripts/notify-ci-approval.sh
timothyaterton: Both fixed in 2d8fe20:

- Dropped the `2>/dev/null` — API errors now surface in the workflow log instead of silently producing an empty string (which would defeat the dedup and re-send every 15 min).
- Added `--paginate` to the comments lookup so the marker isn't missed on PRs with >100 comments.

.github/scripts/notify-ci-approval.sh
timothyaterton: I looked into this — the composite action is a workflow step, so it can't be invoked from inside a bash loop (we send N messages, one per qualifying PR). We'd need to restructure into a matrix job to use it, which adds complexity for what's a single `curl` call.

The composite action also doesn't return the HTTP status code, which we use to log failures without hard-failing the workflow.

Happy to revisit if the team would prefer the matrix approach, but for a fire-and-forget notifier I think keeping the curl inline is the simpler path.

.github/scripts/notify-ci-approval.sh
```diff
@@ -0,0 +1,129 @@
...
+#
+# Requires: gh (authenticated), curl, jq
+# Env:      SLACK_CI_APPROVAL_WEBHOOK_URL (optional — exits cleanly if unset)
+#           GITHUB_REPOSITORY (e.g. dfinity/internet-identity)
+set -euo pipefail
+
+REPO="${GITHUB_REPOSITORY:?GITHUB_REPOSITORY must be set}"
+MARKER_PREFIX="<!-- ci-approval-slack:"
+
+if [[ -z "${SLACK_CI_APPROVAL_WEBHOOK_URL:-}" ]]; then
```
MRmarioruci: pr-revew-requested.yml already has SLACK_PRIVATE_IDENTITY_WEBHOOK_URL. Isn't this going to point to the same channel? If yes we can reuse the same secret

.github/scripts/notify-ci-approval.sh
```diff
@@ -0,0 +1,129 @@
...
+  marker_body="${marker}
+:robot_face: Slack notification sent for CI approval on head \`${head_short}\` (${run_count} pending run(s)).
+
+_Automated by [ci-approval-slack-notifier](https://github.com/${REPO}/actions/workflows/ci-approval-slack-notifier.yml)._"
+
+  gh api "repos/${REPO}/issues/${pr_number}/comments" \
+    -f body="$marker_body" --silent || \
+    echo "PR #${pr_number}: failed to post dedup marker (non-fatal)."
+
+  notified=$((notified + 1))
```
MRmarioruci, line 126: this counter stays at 0, `while` runs in a subshell because of the pipeline, so the final log line always prints "Notifications sent this run: 0" even when we did send some. `done < <(echo "$grouped" | jq -c ".[]")` would keep it in the parent shell.

MRmarioruci (dismissed):

.github/scripts/notify-ci-approval.sh
timothyaterton: Good call — switched to `SLACK_PRIVATE_IDENTITY_WEBHOOK_URL` in 9ceb295. No new secret needed, the workflow reuses the same one as `pr-review-requested.yml`. Updated the PR description accordingly.

.github/scripts/notify-ci-approval.sh
timothyaterton: Nice catch — fixed in 9ceb295. Switched from the pipe to process substitution (`done < <(echo "$grouped" | jq -c '.[]')`) so the `while` loop runs in the parent shell and `notified` increments correctly.

MRmarioruci (dismissed):

MRmarioruci (approved):

PR #3973, opened 2026-06-02
feat(fe): configure frontend feature flags from canister deploy args; split EMAIL_RECOVERY into recovery + set-up flags

sea-snake:
This PR now contains the work from two stacked PRs: the original canister-deploy-arg feature-flag plumbing, and the `EMAIL_RECOVERY` flag split from #3974 (merged into this branch).

---

## Part 1 — Configure frontend feature flags from canister deploy args

Frontend feature flags can currently only be flipped per-browser (localStorage via the console, or a `?feature_flag_*` URL param) or by changing a compile-time default. There is no way for an operator to set a flag's baseline for a deployment (e.g. enable `EMAIL_RECOVERY` on a specific environment) without a code change and release. This adds that, reusing the existing init-arg channel that already carries config from the frontend canister to the browser (the same path `getPrimaryOrigin` uses).

### Changes
- Add `feature_flags : opt vec record { text; bool }` to `InternetIdentityFrontendArgs` and `internet_identity_frontend.did`, with a full `npm run generate` regen of the frontend bindings (the pinned didc also pulls in pre-existing field reordering and the `featured_dashboard_apps` doc comment that the committed bindings were missing).
- `globals.ts`: add `getConfiguredFeatureFlag(name)` reading the decoded init arg.
- `featureFlags.ts`: in each flag's `initialize()`, apply the configured value as the baseline via `temporaryOverride` (non-persistent) before running the flag's init callback. A persisted localStorage value short-circuits this, so user/console/URL choices still win. Unknown flag names in the deploy arg are ignored.

Resolution order (low → high): compile-time default → canister deploy arg → localStorage → init callback → URL param.

Example deploy arg:
```
feature_flags = opt vec {
  record { "EMAIL_RECOVERY"; true };
  record { "DISCOVERABLE_PASSKEY_FLOW"; false };
};
```

### Tests
- `featureFlags/index.test.ts`: `temporaryOverride` changes the value without persisting, and a later `set` persists over it.
- `state/featureFlags.test.ts` (new): resolution order — default held when unconfigured, canister arg overrides the default, localStorage takes precedence over the canister arg.
- Existing `check_candid_interface_compatibility` passes (`.did` matches the Rust type).

---

## Part 2 — Split `EMAIL_RECOVERY` into recovery and set-up flags (was #3974)

Splits the single `EMAIL_RECOVERY` feature flag into two independent flags so the recover-with-email flow and the recovery-email set-up surface can be gated separately:

- `EMAIL_RECOVERY` — recover an identity with a bound recovery email (the recover-with-email button on the `/recovery` sign-in page).
- `EMAIL_RECOVERY_SETUP` — set up, replace or remove a recovery email (the recovery email card in the manage area and the dashboard set-up smart-action).

### Resolution
Both flags resolve identically, via the shared `enableOnIdAiDomains` init callback:
- **on** for `id.ai` and `beta.id.ai`,
- **off** on every other domain,
- but an explicit canister-config value (Part 1) always wins — a configured `false` keeps the flag off on every domain, and a configured `true` turns it on regardless of domain.

The callback returns early when the operator configured the flag, so the domain default never overrides an explicit `false`. localStorage / `?feature_flag_*` overrides still take precedence as before. Note this widens the previous beta-only auto-enable (`beta.id.ai`) to also include `id.ai`.

### Changes
- `featureFlags.ts`: add the `EMAIL_RECOVERY_SETUP` store and the shared `enableOnIdAiDomains(name)` init callback; register both flags in the exported map.
- `manage/(authenticated)/recovery/+page.svelte` and `manage/(authenticated)/(home)/+page.svelte`: gate the set-up card, set-up wizard trigger, and dashboard smart-action with `EMAIL_RECOVERY_SETUP`; the `/recovery` flow stays on `EMAIL_RECOVERY`.

### Tests
- `state/featureFlags.test.ts`: domain defaults (`id.ai`/`beta.id.ai` on, other domains off), configured-`false`-wins-on-`id.ai`, configured-`true`-on-off-domain, and independent resolution of the two flags. The test captures and restores the original `window.location` property descriptor so the override doesn't leak across the Vitest worker.
- e2e: the suite runs against `https://id.ai`, where both flags default on, so the wizard/flow tests rely on the domain default (the fixture's `enableFlag` helper was removed); the off-path fixture writes `false` to both localStorage keys. The shared `ManageRecoveryPage.activate()` helper now targets the phrase button by its full aria-label (`Activate recovery phrase`) so it isn't ambiguous now that the recovery-email card renders on `id.ai`.

### DIFF
--- a/src/internet_identity_frontend/internet_identity_frontend.did
+++ b/src/internet_identity_frontend/internet_identity_frontend.did
@@ -58,6 +58,11 @@ type InternetIdentityFrontendInit = record {
     // Origins of apps to feature on the dashboard home. Each origin is resolved
     // against the bundled dapps catalogue to render name, description and logo.
     featured_dashboard_apps : opt vec text;
+    // Frontend feature flag overrides keyed by flag name, e.g.
+    // record { "EMAIL_RECOVERY"; true }. Sets the deployment-level baseline for
+    // each flag; localStorage, the flag's init callback and ?feature_flag_* URL
+    // params still take precedence. Unknown flag names are ignored.
+    feature_flags : opt vec record { text; bool };
 };
 
 service : (InternetIdentityFrontendInit) -> {
--- a/src/frontend/src/lib/generated/internet_identity_frontend_idl.js
+++ b/src/frontend/src/lib/generated/internet_identity_frontend_idl.js
@@ -17,6 +17,7 @@ export const idlFactory = ({ IDL }) => {
     'dev_csp' : IDL.Opt(IDL.Bool),
     'dummy_auth' : IDL.Opt(IDL.Opt(DummyAuthConfig)),
     'featured_dashboard_apps' : IDL.Opt(IDL.Vec(IDL.Text)),
+    'feature_flags' : IDL.Opt(IDL.Vec(IDL.Tuple(IDL.Text, IDL.Bool))),
   });
   const HeaderField = IDL.Tuple(IDL.Text, IDL.Text);
   const HttpRequest = IDL.Record({
@@ -55,6 +56,7 @@ export const init = ({ IDL }) => {
     'dev_csp' : IDL.Opt(IDL.Bool),
     'dummy_auth' : IDL.Opt(IDL.Opt(DummyAuthConfig)),
     'featured_dashboard_apps' : IDL.Opt(IDL.Vec(IDL.Text)),
+    'feature_flags' : IDL.Opt(IDL.Vec(IDL.Tuple(IDL.Text, IDL.Bool))),
   });
   return [InternetIdentityFrontendInit];
 };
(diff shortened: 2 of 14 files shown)


src/internet_identity_frontend/internet_identity_frontend.did
```diff
@@ -58,6 +58,11 @@ type InternetIdentityFrontendInit = record {
     // Origins of apps to feature on the dashboard home. Each origin is resolved
     // against the bundled dapps catalogue to render name, description and logo.
     featured_dashboard_apps : opt vec text;
+    // Frontend feature flag overrides keyed by flag name, e.g.
+    // record { "EMAIL_RECOVERY"; true }. Sets the deployment-level baseline for
+    // each flag; localStorage, the flag's init callback and ?feature_flag_* URL
+    // params still take precedence. Unknown flag names are ignored.
```
aterga, line 64: "params still take precedence" - which params?

aterga (approved):

src/internet_identity_frontend/internet_identity_frontend.did
sea-snake: "URL params", the line break is probably confusing here.

PR #3966, opened 2026-06-02
feat(fe): auth disambiguation dialog primitives

MRmarioruci:
Foundation for the upcoming homepage and `/authorize` / `/manage` alignment work. Introduces reusable dialog primitives for auth disambiguation, extracts `ContinueOnAnotherDeviceDialog` into its own component, and improves OIDC-cancel UX with a toast. No existing flows are wired to the new dialogs yet — that happens in the stacked PRs.

## Changes

- Add `IdentityNotConnectedDialog`, `IdentityAlreadyLinkedDialog`, `SwitchAccessMethodDialog` components
- Extract `ContinueOnAnotherDeviceDialog` from `AuthWizard` into a standalone component
- Show a toast when OIDC sign-in is canceled by the user

<div align="right">Next: #3967</div>

PR #3975, opened 2026-06-02
fix(fe): de-flake CLI e2e Settings navigation on mobile

aterga:
The `e2e-playwright (mobile, …)` shards intermittently fail on `cli.spec.ts` — most visibly ``` `--app` links the same principal that /authorize gives for that app ```, which times out for 60s clicking the "Settings" link.

The three CLI tests that enable device CLI access open the mobile sidebar nav with `if (await menuButton.isVisible())` before clicking "Settings". `isVisible()` returns immediately without waiting, so on the mobile project (Pixel 5) it races page hydration: when the hamburger button hasn't rendered yet, the guard is `false`, the menu stays closed, and the following "Settings" click never becomes actionable. Desktop is unaffected because its nav is always expanded. The pattern was introduced with the CLI flow in #3956.

# Changes

- Replace the non-waiting `if (await menuButton.isVisible())` guard with a deterministic branch on Playwright's `isMobile` fixture (`true` for the Pixel 5 project, `false` for Desktop Chrome). The menu-button `.click()` auto-waits for actionability, eliminating the race.
- Factor the 12-line "enable CLI access via Settings" block — duplicated verbatim across all three tests — into a shared `enableCliAccessInSettings(page, isMobile)` helper.

The same `isVisible()` menu guard also exists in `manage/index.spec.ts` and `authorize/index.spec.ts` / `index.spec.ts`; this PR fixes the CLI tests that were actively failing, and the others (some operating on secondary device pages) can follow up separately.

# Tests

- `tsc` (after `svelte-kit sync`), `eslint`, and `prettier` pass on the change.
- The change is test-only (no app code touched); the de-flake itself is validated by the `e2e-playwright (mobile)` shard in CI.

### DIFF
--- a/src/frontend/tests/e2e-playwright/routes/cli.spec.ts
+++ b/src/frontend/tests/e2e-playwright/routes/cli.spec.ts
@@ -62,7 +62,7 @@ const enableCliAccessInSettings = async (
   }
   await page.getByRole("link", { name: "Settings" }).click();
   await page.waitForURL(II_URL + "/manage/settings");
-  await page.getByRole("switch").click();
+  await page.getByRole("switch", { name: "CLI access" }).check();
   await page
     .getByLabel("I'm using the official ICP CLI and I trust this device.")
     .check();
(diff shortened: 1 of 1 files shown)


sea-snake (dismissed):

MRmarioruci:
I've fixed the same inside this https://github.com/dfinity/internet-identity/pull/3966 but I am happy if this goes in and I rebase as well.

sea-snake (dismissed):

sea-snake (approved):

PR #3986, opened 2026-06-04
fix(be,fe): support RSA/SHA-512 DNSSEC so email recovery works for mailbox.org

sea-snake:
Email recovery silently stalled for mailbox.org addresses (and any zone signed with RSA/SHA-512): the wizard fired its DNS lookups, then went quiet — no error, no success. mailbox.org is DNSSEC-signed, so the frontend assembled a DNSSEC proof bundle and committed to the DNSSEC path, but it bundled whichever RRSIG the resolver returned first — for mailbox.org a deprecated RSA/SHA-1 signature. The canister verifier only implemented algorithms 8/13/15, so the bundle failed verification; and because supplying any bundle commits prepare to DNSSEC (no DoH fallback once a bundle is present), the flow never recovered or surfaced a clean error.

## Changes
- **Backend:** add RSA/SHA-512 (DNSSEC algorithm 10) to the verifier — a "Validation: MUST" algorithm in RFC 8624 §3.1 and the one mailbox.org actually signs with. The PKCS#1 v1.5 body is shared with RSA/SHA-256; only the digest differs. The SHA-1 family (5, 7) stays unsupported by design (collision-broken; rfc8624-bis moves it to MUST NOT). Corrected doc comments that mislabeled `{8,13,15}` as "the RFC 8624 MUST set".
- **Frontend:** when a zone double-signs an RRset, select an RRSIG whose algorithm the canister can verify (`{8,10,13,15}`) instead of taking the first. If none is supported the walk yields nothing and the wizard falls through to the DoH path; an unsupported domain then lands in the existing "Can't use this email" view, whose explainer now names the supported DNSSEC algorithms.

## Tests
- Real-capture mailbox.org chain vector (RSA-SHA256 root → org → RSA-SHA512 leaf zone + DKIM TXT) verifies end-to-end against the canister verifier.
- Unit tests for the alg-10 dispatch (backend) and the supported-RRSIG selection (frontend).


src/frontend/src/lib/utils/dnssec/chain.ts
```diff
@@ -406,6 +413,38 @@ async function fetchSignedRRsetEither(
...
+ * covers `matchedType` and uses an algorithm the canister can verify
+ * ({@link SUPPORTED_DNSSEC_ALGORITHMS}).
+ *
+ * Returns `undefined` when every covering RRSIG uses an unsupported
+ * algorithm (a SHA-1-only zone, say). The walk then aborts and
+ * `assembleSkeleton` yields nothing, so the wizard falls through to
+ * the canister's DoH path — the same outcome as an unsigned zone.
+ * Exported for unit testing the selection in isolation.
+ */
+export function selectSupportedCoveringRrsig(
```
aterga, line 427: This function should be unit tested.

aterga (changes requested):
LGTM but there should be unit tests for a non-trivial function with bit operations.

aterga (approved):

PR #3991, opened 2026-06-08
refactor(fe): make AuthWizard a black box

MRmarioruci:
Restructures `AuthWizard` so OIDC disambiguation, mode-toggling, and dialog presentation are all internal to the wizard. The parent surface shrinks to four lifecycle callbacks plus `initialMode`; the leak-prone callback-with-functions pattern is removed entirely.

Addresses review comments on #3967.

## Changes

- `AuthFlow` gains view states for disambiguation (`openIdNotConnected`, `openIdAlreadyLinked`, `confirmMethodSwitch`) and parked state; new transition methods (`setMode`, `requestMethodSwitch`, `confirmOpenIdSignUp`, `confirmOpenIdSignIn`, `cancelOpenIdDisambiguation`, `confirmMethodSwitch`, `cancelMethodSwitch`); minimal display getters (`mode`, `configIssuer`, `userName`, `userEmail`, `pendingMethodSwitch`, `providerName`, `providerLogo`).
- `AuthWizard` owns `mode` + `isElevated` and self-elevates to a dialog on toggle when mounted inline. Three callback props (`onOpenIdNotConnected`, `onOpenIdAlreadyLinked`, `onMethodSwitch`), the `onSwitchMode` callback, and the `withinDialog` flag are all removed. `Props` shrinks from 11 to 6.
- `Dialog` reads Svelte context — a nested `<Dialog>` becomes a passthrough automatically, sidestepping `<dialog>.showModal()` stacking issues.
- `IdentityNotConnected`, `IdentityAlreadyLinked` take `issuer` + optional `providerName` / `providerLogo` and look up provider info from `backendCanisterConfig` themselves. `SwitchAccessMethod` takes `previousIdentityNumber` + `newMethod` tag and derives `fromMethod` / `toMethod` / user info from `lastUsedIdentitiesStore`.
- The `AuthWizard.gating.ts` predicate module is deleted; predicates moved inline into `AuthFlow.continueWithOpenId` / `continueWithSso`.
- Page call sites (`(new-styling)/+page.svelte`, `authorize/+layout.svelte`, `authorize/+page.svelte`, `cli/+layout.svelte`, `manage/(authenticated)/+layout.svelte`) lose payload state, dialog render blocks, breadcrumb flags, and callback wirings; the homepage's separate `isCreateIdentityDialogOpen` modal collapses into the single auth modal.

## Tests

- New `authFlow.svelte.test.ts` with 13 unit tests covering mode, method-switch transitions, OpenID disambiguation cleanup, and display getter defaults.


src/frontend/src/lib/components/ui/Dialog.svelte
```diff
@@ -1,11 +1,20 @@
+<script lang="ts" module>
+  import { getContext } from "svelte";
+
+  export const isInsideDialog = (): boolean => getContext("inDialog") === true;
+</script>
+
 <script lang="ts">
-  import { onMount } from "svelte";
+  import { onMount, setContext } from "svelte";
```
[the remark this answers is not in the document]
MRmarioruci: `getContext` is imported in the module `<script>` at the top of the file, and Svelte makes those module-level bindings available to the instance script, so it's in scope where `nested` is set. Compiles fine and CI's green.

src/frontend/src/lib/flows/authFlow.svelte.test.ts
```diff
@@ -0,0 +1,443 @@
...
+
+const requestJWTMock = vi.hoisted(() => vi.fn());
+const decodeJWTMock = vi.hoisted(() => vi.fn());
+
+vi.mock("$lib/utils/openID", () => ({
+  requestJWT: requestJWTMock,
+  decodeJWT: decodeJWTMock,
+  selectAuthScopes: (scopes: string[]) => scopes,
+  extractIssuerTemplateClaims: () => [],
+}));
```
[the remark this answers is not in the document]
MRmarioruci: A `vi.mock` factory replaces the whole module, so a missing export resolves to `undefined` rather than a load error. `findConfig` isn't exercised by these tests and the suite passes. Can add it to the mock as a safety net if we start covering the disambiguation getters.

src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
```diff
@@ -103,10 +118,11 @@
               {#if authenticatingProviderId === provider.client_id}
                 <ProgressRing />
               {:else if provider.logo}
-                <div class="size-6">
+                <div class="size-5">
                   <!-- eslint-disable-next-line svelte/no-at-html-tags -- provider.logo is a trusted SVG string sourced from the backend canister's openid_configs -->
                   {@html provider.logo}
                 </div>
+                <span>{name}</span>
               {/if}
```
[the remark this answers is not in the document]
MRmarioruci: Good catch, the name should render even when there's no logo. Going to fix it in a separate PR to keep this one focused on the refactor.

src/frontend/src/lib/components/ui/Dialog.svelte
```diff
@@ -132,7 +152,9 @@
   On both layouts, spacer elements push the content above the software
   keyboard and safe-area insets.
 -->
-{#if isOpen}
+{#if passthrough}
+  {@render children?.()}
+{:else if isOpen}
```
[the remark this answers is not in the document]
MRmarioruci: Good catch. `onMount` was missing the `nested` guard that `onNavigate` already has, so a nested dialog's cleanup could flip the ancestor's keyboard overlay back off. Fixed in 8aefbfb8.

src/frontend/src/lib/components/wizards/auth/views/SwitchAccessMethod.svelte
```diff
@@ -3,21 +3,98 @@
...
+    ) {
+      return getMetadataString(
+        previousIdentity.authMethod.openid.metadata,
+        "email",
+      );
+    }
+    if ("sso" in previousIdentity.authMethod) {
+      return previousIdentity.authMethod.sso.email;
+    }
+    return undefined;
```
aterga, line 50: Can we enforce full case analysis (e.g., using `satisfied never`)

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
```diff
@@ -161,6 +161,33 @@
...
+    {/if}
+    <button
+      onclick={onClose}
+      class="btn btn-tertiary btn-sm btn-icon absolute inset-e-2 top-2 rounded-full"
+    >
+      <XIcon class="size-5" />
+      <span>{$t`Close`}</span>
+    </button>
+  </div>
+{/snippet}
```
sea-snake, line 187: Why is this needed? A user shouldn't be able to remove an identity they're currently signed in with (UX is already blocking this).

src/frontend/src/lib/components/wizards/auth/views/SwitchAccessMethod.svelte
aterga: Ditto elsewhere

(Can we ask AI to suggest all places in this PR where exclusive case analysis could be enforced?)

src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
```diff
@@ -118,29 +136,15 @@
         the backend canary allowlist are rejected there.
       -->
       <button
-        class="btn btn-secondary btn-xl flex-1"
+        class="border-border-secondary text-fg-primary bg-bg-primary hover:not-disabled:bg-bg-primary_hover disabled:border-border-disabled disabled:text-fg-disabled focus-visible:ring-focus-ring focus-visible:ring-offset-bg-primary flex h-16 w-full flex-col items-center justify-center gap-1.5 rounded-md border text-xs font-semibold outline-none not-disabled:cursor-pointer focus-visible:ring-2 focus-visible:ring-offset-2"
         onclick={signInWithSso}
```
sea-snake: Why is this no longer using the `btn btn-secondary` class and instead specifying all the styles itself?

src/frontend/src/lib/components/wizards/auth/views/IdentityAlreadyLinked.svelte
```diff
@@ -2,24 +2,41 @@
   import { t } from "$lib/stores/locale.store";
   import { InfoIcon, UserIcon } from "@lucide/svelte";
   import ProgressRing from "$lib/components/ui/ProgressRing.svelte";
+  import { findConfig } from "$lib/utils/openID";
 
   interface Props {
-    providerName: string;
+    issuer?: string;
+    providerName?: string;
     providerLogo?: string;
```
sea-snake: Why are these props still here? I'd expect only issuer here.

src/frontend/src/lib/components/wizards/auth/views/IdentityNotConnected.svelte
```diff
@@ -7,26 +7,43 @@
...
     onRecover: () => void;
+    onCancel?: () => void;
     loading?: boolean;
   }
 
   let {
-    providerName,
-    providerLogo,
+    issuer,
+    providerName: providerNameProp,
```
sea-snake, line 25: Same here.

aterga (approved):
✅ LGTM — clean encapsulation refactor that eliminates the callback-with-functions pattern flagged on #3967. View state moves into `AuthFlow` as a proper FSM with solid unit coverage; page call sites shrink by ~600 lines net; CI fully green.

## What the refactor actually does

Previously every page (`+page.svelte`, `authorize/+layout.svelte`, `manage/(authenticated)/+layout.svelte`) held its own `notConnectedPayload`/`alreadyLinkedPayload`/`methodSwitchPayload` `$state` plus `<Dialog>` blocks, and `AuthWizard` handed them function-typed `resume()`/`signIn()`/`proceed()`/`cancel()` callbacks via three `on…` props. Now `AuthFlow` ([`src/frontend/src/lib/flows/authFlow.svelte.ts`](https://github.com/dfinity/internet-identity/blob/main/src/frontend/src/lib/flows/authFlow.svelte.ts)) owns the disambiguation states (`openIdNotConnected`, `openIdAlreadyLinked`, `confirmMethodSwitch`) and `AuthWizard` renders the views itself, swapping content reactively inside a single persistent `<Dialog>`. AuthWizard's Props shrink from 11 to 6. True encapsulation, not a rename.

## #3967 follow-up adherence

- **"Callbacks should return data, not functions"** — verified. Zero `resume:|signIn:|proceed:|cancel:` function-property hits in added types; the `OpenIdNotConnectedArgs`/`OpenIdAlreadyLinkedArgs`/`MethodSwitchArgs` group is deleted. New transitions return `bigint` or `"needs-name"`, never closures.
- **"Data referenced by id from localStorage"** — verified. `IdentityNotConnected`/`IdentityAlreadyLinked` now take `issuer` and look up provider info via `findConfig`; `SwitchAccessMethod` takes `previousIdentity` (snapshot) + `newMethod` tag and derives provider details internally.

## Follow-up suggestions (all deferrable, not blocking)

1. **Missing happy-path unit test for `confirmOpenIdSignIn`** — `authFlow.svelte.test.ts` covers the throw-when-no-pending branch but no test asserts that calling `confirmOpenIdSignIn` after a `"signIn"` result from `continueWithOpenId` returns the parked `identityNumber` and clears `#jwt`/`#configIssuer`/`#pendingOpenIdSignIn`. The e2e specs cover it indirectly.

2. **`AuthWizard.svelte` `onDestroy` cancel-on-unmount heuristic** (~lines 481–485) is a behaviour band-aid for parent-Dialog X clicks. The comment explains it well and the `methodSwitchConfirmed` guard makes it correct today, but an explicit parent close lifecycle would be cleaner than relying on unmount ordering.

3. **`Dialog.svelte`'s `passthrough` prop name is opaque** — it means "render children inline because we're already nested." `nested` or `inline` would self-document better. Pure naming nit.

4. **`findConfig(this.#configIssuer, undefined, [])` is called inline in two AuthFlow getters with the same empty-metadata array** — minor opportunity to extract if an issuer-template-with-claims case for the parked state ever appears.

5. **Deleted `add-identity.spec.ts`** covered the manage-layout sign-in↔sign-up modal toggle path. That UX still exists (now driven by `bind:mode`) but the e2e signal for it is gone. A small spec exercising "open Add identity → toggle sign-up → back to sign-in" would restore coverage.

aterga (approved):
Verdict: ✅ LGTM

What the refactor actually does: True encapsulation, not rename. The OIDC disambiguation flow (NotConnected / AlreadyLinked / SwitchAccessMethod dialogs) was previously a parent-side concern — every page held its own $state payloads + <Dialog> blocks plus received function-typed resume()/signIn()/proceed() callbacks. Now AuthFlow (src/frontend/src/lib/flows/authFlow.svelte.ts) owns the disambiguation states and AuthWizard renders the views itself inside a single persistent <Dialog>. AuthWizard props shrink from 11 to 6; three page files each lose ~200 lines of identical boilerplate.

Addresses #3967 concerns:

* sea-snake's "callbacks should return data, not functions" — ✅ verified, zero resume:|signIn:|proceed:|cancel: function-property hits in added types; the OpenIdNotConnectedArgs/OpenIdAlreadyLinkedArgs/MethodSwitchArgs group is deleted.
* "Data referenced by id from localStorage" — ✅ IdentityNotConnected/IdentityAlreadyLinked now take issuer and look up via findConfig; SwitchAccessMethod takes previousIdentity + newMethod tag.
BLOCKING: None.

FOLLOW-UP (5, all deferrable):

1. authFlow.svelte.test.ts doesn't have an explicit happy-path unit test for confirmOpenIdSignIn returning the parked identityNumber and clearing #jwt/#configIssuer/#pendingOpenIdSignIn (e2e covers it indirectly).
2. The onDestroy cancel-on-unmount heuristic in AuthWizard.svelte (~lines 481-485) is a band-aid for parent-Dialog X clicks. Correct today via methodSwitchConfirmed guard, but an explicit parent close lifecycle would be cleaner.
3. Dialog.svelte's passthrough prop name is opaque — means "render inline because already nested." nested or inline would self-document better. Pure naming nit.
4. findConfig(this.#configIssuer, undefined, []) is called inline in two AuthFlow getters with the same empty-metadata array — minor opportunity to extract if metadata ever matters.
5. Deleted add-identity.spec.ts covered the manage-layout sign-in↔sign-up toggle path. UX still exists (bind:mode), but the e2e signal is gone — a small spec restoring that coverage would close the gap.
CI fully green. Tests cover the new FSM thoroughly.

One-line summary suitable for the PR:

✅ LGTM — clean encapsulation refactor that eliminates the callback-with-functions pattern sea-snake flagged on [#3967](https://github.com/timothyaterton/temate/issues/3967), all view state moves into AuthFlow as a proper FSM with solid unit coverage, page call sites shrink by ~600 lines net, CI fully green; minor follow-ups (one missing happy-path unit test, passthrough naming, an e2e gap from the deleted add-identity spec) can be deferred.

Please also wait for @sea-snake's review

src/frontend/src/lib/components/wizards/auth/views/SwitchAccessMethod.svelte
MRmarioruci: Done — added `satisfies never` to `fromMethod`/`toMethod` here, and to `methodType` in AuthWizard. Going through the rest of the diff, those three are the only genuine closed-union case analyses. The `error.type` switch handles a subset on purpose and rethrows the rest, the `result.type` checks are binary and already total, and the `activeView` `{#if}` chain can't take a clean `satisfies never` in-template — so I left those.

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
MRmarioruci: It's actually reachable: the /manage popover passes `selected={$authenticatedStore.identityNumber}` against `identities={lastUsedIdentities}`, and those diverge after recovery / access-method switch (neither writes last-used). Without the fallback the card renders empty and Sign out becomes unreachable. Kept it, but dropped the contradictory `!` on the derived so the `{#if}` is honest.

src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
MRmarioruci: Good call — back on `btn btn-secondary` now. Added a small `btn-tile` size utility for the vertical logo+label shape (the only part `btn-secondary` didn't already cover) and used it across these and the PickAuthenticationMethod tiles.

src/frontend/src/lib/components/wizards/auth/views/IdentityAlreadyLinked.svelte
MRmarioruci: Dropped `providerLogo` — it was redundant (the view re-derives the logo from `issuer` via `findConfig`). Kept `providerName` because this view is also shown for the SSO disambiguation flow, where there's no configured issuer to look up and the name comes from the discovered domain. `userName`/`userEmail` are the signed-in user's JWT claims, not derivable from the provider issuer.

src/frontend/src/lib/components/wizards/auth/views/IdentityNotConnected.svelte
MRmarioruci: Same as IdentityAlreadyLinked: dropped `providerLogo` (redundant), kept `providerName` (SSO flow has no configured issuer) and `userName`/`userEmail` (the user's JWT claims).

src/frontend/src/lib/components/ui/IdentitySwitcher.svelte
sea-snake: That's strange, I'd expect that switching access method would result in the lastUsedIdentities being updated. Though not sure what happens when last used identities is empty, and a user recovers.

Maybe the identity switcher should use the authentication store state instead for the current identity instead, that would probably make more sense. But it's a larger change, I'd consider that out of scope for this PR.

src/frontend/src/lib/components/wizards/addAccessMethod/views/AddAccessMethod.svelte
sea-snake: I never asked for btn-tile... I expected it to just extend/override the btn styling where needed with inline classes.

Let's avoid pushing one-off design deviations into the design system.

MRmarioruci: Fair point — reverted the utility and inlined the shape delta on each of the 4 sites in [`8620b39e`](https://github.com/dfinity/internet-identity/pull/3991/commits/8620b39e). `btn btn-secondary` stays, only `h-16 w-full flex-col gap-1.5 text-xs whitespace-normal` is local now.

sea-snake (approved):

PR #3992, opened 2026-06-08
refactor(be,fe): explicit DoH fallback for email recovery; de-duplicate wizards

sea-snake:
Follow-up cleanup to #3989, which fixed email recovery hanging for Outlook-style addresses but left two rough edges this PR addresses. No user-facing behaviour change.

The first: the backend signalled "resolve this DKIM leaf over DoH instead of DNSSEC" by having the frontend submit an *empty* `hops` vector to `email_recovery_submit_dkim_leaf` — an overloaded signal on a field that otherwise carries a real resolution chain. The second: the two email-recovery wizards carried ~250 lines of near-identical polling logic copy-pasted between them, plus two copies of the error-mapping helpers that differed only in dead branches and two genuinely-distinct address-error strings.

## Changes

Backend:
- Add a dedicated `email_recovery_submit_dkim_leaf_via_doh(nonce)` method for the DoH fallback. It carries no leaf data; the canister resolves the DKIM key over its own allowlist-gated DoH path.
- `email_recovery_submit_dkim_leaf` requires >=1 hop again — the empty-vec overload is gone. Both methods share a `finalize` tail.

Frontend:
- Extract the shared polling state machine into `wizards/emailRecovery/shared/poll.ts`, and move the shared views + diagnostics into `wizards/emailRecovery/shared/` so neither wizard reaches into the other's folder.
- Collapse the duplicated error mapping into one exhaustive `friendlyFailedReason` whose parameter type excludes the domain-routed variants, so the previously-dead `DomainNotAllowlisted`/`DomainNotSupported` branches can no longer be written, and a trailing `satisfies never` turns a new `EmailRecoveryError` variant into a compile error until it's given copy.
- Wire `submitDkimLeafViaDoh` through both wizards; the empty-`hops` literal is replaced by an explicit call.

## Tests
- Migrated the email-recovery integration tests to the new method; full suite passes (22/22) against PocketIC 9.0.3.
- `cargo clippy` (`internet_identity` + `canister_tests`, `--tests -D warnings`) clean; candid interface-compatibility test passes; `tsc` + `svelte-check` report 0 errors.

---

Rebased on #3990 ("instrument DoH failure cause for the recovery funnel"), which landed after this branch was cut and touched the same wizard. Its `doh_reason` funnel instrumentation is preserved in the shared poll module via a shared `dohSubReason` helper — so it now applies to both the setup and recovery `*-failed` events (#3990 instrumented setup only). All of #3990's other changes (backend `map_doh_error` tokens + test, funnel doc) are unaffected.

---

**Error handling now acts on the contract, not backend internals.** A canister `Err` caught from the leaf submit is the terminal verdict, so it's routed straight to the failed/unsupported view from the typed value (a new `CanisterError.raw` accessor) instead of being discarded in the hope a later `status` poll re-derives the same reason from canister state. And the polling `status` call is now guarded — a transient query failure is retried (bounded) rather than throwing out of the loop and stranding the user on the waiting screen with no Expired backstop. (Addresses review feedback that the FE shouldn't depend on backend implementation, and closes a latent silent-stall from #3989.)


src/internet_identity/src/email_recovery/submit_leaf.rs
```diff
@@ -62,45 +81,64 @@ pub async fn submit_dkim_leaf(
...
-    // material; this is allowlist-gated, so a domain the operator hasn't
-    // enabled comes back as `DomainNotAllowlisted` rather than hanging the
-    // pending entry until it expires.
-    let verification = if hops.is_empty() {
-        run_doh_fallback(&snapshot).await
-    } else {
-        run_submit(&hops, &extra_chains, &snapshot, now_secs)
-    };
+    let verification = run_submit(&hops, &extra_chains, &snapshot, now_secs);
+    finalize(&nonce, &snapshot, verification, now_secs).await
```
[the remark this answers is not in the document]
sea-snake: Intentional — removing the empty-`hops` overload is the goal of this PR. II ships the frontend and canister in lockstep (the FE is served by the canister), and the empty-`hops` signal was an internal FE↔BE protocol detail introduced in #3989, not a documented public contract. The only caller is II's own frontend, updated in this PR to call `email_recovery_submit_dkim_leaf_via_doh`, so there are no external clients pinned to the old behaviour.

src/internet_identity/src/email_recovery/submit_leaf.rs
```diff
@@ -233,6 +271,15 @@ fn run_submit(
...
     now_secs: u64,
 ) -> Result<(), EmailRecoveryError> {
+    // A genuine DNSSEC submission always carries at least the final
+    // TXT hop. An empty `hops` set is malformed input on this path —
+    // the FE that can't walk DNSSEC must call
+    // `submit_dkim_leaf_via_doh` instead. Reject it explicitly rather
+    // than relying on the hop walk below to fail.
+    if hops.is_empty() {
+        return Err(EmailRecoveryError::DkimLeafMismatch);
+    }
```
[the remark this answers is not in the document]
sea-snake: This is a defensive branch that the real client can't reach: after this PR the frontend never submits empty `hops` to this method (it calls `email_recovery_submit_dkim_leaf_via_doh`), and `run_submit` → `verify_hops_with_clock` would itself return `DkimLeafMismatch` for an empty hop set. The explicit guard is just a clearer short-circuit, so keeping the same variant avoids a candid change (new error variant + FE exhaustiveness + regen) for an unreachable path. Worth revisiting if this method is ever exposed beyond II's own frontend.

src/internet_identity/internet_identity.did
```diff
@@ -1530,6 +1523,12 @@ service : (opt InternetIdentityInit) -> {
     email_recovery_status : (text) -> (EmailRecoveryStatus) query;
     email_recovery_diagnostics : (text) -> (opt EmailRecoveryDiagnostics) query;
     email_recovery_submit_dkim_leaf : (EmailRecoverySubmitDkimLeafArg) -> (variant { Ok : EmailRecoveryStatus; Err : EmailRecoveryError });
+    // DoH-fallback sibling of email_recovery_submit_dkim_leaf, called
+    // with just the nonce when the FE can't walk a fully-signed DNSSEC
+    // resolution for the leaf (the DKIM record CNAMEs into an unsigned
+    // zone). The canister resolves the DKIM key over its own
+    // allowlist-gated DoH path.
+    email_recovery_submit_dkim_leaf_via_doh : (text) -> (variant { Ok : EmailRecoveryStatus; Err : EmailRecoveryError });
```
aterga:
```suggestion
    email_recovery_submit_dkim_leaf_via_doh : (record { nonce : opt text }) -> (variant { Ok : EmailRecoveryStatus; Err : EmailRecoveryError });
```

src/frontend/src/lib/components/wizards/emailRecovery/shared/errors.ts
```diff
@@ -0,0 +1,114 @@
...
+ * `*-failed` funnel event so the funnel is segmentable.
+ */
+export const dohSubReason = (
+  status: EmailRecoveryStatus,
+): string | undefined => {
+  if (!("Failed" in status)) return undefined;
+  const reason = status.Failed;
+  if (!("DohFetchFailed" in reason)) return undefined;
+  const token = reason.DohFetchFailed.split(":")[0]?.trim();
+  return token !== undefined && /^[a-z0-9_]+$/.test(token) ? token : undefined;
```
aterga: Why on Earth do we need a regex to parse errors from our own canister?

src/internet_identity/src/email_recovery/submit_leaf.rs
aterga, line 87: Finalize wasn't called at the end of this function before, why did we add it here now?

Without that, `submit_dkim_leaf` could be made synchronous.

src/internet_identity/src/email_recovery/submit_leaf.rs
```diff
@@ -62,45 +81,64 @@ pub async fn submit_dkim_leaf(
...
+/// same post-call polling status as `submit_dkim_leaf`.
+pub async fn submit_dkim_leaf_via_doh(
+    nonce: String,
+    now_secs: u64,
+) -> Result<EmailRecoveryStatus, EmailRecoveryError> {
+    let snapshot = pending::with_mut(&nonce, now_secs, |c| Snapshot::take(c))
+        .ok_or(EmailRecoveryError::NonceUnknown)??;
 
+    let verification = run_doh_fallback(&snapshot).await;
+    finalize(&nonce, &snapshot, verification, now_secs).await
```
aterga, line 109: What would happen if `snapshot` expires between lines 103 and 107? 

(This could happen since control is yielded before and after each `.await` statement)

src/internet_identity/src/email_recovery/submit_leaf.rs
aterga: NCR since it wasn't introduced in this PR, but for the future, this should have been 


```suggestion
    extra_chains: &[DelegationChain],
```

ditto elsewhere.

src/internet_identity/src/email_recovery/submit_leaf.rs
```diff
@@ -233,6 +271,15 @@ fn run_submit(
     snapshot: &Snapshot,
     now_secs: u64,
 ) -> Result<(), EmailRecoveryError> {
+    // A genuine DNSSEC submission always carries at least the final
+    // TXT hop. An empty `hops` set is malformed input on this path —
+    // the FE that can't walk DNSSEC must call
+    // `submit_dkim_leaf_via_doh` instead. Reject it explicitly rather
+    // than relying on the hop walk below to fail.
+    if hops.is_empty() {
+        return Err(EmailRecoveryError::DkimLeafMismatch);
```
aterga: The error name does not trivially map to "zero hops" - is this the right error case?

aterga (changes requested):

src/internet_identity/internet_identity.did
sea-snake: Wrapped in a record as suggested, but with `nonce : text` (required) rather than `opt text`: the nonce is the lookup key, so `None` has no valid meaning and would only add a dead error path. Matches the sibling `EmailRecoverySubmitDkimLeafArg`, whose `nonce` is also a required `text`.

src/frontend/src/lib/components/wizards/emailRecovery/shared/errors.ts
sea-snake: Agreed — removed. The regex came from #3990 (it lived in the setup wizard); this PR had only relocated it to the shared module. `DohFetchFailed` now carries a typed `DohFailureReason` variant (AllProvidersFailed / DedupWaitTimeout / QuorumFailed / ResponseMalformed), and the FE reads it via an exhaustive `satisfies never` switch — no string parsing, and a new backend cause becomes an FE compile error until it's handled.

src/internet_identity/src/email_recovery/submit_leaf.rs
sea-snake: It can't be made synchronous: `finalize` is async because the recovery path awaits `stamp_recovery_delegation` (which awaits `ensure_salt_set`). That await isn't new — on main the same call was inline in `submit_dkim_leaf`; extracting `finalize` just moved it. Only the setup (Register) path is synchronous, but the method serves both kinds through one entry point, so it must stay async.

src/internet_identity/src/email_recovery/submit_leaf.rs
sea-snake: Good catch — fixed. `finalize` now re-grabs a fresh `ic_cdk::api::time()` after the verification await and re-validates the entry is still present, unexpired, and in `NeedDkimLeaf` before binding/stamping; otherwise it returns `NonceExpired`. So an entry that expired (or was evicted by a concurrent poll) during the DoH outcall is no longer finalized against the stale pre-await timestamp. Residual: a small TOCTOU window remains across `stamp_recovery_delegation`'s own internal await (normally a no-op after the first delegation); fully closing it would need an in-progress status flip.

src/internet_identity/src/email_recovery/submit_leaf.rs
sea-snake: Agreed — there was no existing variant that cleanly means "zero hops". Added a dedicated `EmptyDkimLeafHops` so the malformed-request case is unambiguous, distinct from `DkimLeafMismatch` (which now only means a non-empty chain failed to validate).

src/internet_identity/src/email_recovery/submit_leaf.rs
sea-snake: Done — imported `SignedRRset` / `DelegationChain` / `AnchorNumber` so the signatures read `&[DelegationChain]` etc. instead of the fully-qualified paths.

src/internet_identity/internet_identity.did
aterga: Okay - I suppose we don't care about API evolution in this case.

src/internet_identity/src/email_recovery/submit_leaf.rs
aterga: Oh I see - we factored out some async code into `finalize`. 

The claim `on main the same call was inline in submit_dkim_leaf` isn't exactly right, though.

aterga (approved):

PR #3995, opened 2026-06-08
feat(be,fe): generic stale-while-revalidate single-flight cache

sea-snake:
PR #3987 introduced a cold/pending/warm cache with single-flight dedup inside the DoH layer. That pattern is useful beyond DoH (the OpenID/SSO work needs the same fetch-cache shape), so this extracts it into a reusable `single_flight_cache` module for IC canister code and upgrades it to stale-while-revalidate. Design: SSO readiness doc §5.2.2.

## Changes
- New `single_flight_cache` module. **Poll model, not callbacks**: `get(cache, key)` returns `Cached::Ready(v)` or `Cached::Pending` immediately and never blocks on — or wakes — another call (the IC can't suspend one call on another's task without mis-routing its reply). The detached fill's only job is fetch-and-store; whoever needs the value re-issues `get` (the recovery status poll already does this), reading it once `Ready`.
  - **Single-flight dedup** is the in-flight marker alone: concurrent `get`s for a key see the in-flight fill and return `Pending` (no second fan-out). A per-attempt `fill_id` + `abandon_fill_after` recovers a fill whose task trapped after its outcall.
  - **Stale-while-revalidate / stale-if-error**, **exponential-backoff** on failure (reset on success, bounded by the entry lifetime), **LRU eviction** over `max_entries`.
  - **Required `CacheConfig`** (no `Default`, non-optional `backoff`): the cache fronts the subnet's shared, bounded HTTPS-outcall capacity, so every abuse/availability knob is stated explicitly. The waiter-queue knobs (`max_waiters`) and the `CacheFillError` surface are gone — there are no waiters to bound, and the cache never surfaces an error (a transient fill failure reads as `Pending` until backoff clears or the value lands).
  - **Observability**: a read-only `CacheStats` snapshot (live entries by category — fresh / valued / backing-off — plus in-flight count, the cap, and a running inconsistency total) exposed via `stats()`, and wired into the canister `/metrics` endpoint as `internet_identity_doh_cache_*` gauges. Every "can't happen on a correct cache" branch print-logs an `ERROR` and bumps `internet_identity_doh_cache_inconsistencies` (always 0 in healthy operation). The LRU access clock renumbers on `u64` overflow rather than wrapping.
- DoH (`doh/cache.rs`, `doh/mod.rs`): `fetch_txt -> Result<Cached<DohRecord>, DohError>`. A definitive `NoAnswer` is a cached answer (`Ready(NoAnswer)`), never a failure; `DohError::{Throttled, QueueFull}` are removed.
- Email recovery: both DoH-resolution paths now build the same `PartialVerification` at email-arrival (`smtp_request`) and finish through one idempotent, **polled** `email_recovery_resolve_via_doh` (renamed from `email_recovery_submit_dkim_leaf_via_doh`) that reads the DoH cache and verifies off the partial once `Ready`. The pure-DoH (Gmail) path joins this shape, collapsing the inline DoH-completion machinery into a shared `VerifyMaterial` + `resolve_dkim_and_dmarc`. Status `Verifying -> ResolvingDoh`.
- Frontend: the recovery poll loop drives `resolve_via_doh` on each tick while `ResolvingDoh` (pure-DoH and DNSSEC->DoH fallback). A thrown error is handled — canister `Err` is terminal, transport failure surfaces a retryable failure — never silently re-polled. Drops the `queue_full`/`throttled` analytics sub-reasons; regenerates the candid bindings.

## Tests
- Cache unit tests for the poll state machine: fresh/stale/dead, dedup (`Pending` while in flight), stale-if-error serves last-good, cold-failure backoff then retry, backoff growth + reset, LRU, abandonment takeover. Added: backoff overflow saturation, `max_entries` clamp to ≥1, global sweep of all expired entries, `stale_for = 0` (stale-serving disabled), abandoned-fill takeover exercised through the public `get()`, LRU-clock overflow renumbering, and self-healing of a broken size bound.
- DoH unit tests updated for `Cached` (`NoAnswer` as a `Ready` value; transient failure reads as `Pending`, debounced). The TTL-expiry tests now assert the stale-while-revalidate contract (the stale value is served immediately with no synchronous fan-out, then exactly one background re-fetch picks up the rotated record), plus a stale-if-error test (a cached record survives a transient quorum failure and the refresh is debounced).
- DoH integration tests re-orchestrated for the poll model (answer provider outcalls by query name across polls; dedup invariant asserted). Added: `email_recovery_resolve_via_doh` is a silent no-op once the entry is terminal, binding the credential exactly once across repeated polls.
- Frontend: first unit tests for the shared email-recovery poll loop (`poll.ts`) — every status variant routed (`Pending` / `ResolvingDoh` / `NeedDkimLeaf` / `Failed` / `Expired` / success), both DoH paths (direct and the DNSSEC→DoH fallback), the submit-once guard, and canister-`Err` vs. transport-error handling on both the submit and resolve paths, plus poll-error ride-out vs. give-up-after-`MAX_CONSECUTIVE_POLL_ERRORS`.
- 560 bin unit tests + 23 email-recovery integration tests pass under PocketIC; candid compatibility + `cargo clippy --tests -D warnings` clean; frontend format + lint + svelte-check + 17 new `poll.ts` tests clean.


src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+        if let Some(inflight) = self.in_flight.get(key) {
+            let age = now_secs.saturating_sub(inflight.started_at_secs);
+            if age < self.pending_stale_after_secs {
+                return CacheState::Pending;
+            }
+            // Stale: the owner likely died before publishing. Fall through
+            // and take the fill over — the insert below overwrites the
+            // abandoned marker with a fresh `fill_id`, so the old owner's
+            // late `publish` becomes a no-op.
```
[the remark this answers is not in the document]
sea-snake: Addressed — the `publish` doc now states the precise guarantee ("a superseded fill can't clobber fresher state") and notes it still sweeps expired entries. (The PR has also grown since the overview above: stale-while-revalidate + stale-if-error, an LRU `max_entries` cap, exponential-backoff retry, and a `DohRecord { Txt | NoAnswer }` value so a definitive `NoAnswer` is cached rather than debounced.)

src/canister_tests/src/api/internet_identity.rs
```diff
@@ -641,10 +641,7 @@ pub fn email_recovery_submit_dkim_leaf(
     env: &PocketIc,
     canister_id: CanisterId,
     arg: types::email_recovery::EmailRecoverySubmitDkimLeafArg,
-) -> Result<
-    Result<types::email_recovery::EmailRecoveryStatus, types::email_recovery::EmailRecoveryError>,
-    RejectResponse,
-> {
+) -> Result<Result<(), types::email_recovery::EmailRecoveryError>, RejectResponse> {
```
aterga, line 644: Result of result is an anti-pattern. 

NCR as we had it already.

src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,905 @@
...
+                self.touch(key);
+                return Lookup::Ready(v.clone());
+            }
+        }
+        // A fill is in flight (cold, or a refill spawned by an earlier get), or
+        // a recent fill failed and we're backing off. Either way we don't start
+        // a fill here; serve the last-good value if we have one — stale-while-
+        // revalidate while the refill runs, stale-if-error during backoff —
+        // otherwise poll again (a cold fill has nothing to serve yet).
+        if inflight_fresh || is_throttled {
```
aterga, line 274: For my understanding: If `inflight_fresh` is true, how could we not have an entry? 

Maybe (due to the check on https://github.com/dfinity/internet-identity/pull/3995/changes#diff-fcee1fd6fb9e2c555f0d1e4ce54ec12ed5d42c7b296ccd7071ac0d0ec9c063f6R219) this condition could be simplified as

```suggestion
        if is_throttled {
```

src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,905 @@
...
+    fn touch(&mut self, key: &K) {
+        self.access_clock = self.access_clock.wrapping_add(1);
+        let stamp = self.access_clock;
+        if let Some(e) = self.entries.get_mut(key) {
+            e.lru_stamp = stamp;
+        }
+    }
+
+    fn upsert(&mut self, key: &K, mut entry: Entry<V>) {
+        self.access_clock = self.access_clock.wrapping_add(1);
```
aterga: So in the unlikely event that we get the access clock wrapped (for simplicity, let's assume we use 8-bit integers instead of 64-bit), we'd have the following cache configuration (assume the cache can hold 4 entries: a, b, c, d):

* a.lru_stamp = 252
* b.lru_stamp = 253
* c.lru_stamp = 254
* d.lru_stamp = 255

then we want to upsert an entry `e`:

lru_stamp = 255.wrapping_add(1) = 0

and after `self.evict_lru()` we get the same cache contents again:

* a: lru_stamp = 252
* b: lru_stamp = 253
* c: lru_stamp = 254
* d: lru_stamp = 255

because the victim is defined by the min lru_stamp value, which would be `e`, which is supposed to survive the upsert. 

To solve this problem, we should detect overflows (checked_add(1)) and shift left all the lru_stamp values of all the cache elements to end up with, e.g., 

* e: lru_stamp = 0
* a: lru_stamp = 1
* b: lru_stamp = 2
* c: lru_stamp = 3

so in this example, `d` would be evicted, not `e`.

src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,905 @@
...
+        }
+    }
+}
+
+impl<K: Eq + Hash + Clone, V: Clone, E> SingleFlightCache<K, V, E> {
+    /// Classify an incoming request for `key`, claiming the in-flight marker
+    /// for a `StartFill`. Drops a fully-dead entry first.
+    fn lookup(&mut self, key: &K, now: u64) -> Lookup<V> {
+        if self.entries.get(key).is_some_and(|e| !e.is_alive(now)) {
+            self.entries.remove(key);
```
aterga: Here we could call `sweep_expired` to sweep all expired entries, not just the current one.

src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,905 @@
...
+            self.entries.remove(&victim);
+        }
+    }
+
+    fn sweep_expired(&mut self, now: u64) {
+        self.entries.retain(|_, e| e.is_alive(now));
+    }
+
+    #[cfg(test)]
+    fn servable_value(&self, key: &K, now: u64) -> Option<V> {
```
aterga: Why do we need a test-only function like this? Unit tests have access to private fields of types in the same module, right?

src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,905 @@
...
+) where
+    K: Eq + Hash + Clone + 'static,
+    V: Clone + 'static,
+    E: 'static,
+{
+    let fill_fut = cache.with(|c| (c.borrow().fill)(key.clone()));
+    detach(async move {
+        let result = fill_fut.await;
+        let now = now();
+        cache.with(|c| c.borrow_mut().complete_fill(&key, token, result, now));
```
aterga:
```suggestion
        cache.with_borrow_mut(|c| c.complete_fill(&key, token, result, now));
```

src/internet_identity/src/single_flight_cache.rs
sea-snake: Since inflight is tracked separately, an entry could be evicted while the refresh is inflight.

src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,905 @@
...
+        let stamp = self.access_clock;
+        if let Some(e) = self.entries.get_mut(key) {
+            e.lru_stamp = stamp;
+        }
+    }
+
+    fn upsert(&mut self, key: &K, mut entry: Entry<V>) {
+        self.access_clock = self.access_clock.wrapping_add(1);
+        entry.lru_stamp = self.access_clock;
+        if !self.entries.contains_key(key) && self.entries.len() >= self.max_entries {
```
aterga, line 428: The canister should log if it observes `self.entries.len() > self.max_entries` since that would constitute a bug in the cache implementation, I think.

src/internet_identity/src/single_flight_cache.rs
```diff
@@ -0,0 +1,905 @@
...
+impl<K, V, E> SingleFlightCache<K, V, E> {
+    /// Build a cache with its shared async `fill` — run once per cold/stale
+    /// key under single-flight dedup — and its [`CacheConfig`]. Every policy
+    /// knob is required; see `CacheConfig` for why there are no defaults.
+    pub fn new<Fut>(fill: impl Fn(K) -> Fut + 'static, config: CacheConfig) -> Self
+    where
+        Fut: Future<Output = Result<V, E>> + 'static,
+    {
+        Self {
+            entries: HashMap::new(),
```
aterga: If we store entries in a BTreeMap, we'd be able to evict more efficiently at ~O(1), but lookups would be slightly slower at ~O(log(N)). 

I would opt for a BTreeMap here as they are also deterministic.

aterga (approved):
I'm okay merging this if you
1. additionally address the concerns pointed out in the code (do we have a bug in upsert at integer overflows?)
2. add more observability, in particular, all representable code paths that constitute internal inconsistency (e.g., cache size goes out of bounds) should be print-logged, and the cache state should be fully visible to the outside world (either via metrics or a dedicated query function(s))

src/canister_tests/src/api/internet_identity.rs
sea-snake: Yep — kept as-is to match every neighbouring helper in this file (`submit_dkim_leaf`, `prepare_delegation`, …): outer `Result` is the transport-level reject, inner is the candid-level error.

src/internet_identity/src/single_flight_cache.rs
sea-snake: Dug into this a bit more: the simplification would also break single-flight dedup during stale-while-revalidate, independent of the eviction case above. While a refill is in flight over a stale entry, `is_throttled` is false (`retry_at = fresh_until` after a success, and the entry is past that), so `inflight_fresh` is the only thing standing between a concurrent `get` and a second `StartFill` — i.e. a duplicate five-provider fan-out for the same key.

Verified empirically: with `if is_throttled {`, `refreshing_stale_serves_stale_to_concurrent_get` fails (the concurrent lookup claims a second fill instead of being served the stale value) and `stale_refresh_polls_through_to_fresh` counts 3 fills instead of 2.

src/internet_identity/src/single_flight_cache.rs
sea-snake: Good catch — fixed in f134864. `access_clock` now advances with `checked_add`; on overflow every entry's stamp is renumbered to its recency rank (order-preserving) and the clock restarts with full headroom, so a wrapped clock can never mark the newest entry least-recently-used. Covered by `lru_clock_overflow_renumbers_preserving_recency`.

src/internet_identity/src/single_flight_cache.rs
sea-snake: Done in f134864 — `lookup` now opens with `sweep_expired(now)`, which subsumes the single-key removal it replaced.

src/internet_identity/src/single_flight_cache.rs
sea-snake: Right — removed in f134864; the one test that used it reads the private fields directly like the others.

src/internet_identity/src/single_flight_cache.rs
sea-snake: Done in f134864, and converted the other call sites (`get`, `spawn_fill`, the test reset) to `with_borrow`/`with_borrow_mut` for consistency.

src/internet_identity/src/single_flight_cache.rs
sea-snake: Added in f134864, following the canister's `ic_cdk::println!("ERROR: …")` convention used in storage.rs/event_stats.rs. `upsert` also evicts in a loop now, so an over-cap state self-heals down to the bound instead of persisting. Test: `upsert_restores_a_broken_size_bound`.

src/internet_identity/src/single_flight_cache.rs
sea-snake: Switched in f134864 (`K: Ord` instead of `Eq + Hash`) — agreed on determinism: with tied `lru_stamp`s the eviction victim shouldn't depend on hash order. One note on the perf expectation: eviction stays O(n) — a `BTreeMap` keyed by `K` still needs a min-by-`lru_stamp` scan to find the LRU victim; O(1) eviction would need a secondary index ordered by stamp, which doesn't seem worth the bookkeeping at ≤256 entries.

sea-snake:
Addressed the second round (262ca8c):

**Integer overflow in `upsert` — no bug.** Audited the arithmetic: `next_stamp` advances the LRU clock with `checked_add` and, on overflow, renumbers every entry to its recency rank (order-preserving) before continuing — so a wrapped clock can never mark the newest entry least-recently-used. The size-bound logic is `usize` length comparisons with no arithmetic. Locked it down with `upsert_drives_overflow_renumber_and_eviction_together`, which forces the clock to `u64::MAX` so a single insert triggers both the renumber and an eviction, and asserts the just-inserted entry survives while the true LRU is evicted.

**Inconsistency logging.** Every "cannot happen on a correct cache" branch now routes through a `note_inconsistency` helper that print-logs `ERROR: single-flight cache invariant violated: …` and bumps a running counter. Covered branches: size-over-cap in `upsert`, `evict_lru` finding nothing to evict, and the `touch` / `next_stamp` "entry vanished" lookups.

**External visibility.** Added a read-only `CacheStats` snapshot (live entries split into fresh / valued / backing-off, plus in-flight count, the configured cap, and the running inconsistency total) via a `stats()` function mirroring `get()`. Wired into the canister `/metrics` endpoint as `internet_identity_doh_cache_*` gauges, so both the cache state and any invariant violation are observable from outside the canister, not just in the logs — including `internet_identity_doh_cache_inconsistencies`, which should always read 0.

Tests: 3 new cache unit tests + inconsistency-counter assertions, and the DoH cache gauges added to the metrics integration test. 560 unit + 23 email-recovery integration tests pass under PocketIC 9.0.3; clippy clean; fmt applied.

PR #4000, opened 2026-06-09
chore(ci): enable Dependabot version updates

MRmarioruci:
## Problem

The repo runs Dependabot security updates (reactive, vulnerability-triggered) but has no `dependabot.yml`, so version updates never fire. Deps drift forward silently and only surface once they become security issues — exactly what `update-deps.yml` (#3998) does for the toolchain, but for libraries.

## Changes

- New `.github/dependabot.yml` covering 4 ecosystems (npm, cargo, github-actions, docker).
- L3 grouping with two carve-outs:
  - npm: bulk PR + agent-js carve-out (`@dfinity/*` tied to IC release cadence).
  - cargo: bulk PR + `pocket-ic` carve-out (replica-coupled).
- Weekly Mon staggered (08/09/10 UTC) so reviews don't pile.
- Majors ignored across the board — manual upgrade path.
- CODEOWNERS already auto-requests reviews from `@dfinity/identity`.
- Commit prefix `chore(deps)` / `chore(deps-dev)` to match repo convention.

No auto-merge — every Dependabot PR gets human eyes on diff and CI before landing.

## Tests

- YAML schema validates locally before commit.
- First real fire Mon 08:00 UTC post-merge; expected ≤6 PRs/week steady state.


.github/dependabot.yml
```diff
@@ -0,0 +1,121 @@
...
+      - "/demos/vc_issuer"
+      - "/src/try-ii"
+    schedule:
+      interval: weekly
+      day: monday
+      time: "08:00"
+      timezone: Etc/UTC
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "chore(deps)"
```
aterga, line 17: Can we factor out the common stuff like PR prefix?

MRmarioruci: Not really. The dependabot yaml parser rejects things like anchors/aliases, shared defaults blocks etc.. https://github.com/dependabot/dependabot-core/issues/1582. There are workarounds with scripts etc.. but seems too dirty to me for this.

aterga (approved):

PR #4003, opened 2026-06-10
docs(be,fe): design doc for persisting the multiple-accounts toggle

MRmarioruci:
## Problem

The "Enable multiple accounts" toggle on the authorize screen is in-memory only — it resets on every page load and on every identity switch. Account data is persisted backend-side, but the UI gate is not, so users who enable multi-account on a dapp lose their per-app account list on the next visit and have to re-toggle to see their own accounts.

Persisting the toggle backend-side also surfaces a second problem: every II canister call rides on a device-rooted delegation capped at ~30 minutes, so a returning user would pay a full WebAuthn ceremony just to *render* their persisted preference.

## Changes

Adds `docs/persist-multiple-accounts-toggle.md` — a design doc in two parts:

- **Part 1**: per-`(anchor, application)` persistence of the toggle by extending the existing `AnchorApplicationConfig` stable structure. Covers context, alternatives considered (incl. why not localStorage), backend Candid surface, frontend wiring, and upgrade-safety considerations.
- **Addendum — scoped session delegations**: ceremony-free authentication for low-stakes calls (`get_accounts`, `get/set_anchor_application_config`, `get/set_default_account`). II stamps a longer-lived delegation (default 7 days) on a principal derived from `(salt, anchor, scope, epoch)`, recognized by re-derivation in a new opt-in authz branch — the existing `check_authorization` is untouched, so everything else stays full-auth by construction. Revocation is an additive epoch field on the anchor, bumped on any auth-method removal or explicitly. Zero new stable-memory structures; includes a mermaid flow diagram, failure modes, security analysis, and the stored-rows alternative with its adoption trigger.

No code changes.

## Tests

Doc only.

### DIFF
--- a/docs/persist-multiple-accounts-toggle.md
+++ b/docs/persist-multiple-accounts-toggle.md
@@ -1,0 +1,250 @@
+# Persist the "multiple accounts" toggle in the backend
+
+**Date:** 2026-06-10
+
+## TL;DR
+
+The "Enable multiple accounts" toggle on the authorize screen is in-memory only and resets on every page load and every identity switch. I propose persisting it per `(anchor, application)` by extending the existing `AnchorApplicationConfig` stable structure on the backend (please see [why not on browser storage](#dont-persist-use-browser-localstorage) below). Marginal storage is a few bytes per row where the toggle is set, with zero overhead for users who never enable it. No new storage map, no migration.
+
+## Context
+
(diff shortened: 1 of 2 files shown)


MRmarioruci:
Closing this in favor of the implementation in https://github.com/dfinity/internet-identity/pull/4014

PR #4010, opened 2026-06-11
fix(fe): always show OpenID provider name when logo is missing

MRmarioruci:
A configured OpenID provider with an empty or missing logo rendered as a button with no visible text — only an invisible aria-label — on the "Add access method" screen, so it looked blank.

## Changes
- The provider name now always renders on access-method buttons; only the logo is gated on whether the provider has one.
- Applied the same gating to the sign-in provider picker so a logo-less provider no longer leaves an empty icon box above the name.

<div align="left">Previous: #3991</div>


MRmarioruci:
Folding this into #3991 directly — not worth a separate PR for a two-line template fix. The commit is now on `refactor/fe-authwizard-blackbox-67`.

PR #4013, opened 2026-06-12
feat(be): add sso_domain/sso_name to stored OpenID credentials with batched backfill migration

sea-snake:
Implements **PR-M (track M)** of `docs/ongoing/openid-sso-prod-readiness.md` (§8.5 + §8.6): stored OpenID credentials gain `sso_domain` / `sso_name` fields so SSO metadata survives in stable storage independently of the in-memory `DISCOVERY_TASKS` state. This unblocks track C2, which deletes `DISCOVERY_TASKS` and the runtime `sso_fields_for` reverse-scan — after this PR, every SSO credential carries its own stamp.

## Changes

- `StorableOpenIdCredential` and the runtime `OpenIdCredential` gain optional `sso_domain` / `sso_name` fields (CBOR map keys 5 / 6). Records written before this change decode with `None`; a rolled-back wasm skips the unknown keys — both directions are upgrade-safe.
- `DiscoverableProvider::verify` stamps the fields at verification time. Since `openid_prepare_delegation` rewrites the stored credential from a fresh verification on every sign-in, actively-used credentials also self-heal/refresh their stamp. Direct providers (Google / Microsoft / Apple) stamp `None`.
- API response shaping (`OpenIdCredential → OpenIdCredentialData`) prefers the stored stamp and falls back to the live `DISCOVERY_TASKS` lookup for unstamped credentials. The fallback is removed once the migration has completed.
- New `InternetIdentityInit::sso_credential_migration` upgrade arg (`vec` of `{discovery_domain, issuer, client_id, sso_name}`) backfills existing credentials by `(iss, aud)`. The deployer builds the list from the running canister's `discovered_oidc_configs` query before submitting the upgrade proposal; the proposal itself is the audit trail.
- The backfill is a timer-driven batched migration following the prior-migration convention (#3784, #3713): 2 000 index keys examined per 1 s tick so each batch fits in one ingress message, cursor threaded between batches, timer self-clears on completion. Idempotent — already-stamped credentials are skipped, so re-submitting a corrected list is safe.
- Temporary hidden monitoring endpoints `sso_credential_migration_status` (query: `(stamped, is_done)`) and `list_sso_credential_migration_errors`, to be removed with the rest of the scaffolding once the migration completed (same lifecycle as #3997).

## Tests

- Storable CBOR tests: round-trip with the new fields, decoding pre-migration records (keys 5 / 6 absent → `None`).
- `migrate_sso_credentials_batch` unit tests: stamping, non-matching/already-stamped skips, batch splitting with cursor resumption, orphan index entries, idempotent re-run.
- Integration test `should_backfill_sso_fields_via_credential_migration`: link a credential, upgrade with the migration arg, advance the timer, assert `get_anchor_info` returns the stamped fields.
- Full suites pass: 556 unit + 386 integration (incl. upgrade & rollback against the latest release wasm).


src/internet_identity/src/storage.rs
```diff
@@ -864,6 +882,133 @@ impl<M: Memory + Clone> Storage<M> {
...
+                    }
+                    credential.sso_domain = Some(entry.discovery_domain.clone());
+                    credential.sso_name = entry.sso_name.clone();
+                    stamped_any = true;
+                }
+                if stamped_any {
+                    self.stable_anchor_memory
+                        .insert(anchor_number, storable_anchor);
+                    outcome.stamped += 1;
+                }
```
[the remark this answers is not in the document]
sea-snake: Fixed in 1ddd553c: the migration now counts per credential actually stamped (outcome.stamped += stamped_in_anchor) rather than once per anchor, and only writes the anchor back when something changed. Today the index key is the full (iss, sub, aud) so the filter matches at most one credential per anchor, but the counter is now honest if that ever changes.

src/internet_identity/src/openid/generic.rs
aterga, line 881: Why does this doc string no longer belong to `sso_fields_for`?

src/internet_identity/src/storage.rs
```diff
@@ -864,6 +882,138 @@ impl<M: Memory + Clone> Storage<M> {
...
+                    .iter_mut()
+                    .filter(|credential| credential.key() == key)
+                {
+                    // Idempotency: already-stamped credentials are skipped, so
+                    // re-running (e.g. with a corrected entry list) is safe.
+                    if credential.sso_domain.is_some() {
+                        continue;
+                    }
+                    credential.sso_domain = Some(entry.discovery_domain.clone());
+                    credential.sso_name = entry.name.clone();
```
aterga, line 1004: What if `credential.sso_name.is_some`? Should we log that case?

src/internet_identity/src/storage.rs
```diff
@@ -864,6 +882,138 @@ impl<M: Memory + Clone> Storage<M> {
...
+            let Some(entry) = entries
+                .iter()
+                .find(|entry| entry.issuer == key.iss && entry.client_id == key.aud)
+            else {
+                continue;
+            };
+            if anchor_numbers.is_empty() {
+                let err = format!(
+                    "index entry without anchor: iss={} sub={}",
+                    key.iss, key.sub
```
aterga: We should either not aggregate errors containing private data like iss and sub or guard the observability queries with an admin principal.

aterga (approved):

src/internet_identity/src/openid/generic.rs
sea-snake: Good catch — addressed in 7323fcbb. The doc belonged to sso_fields_for; sso_name_for_domain had been inserted between the doc and the function, orphaning it. Moved the helper above the doc block so it reattaches to sso_fields_for.

src/internet_identity/src/storage.rs
sea-snake: Addressed in 7323fcbb. sso_domain and sso_name are always written together, so a set sso_name with an unset sso_domain should not happen; we now log that case (anchor number only, no PII) before overwriting rather than hiding it.

src/internet_identity/src/storage.rs
sea-snake: Addressed in 7323fcbb. Dropped iss/sub from the aggregated error strings since list_sso_credential_migration_errors is anonymous. Kept the anchor number, which is enough to investigate and is not IdP PII. Went with stripping the PII rather than gating the endpoint, to stay consistent with the other (ungated) hidden migration endpoints.

PR #4014, opened 2026-06-12
feat(be,fe): session delegations for ceremony-free account reads

MRmarioruci:
After signing in to id.ai, the authorize screen's account list still re-prompts for a WebAuthn ceremony every time the device-rooted delegation expires (~30 min). This POC introduces anchor-scoped backend session delegations so the account list and default-account reads on return visits render without prompting the user to re-tap a passkey.

## Changes

- Backend: new `prepare_session_delegation` / `get_session_delegation` endpoints. The delegated principal is derived from `H(salt, "session-delegation", anchor)` and recognized by `check_session_authorization`, which sits alongside `check_authorization` in `authz_utils.rs` and returns a sealed `CallerCapability` so it's the only entry point for scoped authz. `check_authorization` itself is unchanged — every other endpoint stays full-auth by construction. Scope is limited to `get_accounts` and `get_default_account` (reads only). Minting requires full auth; email-recovery callers are rejected.
- Frontend: after every full sign-in, mint a session delegation for a non-extractable browser keypair stored in IndexedDB. On return visits the account list reads on the authorize screen route through that delegation. Sign-in (`prepare_account_delegation`), account create/update/rename, and `set_default_account` stay full-auth — those handlers force `authLastUsedFlow.authenticate()` before any out-of-scope call. Mint is fire-and-forget and degrades to the status quo on failure.

## Tests

- Rust integration suite at \`src/internet_identity/tests/integration/session_delegation.rs\` covering the happy path through \`get_accounts\`/\`get_default_account\` via a session principal, default-deny for out-of-scope endpoints (\`create_account\`, \`update_account\`, \`identity_info\`, \`set_default_account\`), TTL clamp and default, upgrade survival of the derived principal, and isolation across anchors.
- Vitest coverage for the session-delegation store (IndexedDB persistence, expiry margin, mint/purge round-trip) and the keypair/signing util.
- Playwright spec at \`src/frontend/tests/e2e-playwright/routes/authorize/session-delegation.spec.ts\` exercising the ceremony-free return-visit path.


src/internet_identity/src/session_delegation.rs
```diff
@@ -0,0 +1,342 @@
...
+fn scope_tag(scope: SessionScope) -> &'static [u8] {
+    match scope {
+        SessionScope::AccountManagement => b"account_management",
+    }
+}
+
+pub(crate) fn session_delegation_seed(
+    anchor_number: AnchorNumber,
+    scope: SessionScope,
+    epoch: u32,
```
aterga:
```suggestion
    expirtion_timestamp_ns: u32,
```

src/internet_identity/src/main.rs
```diff
@@ -381,7 +382,10 @@ fn get_accounts(
     anchor_number: AnchorNumber,
     origin: FrontendHostname,
 ) -> Result<Vec<AccountInfo>, GetAccountsError> {
-    match check_authorization(anchor_number) {
+    match check_authorization_with_scope(
+        anchor_number,
+        internet_identity_interface::internet_identity::types::SessionScope::AccountManagement,
```
aterga:
```suggestion
        SessionScope::AccountManagement,
```

src/internet_identity/src/session_delegation.rs
```diff
@@ -0,0 +1,342 @@
...
+    hasher.update(&blob);
+    hasher.finalize().into()
+}
+
+pub enum CallerCapability {
+    FullAuth(Box<Anchor>, AuthorizationKey),
+    SessionScoped,
+}
+
+pub fn check_authorization_with_scope(
```
aterga: Can we compile-time enforce that this function is the only way to check scoped authorizations?

src/internet_identity/src/session_delegation.rs
```diff
@@ -0,0 +1,342 @@
...
+    anchor_number: AnchorNumber,
+    required_scope: SessionScope,
+) -> Result<CallerCapability, AuthorizationError> {
+    if let Ok((anchor, key)) = check_authorization(anchor_number) {
+        return Ok(CallerCapability::FullAuth(Box::new(anchor), key));
+    }
+
+    let salt_initialised = state::storage_borrow(|storage| storage.salt().is_some());
+    if salt_initialised {
+        let anchor = state::anchor(anchor_number);
```
aterga: Nit. Consider factoring out the expected principal computation into a helper function.

src/internet_identity/src/session_delegation.rs
```diff
@@ -0,0 +1,342 @@
...
+};
+use serde_bytes::ByteBuf;
+use sha2::{Digest, Sha256};
+
+pub const DEFAULT_SESSION_DELEGATION_TTL_NS: u64 = 7 * crate::DAY_NS;
+pub const MAX_SESSION_DELEGATION_TTL_NS: u64 = 30 * crate::DAY_NS;
+
+fn scope_tag(scope: SessionScope) -> &'static [u8] {
+    match scope {
+        SessionScope::AccountManagement => b"account_management",
```
sea-snake: I would avoid concepts like session scoping for now, we can always add that later.

When we get back to this later, it probably makes more sense to use identity-info payload for this purpose instead. Else you'd need to iterate over every possible scope derived principal during authorization, instead of just checking identity info. 

Ideally a principal should indicate who and how is being authenticated while identity-info can be used for additional information e.g. scopes, roles etc.

src/internet_identity/src/storage/storable/anchor.rs
```diff
@@ -28,6 +28,12 @@ pub struct StorableAnchor {
     /// pattern as `passkey_credentials` / `recovery_keys` above.
     #[n(5)]
     pub email_recovery: Option<Vec<StorableEmailRecoveryCredential>>,
+    /// Monotonically-increasing counter used to scope session
+    /// delegations. Bumping it invalidates all outstanding
+    /// session-scoped delegations for this anchor. `None` for
+    /// anchors serialised before this field was added; reads as 0.
+    #[n(6)]
+    pub session_delegation_epoch: Option<u32>,
```
sea-snake: This is a rather creative approach for "invalidate all sessions for anchor".

The only reason it works is due to the delegation being targeted towards a single canister (II backend), so we can rely on stateful checks like this. For delegations towards dapps, this wouldn't work since this would change the principal, thus breaking apps that expect a stable principal.

Not entirely sure if adding this now makes a lot of sense though, since we don't have "sign out from all devices" functionality to begin with in our current UX. I'd argue that given the current scope (read only accounts), invalidation can always be added later when needed with this or another approach.

MRmarioruci: I am not sure as well, cause we dont only have the case where we want to invalidate all sessions for an anchor.. but also the case where you've minted a session delegation with a passkey for example, you remove that passkey, maybe in that case you also need to invalidate all the sessions minted with given passkey. That would require a different approach where we save in the seed the credential id hash or something and also store the epochs per credential. So I will probably remove it for now, but it's an option for the future

src/internet_identity/src/session_delegation.rs
MRmarioruci: not relevant anymore, removed the concept of epochs

src/internet_identity/src/main.rs
MRmarioruci: Moot in 82ecccf8

src/internet_identity/src/session_delegation.rs
MRmarioruci: Done by sealing CallerCapability, both variants wrap a Sealed token whose field is module-private, so no code outside session_delegation.rs can construct one. main.rs can still match on the variants but can't forge one, which makes check_session_authorization the only mint site.

That covers the "forged capability" angle at compile time. The other angle, forgetting the gate entirely on a new scoped endpoint, would want something like #[scoped_query]. II has no proc macros today, and to stay consistent the macro would also have to absorb the existing match check_authorization(...) boilerplate on the full-auth endpoints. Should I explore that path or no? Also I exlored the inspect_message macro but that won't work on query endpoints.

src/internet_identity/src/session_delegation.rs
MRmarioruci: Done in 82ecccf8

src/internet_identity/src/session_delegation.rs
MRmarioruci: removed

src/internet_identity/src/main.rs
```diff
@@ -563,14 +562,22 @@ fn set_default_account(
...
-            account_number,
-        )?;
-        Ok((result, Operation::SetDefaultAccount))
-    })
+    let capability = check_session_authorization(anchor_number)
+        .map_err(|err| SetDefaultAccountError::Unauthorized(err.principal))?;
+
+    if let CallerCapability::FullAuth(mut anchor, authorization_key, _) = capability {
+        anchor_management::activity_bookkeeping(&mut anchor, &authorization_key);
+        state::storage_borrow_mut(|storage| storage.write(*anchor)).map_err(|err| {
```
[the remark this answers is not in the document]
MRmarioruci: This is genuinely how the helper works, and it applies to every endpoint that uses `check_authz_and_record_activity` (10 call sites, not just this one).

I'd rather not fix it locally just for `set_default_account`, that leaves one endpoint with different failure semantics than every other authz-and-bookkeeping endpoint in the canister, which is a confusing inconsistency for the next reader. The right move is either:

1. Redesign the helper to defer the anchor write until after success (matching `anchor_operation_with_authz_check` semantics), or
2. Document that bookkeeping persists on failure as an accepted trade-off.

Either belongs in a follow-up that touches all 10 call sites together.

src/internet_identity/src/session_delegation.rs
```diff
@@ -0,0 +1,207 @@
...
+pub fn get_session_delegation(
+    anchor_number: AnchorNumber,
+    session_key: SessionKey,
+    expiration: Timestamp,
+) -> Result<SignedDelegation, SessionDelegationError> {
+    check_authorization(anchor_number)
+        .map_err(|err| SessionDelegationError::Unauthorized(err.principal))?;
+
+    let seed = session_delegation_seed(anchor_number);
+
```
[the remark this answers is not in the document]
MRmarioruci: Good catch — fixed in [`0081ab82`](https://github.com/dfinity/internet-identity/commit/0081ab82). Added the same `storage.salt().is_some()` guard already used in `check_session_authorization`: if the salt isn't set yet, no session could have been minted (since `prepare_session_delegation` awaits `ensure_salt_set` before stamping), so the call short-circuits with `NoSuchDelegation` instead of trapping.

src/frontend/src/routes/(new-styling)/authorize/views/ContinueView.svelte
```diff
@@ -180,38 +248,76 @@
...
       const index = accounts.findIndex(
-        (account) =>
-          account.account_number[0] === isEditAccountDialogVisibleForNumber,
+        (acc) => acc.account_number[0] === isEditAccountDialogVisibleForNumber,
       );
-      // Only update the name when it has changed
-      if (account.name !== (accounts[index].name[0] ?? primaryAccountName)) {
-        accounts[index] = await actor
+      const nameChanged =
+        account.name !== (accounts[index].name[0] ?? primaryAccountName);
```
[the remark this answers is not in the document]
MRmarioruci: Real bug, fixed in [`21782e5d`](https://github.com/dfinity/internet-identity/commit/21782e5d). Added an early return when `findIndex` returns `-1`.

src/frontend/src/lib/flows/authLastUsedFlow.svelte.ts
```diff
@@ -58,6 +62,10 @@ export class AuthLastUsedFlow {
           identityNumber,
           authMethod: { passkey: { credentialId } },
         });
+        void mintSession({
```
sea-snake: Nit: Session minting isn't awaited, this means that the page could in theory be closed before a session has actually been minted. Or that a session mint can be in progress while the authenticated identity has already been changed.

Awaiting it here isn't ideal either since that would slow all authentication flows down for all users, so for now these concerns can probably be ignored. In the future we can revisit if problems occur ("enable multiple accounts" toggle is not enabled when it should be).

sea-snake (approved):
LGTM, left one minor nit comment but that's not a showstopper for this PR.

PR #4015, opened 2026-06-12
feat(fe): switch OpenID callback to response_mode=form_post

sea-snake:
Sign-in with Apple silently loses the user's name and email, and some strict OIDC providers (Okta, Auth0) reject our hybrid-flow callback (by default): both are artifacts of `response_mode=fragment`, which OAuth 2.1 removes entirely and which also exposes the id_token in the callback URL. This implements Track A (§7) of the OpenID/SSO production-readiness design: the IdP now delivers the OAuth response with `response_mode=form_post` instead.

## Changes

- The IdP POSTs `{id_token, state}` to `/callback`; the frontend canister upgrades the POST to update mode (so the response is certified) and returns an HTML page whose single, CSP-hash-pinned inline script delivers the payload to the frontend: via `BroadcastChannel` for popup flows, via sessionStorage + `/authorize?flow=openid-resume` for 1-click same-tab flows. The flows are discriminated by the `ii-openid-authorize-state` marker — `window.opener` cannot tell them apart since the authorize tab is itself a popup opened by the relying party.
- The POST arrives anonymously (the IdP submits the form), so the handler is a pure transport translator: the salt + nonce + `caller()` JWT redemption binding still happens through the existing signed-ingress flow.
- RFC 6749 error reports (`{error, error_description, state}`) travel through the same page, so a misconfigured SSO app still surfaces its own message as `OAuthProviderError` instead of a generic failure.
- Bodies that can't be translated (malformed, missing fields, oversized) get a `303` redirect to a new in-SPA error page (`/callback-error`) that renders the styled, localized error UI; the machine reason rides along as a query param to aid debugging a misconfigured IdP.
- The callback page's CSP is scoped to its single hash-pinned inline script via a `content_security_policy_override` parameter on the shared header builder, rather than carving the CSP out of the SPA-wide headers.
- Frontend: `createRedirectURL` requests `form_post`; `extractIdTokenFromCallback` validates the structured payload (state check first); `resumeOpenId` reads the payload from sessionStorage.
- Hot-reload dev: a server hook performs the same translation (no canister fronts the dev server) and passes the canister's `303` through unfollowed; SvelteKit's CSRF origin check is disabled since it would reject the IdP's cross-origin form POST before the hook runs — adapter-static ships no server in production, so the check only ever applied to the dev server. In NO_HOT_RELOAD e2e the vite forward plugin routes the POST to the real canister handler.
- The legacy fragment callback page is removed: the live popup utilities moved to `$lib/utils/openID.ts`.

## Tests

- Canister unit tests for the form parser, validation bounds, JSON/HTML escaping and CSP script-hash pinning, plus property tests for the charset predicates and the JSON embedding (output is always free of raw `<`/`>`/`&` and round-trips).
- PocketIC integration tests: POST `/callback` upgrade flag, success/error translation, malformed-body redirect to `/callback-error`, security headers, 405s.
- FE unit tests updated for the payload shape; `createRedirectURL` form_post coverage.
- Existing OpenID/SSO e2e suites pass against the real canister handler (1-click same-tab and wizard popup paths).


sea-snake:
Thanks @copilot — both addressed in 50a8af18:

- **CSP hash pinning:** the `/callback` response now gets its own policy (`default-src 'none'; script-src 'sha384-…'; base-uri 'none'; frame-ancestors 'none'`; the error page has no `script-src`) instead of inheriting the SPA-wide `script-src 'self' 'unsafe-inline' 'unsafe-eval'`. The SPA CSP is replaced on this response, not appended, so the hash actually governs execution rather than relying on the CSP3 rule that a hash makes browsers ignore `unsafe-inline`. Unit + PocketIC integration tests now assert exactly one CSP header carrying the hash and no `unsafe-inline`/`unsafe-eval`.
- **405 Allow header:** narrowed from `GET, POST` to `GET` — `method_not_allowed` is only reached for non-`/callback` resources, which serve GET only (POST is handled on `/callback` via the query→update upgrade).

src/internet_identity_frontend/internet_identity_frontend.did
```diff
@@ -67,4 +67,7 @@ type InternetIdentityFrontendInit = record {
 
 service : (InternetIdentityFrontendInit) -> {
     http_request : (request : HttpRequest) -> (HttpResponse) query;
+    // Handles the OAuth `response_mode=form_post` callback POSTed by the IdP
+    // to /callback, upgraded from `http_request` so the response is certified.
```
aterga: remove comment. 

this function can be used in many ways, no need to clarify one use case in the API

src/internet_identity_frontend/tests/integration/http.rs
```diff
@@ -363,3 +363,184 @@ fn verify_response_certification(
...
+        method: "POST".to_string(),
+        url: "/callback".to_string(),
+        headers: vec![(
+            "Content-Type".to_string(),
+            "application/x-www-form-urlencoded".to_string(),
+        )],
+        body: ByteBuf::from(b"id_token=abc.def.ghi&state=c3RhdGU".to_vec()),
+        certificate_version: None,
+    };
+    let response = http_request(&env, canister_id, &request)?;
```
aterga, line 386: Why is this PR adding integration tests for a pre-existing endpoint?

src/internet_identity_frontend/tests/integration/http.rs
```diff
@@ -363,3 +363,184 @@ fn verify_response_certification(
...
+        install_ii_frontend_canister(&env, II_FRONTEND_WASM.clone(), default_frontend_args());
+
+    let request = HttpRequest {
+        method: "POST".to_string(),
+        url: "/authorize".to_string(),
+        headers: vec![],
+        body: ByteBuf::new(),
+        certificate_version: None,
+    };
+    let response = http_request(&env, canister_id, &request)?;
```
aterga, line 407: Ditto

src/internet_identity_frontend/src/main.rs
```diff
@@ -142,6 +165,21 @@ fn certify_all_assets(args: InternetIdentityFrontendArgs) {
     });
 }
 
+/// Headers for dynamically rendered responses, built from the retained
+/// [`HeaderConfig`] so they match the headers certified onto static assets.
+/// The caller supplies its own `Content-Security-Policy` among
+/// `additional_headers`, so no inline-script integrity hashes are passed here.
+pub(crate) fn dynamic_response_headers(additional_headers: Vec<HeaderField>) -> Vec<HeaderField> {
```
aterga: where is this function used?

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+use ic_http_certification::{HeaderField, HttpRequest, HttpResponse, Method, StatusCode};
+use sha2::Digest;
+use std::borrow::Cow;
+
+pub const CALLBACK_PATH: &str = "/callback";
+
+/// Upper bound on the accepted form body. The largest legitimate body is an
+/// `id_token` near its own cap plus the small `state` and `code` fields;
+/// anything bigger is rejected before parsing.
+const MAX_BODY_BYTES: usize = 16 * 1024;
```
aterga: > its own cap plus

where is this documented?

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+const MAX_ID_TOKEN_BYTES: usize = 8192;
+const MAX_STATE_BYTES: usize = 64;
+const MAX_ERROR_BYTES: usize = 256;
+const MAX_ERROR_DESCRIPTION_BYTES: usize = 1024;
+
+/// The single executable inline script of the landing page. Constant so its
+/// CSP hash is constant; the per-request payload lives in a non-executing
+/// `<script type="application/json">` data block instead.
+const CALLBACK_SCRIPT: &str = r#"(function () {
+  var data = JSON.parse(document.getElementById("cb").textContent);
```
aterga: What is `cb`? Can we make this ID more verbose?

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+}
+
+fn html_response(
+    status_code: StatusCode,
+    html: String,
+    content_security_policy: String,
+) -> HttpResponse<'static> {
+    // Take the shared security headers, then swap in this page's own CSP in
+    // place of the SPA-wide one (see `render_callback_landing`).
+    let mut headers: Vec<HeaderField> = dynamic_response_headers(vec![
```
aterga: This function seems a bit hacky, why not add a parameter to `dynamic_response_headers` to control which CSP it produces rather than carving those out and substituting them here?

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+        "default-src 'none'; script-src '{}'; base-uri 'none'; frame-ancestors 'none'",
+        callback_script_hash()
+    );
+    html_response(StatusCode::OK, html, csp)
+}
+
+/// Static page for bodies that can't be translated. `reason` is always one
+/// of the fixed strings from [`parse_form_post`], never attacker-controlled.
+fn render_error_page(reason: &str) -> HttpResponse<'static> {
+    let html = format!(
```
aterga: Could we move this into an asset file? 

Could we easier to iterate on its styling in the future.

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+pub fn callback_script_hash() -> String {
+    let hash = sha2::Sha384::digest(CALLBACK_SCRIPT.as_bytes());
+    format!("sha384-{}", BASE64.encode(hash))
+}
+
+/// Serialize the payload for embedding in the JSON data block. `serde_json`
+/// escapes `"` and `\`; `<`, `>` and `&` are additionally escaped so the
+/// serialized form can never contain `</script` even if a validation bug
+/// lets a hostile value through.
+fn payload_json(payload: &CallbackPayload) -> String {
```
aterga: This could be a serde decorator on `CallbackPayload` IIUC

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+fn is_state_charset(value: &str) -> bool {
+    !value.is_empty()
+        && value
+            .bytes()
+            .all(|byte| byte.is_ascii_alphanumeric() || matches!(byte, b'_' | b'-' | b'='))
+}
+
+/// RFC 6749 `error` / `error_description` charset: printable ASCII except
+/// `"` (0x22) and `\` (0x5C).
+fn is_rfc6749_charset(value: &str) -> bool {
```
aterga, line 247: Any chance getting these predicates from a lightweight library?

aterga: If not, we need to property-test them.

aterga: Ditto for all other predicates defined in this file.

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+
+/// Parse an `application/x-www-form-urlencoded` body into a
+/// [`CallbackPayload`]. Unknown fields (e.g. the `code` that accompanies our
+/// `response_type=code id_token` request) are ignored; for duplicate keys the
+/// first occurrence wins. All accepted fields are validated against the
+/// charset and length the OAuth/JWT specs allow, so the payload can't carry
+/// anything able to break out of the JSON data block it is embedded in.
+fn parse_form_post(body: &[u8]) -> Result<CallbackPayload, &'static str> {
+    if body.len() > MAX_BODY_BYTES {
+        return Err("form body too large");
```
aterga: Always print the observed size and expected bounds to facilitate potential debugging.

aterga: Ditto elsewhere

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+            None => None,
+        };
+        return Ok(CallbackPayload::ProviderError {
+            error: error.into_owned(),
+            error_description,
+            state: state.into_owned(),
+        });
+    }
+
+    Err("missing id_token")
```
aterga: Ideal code flow:

1. Preconditions are invalid ? return Err : proceed
2. Business logic failed ? return Err : proceed
3. Ok(result)

src/frontend/src/routes/(new-styling)/authorize/+page.svelte
```diff
@@ -165,22 +224,36 @@
...
+    }
+    let jwt: string;
+    try {
+      jwt = extractIdTokenFromCallback(
+        JSON.parse(storedPayload),
+        openIdAuthorizeState,
+      );
+    } catch {
+      // A state mismatch, an IdP error report or a missing token falls
+      // back to the regular flow, matching the fragment-era behavior.
```
aterga: What is the fragment-era behavior?

src/internet_identity_frontend/src/callback.rs
```diff
@@ -0,0 +1,527 @@
...
+        Err(reason) => render_error_page(reason),
+    }
+}
+
+/// Parse an `application/x-www-form-urlencoded` body into a
+/// [`CallbackPayload`]. Unknown fields (e.g. the `code` that accompanies our
+/// `response_type=code id_token` request) are ignored; for duplicate keys the
+/// first occurrence wins. All accepted fields are validated against the
+/// charset and length the OAuth/JWT specs allow, so the payload can't carry
+/// anything able to break out of the JSON data block it is embedded in.
```
aterga: The comment is quite vague about the thread model, while implying that there is one. 

Could you be more specific in the commend re. what are the bad things that we're protecting against?

aterga (approved):
LGTM modulo comments

src/internet_identity_frontend/internet_identity_frontend.did
sea-snake: Done in `24761a2a` — comment removed.

src/internet_identity_frontend/src/main.rs
sea-snake: It's called from the callback module (`callback.rs`) to build the landing-page and redirect response headers. In `24761a2a` it also gained a `content_security_policy_override` parameter (see the reply on the `html_response` thread) so callers can supply their own CSP.

src/internet_identity_frontend/src/callback.rs
sea-snake: Reworded in `24761a2a`: each field cap now documents its own rationale, and `MAX_BODY_BYTES` is defined as `2 * MAX_ID_TOKEN_BYTES`, so the relationship is explicit in code rather than vague prose.

src/internet_identity_frontend/src/callback.rs
sea-snake: Renamed to `callback-payload` in `24761a2a`.

src/internet_identity_frontend/src/callback.rs
sea-snake: Agreed — done in `24761a2a`. `dynamic_response_headers`/`get_asset_headers` now take a `content_security_policy_override: Option<String>`; the landing page passes its scoped CSP through it and the filter-then-re-add hack is gone (the now single-use `html_response` helper was inlined).

src/internet_identity_frontend/src/callback.rs
sea-snake: Reworked in `24761a2a`: untranslatable bodies now `303`-redirect to a new in-SPA Svelte page (`/callback-error`) that renders the styled, localized error UI, so the markup lives in a normal route that's easy to iterate on.

Note for context: there were no standalone *served* error pages to mirror — error UX elsewhere is in-SPA views — and the canister returns this response outside the SPA, so redirecting to a real route was the cleanest fit. Valid RFC 6749 provider errors still come back through the landing page as before; only malformed/untranslatable bodies redirect.

src/internet_identity_frontend/src/callback.rs
sea-snake: Done in `24761a2a` — `CallbackPayload` now derives `Serialize` with `#[serde(untagged)]` (flat per-variant object; `error_description` serialized as `null` when absent, which the frontend schema already normalizes). `payload_json` just serializes and applies the HTML-significant-character escaping.

src/internet_identity_frontend/src/callback.rs
sea-snake: No dependency gives us a bare charset predicate that fits here — a JWT library would parse/verify the token, which is the wrong layer: this endpoint only checks the value is safe to embed in the HTML, and the signed-ingress redemption authenticates the token later. So I property-tested them in `24761a2a`:

- the JWT and `state` charsets reject every embedding-dangerous character;
- the RFC 6749 charset excludes the JSON string delimiters and any control / non-ASCII byte;
- end to end, `payload_json` output is always free of raw `<`/`>`/`&` and round-trips — even for the error fields whose charset deliberately permits those (the escaping layer closes that gap).

src/internet_identity_frontend/src/callback.rs
sea-snake: Done in `24761a2a` — the parser error type is now `String` and the bound errors name the observed size and the limit, e.g. `id_token too large: 8193 bytes exceeds limit of 8192`. The charset and length checks are split so each reports precisely.

src/internet_identity_frontend/src/callback.rs
sea-snake: Restructured in `24761a2a` into the guard-clause order: bounded body → present & valid `state` → if a token is present, validate it and return (success) → otherwise require an error report → single trailing `Ok` for the provider-error variant.

src/frontend/src/routes/(new-styling)/authorize/+page.svelte
sea-snake: Reworded in `24761a2a` to describe the current fallback (a state mismatch, an IdP error report, or a missing token returns to the regular sign-in flow) without the historical reference. I also swept the rest of the diff for stale "fragment used to…" comments and updated those.

src/internet_identity_frontend/src/callback.rs
sea-snake: Expanded in `24761a2a` to name the concrete risks: (1) HTML/script-context breakout — a value containing `</script>` or angle brackets could close the `<script type="application/json">` data block, which the per-field charset checks reject and `payload_json` escapes as defence in depth; (2) unbounded work on this anonymously-reachable endpoint — per-field and whole-body size caps. Plus a note that this layer is transport-safety only; the `id_token` is authenticated later by the signed-ingress redemption.

src/internet_identity_frontend/tests/integration/http.rs
sea-snake: The endpoint (`http_request`) pre-exists, but the behaviour these two tests assert is new in this PR:

- `frontend_canister_upgrades_callback_post_to_update` checks that `http_request` now returns `upgrade = true` for `POST /callback`. That upgrade decision is exactly the change this PR makes to `http_request` in `main.rs`, and it's the precondition for the whole feature — without it the boundary node can't deliver a certified dynamic response, so the callback can't work at all.
- `frontend_canister_rejects_post_to_other_paths` checks the other side of that new branch: a POST to any other path is *not* upgraded (405, no upgrade), so we don't pay for an update call on arbitrary POSTs.

Neither can be a unit test: the upgrade flag is set by the gateway-facing `http_request` in `main.rs`, so asserting it means going through the canister entry point. The body translation itself is covered separately by the `callback.rs` unit tests and by `frontend_canister_translates_form_post_callback` (which additionally needs PocketIC for the certified update call). Happy to rename these if the names read as "testing `http_request`" rather than "testing the new `POST /callback` upgrade routing".

src/internet_identity_frontend/tests/integration/http.rs
sea-snake: Same point as the sibling comment — this one asserts the negative case of the new upgrade branch (a POST to any non-`/callback` path is not upgraded: 405, no update call). See the reply above.

aterga (approved):

PR #4016, opened 2026-06-12
feat(be,fe): read-only mode via delegations with queries-only permissions

aterga:
# Motivation

Let a user grant **read-only (query-only) access** when authorizing a relying party — a dapp (`/continue`), the CLI (`/cli`), or an MCP server (`/mcp`). A read-only authorization can read on the user's behalf but cannot change state: it relies on the optional `permissions = "queries"` field on request delegations (interface-spec extension drafted in dfinity/ic#10449), which the IC enforces at the **protocol level** by rejecting update calls (`/call` requests) authenticated through such a delegation. Enforcement is not up to any app or canister.

# Changes

**Delegation signing & types (backend)**
- `Delegation` gains an optional `permissions` field (`opt text` in candid; `Option` in the Rust interface). `"queries"` restricts the delegation to query calls; absent means unrestricted (`"all"`), per the spec.
- `add_delegation_signature` folds an optional `permissions` value into the representation-independent hash, so the value is covered by the canister signature (byte-identical to the upstream helper when `permissions` is absent).
- A new candid `type Permissions = variant { queries; all }` (mirroring the ICP protocol's request-delegation permission values) is the request-side knob. `prepare_account_delegation` / `get_account_delegation` gain a trailing `permissions : opt Permissions`; `queries` yields a delegation carrying `permissions = "queries"`. An **omitted** arg (or `all`) means **unrestricted**, preserving the original argument-count behavior (see "Backwards compatibility"). Internally the wire enum maps to a `DelegationAccess` domain enum at the boundary (`DelegationAccess::from`), so no blind booleans cross a function signature.
- Regenerated candid bindings (`internet_identity_idl.js`, `internet_identity_types.d.ts`).

**Reusable UI + the three sign-in flows (frontend)**
- New `AccessLevelToggle.svelte` (checkbox + tooltip, bindable `accessLevel`, `prompt` prop selecting which level the checkbox opts into), used in all three flows. Per-flow defaults:
  - `/continue`: **full access** by default ("Read-only mode" opt-in).
  - `/cli`: **read-only** by default ("Full access" opt-in).
  - `/mcp`: **read-only** by default ("Full access" opt-in).
- A typed `AccessLevel = "read-only" | "full-access"` union threads through `authorizationStore` and the flows; the candid `permissions : opt Permissions` argument (`{ queries }` / `{ all }`) is produced only at the actor-call boundary (`toPermissionsArg`). All first-party callers send an explicit value.

**MCP read-only (backend)**
- `mcp_set_access` gains `permissions : opt Permissions`, persisted with the grant (as a bool). The MCP server's **standing delegation stays full-access** (required so it can still call the update endpoint `mcp_prepare_account_delegation`); the read-only restriction is applied to the **per-app delegations** the server mints, by reading the persisted flag.
- Storage unified: the `Principal → AnchorNumber` reverse index and the read-only flag collapse into one `Principal → StorableMcpAccess { anchor_number, read_only }` map. `caller_access()` returns both in a single lookup, so the MCP server still never passes an anchor number. On a fresh memory index — MCP is preview-only, so the old regions are abandoned and any preview grants are recreated on the next connect (no migration).

**Backwards compatibility:** the `permissions` argument is optional and an *omitted* value (or `all`) yields an unrestricted delegation, so existing callers of the pre-feature argument form keep getting update-capable delegations (no silent downgrade), matching the interface spec's default for an absent `permissions` field. Read-only is opt-in via the argument; the per-flow UI defaults above are independent of this (always sent explicitly).

**Note on relying parties:** returning a restricted delegation to a dapp requires the agent library to round-trip the `permissions` field, and no released agent stack does yet (checked `ic-agent` / `ic-transport-types` through 0.47.3, and `@icp-sdk/core`). Until then a restricted delegation **fails closed** on the dapp side — its signature doesn't verify without the field, which the ingress tests below prove at the replica.

# Tests

- **End-to-end ingress enforcement** (`delegation_ingress::`, new): real signed request envelopes submitted to a live PocketIC endpoint (`make_live`), exercising the replica's actual request authentication — which the impersonating PocketIC test transport skips. The envelopes are hand-rolled (CBOR + ed25519 + representation-independent hash) since no released agent can carry `permissions`. Asserted for both mint paths (account delegations, serving `/continue` and `/cli`; MCP per-app delegations):
  - a queries-only delegation authenticates **query** calls and executes as the delegated principal;
  - a queries-only delegation is **rejected at ingress for update** calls with the protocol error ("Update calls are not permitted: a delegation restricts the sender to query calls") — before any canister code runs;
  - an unrestricted delegation authenticates both query and update calls (also confirming dfinity/ic#10226's `subnet_type` cert check doesn't affect pocket-ic v15's canister-sig delegations);
  - **fail closed**: resubmitting a queries-only delegation with the `permissions` field stripped from the envelope is rejected as an invalid signature — omitting the restriction cannot escalate access.
- **Persistence & format (backend units):** `mcp_read_only_grant_persists_across_upgrade` (a read-only grant still mints queries-only after a canister upgrade); `StorableMcpAccess` round-trip + pinned wire-format; a pinned signable-hash vector for the `permissions = "queries"` delegation, plus a check that the absent-permissions message is byte-identical to the upstream helper.
- **Delegation mint semantics (impersonating transport):** read-only account delegation carries `permissions = "queries"` and the signature binds to it (an unrestricted lookup yields `NoSuchDelegation`); an omitted arg yields unrestricted; read-only / full-access MCP grants mint queries-only / unrestricted per-app delegations respectively. `verify_delegation` folds both `permissions` and `targets` into the representation-independent hash.
- **Frontend units:** `toPermissionsArg` mapping (`read-only → { queries }`, `full-access → { all }`); the `AccessLevelToggle` prompt×checked mapping (all combinations); `authorizationStore` records the access level; the `/cli` and `/mcp` authorize helpers send the correct `permissions` variant and keep the MCP standing delegation full-access. Plus e2e assertions that each flow's checkbox starts in its documented default state.
- The `accounts::`, `mcp::`, `session_delegation::`, and `delegation_ingress::` integration suites pass against the built wasm (on `pocket-ic-server 15.0.0`).
- Static checks clean: `cargo clippy --all-targets`, `tsc --noEmit` / `svelte-check` (0 errors), `eslint`, `prettier`, and the candid interface check.

> Bindings note: the pinned `didc` release used by `npm run generate` is currently unreachable from this environment, so the regenerated `internet_identity_idl.js` / `internet_identity_types.d.ts` were produced by hand and validated (tsc + parsing the idl); a `npm run generate` pass on `main` will confirm they match canonically.

https://claude.ai/code/session_018gHRjyFYqumMxAMwbPRLuN


src/internet_identity/src/main.rs
```diff
@@ -612,6 +617,7 @@ fn get_account_delegation(
             account_number,
             session_key,
             expiration,
+            read_only.unwrap_or(false),
```
aterga: Wouldn't it be more sensible to fall back to readonly in the BE? Ditto elsewhere

src/internet_identity/src/main.rs
```diff
@@ -589,6 +592,7 @@ async fn prepare_account_delegation(
                 account_number,
                 session_key,
                 max_ttl,
+                read_only.unwrap_or(false),
```
aterga: Ditto

src/internet_identity/src/main.rs
aterga: Done in 94002fb (covers both this and the `get_account_delegation` "Ditto").

Both endpoints now default an omitted `read_only` to read-only (`unwrap_or(true)`), with a comment noting this deliberately diverges from the interface spec (where an absent `permissions` field means unrestricted/"all" for backward compatibility) and that we err on the side of least privilege.

Because the default flips, first-party callers now pass an explicit value so they aren't silently downgraded to queries-only:
- The ICRC-34 delegation handler sends `[false]` for normal authorizations and `[true]` only when the user enabled read-only mode.
- The CLI authorize flow sends `[false]` (CLI sessions are update-capable).
- The plain `prepare/get_account_delegation` test helpers pass `Some(false)` so existing tests keep exercising unrestricted delegations; the signature-binding check now looks up with `Some(false)` (omitting now means read-only), and a new test `should_default_to_read_only_account_delegation_when_unspecified` locks in the default.

The candid docs on both endpoints now document the omitted-defaults-to-read-only behavior. Full `accounts::` integration suite is green (23 tests).

---

src/internet_identity/internet_identity.did
```diff
@@ -157,6 +157,10 @@ type Delegation = record {
     pubkey : PublicKey;
     expiration : Timestamp;
     targets : opt vec principal;
+    // Restricts the kinds of calls the delegation permits: `"queries"`
+    // restricts the sender to query calls (the IC rejects update calls
+    // authenticated through such a delegation). Absent means unrestricted.
```
aterga, line 162: Wrong. Absent should mean readonly on the BE!

unixway:
I'm concerned that this PR changes the semantics of an existing Candid-compatible API in a way that can break already deployed canister integrations.

Old callers of `prepare_account_delegation` / `get_account_delegation` encode the previous 5-argument form. With the new trailing `opt bool`, those calls still decode successfully, but `read_only = null` now defaults to `true`. That silently changes the delegation from the previous unrestricted behavior to query-only. This is a breaking change: existing canister integrations may keep "working" at the Candid layer while failing later when they try to use the delegation for update calls.

There is a second compatibility concern: the signed delegation format / verification preimage must not change in-place for the existing API. Some deployed canisters verify II delegations inside wasm using the existing signed message shape. If II starts signing `permissions` as part of the delegation while old verifiers reconstruct the previous message, those canisters will reject valid delegations. For immutable or hard-to-upgrade canisters, this is not recoverable after an II mainnet upgrade.

The backward-compatible behavior should be:

- omitted `read_only` / `null` preserves the old unrestricted delegation behavior;
- `read_only = opt true` explicitly opts into query-only delegations;
- the existing signed delegation verification semantics remain valid for existing callers.

If query-only delegations require a different signed format, please introduce that through a versioned delegation type or new endpoint rather than changing the meaning of existing successful calls.

src/internet_identity/internet_identity.did
aterga: I think there are two distinct "absents" here, and the BE default you want is already in place — let me split them:

1. **This `permissions` field** is the *protocol* field from the interface spec (developer-docs#292), where an **absent `permissions` field = `"all"` = unrestricted**. This `Delegation` record is what `get_*_delegation` *returns*, and II only ever emits the field absent when the delegation is genuinely unrestricted (i.e. the caller passed `read_only = false`). So "absent means unrestricted" is correct for the field — flipping it to read-only would contradict #292 (and would mean a queries-only delegation with no `permissions` field, which the IC wouldn't enforce).

2. **The `read_only` argument** on `prepare_account_delegation` / `get_account_delegation` is the BE knob, and that one **does default to read-only when omitted** (`read_only.unwrap_or(true)`) — exactly the behavior you're asking for. It's documented on those args.

So the BE already defaults to read-only when the caller doesn't specify; this line was only documenting the returned protocol field's meaning. I've clarified the comment in 3ba8e73 to make the distinction explicit. If you actually intended the protocol field's semantics to change (absent field ⇒ queries-only at the IC level), that'd be a change to the interface spec in #292 rather than to II — happy to discuss if that's the intent.

---

src/canister_tests/src/framework.rs
```diff
@@ -661,6 +661,12 @@ pub fn verify_delegation(
             Value::Number(signed_delegation.delegation.expiration),
         ),
     ];
+    if let Some(permissions) = &signed_delegation.delegation.permissions {
+        key_value_pairs.push((
+            "permissions".to_string(),
+            Value::String(permissions.clone()),
+        ));
+    }
```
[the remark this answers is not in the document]
aterga: Good catch — fixed in 5ac8fdc. `verify_delegation` now folds `targets` into the representation-independent hash (mirroring the canister signer, between `expiration` and `permissions`; the rep-independent hash is order-independent anyway). No behavior change for the existing tests, which all pass `targets: None`, but the helper now matches the interface spec and its own comment.

---

src/internet_identity/tests/integration/mcp.rs
```diff
@@ -150,6 +151,132 @@ fn mcp_mints_per_app_delegation_authorized_by_caller() -> Result<(), RejectRespo
     Ok(())
 }
 
+/// A read-only access grant (`mcp_set_access` with `read_only = true`) makes the
+/// per-app delegations the server mints query-only: they carry
+/// `permissions = "queries"`, and the signature binds to it (an unrestricted
+/// lookup of the same key yields `NoSuchDelegation`). The standing delegation is
+/// untouched — read-only is a property of the grant, applied per minted
+/// delegation — so the server can still call the (update) prepare endpoint.
```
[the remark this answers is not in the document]
aterga: Fixed in 5ac8fdc — reworded the doc comment to match what the test actually asserts (the minted per-app delegation carries `permissions = "queries"` and verifies against that message). The MCP path derives `read_only` from the persisted grant, so there's no per-call `read_only` argument and thus no unrestricted-lookup variant to contrast here; that signature-binding proof (a differing value ⇒ `NoSuchDelegation`) lives in `accounts::should_get_read_only_account_delegation_with_queries_permissions`, which the comment now cross-references.

---

src/internet_identity/src/main.rs
```diff
@@ -589,6 +592,15 @@ async fn prepare_account_delegation(
...
                 max_ttl,
+                // Default an omitted `read_only` argument to a read-only
+                // (queries-only) delegation: when the caller does not specify,
+                // we err on the side of least privilege. This intentionally
+                // differs from the interface spec, where an absent `permissions`
+                // field means unrestricted ("all") for backward compatibility.
+                // First-party callers (the II frontend and the CLI flow) always
+                // pass an explicit value, so this default only affects callers
+                // that omit the argument entirely.
+                read_only.unwrap_or(true),
```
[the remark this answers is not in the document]
aterga: Confirmed intended. Defaulting an omitted `read_only` to read-only (queries-only) is a deliberate fail-safe chosen by the maintainers (see the adjacent discussion on `internet_identity.did`): err toward least privilege when the caller doesn't specify. All first-party callers — the II frontend authorize flow, the CLI flow, and the MCP connect flow — pass an explicit value, so this default only affects callers that omit the trailing arg entirely. It does diverge from the interface spec's unrestricted default for an absent `permissions` *field*, and that divergence is called out in the `read_only` arg docs in `internet_identity.did`.

---

aterga:
Thanks for raising these concerns, @unixway,

> Old callers of `prepare_account_delegation` / `get_account_delegation` encode the previous 5-argument form. With the new trailing `opt bool`, those calls still decode successfully, but `read_only = null` now defaults to `true`. That silently changes the delegation from the previous unrestricted behavior to query-only. This is a breaking change: existing canister integrations may keep "working" at the Candid layer while failing later when they try to use the delegation for update calls.

This observation is correct. Out of curiosity, could you list some existing canister integrations that would break?

Regardless, we'll consider keeping the more-liberal defaults (similar to the corresponding ICP protocol extension's liberal defaults, e.g., if permissions are not specified, the delegation is treated as a write-access delegation).

> Some deployed canisters verify II delegations inside wasm

Could you clarify what you mean by "inside wasm," please?

Delegations are normally verified by the ICP protocol. 

For context, the new (optional) `permissions` extension is going to be supported by all the ICP subnets ETA = next Monday.

src/frontend/src/routes/(new-styling)/mcp/utils.ts
```diff
@@ -92,6 +102,9 @@ export const mcpAuthorize = async ({
       [],
       ephemeralPublicKey,
       [maxTimeToLiveNanos],
+      // Unrestricted: MCP delegations are update-capable. The backend defaults
+      // an omitted `read_only` to read-only, so pass `[false]` explicitly.
```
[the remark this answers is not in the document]
aterga: Fixed in 28239ed — the comment was stale from before the backwards-compat revert (the backend now defaults an omitted `read_only` to unrestricted). Reworded per your suggestion: the explicit `[false]` documents that the standing delegation is deliberately update-capable, passed explicitly rather than relying on the backend's default.

---

src/internet_identity/Cargo.toml
```diff
@@ -77,3 +77,9 @@ ic-response-verification.workspace = true
 proptest.workspace = true
 flate2 = "1.0"
 getrandom = "0.2"
+# Read-only delegation enforcement tests (`delegation_ingress.rs`): submit
+# real signed request envelopes to a `make_live` PocketIC endpoint. The
+# envelopes are hand-rolled (no ic-agent) because no released agent stack
+# can carry the delegation `permissions` field yet.
+ed25519-dalek = "2.2"
+reqwest = { version = "0.12", features = ["blocking"] }
```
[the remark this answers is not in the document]
aterga: Done in d075a01 — `default-features = false`. The `make_live` endpoint is plain HTTP on localhost, so no TLS backend is needed at all; the openssl/native-tls entries are gone from `Cargo.lock` (199 lines dropped).

---

src/internet_identity/Cargo.toml
```diff
@@ -77,3 +77,9 @@ ic-response-verification.workspace = true
 proptest.workspace = true
 flate2 = "1.0"
 getrandom = "0.2"
+# Read-only delegation enforcement tests (`delegation_ingress.rs`): submit
+# real signed request envelopes to a `make_live` PocketIC endpoint. The
+# envelopes are hand-rolled (no ic-agent) because no released agent stack
+# can carry the delegation `permissions` field yet.
+ed25519-dalek = "2.2"
```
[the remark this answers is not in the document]
aterga: Done in d075a01 — now `ed25519-dalek.workspace = true`. No extra features needed: signing works with the workspace's `default-features = false` definition (all four ingress tests pass).

---

src/internet_identity/tests/integration/delegation_ingress.rs
```diff
@@ -0,0 +1,474 @@
...
+    live_url: &str,
+    canister_id: Principal,
+    endpoint: &str,
+    envelope: Vec<u8>,
+) -> (u16, Vec<u8>) {
+    let url = format!(
+        "{live_url}api/v2/canister/{}/{endpoint}",
+        canister_id.to_text()
+    );
+    let client = reqwest::blocking::Client::new();
```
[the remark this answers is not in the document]
aterga: Done in d075a01 — the blocking client is now built with a 60-second timeout, so a stalled endpoint fails the test deterministically instead of hanging the suite.

---

src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -12,6 +12,11 @@ export type AuthorizationContext = {
 
 export type Authorized = {
   accountNumberPromise: Promise<bigint | undefined>;
+  /** Whether the user restricted this authorization to read-only access:
+   *  the session delegation will carry `permissions = "queries"`, which
+   *  makes the Internet Computer reject update calls authenticated through
+   *  it. Enforcement is protocol-level, not up to the app or canister. */
+  readOnly: boolean;
```
nmattia: nit: would avoid blind booleans; generally not ideal but especially so for sensitive fields like these (throughout the changes)

src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -30,9 +35,13 @@ export const authorizationStore = {
...
   /** Called by the UI when the user authorizes with a specific account.
    *  Accepts a promise so the animation can start immediately while the
-   *  account number resolves asynchronously. */
-  authorize: (accountNumberPromise: Promise<bigint | undefined>): void => {
-    authorizedInternal.set({ accountNumberPromise });
+   *  account number resolves asynchronously. `readOnly` restricts the
+   *  session to read-only access (see {@link Authorized.readOnly}). */
+  authorize: (
+    accountNumberPromise: Promise<bigint | undefined>,
+    readOnly = false,
```
nmattia: same comment regarding boolean blindness, even more so here that the argument is not named. Also applies to the rust changes.

src/frontend/src/routes/(new-styling)/cli/views/CliAuthorizeView.svelte
```diff
@@ -10,17 +11,21 @@
...
+    onAuthorize: (readOnly: boolean) => Promise<void>;
   }
 
   const { domain, onAuthorize }: Props = $props();
 
   let busy = $state(false);
+  // CLI access defaults to read-only (opt-out): a linked CLI usually reads on
+  // the user's behalf, so it gets query-only access unless the user opts into
+  // full access by unchecking.
+  let isReadOnlyMode = $state(true);
```
nmattia: Here it might make sense (from a UX perspective) to flip this around and have an opt-in "full rights" or "RW" checkbox

src/frontend/src/routes/(new-styling)/mcp/utils.ts
```diff
@@ -92,6 +102,10 @@ export const mcpAuthorize = async ({
       [],
       ephemeralPublicKey,
       [maxTimeToLiveNanos],
+      // Unrestricted: the standing delegation is update-capable so the MCP
+      // server can call the (update) prepare endpoint. Passed explicitly
+      // rather than relying on the backend's default for an omitted value.
+      [false],
```
nmattia: I'm a bit out of the loop. What prevents the MCP server from hanging on to this delegation? What's the TTL?

src/internet_identity/src/storage/storable/mcp_access.rs
```diff
@@ -0,0 +1,41 @@
...
+/// the grant here — rather than a bare anchor number plus a parallel read-only
+/// index — means a single lookup by the caller's principal yields everything the
+/// `mcp_*_account_delegation` methods need: the MCP server never passes an
+/// anchor number, and `read_only` travels with the grant it restricts.
+///
+/// The `anchor_number` cannot be dropped: a principal is a one-way hash, and the
+/// anchor is required to derive the per-app account seeds the server acts under.
+#[derive(Encode, Decode, Clone, Debug, Eq, PartialEq)]
+#[cbor(map)]
+pub struct StorableMcpAccess {
```
nmattia, line 19: this is effectively a cache entry, correct? aside from making sure we remember the last selected value for RO/RW for a particular anchor/number, this does not store anything meaningful, right?

src/internet_identity/src/delegation.rs
```diff
@@ -143,6 +143,57 @@ pub fn add_delegation_signature(
...
+    }
+    if let Some(permissions) = permissions {
+        m.push(("permissions".into(), Value::String(permissions.to_string())));
+    }
+    representation_independent_hash(m.as_slice()).to_vec()
+}
+
+/// Like [`add_delegation_signature`], but restricts the delegation to the
+/// given permissions when `read_only` is `true`.
+pub fn add_delegation_signature_with_permissions(
```
nmattia: nit: why not make the "permissions" (`read_only`) an explicit `Option` argument in `add_delegation_signature` instead of duplicating it?

src/internet_identity/Cargo.toml
```diff
@@ -77,3 +77,10 @@ ic-response-verification.workspace = true
 proptest.workspace = true
 flate2 = "1.0"
 getrandom = "0.2"
+# Read-only delegation enforcement tests (`delegation_ingress.rs`): submit
+# real signed request envelopes to a `make_live` PocketIC endpoint. The
+# envelopes are hand-rolled (no ic-agent) because no released agent stack
+# can carry the delegation `permissions` field yet. The endpoint is plain
+# HTTP on localhost, so reqwest needs no TLS backend.
+ed25519-dalek.workspace = true
+reqwest = { version = "0.12", default-features = false, features = ["blocking"] }
```
nmattia: nit: 0.13 is out

src/frontend/src/lib/stores/authorization.store.ts
aterga: Done in 85849c8 — the frontend now threads a typed `AccessLevel = "read-only" | "full-access"` union through the store, the ICRC-34 delegation handler, and all three flows; the candid `read_only : opt bool` is produced only at the actor-call boundary (`toReadOnlyArg`). No boolean crosses a function boundary anymore.

---

src/frontend/src/lib/stores/authorization.store.ts
aterga: Done in 85849c8. `authorize()` now requires an explicit `AccessLevel` (no default), so every call site is self-describing — the 1-click flows (SSO, OpenID, post-upgrade) pass `"full-access"` explicitly. On the Rust side a `DelegationAccess` enum (`Unrestricted` / `ReadOnly`) replaces `read_only: bool` through `account_management` and `mcp`; the candid boundary maps via `DelegationAccess::from_read_only_arg`, and the legacy `prepare/get_delegation` endpoints now read `DelegationAccess::Unrestricted` instead of a bare `false`.

---

src/frontend/src/routes/(new-styling)/cli/views/CliAuthorizeView.svelte
aterga: Done in 85849c8 — the toggle component now takes a `prompt` prop deciding which opt-in the checkbox offers, so ticking the box always *changes* the flow's default. `/cli` now shows an opt-in "Full access" checkbox (unchecked by default; the default remains read-only), while `/continue` and `/mcp` keep their existing presentations.

---

src/frontend/src/routes/(new-styling)/mcp/utils.ts
aterga: Nothing prevents the server from *holding* it — like any delegation it's valid until expiry — but three things bound what holding it is worth:

1. **TTL**: the connect screen lets the user pick the standing delegation's lifetime, from 10 minutes to 1 week (the request suggests a default, e.g. 1 hour; the backend caps any requested TTL at 30 days).
2. **Scope + revocation**: the standing delegation only authenticates the server as the anchor's principal *at the MCP origin*, so it's only useful for calling II's `mcp_*` endpoints — and each of those re-checks the persisted grant on every call. The user can revoke at any time (`mcp_set_access(false)`), which makes the standing delegation useless immediately, well before it expires.
3. **What actually touches apps**: the delegations the server acts with at other apps are the per-app ones it mints on demand, and those are capped at **5 minutes** (`MCP_MAX_EXPIRATION_PERIOD_NS`).

---

src/internet_identity/src/storage/storable/mcp_access.rs
aterga: Not a cache — it's the authorization itself. The `Principal → anchor` mapping is what authorizes the caller: `mcp_prepare/get_account_delegation` recover the anchor from `caller()` through this map, and an absent entry means the caller is rejected. It also can't be re-derived from anything: the principal is a one-way hash of the user key, so without this index there's no way back to the anchor. Removing the entry is exactly how revocation works (`mcp_set_access(false)`), and `read_only` rides along as the restriction of that grant. So it's the opt-in state, not derived data.

---

src/internet_identity/src/delegation.rs
aterga: Done in 85849c8 — merged: `add_delegation_signature` now takes the `permissions` value as an explicit `Option<&str>` argument (callers pass `access.permissions()` or `None`), and the `_with_permissions` duplicate is gone.

---

src/internet_identity/Cargo.toml
aterga: Done in 85849c8 (0.13.4). One note: `pocket-ic` still depends on reqwest 0.12 transitively, so both majors now appear in the test build's lockfile — unavoidable until pocket-ic bumps.

---

nmattia (approved):
Thanks!

src/internet_identity/internet_identity.did
```diff
@@ -1800,7 +1822,11 @@ service : (opt InternetIdentityInit) -> {
     mcp_set_access : (
         anchor_number : UserNumber,
         mcp_server_origin : FrontendHostname,
-        enabled : bool
+        enabled : bool,
+        // `opt read_only` restricts the per-app delegations this server can
+        // later obtain to query calls (queries-only). Omitted (`null`) means
+        // full access.
+        access : opt AccessLevel
```
aterga: why do we reinvent the wheel? 

let's just call this the same way as in the ICP protocol:


```suggestion
        permissions : opt Permissions
```

aterga: Done in 2290448 — all three request args are now `permissions : opt Permissions` with

```candid
type Permissions = variant { queries; all };
```

Case names are the protocol's own permission values (the `pocket-ic` v15 replica knows both `"queries"` and `"all"`; its rejection message is `permissions = "queries"`). `queries` → queries-only; `all` → unrestricted; omitted → unrestricted (backwards-compat unchanged). The internal `DelegationAccess` domain enum + `permissions()` stay, so `account_management`/`mcp` below the boundary are untouched; the wire enum maps in via `DelegationAccess::from`.

Two deliberate choices, flagging in case you'd shape them differently:
- I kept an explicit `all` variant (rather than `variant { queries }` with absence = all) so first-party callers send an explicit value instead of relying on the omitted-default — matching the "always explicit" principle from the earlier boolean discussion. Absence still means `all` for backwards compat.
- The returned `Delegation.permissions` stays `opt text` (`"queries"`) — that's the actual protocol wire string the canister signature covers, so I left it as the spec's text field rather than typing the response too.

Verified: candid interface check, clippy, 43 integration + 12 FE unit tests, tsc/eslint, and the regenerated idl parses to `opt variant { queries; all }`.

---

PR #4020, opened 2026-06-15
feat(be): collapse email-recovery DNS verification to a single DoH path, gating DNSSEC behind a deploy flag

sea-snake:
Email recovery verifies a sender's DKIM/DMARC DNS records through **two parallel code paths**: a client-supplied DNSSEC proof that the canister validates cryptographically, and DoH (DNS-over-HTTPS) records the canister fetches itself via HTTP outcalls. Maintaining both is a significant complexity burden — the DNSSEC path alone is ~85 KB of verification, wire-parsing, and signature code, with its own frontend chain-walking logic and a second canister entry point (`email_recovery_submit_dkim_leaf`).

We want to standardize on the single DoH-outcalls implementation. This PR is the first step: it puts the DNSSEC path behind a deploy flag, defaulting to **off**, so we can run DoH-only in production and delete the DNSSEC code entirely in a follow-up — while keeping it available per-deployment as a fallback during the transition.

## Changes
- Add `enable_dnssec_email_recovery : opt bool` to `InternetIdentityInit`, stored in `PersistentState` (and the storable mirror, so it survives upgrades). Defaults to **off**.
- When off, `email_recovery_credential_prepare_add` / `email_recovery_prepare_delegation` drop any client-supplied `dns_proof` and take the DoH allowlist path; `opt true` re-enables the DNSSEC path. With the path gated at prepare time, `email_recovery_submit_dkim_leaf` is unreachable for new flows.
- Regenerate the frontend candid bindings, and enable the flag in the DNSSEC integration tests, `scripts/dev-e2e-setup`, and the e2e job in `canister-tests.yml` so they keep exercising the path.

## Tests
- New unit tests in `prepare.rs`: flag off drops the proof and takes DoH (rejected when not allowlisted, succeeds when allowlisted); flag on routes the proof to the DNSSEC verifier.


src/internet_identity/tests/integration/email_recovery.rs
```diff
@@ -566,6 +637,56 @@ fn dnssec_path_takes_precedence_over_doh_allowlist() {
...
 
+#[test]
+fn register_with_dnssec_disabled_and_non_allowlisted_email_fails() {
+    // Setup (register) flow with the DNSSEC path disabled (the default):
+    // even when the caller supplies a DNSSEC proof, the canister drops it
+    // and falls through to the DoH allowlist. `example.com` is not on the
+    // allowlist (`setup_canister` only allows `test.example.com`), so the
+    // call fails with `DomainNotAllowlisted`.
+    let env = env();
+    let canister_id = setup_canister(&env);
```
aterga:
```suggestion
    let canister_id = setup_canister_dnssec_disabled(&env);
```

src/internet_identity/tests/integration/email_recovery.rs
```diff
@@ -566,6 +637,56 @@ fn dnssec_path_takes_precedence_over_doh_allowlist() {
...
+
+#[test]
+fn recovery_with_dnssec_disabled_and_non_allowlisted_email_fails() {
+    // Recovery (prepare_delegation) flow, same scenario: DNSSEC disabled
+    // (default) + a supplied proof that gets dropped + a non-allowlisted
+    // domain → `DomainNotAllowlisted`. The allowlist gate runs before any
+    // address→anchor resolution, so an unbound address still surfaces the
+    // allowlist rejection.
+    let env = env();
+    let canister_id = setup_canister(&env);
```
aterga:
```suggestion
    let canister_id = setup_canister_dnssec_disabled(&env);
```

aterga (approved):

PR #4022, opened 2026-06-16
refactor(be,fe): data-only OpenID providers with on-demand SSO discovery/JWKS caches

sea-snake:
OpenID sign-in dispatched through trait objects (`dyn OpenIdProvider`) and resolved SSO organization domains client-side. This reworks it into a single data-driven verification pipeline with two clearly separated JWK sources of truth, moves SSO discovery onto the canister, and makes the "cache still warming" state a first-class retry signal instead of an error. No user-visible behavior changes for the configured Google / Microsoft / Apple providers; SSO (organization) sign-in now resolves and verifies entirely canister-side.

## Changes

**Backend**
- Replace the `dyn OpenIdProvider` trait dispatch and the monolithic `openid/generic.rs` with a data-only provider model: one shared verification pipeline (`openid/verify.rs`, `provider.rs`, `jwks.rs`) that is identical for every provider up to the point it reads the JWK.
- Two JWK sources of truth behind that shared pipeline: configured providers (`openid/configured.rs`) keep their timers + stable-storage-persisted keys (some providers can't reach HTTP-outcall consensus); SSO providers (`openid/sso.rs`) source keys on demand through the single-flight cache.
- Canister-side SSO discovery: `discover_sso : (text) -> ()` (fire-and-forget drive of the two-hop discovery+JWKS fetch) and `get_sso_discovery : (text) -> (SsoDiscoveryState) query` where `SsoDiscoveryState = variant { Resolved : SsoDiscovery; Pending; NotAllowed }`. Gated by the `sso_discoverable_domains` allowlist. Replaces the client-side discovery module.
- The four OpenID JWT methods take an optional SSO discovery domain and return `Pending` as a result arm (not an error) when the SSO discovery/JWKS cache is cold or has been evicted:
  - `openid_credential_add`, `openid_prepare_delegation`, `openid_get_delegation` return `OpenIdResult<T, E> = variant { Ok; Pending; Err }`.
  - `openid_identity_registration_finish` returns `variant { Ok; Pending; Err }`; its verification is hoisted into the endpoint so the JWT is verified exactly once and the pubkey-registration path is untouched.
  - `Pending` is removed from `OpenIdCredentialAddError` / `OpenIdDelegationError`.

**Frontend**
- SSO discovery polls `get_sso_discovery` and drives `discover_sso` while `Pending`; `ssoDiscovery.ts` no longer fetches discovery documents itself.
- Sign-in, account-link, and signup all retry while the canister reports `Pending` (shared `retryWhilePending` / poll loop), so a cold or evicted SSO cache no longer turns into a hard error.

## Tests
- New `config/sso_discovery.rs` integration tests for the `discover_sso` / `get_sso_discovery` allowlist gate.
- Updated OpenID / attributes / v2_api registration integration suites and api helpers for the new return shapes.
- SSO Playwright e2e green (12/12) across the discovery, sign-in, link, and signup paths.

<div align="right">Next: #4045</div>


sea-snake:
Merged `main` (only the two generated candid bindings conflicted — regenerated from the merged `.did` with the pinned didc; `npm run generate` yields no further diff).

Went through the 6 Copilot comments; only one was a real issue:

- **sso.rs — `discovery_domain` stamped verbatim** ✅ Fixed. Canonicalized (trim + lowercase) at the **canister boundary** in all four endpoints that accept `discovery_domain` (`openid_credential_add`, `openid_prepare_delegation`, `openid_get_delegation`, `openid_identity_registration_finish`) via a shared `openid::canonical_discovery_domain`, rather than inside `resolve()` — the arg is untrusted and the allowlist gate is case-insensitive, so it must be canonicalized where it enters, matching the existing `sso_discoverable_domains` setter and `get_sso_discovery`. Added a unit test.
- **verify.rs:20 `Storable` "unused"** ❌ It is used — `caller().to_bytes()` needs the trait in scope; removing it fails the build (E0599). No change.
- **verify.rs:318 `email_verified` "won't compile"** ❌ Current code matches with `ref s`, which binds by reference and doesn't move out of `&Claims`. Compiles. No change.
- **internet_identity.rs 422/441/460 — helpers "fail candid" without the trailing arg** ❌ Candid lets a caller omit trailing optional arguments (decoded as `null`); these are configured-provider helpers that intentionally pass no domain. Confirmed by the integration suite (19/19 OpenID tests green). No change.

Validation: clippy `-D warnings` clean, cargo fmt, frontend check/lint/format clean, OpenID integration 19/19, SSO discovery 2/2, openid unit 15/15.

src/canister_tests/src/api/internet_identity/api_v2.rs
```diff
@@ -75,7 +75,7 @@ pub fn openid_identity_registration_finish(
         "openid_identity_registration_finish",
         (openid_arg.clone(),),
     )
-    .map(|(x,)| x)
+    .map(|(x,)| super::settled(x))
```
aterga, line 78: why do we need this wrapping?

src/internet_identity/src/openid/jwks.rs
```diff
@@ -0,0 +1,118 @@
...
+pub(super) async fn fetch_jwks(jwks_uri: String) -> Result<Vec<Jwk>, String> {
+    use ic_cdk::api::management_canister::http_request::{
+        http_request_with_closure, CanisterHttpRequestArgument, HttpHeader, HttpMethod,
+    };
+
+    let request = CanisterHttpRequestArgument {
+        url: jwks_uri,
+        method: HttpMethod::GET,
+        body: None,
+        max_response_bytes: None,
```
MRmarioruci: Nit. Maybe we wanna limit this a bit? An allowlisted but compromised provider can return a 2 MB(which i see is the default) JWKS or HTML error page; ~2 MB × n nodes × every retry.

sea-snake: You're right, adding limits there and in other places now.

sea-snake:
Bounded the two SSO single-flight caches (`DISCOVERY_CACHE`, `JWKS_CACHE`) so canister state stays bounded once discovery domains / `jwks_uri`s become caller-controlled (the allowlist removal is a separate follow-up).

**Why this shape.** II is on a system subnet — HTTP outcalls cost no cycles — and ingress flooding is handled by the boundary nodes. So these caches don't need outcall-rate-limiting or cycle accounting; their only DDoS-relevant job is **bounded, fairly-evicted state**.

**Single shared budget.** Discovery and JWKS are one coupled flow (`domain → discovery → jwks_uri → JWKS`); a verification needs both the domain's discovery entry and its JWKS entry warm, so they share one `SSO_CACHE_MAX_ENTRIES = 5000` rather than separate caps (separate ones would just leave one cache holding entries the other can't back). Bounded by the larger JWKS entry (≤ 32 KiB): 5000 × 32 KiB ≈ 160 MB + ~10 MB discovery ≈ **~170 MB worst case**, ~5–6% of the ~3 GB Wasm heap — wide eviction headroom so a junk-domain flood can't evict the providers real users rely on.

**Per-fill / per-entry caps** (fills were previously unbounded, `max_response_bytes: None`):
- Response size: discovery hops 16 KiB; JWKS fetch **32 KiB** — broad because real key sets embed `x5c` chains (Microsoft's JWKS is ~14.5 KB with 8 keys) and this fetch is shared with the configured Google/Microsoft refresh, which must not be rejected.
- Content: ≤ 20 JWKs (verification matches by `kid`; providers publish far fewer) and ≤ 32 stored discovery `scopes` (the only unbounded `DiscoveredConfig` field).

No behavior change for honest providers.

src/internet_identity/src/main.rs
```diff
@@ -1450,20 +1478,50 @@ mod openid_api {
...
+            Err(err) => OpenIdResult::Err(err),
         }
     }
+
+    /// Drive the two-hop SSO discovery fetch for `domain`. The frontend calls
+    /// this when `get_sso_discovery` reads `Pending`, then keeps polling the
+    /// query until it returns `Resolved`.
+    #[update]
+    fn discover_sso(domain: String) {
+        openid::discover_sso(&domain)
```
MRmarioruci: Nit. Here the implementation is a bit different from the JWT endppints which call openid::canonical_discovery_domain, which would be fine cause there is a to_lowercase conversion downstream, but there is no trimming so an sso with " domain.com" would be rejected.

src/internet_identity/src/main.rs
```diff
@@ -1450,20 +1478,50 @@ mod openid_api {
...
+            Ok(signed) => OpenIdResult::Ok(signed),
+            Err(err) => OpenIdResult::Err(err),
         }
     }
+
+    /// Drive the two-hop SSO discovery fetch for `domain`. The frontend calls
+    /// this when `get_sso_discovery` reads `Pending`, then keeps polling the
+    /// query until it returns `Resolved`.
+    #[update]
+    fn discover_sso(domain: String) {
```
MRmarioruci, line 1513: What do we do about rate limiting endpoints? Do we have to do this explicitly or not relevant in our case? Since this is unauthenticated. The single-flight cache deduplicates concurrent inflight calls and the backoff bounds the rate per-domain, but I guess a caller cycling through cleared cache entries (e.g. after eviction) can do repeated outcalls

src/frontend/src/lib/utils/openidPoll.ts
```diff
@@ -0,0 +1,32 @@
...
+ * An update call reports `Pending` (and `discover_sso` returns no value) while
+ * a cold cache fills; the frontend retries until it warms. A configured
+ * provider (Google / Microsoft / Apple) is always warm, so callers resolve on
+ * the first attempt.
+ */
+export const POLL_INTERVAL_MS = 500;
+export const MAX_POLL_ATTEMPTS = 60;
+
+export const pollDelay = (): Promise<void> =>
+  new Promise((resolve) => setTimeout(resolve, POLL_INTERVAL_MS));
```
MRmarioruci: Nit. discoverSsoConfig (ssoDiscovery.ts:127) checks signal?.aborted at the start of each iteration. If the user aborts during the 500ms sleep, one more canister call still fires before the loop notices. We could pass the abortSignal and respect it in here as well.

src/internet_identity/src/main.rs
sea-snake: Good catch — fixed in 321704a. `discover_sso` / `get_sso_discovery` now route through `openid::canonical_discovery_domain` (trim + lowercase) at the canister boundary, same as the JWT endpoints, so ` dfinity.org` is accepted everywhere instead of only on the JWT path. The canonicalizer is split into a `&str -> String` core plus a `canonical_discovery_domain_opt` wrapper for the optional JWT args; the unit test now covers both.

src/frontend/src/lib/utils/openidPoll.ts
sea-snake: Done in 9ae5d49. `pollDelay` now takes the signal and resolves early on abort, and `discoverSsoConfig` re-checks `signal.aborted` right before the `discover_sso` update — so an abort during the query or the 500ms sleep stops the loop without firing another canister call. Added a fake-timer test asserting no second update fires after a mid-sleep abort.

src/internet_identity/src/openid/jwks.rs
sea-snake: Shipped earlier in 5308254 / 6701a3a: JWKS fetch capped at 32 KiB (`JWKS_MAX_RESPONSE_BYTES`, broad enough for real x5c key sets — MS JWKS is ~14.5 KB), discovery hops at 16 KiB, plus content caps (<=20 keys, <=32 stored scopes) and bounded single-flight caches (5000 entries, fair eviction). So a compromised allowlisted provider can no longer return a multi-MB body per node per retry.

src/internet_identity/src/main.rs
sea-snake: No explicit per-endpoint rate limiting needed in our case. II runs on a system subnet, so HTTP outcalls cost no cycles, and ingress flooding is absorbed by the boundary nodes — the only DoS-relevant job for these unauthenticated endpoints is keeping canister state bounded, which the single-flight caches now do: bounded entries with fair eviction (5000) plus per-fill byte/content caps. A caller cycling evicted entries can trigger repeated outcalls, but each is free, deduplicated while in flight, and per-domain backoff bounds the rate; it cannot grow state or evict honest providers beyond the cap. (Full reasoning in my cache-bounding PR comment.)

src/canister_tests/src/api/internet_identity/api_v2.rs
sea-snake: `settled()` collapses the new 3-arm `OpenIdResult<Ok | Pending | Err>` back into the classic `Result<Ok, Err>` so the existing configured-provider test call sites keep their original signature instead of every one having to match `Pending`. `Pending` only ever comes from a cold/evicted SSO discovery-or-JWKS cache; configured providers (Google/Microsoft/Apple, which these helpers use) are always warm, so it is unreachable here and the helper panics on it to surface a logic error loudly. It is test-helper-only code. Happy to inline it or expose `OpenIdResult` directly to callers instead if you prefer — let me know.

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -1,167 +1,74 @@
...
- * True if `hostname` is exactly `expected` or a proper subdomain of it.
- * Using `endsWith(expected)` alone would incorrectly accept
- * `evilaccounts.google.com` when `expected` is `accounts.google.com`.
+ * `localhost` / `127.0.0.1`, optionally followed by `:<port>`. IPv6 loopback
+ * (`[::1]`, etc.) is intentionally not handled — the canister doesn't recognise
+ * it either, and the e2e setup uses the hostname form.
  */
-const hostnameMatchesAllowed = (hostname: string, expected: string): boolean =>
-  hostname === expected || hostname.endsWith(`.${expected}`);
+const isLoopbackHost = (host: string): boolean => {
```
aterga, line 59: This low-level parsing could be avoided in favor of a new URL

src/internet_identity/src/openid/verify.rs
```diff
@@ -0,0 +1,449 @@
...
+            },
+        };
+        let credential = verify_and_build(VALID_JWT, &descriptor, &test_certs(), &test_salt())
+            .expect("expected verification to succeed");
+        assert_eq!(credential.sso_domain, Some("example.org".to_string()));
+        assert_eq!(credential.sso_name, Some("Example".to_string()));
+    }
+
+    #[test]
+    fn should_reject_certificate_not_found() {
```
aterga, line 476: Do we already have tests for other negative cases?

For example, expired certificate / invalid certificate / expired JWT / invalidated JWT / etc.

src/internet_identity/src/openid/verify.rs
```diff
@@ -0,0 +1,449 @@
...
+    if claims.nonce != expected_nonce {
+        return Err(OpenIDJWTVerificationError::GenericError(format!(
+            "Invalid nonce: {}",
+            claims.nonce
+        )));
+    }
+    if now > secs_to_nanos(claims.exp) {
+        return Err(OpenIDJWTVerificationError::JWTExpired);
+    }
+    if now > secs_to_nanos(claims.iat + MAX_VALIDITY_WINDOW_SECONDS) {
```
aterga: Please always use panic-free code in canister code.


```suggestion
    let Some(max_expected_value) = claims.iat.checked_add(MAX_VALIDITY_WINDOW_SECONDS) else {
        return Err(OpenIDJWTVerificationError::GenericError { ... });
    };
    
    if now > max_expected_value {
        return Err(OpenIDJWTVerificationError::JWTExpired);
    }
```

src/internet_identity/src/openid/verify.rs
```diff
@@ -0,0 +1,449 @@
...
+
+    // Compute the effective issuer. For a placeholder issuer (Microsoft's
+    // `https://login.microsoftonline.com/{tid}/v2.0`) the `{tid}` is filled
+    // from the JWT claims; a concrete issuer has no placeholders and passes
+    // through unchanged, so SSO and Google/Apple share this code path.
+    let issuer_placeholders = get_issuer_placeholders(&descriptor.issuer);
+    let issuer_claims = get_all_claims(validation_item.claims(), issuer_placeholders);
+    let effective_issuer = replace_issuer_placeholders(&descriptor.issuer, &issuer_claims);
+
+    verify_claims(&effective_issuer, &descriptor.client_id, &claims, salt)?;
```
aterga, line 143: Nit. Deconstruct `claims`:


```suggestion
    // Audience must not be reused since only weak verification was done for it so far
    let Claims { email, email_verified, issuer_claims, email, iss, sub, aud: _aud } = claims;
```

src/internet_identity/src/openid/verify.rs
aterga: Added negative-path unit tests in `openid::verify` covering: invalid signature (tampered JWT), wrong issuer, wrong audience, wrong nonce, expired JWT (`now > exp`), JWT past our validity window (`iat + MAX_VALIDITY_WINDOW_SECONDS`), JWT issued in the future (`now < iat`), and a far-future `iat` overflow guard.

On the certificate side: invalid / rotated-away certificates are already covered by `should_reject_certificate_not_found` (a non-empty key set that lacks the JWT's `kid`). There's no separate cert-expiry concept in this pipeline — keys are matched by `kid`, so a rotated/expired key simply yields "certificate not found"; key freshness is handled by the JWKS refresh timer (configured) / on-demand cache (SSO), not a per-cert expiry check.

Each test resets the test clock and caller at its start so parallel thread reuse can't leak mutated state between tests.

---

aterga:
Addressed the open review comments and got the branch compiling again.

**Build fixes** — the branch did not compile at `e440e0c`:
- `openid.rs` imported `std::ffi::os_str::Display` (a struct) instead of the `std::fmt::Display` trait, so `impl Display for AudClaim` failed with `E0404`.
- The SSO unit test `resolve_cross_checks_issuer` still called `resolve()` with the old 2-arg signature, with the new `&AudClaim` mis-placed as an `assert!` format argument.

**Review comments:**
- `verify.rs:143` — destructured the verified `Claims`, explicitly dropping `aud` (only weakly checked via `AudClaim::matches`; the canonical `client_id` is stored instead) and the consumed `nonce`/`exp`/`iat`.
- `verify.rs:292` — the JWT time checks are now panic-free: validity is compared in whole seconds and `iat + window` uses `checked_add`, so a malicious far-future `iat`/`exp` can no longer overflow `secs_to_nanos`.
- `verify.rs:429` — added the negative-path tests (details in the inline reply).
- `ssoDiscovery.ts:59` — the loopback host is now parsed with `new URL` instead of hand-splitting on `:` (and a host carrying a path/query/fragment is rejected).

Locally verified: `cargo fmt --check`, `cargo clippy -D warnings`, the `openid` unit tests (23/23), and the `ssoDiscovery` vitest suite (15/15) all pass. The remaining canister-tests / e2e jobs are running in CI.

---

aterga (approved):

aterga (approved):

PR #4046, opened 2026-06-19
feat(be,fe): verified emails as a new first-class anchor primitive

MRmarioruci:
Internet Identity gains a new anchor primitive: a verified-emails list, separate from the recovery email. Users can verify one or more addresses (capped at 5) on the manage page, and relying parties that request an email attribute can now reach the user through one of those verified addresses — without it doubling as a recovery credential. When a dapp requests an email and the anchor has none, the /authorize consent view offers an inline "Verify an email" affordance so the user doesn't have to bounce out to /manage and back.

The verification path reuses every piece of the recovery flow's SMTP/DKIM/DMARC/DNSSEC machinery; the disambiguator is a new canister-issued Subject prefix (`II-Verify-` vs `II-Recovery-`) that `PendingKind` uses to route the inbound to the right anchor field.

## Changes

**Backend**
- New `StorableVerifiedEmail` (cbor-map, additive forward-compat) and new `Anchor.verified_emails` field, capped at 5 entries.
- New `PendingKind::VerifyEmail { anchor }` dispatch, gated on the `II-Verify-` Subject prefix.
- New canister methods `verified_email_prepare_add` (authenticated) and `verified_email_remove` (authenticated). The status / diagnostics / submit-dkim-leaf / resolve-via-doh queries are shared with the recovery flow — keyed by nonce, flow-neutral.
- Integration tests cover cap enforcement, dedup, removal, and the verify-vs-recovery dispatch.

**Frontend**
- `VerifiedEmailWizard` parallels the recovery setup wizard — same DNSSEC skeleton + DKIM + DoH path, differing only in the canister call and the user-visible nonce prefix.
- `VerifiedEmailsPanel` renders on /manage/recovery alongside the existing recovery card: counter, list with relative-time tooltip, "Verify an email" CTA, per-row remove via confirmation dialog.
- /authorize attribute consent view now offers an inline "Verify an email" affordance when an email attribute is requested but the anchor has none — after verification the view re-resolves groups against the live attribute set; the user can also skip and continue without sharing.

## Tests

- Backend: `tests/integration/verified_emails.rs` (cap, dedup, remove, dispatch).
- Frontend: vitest coverage for the empty-state branch + attribute-resolution helper; Playwright e2e in `tests/e2e-playwright/routes/verifiedEmails.spec.ts` covers the wizard surface and a full DNSSEC + DKIM verify-then-remove flow.
- Clippy clean (1.91.1, CI-matching flags); `npm run check` + `lint` + `format` clean.

<div align="right">Next: #4056</div>


src/internet_identity/src/verified_emails/prepare.rs
```diff
@@ -0,0 +1,58 @@
...
+) -> Result<EmailChallenge, EmailChallengeError> {
+    // Per-anchor cap. Checked before issuing the nonce so the FE
+    // wizard can show a "limit reached" notice without the user
+    // having to send a magic email first. Compared inclusively
+    // because we're about to add one more.
+    let current = state::anchor(anchor).verified_emails.len();
+    if current >= MAX_VERIFIED_EMAILS_PER_ANCHOR {
+        return Err(EmailChallengeError::InternalCanisterError(format!(
+            "anchor already holds {current} verified emails — limit is {MAX_VERIFIED_EMAILS_PER_ANCHOR}",
+        )));
```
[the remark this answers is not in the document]
MRmarioruci: Done in f347fec5 — new `EmailChallengeError::LimitReached { limit: nat8 }` variant, used at both prepare time and the commit-time race recheck. The wizard catches it and shows "You've reached the limit of 5 verified emails. Remove one to add another." inline on the address-entry view.

src/internet_identity/src/email_inbound/prepare.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    // Sanity check the address shape. Detailed RFC 5321/5322 validation
+    // is the verifier's job; here we just want to fail fast on the
+    // obviously-malformed inputs (no `@`, leading/trailing whitespace,
+    // empty parts) so the FE doesn't get back an opaque rejection
+    // half a flow later.
+    let address = normalize_address(&address)
+        .ok_or_else(|| EmailChallengeError::DomainNotSupported(address.clone()))?;
+    let registered_domain = registered_domain_of(&address)
+        .ok_or_else(|| EmailChallengeError::DomainNotSupported(address.clone()))?;
```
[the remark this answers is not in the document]
MRmarioruci: Done in f347fec5 — added `EmailChallengeError::InvalidEmailAddress(text)`. `normalize_address` failures route to it (wizard surfaces an inline form error); `registered_domain_of` failures stay on `DomainNotSupported` and still route to the dedicated unsupported-domain view.

src/frontend/src/lib/components/settings/VerifiedEmailsPanel.svelte
```diff
@@ -0,0 +1,262 @@
...
+      handleError(new Error(JSON.stringify(result.Err)));
+      return;
+    }
+    removingAddress = undefined;
+    void invalidateAll();
+    toaster.success({
+      title: $t`Verified email removed`,
+      description: $t`${address} has been removed from your verified emails.`,
+    });
+  };
```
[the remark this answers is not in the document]
MRmarioruci: Done in f347fec5 — `handleRemove` now uses `.then(throwCanisterError)` with try/catch like the rest of the file.

src/internet_identity_interface/src/internet_identity/types/email_challenge.rs
aterga, line 1: I assume this file's contents are not new, right?

src/internet_identity/src/anchor_management.rs
```diff
@@ -218,8 +218,16 @@ pub fn add_openid_credential_skip_checks(
...
     openid_credential: OpenIdCredential,
 ) -> Result<Operation, AnchorError> {
+    let mirrored_address = openid_credential.get_verified_email();
     anchor.add_openid_credential(openid_credential.clone())?;
 
+    // Silently mirror an IdP-vouched email into the anchor's verified-emails
+    // list so the address survives an OIDC unlink. The user can still remove
+    // the entry from the panel; re-adding then goes through the DKIM flow.
+    if let Some(address) = mirrored_address {
+        anchor.mirror_verified_email_from_oidc(&address, time());
```
aterga: I'm not sure why we need to duplicate data inside the anchor. 

Couldn't we fetch it from `anchor.openid_credentials` when needed?

src/internet_identity/src/verified_emails/prepare.rs
```diff
@@ -0,0 +1,58 @@
...
+    now_secs: u64,
+) -> Result<EmailChallenge, EmailChallengeError> {
+    // Per-anchor cap. Checked before issuing the nonce so the FE
+    // wizard can show a "limit reached" notice without the user
+    // having to send a magic email first. Compared inclusively
+    // because we're about to add one more.
+    let current = state::anchor(anchor).verified_emails.len();
+    if current >= MAX_VERIFIED_EMAILS_PER_ANCHOR {
+        return Err(EmailChallengeError::LimitReached {
+            limit: MAX_VERIFIED_EMAILS_PER_ANCHOR as u8,
```
aterga: Avoid casting numerical types in smart contracts (as it could in theory panic). 

Use the same type for the field and the constant.

src/internet_identity/src/anchor_management.rs
MRmarioruci: Sure but then we are coupled to the openid credentials. which means if i remove it as an access method, I don't have it as a verified email anymore.

src/internet_identity/src/verified_emails/prepare.rs
```diff
@@ -0,0 +1,33 @@
...
+    // Cap-check at prepare so the FE can show a "limit reached"
+    // notice without the user sending the magic email first.
+    let current = state::anchor(anchor).verified_emails.len();
+    if current >= MAX_VERIFIED_EMAILS_PER_ANCHOR {
+        return Err(EmailChallengeError::LimitReached {
+            limit: MAX_VERIFIED_EMAILS_PER_ANCHOR as u8,
+        });
+    }
+
+    prepare_common(
```
aterga, line 26: Let's make function names verbose!

src/internet_identity/src/storage/anchor.rs
```diff
@@ -845,6 +874,31 @@ impl Anchor {
         Ok(())
     }
 
+    /// Returns `true` when appended, `false` if already present or cap
+    /// reached. Never errors — the surrounding OIDC link must not fail
+    /// just because the verified-emails bucket is full.
+    pub fn mirror_verified_email_from_oidc(
```
aterga, line 880: I would drop this function. 

Just aggregate verified and OpenID emails in the FE, perhaps?

src/internet_identity/src/email_recovery/mod.rs
```diff
@@ -29,53 +32,14 @@
...
-//!
-//! The pending-challenge map has a 30-minute TTL and is a strict
-//! optimisation — losing it on canister upgrade just costs every
-//! in-flight registration its retry. We do **not** want stable-memory
-//! churn for entries that expire in half an hour. The heap-side
-//! `HashMap` is naturally bounded by the policy in §8.8 of the design
-//! doc (size cap + oldest-first eviction).
 
-#![allow(dead_code)] // Exposed as the module fills in.
+#![allow(dead_code)] // Some recovery-side helpers are reserved for follow-up flows.
```
aterga: Could you clarify what you mean with this comment, please?

src/internet_identity/src/email_inbound/prepare.rs
```diff
@@ -0,0 +1,298 @@
...
+};
+
+/// Shared input-validation + nonce-issuing core. `kind`
+/// parametrises over which flow we're starting. The challenge no
+/// longer carries a `mailbox` field — the FE pairs the user-part
+/// (`register` / `recover`) with `window.location.hostname` to
+/// render the user-facing label, and the canister accepts mail at
+/// any of the configured `related_origins` aliases (see
+/// [`super::mailbox_domains`]).
+pub(crate) async fn prepare_common(
```
aterga, line 19: Let's double check that this code matches the exact code from before module names, etc.

src/internet_identity/src/email_inbound/smtp.rs
```diff
@@ -738,13 +762,58 @@ pub(super) fn bind_credential(
...
+    // Refresh in place if the address is already verified on this
+    // anchor; otherwise append a fresh entry.
+    if let Some(existing) = a
+        .verified_emails
+        .iter_mut()
+        .find(|e| e.address.eq_ignore_ascii_case(claimed_address))
+    {
+        existing.verified_at = now_ns;
+    } else {
+        if a.verified_emails.len() >= super::MAX_VERIFIED_EMAILS_PER_ANCHOR {
```
aterga: This is an odd place for checking this condition, maybe it should be at the top of the function?

src/internet_identity/src/email_inbound/smtp.rs
```diff
@@ -738,13 +762,58 @@ pub(super) fn bind_credential(
...
+/// onto the anchor's `verified_emails` list.
+///
+/// Unlike `bind_credential` this does *not* touch the recovery
+/// reverse-address index — verified emails aren't queryable by
+/// address, only via the owning anchor. Duplicates (same lowercased
+/// address already present on this anchor) are accepted as no-ops:
+/// the inbound email already proved control, and refreshing
+/// `verified_at` is also a legitimate side effect of running the
+/// wizard again.
+pub(super) fn bind_verified_email(
```
aterga: Function name could be improved

src/internet_identity/src/email_inbound/smtp.rs
```diff
@@ -738,13 +762,58 @@ pub(super) fn bind_credential(
...
+    anchor: AnchorNumber,
```
aterga, line 784:
```suggestion
    anchor: &mut Anchor,
```

src/internet_identity/src/email_inbound/smtp.rs
aterga:
```suggestion
pub(super) fn append_or_refresh_verified_email(
```

aterga (approved):

src/internet_identity_interface/src/internet_identity/types/email_challenge.rs
MRmarioruci: Right — extracted from `email_recovery.rs` (which has been around for a few releases). Renamed to `email_challenge.rs` once the shared inbound primitive was lifted out, but the type definitions are unchanged in shape (only renamed alongside `EmailRecovery* → EmailChallenge*`). Wire-compatible with the old FE.

src/internet_identity/src/anchor_management.rs
MRmarioruci: Keeping the mirror, intentionally. The point is that a verified address survives an OIDC unlink: a user picks a 'reach' email once, the dapp learns to request it via `verified:<H>:email`, and if the user later unlinks the IdP the address stays on the anchor instead of silently disappearing. FE-side aggregation can't model that because `openid_credentials` no longer carries the address by then. The mirror's properties (idempotent, cap-respecting, never errors) are tested at the unit level in `storage/anchor/tests.rs` and now end-to-end in `anchor_management::oidc_link_at_verified_emails_cap_skips_mirror_but_succeeds`.

src/internet_identity/src/verified_emails/prepare.rs
MRmarioruci: Done in 86339630a — `MAX_VERIFIED_EMAILS_PER_ANCHOR` is now `u8` (matches the `LimitReached { limit: u8 }` field). The two comparison sites use `usize::from(MAX_VERIFIED_EMAILS_PER_ANCHOR)` which is infallible — no `as` cast anywhere.

src/internet_identity/src/verified_emails/prepare.rs
MRmarioruci: Done in 86339630a — `prepare_add` → `prepare_add_verified_email`; export and the single call site in main.rs updated.

src/internet_identity/src/storage/anchor.rs
MRmarioruci: Keeping — see the longer reply on the `add_openid_credential_skip_checks` mirror call: aggregating in the FE loses the address as soon as the OIDC link goes away, which defeats the design's goal of stable attribute resolution across unlinks.

src/internet_identity/src/email_recovery/mod.rs
MRmarioruci: Done — the `#![allow(dead_code)]` was masking exactly one unused helper (`prepare_add_now`). Deleted both the helper and the module-wide allow in 422b8c5b6.

src/internet_identity/src/email_inbound/prepare.rs
MRmarioruci: Diffed and confirmed in 86339630a. Substantive changes vs. the pre-refactor body: (1) type renames `EmailRecovery* → EmailChallenge*` (structurally identical), (2) a new `nonce_prefix: &str` parameter so the same helper can issue both `II-Recovery-` and `II-Verify-` nonces, (3) malformed addresses now map to the new `InvalidEmailAddress` variant instead of overloading `DomainNotSupported` (separate Copilot review). No other logic drift.

src/internet_identity/src/email_inbound/smtp.rs
MRmarioruci: Leaving the placement as-is — the function intentionally early-returns when the address is already verified (refresh-in-place path), and the cap is only relevant on the append branch. Hoisting the check to the top would reject a refresh of an already-verified address when at cap, which is a regression from the design.

src/internet_identity/src/email_inbound/smtp.rs
MRmarioruci: Done in 86339630a — `bind_verified_email` → `append_or_refresh_verified_email` (matches your other suggestion on 783).

src/internet_identity/src/email_inbound/smtp.rs
MRmarioruci: Adopted in spirit in 86339630a: kept the param as `anchor: AnchorNumber` for symmetry with `bind_credential` (which also loads from state internally), but renamed the inner local `mut a` → `mut anchor` so the body reads cleanly. Changing the param to `&mut Anchor` would push the load/write responsibility to the caller — fine for one of the two functions but asymmetric with the recovery side, so I'd rather keep them parallel.

src/internet_identity/src/email_inbound/smtp.rs
MRmarioruci: Adopted in 86339630a.

src/internet_identity/src/main.rs
```diff
@@ -1815,20 +1883,64 @@ mod email_recovery_api {
...
+    #[update]
+    async fn verified_email_prepare_add(
+        identity_number: IdentityNumber,
+        dns_input: EmailChallengeDnsInput,
+    ) -> Result<EmailChallenge, EmailChallengeError> {
+        check_authorization(identity_number)
+            .map_err(|err| EmailChallengeError::Unauthorized(err.principal))?;
+
+        let now_secs = ic_cdk::api::time() / 1_000_000_000;
+        crate::verified_emails::prepare_add_verified_email(identity_number, dns_input, now_secs)
```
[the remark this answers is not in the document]
MRmarioruci: not valid

PR #4034, opened 2026-06-18
feat(be): backend /mcp account-delegation methods (MCP server acts as anchor)

sea-snake:
Canister side of the **backend-direct** MCP delegation model: a single, deploy-configured MCP server fetches per-app account delegations for an anchor by calling II directly — no per-app browser flow. Pairs with the MCP server PR (sea-snake/imcp2#10), which calls these methods.

## Model
At connect, the MCP server gets (via the existing `/mcp` browser delegation flow) a 60-min standing delegation `anchor → MCP-server key` for `mcp_server_origin`. Its principal is the principal II derives for that anchor's default account at `mcp_server_origin`. When the anchor opts in, II records `that principal → anchor`. The MCP server then calls the new methods *as that principal*; II recovers the anchor from `caller()` — **no `anchor_number` parameter**, being the right caller is the authorization.

## Changes (backend)
- **`mcp_server_origin` deploy arg** — `InternetIdentityInit` + `PersistentState`, persisted via `StorablePersistentState` exactly like `backend_origin`. Unset ⇒ path disabled.
- **Reverse index** `lookup_anchor_with_mcp_principal` — `StableBTreeMap<Principal, AnchorNumber>` at `MemoryId(25)` (mirrors the passkey/recovery-phrase principal indexes).
- **`mcp_set_access(anchor_number, enabled)`** / **`mcp_access_enabled(anchor_number)`** — opt-in toggle, authorized as the anchor; binds/unbinds the anchor's `mcp_server_origin` principal in the index.
- **`mcp_prepare_account_delegation(target_origin, session_key, max_ttl)`** + **`mcp_get_account_delegation(target_origin, session_key, expiration)`** — anchor recovered from `caller()` via the index; delegate the anchor's **default account** at `target_origin`, **capped at 5 minutes**, reusing the existing account-delegation machinery.
- Candid updated for all of the above.

`cargo build -p internet_identity` is clean.

## Not in this PR (draft) — remaining work
- **Frontend `/mcp` connect screen**: repurpose the existing `/mcp` route from the per-app "Allow MCP access" screen into the one-time **connect** screen ("allow MCP for this session"), which issues the 60-min standing delegation to the MCP server key **and** calls `mcp_set_access(anchor, true)`. Regenerate FE candid bindings for the new methods.
- E2E coverage; a Rust test for the new methods (auth via the index, 5-min cap).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4052, opened 2026-06-22
feat(mcp): per-user trusted MCP server, synced across devices, with account-bound delegations

aterga:
## What this is

Internet Identity's `/mcp` flow lets an MCP server obtain a standing delegation so it can act on the user's behalf across apps. This PR reworks it from a **single, deploy-configured MCP server** to **per-user MCP server trust**: each user enables MCP and designates the one (remote, https) MCP server they trust, in Settings; that choice **syncs across all their devices**; and connecting binds it to a specific account.

Builds on #4034 (backend `/mcp` account-delegation methods), redesigned per review.

## Design

**Per-user, Settings-managed, synced trust.** Instead of a global `mcp_server_origin` deploy arg (removed from the frontend-canister init, deploy scripts, and CI), each identity, under **Settings → Trusted MCP server**, flips a master toggle and sets the one MCP server URL it trusts. Both are stored **on-chain, per anchor**, so the choice **syncs across all the identity's devices**, and only the authenticated user can change it (never a page that initiates a connect request). (The CLI-access toggle stays intentionally device-local for contrast: its badge reads "Enabled on this device", the MCP one "Enabled on all your devices".)

**Storage: a dedicated, persisted per-anchor map.** The config (`{ enabled, url }`) lives in its own `StableBTreeMap` keyed by anchor number (memory index 26), kept **separate from the core anchor record so it never touches anchor serialization**. It is written via a new authenticated `mcp_set_config` (full authorization — only the user can change what their identity trusts) and read via `mcp_get_config` (session-scoped read; an unauthorized caller gets the safe disabled/no-server default). Candid gains an `McpConfig` record and these two methods; FE bindings regenerated. (An earlier revision tried to reuse `identity_metadata` for this — that field is **not persisted to stable memory** on anchor write, so it never actually synced; hence the dedicated store.)

**Verify at connect (canister is the source of truth).** Because the synced config is only readable once authenticated, the `/mcp` connect screen opens optimistically; right after the user authenticates ("Allow access"), II reads the identity's synced config via `mcp_get_config` and connects only when MCP is enabled **and** the request's callback origin matches the trusted server's origin — otherwise it shows an "isn't trusted yet" screen pointing to Settings. The decision is identical on every device, with no device-local state involved.

**Device master toggle.** The trusted-server URL box appears only once the toggle is on; turning it off disables MCP completely (the connect flow rejects every request, independent of any stored URL). An "Enabled on all your devices" badge shows next to the title once a server is set.

**Remote (https) only.** MCP connections target remote servers, so the trusted URL and the connect callback must be https — loopback/plain-http is rejected in the Settings input, in the connect-flow check, and by the `/mcp` `form-action` CSP.

**Real-MCP verification.** When a server is set (and via a re-check button / on reopening Settings), the UI verifies it through the OAuth **Protected Resource Metadata** the MCP authorization spec requires (RFC 9728): a CORS-friendly GET to `/.well-known/oauth-protected-resource` (bare and resource-path-suffixed forms), requiring the doc's `resource` to name the exact URL entered. This works even though the MCP endpoint itself answers `401` and blocks cross-origin reads. The MCP `initialize` JSON-RPC handshake is kept as a fallback for CORS-permissive / unprotected servers. The probe is advisory, so a server it can't confirm is still activated but flagged with a connection-status indicator and a re-check button — never a hard block.

**Account-bound delegations.** The connect screen reuses the regular continue-with-ii account picker (`ContinueView`) so the user binds a **specific** account, and lets the user choose the session duration. `mcp_set_access(origin, account, true)` and the standing delegation derive the **same** principal, and the backend's disable re-derives it to revoke — fixing the revocation bug where the binding followed the *mutable* default account and orphaned the old principal when the default changed (per the copilot note + the account-picker review thread). On enable, a cross-anchor principal-bind collision now surfaces as an `Err` rather than silently no-op'ing (per @MRmarioruci's review). The MCP server-side methods are kept consistent too: `mcp_prepare_account_delegation` returns the resolved `account_number` (new `McpPrepareDelegation` type) which `mcp_get_account_delegation` then takes, so `get` reads the exact account `prepare` signed for instead of re-resolving the mutable default (per @MRmarioruci's review). A loading screen covers the gap between "Allow access" and the redirect to the server.

**Session handoff to manage tabs.** Opening management from the `/mcp` flow — the untrusted screen's "Manage trusted server" button, or the header's "Manage your Internet Identity" — authenticates the selected identity in place and hands the session to the opened tab via the same postMessage handoff `/authorize` uses, so the user doesn't have to sign in again to set their trusted server. (The passkey/IdP prompt consumes the click's transient activation on Safari/strict Firefox, so when a fresh sign-in was needed a confirmation step supplies a new gesture for `window.open`.) The ceremony lives in a shared `ManageHandoffFlow` + `ManageHandoff` overlay, which `/authorize` now also uses (deduplicated from its previous inline copy, per @MRmarioruci's review).

## Security note — CSP `form-action`

The delegation is delivered to the chosen server via a top-level form-POST, which `form-action` governs.

- The relaxation is **scoped to the `/mcp` landing asset only** (`is_mcp_landing_path` → `/mcp`, `/mcp/`, `/mcp/index.html`), where it is `'self' https:`. The SPA-wide policy stays `'self' http://127.0.0.1:*` (the loopback is for the `/cli` flow), so `/authorize`, `/cli`, and every other route keep the tight policy — an HTML injection elsewhere still can't form-POST data cross-origin.
- Delivering an origin-scoped delegation to the chosen origin is consistent with II's existing `/authorize` model: II derives a **different** principal per origin, so a delegation delivered to origin X only grants the user's identity-at-X — exactly what a normal `/authorize` sign-in to X would yield. The standing credential is an ephemeral-key sub-delegation with a user-chosen TTL (capped at 30 days); the canister only ever signs to a freshly-generated, non-extractable browser key, never to the public key from the URL.

## Known follow-up

- **Revoke on remove.** Removing/disabling the trusted server in Settings is a pre-gate for **new** connects; it does **not** yet revoke the backend `mcp_set_access` binding, so an already-connected server keeps access until its standing delegation expires (the user-chosen TTL, ≤ 30 days). Tracked in #4064 (revoke-on-remove + a possible "connected servers" management surface).

## Verification
- FE: `npm run check` (svelte-check + tsc) — 0 errors; eslint + prettier clean.
- Backend: `cargo check` / `clippy` (wasm) clean; the candid `service_equal` compatibility test passes against the regenerated `.did`. Integration tests cover the config **round-trip**, **persistence across a canister upgrade** (proving it's in stable memory), **identity gating** (a different caller can neither overwrite nor read the config), and the **prepare/get account threading** (changing the default account between prepare and get; the threaded account still resolves while the changed default returns `NoSuchDelegation`).
- e2e `/mcp` specs updated for the synced model: trust is seeded through the Settings UI (real canister writes) and verified at connect; the connect screen shows optimistically and the "isn't trusted yet" screen now appears after authenticating; non-https rejected; MCP-verified add via protected-resource metadata. The connect tests are marked `test.slow()` because they now make real canister round-trips (Settings write + verify-at-connect + delegation) rather than reading instant device-local state.

## History — rebase of #4034
This branch began as **#4034 (`sea-snake:feat/mcp-backend-delegation`) brought up to date with `main`**, preserving the original commits/authorship; the one conflict (the generated candid binding `internet_identity_types.d.ts`) was resolved by regenerating it from the merged `.did`. #4034's head lives on an external fork this environment can't push to, so the merged + reworked result is published here. See #4034 for the original backend description.

https://claude.ai/code/session_01CGvgPmEVcgyUP2mAnwnDcj


src/internet_identity/src/mcp.rs
```diff
@@ -0,0 +1,148 @@
...
+const MCP_MAX_EXPIRATION_PERIOD_NS: u64 = 5 * 60 * 1_000_000_000;
+
+/// The configured MCP server origin, or an error if the path is disabled.
+fn mcp_server_origin() -> Result<FrontendHostname, String> {
+    state::persistent_state(|ps| ps.mcp_server_origin.clone())
+        .ok_or_else(|| "MCP delegation is not enabled on this canister".to_string())
+}
+
+/// The anchor's default account at `origin` (synthetic when none is reserved).
+fn default_account(anchor_number: AnchorNumber, origin: &FrontendHostname) -> Account {
```
aterga, line 36: Why invent a new type of flow, let's just use the same account picker as we have for the regular continue-with-ii flow.

src/internet_identity/src/storage.rs
```diff
@@ -1039,6 +1062,24 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// principal (the caller of the `mcp_*_account_delegation` methods).
+    pub fn lookup_anchor_with_mcp_principal(&self, principal: Principal) -> Option<AnchorNumber> {
+        self.lookup_anchor_with_mcp_principal_memory.get(&principal)
+    }
+
+    /// Record (enable) MCP access: bind an anchor's MCP-server principal to it.
+    pub fn set_anchor_mcp_principal(&mut self, principal: Principal, anchor_number: AnchorNumber) {
+        self.lookup_anchor_with_mcp_principal_memory
+            .insert(principal, anchor_number);
+    }
```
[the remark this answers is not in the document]
aterga: Fixed in 6ecd598 — `set_anchor_mcp_principal` now mirrors the passkey / recovery-phrase guard: it refuses to overwrite a principal already bound to a *different* anchor (warn-and-skip), and is idempotent when re-binding the same anchor.

---

src/internet_identity/src/mcp.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Whether `anchor_number` currently has MCP access enabled.
+pub fn is_mcp_access_enabled(anchor_number: AnchorNumber) -> bool {
+    let Ok(origin) = mcp_server_origin() else {
+        return false;
+    };
+    let principal = mcp_principal_for(anchor_number, &origin);
+    storage_borrow(|storage| storage.lookup_anchor_with_mcp_principal(principal))
+        == Some(anchor_number)
+}
```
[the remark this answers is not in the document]
aterga: Valid — and it's the same root cause as @aterga's note about reusing the account picker: the binding follows the *mutable* default account, so changing the default orphans the old principal. The fix is to bind to a specific, user-chosen account (stable across default changes), which makes enable/disable derive the same principal and revoke correctly. Tracking it together with that account-binding change rather than patching the symptom here.

---

src/internet_identity/internet_identity.did
```diff
@@ -324,6 +324,9 @@ type InternetIdentityInit = record {
     backend_canister_id : opt principal;
     // Backend origin, needed to sync configuration with frontend.
     backend_origin : opt text;
+    // Origin of the trusted MCP server (no trailing slash). Enables the backend
+    // /mcp delegation path; `null` leaves it disabled.
+    mcp_server_origin : opt text;
```
aterga: Instead of all users having to trust the same MCP server, 

we should change the feature to enable each user to specify the MCP server they want to trust.

this way, it's up to the user to enforce that this server can be trusted to manage their own sessions keys (e.g., by running the server on their own machine)

src/frontend/src/lib/constants/store.constants.ts
```diff
@@ -2,7 +2,7 @@ export const storeLocalStorageKey = {
   LastUsedIdentities: "ii-last-used-identities",
   Locale: "ii-locale",
   CliAccess: "ii-cli-access",
-  McpAccess: "ii-mcp-access",
+  McpTrustedServers: "ii-mcp-trusted-servers",
```
aterga:
```suggestion
  McpTrustedServers: "ii-mcp-trusted-server",
```

src/internet_identity/src/mcp.rs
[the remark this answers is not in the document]
aterga: Landed. The connect flow now binds to a specific, user-chosen account via the standard account picker (`ContinueView`), and `mcp_set_access` enable/disable both derive the principal from that fixed account — so disable revokes the exact principal that was granted, and changing the default account for an origin no longer orphans an old binding. Resolving.

---

src/internet_identity/src/mcp.rs
aterga: Done. The `/mcp` connect screen now reuses the regular continue-with-ii account picker (`ContinueView`) instead of a bespoke flow, so the user binds a specific account exactly as in the normal authorize flow (multi-account toggle, add/edit account, etc.). Resolving.

---

src/internet_identity/internet_identity.did
aterga: Implemented. The feature is now per-user: each identity sets the one (remote, https) MCP server it trusts under Settings → Trusted MCP server, behind a device master toggle, and the connect flow only proceeds when the request's callback origin matches that user's trusted server. The global/deploy-configured `mcp_server_origin` is gone (removed from the frontend-canister init, deploy scripts, and CI). Resolving.

---

src/frontend/src/lib/constants/store.constants.ts
aterga: Applied — the key is now `ii-mcp-trusted-server`. Resolving.

---

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpTrustedServersSection.svelte
```diff
@@ -0,0 +1,337 @@
...
+    }
+    const next = event.currentTarget.checked;
+    const previous = enabled;
+    enabled = next; // optimistic; reverted below if the write fails
+    saving = true;
+    try {
+      await setMcpEnabled($authenticatedStore.actor, identityNumber, next);
+      // Re-check a previously-set server when re-enabling.
+      if (next && trusted !== undefined) {
+        void verify({ notify: true });
```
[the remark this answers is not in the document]
aterga: Good catch, and accurate. This is pre-existing behaviour: the trusted-server setting is a pre-gate for *new* connects, not a kill switch for already-issued standing delegations (the earlier device-local version didn't revoke either). We're keeping that semantics for this PR and tracking revoke-on-remove (call `mcp_set_access(false)` for the origin's bound accounts on remove/disable) plus a possible "connected servers" management surface in #4064. Existing access still ends when the standing delegation expires (the user-chosen TTL, ≤ 30 days).

---

src/internet_identity_frontend/src/main.rs
```diff
@@ -299,6 +304,15 @@ fn get_asset_headers(
     headers
 }
 
+/// Whether `path` is one of the url_paths the prerendered `/mcp` connect page
+/// is certified at. A top-level `mcp.html` (adapter-static, `prerender = true`)
+/// is served at `/mcp`, `/mcp/`, and `/mcp/index.html` (see
+/// `asset_util::filepath_to_urlpaths`), so the `form-action` relaxation must
+/// cover all three. This is the *only* asset that gets the relaxed policy.
+fn is_mcp_landing_path(path: &str) -> bool {
+    matches!(path, "/mcp" | "/mcp/" | "/mcp/index.html")
```
MRmarioruci, line 313: Could we do a wildcard match here?? like /mcp*

MRmarioruci: if it's only achievable by regex, I prefer not to and this would be fine

src/internet_identity/src/mcp.rs
```diff
@@ -0,0 +1,181 @@
...
+) -> Result<(), String> {
+    let principal = mcp_principal_for(anchor_number, account_number, &mcp_server_origin);
+    storage_borrow_mut(|storage| {
+        if enabled {
+            storage.set_anchor_mcp_principal(principal, anchor_number);
+        } else {
+            storage.remove_anchor_mcp_principal(principal);
+        }
+    });
+    Ok(())
```
MRmarioruci: This always returns `Ok(())`, but `set_anchor_mcp_principal` silently does nothing if the principal is already bound to a different anchor (`storage.rs:1082-1091`). On a collision the UI shows "enabled" while nothing was written, so every later `mcp_*` call then fails `Unauthorized`, and toggling off is a no-op too. Could `set_anchor_mcp_principal` report whether it bound, and surface an `Err` here? Fails closed, so not a security hole, just a failure dressed as success.

src/internet_identity/src/mcp.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/// `mcp_get_account_delegation`: fetch the signed delegation prepared above.
+pub fn get_account_delegation(
+    target_origin: FrontendHostname,
+    session_key: SessionKey,
+    expiration: Timestamp,
+) -> Result<SignedDelegation, AccountDelegationError> {
+    let anchor_number = caller_anchor()?;
+    let account_number = default_account_number(anchor_number, &target_origin);
```
MRmarioruci: `prepare` (line 154) and `get` (here) each resolve `default_account_number` independently. If the user's default account at `target_origin` changes between the two calls, `get` looks under a different account than `prepare` signed and returns `NoSuchDelegation`. The browser flow avoids this by threading one explicit `account_number` through both. Resolve once in `prepare` and reuse it?

src/frontend/src/lib/components/ui/ManageHandoff.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
```
MRmarioruci, line 1: Not sure if this duplicated with the other dialog we have in the /authorize flow. If yes, then please try to use the same dialog to avoid duplication

src/frontend/src/lib/flows/manageHandoffFlow.svelte.ts
```diff
@@ -0,0 +1,98 @@
+import { get } from "svelte/store";
+import { goto } from "$app/navigation";
+import {
```
MRmarioruci, line 3: Same as ManageHandoff.. If this is duplicated then would be nice to consolidate. Otherwise ignore

MRmarioruci (approved):

src/internet_identity_frontend/src/main.rs
aterga: We can improve this later

src/internet_identity/src/mcp.rs
aterga: Principal collisions are extremely unlikely, but yes

aterga: Done — surfaced in the latest push. `set_anchor_mcp_principal` now returns `Result<(), AnchorNumber>` (Err = the other anchor on a cross-anchor collision), and `set_mcp_access` maps that to an `Err(String)` instead of returning `Ok` on a silent no-op. So a collision is now reported rather than dressed as success. Agreed it's extremely unlikely, but it's no longer a silent failure.

---

src/internet_identity/src/mcp.rs
aterga: Valid — `prepare` and `get` resolve the default account independently, so a default-account change between the two calls makes `get` look under a different account and return `NoSuchDelegation` (fails closed, no security impact). The clean fix threads one resolved `account_number` through both, which is a small candid change to `mcp_get_account_delegation`. I've noted it under "Known follow-up" in the PR description to keep this PR's surface focused (alongside revoke-on-remove, #4064) — happy to fold it in here instead if you'd prefer.

---

src/frontend/src/lib/components/ui/ManageHandoff.svelte
aterga: Good catch — it was duplicated. Consolidated in 89e56ab: `/authorize/+layout.svelte` now routes its "Manage identity" handoff through this same `ManageHandoffFlow` + `ManageHandoff` component (target `/manage`) instead of its own inline `pendingManageOpen` state, open-tab/dismiss handlers, and "You're signed in" dialog. The component's default copy matches the old inline dialog verbatim, so behaviour is unchanged. Resolving.

---

src/frontend/src/lib/flows/manageHandoffFlow.svelte.ts
aterga: Yes — this flow + the `ManageHandoff` dialog were duplicated in `/authorize/+layout.svelte`. Consolidated in 89e56ab so the authorize layout now uses this shared `ManageHandoffFlow` (and `<ManageHandoff>`) directly, dropping its inline copy. Resolving.

---

src/internet_identity/src/mcp.rs
aterga: Done in 0fb42c4. `mcp_prepare_account_delegation` now returns the `account_number` it resolved (via a new `McpPrepareDelegation` type, leaving the shared `PrepareAccountDelegation` untouched), and `mcp_get_account_delegation` takes that `account_number` as a parameter — so `get` reads the exact account `prepare` signed for instead of independently re-resolving the (mutable) default. The caller is still authorized only for its own anchor (recovered from `caller()`), so it can only fetch what it prepared.

Added a regression test (`mcp_get_uses_prepared_account_despite_default_change`) that reserves a *different* default account at the target origin between prepare and get, then asserts the threaded account still resolves while the now-current default returns `NoSuchDelegation` — i.e. exactly the divergence the old re-resolving `get` would have hit. Resolving.

---

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -594,6 +594,21 @@ pub struct PrepareAccountDelegation {
...
 }
 
+/// Result of `mcp_prepare_account_delegation`. Unlike the browser flow — where
+/// the caller chooses the account and so already knows it for the matching
+/// `get` — the MCP server doesn't pass an account; the canister resolves the
+/// anchor's default account at `target_origin`. We return that resolved
+/// `account_number` so the server can thread the *same* account into
+/// `mcp_get_account_delegation`; otherwise `get` would re-resolve the (mutable)
+/// default and, if it changed in between, look under a different account and
+/// return `NoSuchDelegation`.
```
aterga: Why don't we pass the account selected by the user (either explicitly or implicitly)? 

How would if the user selects a non-default account?

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -594,6 +594,21 @@ pub struct PrepareAccountDelegation {
     pub expiration: Timestamp,
 }
 
+/// Result of `mcp_prepare_account_delegation`. Unlike the browser flow — where
+/// the caller chooses the account and so already knows it for the matching
+/// `get` — the MCP server doesn't pass an account; the canister resolves the
```
aterga: The comment seems stale!

src/internet_identity_interface/src/internet_identity/types.rs
aterga: Good call — done in a61d756. `mcp_prepare_account_delegation` now takes an optional `account_number`, mirroring the browser `prepare_account_delegation`:
- **explicitly**: the MCP server names one of the anchor's accounts at `target_origin`;
- **implicitly** (`None`): it acts as the anchor's default account there.

So a **non-default** account is now reachable per app. An `account_number` that isn't the anchor's at `target_origin` is rejected as `Unauthorized` (`read_account` validates it via `find_account_reference`).

One subtlety worth flagging on "the account selected by the user [at connect]": accounts are **per-origin** — the seed is derived from `account_number`/anchor **plus** the origin (`calculate_account_seed`/`calculate_anchor_seed`), and account records are scoped per `(anchor, origin)`. So the account the user picks at the *MCP server* origin during connect determines the **standing-credential principal** at that origin, but it has no counterpart at an unrelated *target app* origin — the per-app delegation is necessarily for one of the anchor's accounts **at the app** (default, or the now-explicit one). The selection at connect can't be transplanted across origins.

Added `mcp_prepares_for_an_explicitly_named_account` (creates a specific account at the target, asserts `prepare` mints for it and echoes it back, and `get` resolves the same number). Resolving — happy to adjust if you had a different threading in mind.

---

src/internet_identity_interface/src/internet_identity/types.rs
aterga: Fixed in a61d756 — the comment was self-contradictory ("the MCP server doesn't pass an account" while also describing it threading the account into `get`). Rewrote it to describe the new explicit/implicit account selection and the per-origin point (it's an account at `target_origin`, not the one picked at the MCP server origin during connect). Resolving.

---

alexeychirkov:
@aterga sorry for the extra ping here.

I left a follow-up question on #4064 about the MCP access lifecycle and the assumption that access ends when the standing delegation expires:

https://github.com/dfinity/internet-identity/issues/4064#issuecomment-4835482544

As far as I understand, the edge case mentioned there may already be active in production, so I wanted to make sure it is not missed.

Since this PR introduced the relevant MCP access flow, could you please take a look when you have a chance?

PR #4056, opened 2026-06-23
feat(be,fe): expose verified emails as attribute sources

MRmarioruci:
<div align="left">Previous: #4046</div>

Stacks on #4046 (Phase 1 — verified emails as a first-class anchor primitive). Implements [Phase 2 of the design doc](docs/design/verified-email-attributes.md#phase-2--verified-emails-as-attribute-sources): dapps that ask for `email` / `verified_email` on a passkey-only anchor with a verified email now receive a value via the existing ICRC-3 attribute flow, without needing an OIDC link.

## Changes

**Backend**
- `list_available_attributes` walks `Anchor.verified_emails` after the OIDC walk and surfaces each entry as two unscoped wire rows: `("email", <address>)` and `("verified_email", <address>)`. An anchor with N verified emails produces 2N rows that share keys but carry distinct values. `name` is not surfaced (no name claim from the verified-emails flow). `Anchor.email_recovery` is intentionally not walked — the recovery email stays private.
- **Dedup against the OIDC mirror**: Phase 1 silently writes a `verified_emails` entry whenever OIDC vouches for an address; without dedup the consent dialog would offer the same address twice. The OIDC source wins because it carries the IdP brand the user recognises. After unlinking the IdP, the mirrored entry stays and the next consent surfaces it unscoped.
- `prepare_icrc3_attributes` accepts unscoped `email` / `verified_email` specs (scope `None`) and resolves them by case-insensitively matching the spec's `value` field against `Anchor.verified_emails`. Unscoped `name` is rejected. Scoped OIDC / SSO arms are unchanged.

**Frontend**
- `AttributeConsentView.svelte` — `getProviderName` returns `Verified email` as the picker label for unscoped rows (no IdP config to look up). Picker option ids encode `key + "|" + displayValue` so multiple unscoped rows sharing the same wire key (one anchor, N verified addresses) stay distinct in the `{#each}`.
- `AttributePicker.svelte` — adds `onVerifyNew` prop; when set, the dropdown shows an "Add another email" CTA. The CTA sits as a sibling of the `role="listbox"` container so the listbox holds only options.
- `attributeConsentView.utils.ts` — bucket option keys discriminate by `(scope, displayValueLowercase)` so unscoped multi-value buckets enumerate each address as its own picker option. The merge of `email` + `verified_email` pairs by the same key.
- `channelHandlers/attributes.ts` — `resolveKey` fans out exact + suffix matches for unscoped requests so an unscoped `email` request picks up both OIDC `openid:google:email` rows and unscoped `email` rows.

## Tests

- BE unit tests in `attributes::tests::verified_email_attributes_tests` cover: unscoped `email` + `verified_email` listing per entry, multiple verified addresses producing multiple rows, OIDC dedup (case-insensitive), recovery-email exclusion, unscoped resolution by value match, unscoped `name` rejection, value-required rejection.
- **E2E pending** — full attribute-flow spec (dapp requests `email` → consent dialog shows "alice@example.com" tagged "Verified email" → accept → value returned) will land before flipping to ready.

## Notes

- No `.did` change for the attribute surface — verified emails travel as bare `email` / `verified_email` strings; the request shape already carries an optional `value` field that doubles as the picker selector.
- Phase 1.5 (Reach page) and Phase 3 (smart-routing) remain separate follow-ups.


src/internet_identity/src/main.rs
```diff
@@ -2064,6 +2091,24 @@ mod attribute_sharing {
...
+        if let Some(scope) = new_last_shared {
+            if anchor.last_shared_email_source.as_deref() != Some(scope.as_str()) {
+                anchor.last_shared_email_source = Some(scope);
+                if let Err(err) = state::storage_borrow_mut(|storage| storage.write(anchor)) {
+                    ic_cdk::println!(
+                        "Warning: failed to persist last_shared_email_source: {err:?}"
+                    );
+                }
+            }
+        }
```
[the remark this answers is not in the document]
MRmarioruci: Good catch — fixed in 44d8c8374. Re-reading the latest anchor inside the `storage_borrow_mut` closure and only mutating `last_shared_email_source` on that fresh copy.

src/frontend/src/routes/(new-styling)/authorize/views/AttributeConsentView.svelte
```diff
@@ -408,6 +438,10 @@
                   selectedIndex: index,
                 });
               }}
+              onVerifyNew={group.name === "email" ||
+              group.name === "verified_email"
+                ? () => (showVerifyWizard = true)
+                : undefined}
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in 44d8c8374. The probe placeholder now reserves chevron space when the group is `email`/`verified_email`, mirroring the picker's expandability rule.

src/frontend/src/routes/(new-styling)/authorize/views/AttributePicker.svelte
```diff
@@ -229,6 +240,21 @@
           {/if}
         </button>
       {/each}
+      {#if onVerifyNew !== undefined}
+        <button
+          onclick={() => {
+            isOpen = false;
+            onVerifyNew();
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in 44d8c8374. Split into an outer flex container that holds the `role="listbox"` div (options only) and the "Add another email" CTA as a sibling below.

src/internet_identity/internet_identity.did
```diff
@@ -987,6 +987,7 @@ type IdentityInfo = record {
     // shows a "limit reached" notice in the wizard when adding
     // beyond the cap.
     verified_emails : opt vec VerifiedEmail;
+    last_shared_email_source : opt text;
```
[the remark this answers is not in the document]
MRmarioruci: Fair — PR description updated to note the additive `last_shared_email_source : opt text` field on `IdentityInfo` (backwards-compatible).

aterga: It's unclear what "email source" means. Why not save the last-shared email?

src/internet_identity_interface/src/internet_identity/types/attributes.rs
```diff
@@ -99,15 +99,38 @@ impl std::fmt::Display for AttributeName {
...
+/// collision resistance for a per-anchor scope identifier and short
+/// enough to keep `list_available_attributes` responses tight.
+pub const VERIFIED_EMAIL_SCOPE_HASH_HEX_LEN: usize = 16;
+
 impl std::fmt::Display for AttributeScope {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             AttributeScope::OpenId { issuer } => write!(f, "openid:{}", issuer),
             AttributeScope::Sso { domain } => write!(f, "sso:{}", domain),
+            AttributeScope::Verified { address_hash } => write!(f, "verified:{}", address_hash),
```
aterga: Interesting, 

What is the use case for a scoped verified email?

If an app wants a verified email, it usually doesn't care where it comes from (as long as it can trust it)

(For OpenID and SSO we needed this for the 1-Click flows; but I don't think we need a new scope now)

src/internet_identity/src/attributes.rs
```diff
@@ -10,13 +10,22 @@ use internet_identity_interface::internet_identity::types::{
...
     },
     icrc3::Icrc3Value,
     Timestamp,
 };
+use sha2::{Digest, Sha256};
 use std::collections::{BTreeMap, BTreeSet};
 
+/// Stable short hash used as the `<H>` in `verified:<H>:...` scope
+/// keys so the plaintext address never appears in the candid surface.
+pub fn verified_email_scope_hash(address: &str) -> String {
```
aterga: Let's not implement this in this PR (interesting use case though)

src/internet_identity/internet_identity.did
MRmarioruci: Field dropped in 3c32a3247.

src/internet_identity_interface/src/internet_identity/types/attributes.rs
MRmarioruci: Agreed — `AttributeScope::Verified` is gone in 3c32a3247. Verified emails now surface as unscoped `email` / `verified_email` rows; the canister resolves them in `prepare_icrc3_attributes` by case-insensitively matching the spec 's `value` against `Anchor.verified_emails`.

src/internet_identity/src/attributes.rs
MRmarioruci: `verified_email_scope_hash` is gone in 3c32a3247 along with the rest of the scope plumbing.

aterga (approved):

PR #4057, opened 2026-06-23
fix(be): pin DNSKEY RRset validation to the DS-matched KSK (RFC 4035 §5.2)

aterga:
## Summary

Fixes a DNSSEC chain-of-trust bypass in the email-recovery verifier (`src/internet_identity/src/dnssec/verify.rs`). `verify_link` accepted a child DNSKEY RRset whose self-signature was made by **any** key present in the RRset, instead of the specific KSK the parent DS commits to (RFC 4035 §5.2).

- **2b** (`verify.rs`) looked up a DS-matching KSK only to confirm one exists, then **discarded which key it was** (`let _ = pick_matching_ksk(...)`).
- **2c** then validated the DNSKEY RRset's RRSIG against the **whole** rrset (`verify_rrsig_under_dnskey_rrset(&link.child_dnskey, &link.child_dnskey)`).

## Impact

A zone's KSK public bytes and its DS record are both public DNS data, so an attacker who controls **neither the domain nor any private key** could:

1. Build a forged DNSKEY RRset = `{ genuine DS-pinned KSK, attacker key }`,
2. Self-sign it with the **attacker** key (passing 2c), while the genuine KSK satisfies the 2b digest match,
3. Have the attacker key admitted as a trusted signer for the zone — and then forge any DKIM/TXT record under it, even for a victim domain that isn't itself DNSSEC-signed (by injecting at a signed parent such as a TLD).

Through the email-recovery flow this lets a fully fabricated DNS proof validate a forged DKIM record, enabling credential binding / account recovery off DNS the attacker never controlled.

## Fix

Collect **every** DS-pinned KSK (`matching_ksks`, so a multi-DS KSK rollover still validates) and require the DNSKEY RRset's RRSIG to verify under one of those keys **specifically** — never under an unpinned in-set key. A new `verify_rrsig_under_any_dnskey` takes an explicit candidate set; `verify_rrsig_under_dnskey_rrset` (still used for 2a and hop verification over an already-trusted full set) delegates to it.

## Behavioural change (please review)

A zone whose **DS-pinned KSK uses an algorithm this verifier does not implement** (e.g. `mailbox.org`'s RSA-SHA1 / algorithm-7 KSK) can no longer be validated via DNSSEC and **falls back to the DoH path** (design doc §7.6). This is correct fail-closed behaviour — the previous "success" relied on accepting a non-DS ZSK's signature, i.e. the bug itself. The `mailbox.org` vector test is updated to assert rejection.

**Front-end coordination required:** bundle assembly must supply the **KSK's** RRSIG over the DNSKEY RRset (the one whose `key_tag` matches the DS), not the ZSK's. In real DNSSEC the KSK signs the DNSKEY RRset, so this RRSIG always exists for validatable zones; the FE just needs to select it.

## Tests

`cargo test -p internet_identity --bin internet_identity dnssec::verify::` → **22 passed, 0 failed**:

- All five real captured chains still validate (cloudflare, proton.me, protonmail, tutanota, ed25519-nl).
- New regression module (`ds_pin_regression`) with a self-contained Ed25519 signer:
  - forged-key-self-signed rrset → **rejected** (`BadSignature`)
  - ZSK-only-signed rrset → **rejected**
  - no DS-pinned KSK present → **rejected** (`DsMismatch`)
  - legit KSK-signed rrset → **accepted**
  - KSK rollover (DS pins two KSKs, rrset signed by the second) → **accepted**
- `mailbox.org` vector → now asserts rejection (unsupported RSA-SHA1 KSK).

Also verified: `clippy` clean, and the canister `wasm32-unknown-unknown` release build succeeds. The fix was additionally confirmed end-to-end against a local PocketIC reproduction (the forged-bundle takeover is rejected post-fix).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GvNUrqYu86TKRgpefnDWaT)_


MRmarioruci (approved):
I guess there should be a FE PR as well?

aterga:
Yes — good catch, there is. I've opened it: #4072.

This PR's verifier change requires a child DNSKEY RRset to be signed by the **DS-pinned KSK** (RFC 4035 §5.2). The browser-side bundle assembler (`selectSupportedCoveringRrsig` in `src/frontend/src/lib/utils/dnssec/chain.ts`) currently attaches the *first* covering, supported-algorithm RRSIG, which for a zone that signs its DNSKEY RRset with both a KSK and a ZSK can be the ZSK's — and that would now be rejected even for a legitimate zone. #4072 makes the FE bundle the RRSIG matching a SEP (KSK) key in the RRset (and fall back to DoH when no supported KSK signature exists, e.g. mailbox.org's RSA-SHA1 KSK).

They can land independently — preferring the KSK's RRSIG is correct regardless of whether this verifier is deployed yet — but #4072 is required for legitimate KSK+ZSK-signed zones once this one merges.

---

PR #4062, opened 2026-06-24
fix(fe): show correct SSO error when a domain isn't allowlisted

aterga:
## Problem

On id.ai, typing an SSO domain that isn't on the canister's discovery allowlist (e.g. `beta.dfinity.org` against the production canister, whose allowlist is `["dfinity.org"]`) shows:

> Couldn't load SSO settings from beta.dfinity.org. Ask your SSO admin to check that /.well-known/ii-openid-configuration is reachable.

This is the wrong message. When a domain isn't allowlisted, the canister rejects it at the allowlist gate (`get_sso_discovery` → `NotAllowed`) and **never even attempts** the `/.well-known/ii-openid-configuration` fetch — so advising the admin to check that endpoint's reachability is misleading.

## Root cause

`discoverSsoConfig` maps the canister's `NotAllowed` to `DomainNotConfiguredError("rejected")`. In `SignInWithSso.svelte`, the error mapper had the two `DomainNotConfiguredError` reasons mapped to semantically-mismatched copy:

- `rejected` (not allowlisted) → "...check that /.well-known/ii-openid-configuration is reachable." ❌
- `timeout` (allowlisted, but discovery never resolved) → "took too long to respond." 

The correct "domain isn't registered for SSO" copy already existed, but only inside a dead `canary allowlist` string-match branch that nothing ever triggers (no error in the codebase carries that text).

## Fix

- **`rejected`** now reports that SSO isn't available for the domain yet, reusing the message previously stranded in the dead branch.
- The **`/.well-known/ii-openid-configuration` reachability advice** moves to the **`timeout`** case — the one situation where the domain *is* allowlisted but discovery never resolved (a failed/unreachable fetch surfaces as a timeout because the canister keeps reporting `Pending`), so checking that endpoint actually helps.
- Removed the now-dead `canary allowlist` branch and the orphaned `took too long to respond` catalog entry across all locales.

Both messages I now use already existed as msgids, so existing translations carry over and no new untranslated strings are introduced.

## Testing

No automated tests assert on these UI strings (verified). The change is a 1:1 remapping of which existing message is shown for each `DomainNotConfiguredError` reason, plus removal of dead code. Manual verification on a deployment where the typed domain is off the allowlist will now show "SSO is not available for ... yet. Ask an II admin to register this domain." instead of the endpoint-reachability message.

https://claude.ai/code/session_01GgPmHHcAwFqjo2j64KXr7G

---
_Generated by [Claude Code](https://claude.ai/code/session_01GgPmHHcAwFqjo2j64KXr7G)_

### DIFF
--- a/src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
+++ b/src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
@@ -96,9 +96,18 @@
   ): string | undefined => {
     if (e instanceof DomainNotConfiguredError) {
       if (e.reason === "timeout") {
-        return $t`${domainInput} took too long to respond. Try again in a moment.`;
+        // The domain IS on II's allowlist, but discovery never resolved.
+        // A failed or unreachable discovery fetch surfaces here too — the
+        // canister keeps reporting `Pending` until we time out — so this is
+        // the case where pointing the SSO admin at the discovery endpoint
+        // actually helps.
+        return $t`Couldn't load SSO settings from ${domainInput}. Ask your SSO admin to check that /.well-known/ii-openid-configuration is reachable.`;
       }
-      return $t`Couldn't load SSO settings from ${domainInput}. Ask your SSO admin to check that /.well-known/ii-openid-configuration is reachable.`;
+      // `rejected` is the canister's `NotAllowed`: the domain isn't on II's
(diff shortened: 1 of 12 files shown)


MRmarioruci (approved):
maybe removee the locales? not sure if Thomas's bot also removes unused ids..

PR #4063, opened 2026-06-25
fix(fe): pass SSO discovery domain when switching to an SSO method

MRmarioruci:
On /manage/access, switching the active access method to an SSO method (More options → Switch → Continue) failed with "Authorization invalid". The switch handler extracted `ssoDomain` to fetch the JWT, but dropped it before calling `authenticateWithJWT`, so the subsequent `openid_prepare_delegation` / `openid_get_delegation` calls ran without a discovery domain and the canister had no JWKS to verify the JWT against.

Sign-in, account-link, and signup paths thread `discoveryDomain` through correctly (see [`authFlow.svelte.ts:542`](../blob/main/src/frontend/src/lib/flows/authFlow.svelte.ts#L542), [`authFlow.svelte.ts:816`](../blob/main/src/frontend/src/lib/flows/authFlow.svelte.ts#L816)); only the switch path was missing it.

While in the same block, the lastUsed-identity entry written after a successful switch always recorded `authMethod: { openid: { iss, sub, ... } }`, even when the switched-to method was SSO — that would have made the next "last used" sign-in go down the configured-OpenID branch and miss the domain. The sign-in flows record SSO as `authMethod: { sso: { domain, name, email, loginHint } }`; this PR matches that shape.

## Changes

- Pass `discoveryDomain: ssoDomain` to `authenticateWithJWT` in `handleSwitchConfirmed`.
- Cache the `discoverSsoConfig` promise so its `name` field can be reused without re-running discovery, then store the lastUsed entry as `sso` (with `domain`, `name`, `email`, `loginHint`) when the method is SSO, falling back to `openid` for configured providers.

## Tests

- `npm run check` (svelte-check + tsc): 0 errors.
- `npm run lint`: 0 problems.
- New Playwright spec `tests/e2e-playwright/routes/manage/access.spec.ts` — describe `"Access methods > with an SSO method"` — signs in with a passkey, links an SSO method via the Add access method dialog, then exercises More options → Switch → Continue on the SSO item. Asserts (a) the success toast fires, (b) the Active badge moves to the SSO item, and (c) the lastUsed entry is tagged as `sso` (with `domain`, `email`, `loginHint`) so a subsequent "last used" sign-in routes through the SSO branch. Local e2e run was skipped (port 8000 was held by a parallel worktree's `icp network`); relying on this PR's `canister-tests / e2e-playwright` CI job to verify.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/access/+page.svelte
@@ -271,9 +271,13 @@
               ? lastUsedAuthMethod.sso.loginHint
               : undefined;
         let jwt: string;
+        let ssoDiscoveryPromise:
+          | ReturnType<typeof discoverSsoConfig>
+          | undefined;
         if (ssoDomain !== undefined) {
+          ssoDiscoveryPromise = discoverSsoConfig(ssoDomain);
           jwt = await requestWithPopup(
-            discoverSsoConfig(ssoDomain).then((ssoResult) => ({
+            ssoDiscoveryPromise.then((ssoResult) => ({
               clientId: ssoResult.clientId,
               authURL: ssoResult.discovery.authorization_endpoint,
               authScope: selectAuthScopes(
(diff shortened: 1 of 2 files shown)


aterga (approved):
Approving, but please test it on staging before you merge

PR #4071, opened 2026-06-29
docs(be,fe): push notifications PoC design

MRmarioruci:
Design proposal for routing dApp → user push notifications through Internet Identity as a single permission + delivery hub. The doc settles the shape before code lands: what it is for, what it deliberately is not, the two paths that decide whether it scales, the security and privacy model, and what is still open. The implementation PoC is tracked separately in #4109.

Start with [How it works, in plain terms](https://github.com/dfinity/internet-identity/blob/docs/push-notifications-poc/docs/push-notifications.md#how-it-works-in-plain-terms) — the whole design in six bullets. The doc opens with a reading path routed by what you came for: reviewing, implementing, integrating, or just wanting to know what exists today.

Positions the doc takes:

- **II is the push hub.** The user grants notification permission to id.ai once and consents per dApp; every consented dApp delivers through II instead of running its own service worker + VAPID keypair + subscriptions. PWA install is optional on Android/desktop and required only on iOS Safari. The alternative — every dApp running its own pipeline — is now argued and rejected explicitly, because the scarce resource is the permission, not the plumbing: a user declines the *n*th prompt far more often than the first.
- **Two scalability paths.** dApp → II (chunked send with mandatory II-side admission control plus cooperative client-side pacing) and II → device (batched through a trusted web2 gateway, with direct per-device outcalls as the documented alternative).
- **II's handoff to the gateway is a non-replicated outcall.** This is the design, not a future option. It makes the gateway genuinely stateless (no idempotency window needed to collapse 34 duplicate copies), lets per-device status come back inline — which is what makes `410`-driven subscription cleanup possible at all — and drops the payload cost by the full factor of *n*: ~0.7 T → **~0.02 T** for a 10k-user blast.
- **Stateless for campaigns.** The durable campaign lives in a dApp-side client library; II keeps only user-scoped data (subscriptions, consent) plus a small transient buffer, so its stable storage stays flat regardless of send volume.
- **The buffer is now sized, not just called "small".** Entries are per recipient rather than per device (~200 B, or ~1 KB when every recipient gets different text). Memory is not the constraint by three orders of magnitude — drain rate is, and it is a pie shared by every origin, so ten origins blasting at once slows delivery for everyone rather than for the noisy one. Buffer depth should therefore be drain rate × acceptable queue latency (~6 MB at 60 s), because a memory-sized buffer would accept hours of backlog and report `admitted` for messages whose TTL expires first — a lying success rather than an out-of-memory.
- **Throughput is capped by a budget, not by the platform.** The cost is RFC 8291's ECDH, two scalar multiplications per device-message; the VAPID signature is cached per audience and irrelevant. The slice is kept small because this canister also serves every login on the network. Moving the sealing to the gateway is not merely disallowed but useless — deriving the content key and being able to decrypt are the same capability — so the alternative that would actually lift the ceiling is a separate push canister, recorded as needing its own review.
- **Targeting by in-app principal**, per-origin consent with origin pinning, and per-origin admission control (Layer 1).
- **Sender verification is automatic — a dApp's only obligation is publishing a file.** `origin_sha256 → sender canister` is what authorizes a send, because the original `caller() == in_app_principal` rule can never match an inter-canister call and so admitted only self-sends. But a dApp should not have to call a setup endpoint: it publishes `/.well-known/ii-push-senders` listing its backend canister, and II verifies that itself — on first consent (preferred: user-paced, so the row exists before the first send) or on first send as a backstop, returning `SenderUnverified` with an actionable hint without blocking on the outcall. `push_register_sender` remains only as "re-check me now" after editing the file, and the client library owns the retry so the dApp writes none of it. **The gate that makes this safe: only ever fetch for origins a user has already consented to** — otherwise any canister could make II fetch a URL of its choosing by naming an arbitrary origin, an SSRF/DoS amplifier of the kind the endpoint allowlist exists to prevent. Consent bounds the fetchable set to grants that already happened, and costs nothing since a send to a non-consenting user is refused anyway.
- **VAPID key custody.** Generated via `raw_rand` and held in stable memory today because IC threshold ECDSA is secp256k1-only while Web Push needs P-256; migrate to `sign_with_ecdsa` if P-256 threshold ECDSA ships.
- **E2E.** A `Hidden` content-hidden baseline that ships first (generic notification + tap-through reveal), plus two documented future paths for rich content on the lock screen — vetKeys-sealed (II's SW decrypts on-device, II-the-service stays blind) and dApp-fetch.
- **VAPID keys are never rotated as maintenance.** There is no expiry and no hygiene schedule; a rotation is an incident. A *planned* migration to a subnet-held key needs no user-visible event — a subscription is bound to the `applicationServerKey` it was created with, so II honours the old key for devices already on it while issuing new subscriptions under the new one, and the fleet migrates as endpoints churn. Only *compromise recovery* invalidates everything at once, and that is what the "re-enable on your devices" UX is for. **P-256 threshold ECDSA is the end state** — the subnet signs, no stored secret exists to extract, and the custody caveat is deleted rather than mitigated.
- **Per-anchor caps are sized at 20 subscriptions / 100 consents** (~19.5 KB per anchor typical, ~35 KB worst case, since every consent costs a principal-index row too). Read as anti-griefing, not a storage bound: all 10M anchors saturating it would be 200–360 GB against a 537 GB limit, so it stops one anchor mining II's storage while aggregate growth stays bounded by real usage plus stale-row cleanup.
- **Shared devices** bind push per identity, never per device — a shared endpoint under one identity implies no consent for another.

## Changes

- `docs/push-notifications.md` — the design doc.
- `docs/push-api.md` — **new.** The Candid interface and what a dApp must do to integrate.
- `docs/push-client-library.md` — **new.** The client library's design, send loop, the state it must keep, and the failure modes it owns.

The split follows the guidance in [Design Docs at Google](https://www.industrialempathy.com/posts/design-docs-at-google/): the Candid surface and the library internals are needed to show the design is buildable, but they are reference material rather than decision analysis, and reading through them to reach the next decision is what made the doc hard to follow. The design doc keeps a stub for each, carrying only the part that is a design argument.

The doc was also restructured to that post's shape, which it was missing three sections of: **context and scope** (why a dApp cannot solve this alone), **goals and non-goals** (the non-goals being the load-bearing half — not a messaging product, not exactly-once, not guaranteed delivery, no billing in v1, no way to address anyone who has not consented), and **alternatives considered**. "How it works, in plain terms" now actually comes first, as it always told the reader to read it first; build status moved to the end, being project tracking rather than design. All 13 question-phrased headings became statements, with every anchor link updated.

Corrections worth calling out for anyone who reviewed an earlier revision:

- Stable-memory limits were off by a column: **512 MiB touched per message and 2 GiB per upgrade**, not 2 GiB and 8 GiB. 500 GiB per canister was right.
- The opt-in description conflated two prompts. II's own opt-in screen **is** shown for each new dApp, once per dApp per device; the **browser's** permission dialog is the part that never returns, because it is granted to id.ai and every dApp is delivered through that one origin.
- Tapping a notification opens the dApp's deep link **directly**, and the dApp signs the user in on arrival; II's `/notify` screen is only the fallback when no target was supplied. Four places still described everything routing through `/notify`, including the architecture diagram.
- Notifications can be turned off from the **browser** as well as from II, and a browser-level block cannot be undone from inside II — which is also why the opt-in screen is skipped rather than shown when permission is already denied.

Further corrections, from a pass reviewing the doc for duplicated material:

- **The stable-memory index table was wrong in the dangerous direction** — it omitted the SSO stable-id index at 32 and shifted all three push indices down one. Following it would have reintroduced the collision that actually happened during the PoC: 32 was claimed after `main` had taken it, the two `StableBTreeMap`s interleaved, and the canister trapped inside the B-tree on first read. Corrected to 31–36 against `storage.rs`, with that history recorded under the table.
- **The storage estimate was understated ~2.5×.** Both 100M-row maps were priced as if they held ~17M rows: 10M users × 10 consented dApps is 100M consent rows (~6 GB, not 1 GB) and 100M principal-index rows (~8 GB, not 1.2 GB). Total ~20 GB, not ~8 GB. Still comfortable against the limit, but it changes which axis matters — consents are expensive because each costs two rows; devices are cheap.
- **Duplicated material removed** so each fact has one home: Status no longer restates five Open items, "Feasibility and scale" no longer re-derives the shared-throughput ceiling or the gateway rationale a third time, and the vetKD cost and the `ic-cdk-timers` duplicate-execution note are each stated once. Cross-reference links were deliberately left alone.

Supersedes the earlier `docs/push-notifications-poc.md` and `docs/push-notifications-flow.mmd`; the architecture diagram is now inline.

Two gaps are known and deliberately not yet addressed: the local-development constraint (a dApp backend cannot reach a mainnet II from a local replica, because inter-canister calls only route within one domain — every integrator will hit this), and per-category consent (consent is all-or-nothing per origin, so a venue cannot offer "liquidation alerts yes, every fill no").

### DIFF
--- a/docs/push-api.md
+++ b/docs/push-api.md
@@ -1,0 +1,137 @@
+# Push notifications — II interface and dApp integration
+
+The Candid surface a dApp sends through, and the steps to integrate one.
+
+Design rationale lives in [push-notifications.md](push-notifications.md); this
+document is reference material and assumes you have read at least
+[how it works, in plain terms](push-notifications.md#how-it-works-in-plain-terms).
+
+## dApp developer integration
+
+
+
+One-time setup (~15 min): serve `.well-known/ii-push-senders`, call
+`push_register_sender(origin)` from the backend. Steady state: hand the client
+library a campaign — it chunks, paces, retries and reports:
+
+```rust
+// Non-sensitive app: show the content.
+push.broadcast(
+    PushContent::Display { title, body },
+    PushDelivery { urgency: High, ttl_seconds: 3600, topic: Some("balance") },
+    &my_user_principals,        // library chunks + paces these
+).await?;
+
+// E2E app: send nothing sensitive; content is revealed on tap.
+push.broadcast(
+    PushContent::Hidden { category: Some(Message) },
+    PushDelivery { urgency: High, ttl_seconds: 3600, topic: None },
+    &my_user_principals,
+).await?;
+```
+
+Under the hood the library splits the audience into ≤ ~1000-target chunks,
+calls `push_send` per chunk, **paces on `ready` / `retry_after_ms`**, retries
+with a stable `chunk_id`, and aggregates per-target results into campaign
+status. The dApp sees a campaign API; II sees paced, bounded chunks. No keys,
+no crypto, no Web Push knowledge, and no per-user state beyond the principals
+from auth (the library owns campaign status). Delivery is best-effort with no
+receipts; the only per-user signal is `NoConsent`.
+
+
+## The Candid interface
+
+The exact API a dApp's backend calls, written in Candid (the IC's interface
+language). There is a single send entry point, `push_send`: it carries a
+shared `default_alert` plus a list of recipients, each of which may override
+that alert with its own — so the same call does both "same text for everyone"
+and "different text per user." The rest are the types it uses; skim the
+comments — each explains what the field is for.
+
+```candid
+type PushCategory = variant { Message; Transfer; Update; Generic };
+
+// The content variant is what makes end-to-end encryption possible later
+// (see the end-to-end-encryption section below). Display content is read by II
+// and shown verbatim — fine for non-sensitive notifications, but II sees it.
+// Hidden content is never sent to II at all: the payload carries no message
+// text, so an E2E sender structurally cannot leak content. The service
+// worker renders an II-controlled generic string keyed by category, and the
+// real message is revealed on tap-through when the app decrypts it.
+type PushContent = variant {
+  Display : record {             // II-visible; transport-encrypted only, NOT E2E
+    title : text;                // ≤ 64 bytes
+    body : text;                 // ≤ 256 bytes
+  };
+  Hidden : record {              // content-free; E2E-safe by construction
+    category : opt PushCategory; // maps to II-controlled copy ("New message", …)
+  };
+  Dismiss;                       // close the shown notification named by notification_id; renders nothing
+};
+
+type PushAlert = record {
+  content : PushContent;
+  url : opt text;                // tap-through target; must be same-origin as sender
+  notification_id : opt text;    // dApp's id for this notification (maps to the Web Notification `tag`):
+                                 // reuse it to UPDATE the shown one, or pair with Dismiss to close it
+};
+
+type PushUrgency = variant { VeryLow; Low; Normal; High };
+
+type PushDelivery = record {     // RFC 8030 relay headers; plaintext, relay-visible
+  urgency : opt PushUrgency;     // default Normal; also orders II's drain
+  ttl_seconds : opt nat32;       // default ~4h; 0 = only if online now; clamped to a max
+  topic : opt text;              // ≤ 32 chars base64url; collapse key
+};
+
+type PushRejection = variant {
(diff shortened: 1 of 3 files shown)


MRmarioruci:
Served it's purpose, no longer relevant

PR #4095, opened 2026-07-09
feat(be): scope OpenID anchor lookup by sso_domain (chunked re-key migration)

sea-snake:
**PR 1 of the enterprise-SSO per-app-gating stack.** Backend-only, behavior-preserving — no gating yet.

## Why
For per-app SSO gating to be safe, an OpenID login must resolve to an anchor **scoped by the SSO discovery domain**. Today the anchor lookup index keys on `(iss, sub, aud)` only, with `sso_domain` a non-key field — so a login via an attacker-controlled discovery domain that points at an org's *real* IdP (same `iss`/`sub`/`aud`) collides onto the org's anchor. This PR closes that by putting `sso_domain` in the lookup key, with a chunked migration to re-key existing entries.

Design doc (forthcoming in a docs PR): `docs/ongoing/enterprise-sso-idp-side-gating.md` §6 / §6.1.

## What
Scope is strictly the internal lookup index:
- `StorableOpenIdCredentialKey` gains CBOR field `3` (`sso_domain`) with a **dual-format `from_bytes`** (legacy 3-field + new 4-field).
- `lookup_anchor_with_openid_credential` does a **dual lookup** (full key, then legacy fallback) so logins keep resolving during the migration.
- **Chunked, timer-driven re-key migration** — 2000 keys/batch, cursor-resumable, with hidden `openid_key_migration_status` / `list_openid_key_migration_errors` scaffolding — following the `#4013` / `#3784` convention. Re-keys SSO entries from their already-stored `sso_domain`; direct-provider entries (`None`) are left untouched. Scaffolding to be removed in a follow-up once the migration completes.

## Explicitly unchanged (invariants)
- **Public API / `.did`** — `OpenIdCredentialKey` and `openid_credential_remove` stay the `(iss, sub, aud)` 3-tuple (verified: no `.did` diff vs `main`). The 4-tuple exists only as the internal `OpenIdCredentialLookupKey`.
- **Delegation seed / principals** — `calculate_delegation_seed` is byte-identical to `main`; the anchor it already hashes provides the domain separation transitively, so principals do not shift.

## Testing
- `cargo test -p internet_identity --bin internet_identity` — 610 pass, incl. dual-format key, batch re-key (cursor resume / done / idempotency / orphan), dual-lookup fallback, seed-independence, and `check_candid_interface_compatibility`.
- `cargo clippy`/`fmt` clean; integration target compiles.
- PocketIC integration tests: exercised by CI (not run locally — env version mismatch).

### DIFF
--- a/src/internet_identity/src/anchor_management.rs
+++ b/src/internet_identity/src/anchor_management.rs
@@ -1,5 +1,5 @@
 use crate::archive::{archive_operation, device_diff};
-use crate::openid::{OpenIdCredential, OpenIdCredentialKey};
+use crate::openid::{OpenIdCredential, OpenIdCredentialKey, OpenIdCredentialLookupKey};
 use crate::storage::anchor::{Anchor, AnchorError, Device};
 use crate::storage::Storage;
 use crate::{state, stats::activity_stats};
@@ -200,10 +200,10 @@ pub fn identity_properties_replace(
 
 pub fn check_openid_credential_is_unique<M: Memory + Clone>(
     storage: &Storage<M>,
-    openid_credential_key: &OpenIdCredentialKey,
+    openid_credential_lookup_key: &OpenIdCredentialLookupKey,
 ) -> Result<(), AnchorError> {
     if storage
-        .lookup_anchor_with_openid_credential(openid_credential_key)
+        .lookup_anchor_with_openid_credential(openid_credential_lookup_key)
         .is_some()
     {
         return Err(AnchorError::OpenIdCredentialAlreadyRegistered);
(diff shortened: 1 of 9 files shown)


sea-snake:
Closing: on review this migration isn't needed. The rogue-domain concern it addressed is already covered by the **certified `sso_domain`** — it's stamped per-login from the discovery domain II actually resolves, so it's unforgeable, and enforcement can verify it directly instead of re-keying the anchor index. The simpler approach (check the certified `sso_domain` at the gate; no stable-memory migration) supersedes this. Design doc will be updated accordingly.

PR #4096, opened 2026-07-09
feat(be,fe): IdP-side per-app SSO gating

sea-snake:
Lets an org gate **which IC dapps its users can reach through II SSO** using the IdP's own per-app assignment (Okta / Entra / OneLogin), with no id.ai-side policy, directory, proxy, or panel. Inert until an org opts in via its well-known.

**Design doc:** `docs/ongoing/enterprise-sso-idp-side-gating.md` (companion docs PR).

## How it works
- Each gated dapp gets its own OIDC client in the org's IdP; the org lists `origin → client_id` in `/.well-known/ii-openid-configuration` (`app_clients`, plus `gate_all_apps`, `stable_identifier_claim`).
- **Gate = mint-or-refuse.** `sso_prepare_delegation` verifies the JWT and mints the SSO sign-in only if `aud` matches the client the origin resolves to (per-app / primary / deny). It mints a **plain openid-credential delegation** — identity always resolves to the org's *primary* client — and returns a small **II-canister-signed bundle** carrying `(sso_domain, origin, expiry)` (~5 min).
- **The gating context travels as a certified `sender_info` bundle.** The frontend attaches it via the SDK's `AttributesIdentity`; the canister reads it back with `msg_caller_info_signer/_data` (raw `ic0 1.1`) in `read_certified_sso_bundle`. The replica verifies the bundle's canister signature under the caller's own credential seed, so it's bound to the identity that presented it.
- **`sso:<domain>` is certified only from an SSO session.** `prepare_icrc3_attributes` / `list_available_attributes` include the SSO rows only when a valid bundle whose `origin` matches is presented; a passkey/plain session carries no bundle → no SSO attributes. Delegation/account methods stay permissionless (deriving `f(anchor, origin)` always was); the gate rides solely on the certified attribute a dapp already verifies.
- **First gated login registers directly** (`OpenIDRegFinishArg.origin`): a `sub` IdP registers in one trip; a non-`sub` IdP (Entra `oid`) fails safe with `IdRegFinishError::SsoNormalLoginRequired`, and the frontend guides a normal sign-in first. Non-`sub` identities resolve through a **storage-maintained SSO stable-id index** — the stable id (`oid`) is stored on the primary credential and reconciled into the index (keyed `(iss, primary_client_id, stable_id) → anchor`) on every anchor `write()`, so it **self-cleans** when the credential is removed or repoints when it moves.
- **Frontend ceremony routing.** The ceremony runs against the origin's `resolved_client_id`: the manual "Sign in with SSO" wizard uses the dapp's effective origin; the 1-click `?sso=` path uses `remapToLegacyDomain(derivationOrigin ?? channel.origin)` so client selection keys the same origin the gate does. The FE-supplied origin only selects the client; the canister independently re-validates `aud`.

## `.did` / deps
- Adds `sso_prepare_delegation(SsoPrepareDelegationRequest)` / `sso_get_delegation(SsoGetDelegationRequest)` (+ response records carrying the `sso_attr_bundle` + its signature) and `opt origin` on `OpenIDRegFinishArg`. The discovery query is `get_sso_discovery_status(GetSsoDiscoveryStatusRequest) -> SsoDiscoveryStatus`, reporting the per-origin `resolved_client_id`. `openid_prepare_delegation` / `openid_get_delegation` and the attribute request records are unchanged.
- **Deploy config:** removes the SSO discovery domain allowlist — the `sso_discoverable_domains` and `sso_allow_any_domain` init args are gone, so any bare-authority domain is discoverable over `https`. Adds `sso_allow_insecure_discovery` (`opt bool`, default off): a deploy flag that relaxes the `https` requirement **only for loopback hosts** (`localhost` / `127.0.0.1`) so e2e can point at local `http` mock IdPs. Non-loopback always requires `https` and prod runs with the flag off; CI and `dev-e2e-setup` deploy with it on.
- **Non-additive changes** (all safe — the feature is unshipped): removing the two init args above; renaming `get_sso_discovery` → `get_sso_discovery_status` and wrapping the delegation args in request records; renaming `SsoDiscoveryState` → `SsoDiscoveryStatus` and dropping its `NotAllowed` variant; and the `IdRegFinishError::SsoNormalLoginRequired` variant (inert for existing clients — they never send `origin`).
- Adds `ic0 = "1.1"` for `msg_caller_info_*` (coexists with the transitive `ic0 0.23`; avoids an `ic-cdk 0.20` bump that conflicts with the vc-sdk git deps). No `ic-cdk` bump, no credential-key or delegation-seed change; one new **additive** stable structure — the SSO stable-id index (memory id 32), reconciled on write like II's other lookup indexes — not a migration.

## Length caps
Caller-supplied values are length-bounded: the `app_clients` client-ids and the `stable_identifier_claim` value (this PR); JWT `iss`/`sub` and the discovery `client_id`/`jwks_uri`/`issuer`/`name`/`domain` in the companion base PR #4113.

## Testing (run locally + CI)
- Unit `cargo test -p internet_identity --bins` and integration `openid::sso_gating` green. Coverage: linchpin (gated == ungated → same dapp principal); cross-identity / expired / cross-origin bundle rejected; passkey gets no SSO attrs; `gate_all_apps` deny; SSO-aware registration incl. non-`sub` fail-safe (and a non-`sub` token missing the stable-id claim); the stable-id index self-cleans on credential removal, repoints on move, survives upgrade, and is per-primary-client scoped. clippy/fmt clean.
- Frontend: `check` / `lint` / `format-check` clean. e2e `routes/authorize/` green — 1-click (incl. gated via channel origin and via `derivationOrigin`), manual "Sign in with SSO", last-used SSO continue, consent, and gating.

## Notes for review
- Behavior change (deliberate): `sso:<domain>` attributes now require an SSO session (previously certified from the stored credential in any session).
- The legacy `icp0.io / icp.net → ic0.app` origin canonicalization applies before the `app_clients` lookup — so `app_clients` keys must be the canonical origin.
- Bounded reconfiguration window: flipping an app ungated → gated leaves already-minted bundles valid until expiry (~5 min) plus the ~1 h discovery-cache refresh.


src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
```diff
@@ -94,7 +103,18 @@
     e: unknown,
     domainInput: string,
   ): string | undefined => {
+    if (e instanceof SsoNormalLoginRequiredError) {
+      // A gated dapp whose org uses a non-`sub` stable identifier (e.g. Entra
+      // `oid`): the first gated login can't be bridged to an existing identity
+      // until the user has signed in normally once (§6.5).
+      return $t`Sign in to ${domainInput} normally once before using this app, then try again.`;
```
sea-snake: The user should be guided to sign-in normally once, there's no need to ask them to do it themselves, else they'll be confused and not understand what "normally" even means.

src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -0,0 +1,290 @@
...
+        let len_end = self
+            .pos
+            .checked_add(8)
+            .ok_or(SsoBundleDecodeError::Truncated)?;
+        let len_bytes = self
+            .bytes
+            .get(self.pos..len_end)
+            .ok_or(SsoBundleDecodeError::Truncated)?;
+        let len = u64::from_be_bytes(len_bytes.try_into().expect("slice is 8 bytes")) as usize;
+        let data_end = len_end.checked_add(len).ok_or(SsoBundleDecodeError::Truncated)?;
```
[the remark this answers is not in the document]
sea-snake: Good catch — fixed in `6d71034fb`. `read_field` now does a checked `usize::try_from` on the length prefix and returns `Truncated` on overflow, so on wasm32 (32-bit `usize`) an oversized length fails closed instead of decoding a short read. Added a unit test with a `> u32::MAX` prefix.

src/internet_identity/src/openid.rs
```diff
@@ -506,6 +527,432 @@ fn salt() -> [u8; 32] {
...
+// --- SSO attribute bundle: thin `msg_caller_info` wrappers + consumer -------
+
+/// Copy this call's `sender_info` data payload (the attribute bundle bytes the
+/// caller attached). Empty when no `sender_info` was attached.
+fn msg_caller_info_data() -> Vec<u8> {
+    let n = ic0::msg_caller_info_data_size();
+    let mut b = vec![0u8; n];
+    ic0::msg_caller_info_data_copy(&mut b, 0);
+    b
+}
```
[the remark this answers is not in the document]
sea-snake: Fixed in `6d71034fb`: added a `MAX_SSO_BUNDLE_BYTES` (4096) cap checked before allocating, so an oversized `sender_info` is treated as no bundle. In practice this allocation was already gated behind `signer == ic_cdk::id()` (only II-signed payloads, which are small, reach it), so this is defense-in-depth that makes the bound explicit/local rather than dependent on call ordering.

src/internet_identity/internet_identity.did
```diff
@@ -1670,8 +1692,17 @@ service : (opt InternetIdentityInit) -> {
...
+    // The optional second argument is the target dapp origin; when supplied,
+    // `resolved_client_id` reports the client that origin must use (`null` = denied).
     discover_sso : (text) -> ();
-    get_sso_discovery : (text) -> (SsoDiscoveryState) query;
+    get_sso_discovery : (text, opt FrontendHostname) -> (SsoDiscoveryState) query;
+
+    // SSO sign-in path. Mints a delegation only if the JWT's `aud` matches the
+    // client the target origin resolves to. `Pending` means the discovery/JWKS
+    // cache is cold — re-call to drive the fetch.
+    sso_prepare_delegation : (JWT, Salt, SessionKey, text, FrontendHostname) -> (variant { Ok : SsoPrepareDelegationResponse; Pending; Err : OpenIdDelegationError });
```
aterga: These input args are fairly complex, let's wrap them into a Request type


```suggestion
    sso_prepare_delegation : (SsoPrepareDelegationRequest) -> (variant { Ok : SsoPrepareDelegationResponse; Pending; Err : OpenIdDelegationError });
```

src/internet_identity/internet_identity.did
```diff
@@ -1670,8 +1692,17 @@ service : (opt InternetIdentityInit) -> {
...
+    sso_get_delegation : (JWT, Salt, SessionKey, Timestamp, text, FrontendHostname, blob) -> (variant { Ok : SsoGetDelegationResponse; Pending; Err : OpenIdDelegationError }) query;
```
aterga: ditto


```suggestion
    sso_get_delegation : (SsoGetDelegationRequest) -> (variant { Ok : SsoGetDelegationResponse; Pending; Err : OpenIdDelegationError }) query;
```

src/internet_identity/internet_identity.did
```diff
@@ -1670,8 +1692,17 @@ service : (opt InternetIdentityInit) -> {
     // `get_sso_discovery` (query) and, while it reads `Pending`, drives the
     // on-demand two-hop discovery fetch with `discover_sso` (update); once the
     // fetch completes the query returns `Resolved` with the config.
+    //
+    // The optional second argument is the target dapp origin; when supplied,
+    // `resolved_client_id` reports the client that origin must use (`null` = denied).
     discover_sso : (text) -> ();
-    get_sso_discovery : (text) -> (SsoDiscoveryState) query;
+    get_sso_discovery : (text, opt FrontendHostname) -> (SsoDiscoveryState) query;
```
aterga: Let's commit to making the code self-explanatory. 


```suggestion
    type GetSsoDiscoveryRequest = record {
        org_domain : text;
        target_app_origin : opt FrontendHostname;
    };
    get_sso_discovery : (GetSsoDiscoveryRequest) -> (SsoDiscoveryState) query;
```

aterga: Side note

a better change for this function would be 


```suggestion
    get_sso_discovery_status : (GetSsoDiscoveryStatusRequest) -> (SsoDiscoveryStatus) query;
```

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
```diff
@@ -216,12 +216,29 @@ pub fn verify_openid_for_registration(
...
 
     openid::prefetch_sso(discovery_domain.as_deref());
-    let openid_credential = match openid::verify_jwt(jwt, salt, discovery_domain.as_deref())? {
-        openid::Cached::Ready(credential) => credential,
-        openid::Cached::Pending => return Ok(openid::Cached::Pending),
+    // Both discovery domain and origin present means a gated SSO login: verify
+    // through the SSO gate and store the primary-keyed credential.
+    let openid_credential = match (discovery_domain.as_deref(), origin.as_deref()) {
+        (Some(discovery_domain), Some(origin)) => {
+            match openid::verify_sso_for_registration(jwt, salt, discovery_domain, origin)? {
```
aterga: It's a bit strange that instead of factoring out the common code between `verify_openid_for_registration` and `verify_sso_for_registration` into a separate 3rd function, we first always call `verify_openid_for_registration` and sometimes digress into `verify_sso_for_registration`.

I recommend refactoring the code to follow the more predictable pattern (common code as a separate private function; bubble up the branching decision, e.g., `Some(discovery_domain), Some(origin)`, to the caller).

src/internet_identity/src/openid/sso.rs
```diff
@@ -18,8 +18,16 @@ use super::OpenIDJWTVerificationError;
 use crate::openid::AudClaim;
 use crate::single_flight_cache::{self, CacheConfig, Cached, RetryBackoff, SingleFlightCache};
 use identity_jose::jwk::Jwk;
+use sha2::{Digest, Sha256};
 use std::cell::RefCell;
 
+pub(super) const DEFAULT_STABLE_IDENTIFIER_CLAIM: &str = "sub";
+
+pub(super) const MAX_APP_CLIENTS: usize = 100;
```
aterga, line 42: Constants are exactly the code that MUST be documented.

src/internet_identity/src/openid/sso.rs
```diff
@@ -18,8 +18,16 @@ use super::OpenIDJWTVerificationError;
 use crate::openid::AudClaim;
 use crate::single_flight_cache::{self, CacheConfig, Cached, RetryBackoff, SingleFlightCache};
 use identity_jose::jwk::Jwk;
+use sha2::{Digest, Sha256};
 use std::cell::RefCell;
 
+pub(super) const DEFAULT_STABLE_IDENTIFIER_CLAIM: &str = "sub";
```
aterga, line 30: [Ditto](https://github.com/dfinity/internet-identity/pull/4096/changes#r3620480249)

src/internet_identity/src/openid/sso.rs
```diff
@@ -98,11 +106,118 @@ const MAX_DOMAIN_LENGTH: usize = 255;
...
+    /// Origin unlisted and `gate_all_apps` off — served by the primary client.
+    Primary(String),
+    /// Origin unlisted and `gate_all_apps` on — denied.
+    NotAllowed,
+}
+
+/// Parse the raw `app_clients` map, rejecting (never truncating) a map over
+/// `MAX_APP_CLIENTS` — truncation could silently drop a gated origin into the
+/// open fallback.
+pub(super) fn parse_app_clients(
```
aterga: Parse isn't the right verb here, are we sanitizing or validating or what?

aterga: This function smells like it should be a TryFrom

src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -0,0 +1,361 @@
...
+use super::{calculate_delegation_seed, SSO_ATTR_BUNDLE_TTL_NS};
+use crate::{state, update_root_hash};
+use candid::Principal;
+use ic_canister_sig_creation::signature_map::CanisterSigInputs;
+use ic_cdk::api::time;
+use internet_identity_interface::internet_identity::types::openid::OpenIdDelegationError;
+use internet_identity_interface::internet_identity::types::{AnchorNumber, Timestamp};
+
+/// Domain separator + version for the SSO attribute bundle.
+const SSO_ATTR_BUNDLE_DOMAIN: &[u8] = b"ii-sso-attr-v1";
```
aterga: Why the v1?

src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -0,0 +1,361 @@
...
+    let n = ic0::msg_caller_info_signer_size();
+    if n == 0 {
+        return None;
+    }
+    let mut b = vec![0u8; n];
+    ic0::msg_caller_info_signer_copy(&mut b, 0);
+    Some(Principal::try_from(b.as_slice()).expect("valid principal"))
+}
+
+const MAX_SSO_BUNDLE_BYTES: usize = 4096;
```
aterga: Ditto (doc string missing)

aterga: Also, remember these rules FOREVER, 

you should bubble up the constant declarations to the top of the file

src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -0,0 +1,361 @@
...
+
+/// The certified SSO attribute bundle attached to this call, if signed by this
+/// canister and not expired.
+///
+/// A `Some` result only proves II signed the bundle under the caller's credential
+/// seed; the caller MUST still check `bundle.origin` matches the serving origin —
+/// the bundle is bound to the identity, not the origin.
+pub fn read_certified_sso_bundle() -> Option<SsoAttrBundle> {
+    if msg_caller_info_signer()? != ic_cdk::id() {
```
aterga, line 161: This could be checked by the protocol, right?

src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -0,0 +1,361 @@
...
+    if time() > bundle.expiry_ns {
+        return None;
+    }
+    Some(bundle)
+}
+
+/// Prepare the certified SSO attribute bundle: encode `(sso_domain, origin,
+/// expiry)` and sign it under the credential seed. Returns the message bytes and
+/// its expiry; [`get_sso_attr_bundle_signature`] witnesses the signature.
+pub fn prepare_sso_attr_bundle(
```
aterga, line 178: The doc string deserves to explain the difference between this and regular OpenID attribute bundles

src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -0,0 +1,361 @@
...
+) -> (Vec<u8>, Timestamp) {
+    let expiration = time().saturating_add(SSO_ATTR_BUNDLE_TTL_NS);
+    let message = encode_sso_attr_bundle(sso_domain, origin, expiration);
+    let seed = calculate_delegation_seed(
+        &(iss.to_string(), sub.to_string(), aud.to_string()),
+        anchor_number,
+    );
+    state::signature_map_mut(|sigs| {
+        sigs.add_signature(&CanisterSigInputs {
+            domain: crate::attributes::ICRC3_ATTRIBUTES_CERTIFICATION_DOMAIN,
```
aterga: Prefer import to qualified:


```suggestion
            domain: ICRC3_ATTRIBUTES_CERTIFICATION_DOMAIN,
```

aterga: ^ Patch your memory

aterga: ^ Ditto elsewhere

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,437 @@
+//! SSO per-app gating and primary-identity resolution: verify an SSO JWT
```
aterga: Long sentence, but I still don't grasp (from reading it) what the problem is that we're solving here. 

Each file should have a doc string that says, e.g., 

* There is a problem. 
* This file solves that problem by A, B, ...

^ Patch your memory 
^ Ditto elsewhere

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,437 @@
...
+use internet_identity_interface::internet_identity::types::openid::OpenIdDelegationError;
+use internet_identity_interface::internet_identity::types::{AnchorNumber, IdRegFinishError};
+
+/// Maximum length of the configured stable-identifier claim value.
+const MAX_STABLE_IDENTIFIER_LENGTH: usize = 255;
+
+/// A verified SSO login. `credential.aud` is the resolved (per-app or primary)
+/// client; `primary_client_id` is always the primary, on which identity is keyed.
+#[derive(Debug)]
+pub struct SsoVerification {
```
aterga: Bad type name

Verification is a process. Whereas, this is a representation of an SSO session.

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,437 @@
...
+/// drive it via [`super::prefetch_sso`], then retry.
+pub fn verify_sso_jwt(
+    jwt: &str,
+    salt: &[u8; 32],
+    discovery_domain: &str,
+    origin: &str,
+) -> Result<Cached<SsoVerification>, OpenIDJWTVerificationError> {
+    sso::validate_allowed_discovery_domain(discovery_domain)
+        .map_err(OpenIDJWTVerificationError::GenericError)?;
+    let cfg = match sso::peek_discovery(discovery_domain) {
```
aterga: What is this? Control Flow Graph? (Be more verbose with code-level names)

^ PYM (Patch Your Memory)
^ DE (Ditto Elsewhere)

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,437 @@
...
+        Cached::Pending => return Ok(Cached::Pending),
+        Cached::Ready(cfg) => cfg,
+    };
+
+    let (expected_client, gated) = match cfg.resolve_client_for_origin(origin) {
+        sso::ClientResolution::PerApp(client) => (client, true),
+        sso::ClientResolution::Primary(client) => (client, false),
```
aterga: While I kind (kind of) understand `ClientResolution::PerApp`, what in the world is `ClientResolution::Primary`?

aterga: Use better type names!

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,437 @@
...
+) -> Result<Cached<SsoVerification>, OpenIDJWTVerificationError> {
+    sso::validate_allowed_discovery_domain(discovery_domain)
+        .map_err(OpenIDJWTVerificationError::GenericError)?;
+    let cfg = match sso::peek_discovery(discovery_domain) {
+        Cached::Pending => return Ok(Cached::Pending),
+        Cached::Ready(cfg) => cfg,
+    };
+
+    let (expected_client, gated) = match cfg.resolve_client_for_origin(origin) {
+        sso::ClientResolution::PerApp(client) => (client, true),
```
aterga:
```suggestion
        sso::ClientGating::Gated(client) => (client, true),
```

src/internet_identity/src/openid/sso_gating.rs
aterga:
```suggestion
        sso::ClientResolution::Ungated(client) => (client, false),
```

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,437 @@
...
+        sso::ClientResolution::Primary(client) => (client, false),
+        sso::ClientResolution::NotAllowed => {
```
aterga:
```suggestion
        sso::ClientResolution::Forbidden => {
```

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,437 @@
...
+pub struct SsoPrimaryIdentity {
+    /// Primary-keyed credential to store: `aud` is the primary client. For a
+    /// non-`sub` org it carries `stable_id = Some(oid)` so the anchor `write()`
+    /// (re)establishes the SSO stable-id index entry; the gated resolve below
+    /// sets the same value so re-writing the primary credential leaves the
+    /// index unchanged.
+    pub credential: OpenIdCredential,
+}
+
+/// Resolve a verified SSO login to the primary identity. Reads only (safe from a
```
aterga: IT is not clear what primary means here - was that terminology established before?

src/internet_identity/src/openid/verify.rs
```diff
@@ -218,6 +218,9 @@ pub(super) fn verify_and_build(
         metadata,
         sso_domain,
         sso_name,
+        // Set later at SSO write time for a non-`sub` primary credential; a
+        // freshly verified token never carries it.
+        stable_id: None,
```
aterga: Explain in the comment why this field is needed

aterga (changes requested):
Some questions

> Behavior change (deliberate): sso:<domain> attributes now require an SSO session (previously certified from the stored credential in any session).

Are there integration / e2e tests that show the old behavior that is not allowed anymore?

> Bounded reconfiguration window: flipping an app ungated → gated leaves already-minted bundles valid until expiry (~10 min) plus the ~1 h discovery-cache refresh.

Do we have integration / e2e tests that cover flipping this flag? 

Also interested in integration / e2e tests that show app assignment followed by revocation.

src/internet_identity/src/openid/sso_bundle.rs
sea-snake: No, the protocol only checks the signature, it doesn't enforce where the signature comes from. That's why the certified ICRC-3 attributes lib requires specifying the signer principal (II). In this PR, the bundle is coming from II itself and consumed by II itself, so it does a comparison with itself (`id()`).

src/internet_identity/src/openid/sso_bundle.rs
sea-snake: There are no regular OpenID attribute bundles, there's only the SSO bundle (that's for internal use) and the ICRC-3 certified attributes bundle with any arbitrary attributes incl. unscoped, openid scope, sso scope etc (that's for external use).

src/internet_identity/src/openid/sso_gating.rs
sea-snake: "Ungated" isn't a correct term either, since we're referring to the client for a given origin here _not_ if things are ungated or not, that's currently the second boolean in the tuple.

In practice it's possible to get the primary client resolution even though you're not passing the primary origin (app isn't listed and unknown apps aren't restricted). You'd only get the not allowed variant if the app isn't listed and unknown apps are restricted. 

The app gating itself is supposed on the IDP side, on the II side, it's more about client resolution only. So maybe the whole gating and PerApp/Primary enum should not be exposed, it should maybe just return a result with either a client of forbidden error instead since the two different enum variants don't even seem to be used in the first place.

src/frontend/src/lib/components/wizards/auth/views/SignInWithSso.svelte
sea-snake: Addressed — the first-sign-in dialog drives the one normal sign-in automatically; the user never has to trigger it themselves or see the word "normally".

src/internet_identity/internet_identity.did
sea-snake: Wrapping the args into a request record and renaming to `get_sso_discovery_status` → `SsoDiscoveryStatus`. We're also removing the SSO discovery allowlist entirely (prod already runs `sso_allow_any_domain`), which drops the `NotAllowed` variant — so `State` → `Status` fits cleanly.

src/internet_identity/internet_identity.did
sea-snake: Agreed — wrapping into `SsoPrepareDelegationRequest`.

src/internet_identity/internet_identity.did
sea-snake: Agreed — `SsoGetDelegationRequest`.

src/internet_identity/src/anchor_management/registration/registration_flow_v2.rs
sea-snake: Agreed — and it lines up with authentication, where SSO and direct are already separate top-level methods (`sso_prepare_delegation`/`openid_prepare_delegation`, `sso_get_delegation`/`openid_get_delegation`), each calling its own verify. Registration is the odd one out with the internal digression. I'll bubble the `(Some(domain), Some(origin))` branch into `openid_identity_registration_finish`, so `verify_openid_for_registration` becomes the direct path (matching its name) and `verify_sso_for_registration` the SSO path — same shape as auth.

src/internet_identity/src/openid/sso.rs
sea-snake: Will document: it's the default for the well-known's optional `stable_identifier_claim`. `sub` is the OIDC subject; when an org's `sub` is the same across its OIDC clients II keys identity on it directly (no bridge), while an org with a pairwise `sub` (per-client, e.g. Entra) overrides it with a cross-client-stable claim like `oid`.

src/internet_identity/src/openid/sso.rs
sea-snake: Will document: max `app_clients` entries from an org's well-known. The parsed map is retained in the in-memory SSO discovery cache (one per cached domain), so this primarily bounds heap footprint there; an over-cap map is rejected, not truncated.

src/internet_identity/src/openid/sso.rs
sea-snake: Agreed "parse" is wrong — it validates the raw map (entry-count + client_id caps) and builds the typed `Vec<AppClient>`, so I'll rename it `validate_app_clients`. I'd skip `TryFrom` here though: the source is a plain `HashMap<String, String>`, so `TryFrom<&HashMap<String, String>>` would read as "any string map converts to app-clients," which is too broad — there's no distinct raw type to hang the conversion on.

src/internet_identity/src/openid/sso_bundle.rs
sea-snake: Good point — it isn't load-bearing (II signs and consumes it, upgrades atomically, and the bundle is short-lived), so I'll drop the version: `ii-sso-attr`.

src/internet_identity/src/openid/sso_bundle.rs
sea-snake: Will document it and move it up with the other constants.

src/internet_identity/src/openid/sso_bundle.rs
sea-snake: Will import it and use the short name; swept the other qualified refs too.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Restructured into problem → solution: the problem is that a gated app signs in against its own per-app client, so the token's identity is scoped to that client (pairwise `sub` for Entra), not the primary client the anchor is keyed on — a gated login can't match the anchor directly. This file solves it by (a) verifying against the origin's resolved client (the gate) and (b) bridging to the primary identity via the stable-id index.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Agreed it's a representation, not the process. Renaming to `VerifiedSsoLogin` (matches the doc). I'd steer clear of `SsoSession` since "session" already refers to the session key in this flow.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: That's the discovery config — renaming `cfg` → `discovery_config` here and in the other spots it appears.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Collapsing to `resolve_client_for_origin(origin) -> Result<String, Forbidden>` — just the client id to verify against, or `Forbidden`. The per-app-vs-primary bit doesn't need storing: it's `credential.aud != primary_client_id`, so `VerifiedSsoLogin` drops the `gated` field and `resolve_primary_identity` derives it via an `is_gated()` helper. Removes the confusing variants and the redundant bool.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: See the collapse in the `Primary` thread — `resolve_client_for_origin` now returns `Result<String, Forbidden>`, so there's no `Gated`/`PerApp` variant to rename.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: See the collapse in the `Primary` thread — the denial becomes the `Forbidden` error variant of the `Result`.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: "Primary client" = the org's main OIDC client that the user's anchor is keyed on, as opposed to per-app clients. I'm defining it explicitly in the module doc (the file-level rewrite above), so the term is established before it's used.

src/internet_identity/src/openid/verify.rs
sea-snake: The field's "why" is documented at its definition in `openid.rs` — it seeds the SSO stable-id index `(iss, aud, stable_id) → anchor` that bridges per-app logins to the primary identity (`None` for `sub` orgs / direct providers). The construction site you flagged here was reworked in 25d05ca46, where the stable id is now built into the credential with a comment.

src/internet_identity/src/main.rs
```diff
@@ -1676,20 +1684,158 @@ mod openid_api {
...
+                &discovery_domain,
+                &origin,
+            );
+
+            // The association could change during the `.await`.
+            let still_anchor_number =
+                state::storage_borrow(|storage| storage.lookup_anchor_with_openid_credential(&key))
+                    .ok_or(OpenIdDelegationError::NoSuchAnchor)?;
+            if anchor_number != still_anchor_number {
+                return Err(OpenIdDelegationError::NoSuchAnchor);
```
aterga, line 1773: error case seems misleading 


```suggestion
                return Err(OpenIdDelegationError::CredentialAnchorAssociationChanged);
```

src/internet_identity/src/main.rs
```diff
@@ -1676,20 +1684,158 @@ mod openid_api {
...
+                anchor_number,
+                sso_attr_bundle: ByteBuf::from(sso_attr_bundle),
+            })
+        }
+        .await;
+
+        match prepared {
+            Ok(response) => OpenIdResult::Ok(response),
+            Err(err) => OpenIdResult::Err(err),
+        }
```
aterga:
```suggestion
        prepared.into()
```

src/internet_identity/src/main.rs
```diff
@@ -1676,20 +1684,158 @@ mod openid_api {
...
+            &identity.credential.aud,
+            anchor_number,
+            &sso_attr_bundle,
+        ) {
+            Ok(signature) => signature,
+            Err(err) => return OpenIdResult::Err(err),
+        };
+        OpenIdResult::Ok(SsoGetDelegationResponse {
+            signed_delegation,
+            sso_attr_bundle_signature: ByteBuf::from(sso_attr_bundle_signature),
```
aterga:
```rust
let sso_attr_bundle_signature = ByteBuf::from(sso_attr_bundle_signature);
```

and then 


```suggestion
            sso_attr_bundle_signature,
```

aterga: Nit obviously

src/internet_identity/src/openid.rs
```diff
@@ -506,12 +531,17 @@ fn salt() -> [u8; 32] {
     [0; 32]
 }
 
+/// How long a certified SSO attribute bundle is accepted after it is minted
+/// (`sso_bundle.rs`) — the bundle acceptance window, not a session or delegation
+/// lifetime.
+pub const SSO_ATTR_BUNDLE_TTL_NS: u64 = 10 * MINUTE_NS;
```
aterga: why not 5?

src/internet_identity/Cargo.toml
```diff
@@ -49,6 +49,9 @@ futures = { version = "0.3", default-features = false, features = ["alloc"] }
 
 # All IC deps
 candid.workspace = true
+# Separate `ic0 1.1` for the `msg_caller_info_*` system API; it has no `links`
+# key, so it coexists with the `ic0 0.23` pulled in transitively by `ic-cdk`.
+ic0 = "1.1"
```
aterga, line 54: Consider bumping ic-cdk instead.

aterga (approved):

src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -0,0 +1,369 @@
...
+use ic_canister_sig_creation::signature_map::CanisterSigInputs;
+use ic_cdk::api::time;
+use internet_identity_interface::internet_identity::types::openid::OpenIdDelegationError;
+use internet_identity_interface::internet_identity::types::{AnchorNumber, Timestamp};
+
+/// Domain separator for the SSO attribute bundle signature.
+const SSO_ATTR_BUNDLE_DOMAIN: &[u8] = b"ii-sso-attr";
+
+/// Max accepted size of an attached bundle payload (`sender_info`). A real
+/// bundle is a few hundred bytes; anything larger is treated as no bundle,
```
aterga, line 24: > anything larger is treated as no bundle

have you considered encoding a special error bundle? 

what would break, and what would be the UX, if a bundle is dropped silently?

src/internet_identity/src/openid/sso_gating.rs
aterga: Why not call it the II client?

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,444 @@
...
+        let anchor_number = state::storage_borrow(|storage| {
+            storage.lookup_anchor_by_sso_stable_id(
+                sso_domain,
+                &verification.credential.iss,
+                &verification.primary_client_id,
+                &stable_id,
+            )
+        })
+        .ok_or(OpenIdDelegationError::NoSuchAnchor)?;
+        primary_sub_on_anchor(
```
aterga: it's unclear from the name of this function what it does

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,444 @@
...
+    };
+    Ok(SsoPrimaryIdentity { credential })
+}
+
+/// Load `anchor_number` and return the `sub` of its primary-keyed OpenID
+/// credential for `(iss, primary_client_id)`, if present. `None` when the
+/// anchor is gone or no longer carries that credential — the index pointed
+/// here but the credential has since been removed, so the gated login fails
+/// safe.
+fn primary_sub_on_anchor(
```
aterga:
```suggestion
fn lookup_iss_and_primary_client_id_by_anchor_number(
```

Or consider inlining this function.

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -0,0 +1,444 @@
...
+/// store — a first gated login registers under the primary client, never a
+/// per-app one. Fails safe: a non-`sub` per-app token with no bridge yet returns
+/// an error and stores nothing.
+pub fn verify_sso_for_registration(
+    jwt: &str,
+    salt: &[u8; 32],
+    discovery_domain: &str,
+    origin: &str,
+) -> Result<Cached<OpenIdCredential>, IdRegFinishError> {
+    let verification = match verify_sso_jwt(jwt, salt, discovery_domain, origin) {
```
aterga:
```suggestion
    let verification_outcome = match verify_sso_jwt(jwt, salt, discovery_domain, origin) {
```

src/internet_identity/src/openid/verify.rs
```diff
@@ -98,20 +98,32 @@ pub(super) struct Descriptor {
...
+pub(super) struct SsoProvider {
+    pub domain: String,
+    pub name: Option<String>,
+    /// Claim holding the cross-client-stable id (e.g. Entra `oid`). `None` for a
+    /// `sub` org, which keys identity on the token `sub` and needs no bridge.
+    pub stable_identifier_claim: Option<String>,
+}
+
+/// Maximum length of a configured stable-identifier claim value.
+const MAX_STABLE_IDENTIFIER_LENGTH: usize = 255;
```
aterga: In bytes??


```suggestion
const MAX_STABLE_IDENTIFIER_BYTES: usize = 255;
```

src/internet_identity/src/openid/verify.rs
aterga:
```suggestion
    let jwt_under_validation = Decoder::new()
```

src/internet_identity/src/attributes.rs
```diff
@@ -576,6 +590,13 @@ impl Anchor {
                 continue;
             };
 
+            // List `sso:<domain>` rows only for a session that can certify that exact domain.
+            if let AttributeScope::Sso { domain } = &scope {
+                if sso_session_domain.as_deref() != Some(domain) {
+                    continue;
```
aterga, line 601: Ummm, maybe in this case we should error out instead of just ignoring this very strange situation.

src/internet_identity/src/openid.rs
sea-snake: Good call — shortened to 5 min.

5 minutes comfortably covers the post-prepare ceremony (IdP 2FA/redirect, `sso_get_delegation`, then the dapp's attribute call). My initial instinct was to keep an extra ~5 min of slack for client clock skew, but that reasoning does not apply here: the expiry is both minted (`time().saturating_add(TTL)`) and checked (`if time() > bundle.expiry_ns`) entirely by II using IC time — the client's system clock never enters the comparison. So there is nothing to budget skew for, and the window can be tight.

src/internet_identity/Cargo.toml
sea-snake: Kept `ic0 = "1.1"` deliberately rather than bumping `ic-cdk`. The only thing this crate needs from the newer system API is `msg_caller_info_signer`/`_data`, which `ic0 1.1` exposes directly. Bumping `ic-cdk` to a version that ships those instead pulls a transitive set that conflicts with the git-pinned vc-sdk / identity.rs deps this crate depends on, so it turns into a much larger and riskier dependency untangle for no additional benefit. `ic0` has no `links` key, so it coexists cleanly with the `ic0 0.23` that `ic-cdk` already pulls transitively (as noted in the comment above the dep). Happy to revisit the `ic-cdk` bump as its own change if/when the vc-sdk deps move.

src/internet_identity/src/openid/sso_bundle.rs
sea-snake: Considered it, but kept the silent drop. The behaviour is fail-safe: an oversized (or otherwise unparseable) `sender_info` is treated as "no bundle", so no `sso:<domain>` attributes are certified for that session. A gated dapp then sees an ungated session and refuses, and the user retries — no security hole, no wrong-identity outcome, just a sign-in that has to be re-attempted.

In practice this path is essentially unreachable: a real bundle is a few hundred bytes against the 4096-byte cap. The cap exists only to bound the decode allocation against a hostile caller, not to reject legitimate bundles. An explicit error bundle would let us distinguish *malformed* from *absent* for a slightly nicer diagnostic, but it adds encode/decode surface and a new failure mode to reason about for a case that should never occur legitimately — so it does not seem worth the added complexity here.

src/internet_identity/src/attributes.rs
sea-snake: This is intended behaviour rather than a strange situation, so I kept the `continue`. Two reasons:

1. It is the normal per-domain filter. An anchor can hold SSO credentials for multiple providers, but a single session certifies exactly one domain, so we list only the rows for that session's domain and skip the others. Erroring here would break listing for any anchor with more than one SSO credential.

2. More fundamentally, the list-attributes API is best-effort and non-blocking by design. Its job is to list what is *available*; it is intended that attributes can be requested but not available, in which case they are simply not listed and the app does not receive them. The attributes flow deliberately has no error paths — it must never block the sign-in flow. Erroring out here would contradict that design.

I added a short comment making the "list only what this session can certify" intent explicit at the call site.

src/internet_identity/internet_identity.did
sea-snake: Done — the delegation args are wrapped in `SsoPrepareDelegationRequest` / `SsoGetDelegationRequest` (and the discovery query in `GetSsoDiscoveryStatusRequest`).

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Renamed throughout (7635c4d81). The client the anchor is keyed on is Internet Identity’s own OIDC client registered at the org’s IdP — as opposed to the per-app clients — so “primary” is now **II client**: `primary_client_id`→`ii_client_id`, `resolve_primary_identity`→`resolve_ii_client_identity`, `SsoPrimaryIdentity`→`SsoResolvedIdentity`, `primary_sub_on_anchor`→`ii_client_sub_on_anchor`. The module doc now introduces the term explicitly.

src/internet_identity/src/main.rs
sea-snake: Done (7635c4d81) — added `impl<T, E> From<Result<T, E>> for OpenIdResult<T, E>`, and the site is now `prepared.into()`.

src/internet_identity/src/main.rs
sea-snake: Done (7635c4d81) — hoisted into a `let` binding and used field shorthand.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Done (7635c4d81) — renamed to `verification_outcome`.

src/internet_identity/src/openid/verify.rs
sea-snake: Good catch — bytes. Renamed to `MAX_STABLE_IDENTIFIER_BYTES` (7635c4d81); the cap is enforced with `.len()`, which is the UTF-8 byte length.

src/internet_identity/src/openid/verify.rs
sea-snake: Done (7635c4d81) — renamed to `jwt_under_validation`.

src/internet_identity/src/main.rs
sea-snake: Kept the coarse `NoSuchAnchor` here and added a comment explaining why, rather than introducing a new variant (7635c4d81).

You’re right that it reads as misleading — the anchor exists, the credential just re-associated to a different one during the `.await`. But I opted not to surface that distinction on the wire: the frontend has no distinct recovery for “re-associated” vs “gone” (both collapse to the same retry / re-auth outcome), so a dedicated variant would only leak a backend state-transition signal to the caller for no client benefit. The comment documents the deliberate coarsening. Happy to add a `CredentialAnchorAssociationChanged` variant instead if you’d rather have the precision on the wire.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Agreed — renamed to `ii_client_sub_on_anchor` (7635c4d81), which now says what it returns: the `sub` of the given anchor’s II-client credential. Part of the broader `primary`→`II client` rename.

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Renamed, but not to the suggested name. `lookup_iss_and_primary_client_id_by_anchor_number` reads as “look up the iss + client id by anchor”, but the function does the opposite: it takes `(anchor_number, iss, ii_client_id)` and returns that anchor’s II-client `sub` (failing safe if the credential has since been removed). So I went with `ii_client_sub_on_anchor` (7635c4d81). I kept it as a named helper rather than inlining — it’s a distinct read-with-fail-safe step, and the name + doc carry the “the index pointed here but the credential may be gone” semantics that are easy to miss when inlined.

PR #4100, opened 2026-07-12
feat(fe): replace read-only checkbox with a permissions radio group

aterga:
# Motivation

On the sign-in screens the access level was offered as a single
"Read-only mode" / "Full access" checkbox. In practice nobody noticed or
understood it, so it was left at whatever it defaulted to — the choice was
effectively invisible.

This refines the UI so the access level is a deliberate, understandable
choice, and remembers what the user picked so they don't have to re-decide
every time.

# Changes

- **Radio group instead of a checkbox.** A new `AccessLevelSelector`
  component replaces `AccessLevelToggle`, offering two explicit options:
  - **Queries only** — read your data without making changes on your behalf (query-only delegation).
  - **Actions & queries** — read your data and make changes on your behalf (full access).
- **No pre-selection on a first-time sign in.** Both options start
  unselected and the confirm/continue action is disabled until the user
  picks one, so the access level is chosen on purpose rather than inherited
  silently.
- **Per-flow persistence.** Once chosen, the level is stored in browser
  storage (`ii:access-level:{continue,cli,mcp}`) and pre-fills the next
  connect. Preferences are kept **separate** across the normal `/authorize`
  continue flow, the `/cli` flow, and the `/mcp` flow, via
  `read/writeAccessLevelPreference` in `accessLevel.ts`.
- **Gating unchanged.** `/continue` and `/cli` still show the selector only
  when the `READ_ONLY_MODE` feature flag is on (off by default); `/mcp`
  keeps offering it ungated, where the choice applies to the per-app
  delegations the server later obtains.
- Removed the now-unused toggle-mapping helpers and updated a couple of
  stale comments referring to the old checkbox.

# Tests

- **Unit** (`accessLevel.test.ts`): preference round-trip, first-time
  returns `undefined`, per-flow separation, and rejection of unrecognized
  stored values. Existing `toPermissionsArg` coverage retained.
- **e2e** (`mcp.spec.ts`, `cli.spec.ts`, `authorize/index.spec.ts`):
  updated the connect flows for the radio UI (choose a level, then
  connect), assert "Allow access" is disabled until a level is chosen, and
  added a "Remembers the access-level choice for the next connect" test
  that verifies the stored preference pre-fills a second connect. The
  flag-off flows assert the selector is absent.
- Local checks green: `svelte-check` (0 errors), eslint, prettier, and the
  `accessLevel` unit suite.

https://claude.ai/code/session_01N4S19riABjX8CF6JhTuQ1L

---
_Generated by [Claude Code](https://claude.ai/code/session_01N4S19riABjX8CF6JhTuQ1L)_


src/frontend/src/lib/utils/accessLevel.ts
```diff
@@ -19,18 +19,44 @@ export const toPermissionsArg = (accessLevel: AccessLevel): [Permissions] => [
...
+ *  time they use this flow. */
+export const writeAccessLevelPreference = (
+  flow: AccessLevelFlow,
+  accessLevel: AccessLevel,
+): void => {
+  if (typeof localStorage === "undefined") {
+    return;
+  }
+  localStorage.setItem(`${ACCESS_LEVEL_STORAGE_PREFIX}${flow}`, accessLevel);
+};
```
sea-snake: This implementation writes the setting for all anchor numbers instead of per anchor number, particularly on a public system, this behavior leads to one user being able to change the defaults for another.

The current pattern for per-anchor stored auxiliary state can be found here for example here`src/frontend/src/lib/stores/cli-access.store.ts`.

sea-snake (changes requested):

src/frontend/src/lib/utils/accessLevel.ts
aterga: Good catch — fixed in e071b41. The preference is now keyed per anchor as well as per flow (`ii:access-level:{flow}:{anchorNumber}`), so one anchor's choice can no longer become another's default on a shared browser. The selector re-hydrates from the selected anchor's stored value when the identity is switched, following the same per-anchor pattern as the multi-accounts toggle and `cli-access.store`. Added a unit test covering cross-anchor separation.

---

sea-snake: Why did you not exactly follow the `cli-access.store.ts` pattern I shared? I'm still looking at a util instead of store file and it's still writing to local storage directly instead of using the writable storage.

sea-snake (changes requested):

src/frontend/src/lib/utils/accessLevel.ts
aterga: You're right — done properly now in 9ff1604. The persistence lives in a dedicated store, `src/frontend/src/lib/stores/access-level.store.ts`, backed by `writableStored` (new `storeLocalStorageKey.AccessLevel`, versioned), with `getPreference`/`setPreference` over a per-flow, per-anchor map — same shape as `last-shared-emails.store` / `cli-access.store`. No more direct `localStorage` access. `accessLevel.ts` is back to just the `AccessLevel` type and `toPermissionsArg`, and the persistence tests moved to `access-level.store.test.ts`. Sorry for the half-measure on the first pass.

---

sea-snake (approved):

PR #4101, opened 2026-07-12
fix(fe): persist newly-minted identity from the 1-click flow

aterga:
# Motivation

When a user signs up with II through the **1-click flow** (e.g. the "Continue with Google" button that a dapp such as caffeine.ai launches directly), the newly-minted identity was never written to local storage. So when the user later opened id.ai directly, they saw an empty identity list — their brand-new identity was invisible.

Root cause: the authorize page's OpenID/SSO resume handler (`resumeOpenId` in `authorize/+page.svelte`) built its `AuthFlow` with `trackLastUsed: false`, so neither a sign-up nor a sign-in in that flow ever recorded the identity in `ii-last-used-identities`.

# Changes

- **`flows/authFlow.svelte.ts`** — the resume flow redeems *both* direct-OpenID and SSO JWTs through `continueWithOpenId` / `completeOpenIdRegistration`, with SSO signalled only by a discovery domain. Made both the sign-in entry (`continueWithOpenId`) and the sign-up commit (`#registerWithOpenId`) record an **`sso`-typed** last-used entry (keyed by domain) when a discovery domain is present, and keep the existing `openid` entry otherwise.
  - This matters because re-authentication (`authLastUsedFlow`) dispatches on the stored `authMethod`: an `openid` entry is resolved against the static `openid_configs` via `findConfig(issuer)` — which cannot resolve an SSO issuer and passes no discovery domain — whereas an `sso` entry re-runs `discoverSsoConfig(domain)`. Storing an SSO credential as `openid` would silently break its later sign-in.
- **`authorize/+page.svelte`** — `resumeOpenId` now uses `new AuthFlow()` (tracking enabled) and commits the `pendingLastUsedEntry` the flow returns for the sign-in case. The write happens before `authorize()`/redirect, so it always lands.

# Tests

- **Unit** (`authFlow.svelte.test.ts`): added coverage that `continueWithOpenId` returns — and `completeOpenIdRegistration` commits — an `sso`-typed entry when a discovery domain is present and an `openid`-typed entry otherwise, for both sign-in and sign-up.
- **E2E** (`authorize/openid.spec.ts`, `authorize/sso.spec.ts`): regression guards that, after a 1-click sign-up, the identity is present in `ii-last-used-identities` on the II origin — tagged `openid` for direct OpenID and `sso` (with the discovery domain) for SSO.
- Local checks: `tsc` (full project), `svelte-check` (0 errors), eslint (0 warnings), prettier, and the `authFlow` unit suite all pass.

https://claude.ai/code/session_01KsFeaQQFuJeU1aw1k9yCoR

---
_Generated by [Claude Code](https://claude.ai/code/session_01KsFeaQQFuJeU1aw1k9yCoR)_


src/frontend/src/routes/(new-styling)/authorize/+page.svelte
```diff
@@ -259,9 +262,20 @@
...
       );
+    } else if (authFlowResult?.type === "signIn") {
+      // Sign-in returns the entry for the caller to commit (it defers the
+      // write so the interactive wizard can gate it behind disambiguation);
+      // here there is no disambiguation, so commit it right away.
+      if (authFlowResult.pendingLastUsedEntry !== undefined) {
+        lastUsedIdentitiesStore.addLastUsedIdentity(
+          authFlowResult.pendingLastUsedEntry,
+        );
+      }
```
sea-snake: Shouldn't the last used flow store the last used entry? Why is the page doing it here instead? That seems like an anti pattern considering that `trackLastUsed` is here specifically to manage this.

sea-snake (changes requested):

src/frontend/src/routes/(new-styling)/authorize/+page.svelte
aterga: Good call — moved it into `AuthFlow` (1bdc799). When `continueWithOpenId` is handed a JWT (the non-interactive 1-click resume flow), there's no method-switch disambiguation to gate the write, so it now records the sign-in entry itself and returns no `pendingLastUsedEntry`; the page no longer touches `addLastUsedIdentity`.

The one case that still returns the entry for the caller to commit is the interactive `AuthWizard` path (`existingJwt === undefined`), which has to defer the write until the method-switch dialog is confirmed/cancelled (`commitLastUsedEntry`). Left that as-is so the disambiguation gating keeps working. Unit tests now cover both the resume path (commits inside `AuthFlow`) and the wizard path (returns uncommitted).

---

src/frontend/tests/e2e-playwright/routes/authorize/openid.spec.ts
```diff
@@ -86,6 +86,133 @@ test.describe("Authorize with 1-click OpenID", () => {
...
+        await iiPage.close();
+      }
+    });
+  });
+
+  // The sign-up regression tests above only exercise the sign-UP branch of
+  // resumeOpenId (fresh OpenID users). This covers the sign-IN branch — a
+  // returning user who signs in via 1-click on a device with no local
+  // "last used" entry — which commits the identity through a separate code
+  // path (`pendingLastUsedEntry`) that a regression could break independently.
```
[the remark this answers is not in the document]
aterga: Fixed in ad2bb81 — updated this comment to say the JWT-supplied sign-in now commits inside `continueWithOpenId` (no `pendingLastUsedEntry`). Also addressed the two `JSON.parse(lastUsedRaw!)` comments by asserting `lastUsedRaw` is present first (matching the manage/access.spec.ts convention) across all three localStorage reads.

---

sea-snake (approved):

PR #4109, opened 2026-07-15
feat(be,fe): push notifications PoC — RFC 8291 + VAPID + notify_user

MRmarioruci:
II canisters have no way to reach users when they're offline or have the tab closed. This PR lands the backend for a Web Push notification pipeline so dApps can send encrypted, VAPID-signed pushes through Apple/Google/Mozilla's relays via a single `notify_user` call. Design was accepted in #4071; this is the implementation.

## Changes

- New `push` module: `rfc8291.rs` (RFC 8291 `aes128gcm` payload encryption), `vapid.rs` (ECDSA-P256-SHA256 JWT signer, lazy-generated stable-memory keypair), `api.rs` (subscribe / unsubscribe / grant_consent / revoke_consent / notify_user).
- Three new stable-memory regions (indices 31 / 32 / 33): `push_subscriptions_memory` `(anchor, endpoint_hash) → StorablePushSubscription`, `push_consent_memory` `(anchor, origin_hash) → StorablePushConsent`, `push_principal_index_memory` `Principal → StorablePushSenderInfo`.
- Six new Candid methods: `push_subscribe_device`, `push_unsubscribe_device`, `push_grant_consent`, `push_revoke_consent`, `notify_user`, `push_vapid_public_key`.
- New crypto deps: `hkdf`, `aes-gcm`; `p256` gains the `ecdh` feature.
- Frontend IDL regenerated from the `.did` change.

## VAPID key handling

The private key is a random P-256 scalar generated via `raw_rand` on the first push-related call after install/upgrade, rejection-sampled against the curve order, and persisted in `PersistentState` alongside the anchor salt. Same threat profile as II's existing long-lived secrets. When secp256r1 threshold ECDSA lands on IC the plan is to migrate to `sign_with_ecdsa` (~two-hour swap) so nothing is persisted at all.

**Deployment note (staging):** the first call after this rolls out generates a brand-new random VAPID key. Existing subscriptions on the previous hardcoded PoC key become invalid — browsers still have the old pubkey baked into their relay registration, so FCM/Mozilla will 401 every push until each device hits Disable→Enable in Settings once. Only affects PoC testers on staging D.

Design: #4071


MRmarioruci:
no longer relevant

PR #4113, opened 2026-07-16
Length-cap OpenID/SSO claim and discovery string values

sea-snake:
## What

Add byte-length caps (255) on the OpenID/SSO string values II stores and derives identities from — the JWT `iss` / `sub` claims and the SSO discovery fields (`client_id`, `jwks_uri`, `issuer`, `name`, `domain`). Defensive bounds on these caller-supplied values, matching the existing `email` / `name` claim caps.

| Field | Capped in |
|---|---|
| JWT `iss`, `sub` | `verify_claims` |
| discovery `client_id`, `jwks_uri`, `issuer`, `name` | `discovery_fill` |
| discovery `domain` | `is_allowed_discovery_domain` |

## Tests

`verify.rs`: `should_reject_over_long_sub` / `_iss`. `sso.rs`: `over_long_domain_is_never_allowed`. Integration: over-length `client_id` / `jwks_uri` / `issuer` / `name` each fail discovery; over-length domain reads `NotAllowed`.

No `.did` change; pure backend validation.

### DIFF
--- a/src/internet_identity/src/openid/sso.rs
+++ b/src/internet_identity/src/openid/sso.rs
@@ -76,9 +76,20 @@ const DISCOVERY_MAX_SCOPES: usize = 32;
 #[cfg(not(test))]
 const DEFAULT_SCOPES: [&str; 3] = ["openid", "profile", "email"];
 
-/// Untrusted well-known values (cache keys/values; `client_id` also feeds the
-/// seed). Bounded at 255, like `MAX_SEED_FIELD_LENGTH` in `verify.rs`.
-const MAX_DISCOVERY_FIELD_LENGTH: usize = 255;
+/// Maximum length of the client_id from the well-known.
+#[cfg(not(test))]
+const MAX_CLIENT_ID_LENGTH: usize = 255;
+/// Maximum length of the issuer from the OIDC configuration.
+#[cfg(not(test))]
+const MAX_ISSUER_LENGTH: usize = 255;
+/// Maximum length of the jwks_uri from the OIDC configuration.
+#[cfg(not(test))]
+const MAX_JWKS_URI_LENGTH: usize = 255;
+/// Maximum length of the display name from the well-known.
+#[cfg(not(test))]
+const MAX_SSO_NAME_LENGTH: usize = 255;
+/// Maximum length of an SSO discovery domain.
+const MAX_DOMAIN_LENGTH: usize = 255;
 
 /// The result of resolving a discovery domain (hop 1 + hop 2). Carries
 /// everything the redemption path needs (`issuer`, `jwks_uri`) and everything
@@ -254,11 +265,7 @@ fn is_explicitly_allowlisted(domain: &str) -> bool {
 }
 
 pub fn is_allowed_discovery_domain(domain: &str) -> bool {
-    // Reject an over-length domain before it can become a discovery cache key.
-    // This is the single chokepoint every path that inserts a domain key
-    // (`prefetch`, `drive_discovery`) gates on, so a >255-byte domain never
-    // reaches `single_flight_cache::get`. See `MAX_DISCOVERY_FIELD_LENGTH`.
-    if domain.len() > MAX_DISCOVERY_FIELD_LENGTH {
+    if domain.len() > MAX_DOMAIN_LENGTH {
         return false;
     }
     // An explicitly allowlisted domain is admin-curated and trusted verbatim.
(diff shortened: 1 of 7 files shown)


src/internet_identity/src/openid/verify.rs
```diff
@@ -43,6 +43,11 @@ const MAX_EMAIL_LENGTH: usize = 256;
 // already validate it on their end for a sane maximum length. This is an additional sanity check.
 const MAX_NAME_LENGTH: usize = 128;
 
+// Maximum length of the sub claim in the JWT. The subject is stored verbatim on
+// the credential and feeds the delegation seed, so bound it as a sanity check —
+// the identity provider is expected to keep it short.
+const MAX_SUB_LENGTH: usize = 256;
```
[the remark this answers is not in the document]
sea-snake: Fixed — capped at 255 (a byte limit, since `String::len()` counts bytes). Applied the same 255 bound to `iss` and the SSO discovery fields (`client_id`/`jwks_uri`/`issuer`/`name`/`domain`).

src/internet_identity/src/openid/verify.rs
```diff
@@ -43,6 +43,12 @@ const MAX_EMAIL_LENGTH: usize = 256;
 // already validate it on their end for a sane maximum length. This is an additional sanity check.
 const MAX_NAME_LENGTH: usize = 128;
 
+// Maximum length of the iss claim in the JWT; a sanity bound.
+const MAX_ISS_LENGTH: usize = 255;
```
MRmarioruci, line 47: Why not 256 like the max_email_length cap?

src/internet_identity/src/openid/sso.rs
```diff
@@ -76,6 +76,21 @@ const DISCOVERY_MAX_SCOPES: usize = 32;
 #[cfg(not(test))]
 const DEFAULT_SCOPES: [&str; 3] = ["openid", "profile", "email"];
 
+/// Maximum length of the client_id from the well-known.
+#[cfg(not(test))]
+const MAX_CLIENT_ID_LENGTH: usize = 255;
+/// Maximum length of the issuer from the OIDC configuration.
+#[cfg(not(test))]
+const MAX_ISSUER_LENGTH: usize = 255;
```
MRmarioruci, line 84: Is this the same as MAX_ISS_LENGTH just on different layers? If yes then we could name them the same

MRmarioruci (approved):

src/internet_identity/src/openid/sso.rs
```diff
@@ -76,6 +76,21 @@ const DISCOVERY_MAX_SCOPES: usize = 32;
...
+/// Maximum length of the jwks_uri from the OIDC configuration.
+#[cfg(not(test))]
```
aterga, line 86: Why can't tests work with these constants?

src/internet_identity/src/openid/sso.rs
```diff
@@ -250,6 +265,9 @@ fn is_explicitly_allowlisted(domain: &str) -> bool {
 }
 
 pub fn is_allowed_discovery_domain(domain: &str) -> bool {
```
aterga: You'll thank me later. 

```suggestion
pub fn validate_allowed_discovery_domain(domain: &str) -> Result<(), String> {
```

src/internet_identity/src/openid/sso.rs
aterga, line 393: Why don't we use a URL library here?

NCR if orthogonal to this PR

src/internet_identity/src/openid/sso.rs
```diff
@@ -367,6 +385,26 @@ async fn discovery_fill(domain: String) -> Result<DiscoveredConfig, String> {
     // discovery document can't inflate a cached `DiscoveredConfig`.
     scopes.truncate(DISCOVERY_MAX_SCOPES);
 
+    // Reject over-length values from the untrusted well-known / OIDC config.
+    if ii_config.client_id.len() > MAX_CLIENT_ID_LENGTH {
```
aterga: All of these checks should probably be under 

```suggestion
    validate_ii_config(&ii_config)?;
```

src/internet_identity/src/openid/sso.rs
aterga, line 439: Let's start following proper Rust style guidelines and using perfect constructors and destructors

```suggestion
    let Doc {
        issuer,
        jwks_uri,
        ..
    } = doc;
    
    let IiConfig {
        name,
        client_id,
        ..
    } = ii_config;

    Ok(DiscoveredConfig {
        issuer,
        client_id,
        jwks_uri,
        authorization_endpoint,
        scopes,
        name,
    })
```

aterga (approved):

src/internet_identity/src/openid/verify.rs
sea-snake: Because a Rust `u8` only fits values between 0 and 255.

src/internet_identity/src/openid/sso.rs
sea-snake: The types they bound — `IIOpenIdConfiguration` / `DiscoveryDocument` — are `#[cfg(not(test))]`: tests inject a `DiscoveredConfig` directly and skip the two-hop parse, so those fields (and their bounds) don't exist in the test build.

src/internet_identity/src/openid/sso.rs
sea-snake: Done in `90ac30c2` — now `validate_allowed_discovery_domain(&str) -> Result<(), String>`, so the verify path propagates the reason.

src/internet_identity/src/openid/sso.rs
sea-snake: Pre-existing parsing, orthogonal to the caps — leaving it as-is per your NCR.

src/internet_identity/src/openid/sso.rs
sea-snake: Done — extracted `validate_ii_config` (and `validate_discovery_document` for the OIDC-doc fields).

src/internet_identity/src/openid/sso.rs
sea-snake: Done — `DiscoveredConfig` is now built by destructuring `DiscoveryDocument` / `IIOpenIdConfiguration`.

src/internet_identity/src/openid/sso.rs
sea-snake: They bound different inputs at different layers — the JWT `iss` claim (verify.rs) vs the discovery `issuer` field from the OIDC config (here) — and live in different modules, so I've kept them separate.

src/internet_identity/src/openid/sso.rs
```diff
@@ -143,7 +143,7 @@ fn new_jwks_cache() -> JwksCache {
 /// disallowed domain.
 pub(super) fn prefetch(domain: &str) {
     let domain = domain.to_ascii_lowercase();
-    if !is_allowed_discovery_domain(&domain) {
+    if validate_allowed_discovery_domain(&domain).is_err() {
```
aterga, line 146: Why drop error messages? 

We should have a way to bubble them up to the user or at least log them.

aterga: NCR for this PR

sea-snake: I'll look into this in another PR.

PR #4114, opened 2026-07-17
feat(fe): revamp MCP settings into an AI access card

MRmarioruci:
The trusted-MCP-server settings section used a plain inline URL box and framed the feature around "Trusted MCP server". The new design reframes it as an **AI access** card and moves adding a connector into a focused dialog, so the settings page reads clearly and the risky "add a server" step gets its own guardrails.

## Changes

- Renamed the section to **AI access** with a **Preview** badge and updated copy ("Ask questions and perform actions across your apps by chatting with AI.").
- Replaced the inline URL input with a list-style **Trusted connectors** row: globe icon, host name, a **Custom** pill (with an explanatory tooltip), the URL in monospace, and a remove button.
- Added a dedicated **Add connector** dialog: featured MCP icon, safety tips (only add trusted connectors / revoke access anytime), a link to the MCP docs on "MCP connector", a URL field with a debounced verification indicator (checking → reachable → "Couldn't reach this connector", the last one advisory and still savable), and a **hold-to-confirm** action styled as the primary button.
- Added a `variant` prop to the shared `HoldToConfirm` component so the dialog's action can use the primary (white) style without affecting the confirmation-code wizard.
- Reworked the toggle flow: turning AI access on with no connector yet opens the dialog and pairs the two on-chain writes. Removing the only connector clears the URL **and** disables the feature in a single atomic write, with the UI collapsing optimistically so no stray "Add connector" button appears in between.

The on-chain data model (a single trusted URL per identity), the backend, and the locale catalog are unchanged.

## Tests

- Updated the Playwright fixture and `mcp.spec.ts` selectors for the new UI (AI access heading/switch, hold-to-confirm add flow, removal now disables the feature).
- `svelte-check` clean, MCP unit tests pass (`mcpConfig`, `mcpServer`), eslint clean, production `vite build` succeeds.

### DIFF
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
@@ -186,12 +186,12 @@
             </Tooltip>
           {:else if verifyState === "unverified" && parsedUrl !== undefined}
             <Tooltip
-              label={$t`Couldn't reach this connector`}
-              description={$t`We couldn't reach this URL to confirm it, but you can still trust it.`}
+              label={$t`Couldn't validate MCP support`}
+              description={$t`We couldn't validate MCP support at this URL, but you can still trust it.`}
             >
               <span
                 class="border-border-secondary text-fg-tertiary pointer-events-auto flex size-5 items-center justify-center rounded-full border"
-                aria-label={$t`Couldn't reach this connector`}
+                aria-label={$t`Couldn't validate MCP support`}
               >
                 <TriangleAlertIcon class="size-3" />
               </span>
--- a/src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
+++ b/src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
@@ -161,47 +161,56 @@
       </li>
     </ul>
 
-    <Input
-      bind:value={urlInput}
-      oninput={handleInput}
-      onblur={handleBlur}
-      label={$t`Connector URL`}
-      placeholder="https://mcp.example.com/mcp"
-      aria-label={$t`MCP server URL`}
-      error={errorText}
-      disabled={saving}
-      autocomplete="off"
-      autocapitalize="off"
-      spellcheck={false}
-      inputClass="pe-11 font-mono text-sm"
-    >
-      {#snippet trailing()}
-        {#if verifyState === "checking"}
-          <ProgressRing class="text-fg-tertiary size-5" />
-        {:else if verifyState === "ok"}
-          <Tooltip label={$t`Reachable MCP server`}>
-            <span
-              class="bg-bg-success-primary text-fg-success-primary pointer-events-auto flex size-5 items-center justify-center rounded-full"
-              aria-label={$t`Reachable MCP server`}
+    <div class="flex flex-col gap-1.5">
+      <span class="text-text-secondary text-sm font-medium">
+        {$t`Connector URL`}
+      </span>
+      <!-- The verify indicator is centered inside the field, so it's wrapped
+           with the input (not the label) in the positioning context. -->
+      <div class="relative">
+        <Input
+          bind:value={urlInput}
+          oninput={handleInput}
+          onblur={handleBlur}
+          placeholder="https://mcp.example.com/mcp"
+          aria-label={$t`MCP server URL`}
+          error={errorText}
+          disabled={saving}
+          autocomplete="off"
+          autocapitalize="off"
+          spellcheck={false}
+          inputClass="pe-11 font-mono text-sm"
+        />
+        <div
+          class="pointer-events-none absolute inset-y-0 inset-e-3 flex items-center"
+        >
+          {#if verifyState === "checking"}
+            <ProgressRing class="text-fg-tertiary size-5" />
+          {:else if verifyState === "ok"}
+            <Tooltip label={$t`Reachable MCP server`}>
+              <span
+                class="bg-bg-success-primary text-fg-success-primary pointer-events-auto flex size-5 items-center justify-center rounded-full"
+                aria-label={$t`Reachable MCP server`}
+              >
+                <CheckIcon class="size-3.5" />
+              </span>
+            </Tooltip>
+          {:else if verifyState === "unverified" && parsedUrl !== undefined}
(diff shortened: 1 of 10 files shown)


src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
```diff
@@ -0,0 +1,204 @@
...
+  let debounceTimer: ReturnType<typeof setTimeout> | undefined;
+
+  const verify = async (url: string) => {
+    verifyState = "checking";
+    parsedUrl = url;
+    const ok = await probeMcpServer(url);
+    if (urlInput.trim() === url) {
+      verifyState = ok ? "ok" : "unverified";
+    }
+  };
```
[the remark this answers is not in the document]
MRmarioruci: Good catch — added an onDestroy that clears the debounce timer and set a destroyed guard so a probe that's already in flight won't touch state after the dialog closes. (de79eb2)

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
```diff
@@ -0,0 +1,204 @@
...
+        </div>
+      </div>
+    </div>
+
+    <HoldToConfirm
+      label={$t`Hold to continue`}
+      variant="primary"
+      onComplete={handleConfirm}
+      class={canConfirm ? "" : "pointer-events-none opacity-60"}
+    />
```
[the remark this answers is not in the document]
MRmarioruci: Agreed. The button now takes a real disabled prop instead of pointer-events-none, so it drops out of the tab order and can't be activated by keyboard when the URL isn't confirmable. (de79eb2)

aterga (approved):

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpTrustedServersSection.svelte
```diff
@@ -225,112 +187,62 @@
...
+      {/if}
+    </div>
   {/if}
 </section>
+
+{#if showAdd}
+  <McpAddConnectorDialog
+    onClose={handleAddClose}
+    onSave={handleAddSave}
+    {saving}
```
sea-snake: Nit: The `saving` boolean can be avoided by making `onSave` prop async and awaiting the promise in the component itself to derive the 'saving' state, as seen in other dialog components.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
```diff
@@ -0,0 +1,214 @@
...
+  const handleBlur = () => {
+    const trimmed = normalizeUrl(urlInput);
+    if (trimmed === "") return;
+    if (parseMcpServerUrl(trimmed) === undefined) {
+      verifyState = "invalid";
+    }
+  };
+</script>
+
+<Dialog {onClose} width="wider">
```
sea-snake, line 107: Nit: Dialog component is rendered by parent component in this codebase, this component could be simplified to just be `McpAddConnect` rendered within a `<Dialog>` by it's parent.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpTrustedServersSection.svelte
```diff
@@ -41,15 +34,10 @@
...
-  // A canister write (toggle / add / remove) is in flight.
   let saving = $state(false);
 
-  let urlInput = $state("");
-  let error = $state<string | undefined>();
-  // Live MCP status of the trusted server: undefined while unknown / in flight,
-  // true when it verified as MCP, false when it couldn't be.
-  let verified = $state<boolean | undefined>(undefined);
-  let checking = $state(false);
+  let showAdd = $state(false);
```
sea-snake, line 36: Nit: Currently we only support a single connector, so not sure what this state variable is used for? In practice an add button is never shown.

src/frontend/tests/e2e-playwright/routes/mcp.spec.ts
```diff
@@ -7,6 +7,19 @@ import { addVirtualAuthenticator, II_URL } from "../utils";
  *  server-side per call), but kept to exercise that the param is tolerated. */
 const APP = "nice-name.com";
 
+const holdSettingsConfirm = async (page: Page, name: string): Promise<void> => {
```
sea-snake: Holding a button could be a generic e2e playwright util.

sea-snake (approved):
Added a few nit comments, nothing's blocking.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpTrustedServersSection.svelte
MRmarioruci: Done — the dialog now owns its saving state by awaiting the async `onSave`, so the parent no longer passes `saving` down. (ff1481c)

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpTrustedServersSection.svelte
MRmarioruci: Good catch — removed. Since enabling always goes through the dialog (which trusts-and-enables in one write), the `enableOnSave` branch and the add-connector button were unreachable. Dropped both. (ff1481c)

src/frontend/tests/e2e-playwright/routes/mcp.spec.ts
MRmarioruci: Done — there was already a `holdToConfirm(page)` util hardcoded to the confirmation-code flow. Generalized it to take a button name (and wait for it to be enabled), and both MCP helpers now use it. (ff1481c)

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/McpAddConnectorDialog.svelte
MRmarioruci: Left as-is: self-wrapping `<Dialog>` is the prevailing pattern here (~21 components, including the sibling CliConfirmDialog in this same folder), so keeping it consistent with those. Happy to flip it if you'd prefer the parent-wraps approach across the board.

PR #4121, opened 2026-07-21
fix(fe): reset the consent checkboxes when switching identity

MRmarioruci:
When an app asks which details to share, Internet Identity lets you switch to another identity from the switcher without leaving the screen. Until now the checkboxes you'd ticked for the first identity stayed ticked after switching, and the details of that first identity are still visible (not submittable though). Switching now takes you back to the account-selection screen for the newly selected identity, so you can choose a different account, change the access level, or decide not to continue at all before anything is shared.

## Changes

- Switching identity on the consent screen clears the authorization, dropping the normal flow back to the account-selection (Continue) view for the newly selected identity instead of re-authorizing straight back into consent.
- The OpenID consent flow has no account-selection step, so switching there keeps authorizing the new identity's default account directly.
- The attribute consent handler now loops: any change to the authorization (cleared or replaced) restarts attribute resolution for the newly selected identity, and certification always binds to that identity.
- The consent view is re-keyed on its context so it rebuilds its selections from scratch rather than carrying the previous identity's ticked boxes over.


src/frontend/src/lib/stores/authorization.store.ts
```diff
@@ -45,6 +45,12 @@ export const authorizationStore = {
   ): void => {
     authorizedInternal.set({ accountNumberPromise, accessLevel });
   },
+  /** Clear the authorization outcome without touching the effective origin
+   *  or flow. Switching identity on the consent screen uses this to drop
+   *  back to account selection for the newly selected identity. */
+  reset: (): void => {
+    authorizedInternal.set(undefined);
```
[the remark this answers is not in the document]
MRmarioruci: Added — `reset() clears the authorization outcome` in authorization.store.test.ts.

src/frontend/src/lib/stores/channelHandlers/attributes.ts
```diff
@@ -658,76 +705,88 @@ export const handleIcrc3ConsentAttributes =
...
-          attributeSpecs = consent.attributes.map((attr) => ({
-            key: attr.key,
-            value: [new Uint8Array(attr.rawValue)] as [Uint8Array],
-            omit_scope: attr.omitScope,
-          }));
+        // Switching identity on the consent screen clears the authorization
+        // and drops back to account selection, which restarts this loop:
+        // once the new identity re-authorizes, re-resolve its available
+        // attributes and repaint the screen from scratch so the previous
+        // identity's selections are never carried over or certified.
```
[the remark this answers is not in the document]
MRmarioruci: Extracted the restart decision into `consentRestart.ts` and unit-tested it in `consentRestart.test.ts`: consent-commit vs restart on the authorization changing (cleared or replaced), plus staying pending while unchanged. The full end-to-end switch flow is better exercised by a Playwright test, which I'll add separately.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
```diff
@@ -293,7 +292,11 @@
                 identities={lastUsedIdentities}
                 onSwitchIdentity={async (identityNumber) => {
                   await handleSignIn(identityNumber);
-                  await authorizeDefault();
+                  if (flow === "openid-resume") {
+                    await authorizeDefault();
```
sea-snake: Why is this condition needed? The flow is only ever "openid-resume" in the 1-click flow as far as I remember, which doesn't show an identity switcher to begin with.

MRmarioruci: a returning user hitting a 1 click flow whose dapp requests consent-requiring attributes would see the switcher.

src/frontend/src/lib/stores/channelHandlers/attributes.ts
```diff
@@ -702,32 +670,74 @@ export const handleIcrc3ConsentAttributes =
...
+            omit_scope: boolean;
+          }>;
+
+          if (pipeline.groups.length === 0 && !emailRequested) {
+            // Nothing to share and no inline "Verify an email" affordance —
+            // certify an empty set so the UI skips the empty picker view.
+            attributeSpecs = [];
+            attributeConsentStore.setConsent({ attributes: [] });
+          } else {
+            const outcome = await waitForConsentOrRestart(pipeline.authorized);
```
sea-snake: Could we inline the race here and drop `waitForConsentOrRestart` entirely? `waitForStore` already gives us the "fire-on-subscribe, wait-on-undefined" semantics, so `Promise.race` of two `waitForStore` calls expresses the same intent without the hand-rolled `settled` flag + manual unsubscribe. Tagging each branch with `as const` keeps the inferred discriminated union, so the `outcome.type === "restart"` check below still narrows cleanly:

```suggestion
            const outcome = await Promise.race([
              waitForStore(attributeConsentResultStore).then(
                (consent) => ({ type: "consent" as const, consent }),
              ),
              waitForStore(authorizedStore, (authorized) =>
                authorized !== pipeline.authorized
                  ? ({ type: "restart" as const })
                  : undefined,
              ),
            ]);
```

This lets us delete `consentRestart.ts` (and the `ConsentOutcome` type + the import on line 28), adding just `attributeConsentResultStore` to the `attributeConsent.store` import. `pipeline.authorized` is the baseline in-place, so the `authorized` field on `ConsentPipeline` can stay as-is.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
sea-snake: With the above comment's code suggestion (`authorized !== pipeline.authorized`), this PR change should now be revert-able to just `await authorizeDefault();`. Avoiding the need for `authorizationStore.reset();` in the first place. 

Practically this would mean that users don't have to re-click the authorize button when switching identities, something that the OISY team requested a while back, and the reason why `authorizeDefault();` is called here.

sea-snake (changes requested):
Besides the two comments, could you add e2e coverage for the scenarios where a user changes identity:
- user is on authorize screen and changes identity
- user is on consent screen and changes identity

Verifying that the principals and attributes received by the app is crucial here, we want to make sure to catch any regressions that could make the user authenticate or return attributes for the wrong identity.

An e2e test here would give us more reliable coverage than the unit tests on the util, hence why I'm suggesting to replace the util with a plain Promise.race().

src/frontend/src/lib/stores/channelHandlers/delegation.ts
```diff
@@ -83,9 +88,31 @@ export const handleDelegationRequest =
...
-        // Authorization is the commit point — the user may switch identities
-        // freely before this. Once authorized, the UI is no longer needed.
-        const authorized = await waitForStore(authorizedStore);
+        // Authorization is the commit point for a delegation-only flow. But
+        // when the dapp also requests attributes, the identity switcher stays
+        // live on the consent screen, so the user can re-authorize a
+        // different identity after this point. Wait for the consent to be
+        // committed (restarting on a switch) so the delegation is issued for
+        // the same identity whose attributes get certified — not whichever
+        // identity authorized first.
```
sea-snake: This might break applications that request delegations and attributes sequentially since `attributeConsentStore` is only set after the app received the delegation and sent the attribute request.

Likely, in that case `get(attributeConsentStore) !== undefined` will evaluate as `false`, so a delegation wil be actually be returned, after which the attribute screen will be shown and then consequently the attributes will be returned too. 

So nothing seems to be blocking for sequential request flows when I reason through it. But compared to parallel request flows, the app might receives attributes for a different principal than the delegation if the user switches the identity at the attributes screen.

This seems acceptable to me, II can only optimize for the attribute request when it already has it at this point. Nonetheless, I think it's important to clarify here in the code comments that basically the whole while loop section is an optimization (avoid mismatching principals on switch) for specifically the parallel requests scenario.

sea-snake (approved):
Added one nit comment regarding a code comment in the delegation request handler, LGTM, approving PR

src/frontend/src/lib/stores/channelHandlers/delegation.ts
MRmarioruci: removed comment entirely.. THe no comments rule doesn't seem to be working..

PR #4122, opened 2026-07-22
feat(fe): Add session duration picker to the sign-in / Continue screen

aterga:
# Motivation

The MCP connect screen already lets the user pick how long a session lasts, but the regular dapp sign-in / "Continue" screen offered no control over session duration — the delegation's `maxTimeToLive` was determined entirely by the requesting app.

This adds the same session-duration control to the Continue screen so users can shorten their session before signing in. Per the requirement, the duration the app requested is treated as the **maximum** the user can pick (unlike the MCP screen, where the request is only the default and the user may go up to the 30-day cap).

# Changes

- **New shared `SessionDurationSelect` component** (`$lib/components/ui/`): the dropdown control (presets + cap handling) shared by both screens, so they look and feel identical.
- **New `sessionDuration` util** (`$lib/utils/`): pure helpers for the trickier logic — `formatSessionDuration` (compact `2h`-style labels), `cappedSessionDurations` (presets up to the cap, plus the cap itself when off-preset), `sessionDurationCeilingSeconds` (requested nanoseconds → capped seconds), and `sessionDurationToNanos`.
- **`ContinueView`**: adds a "Session" section with the picker, capped at the app's requested `maxTimeToLive` and defaulting to it — so an untouched "Continue" grants exactly what the app asked for, matching previous behavior. The chosen value flows out via `onAuthorize`.
- **Authorization store**: `AuthorizationContext` now carries the app-requested `maxTimeToLive`; `Authorized` carries the user-chosen value; `authorize()` gained an optional `maxTimeToLive` argument.
- **`delegation.ts`**: surfaces the requested `maxTimeToLive` to the UI, and passes the user's chosen value (falling back to the app's request, then the backend default) into `prepare_account_delegation`.
- **`McpAuthorizeView`**: refactored to use the shared component (behavior unchanged — it still offers durations up to the 30-day maximum, defaulting to the requested value).

Durations are handled in seconds through the UI and converted to nanoseconds only at the canister boundary. The 30-day ceiling matches the backend's `MAX_SESSION_DELEGATION_TTL_NS` (also its default when a request omits `maxTimeToLive`).

# Tests

- New unit tests for `sessionDuration` (formatting, capping, nanosecond conversion, and the 30-day clamp).
- Extended `authorization.store` tests to cover the new `maxTimeToLive` field.
- `npm run check` (tsc + svelte-check) passes with 0 errors; ESLint clean on changed files; existing MCP unit tests still pass.
- The existing `delegationTtl` e2e tests remain valid: the picker defaults to the requested duration, so clicking "Continue" without interacting yields the same expiration as before.

---
_Generated by [Claude Code](https://claude.ai/code/session_011u3zkwb2JzEZgvcZhCPm2P)_


src/frontend/src/lib/stores/channelHandlers/delegation.ts
```diff
@@ -82,6 +82,11 @@ export const handleDelegationRequest =
           params.icrc95DerivationOrigin ?? channel.origin,
         );
         authorizationStore.setEffectiveOrigin(effectiveOrigin);
+        // Surface the app's requested session duration so the sign-in screen can
+        // offer durations up to it (the request is the ceiling the user picks
+        // under). `undefined` when the app didn't specify one — the backend then
+        // applies its default.
+        authorizationStore.setMaxTimeToLive(params.maxTimeToLive);
```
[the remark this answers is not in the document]
aterga: Addressed in 49a67fb. The two separate setters were replaced with a single atomic `authorizationStore.setRequestContext(effectiveOrigin, maxTimeToLive)` update, so the effective origin (which triggers the sign-in UI) and the requested duration are always set together — the Continue screen can never render with the origin known but the requested duration (the picker's ceiling) still missing. The picker also defaults to and caps at the requested value, so the chosen TTL can't exceed what the app asked for.

---

src/frontend/src/lib/utils/sessionDuration.ts
```diff
@@ -0,0 +1,82 @@
...
+  requestedMaxTimeToLive: bigint | undefined,
+): number => {
+  if (requestedMaxTimeToLive === undefined) {
+    return MAX_SESSION_DURATION_SECONDS;
+  }
+  const capped =
+    requestedMaxTimeToLive < MAX_SESSION_DURATION_NANOS
+      ? requestedMaxTimeToLive
+      : MAX_SESSION_DURATION_NANOS;
+  return Number(capped / NANOS_PER_SECOND);
```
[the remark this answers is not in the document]
aterga: Addressed in 49a67fb. `sessionDurationCeilingSeconds()` now treats a missing or non-positive `maxTimeToLive` as unspecified — falling back to the 30-day maximum, the same duration an omitted request yields — and floors the ceiling at one second. So a negative/zero value can't produce a negative/empty picker option or pass a nonsensical TTL onward, and a sub-second value can't collapse to a `0s` option. Added unit tests covering the zero, negative, and sub-second cases.

---

src/frontend/src/lib/components/ui/SessionDurationSelect.svelte
```diff
@@ -0,0 +1,84 @@
...
+  const MINUTE = 60;
+  const HOUR = 60 * MINUTE;
+  const DAY = 24 * HOUR;
+  const WEEK = 7 * DAY;
+
+  // The named durations the picker offers, spanning the allowed range (10
+  // minutes to 30 days). `$derived` so the labels re-translate when the locale
+  // changes.
+  const presets = $derived([
+    { value: 10 * MINUTE, label: $t`10 minutes` },
```
sea-snake, line 42: See e.g. `$formatRelative(dateObj, { style: "long" })` util in `src/frontend/src/lib/stores/locale.store.ts`, there's utils for formatting numbers, dates and relative dates there. If there's an `Intl` api for duration, consider adding duration utils and using those instead over adding translation labels here.

src/frontend/src/lib/utils/sessionDuration.ts
```diff
@@ -0,0 +1,91 @@
...
+const NANOS_PER_SECOND = BigInt(1_000_000_000);
+const MAX_SESSION_DURATION_NANOS =
+  BigInt(MAX_SESSION_DURATION_SECONDS) * NANOS_PER_SECOND;
+
+/**
+ * Compact label for an off-preset duration, e.g. `7200 -> "2h"`,
+ * `90_060 -> "1d 1h 1m"`. Used when the selected/cap value isn't one of the
+ * named presets.
+ */
+export const formatSessionDuration = (seconds: number): string => {
```
sea-snake, line 29: See my other comment:
> See e.g. `$formatRelative(dateObj, { style: "long" })` util in `src/frontend/src/lib/stores/locale.store.ts`, there's utils for formatting numbers, dates and relative dates there. If there's an `Intl` api for duration, consider adding duration utils and using those instead over adding translation labels here.

Ideally, we wouldn't need to implement time/date formatting functions like these, which in this particular case aren't even multi language compatible (d, h, m and s prefixes are hardcoded).

src/frontend/src/lib/components/ui/SessionDurationSelect.svelte
aterga: Good pointer, thanks. For now we're keeping the `$t` labels for the presets: they're a small, human-curated set (10 minutes … 30 days) where we want explicit control over the exact wording, and they're already translatable via the catalog. `Intl.DurationFormat` would be the natural fit, but it isn't broadly supported yet (nor in our current TS lib target), so we'd rather not take a dependency on it in this PR. Happy to revisit as a follow-up if you'd prefer the Intl route once it's viable.

---

src/frontend/src/lib/utils/sessionDuration.ts
aterga: Fair — the hardcoded `d/h/m/s` here isn't localizable, agreed. This formatter only labels an *off-preset* cap: i.e. when a dapp requests an arbitrary `maxTimeToLive` that isn't one of the presets (e.g. 2h), the picker still surfaces that exact value. So it's an edge path rather than the common case (which hits the translated presets). `Intl.DurationFormat` is the right tool but isn't broadly available yet, so we kept the simple helper to keep this PR focused. I can follow up with a proper localized duration util once `Intl.DurationFormat` is viable for us (or add it here if you'd rather block on it).

---

sea-snake (approved):

PR #4143, opened 2026-07-27
refactor(be): unify SSO client-identity resolution across org kinds

sea-snake:
## What

`resolve_ii_client_identity` had two divergent code paths for the SSO gating feature — one for `sub`-based orgs and one for stable-id orgs — that duplicated lookup logic and threaded the org's domain through inconsistently. This folds both into a single helper so the two kinds resolve the same way.

- Both kinds now look the anchor up by its stored SSO credential and read the II-client `sub` off that credential through one shared path.
- `ii_client_sub_on_anchor` takes the `sso_domain` alongside `iss` / `ii_client_id`, so all three are part of the lookup key instead of the domain being carried implicitly.

Net effect: less duplication, and the `sub` and stable-id paths behave identically instead of each having its own resolution rules.

## Behaviour note

As a consequence of unifying the paths, a first *gated* login for a `sub`-based org now follows the same "normal-login-first" shape the non-`sub` orgs already use, rather than its own one-trip path. The frontend already drives this: the canister returns `SsoNormalLoginRequired` and the existing recovery dialog replays the gated login, so there's no FE change.

## Tests

- Integration: `sub_org_first_gated_login_registers_directly` → `sub_org_first_gated_login_needs_normal_login_first` to match the unified flow; the shared positive path stays covered by `gated_and_ungated_resolve_to_same_dapp_principal`.
- e2e gating specs updated for the same flow.

### DIFF
--- a/src/internet_identity/src/openid/sso_gating.rs
+++ b/src/internet_identity/src/openid/sso_gating.rs
@@ -162,11 +162,11 @@ fn resolve_gated_ii_client_sub(
             let anchor = state::storage_borrow(|storage| {
                 storage.lookup_anchor_by_sso_stable_id(sso_domain, iss, ii_client_id, stable_id)
             })?;
-            ii_client_sub_on_anchor(anchor, iss, ii_client_id, sso_domain, stable_id)
+            ii_client_sub_on_anchor(anchor, sso_domain, iss, ii_client_id, stable_id)
         }
         None => {
             let sub = &verification.credential.sub;
-            if anchor_established_through_domain(iss, sub, ii_client_id, sso_domain) {
+            if anchor_established_through_domain(sso_domain, iss, ii_client_id, sub) {
                 Some(sub.clone())
             } else {
                 None
@@ -189,9 +189,9 @@ fn resolve_gated_ii_client_sub(
 /// login fails safe.
 fn ii_client_sub_on_anchor(
     anchor_number: AnchorNumber,
+    sso_domain: &str,
     iss: &str,
     ii_client_id: &str,
-    sso_domain: &str,
     stable_id: &str,
 ) -> Option<String> {
     state::storage_borrow(|storage| {
@@ -214,10 +214,10 @@ fn ii_client_sub_on_anchor(
 /// that `sso_domain` on the credential, so a credential established through a
 /// different domain — or a non-SSO one (`sso_domain == None`) — does not match.
 fn anchor_established_through_domain(
-    iss: &str,
-    sub: &str,
-    ii_client_id: &str,
     sso_domain: &str,
+    iss: &str,
+    ii_client_id: &str,
+    sub: &str,
 ) -> bool {
     let key = (iss.to_string(), sub.to_string(), ii_client_id.to_string());
     state::storage_borrow(|storage| {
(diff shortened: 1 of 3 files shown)


src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -100,89 +100,136 @@ pub struct SsoResolvedIdentity {
...
         let anchor = storage.read(anchor_number).ok()?;
         anchor
             .openid_credentials()
             .iter()
-            .find(|cred| cred.iss == iss && cred.aud == ii_client_id)
+            .find(|cred| {
+                cred.iss == iss
+                    && cred.aud == ii_client_id
+                    && cred.sso_domain.as_deref() == Some(sso_domain)
+            })
```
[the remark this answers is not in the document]
sea-snake: Good catch — the read side now takes `stable_id` and matches on the same full key the index used to find the anchor, so it selects the intended credential when an anchor carries several for the same `(iss, ii_client_id, sso_domain)`. Added a unit test for that case. Fixed in 5ecbcae.

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -100,89 +100,142 @@ pub struct SsoResolvedIdentity {
...
+    match stable_id {
+        Some(stable_id) => {
+            let anchor = state::storage_borrow(|storage| {
+                storage.lookup_anchor_by_sso_stable_id(sso_domain, iss, ii_client_id, stable_id)
+            })?;
+            ii_client_sub_on_anchor(anchor, iss, ii_client_id, sso_domain, stable_id)
+        }
+        None => {
+            let sub = &verification.credential.sub;
+            if anchor_established_through_domain(iss, sub, ii_client_id, sso_domain) {
```
aterga: Nit. The order of arguments should ideally match that of `lookup_anchor_by_sso_stable_id`:



```suggestion
            if anchor_established_through_domain(sso_domain, iss, ii_client_id, sub) {
```

MRmarioruci (approved):

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -100,89 +100,142 @@ pub struct SsoResolvedIdentity {
...
+            return false;
+        };
+        let Ok(anchor) = storage.read(anchor_number) else {
+            return false;
+        };
+        anchor.openid_credentials().iter().any(|c| {
+            c.iss == iss
+                && c.sub == sub
+                && c.aud == ii_client_id
+                && c.sso_domain.as_deref() == Some(sso_domain)
```
aterga, line 234: In many places throughout the code, we build conditions on multiple factors matching from two sides. 

These conditions should be organized much more responsibly. 

In particular, here I would like to see something like (don't mind the exact name of the types and functions):

```suggestion
                anchor.openid_credentials().satisfies_my_condition(iss, sub, aud, sso_domain);
```

aterga: More generally, make it hard for coding mistakes to slip through by making nontrivial decisions in one place and reusing those decisions. Here, it's not 100% clear why we match against these exact 4 factors, not more and not less.

aterga (approved):

src/internet_identity/src/openid/sso_gating.rs
sea-snake: Done — reordered both `anchor_established_through_domain` and `ii_client_sub_on_anchor` to `(sso_domain, iss, ii_client_id, …)` so they line up with `lookup_anchor_by_sso_stable_id`. b1b1ca1

PR #4153, opened 2026-07-27
feat(be,fe): show the official MCP connector in AI access settings

MRmarioruci:
Setting up AI access today means finding an MCP server URL yourself and typing it into Settings. Anyone arriving at `/mcp` for a server their identity doesn't trust yet hits a dead end that sends them here first, types a URL, and goes back.

This adds an MCP connector DFINITY hosts, configured per deployment, so the common case needs no setup at all. A fresh identity can connect to it straight from `/mcp`; the AI access card shows whichever single connector is trusted — the official one, or the user's own once they replace it.

The install arg is unset by default. With no deployment change, everything behaves exactly as it does today.

## Behaviour

Connecting and acting are different moments, and they follow different rules:

| Moment | Rule |
| --- | --- |
| A **new connect** (`/mcp`) | An identity that never configured MCP may connect the official connector — completing the consent is what enables the feature for them. One that *switched it off* may not: they are sent back to Settings rather than silently re-enabled by a link. |
| An **existing session** | Strict: `authorize_mcp_session` still requires an enabled config, so a session can't outlive a disable. |
| **Settings** | `mcp_get_config` reports the raw stored row, so an identity that never used the feature sees the toggle **off**. Visiting the dashboard enables nothing; only connecting does. |

A custom connector displaces the official one while it is set — exactly one server is trusted at a time.

Telling those two apart needs more than `enabled`, which reads `false` for both. `mcp_get_config` now returns `opt McpConfig`: `null` is an identity that never wrote a config, a record is one that did. On-chain nothing new is stored — "never configured" is simply the absence of a row in the per-anchor config map, which `read_mcp_config` used to flatten away.

That also makes the unauthorized-read path safer by construction. It answers with the switched-off shape rather than `null`, because `null` is the branch that lets a connect proceed.

Note this is a breaking Candid change on a shared type. Our two canisters deploy independently, so until both land the MCP surface degrades: an older frontend fails to decode the response, and a newer frontend against an older backend reads every identity as unconfigured. Neither grants anything — an older backend has no official connector to offer — but AI access settings and `/mcp` are broken for that window.

## Changes

**Backend**

- New `mcp_official_url : opt text` install arg, following the same pattern as `backend_origin`: applied only when supplied, persisted in `PersistentState`, reported back by `config()` so the effective value is auditable. Copied into `InternetIdentitySynchronizedConfig`, so the frontend reads it from `/.config.did.bin` on page load.
- `connect_trusted_url` / `session_trusted_url` replace reading `config.url` directly, and `register`, `authorize_mcp_session`, `prepare_mcp_registration_delegation` and the redemption URL-hash check all resolve through them.
- `register` persists `enabled: true` on a successful connect, so a first-time identity ends up with a real stored config rather than depending on a default.
- `mcp_set_config` rejects a disabled config that still carries a URL, and revokes the grant whenever the *resolved* server changes rather than only when the stored URL string does.
- `mcp_get_config` returns `opt McpConfig`, and its authorization-failure path answers with the disabled shape rather than `null`, so an unauthorized caller is never told an identity trusts a connector nor that it is free to connect one.
- `prepare_mcp_registration_delegation` writes the config row it materialises as enabled. Leaving it disabled made a first-time identity read as "switched off" by the time the connect was redeemed, so `register_v2` refused the connect it had just authorized.

**Frontend**

- `officialMcpUrl()` in `globals.ts` reads the URL out of the synchronized config; `connectTrustedUrl` / `trustedUrl` mirror the two canister rules and take `McpConfig | undefined`.
- The settings card holds a single `McpConfig | undefined` rather than separate `enabled` / `url` state, so the two can't drift apart.
- `setMcpEnabled(false)` clears the custom URL as well as flipping the flag, so the config never holds a server it isn't using.
- The AI access card renders a single **Trusted connector** row: the official connector (*Internet Computer MCP · Official · by DFINITY*) with **Customize**, or the user's own (*Added by you · Replaces the official connector*) with **Restore default**. With no official connector configured that action reads **Remove** and turns the feature off, since there is nothing to fall back to.
- The add dialog is titled *Customize AI access* when an official connector exists. Dropped the *Preview* badge from the card header.

**Deploy tooling** — `scripts/deploy-common.bash` gains an `mcp_official_url` prompt, emits the field in the backend install-arg record, and adds it to the fields `--be-extra-args-file` may not set.

## Deploying

| Path | How |
| --- | --- |
| `deploy-local-to-beta`, `deploy-pr-to-beta` | New `mcp_official_url (opt text)` prompt in the backend block |
| `make-upgrade-proposal` (staging + NNS) | No change needed — backend args are filled by `didc assist` against the `.did`, so the field is prompted automatically |
| Local `icp deploy` | `src/internet_identity/local_test_arg.did` |
| Playwright e2e stack (local) | `scripts/dev-e2e-setup` |
| Playwright e2e (CI) | install args are inlined in `.github/workflows/canister-tests.yml` — it reads neither of the files above |

Prompt values are verbatim Candid, so a text field needs the wrapper — `opt "https://mcp.internetcomputer.org/mcp"`, not a bare URL. Enter leaves it `null`, preserving the on-chain value.

`--be-extra-args-file` rejects `mcp_official_url`, matching every other field the script emits itself. The script always writes the field into the record, so a second copy would be a duplicate Candid field; the guard turns that into a clear message pointing at the prompt.

## Tests

- `mcp.rs` unit tests for both resolution rules: the official connector offered to a fresh identity, blocked for a switched-off one, a custom connector displacing it, a custom connector still needing the feature enabled, and a live session requiring it. Plus `set_mcp_config` rejecting a disabled config that keeps a URL.
- `mcp_fresh_identity_connects_the_official_connector` (PocketIC) — the headline flow: a never-configured identity connects without visiting Settings and ends up with `{enabled: true, url: None}`, with a session that can act. Regression test for the `prepare` bug above.
- `mcp_switching_off_blocks_the_official_connector` (PocketIC) — the other half: once switched off, a connect link cannot re-enable.
- `ii_canister_serves_mcp_official_url_in_synchronized_config` (PocketIC) — installs with the arg set, asserts it reaches `/.config.did.bin`, then upgrades without re-supplying it and asserts it survives.
- `mcpConfig.test.ts` — 29 cases covering both frontend rules and the config writes.
- New `mcp-settings.spec.ts` covering the card's states, plus the existing MCP specs updated for the behaviour changes. Last full local run of both files: 44 passing.
- `cargo test -p internet_identity --bin internet_identity` — 645 passing, including `check_candid_interface_compatibility`. `cargo fmt`, `clippy -D warnings`, `npm run check` and `npm run lint` clean.

## Follow-up

Identities with a stored config from before this change keep it, so anyone who had switched AI access off stays off until they turn it on or connect. A separate PR migrates those rows.

<div align="right">Next: #4157</div>


scripts/deploy-common.bash
```diff
@@ -696,8 +696,9 @@ prompt_be_extra_args() {
     # Defaults are `null` (preserve). To set a value, type the full
     # Candid expression at the prompt — e.g. `opt "https://backend.beta.id.ai"`
     # or `opt vec { "https://beta.id.ai"; "https://beta.identity.icp0.io" }`.
-    BE_BACKEND_ORIGIN_ARG=$(prompt_default  "backend_origin (opt text)"       "null")
-    BE_RELATED_ORIGINS_ARG=$(prompt_default "related_origins (opt vec text)"  "null")
+    BE_BACKEND_ORIGIN_ARG=$(prompt_default    "backend_origin (opt text)"       "null")
+    BE_RELATED_ORIGINS_ARG=$(prompt_default   "related_origins (opt vec text)"  "null")
+    BE_MCP_OFFICIAL_URL_ARG=$(prompt_default  "mcp_official_url (opt text)"     "null")
```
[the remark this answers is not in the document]
MRmarioruci: Good catch — the description predated the prompt. Rewritten with a table of every path that sets install args.

Kept the managed-field guard though: the script always emits `mcp_official_url` in the record, so letting `--be-extra-args-file` set it too would be a duplicate Candid field and a hard didc error. Same treatment as `backend_origin` and the rest. The escape hatch is for fields without a dedicated path, and this one now has the prompt.

Also verified `make-upgrade-proposal` needs no change — it fills backend args via `didc assist` against the .did, so the field is prompted there automatically.

src/frontend/src/lib/utils/mcpConfig.ts
```diff
@@ -50,19 +50,58 @@ export const originOf = (url: string): string | undefined => {
...
+    : undefined;
+
+/**
+ * The URL a *new* connect may target. An identity that never configured MCP
+ * (`undefined`) may connect the official connector — completing the consent is
+ * what enables the feature for them. One that switched the feature off may not:
+ * they are sent back to Settings rather than silently re-enabled by a link.
+ * Mirrors `connect_trusted_url` in the canister, which is what enforces it.
+ */
+export const connectTrustedUrl = (
```
sea-snake: This function only seems to be used within `isOriginTrusted`, given that it's confusing in relation to `trustedUrl`, I'd recommend to inline it within `isOriginTrusted` instead so it can't be imported and used accidentally over the other function.

src/frontend/src/lib/utils/mcpConfig.ts
```diff
@@ -50,19 +50,58 @@ export const originOf = (url: string): string | undefined => {
...
- * of that origin is set. Matching is by origin — the same security boundary the
- * delegation uses (II derives a per-origin principal; the path can't scope it).
+ * The URL `config` trusts for display: the identity's own server when they set
+ * one, otherwise `officialUrl`. Nothing while the feature is off, or when the
+ * identity has never configured it. Mirrors `session_trusted_url` in the
+ * canister.
  */
-export const isOriginTrusted = (config: McpConfig, origin: string): boolean =>
-  config.enabled && config.url !== undefined && originOf(config.url) === origin;
+export const trustedUrl = (
```
sea-snake: This function only seems to be used by the MCP manage page, I'd recommend co-locating it with that route (e.g `utils.ts` or `./utils/mcp.ts`) so it's not confused with the functions below.

src/frontend/src/lib/utils/mcpConfig.ts
```diff
@@ -50,19 +50,58 @@ export const originOf = (url: string): string | undefined => {
...
+  }
+  return config.enabled ? (config.url ?? officialUrl) : undefined;
+};
+
+/**
+ * Whether `origin` may be connected. Matching is by origin — the same security
+ * boundary the delegation uses (II derives a per-origin principal; the path
+ * can't scope it).
+ */
+export const isOriginTrusted = (
```
sea-snake: This function only seems to be used by the MCP consent page, I'd recommend co-locating it with that route (e.g `utils.ts` or `./utils/mcp.ts`) so it's not confused with the functions above.

src/frontend/src/lib/utils/mcpConfig.ts
MRmarioruci: Agreed — inlined it. It had exactly one caller, and two exported near-synonyms invited importing the permissive rule where the strict one was meant. Left a note on `isOriginTrusted` explaining why the connect rule and `trustedUrl` differ, mirroring the canister's `connect_trusted_url` / `session_trusted_url` split. Its tests folded into the `isOriginTrusted` suite.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * The URL `config` trusts for display: the identity's own server when they set
+ * one, otherwise `officialUrl`. Nothing while the feature is off, or when the
+ * identity has never configured it. Mirrors `session_trusted_url` in the
+ * canister.
+ */
+export const trustedUrl = (
+  config: McpConfig | undefined,
+  officialUrl: string | undefined,
```
sea-snake: I'd pass Pick<BackenCanisterConfig, 'officialUrl'>  into function instead of only the officialURL string param to make sure the function can't be called accidentally with an arbitrary URL as officialURL.

src/frontend/src/routes/(new-styling)/mcp/utils.ts
```diff
@@ -154,3 +155,31 @@ export const mcpAuthorize = async ({
...
+ * `connect_trusted_url` / `session_trusted_url`; this lives here, next to the
+ * consent flow that needs it, so the two can't be confused at a call site.
+ *
+ * Matching is by origin, the same boundary the delegation uses (II derives a
+ * per-origin principal; the path can't scope it).
+ */
+export const isOriginTrusted = (
+  config: McpConfig | undefined,
+  origin: string,
+  officialUrl?: string,
```
sea-snake: I'd pass Pick<BackenCanisterConfig, 'officialUrl'>  into function instead of only the officialURL string param to make sure the function can't be called accidentally with an arbitrary URL as officialURL.

src/frontend/src/lib/globals.ts
```diff
@@ -73,6 +75,9 @@ export let agentOptions: HttpAgentOptions;
 export let anonymousAgent: HttpAgent;
 export let anonymousActor: ActorSubclass<_SERVICE>;
 
+export const officialMcpUrl = (): string | undefined =>
```
sea-snake: See my other comments regarding passing the BackendCanisterConfig as a whole (filtered with Pick<>) as function argument.

sea-snake (approved):

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/utils.ts
MRmarioruci: Done — both now take `Pick<BackendCanisterConfig, "mcp_official_url">` and unwrap it themselves, so the official connector can only come from the synchronized config. Also made `isOriginTrusted`'s third argument required: omitting it silently meant "no official connector", which is the wrong default for a trust check.

src/frontend/src/routes/(new-styling)/mcp/utils.ts
MRmarioruci: Done — both now take `Pick<BackendCanisterConfig, "mcp_official_url">` and unwrap it themselves, so the official connector can only come from the synchronized config. Also made `isOriginTrusted`'s third argument required: omitting it silently meant "no official connector", which is the wrong default for a trust check.

src/frontend/src/lib/globals.ts
MRmarioruci: Removed it. With the trust rules taking `Pick<BackendCanisterConfig, "mcp_official_url">`, this was a second way to reach the same field — the two remaining call sites were display-only and now read it off the config directly.

src/frontend/src/lib/utils/mcpConfig.ts
MRmarioruci: Moved — `trustedUrl` now lives in `settings/utils.ts` next to the card that uses it, with its tests. (Sorry for the late reply: the change shipped in 57aa1c2fb, I just didn't close the thread.)

src/frontend/src/lib/utils/mcpConfig.ts
MRmarioruci: Moved — `isOriginTrusted` now lives in `mcp/utils.ts` next to the consent flow, with its tests folded into that module's suite. `mcpConfig.ts` is left with the config type, `originOf`, and the canister accessors. (Shipped in 57aa1c2fb; thread just wasn't closed.)

aterga (approved):

PR #4155, opened 2026-07-28
test(be): SSO gating cross-domain takeover regression test + threat-model docs

sea-snake:
## What

Follow-up to #4143, now with explicit framing (the fix itself shipped intentionally quiet):

1. **Integration regression test** — `openid::sso_gating::gated_login_cannot_resolve_across_discovery_domains`. It stages the cross-domain audience-relabel takeover end to end:
   - An anchor establishes its II-client identity through domain **A** (`gate.example.org`).
   - A gated login is then driven through domain **B** (`other.example.org`) — a discovery domain on a host someone else controls, declaring the *same* `client_id`/issuer and mapping the gated origin to its own per-app client — by a different caller, carrying A's cross-client `sub`.
   - It asserts `NoSuchAnchor`: B's login must not reach A's anchor.

2. **Comments** — `sso_gating.rs` resolution now states the security model plainly (the per-app→II-client relabel is the security-critical step; resolution is scoped to the discovery domain the identity was established through) instead of the deliberately vague wording it shipped with.

## Why it's a real guard

Verified fail = pass: with the domain check removed from the `sub`-org resolution path (pre-fix behaviour) and the wasm rebuilt, the new test **fails** — B's login resolves onto A's anchor and mints a real delegation (`user_key` + certified `sso_attr_bundle` for `other.example.org`). With the fix in place it passes.

## Testing

- Unit `openid::sso_gating`: 9/0
- Integration `openid::sso_gating`: 20/0 (+ the new test)
- fmt + clippy clean

No production code changes — test + comments only.


MRmarioruci (approved):

aterga (approved):
Please double check that the test expectedly fails with the old code.

PR #4156, opened 2026-07-28
fix(fe): spell out off-preset session durations in the picker

aterga:
# Motivation

The session-duration dropdown on the sign-in screen labels its named presets in words ("10 minutes", "1 hour", "8 hours", …), but fell back to a compact format for a duration that isn't a preset — the cap an app requests via `maxTimeToLive`, which is always offered as the top option. A 20-minute cap therefore rendered as "20m" directly below "10 minutes" in the same list, mixing two label styles in one dropdown. The compact form was also untranslated: `d`/`h`/`m`/`s` are hardcoded English abbreviations.

# Changes

- `locale.store.ts`: new `formatDuration` method, alongside the existing `formatDate`/`formatNumber`/`formatRelative`. It takes a duration as a count per unit (`{ day: 1, hour: 2 }`) and lets Intl do the wording: `Intl.NumberFormat` with `style: "unit"` / `unitDisplay: "long"` spells each unit out and pluralizes it for the locale, and `Intl.ListFormat` with `type: "unit"` joins them the way the locale joins a list of units — "1 day 2 hours", "1 Tag und 2 Stunden". Because the plural form comes from the locale, this is also correct for locales whose rules need more than one/other.
- `sessionDuration.ts`: `formatSessionDuration` (which returned `"1d 1h 1m"`) is replaced by `sessionDurationBreakdown`, which splits a duration into its non-zero units for `formatDuration`. Zero yields `{ second: 0 }` so a label is never empty.
- `SessionDurationSelect.svelte`: off-preset durations get their label from `$formatDuration`, so a 20-minute cap reads "20 minutes". Presets keep their existing labels. No new translatable strings.

`Intl.DurationFormat` would replace the two Intl calls with one, but it only reached the last of the browsers we support during 2025 (Safari 18.4, Firefox 140) and TypeScript 5.9.3 doesn't declare it under our `lib` setting, so `formatDuration` is the single place to switch once both land — there's a comment there saying so.

# Tests

- `locale.store.test.ts` (new): `formatDuration` spells out and pluralizes a single unit, joins several, and formats only the units it's given (including a deliberate `{ second: 0 }`).
- `sessionDuration.test.ts`: the `formatSessionDuration` cases are rewritten for `sessionDurationBreakdown` (single unit, every non-zero unit, zero duration).
- `npx vitest run` on both files: 19 passed. `tsc --noEmit` (via `tsconfig.all.json`), `svelte-check` (0 errors), `eslint --max-warnings 0` and `prettier --check` on the changed files all clean.
- Checked the rendered output across all locales in the catalogs (en, de, es, fr, id, it, nl, pl, ru, uk, ur) for single-unit and multi-unit durations; sample in [this thread](https://github.com/dfinity/internet-identity/pull/4156#discussion_r3667067685).

### DIFF
--- a/src/frontend/src/lib/components/ui/SessionDurationSelect.svelte
+++ b/src/frontend/src/lib/components/ui/SessionDurationSelect.svelte
@@ -1,11 +1,10 @@
 <script lang="ts">
   import Select from "$lib/components/ui/Select.svelte";
   import { ChevronDownIcon } from "@lucide/svelte";
-  import { plural, t } from "$lib/stores/locale.store";
+  import { formatDuration, t } from "$lib/stores/locale.store";
   import {
     cappedSessionDurations,
-    sessionDurationParts,
-    type SessionDurationPart,
+    sessionDurationBreakdown,
   } from "$lib/utils/sessionDuration";
 
   type Align = "start" | "center" | "end";
@@ -48,28 +47,12 @@
     { value: 30 * DAY, label: $t`30 days` },
   ]);
 
-  // One unit of an off-preset duration, spelled out and pluralized in the
-  // current locale, e.g. `{ unit: "minute", value: 20 } -> "20 minutes"`.
-  const partLabel = ({ unit, value }: SessionDurationPart): string => {
-    switch (unit) {
-      case "day":
-        return $plural(value, { one: `# day`, other: `# days` });
-      case "hour":
-        return $plural(value, { one: `# hour`, other: `# hours` });
-      case "minute":
-        return $plural(value, { one: `# minute`, other: `# minutes` });
-      case "second":
-        return $plural(value, { one: `# second`, other: `# seconds` });
-    }
-  };
-
-  // A named preset's label, or a spelled-out label built from the duration's
-  // units for an off-preset one (e.g. an off-preset cap shown as "2 hours").
-  // Off-preset values are written out like the presets rather than compactly,
-  // so the dropdown doesn't mix "10 minutes" with "20m".
+  // A named preset's label, or — for an off-preset duration such as a cap the
+  // app asked for — its units spelled out by the locale ("2 hours"), so the
+  // dropdown doesn't mix "10 minutes" with a compact "20m".
   const labelFor = (seconds: number): string =>
     presets.find((preset) => preset.value === seconds)?.label ??
-    sessionDurationParts(seconds).map(partLabel).join(" ");
+    $formatDuration(sessionDurationBreakdown(seconds));
 
   // Every preset up to the cap, plus the cap itself when it isn't a preset, each
   // with its label and a flag for the current selection.
--- a/src/frontend/src/lib/components/ui/SessionDurationSelect.svelte
+++ b/src/frontend/src/lib/components/ui/SessionDurationSelect.svelte
@@ -1,10 +1,11 @@
 <script lang="ts">
   import Select from "$lib/components/ui/Select.svelte";
   import { ChevronDownIcon } from "@lucide/svelte";
-  import { t } from "$lib/stores/locale.store";
+  import { plural, t } from "$lib/stores/locale.store";
   import {
     cappedSessionDurations,
-    formatSessionDuration,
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/ui/SessionDurationSelect.svelte
```diff
@@ -47,11 +48,28 @@
...
+  const partLabel = ({ unit, value }: SessionDurationPart): string => {
+    switch (unit) {
+      case "day":
+        return $plural(value, { one: `# day`, other: `# days` });
+      case "hour":
+        return $plural(value, { one: `# hour`, other: `# hours` });
+      case "minute":
+        return $plural(value, { one: `# minute`, other: `# minutes` });
+      case "second":
+        return $plural(value, { one: `# second`, other: `# seconds` });
```
sea-snake: Why not use the `Intl` duration browser api (adding another method in `$lib/stores/locale.store`) instead of manually implementing duration strings?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat

aterga: Good call — done in 650fc8c: the hand-written plural messages are gone, replaced by a `formatDuration` method in `$lib/stores/locale.store`, and `sessionDurationBreakdown` now returns the unit-keyed duration it takes.

I did not use `Intl.DurationFormat` itself, for two reasons:

- It only reached the last of the browsers we support during 2025 (Safari 18.4, Firefox 140), so on anything older `new Intl.DurationFormat(...)` throws — and this label renders as part of the sign-in screen.
- TypeScript doesn't declare it yet: `tsc` 5.9.3 rejects it under our `lib: ["ES2022", "DOM"]` and points at `es2025`, which doesn't exist even in `esnext` in that version, so we'd have to hand-write an ambient declaration.

So `formatDuration` composes two Intl APIs that have been available for years and produce the same thing: `Intl.NumberFormat` with `style: "unit"` / `unitDisplay: "long"` spells each unit out and pluralizes it per locale, and `Intl.ListFormat` with `type: "unit"` joins them the way the locale joins a list of units. Sampling our locales:

| locale | `{ minute: 20 }` | `{ day: 1, hour: 1, minute: 1 }` |
| --- | --- | --- |
| en | 20 minutes | 1 day 1 hour 1 minute |
| de | 20 Minuten | 1 Tag, 1 Stunde und 1 Minute |
| pl | 20 minut | 1 dzień, 1 godzina i 1 minuta |
| ru | 20 минут | 1 день 1 час 1 минута |
| ur | 20 منٹ | 1 دن، 1 گھنٹہ، اور 1 منٹ |

The wording, the plural form and the joining all come from the locale, so this also fixes locales whose plural rules need more than `one`/`other` — which the `$plural` version got wrong. `formatDuration` is the single place to swap in `Intl.DurationFormat` once support and typings are there; there's a comment saying so.

---

src/frontend/src/lib/utils/sessionDuration.ts
```diff
@@ -21,22 +21,36 @@ const NANOS_PER_SECOND = BigInt(1_000_000_000);
...
+  value: number;
+}
+
 /**
- * Compact label for an off-preset duration, e.g. `7200 -> "2h"`,
- * `90_060 -> "1d 1h 1m"`. Used when the selected/cap value isn't one of the
- * named presets.
+ * An off-preset duration split into its non-zero units, largest first, e.g.
+ * `7200 -> [{ hour, 2 }]`, `90_060 -> [{ day, 1 }, { hour, 1 }, { minute, 1 }]`.
+ * The caller turns each part into a localized, spelled-out label (see
```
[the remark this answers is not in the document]
aterga: Fair point, and it's moot as of 650fc8c: that function is now `sessionDurationBreakdown` and returns a unit-keyed object, so the doc examples (`7200 -> { hour: 2 }`) are the literal return value.

---

sea-snake (approved):

PR #4157, opened 2026-07-28
feat(be): batched migration enabling AI access on stored MCP configs

MRmarioruci:
> Replaces #4154, which GitHub auto-closed as merged when its head commits briefly landed on the base branch by mistake. Same branch, same commits, nothing was lost.

`enabled` on a stored MCP config predates the official connector, so a stored `false` only ever meant "no custom server of mine" — nobody could have declined a connector that didn't exist yet. Without a migration those identities would sit with AI access off indefinitely, even though the official connector is now there for them.

Anchors with **no** stored config need nothing: they already resolve to the official connector at connect time (see the parent PR). This migration is only about rows that already exist.

## Changes

- New `mcp_config_migration : opt bool` upgrade arg. `opt true` runs the migration; unset means nothing happens.
- `Storage::migrate_mcp_configs_batch(cursor, batch_size)` rewrites stored configs to `{ enabled: true, url: None }`, returning `{ migrated, is_done, next_cursor }`.
- Driven by an interval timer at 1s with a 2 000-row batch, following the SSO credential migration (#4013) and the anchor migrations before it. Rewriting every config in `post_upgrade` would blow the instruction limit.
- Hidden `mcp_config_migration_status() -> (migrated, done)` query for monitoring.
- `scripts/deploy-common.bash` gains the prompt; `make-upgrade-proposal` picks the field up automatically through `didc assist`.

**Every rewritten row loses its grant.** A grant that was inert only because the config was disabled must not come back to life authorised against a connector it was never granted for. This is the part that would be a security bug if it were skipped.

## Two departures from the #4013 pattern

**Completion is persisted, not heap-only.** That backfill was naturally idempotent — re-stamping a stamped credential is a no-op. This one isn't: re-running would re-enable anyone who deliberately switched AI access off *after* the migration. So `mcp_config_migration_done` lives in `PersistentState` and passing the arg a second time is a no-op.

**No errors endpoint.** #4013 needed one because it joined a credential index to anchors and could hit orphans. Every row here is self-contained, so there is nothing to report; a permanently-empty endpoint would be noise.

## Validating on beta

Deploy with `mcp_config_migration = opt true`, then poll:

```
icp canister call <ii> mcp_config_migration_status
```

It returns `(migrated, done)`. The count should climb in 2 000-row steps and `done` flip to `true` within a few seconds of the upgrade, after which the timer clears itself and the canister logs `MCP config migration COMPLETED (N configs migrated).`

## Tests

- `mcp_config_migration_enables_stored_configs_and_revokes_their_grants` — a disabled row with a live grant is rewritten and the grant deleted; a row already in the migrated shape is skipped; a second run migrates nothing.
- `mcp_config_migration_resumes_from_the_cursor` — a short batch reports `is_done: false` with a cursor, and the follow-up batch finishes the scan.
- `cargo test -p internet_identity --bin internet_identity` — 650 passing, including `check_candid_interface_compatibility`. `cargo fmt`, `clippy -D warnings`, `npm run check` and `npm run lint` clean.

<div align="left">Previous: #4153</div>


src/internet_identity/src/storage.rs
```diff
@@ -1195,6 +1207,80 @@ impl<M: Memory + Clone> Storage<M> {
...
+                    }
+                }
+            }
+            self.mcp_config_memory.insert(
+                anchor_number,
+                StorableMcpConfig {
+                    enabled: true,
+                    url: None,
+                    session_principal: None,
+                    pending_registration: config.pending_registration,
```
aterga: Why not reset this?

src/internet_identity/src/storage.rs
```diff
@@ -1195,6 +1207,80 @@ impl<M: Memory + Clone> Storage<M> {
...
+                        .lookup_mcp_grant(principal)
+                        .is_some_and(|grant| grant.anchor_number == anchor_number)
+                    {
+                        self.remove_mcp_grant(principal);
+                    }
+                }
+            }
+            self.mcp_config_memory.insert(
+                anchor_number,
+                StorableMcpConfig {
```
aterga, line 1293: Best to implement Default for StorableMcpConfig and then 


```suggestion
                StorableMcpConfig::default()
```

aterga (approved):

src/internet_identity/src/storage.rs
MRmarioruci: Good catch, and it's worse than untidiness. `set_mcp_config` carries it over because the redemption's URL-hash check invalidates it — but that check compares against the *current* trusted URL, and this migration can restore the very URL the entry was minted under. Enable → start a connect → switch AI access off → migrate, and a delegation consented to before the switch-off is redeemable again inside its 5-minute window. Same resurrection the grant deletion prevents, via the registration index. Now cleared alongside the grant, with test coverage.

src/internet_identity/src/storage.rs
MRmarioruci: Done — `Default` is already derived on `StorableMcpConfig`, so this is now `StorableMcpConfig { enabled: true, ..Default::default() }`. Better than the exhaustive literal for the reason you're getting at: a field added later gets cleared rather than carried over by hand.

sea-snake (approved):

src/internet_identity/src/storage.rs
```diff
@@ -1195,6 +1207,96 @@ impl<M: Memory + Clone> Storage<M> {
...
+            .range(range)
+            .take(batch_size as usize)
+        {
+            examined += 1;
+            outcome.next_cursor = Some(anchor_number);
+            // Idempotency: a row already in the migrated shape is left alone,
+            // so a re-run is a no-op.
+            if config.enabled && config.url.is_none() {
+                continue;
+            }
```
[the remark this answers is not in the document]
MRmarioruci: Correct reading of the code, but that rewrite is intended rather than accidental — sorry, nothing said so. The premise is that `enabled`/`url` both predate the official connector, so choosing a custom server was never a choice *against* it either. Those rows are reset to the default and the identity re-adds their server if they still want it; only the target shape is skipped, which is what makes a re-run a no-op.

I'd originally proposed exactly what you suggest (skip all enabled rows) and was overruled on that premise. Now spelled out in the doc comment and pinned by a test, since you're the second reviewer to read it as a bug — if that keeps happening it's a sign the trade is wrong, not just underdocumented.

src/internet_identity/src/storage/tests.rs
```diff
@@ -2373,3 +2373,106 @@ mod migrate_sso_credentials_batch_tests {
...
+    assert!(outcome.is_done);
+    assert_eq!(
+        storage.read_mcp_config(10_000),
+        StorableMcpConfig {
+            enabled: true,
+            url: None,
+            session_principal: None,
+            pending_registration: None,
+        }
+    );
```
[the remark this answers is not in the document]
MRmarioruci: Added — `mcp_config_migration_also_resets_an_enabled_custom_connector`. It asserts the opposite of what you'd expect though: the custom URL *is* dropped and its session revoked, because that's the intended behaviour (see the other thread). Either way the case is now pinned rather than unspecified.

PR #4164, opened 2026-07-29
feat(be,fe): turn AI access on at registration and let the upgrade arg control the migration

MRmarioruci:
AI access should be on by default and reachable from one place, whichever way an identity got here.

A freshly registered identity had no stored MCP config, so the Settings card read the toggle off while `/mcp` would already have let that same identity connect against the official connector. Settings said off, the connect flow said yes. Seeding the config at registration removes the split, and once every identity that can connect has a stored config, the connect flow no longer needs a rule of its own.

Separately, the migration that brings older identities to the same shape could only ever run once. It recorded completion in a persisted `mcp_config_migration_done` flag, which contradicts the design where the upgrade arg decides whether a migration runs, and means a migration that ships with a bug can never be corrected. Every other migration here, including `sso_credential_migration`, is controlled by its arg and nothing else.

That flag also made the status endpoint unreadable: `init_mcp_config_migration_timer` set the heap `done` flag whenever the arg was absent, so a canister that had never been asked to migrate reported `(0, true)`, indistinguishable from "ran and found nothing".

## Changes

**Seed the config at registration.** `Storage::init_mcp_config` writes `{ enabled: true, url: None }`, the same shape `migrate_mcp_configs_batch` gives identities that registered earlier. It runs at both anchor-creating sites, next to the `registration_rates.new_registration()` call that already has to be duplicated the same way:

- `anchor_management::registration::register` for the legacy v1 endpoint;
- `registration_flow_v2::create_identity`, which `identity_registration_finish` and `openid_identity_registration_finish` both funnel through.

Neither path is `async` after the anchor is written, so the row cannot be orphaned by a later failure and a failed registration cannot leave one behind.

The write is gated on the deployment shipping an official connector, via `mcp::init_config_for_new_identity` so the two paths cannot drift. Without one, `{ enabled: true, url: None }` resolves to no trusted URL at all — a switch that reads on with nothing behind it — so such a deployment leaves the identity with no row until it adds a server of its own.

**One trust rule instead of two.** `connect_trusted_url` and `session_trusted_url` differed only in that the former let an identity with *no* stored config connect the official connector. With registration seeding one, that arm is gone and the two are the same function:

```rust
pub(crate) fn trusted_url(config: &StorableMcpConfig) -> Option<String>
```

A config that is switched off, or absent, trusts nothing; the identity turns AI access on in Settings. `resolve_connect_url` becomes `lookup_mcp_config(anchor).as_ref().and_then(trusted_url)`, and `isOriginTrusted` mirrors it on the frontend.

**Drop the persisted migration gate.** `mcp_config_migration_done` is removed from `PersistentState`, `StorablePersistentState` (both `From` impls, both default-value tests) and from the write in `run_mcp_config_migration_batch`. The init gate reduces to the arg:

```diff
 fn init_mcp_config_migration_timer() {
-    let requested = MCP_CONFIG_MIGRATION_REQUESTED.with_borrow(|requested| *requested);
-    let already_done =
-        state::persistent_state(|persistent_state| persistent_state.mcp_config_migration_done)
-            == Some(true);
-    if !requested || already_done {
-        MCP_CONFIG_MIGRATION_DONE.replace(true);
+    if !MCP_CONFIG_MIGRATION_REQUESTED.with_borrow(|requested| *requested) {
         return;
     }
```

Dropping `MCP_CONFIG_MIGRATION_DONE.replace(true)` from that branch is what makes the status endpoint honest. Both values are per-boot:

| `mcp_config_migration_status()` | Meaning |
| --- | --- |
| `(0, false)` | not requested this boot, **or** requested and still scanning |
| `(N, true)` | a run completed this boot, migrating N configs |

Because the values reset on every upgrade, `(0, false)` on a canister that migrated last week is expected rather than alarming.

## Re-running the migration

Passing `mcp_config_migration = opt true` again re-runs it. `migrate_mcp_configs_batch` skips rows already at `{ enabled: true, url: None }`, so a re-run over untouched data is a no-op. It would re-enable an identity that switched AI access off after the first run, which is the cost of an arg-controlled migration and the same exposure `sso_credential_migration` carries.

## Unsetting the official connector

`mcp_official_url` widens from `opt text` to `opt opt text`, the same set/clear shape as `dnssec_config` and `doh_config`:

| arg | effect |
| --- | --- |
| `null` | keep the stored value |
| `opt null` | clear it — the deployment has no official connector |
| `opt opt "https://…"` | set it |

Setting was the only thing the old type could express: `null` means "keep" and `opt "…"` means "store this", so once a deployment had a connector there was no encodable way to say it no longer has one. That matters more now that a stored `{ enabled: true, url: None }` resolves its trusted URL from this arg — with no way to clear it, the only way to stand a deployment down was to re-point it somewhere.

Two readers follow the extra nesting. `config()` reports the stored value wrapped in `Some`, and the synchronized-config asset flattens it back to the plain `opt text` the frontend already consumes, so `InternetIdentitySynchronizedConfig` and every frontend call site are untouched.

Args written the old way still work: under the new type a wire `opt "https://…"` coerces to `opt opt "https://…"`, so an arg prepared against a released build keeps setting the URL rather than clearing it. `didc encode` under the old type followed by `didc decode` under the new one confirms it.

**Clearing revokes the identities still on the official connector.** They resolve their trusted URL from this arg, so unsetting it leaves them trusting nothing until they add a server of their own — deployment-wide, and the point of the switch.

## The AI access switch

The Settings switch reported the stored `enabled` flag, which is not the same question as whether anything is trusted. Two consequences, both visible on a deployment with no official connector:

- Turning it on with no connector to enable opens the add-connector dialog without writing anything. Cancelling left the switch on — the value driving it never changed, so nothing re-wrote the position the browser had already flipped, and only a reload corrected it.
- A stored `{ enabled: true, url: None }` read as on while trusting nothing.

It now reports whether a connector actually resolves, official or custom, and stays on while the add dialog is open so cancelling puts it back down. That makes the "Add connector" empty state unreachable — the switch reads off in the state that rendered it, and flipping it opens the dialog — so that branch is gone.

Not reachable from the e2e suite: `local_test_arg.did` configures `mcp_official_url`, so the no-connector paths only exist on a deployment that clears it.

## Upgrade safety

`mcp_config_migration_done` is a trailing `Option` in a Candid record, so removing it decodes cleanly in both directions. No released build has written it: it exists only in `release-2026-07-28`, for which no NNS proposal has been submitted.

## Tests

New in `tests/integration/mcp.rs`:

- `mcp_registration_enables_the_official_connector` drives both registration paths, v1 `register` and v2 `identity_registration_finish`, and asserts each lands on `{ enabled: true, url: null }`.
- `mcp_registration_writes_no_config_without_an_official_connector` pins the gate: on a deployment with no connector configured, registration writes nothing.

Updated:

- `ii_canister_serves_mcp_official_url_in_synchronized_config` now drives the whole cycle through `/.config.did.bin`: install sets the URL, an upgrade that omits it preserves it, `opt null` clears it, and it can be set again afterwards.
- `mcp_config_round_trips_and_persists_across_upgrade` spells out why nothing is stored right after registration there: that deployment ships no connector.
- The `mcp.rs` trust-rule unit tests collapse with the functions they cover: the absent-config cases are gone, the rest now exercise `trusted_url_with`.
- `mcp-settings.spec.ts` no longer opens each test by switching AI access on, since the identity now arrives enabled. `beforeEach` asserts the seeded state instead, and the disable/re-enable test starts from on.
- `mcp.spec.ts` and the `trustServer` fixture followed the same seeded state: three sites still clicked the AI access toggle to enable it, which now turns it off, so `Customize` never rendered and five specs timed out on both shards. They assert the seeded state instead.
- `isOriginTrusted`'s frontend tests cover an absent config trusting nothing.

`cargo test -p internet_identity --bin internet_identity` passes 648 tests; `cargo fmt --all --check` and `cargo clippy -p internet_identity --all-targets -- -D warnings` are clean.


src/internet_identity/src/main.rs
```diff
@@ -223,7 +223,9 @@ thread_local! {
 
 /// Temporary hidden endpoint: returns `(migrated_configs, is_done)` so beta
 /// monitoring can watch the migration land before it is proposed for
-/// production.
+/// production. Both values are per-boot, so `(0, false)` on a canister that
+/// was upgraded without `mcp_config_migration` means the migration was never
+/// asked for, not that it found nothing.
```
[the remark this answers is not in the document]
MRmarioruci: Good catch, fixed in 595042b81. The comment now names both readings of `(0, false)` (arg absent vs. scan still running) and points at `(_, true)` as the only completion signal.

sea-snake (approved):

src/internet_identity/src/storage.rs
```diff
@@ -1429,6 +1430,20 @@ impl<M: Memory + Clone> Storage<M> {
...
     }
 
+    /// Give `anchor_number` AI access on the official connector, the same shape
+    /// `migrate_mcp_configs_batch` gives identities that registered earlier.
+    /// Call it through `mcp::init_config_for_new_identity`, which holds the rule
+    /// that this is only written when the deployment has a connector to trust.
+    pub fn init_mcp_config(&mut self, anchor_number: AnchorNumber) {
+        self.mcp_config_memory.insert(
+            anchor_number,
+            StorableMcpConfig {
```
aterga, line 1440:
```suggestion
            StorableMcpConfig::default()
```

aterga: + implement a custom `Default for StorableMcpConfig`

aterga (approved):

src/internet_identity/src/storage.rs
MRmarioruci: `Default` is the disabled state (`enabled: false`), so this would seed every new identity switched off — nothing trusted, and `/mcp` refusing the connect it should allow. The spread is only there to keep `enabled: true` while leaving `url` / `session_principal` / `pending_registration` at their defaults.

MRmarioruci: Rather not: `Default` is the value for an *absent* row, not just a base to spread from — `read_mcp_config` is `get(..).unwrap_or_default()`, and `authorize_mcp_session` / `set_mcp_config` read it through that. Making it enabled means "no stored config" starts resolving to the official connector, which is the rule 1a53110 deliberately removed (and `storage::tests` pins the disabled shape).

Happy to add a named constructor — `StorableMcpConfig::enabled_on_official()` — if the spread is what reads badly here; that keeps the absent-row default fail-safe.

PR #4165, opened 2026-07-29
fix(fe): use the current sign-in picker on /mcp and /cli

MRmarioruci:
Starting an MCP connect (`/mcp`) with no saved identities showed a sign-in screen that looked and behaved like the old one: buttons read "Continue with passkey" / "Continue with Google" instead of "Sign in with …", there was no way to create an identity from that screen at all, and choosing a passkey pushed the user through an extra "Simplify your sign-in" step that `/authorize` dropped a while ago. `/cli` had the same problem.

Two separate causes, both at the call site rather than in the wizard:

`AuthWizard` figures out how it's framed by calling `isInsideAuthPanel()`. That helper was exported from `ui/AuthPanel.svelte`, a marker component with no markup, while `/mcp` and `/cli` imported `layout/AuthPanel.svelte`, the visual card — two components with the same name, only one carrying the context. The wizard therefore believed it wasn't inside a panel, which dropped the picker footer's styling and hid the switch-mode CTA for exactly the case reported: a user with no stored identities fails every clause of the `switchModeAvailable` check.

Neither route passed `mode`, so the wizard ran in its `"both"` default — the presentation from before #4043. That's where the generic copy, the missing sign-up CTA, and the extra interstitial came from. It also masked the first cause, since the footer only renders in `signin`/`signup` mode.

## Changes

- Collapsed the two `AuthPanel` components into one (`layout/AuthPanel.svelte`), which now both sets the context and renders the card, so the marker and the card can't drift apart again. `ui/AuthPanel.svelte` is deleted; `/authorize` uses the shared component with a `class` prop — its inlined classes were already identical apart from `z-1` and a conditional `rounded-t-none`.
- `/mcp` and `/cli` mount the wizard with `mode="signin"`, matching `/authorize`. Users get sign-in copy and a "Create new identity → Create" CTA to toggle into sign-up.

`/pair`, `/recovery` and `/callback-error` also import the panel and pick up the context fix, though none of them mount the wizard today.

## Tests

- New `/mcp` case asserting a user with no stored identity gets the sign-in picker and the sign-up CTA, and that the mode-less label is absent — it fails if this surface regresses.
- Fixed an assertion in the `/cli` spec that proved "the wizard isn't shown" by checking `Continue with passkey` was hidden. That label no longer exists on the route, so the assertion would have passed regardless of what rendered.

Playwright specs have not been run yet locally; CI covers them.


MRmarioruci:
image

sea-snake (approved):

PR #4169, opened 2026-07-31
feat(be): migrate every anchor to AI access on the official connector

MRmarioruci:
## Problem

The MCP config migration added in #4157 walks the *config* map, so it only ever reaches identities that already stored a config. That was justified at the time by "anchors with no stored config already resolve to the official connector at connect time" — but #4164 replaced that read-path default with registration-time seeding (`mcp::init_config_for_new_identity`), which only covers anchors registered *after* it shipped.

So the gap is: an anchor that registered earlier and never touched the feature has no config row, `resolve_connect_url` returns `None` for it, and nothing would ever switch it on. Those identities would sit with AI access off indefinitely while the official connector is right there for them.

`enabled` predating the official connector is the same argument that justifies rewriting stored rows: a stored `false` only ever meant "no custom server of mine". An *absent* row is no more a choice about the connector than a stored `false` is.

## Changes

- `Storage::migrate_mcp_configs_batch` scans `stable_anchor_memory` via `keys_range` instead of `mcp_config_memory.range`, writing `{ enabled: true, url: None }` for every anchor it reaches — including anchors that never stored a config. `keys_range` keeps the scan from decoding a `StorableAnchor` per anchor; the batch only needs the number. Cursor type is unchanged.
- The write is **unconditional** — no "already in the target shape" check. Rows enabled with a custom server are reset like everything else, so those identities re-add their server if they still want it.
- Rewriting a row is what invalidates a live session: `session_principal` returns to `None`, and `authorize_mcp_session` only authorizes a caller the config still names. The grant row is left to the existing lazy expiry cleanup on the update path, since it can no longer authorize anything.
- In-flight registrations are removed explicitly, because they are *not* covered by that: redemption resolves the entry from the registration index by `p_reg` and compares its hash against the **current** trusted URL, never consulting the config. Switching AI access off leaves the entry in place and relies on `trusted_url` being `None` to refuse it, so re-pointing an identity at the official connector would make an entry minted under that connector redeemable again.
- `init_mcp_config_migration_timer` refuses to run when the deployment ships no `mcp_official_url`, marking the migration done and logging why. `{ enabled: true, url: None }` resolves to no trusted URL without a connector, so running it there would leave every identity with AI access reading on and nothing behind it. Same rule `mcp::init_config_for_new_identity` already applies.
- `mcp::official_url()` widened to `pub(crate)` for that gate.

### Re-running is not idempotent

Unlike `migrate_sso_credentials_batch`, this migration has no skip for rows already in the target shape. A healthy identity on the official connector sits at exactly that shape with a live `session_principal`, so a second run rewrites its row and drops that session along with everyone else's — passing the arg twice logs every AI-access user out of MCP. Completion is heap-only (not persisted in `PersistentState`), so withholding the upgrade arg is the only thing that prevents it. Accepted deliberately while the feature is in Preview; pinned by a test so it can't regress silently.

## Errors — addressing the review

Thanks @gladdy and @arshavir for both points; they're handled as follows.

**What is exposed.** A registration pointer that can't be resolved to an entry the anchor owns is reported, in two cases: it resolves to another anchor's entry (left in place), or the bytes aren't a valid principal. The strings carry the **anchor number and the anomaly only, never a principal** — the same rule `migrate_sso_credentials_batch` already documents as *"No PII: anchor number only"*. There's a test asserting no principal text appears in the reported strings.

**Bounding.** `list_mcp_config_migration_errors(lo, hi)` now serves the half-open range `[lo, hi)` instead of the whole list, so the response is bounded by what you ask for rather than by how many anchors went wrong. `hi` is clamped to the number retained and to 500 past `lo`, so page until a call returns fewer than requested; an out-of-range `lo` yields an empty vector rather than trapping. Separately, retention is capped at 10 000 entries so the accumulator can't grow the heap without limit, and the moment that cap truncates anything is logged so a short list is never silent.

Every error also goes to the canister log via `ic_cdk::println!`, paired inside `push_error` so the two channels can't drift apart. That makes Thomas's "read them off the nodes" the complete record, with the endpoint as the convenient paged view.

Worth noting separately: `SSO_CREDENTIAL_MIGRATION_ERRORS` has no such cap and *is* populated. Small blast radius there, as Thomas said, but it's the instance that actually has the bug. Happy to send a follow-up rather than widen this PR.

### Note on storage

This writes one `mcp_config_memory` row per anchor across the whole anchor set — permanent per-anchor growth. The alternative, defaulting an absent row to "enabled, official connector" in the read path, would cost nothing, but the explicit-row approach was chosen so every identity has a real row to revoke against.

### Note on `url`

The target shape is `url: None`, not `Some("")`. `trusted_url_with` does `config.url.clone().or(official)`, so an empty string would short-circuit the official-connector fallback and leave the identity trusting `""`.

## Tests

Restructured into `migrate_mcp_configs_batch_tests` with a `run_to_completion` driver that threads the cursor the way the interval timer does, mirroring the SSO migration's test module.

- `writes_a_config_for_anchors_that_never_stored_one` — three bare anchors with no config row all end up `{ enabled: true, url: None }`. The case the old config-map scan could not reach.
- `rewrites_a_disabled_row_and_invalidates_its_session` — a disabled row with a live session and an in-flight registration is rewritten, its session pointer cleared and its registration removed; the grant row is left for lazy cleanup.
- `resets_an_enabled_custom_connector` — custom-server row reset to the official connector.
- `reports_a_registration_pointer_it_cannot_resolve` — both error cases reported, the other anchor keeps its entry, the rows still migrate, and no principal text leaks into the strings.
- `rerun_drops_sessions_established_since_the_first_run` — pins the accepted cost above: a session established after the first run does not survive a second.
- `resumes_from_the_cursor` — a short batch reports `is_done: false` with a cursor, the follow-up finishes, and the driver completes three anchors in two batches at `batch_size: 2`.

Plus `list_mcp_config_migration_errors_tests` for the paging itself: the half-open range, `hi` clamped to what's retained, the window capped at the page limit, and empty (never a trap) for out-of-range, empty and inverted bounds including `u64::MAX`.

- `cargo test -p internet_identity --bin internet_identity` — 655 passing, including `check_candid_interface_compatibility`. `cargo fmt` and `cargo clippy --all-targets -- -D warnings` clean. PocketIC integration tests left to CI.

### DIFF
--- a/src/internet_identity/src/http/metrics.rs
+++ b/src/internet_identity/src/http/metrics.rs
@@ -80,6 +80,21 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
             storage.count_live_mcp_grants(time()) as f64,
             "Number of live (non-expired) MCP session grants: the currently-authorized MCP sessions. Computed by scanning the grant map at scrape time (O(n) in stored grants).",
         )?;
+        w.encode_gauge(
+            "internet_identity_mcp_registration_count",
+            storage.mcp_registration_count() as f64,
+            "Number of pending MCP registration entries stored (in-flight registrations plus expired residue not yet reclaimed). Expired entries are swept by a bounded amortized GC driven by prepare writes, so the gap against the live count persists while registration traffic is idle.",
+        )?;
+        w.encode_gauge(
+            "internet_identity_mcp_live_registration_count",
+            storage.count_live_mcp_registrations(time()) as f64,
+            "Number of live (non-expired) MCP registration entries: the registration ceremonies currently redeemable. Computed by scanning the registration map at scrape time (O(n) in stored entries).",
+        )?;
+        w.encode_gauge(
+            "internet_identity_mcp_config_count",
+            storage.mcp_config_count() as f64,
+            "Number of stored per-anchor MCP configs. Configs never expire; after the MCP config migration this is one row per anchor.",
+        )?;
         if let Some(registration_rates) = storage.registration_rates.registration_rates() {
             w.gauge_vec(
                 "internet_identity_registrations_per_second",
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -196,22 +196,29 @@ fn init_sso_credential_migration_timer() {
 // `enabled` on a stored `StorableMcpConfig` predates the official connector,
 // so a stored `false` only ever meant "no custom server of mine" — nobody
 // could have declined a connector that did not exist. This one-shot migration
-// rewrites every stored config to "enabled, official connector". Anchors with
-// no stored config are left alone: they trust nothing until the identity turns
-// the feature on in Settings, and an identity registered on a deployment with
-// an official connector starts out with a config already.
+// gives AI access on the official connector to *every* anchor: it walks the
+// anchor map, so anchors that never wrote a config get a row too. Those are the
+// anchors that registered before configs were seeded at registration time, and
+// nothing else would ever switch them on.
 //
 // Driven by the `mcp_config_migration` upgrade arg and batched via an interval
 // timer, using the same convention as the SSO credential migration above:
-// rewriting every config in `post_upgrade` would blow the instruction limit.
+// writing a config for every anchor in `post_upgrade` would blow the
+// instruction limit.
 
 /// How long to wait between migration batches.
 const MCP_CONFIG_MIGRATION_BATCH_BACKOFF_SECONDS: Duration = Duration::from_secs(1);
 
-/// Maximum number of stored configs to examine per batch (= per ingress
-/// message). Matches the 2 000-per-batch convention used for previous
-/// migrations.
+/// Maximum number of anchors to examine per batch (= per ingress message).
+/// Matches the 2 000-per-batch convention used for previous migrations.
 const MCP_CONFIG_MIGRATION_BATCH_SIZE: u64 = 2_000;
+
+/// Largest window [`list_mcp_config_migration_errors`] serves in one call, so a
+/// wide `lo`/`hi` range can't push the response past the 2 MiB limit. The full
+/// set is accumulated regardless — even the pathological case of every anchor
+/// erroring is a few hundred MB of strings, which fits the heap — so paging is
+/// only about the response size, not about bounding what is kept.
+const MCP_CONFIG_MIGRATION_ERROR_PAGE_LIMIT: u64 = 500;
 
 thread_local! {
     // TODO: Remove these after the data migration is complete.
@@ -219,7 +226,30 @@ thread_local! {
     static MCP_CONFIG_MIGRATION_CURSOR: RefCell<Option<AnchorNumber>> = const { RefCell::new(None) };
     static MCP_CONFIG_MIGRATION_DONE: RefCell<bool> = const { RefCell::new(false) };
     static MCP_CONFIG_MIGRATION_COUNT: RefCell<u64> = const { RefCell::new(0) };
+    static MCP_CONFIG_MIGRATION_ERRORS: RefCell<Vec<String>> = const { RefCell::new(Vec::new()) };
     static MCP_CONFIG_MIGRATION_TIMER_ID: RefCell<Option<TimerId>> = const { RefCell::new(None) };
+}
+
+/// Temporary hidden endpoint: per-anchor errors encountered during the MCP
+/// config migration, as the half-open range `[lo, hi)`. The strings name an
+/// anchor number and the anomaly, never a principal.
+///
+/// Every error is collected, but paginated on read because this migration
+/// touches every anchor: returning the whole list at once could exceed the
+/// 2 MiB response limit and fail exactly when it is needed. `hi` is clamped to
+/// the number stored and to [`MCP_CONFIG_MIGRATION_ERROR_PAGE_LIMIT`] past `lo`,
+/// so page until a call returns fewer than requested. An out-of-range `lo`
+/// yields an empty vector rather than trapping.
+#[update(hidden = true)]
+fn list_mcp_config_migration_errors(lo: u64, hi: u64) -> Vec<String> {
+    MCP_CONFIG_MIGRATION_ERRORS.with_borrow(|errors| {
+        let len = errors.len() as u64;
+        if lo >= len || hi <= lo {
+            return vec![];
+        }
+        let end = hi.min(len).min(lo + MCP_CONFIG_MIGRATION_ERROR_PAGE_LIMIT);
+        errors[lo as usize..end as usize].to_vec()
+    })
 }
 
 /// Temporary hidden endpoint: returns `(migrated_configs, is_done)` so beta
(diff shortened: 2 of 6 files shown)


src/internet_identity/src/storage.rs
```diff
@@ -1237,55 +1271,47 @@ impl<M: Memory + Clone> Storage<M> {
...
         let range = match cursor {
             Some(cursor) => (RangeBound::Excluded(cursor), RangeBound::Unbounded),
             None => (RangeBound::Unbounded, RangeBound::Unbounded),
         };
         let mut examined: u64 = 0;
-        let mut stale: Vec<(StorableAnchorNumber, StorableMcpConfig)> = vec![];
-        for (anchor_number, config) in self
-            .mcp_config_memory
-            .range(range)
+        let mut pending: Vec<(StorableAnchorNumber, Option<StorableMcpConfig>)> = vec![];
```
aterga: This whole loop is probably redundant.

MRmarioruci: Right, gone in e9c5892. It was only needed back when the scan and the writes hit the same map — now the scan reads the anchor map while the writes land in the config and registration maps, so there is no aliasing and it is a single pass with no intermediate Vec.

One wrinkle worth noting: the registration index has to be reached by field rather than via `lookup_mcp_registration`/`remove_mcp_registration`, since those borrow all of `self` and conflict with the iterator.

src/internet_identity/src/main.rs
```diff
@@ -273,10 +284,23 @@ fn run_mcp_config_migration_batch() {
...
         return;
     }
+    if mcp::official_url().is_none() {
+        ic_cdk::println!(
+            "MCP config migration SKIPPED: mcp_config_migration was requested but this deployment \
+             has no official connector."
+        );
+        MCP_CONFIG_MIGRATION_DONE.replace(true);
+        return;
+    }
```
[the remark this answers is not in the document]
MRmarioruci: Good catch. The refusal now goes into the migration-errors vector as well as the log, so `(0, true)` can be told apart from a clean completion — also called out in the `mcp_config_migration_status` doc comment. e9c5892

src/internet_identity/src/http/metrics.rs
```diff
@@ -80,6 +80,21 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
         )?;
+        w.encode_gauge(
+            "internet_identity_mcp_registration_count",
+            storage.mcp_registration_count() as f64,
+            "Number of pending MCP registration entries stored (in-flight registrations plus expired residue not yet reclaimed). Expired entries are swept by a bounded amortized GC driven by prepare writes, so the gap against the live count persists while registration traffic is idle.",
+        )?;
+        w.encode_gauge(
+            "internet_identity_mcp_live_registration_count",
+            storage.count_live_mcp_registrations(time()) as f64,
+            "Number of live (non-expired) MCP registration entries: the registration ceremonies currently redeemable. Computed by scanning the registration map at scrape time (O(n) in stored entries).",
```
sea-snake, line 91: We're looping over all entries?

src/internet_identity/src/storage.rs
```diff
@@ -1236,56 +1270,49 @@ impl<M: Memory + Clone> Storage<M> {
...
+                        Some(entry) if entry.anchor_number == anchor_number => {
+                            self.mcp_registration_memory.remove(&principal);
+                        }
+                        Some(_) => outcome.push_error(format!(
+                            "anchor {anchor_number}: registration pointer resolves to an entry \
+                             held by another anchor; left in place"
+                        )),
+                        None => {}
+                    },
+                    Err(_) => outcome.push_error(format!(
```
aterga, line 1319: Nit. We should clean up invalid entries if we detect any.

src/internet_identity/src/storage.rs
```diff
@@ -1347,6 +1374,36 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// Mirrors [`Self::count_live_mcp_grants`]. The gap against
+    /// [`Self::mcp_registration_count`] is residue awaiting the sweep, which is
+    /// worth watching because the sweep is driven by writes: a deployment that
+    /// goes quiet stops reclaiming and the gap persists. O(n) in stored entries.
+    pub fn count_live_mcp_registrations(&self, now_ns: u64) -> u64 {
+        // Accumulate directly into a `u64`: `Iterator::count` returns `usize`,
+        // which is 32-bit on wasm32 and would wrap in a release build.
+        self.mcp_registration_memory
+            .iter()
+            .filter(|(_, entry)| entry.expires_at_ns > now_ns)
```
sea-snake, line 1402: Seems like we are indeed iterating over all entries, not sure if this will hit limits since it actually filters on entry values, thus needs to read every entry.

aterga: these are just for open registrations

aterga: so practically we don't loop over 3M entries each time

MRmarioruci: true, only for open registrations, which yes is O(n) but on a handful of entries with a 5 minute TTL. leaving it as is for now.

src/internet_identity/src/storage.rs
MRmarioruci: The stale pointer itself is cleaned up — the rewrite resets `pending_registration` to `None`, so it does not survive the batch. Added an assertion for that since it was previously incidental.

Going further is not reachable here: undecodable bytes give no key to look up, and the registration index is keyed by principal with no anchor reverse lookup, so an entry it may have referenced cannot be located without scanning the whole index. Those expire in 5 min and the amortized sweep reclaims them.

src/internet_identity/src/main.rs
```diff
@@ -196,39 +196,76 @@ fn init_sso_credential_migration_timer() {
...
+/// rather than trapping. The canister log holds every error either way.
+#[update(hidden = true)]
+fn list_mcp_config_migration_errors(lo: u64, hi: u64) -> Vec<String> {
+    MCP_CONFIG_MIGRATION_ERRORS.with_borrow(|errors| {
+        let len = errors.len() as u64;
+        if lo >= len || hi <= lo {
+            return vec![];
+        }
+        let end = hi.min(len).min(lo + MCP_CONFIG_MIGRATION_ERROR_PAGE_LIMIT);
+        errors[lo as usize..end as usize].to_vec()
```
aterga, line 251: would this panic if `errors` is `vec![]`?

MRmarioruci: no, when errors is empty, len == 0, and lo >= len is lo >= 0 always true for a u64. So it returns vec![] at line 251 before the slice.

src/internet_identity/src/main.rs
```diff
@@ -253,6 +290,19 @@ fn run_mcp_config_migration_batch() {
     MCP_CONFIG_MIGRATION_COUNT.with_borrow_mut(|count| {
         *count = count.saturating_add(outcome.migrated);
     });
+    if !outcome.errors.is_empty() {
+        MCP_CONFIG_MIGRATION_ERRORS.with_borrow_mut(|errors| {
+            let room = MCP_CONFIG_MIGRATION_MAX_ERRORS.saturating_sub(errors.len());
```
aterga: We should collect all errors - even if each anchor produces an error, it should still fit into the heap.

aterga (approved):

src/internet_identity/src/main.rs
MRmarioruci: removed limit

PR #4170, opened 2026-07-31
feat(fe): add Mission Planets: Astronauts to dapps explorer

MissionPlanets:
Register logo and authOrigins so the II authorize screen shows the Mission Planets mark for astronauts/astronautas domains.

# Motivation

# Changes

# Tests


sea-snake:

MissionPlanets:
Thank you for your answer and clear explanation regarding the WASM size.

In the meantime, is there currently any other official channel or process that would allow our Internet Identity login flow to display our logotype?

We’re happy to follow whatever official path exists (or wait for the re-architecture if that’s the only viable option). Just wanted to make sure we haven’t missed any current workaround or review process.

Thanks again for the update.

PR #4174, opened 2026-08-03
docs: check in agent rules under .claude/rules/

sea-snake:
The conventions coding agents follow in this repo have been maintained in a separate repository and symlinked into each developer's working copy by a `SessionStart` hook. That has two costs: the rules are invisible to anyone reading this repository — including new contributors and reviewers — and they drift from the code they describe, because nothing ties a rule change to the commit that invalidates it.

Checking them in fixes both. The rules land in code review alongside the code, and a change that moves a port or removes an init arg can update its rule in the same PR.

Reviewing them against the current tree turned up three claims that had already gone stale, corrected here.

# Changes

Add five rules under `.claude/rules/`, one topic per file. Each carries `paths:` frontmatter where it applies to a subtree, so it only enters an agent's context when the files it governs are being worked on:

| File | Applies to |
| --- | --- |
| `workflow.md` | all files — formatting, PR titles and descriptions, stacked PRs, CI and reviews |
| `handle-all-type-variants.md` | all files — exhaustive handling of candid variants, Rust enums, TS unions |
| `backend.md` | `src/internet_identity*/**`, `src/asset_util/**` — no-trap rule, clippy flags, PocketIC tests |
| `frontend.md` | `src/frontend/**` — design tokens, buttons, i18n, Svelte and TypeScript style |
| `e2e-testing.md` | `src/frontend/tests/**`, `scripts/dev-e2e*`, `playwright.config.ts` — the dev-e2e stack |

Correct three stale facts rather than importing them as-is:

- The e2e rule listed the local OpenID providers as `11105/11106/11107`. A fourth provider on `11108` (Entra-style, pairwise `sub` plus stable `oid`) has since been added. Rather than restate a list that just drifted, the rule now points at `OPENID_PORTS`, `DIRECT_OPENID_PORTS` and `PAIRWISE_OPENID_PORT` in `scripts/dev-e2e-setup`, and the teardown steps resolve ports from there too.
- The same rule documented a `sso_discoverable_domains` init arg and a matching-precedence gotcha. That field no longer exists anywhere in `src/` — SSO discovery is domain-based now. Removed.
- The backend rule hardcoded the `arm64-linux` PocketIC asset. The release ships all four arch/OS combinations, so it now says to match the host, and notes that `scripts/test-canisters.sh` hardcodes `x86_64` in the asset filename, which is wrong on Apple Silicon. It also points at `POCKET_IC_SERVER_VERSION` for the pin instead of quoting a version that would itself go stale.

Un-ignore the rules directory in `.gitignore`. `.claude/` becomes `.claude/*` plus `!.claude/rules/` — a directory-level ignore stops git descending, so a bare negation inside it never fires. Everything else Claude Code writes there (`settings.local.json`, worktrees, shell snapshots) stays ignored.

Reduce `CLAUDE.md` to an index of the rules directory. Its previous content, the PR review re-request guidance from #4104, moves into the "CI and reviews" section of `workflow.md` rather than being duplicated.

### DIFF
--- a/.claude/rules/frontend.md
+++ b/.claude/rules/frontend.md
@@ -1,0 +1,55 @@
+---
+paths:
+  - "src/frontend/**"
+---
+
+# Frontend
+
+## Design tokens and components
+
+- Always use Tailwind design token classes instead of hardcoded values. Colors: `bg-bg-primary`, `text-text-tertiary`, `border-border-secondary` -- never raw hex like `bg-[#ecfdf3]`. Border radius: `rounded-lg`, `rounded-xl`, `rounded-full` -- never `rounded-[10px]`. Avoid arbitrary bracket values when a named token exists.
+- Check `src/frontend/src/lib/components/ui/` for existing components before writing raw markup. The codebase has a design system; duplicating its markup inline leads to inconsistency.
+
+## Template comments
+
+- Don't write template comments that solely narrate Tailwind utility classes (e.g. "`-mx-4` bleeds outside the padding", "`@container` switches breakpoints to container query"). Utility classes are self-documenting; the only thing worth committing is the non-obvious WHY — the constraint, the bug avoided, the visual intent. If after stripping the class-paraphrase nothing remains, drop the comment entirely.
+  - Good: `<!-- React to the manage pane's actual width, not the viewport — the sidebar can narrow it -->`.
+  - Bad: `<!-- @container sets up a container query, @xl:grid-cols-3 switches to 3 columns at 36rem container width -->`.
+
+## Buttons
+
+- Never assign color classes (e.g. `text-fg-quaternary`) to icons inside `btn` elements. The button component handles icon colors; adding explicit color classes overrides built-in states (hover, focus, disabled). Only add sizing classes (e.g. `size-5`).
+- Don't add `rounded-*` classes to `btn` elements unless explicitly asked or visible in a design. Exception: standalone `btn-icon` buttons (not in a row of icon buttons) should get `rounded-full`.
+- `btn-danger` is a separate modifier class added alongside the variant. Correct: `btn btn-tertiary btn-danger`. Wrong: `btn btn-tertiary-danger`.
+
+## i18n
+
+- Content inside `<Trans>` must be on its own indented line, not inline with the tags:
+  ```svelte
+  <Trans>
+    Content here.
+  </Trans>
+  ```
+- Use `<Trans>` for paragraph (`<p>`) text content, not `$t` tagged template literals. `$t` is fine for short inline strings (button labels, headings).
+
+## Svelte patterns
+
+- When a derived value is only used within a single snippet, use `{@const}` inside the snippet rather than a script-level `$derived`. Reference the snippet's local parameter instead of component-level state.
+
+## TypeScript
+
+- Never use `as` type casts or `!` non-null assertions. Use control flow narrowing (guards, early returns) or utility functions that narrow the type. Restructure code so TypeScript can narrow through the pattern.
+- Never use inline `import("module").Type` syntax. Always import the type at the top of the file with `import type { Foo } from "module"`.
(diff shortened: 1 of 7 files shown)


.claude/rules/frontend.md
```diff
@@ -0,0 +1,55 @@
...
+---
+
+# Frontend
+
+## Design tokens and components
+
+- Always use Tailwind design token classes instead of hardcoded values. Colors: `bg-bg-primary`, `text-text-tertiary`, `border-border-secondary` -- never raw hex like `bg-[#ecfdf3]`. Border radius: `rounded-lg`, `rounded-xl`, `rounded-full` -- never `rounded-[10px]`. Avoid arbitrary bracket values when a named token exists.
+- Check `src/frontend/src/lib/components/ui/` for existing components before writing raw markup. The codebase has a design system; duplicating its markup inline leads to inconsistency.
+
+## Template comments
```
MRmarioruci, line 13: I would probably make this a bit more generic. Not only about tailwind template comments but generally we should find a way to minimize comments, and then replicate this into backend.md as well. Not sure how well it's gonna work, doesn't really work for me locally.

Something along the lines of avoid excessive commenting on code that is self-explanatory, naming (variables, functions etc..) should be as descriptive as possible so that less commenting over is needed. Make sure to update pre-existing comment blocks that would otherwise go stale etc..

.claude/rules/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+Run these even if a pre-commit hook would normally do it — hooks aren't installed in fresh clones, containers, or CI.
+
+## Code style
+
+- Prefer early-return guard clauses over nested if/else. Flatten the happy path.
+- Comments must describe the current state of the code — never reference removed code, previous implementations, or what's deliberately absent. History belongs in git.
+
+## Full-stack mindset
+
+- Don't limit fixes to frontend or backend only. When a behavior change is needed, consider whether the fix belongs in the Rust canister code, the frontend handler, the test app, or a combination. Fix the root cause in the right layer.
```
MRmarioruci, line 19: Not sure if this does anything to higher end models. It's pretty generic and probably has no measurable effect

.claude/rules/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+
+- First PR: `<div align="right">Next: #XXXX</div>`
+- Middle PR: `<div align="left">Previous: #XXXX</div>` and `<div align="right">Next: #XXXX</div>`
+- Last PR: `<div align="left">Previous: #XXXX</div>`
+
+Update all PRs in the stack when a new one is added.
+
+## CI and reviews
+
+- After pushing changes to a PR, monitor the CI pipeline for failures and fix them proactively. Poll with a loop that emits one event per check as it resolves and exits when the run completes:
```
MRmarioruci, line 52: This is probably not needed as well. There is built in functionality in claude code now that does this. Not sure about the terminal version but you can enable it via / commands probably. 

Image

MRmarioruci: This replies with the "addressed with claude" footer though, maybe we wanna remove that.

.claude/rules/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+  ```bash
+  gh api -X POST repos/OWNER/REPO/pulls/NUMBER/requested_reviewers --raw-field 'reviewers[]=Copilot'
+  ```
+  Wait for Copilot feedback and address valid suggestions before considering the PR ready.
+- After pushing changes that address a reviewer's feedback, re-request a review from that reviewer — GitHub does not re-request automatically once a reviewer has left comments or requested changes, so without this the reviewer has to notice the update on their own and the PR can stall.
+
+## Writing
+
+- Bullet points that are complete sentences must end with punctuation.
+- Always include `https://` in URLs so they render as clickable links.
```
MRmarioruci, line 75: +1. Never add "Generated with Claude" in PR descriptions, commits, code.

PR #4184, opened 2026-08-04
fix(fe): identity switcher shows two scrollbars when the identity list is long

MRmarioruci:
With enough identities on the device, the identity switcher shows two scrollbars side by side — one for the list, one for the dialog around it. Both are draggable and neither reaches the end of the list on its own. This is visible on `/mcp` when the switcher is opened from the identity row, which is the only place it renders inside a modal at desktop width.

The switcher sized its list from `window.innerHeight` (`max(2, floor((innerHeight - 380) / 74)) * 74 - 41`), but the surface holding it is capped independently: `Dialog`'s scroll area stops at `min(100dvh - keyboard, 48rem)`. On any viewport taller than ~48rem the two disagree, so the dialog scrolls *and* the list scrolls.

The `380` was calibrated for the popover the switcher usually opens in — `/cli`, `/authorize`, `/manage` and the `/mcp` header button. Those never showed a second scrollbar, because a popover has no scroll container at all; instead the popover overshoots the bottom of the viewport by up to ~20px on some heights, clipping the edge of *Add identity*.

While fixing the heights, the bottom sheet also gets the drawer cap it was always meant to have: it could previously grow to the full viewport once its content was tall enough.

# Changes

The surrounding surface now tells the switcher how much room it has, instead of the switcher guessing from the window.

- `Dialog` publishes `--max-panel-height` and uses it for its own `max-h`: `85dvh` for the mobile bottom sheet so it reads as a drawer over the page, `48rem` for the centered dialog, both clamped by `--max-content-height` and with the bottom safe-area spacer counted inside the cap — so the number is the height of the whole sheet. The `min()` keeps the keyboard-adjusted height winning whenever it is the smaller of the two, so the sheet still sits above an open keyboard.
- `Popover` publishes the same variable from its tracking loop: the room left above or below the anchor, less a 1rem viewport margin. It had no cap of its own, so this is what keeps a tall popover on screen now that the switcher no longer self-limits. The write is skipped when the value is unchanged, and the four duplicated `top` alignment maps collapsed into one `sideTop`.
- `IdentitySwitcher` caps itself at `var(--max-panel-height, 100dvh)` and makes the list the only scroll area, with the identity card, the list heading and *Add identity* staying pinned. The `window.innerHeight` binding and the magic numbers are gone.

Note that the `Dialog` changes apply to every dialog, not only the ones hosting the switcher: tall mobile sheets now stop at 85dvh, and on iOS home-bar devices tall content fits instead of being pushed off the top of the screen.

# Tests

Verified in a browser harness replicating both the dialog and popover structure: exactly one scroll area at 500px / 1000px / 1568px viewport heights with 1, 2, 3, 12 and 20 identities, the popover stays inside the viewport, and short lists do not stretch. Before the change the same harness reported two scrollers at 1000px (dialog 768/917 and list 551/880). On a simulated 800px-tall mobile viewport the sheet measures 682px with 118px of page visible above it and the list as the only scroller.

`npm run check` passes, eslint and prettier are clean, and the Tailwind utilities were compiled to confirm the arbitrary `--max-panel-height` properties emit CSS at both breakpoints. Unit-test results are unchanged from `main` in this environment.


sea-snake (approved):
Very interesting approach with the `--max-panel-height` CSS variable, a lot more elegant than the `max-height: ${Math.max(2, Math.floor((windowHeight - 380) / 74)) * 74 - 41}px` workaround I've implemented before (which only worked for the panel position in the header).

PR #4188, opened 2026-08-06
feat(be): resolve an SSO domain's session_max_age_seconds

sea-snake:
Bottom of stack #4191. Resolves the policy; nothing consumes it yet.

# Motivation

An SSO organization has no way to say how long a sign-in through its domain stays valid. This adds `session_max_age_seconds` to the `ii-openid-configuration` document and resolves it through discovery, so the layers above can act on it.

While in there: both discovery documents were validated in place and then read field by field, so using an unchecked value was a matter of remembering to call the right helper first.

# Changes

- `session_max_age_seconds` is parsed from the well-known document, defaults to eight hours, and is rejected if zero or above the 30-day delegation ceiling. The seconds-to-nanoseconds conversion happens once, in the parser, and uses `checked_mul` so a garbage value is rejected rather than saturated to the ceiling. Seconds exist only in the document.
- The resolved value rides the existing discovery cache on `DiscoveredConfig` and is exposed on `SsoDiscovery` as `session_max_age_ns`, alongside the other durations on that interface, which are all nanoseconds.
- `validate_ii_config` and `validate_discovery_document` now consume the raw documents and return `ValidatedIIOpenIdConfiguration` and `ValidatedDiscoveryDocument`, absorbing the length bounds, the URL checks, the `app_clients` parse, the stable-claim default, the host self-assertion checks and the scope defaulting. The hop-2 validator takes the validated hop-1 document, so the URL its host comparison comes from is itself checked. `discovery_fill` is now a fetch and a validate per hop.
- `validate_discovery_url` returns the URL it checked instead of `()`, so its callers bind the checked value.

# Tests

- New unit tests for the conversion: the default when absent, a plain conversion, zero, the ceiling boundary either side, and a value that overflows nanoseconds.
- `cargo test -p internet_identity --lib --bins`: 662 pass. `cargo clippy -p internet_identity -p internet_identity_interface --all-targets`: clean. `tsc --project tsconfig.all.json`: clean.
- Integration tests were not run locally.

### DIFF
--- a/src/frontend/src/lib/generated/internet_identity_types.d.ts
+++ b/src/frontend/src/lib/generated/internet_identity_types.d.ts
@@ -1027,10 +1027,11 @@ export interface InternetIdentityInit {
    */
   'is_production' : [] | [boolean],
   /**
-   * One-shot upgrade arg driving the MCP config migration: `opt true`
-   * rewrites every stored config to "enabled, official connector", since a
-   * value stored before the official connector existed was never a choice
-   * about it. Runs at most once per deployment; unset means no migration.
+   * One-shot upgrade arg driving the MCP config migration: `opt true` gives
+   * every anchor "enabled, official connector", since neither a value stored
+   * before the official connector existed nor an absent config was ever a
+   * choice about it. Unset means no migration, and the migration is refused
+   * when the deployment ships no `mcp_official_url`.
    */
   'mcp_config_migration' : [] | [boolean],
   /**
@@ -1430,15 +1431,19 @@ export interface PrepareIdAliasRequest {
 }
 /**
  * Result of prepare_mcp_registration_delegation: the canister-signature public
- * key the registration delegation is rooted at (P_reg), and the (short)
- * expiration of that delegation. The frontend fetches the signed delegation via
- * get_mcp_registration_delegation and delivers the chain to the trusted MCP
+ * key the registration delegation is rooted at (P_reg), the (short) expiration
+ * of that delegation, and the identity's resolved trusted_url (the anchor's own
+ * server if set, else the official connector). trusted_url is certified (this
+ * is an update call), so the frontend can gate delivery on it - delivering the
+ * chain only when the connect link's origin matches - rather than trusting an
+ * uncertified mcp_get_config query. The frontend fetches the signed delegation
+ * via get_mcp_registration_delegation and delivers the chain to the trusted MCP
  * server, which redeems it with mcp_register_v2.
  */
 export interface PrepareMcpRegistrationDelegation {
   'user_key' : UserKey,
-  'expiration' : Timestamp,
   'trusted_url' : string,
+  'expiration' : Timestamp,
 }
 export interface PrepareSessionDelegation {
   'user_key' : UserKey,
@@ -1665,6 +1670,12 @@ export interface SsoDiscovery {
    * The org's primary OIDC client.
    */
   'client_id' : string,
+  /**
+   * How long a sign-in through this domain stays valid, in nanoseconds, from
+   * the domain's optional `session_max_age_seconds` (8 hours when unset). The
+   * sign-in flow caps the account delegation it requests at this value.
+   */
+  'session_max_age_ns' : bigint,
 }
 /**
  * Status of a domain's SSO discovery, read by `get_sso_discovery_status`. A
--- a/src/frontend/src/lib/generated/internet_identity_types.d.ts
+++ b/src/frontend/src/lib/generated/internet_identity_types.d.ts
@@ -1672,8 +1672,7 @@ export interface SsoDiscovery {
   'client_id' : string,
   /**
    * How long a sign-in through this domain stays valid, in nanoseconds, from
-   * the domain's optional `session_max_age_seconds` (8 hours when unset). The
-   * sign-in flow caps the account delegation it requests at this value.
+   * the domain's optional `session_max_age_seconds` (8 hours when unset).
    */
   'session_max_age_ns' : bigint,
 }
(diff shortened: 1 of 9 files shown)


src/internet_identity/internet_identity.did
```diff
@@ -501,6 +501,10 @@ type SsoDiscovery = record {
     name : opt text;
     // Client the target origin runs its ceremony against; `null` when denied.
     resolved_client_id : opt text;
+    // How long a sign-in through this domain stays valid, in nanoseconds, from
+    // the domain's optional `session_max_age_seconds` (8 hours when unset). The
+    // sign-in flow caps the account delegation it requests at this value.
+    session_max_age_ns : nat64;
```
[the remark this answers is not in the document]
sea-snake: The capping applies to the account delegation (`prepare_account_delegation`'s `max_ttl`), not to the delegation minted by `prepare_jwt_delegation` — `OPENID_SESSION_DURATION_NS` governs II's own sign-in path, not what a relying party holds. So the claim is accurate for the stack, and the consumer is #4190.

The conclusion holds for a different reason though: a reviewer of this PR alone can't see that consumer, and the field ships in the interface as soon as this merges. Reworded to state what the value is and where it comes from, with the capping documented where it happens.

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -516,6 +516,11 @@ pub struct SsoDiscovery {
     /// Client the frontend runs the ceremony against for the requested origin;
     /// `None` when the origin is denied.
     pub resolved_client_id: Option<String>,
+    /// How long a sign-in through this domain stays valid, in nanoseconds, from
+    /// the domain's `session_max_age_seconds`. The sign-in flow caps the account
+    /// delegation it requests at this value. Nanoseconds, like every other
+    /// duration on this interface: seconds exist only in the well-known itself.
+    pub session_max_age_ns: u64,
```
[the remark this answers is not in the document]
sea-snake: Same distinction as on the candid field: the value bounds the account delegation's `max_ttl`, while `OPENID_SESSION_DURATION_NS` is II's own sign-in delegation. Reworded regardless, since this layer has no consumer — the docstring now says what the value is and that it defaults to eight hours.

src/frontend/src/lib/utils/ssoDiscovery.ts
```diff
@@ -21,6 +21,12 @@ export interface SsoDiscoveryResult {
   clientId: string;
   /** The client to run the ceremony against for the target origin; equals {@link clientId} when no origin was passed. */
   resolvedClientId: string;
+  /**
+   * How long a sign-in through this domain stays valid, in nanoseconds, from the
+   * domain's `session_max_age_seconds`. The sign-in flow caps the account
+   * delegation it requests at this value.
+   */
+  sessionMaxAgeNs: bigint;
```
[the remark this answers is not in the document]
sea-snake: Right that there's no consumer in this PR. It's #4190, which threads this onto the authenticated session and bounds `prepare_account_delegation`'s `max_ttl` with it. Comment reworded to drop the enforcement claim.

src/internet_identity/src/openid/sso.rs
```diff
@@ -458,9 +472,29 @@ struct DiscoveryDocument {
...
-/// Reject over-length values from the untrusted well-known configuration.
+/// The hop-1 document once every field has been checked, bounded, defaulted and
+/// converted. Produced only by [`validate_ii_config`], which consumes the raw
+/// [`IIOpenIdConfiguration`], so nothing downstream can reach an unvalidated
+/// field by accident.
+#[cfg(not(test))]
+struct ValidatedIIOpenIdConfiguration {
+    client_id: String,
+    /// The hop-2 URL, checked before it is fetched.
+    openid_configuration: String,
```
MRmarioruci, line 483: Is it true that this is a URL?

sea-snake: Yes, see the following example in production: https://dfinity.org/.well-known/ii-openid-configuration

src/frontend/src/lib/generated/internet_identity_types.d.ts
```diff
@@ -1027,10 +1027,11 @@ export interface InternetIdentityInit {
    */
   'is_production' : [] | [boolean],
   /**
-   * One-shot upgrade arg driving the MCP config migration: `opt true`
-   * rewrites every stored config to "enabled, official connector", since a
-   * value stored before the official connector existed was never a choice
-   * about it. Runs at most once per deployment; unset means no migration.
+   * One-shot upgrade arg driving the MCP config migration: `opt true` gives
```
MRmarioruci, line 1030: I guess opus 5 just decided that this is relevant

MRmarioruci: this is a good reminder to maybe remove the migration

src/internet_identity/src/openid/sso.rs
MRmarioruci: if it's a URL could we rename it to openid_configuration_url or something/

MRmarioruci (approved):

src/frontend/src/lib/generated/internet_identity_types.d.ts
sea-snake: Actually these files are generated from the `.did` file, so likely that was updated in a prior unrelated PR without regenerating the types from it. Now that this PR updates the `.did` file, the type are regenerated and those changes also appear.

src/internet_identity/src/openid/sso.rs
sea-snake: Not at this point anymore, we'd be breaking what we already have in production 😅

PR #4189, opened 2026-08-06
feat(be): carry the SSO session deadline into the attribute bundle

sea-snake:
Stacked on #4188.

# Motivation

The organization's deadline has to reach the relying party that consumes the user's SSO attributes. The ICRC-3 bundle carries an issued-at and nothing else today, so a canister has no basis to decide when the organization's claims about a user stop being true.

# Changes

- `VerifiedSsoLogin` carries the resolved policy, and `sso_prepare_delegation` computes `session_expiry_ns` once, at the ceremony. That is deliberate: it is the only point with a warm discovery cache, so an evicted entry or a refreshed policy can never move a live session's deadline.
- The internal SSO bundle gains a fourth length-prefixed field carrying that deadline, so attribute certification never re-reads the cache. Its own five-minute `SSO_ATTR_BUNDLE_TTL_NS` is untouched: marking a ceremony as fresh is a different question from when the session ends.
- Attribute certification stamps `sso:<domain>:expires_at_timestamp_ns` as a `Nat` when the bundle carries an attribute for that domain.

The key is scoped to the domain whose policy set it rather than living under `implicit:`, because one bundle can carry attributes from several scopes with different lifetimes and a verifier must enforce the deadline belonging to the scope it reads. It cannot be requested, only stamped: `AttributeName` accepts `email`, `name` and `verified_email` only, so a request for it fails validation, and `list_available_attributes` iterates `AttributeName::all()` and won't list it.

**This is fail-open until verifiers check it.** This PR only signs the deadline. `mo:identity-attributes` and any hand-rolled verifier need to reject a bundle whose deadline has passed, and only once II always stamps it can a missing key be treated as invalid rather than as an older II.

# Tests

- New unit tests for the stamping decision: the scoped key, a bundle with no attribute for that domain, two SSO domains in one bundle, and non-SSO sessions. The success path of `prepare_icrc3_attributes` needs canister salt, so the decision is tested through an extracted helper, matching how the existing ICRC-3 tests in that module are structured.
- New round-trip and malformed-field coverage for the bundle's added field.
- `cargo test -p internet_identity --lib --bins`: 667 pass. `cargo clippy -p internet_identity --all-targets`: clean.
- Integration tests were not run locally. The one exact-count assertion on a bundle (`entries.len(), 5`) is a non-SSO session, so the new key is not added there.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -1899,6 +1899,7 @@ mod openid_api {
             Err(err) => return OpenIdResult::Err(err.into()),
         };
 
+        let session_max_age_ns = verification.session_max_age_ns;
         let identity = match openid::resolve_ii_client_identity(&verification) {
             Ok(identity) => identity,
             Err(err) => return OpenIdResult::Err(err),
@@ -1922,6 +1923,12 @@ mod openid_api {
                 .prepare_jwt_delegation(session_key, anchor_number)
                 .await;
 
+            // The session deadline is fixed here, at the ceremony, from the
+            // policy captured while the discovery cache was warm. Everything
+            // downstream reads this value rather than the cache, so an evicted
+            // entry or a refreshed policy can't move a live session's deadline.
+            let session_expiry_ns = ic_cdk::api::time().saturating_add(session_max_age_ns);
+
             let (sso_attr_bundle, _bundle_expiration) = openid::prepare_sso_attr_bundle(
                 &identity.credential.iss,
                 &identity.credential.sub,
@@ -1929,6 +1936,7 @@ mod openid_api {
                 anchor_number,
                 &discovery_domain,
                 &origin,
+                session_expiry_ns,
             );
 
             // The association could change during the `.await`.
@@ -2517,9 +2525,10 @@ mod attribute_sharing {
         } = request.try_into()?;
 
         // SSO session iff a certified bundle for this origin is attached; gates `sso:<domain>` attributes.
-        let sso_session_domain = openid::read_certified_sso_bundle()
-            .filter(|bundle| bundle.origin == origin)
-            .map(|bundle| bundle.sso_domain);
+        let sso_session =
+            openid::read_certified_sso_bundle().filter(|bundle| bundle.origin == origin);
+        let sso_session_expiry_ns = sso_session.as_ref().map(|bundle| bundle.session_expiry_ns);
+        let sso_session_domain = sso_session.map(|bundle| bundle.sso_domain);
         let (anchor, _) =
             check_authorization(identity_number).map_err(|AuthorizationError { principal }| {
                 PrepareIcrc3AttributeError::AuthorizationError(principal)
@@ -2540,6 +2549,7 @@ mod attribute_sharing {
             issued_at_timestamp_ns,
             account,
             sso_session_domain,
+            sso_session_expiry_ns,
         )?;
 
         Ok(PrepareIcrc3AttributeResponse { message })
--- a/src/internet_identity/src/openid/sso_bundle.rs
+++ b/src/internet_identity/src/openid/sso_bundle.rs
@@ -6,6 +6,7 @@
 //! u64-BE len || sso_domain bytes       SSO discovery domain
 //! u64-BE len || origin bytes           certified dapp origin
 //! u64-BE len || expiry_ns (u64 BE)     bundle expiry, ns since epoch
+//! u64-BE len || session_expiry_ns  session deadline, ns since epoch
 //! ```
 
 use super::{calculate_delegation_seed, SSO_ATTR_BUNDLE_TTL_NS};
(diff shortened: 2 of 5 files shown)


src/internet_identity/src/openid/sso_bundle.rs
```diff
@@ -109,6 +123,7 @@ pub fn decode_sso_attr_bundle(bytes: &[u8]) -> Result<SsoAttrBundle, SsoBundleDe
     let sso_domain_bytes = reader.read_field()?;
     let origin_bytes = reader.read_field()?;
     let expiry_bytes = reader.read_field()?;
+    let session_expiry_bytes = reader.read_field()?;
```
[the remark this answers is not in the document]
sea-snake: Deliberately not backwards compatible. The bundle lives five minutes, so the only bundles that fail to decode across an upgrade belong to a ceremony in flight, and the user re-runs it. Carrying a compatibility path for that would mean keeping a second interpretation of the format alive indefinitely to cover a five-minute window.

src/internet_identity/src/main.rs
```diff
@@ -2517,9 +2525,10 @@ mod attribute_sharing {
         } = request.try_into()?;
 
         // SSO session iff a certified bundle for this origin is attached; gates `sso:<domain>` attributes.
-        let sso_session_domain = openid::read_certified_sso_bundle()
-            .filter(|bundle| bundle.origin == origin)
-            .map(|bundle| bundle.sso_domain);
+        let sso_session =
+            openid::read_certified_sso_bundle().filter(|bundle| bundle.origin == origin);
+        let sso_session_expiry_ns = sso_session.as_ref().map(|bundle| bundle.session_expiry_ns);
+        let sso_session_domain = sso_session.map(|bundle| bundle.sso_domain);
```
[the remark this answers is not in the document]
sea-snake: This is expected behaviour rather than a gap. The deadline is enforced where it's read, not where it's minted: a bundle whose deadline passes shortly after issuance is the normal case for any policy, and a verifier rejects it. Certifying during that window leaks nothing, because `sso:<domain>:email|name` travel in the same signed map as the deadline, so a consumer enforcing it drops the whole bundle.

An issuance-time check would be defence in depth at best, and it has two costs: it adds a second enforcement point that can disagree with the first at the boundary, and since `prepare_icrc3_attributes` fails the entire request on any problem, refusing there would also take down a request's non-SSO attributes. The upside is only that a dapp finds out sooner, which is a UX preference, not a security property.

src/internet_identity/src/openid/sso_gating.rs
```diff
@@ -35,6 +35,9 @@ pub struct VerifiedSsoLogin {
     pub stable_identifier_claim: String,
     /// The cross-client-stable identifier; `None` when the claim is `sub`.
     pub stable_id: Option<String>,
+    /// The org's session length in nanoseconds, captured here while the
```
MRmarioruci, line 38: In the rules PR.. maybe we should agree where to comment on. LLMs will just add the same comment across 50 different places where the property lives.

sea-snake: Good point, this will be a rather nuanced rule since duplicate comments can make sense if those add actual value in their respective places (e.g. same comment internally and on an external .did file).

src/internet_identity/src/main.rs
```diff
@@ -1922,13 +1923,20 @@ mod openid_api {
                 .prepare_jwt_delegation(session_key, anchor_number)
                 .await;
 
+            // The session deadline is fixed here, at the ceremony, from the
+            // policy captured while the discovery cache was warm. Everything
+            // downstream reads this value rather than the cache, so an evicted
+            // entry or a refreshed policy can't move a live session's deadline.
+            let session_expiry_ns = ic_cdk::api::time().saturating_add(session_max_age_ns);
```
MRmarioruci, line 1930: `validate_session_max_age` accepts 1s–30d, but this flow issues a flat 30-minute delegation (`OPENID_SESSION_DURATION_NS`) and nothing compares the two.

A domain publishing `session_max_age_seconds: 60` therefore gets a 60-second deadline in the bundle and a 30-minute delegation alongside it. Under 5 minutes it's shorter than the bundle TTL too, which breaks the "must outlive the bundle's freshness window" invariant the new test asserts — it just happens to hold for the 15-minute value under test.

Either clamp the delegation to `min(OPENID_SESSION_DURATION_NS, session_max_age_ns)`, or reject anything below `OPENID_SESSION_DURATION_NS` at parse time so we don't publish a deadline we don't honor.

src/internet_identity/tests/integration/openid.rs
```diff
@@ -2941,4 +2955,108 @@ mod sso_gating {
         );
         Ok(())
     }
+
+    /// Reads `sso:<domain>:expires_at_timestamp_ns` out of a certified bundle.
+    fn session_expiry_from_message(message: &[u8]) -> u64 {
+        use candid::Decode;
```
MRmarioruci, line 2961: Not sure if this a Rust idiomatic way but I prefer imports to be at the top of the module. Not a deal-breaker. Maybe a new rule?

MRmarioruci (approved):

src/internet_identity/src/main.rs
sea-snake: Delegations are clamped on the FE side in the next PR since they're not an SSO security boundary (the attributes are). (Also `OPENID_SESSION_DURATION_NS ` is for a different delegation than the app receives, so the 30min fixed value isn't correct in this regard either).

src/internet_identity/tests/integration/openid.rs
sea-snake: Yeah, let's discuss it and put whatever we agree on in a rule.

PR #4195, opened 2026-08-07
fix(be): pin AuthorizationKey on email challenges and drop pending on…

unixway:
… auth changes

Store the prepare-time authn method (not caller/temp-key principal) so finalize still works after temp-key expiry, and fail closed if that method was removed. Invalidate in-flight setup/verify challenges on security-relevant control-set mutations.

# Motivation

Email recovery setup is a long-running, two-phase flow. An authenticated prepare_add issues a challenge that remains valid for 30 minutes; binding completes later over the unauthenticated inbound path (SMTP plus DoH or DNSSEC). Between those steps the identity’s control set can change, but the pending challenge only tracked the target anchor—not which authentication method started the flow.

Stale challenges after control-set changes. If the method that started the wizard is removed or replaced mid-flow—device remove/replace, OpenID unlink, recovery or verified-email remove, or confirming a new method out of registration mode—an in-flight challenge can still complete and bind email state, or linger as a zombie entry and race with cleanup.

This change hardens the flow without altering the external Candid happy path: pin the prepare-time AuthorizationKey (device / OpenID / recovery credential), re-check that method’s membership on the anchor at finalize (independent of temp-key lifetime), and drop in-flight setup and verified-email challenges on security-relevant auth mutations in the same update that mutates the control set.

# Changes

- Pin prepare-time `AuthorizationKey` (device/OpenID/recovery), not live caller / temp-key principal.
- Re-auth at finalize via method still on the anchor (temp-key TTL no longer cuts a 30m challenge to 10m).
- Drop in-flight email setup + verified-email challenges on authn remove/replace, OpenID remove, recovery/verified-email remove, and registration-mode device confirm.

# Notes

- No Candid happy-path API change.
- Heap-only pending map; no stable migration.

# Tests

- Integration: `finalize_succeeds_after_temp_key_expires`, `finalize_fails_after_prepare_device_removed`
- Full `email_recovery` suite: 29 passed


src/internet_identity/src/main.rs
```diff
@@ -416,6 +416,9 @@ fn verify_tentative_device(
     ) {
         Ok(maybe_confirmed_device) => {
             if let Some(confirmed_device) = maybe_confirmed_device {
+                // Registration-mode exit that adds a method changes the
+                // control set — drop in-flight email challenges first.
+                email_inbound::drop_challenges_for_anchor(anchor_number);
```
aterga: These registration-mode drops (here, and the `v2_api` sibling around L1616) fire on device **addition**. Adding a method doesn't invalidate the `AuthorizationKey` pinned at prepare, so `finalize_authenticated_bind` would already let such a challenge complete correctly — the drop isn't needed for the security property here, and it's asymmetric with the regular add paths (`authn_method_add`, OpenID add) that don't drop. Net effect: confirming a device during the 30-minute window silently kills an unrelated in-flight email setup. Intended, or can these two drops come out?

---

src/internet_identity/src/email_recovery/remove.rs
```diff
@@ -57,6 +57,11 @@ pub fn remove_credential(anchor: &mut Anchor, address: &str) -> Result<Operation
         .position(|c| c.address.eq_ignore_ascii_case(address))
         .ok_or(RemoveError::NotRegistered)?;
 
+    // Drop pending email challenges before clearing the credential so
+    // a prepare started under recovery-delegation auth cannot finalize
+    // after the recovery email is detached.
+    crate::email_inbound::drop_challenges_for_anchor(anchor.anchor_number());
```
aterga, line 63: Optional: an integration test that prepares a challenge, calls `email_recovery_credential_remove`, then asserts the challenge reads `Expired` would pin the original persistence scenario (recovery-authenticated rebind) end-to-end — a nice complement to the device-removal test already added.

---

aterga (commented):
Reviewed the full diff against `main`. The core design is sound: the finalize-time re-auth (`finalize_authenticated_bind`) is the real guarantee — all four bind sites route through it — and the eager `drop_challenges_for_anchor` is solid defense-in-depth. The OpenID pre-check-before-drop is correctly placed, too: OpenID remove returns `Err` (no state rollback), unlike the device paths which trap and roll back, so dropping first there would have stranded a legitimate challenge on a mistyped remove.

Two non-blocking notes inline; neither gates the change.

---

src/internet_identity/src/email_recovery/remove.rs
unixway: fixed

src/internet_identity/src/main.rs
```diff
@@ -2002,11 +2008,14 @@ mod email_recovery_api {
         identity_number: IdentityNumber,
         dns_input: EmailChallengeDnsInput,
     ) -> Result<EmailChallenge, EmailChallengeError> {
-        check_authorization(identity_number)
+        // Pin the authorization *method* (device pubkey / OpenID key /
+        // recovery address), not `caller()`: a temp-key principal expires
+        // in 10 minutes while the challenge lives 30.
+        let (_anchor, authorization_key) = check_authorization(identity_number)
             .map_err(|err| EmailChallengeError::Unauthorized(err.principal))?;
```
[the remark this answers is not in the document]
unixway: Regression resolved.

src/internet_identity/src/main.rs
```diff
@@ -2119,12 +2128,19 @@ mod verified_email_api {
         identity_number: IdentityNumber,
         dns_input: EmailChallengeDnsInput,
     ) -> Result<EmailChallenge, EmailChallengeError> {
-        check_authorization(identity_number)
+        // Pin the authorization *method*, not `caller()` — see
+        // `email_recovery_credential_prepare_add`.
+        let (_anchor, authorization_key) = check_authorization(identity_number)
             .map_err(|err| EmailChallengeError::Unauthorized(err.principal))?;
```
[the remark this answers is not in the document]
unixway: Regression resolved.

sea-snake (dismissed):

src/internet_identity/src/email_inbound/mod.rs
```diff
@@ -72,6 +72,14 @@ pub fn pending_diagnostics(
     pending::diagnostics_of(nonce, now_secs)
 }
 
+/// Drop in-flight setup / verified-email challenges for `anchor`.
+/// See [`pending::drop_challenges_for_anchor`].
+pub fn drop_challenges_for_anchor(
+    anchor: internet_identity_interface::internet_identity::types::AnchorNumber,
```
aterga, line 78: Nit 

```suggestion
    anchor: AnchorNumber,
```

src/internet_identity/src/email_inbound/mod.rs
```diff
@@ -72,6 +72,14 @@ pub fn pending_diagnostics(
     pending::diagnostics_of(nonce, now_secs)
 }
 
+/// Drop in-flight setup / verified-email challenges for `anchor`.
+/// See [`pending::drop_challenges_for_anchor`].
+pub fn drop_challenges_for_anchor(
```
aterga, line 77: Is this function needed? Or is it just reexporting?

src/internet_identity/src/email_inbound/pending.rs
```diff
@@ -392,6 +409,41 @@ fn is_expired_at(c: &PendingChallenge, now_secs: u64) -> bool {
...
+/// now-revoked method can't complete, and so the pending map doesn't keep
+/// "zombie" entries that only the finalize-time re-auth would reject.
+/// Method *additions* (including registration-mode confirm) do not drop:
+/// they leave the pinned AuthorizationKey valid.
+///
+/// Recovery-flow entries (`PendingKind::Recover`) are *not* keyed by
+/// anchor (the anchor is resolved from the verified `From:` later)
+/// and are left alone.
+///
+/// Order contract with callers: invoke this **before** mutating the
```
aterga, line 427: Is there an elegant way to enforce this contract in the code? 

Otherwise, it's scattered across multiple files / functions, hard to verify if it's called in all the places where it's needed.

aterga (dismissed):
Pre-approving, 

but I hope we can avoid calling the `drop_challenges_for_anchor` function in multiple places. This could be addressed in a follow-up if desired. 

See https://github.com/dfinity/internet-identity/pull/4195/changes#r3751725339

src/internet_identity/src/email_inbound/mod.rs
unixway: Well, yes, I believe we can drop it, and call 'pending::drop_challenges_for_anchor' directly.

unixway: Just small note that calling crate::email_inbound::pending::drop_challenges_for_anchor directly would unnecessarily couple anchor_management, email_recovery, and verified_emails to the internal storage implementation.

src/internet_identity/src/email_inbound/pending.rs
unixway: We could move drop_challenges_for_anchor into post_operation_bookkeeping and trigger it based on the Operation variant?

aterga: post_operation_bookkeeping is mostly for archiving operations, so unfortunately we don't seem to have a great single place for calling this at the moment. Let's keep as-is for now.

aterga (dismissed):

src/internet_identity/src/email_inbound/submit_leaf.rs
```diff
@@ -266,6 +277,23 @@ async fn finalize(
...
+/// removed mid-flow, the challenge is stale (`Unauthorized`) and the
+/// anchor is left untouched. Does **not** depend on temp-key TTL.
+fn finalize_authenticated_bind(
+    anchor: &AnchorNumber,
+    authorization_key: &AuthorizationKey,
+    bind: impl FnOnce() -> Result<(), EmailChallengeError>,
+) -> Result<(), EmailChallengeError> {
+    if !authorization_key_is_valid(*anchor, authorization_key) {
+        return Err(EmailChallengeError::Unauthorized(
+            unauthorized_principal_for_key(authorization_key),
```
sea-snake, line 291: The `unauthorized_principal_for_key` method is only used here, I'd inline it's implementation inside `finalize_authenticated_bind` itself to avoid the need for the `unauthorized_principal_for_key` method to exist. 

The `unauthorized_principal_for_key` method is a potential risk since AI could accidentally rename this and use it for `caller()` principal authentication, which could lead to all kinds of risky unintended authentication code.

src/internet_identity/src/email_inbound/mod.rs
unixway: reexported pending::drop_challenges_for_anchor

src/internet_identity/src/email_inbound/submit_leaf.rs
unixway: fixed

sea-snake (approved):
LGTM

PR #4197, opened 2026-08-10
docs(be,fe): push notifications design doc and candid interface

MRmarioruci:
Draft, for a discussion on the interface before anything gets built.

## Changes

- **`src/internet_identity/internet_identity.did`** — the `notification_*` / `webpush_*` methods and types, split into two labelled groups: **app-facing** (a sending dApp) and **II-facing** (II's own frontend / service worker).
- **`docs/ii-notifications-dapp.did`** — the interface a sending dApp implements, plus the `/.well-known/ii-notification-senders` file it serves.

Both parse under `didc check`.

Two things to know before reading it:

- **The notification carries no text.** II sends a content-free ping; the device fetches the message from the sending app. II never sees the content, so end-to-end privacy is free, no vetKeys.
- **Sealing is cached.** The curve math depends only on the device, so it happens once at approval, not per notification.

## App-facing surface

```
notification_send : (NotificationSendRequest) -> (NotificationSendResponse);
```

- Request and response are records with **all-`opt` fields** (the conventional signature, per @aterga's review), so the surface can grow without a breaking candid change.
- **No `origin` parameter.** A dApp authorizes its sender canisters by serving `/.well-known/ii-notification-senders` at its origin; II fetches that list and derives the send origin from `caller()`. Serving the file proves ownership, so there is deliberately no register endpoint.
- **`drain_epoch`** in the response: II's buffer is transient and drops on upgrade, so a sender that sees the epoch change resends whatever it hadn't heard back on. II is fire-and-forget; the dApp holds the campaign.

## II-facing surface

Consent (`grant`/`revoke`/`status`/`consented_origins`) and device subscription (`webpush_*`) are called by II's own frontend, so they keep simple signatures and take `anchor_number` for authorization.

## Worth arguing over

- **`not_subscribed` vs `no_consent`.** `not_subscribed` reveals device state for a user the app already knows; `no_consent` merges "unknown identity" with "not consented" so existence can't be probed.
- **`rejected` is keyed by id**, not a position, so it survives II reordering.
- **`opt message` is reserved** (commented) on `Notification` for a channel that can't fetch (email/SMS); Web Push ignores it, so there's never a second sealing path.
- **Sealed payload** (not in candid): a canister id and a label, no callback URL, frozen at approval.

The design doc is kept separately for now; happy to add it if it helps.


src/internet_identity/internet_identity.did
```diff
@@ -2003,4 +2035,23 @@ service : (opt InternetIdentityInit) -> {
 
     // Looks up identity number when called with a recovery phrase
     lookup_caller_identity_by_recovery_phrase : () -> (opt IdentityNumber);
+
+    notification_send : (
+        chunk_id : blob,
+        headers : NotificationHeaders,
+        recipients : vec NotificationRecipient,
+    ) -> (NotificationSendResult);
```
sea-snake: I don't see a way to send multiple notifications with a single inter canister call?

src/internet_identity/internet_identity.did
```diff
@@ -1558,6 +1558,38 @@ type ListAvailableAttributesError = variant {
...
+type NotificationContent = record {
+    title : text;
+    body : text;
+    url : opt text;
+};
+
+type NotificationRecipient = record {
+    user : principal;
+    // message : opt NotificationContent;
+};
```
sea-snake: This seems to be a vector, but that means that e.g. urgency is the same for all users?

src/internet_identity/internet_identity.did
MRmarioruci: wdym multiple notifications? The campaign lives on the app's side not II's. So it will just say, I wanna reach these users, and then on the SW we pull the pending notifications for any given user.

src/internet_identity/internet_identity.did
MRmarioruci: yes, we could add Notificationheaders per user as well but do you see a use-case for this? Sorting the notifications then becomes a bit more complex, do you order by user or by notification urgency first?

src/internet_identity/internet_identity.did
sea-snake: I was thinking along the lines of a flattened abstract structure:
```
type Notification = record {
    id : blob; // Used to reference the notification in the sender canister
    recipient : principal;

    // Optional attributes that can change over time
    urgency : opt NotificationUrgency;
    expires_at : opt nat64;
};

type NotificationResult = variant {
    accepted;
    rejected : NotificationRejection;
    retry_after_ms : nat32;
}

// II
service : () -> {
    notification_send : (vec Notification) -> (vec NotificationResult);
}

// App
service : () -> {
    notification_content : (id : blob) -> (opt NotificationContent);
}
```

If the service worker pulls the notification content, this means that content can always be updated by the app without contacting II, though it's probably still best to always contact II so the service worker is triggered and updates any already visible notification. The abstract id should probably still be used by II to coalesce so the user doesn't get multiple notifications, just an updated one.

MRmarioruci:
```rust
type Notification = record {
    id : blob;
    recipient : principal;
    urgency : opt NotificationUrgency;
    expires_at : opt nat64;
};

type NotificationSendResult = record {
    accepted : nat32;                                       // just a count
    rejected : vec record { id : blob; reason : NotificationRejection };
    retry_after_ms : opt nat32;
};

notification_send : (chunk_id : blob, notifications : vec Notification) -> (NotificationSendResult);```

Comment (MRmarioruci) [reply src/internet_identity/internet_identity.did]: I like the flattened layout better, but maybe would be nice to keep chunk_id so that we don't have to do O(n) on retry just to see that "all these notifications are already done"?

File: src/internet_identity/internet_identity.did [hunk 3]
```diff
@@ -2003,4 +2032,22 @@ service : (opt InternetIdentityInit) -> {
 
     // Looks up identity number when called with a recovery phrase
     lookup_caller_identity_by_recovery_phrase : () -> (opt IdentityNumber);
+
+    notification_send : (
```
aterga:
```suggestion
    type NotificationSendRequest = record {
         notifications : opt vec Notification;
    };
    notification_send : (
        notifications : NotificationSendRequest
```

src/internet_identity/internet_identity.did
```diff
@@ -2003,4 +2032,22 @@ service : (opt InternetIdentityInit) -> {
...
+        notifications : vec Notification,
+    ) -> (NotificationSendResult);
+
+    notification_consent_status : (origin : text) -> (bool) query;
```
aterga: Let's use the conventional API signature here

aterga: Ditto for all functions that should be called by non-II clients.

PR #4198, opened 2026-08-10
feat(fe): silent re-issue of delegations via ?prompt= and ?hint=

sea-snake:
Internet Identity keeps no session that can authorize anything, so every sign-in costs a passkey — including the one a user does thirty seconds after signing in to a sibling app on the same domain. There is nothing to fall back on: `ii-session-delegations` is accepted by exactly two canister endpoints, both reads.

This keeps the delegation for an app in Internet Identity's own frontend and extends it locally on the next request, so an app that asks can get a delegation again without a ceremony. The user sees an app they already signed in to simply being signed in, instead of another biometric prompt.

Two optional query params on `/authorize`, both borrowed from OpenID Connect:

| Param | Effect |
| --- | --- |
| `prompt=none` | Answer from the cached delegation, or error without ever rendering UI. The app promises it can handle the failure. |
| `prompt=login` | Run a full ceremony, ignoring anything cached. |
| `hint=<principal>` | The principal the app last received, picking between cached delegations when there is more than one. |

**An absent `prompt` behaves as `login`**, and is also what keeps an app out of the cache entirely: an app that has never heard of the param cannot call `ii-forget-delegation` either, so keeping a delegation for it would leave one nothing clears until it expires, and one it could never use since re-issue needs `prompt=none`. So nothing at all changes for an app that has not opted in, down to the shape of the delegation it receives. Unrecognised values (including `consent` and `select_account`, which are not implemented) are ignored and behave the same way, leaving room for an `auto` value later meaning "re-issue when unambiguous, otherwise show the sign-in screen".

Silence is opt-in rather than the default deliberately. The first version of this followed the OpenID Connect convention and re-issued silently whenever it was unambiguous, which removed the sign-in screen from every return visit. That screen is the only place a user can switch identity or account, and 11 e2e specs exist precisely to exercise it on a return visit. Until there is an escape hatch for switching (`select_account` is the conventional one), silence stays something an app asks for.

Even under `prompt=none` it needs all of: something cached for the effective origin with more than five minutes left, exactly one identity in the last-used list, and the multiple-accounts toggle off for it. A matching `hint` satisfies the ambiguity guards on its own, since a principal names an identity and an account together. Ambiguity is judged from the identity list rather than from what is cached, so a user is never silently signed in as their only cached identity when they have another they might have wanted.

**Why this does not widen the blast radius.** The stored delegation carries exactly the expiration the app's own copy carries, so an attacker with the browser profile already holds an equivalent credential for that same window; the two live and die together. The key pair is non-extractable and confined to one origin and account, and the canister-signed half is inert without it. Re-issuing can only restore a session, never lengthen one, because the extension expires with its parent. How long any of this lasts is the app's own `maxTimeToLive`, visible in the app's code rather than being an Internet Identity policy.

# Changes

- `?prompt=` and `?hint=` read once in the authorize layout, kept in `authorization.store`, and stripped from the address bar so a principal does not linger in history. Mirrored through `sessionStorage` across a 1-click identity provider round-trip: without that, the resumed load would read no `prompt` and could answer a request from a delegation the user had just been made to sign in past.
- New `ii-app-delegations` store, keyed by effective origin. Every alternative origin of a derivation origin shares one entry, because Internet Identity already treats them as one app. Keyed that way because `idb-keyval` has no secondary indexes, so the hot path is a single keyed read and only the cross-origin identity purge scans.
- The delegation handler now has the canister delegate to a key the frontend holds and extends that to the app's session key, instead of delegating straight to the app. Same number of canister calls as before, and it populates the cache as a side effect. Chains gain one hop, but only for a request that sent `prompt`: the popup transport goes from one delegation to two and ICRC-167 from two to three, well inside the limit of ten. A request without `prompt` keeps the shape it has today. If the key cannot be created the delegation goes straight to the app's key as before, simply without a cache.
- `ii-forget-delegation`, a new channel method for an app to drop what Internet Identity holds when it signs the user out. Named "forget" rather than "revoke" because nothing is invalidated: the app's own delegation stays valid until it expires and is the app's to clear. The next request then falls through to the interactive path with no extra state.
- New error code `3002 Interaction required`, with `data.reason` of `login_required`, `account_selection_required` or `consent_required`. ICRC-25's ranges are categorical and 3xxx is user action, so a client that categorises by range reads it correctly. An Internet Identity extension, like the existing non-standard `permissions` field.
- Cached delegations are purged wherever `purgeSession` already runs, at the eight sign-out and identity-removal sites. Deliberately not at the three `ContinueView` sites that only discard a stale session key, which says nothing about the per-origin delegations. At the two sites that navigate away immediately the purges are now awaited, since navigation can cancel a pending IndexedDB delete and a surviving delegation would allow silent re-issue after sign-out.
- The multiple-accounts toggle helper moved out of `ContinueView` now that the delegation handler reads it too.
- `AuthorizeConfig` in the e2e authorize fixture takes `prompt` and `hint`, built into the same query string as `openid` and `sso`. The round the `authorizePage` fixture performs is extracted into an exported `performAuthorize`, because re-authorizing needs one round to leave the delegation and a second to re-issue it, and both should go through the path the fixture uses. `redirect.spec.ts` keeps its two-hop assertion, since none of those flows sends `prompt`; a comment there now says so, so a third hop appearing would be read as II having started caching for apps that never asked.

### Two things worth a reviewer's attention

**Attributes are out of scope, and had to be made to fail fast.** Attributes are certified by the canister over a nonce the app picks per request, so nothing cached can answer them. Left alone, an attributes request under `prompt=none` would block on an authorization that never comes, and on the redirect transport a batch that never completes leaves the user parked on Internet Identity — the one outcome `prompt=none` promises to avoid. It now errors immediately instead. Note the consequence on ICRC-167: a JSON-RPC error aborts the batch, so an app batching a delegation and attributes under `prompt=none` gets an error for both rather than a delegation plus one error. On the postMessage transport the two are independent and it gets exactly that.

**Silent re-issue makes no canister call**, so `delegation_bookkeeping`, `set_account_last_used` and activity stats do not record those sign-ins. Dashboards will under-report by however much `prompt=none` gets used.

# Tests

40 new unit tests across four files:

- The silence rule's truth table: the single-identity case, both denial reasons, and a matching `hint` overriding every ambiguity guard.
- That a read-only restriction survives storage plus a local extension, and that an extension never outlives the delegation it extends. These were the two properties the design rested on that I did not want to take on faith.
- The store: origin isolation, several records per origin, replacement by principal, the five-minute expiry margin, and that purging an identity spans origins while sparing other identities.
- The params: unrecognised `prompt` values ignored, non-principal hints dropped, survival across an identity provider round-trip, and that a stored value cannot leak into an unrelated later flow.
- That an attributes request under `prompt=none` errors rather than blocking.

The 18 `vitest` failures in `findWebAuthnFlows` and `iiConnection` fail identically on `main` at this commit and are unrelated.

One e2e spec drives the silent path end to end on the popup transport: a normal sign-in, then `?prompt=none` with no interaction at all, asserting the same principal comes back. Both transports are covered, differing by one config value. The remaining gap is `ii-forget-delegation`, which has no e2e coverage.

### DIFF
--- a/src/frontend/src/lib/stores/app-delegation.store.test.ts
+++ b/src/frontend/src/lib/stores/app-delegation.store.test.ts
@@ -1,0 +1,170 @@
+import {
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from "vitest";
+import { IDBFactory } from "fake-indexeddb";
+import { createStore as idbCreateStore, clear as idbClear } from "idb-keyval";
+import { ECDSAKeyIdentity } from "@icp-sdk/core/identity";
+import {
+  type AppDelegationRecord,
+  appDelegationsForOrigin,
+  discardAppDelegation,
+  forgetAppDelegations,
+  purgeAppDelegations,
+  storeAppDelegation,
+} from "./app-delegation.store";
+
+const NOW = 1_700_000_000_000;
+const FAR_FUTURE = NOW + 60 * 60 * 1000;
+// Inside the 5 minute margin, so treated as already expired.
+const NEAR_EXPIRY = NOW + 4 * 60 * 1000;
+
+const ORIGIN = "https://docs.example.com";
+const OTHER_ORIGIN = "https://chat.example.com";
+const IDENTITY = BigInt(42);
+const OTHER_IDENTITY = BigInt(43);
+
+const TEST_STORE = idbCreateStore("ii-app-delegations", "origins");
+
+let keyPair: CryptoKeyPair;
+
+const record = (
+  principal: string,
+  overrides: Partial<AppDelegationRecord> = {},
+): AppDelegationRecord => ({
+  principal,
+  identityNumber: IDENTITY,
+  keyPair,
+  chainJson: "{}",
+  expiresAtMillis: FAR_FUTURE,
+  ...overrides,
+});
+
+const principalsFor = async (origin: string): Promise<string[]> =>
+  (await appDelegationsForOrigin(origin)).map(({ principal }) => principal);
+
+beforeAll(async () => {
+  global.indexedDB = new IDBFactory();
+  keyPair = (
+    await ECDSAKeyIdentity.generate({ extractable: false })
+  ).getKeyPair();
+});
+
+beforeEach(async () => {
+  vi.useFakeTimers();
+  vi.setSystemTime(NOW);
+  await idbClear(TEST_STORE);
+});
+
+afterEach(() => {
+  vi.useRealTimers();
+});
+
+describe("appDelegationsForOrigin", () => {
+  it("round-trips a record", async () => {
+    await storeAppDelegation(ORIGIN, record("aaaaa-aa"));
+
+    const [stored] = await appDelegationsForOrigin(ORIGIN);
+    expect(stored.principal).toBe("aaaaa-aa");
+    // The non-extractable key pair has to survive structured clone, which is the
+    // whole reason a record can be stored at all.
+    expect(stored.keyPair.privateKey.extractable).toBe(false);
+  });
+
+  it("keeps origins apart", async () => {
+    await storeAppDelegation(ORIGIN, record("aaaaa-aa"));
+
+    expect(await principalsFor(OTHER_ORIGIN)).toEqual([]);
+  });
+
+  it("holds several records for one origin", async () => {
+    await storeAppDelegation(ORIGIN, record("aaaaa-aa"));
+    await storeAppDelegation(
+      ORIGIN,
+      record("bbbbb-bb", { identityNumber: OTHER_IDENTITY }),
+    );
+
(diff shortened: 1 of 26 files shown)


sea-snake:
Superseded by #4248, which implements silent re-issue on the revocable-session design (#4228) instead of on a client-stored delegation.

The two differ in substance, not just in code: this PR extended a delegation the browser had stored, so what an app ended up holding was still an unrevocable artifact with the original expiry. #4248 answers `prompt=none` with a canister-checked mint from a session record, so re-issue is revocable and `app_revoke_session` ends it. `app-delegation.store.ts` and `forgetDelegation.ts` have no counterpart there by design — an app revokes its own session directly against the canister rather than asking the II frontend to forget something.

`promptParams.ts`, the two authorize-URL parameters and the `interaction_required` outcome carried over to #4248.

Not carried over: `tests/e2e-playwright/routes/authorize/silentReauth.spec.ts` from this branch. #4248 has unit coverage for parameter parsing and session selection but no end-to-end run of the silent flow.

PR #4204, opened 2026-08-12
feat(dapps): add Mingle Chat to the dapps list

hankolious:
Adds **Mingle Chat** to the dapps list so that the sign-in dialog shows the app's name and logo instead of a globe and a bare domain.

### What it is

An end-to-end encrypted messenger running on the Internet Computer — frontend and backend are both canisters. Messages use a Double Ratchet; the initial key exchange is post-quantum (X-Wing hybrid KEM) and bootstrapped through vetKD, so a new contact can be reached before they have ever been online. Beyond 1:1 chats there are groups with sender keys, public communities, voice messages, images and files, and 1:1 voice calls over WebRTC. ICRC-1/ICRC-2 transfers are sent straight from a chat.

**Internet Identity is the only sign-in method.**

### Why two auth origins

Users reach the app under the custom domain `https://chat.mingleberrymedia.com`, but the app pins

```js
derivationOrigin: "https://ncgu7-eyaaa-aaaad-qmaza-cai.icp0.io"
```

so that principals stay stable across both addresses (the canister URL remains reachable as well). Both origins are therefore listed, in the same way as the existing `distrikt` and `NeuronPool` entries. `https://chat.mingleberrymedia.com` is listed in the app's `/.well-known/ii-alternative-origins`.

### Assets

`mingle-chat_logo.webp`, 128×128 with transparency, 6.6 kB — within the size range of the existing logos.

### Checks

Ran the two assertions from `dapps.test.ts` against the modified list: the logo resolves in `$lib/icons`, and every entry (including the new one) has at least one valid `authOrigins` entry. The JSON parses and the diff is limited to the appended entry.

| | |
|---|---|
| Website | https://chat.mingleberrymedia.com |
| Frontend canister | `ncgu7-eyaaa-aaaad-qmaza-cai` |
| Backend canister | `nfhsl-jaaaa-aaaad-qmazq-cai` |

Happy to adjust the wording, the one-liner or the artwork.


sea-snake:

PR #4205, opened 2026-08-12
feat(dapps): add Mingle Cloud Migrator to the dapps list

hankolious:
Adds **Mingle Cloud Migrator** to the dapps list so that the sign-in dialog shows the product's name and logo instead of a globe and a bare domain.

### What it is

A migration tool that reads an existing Web2 codebase and rebuilds it on the Internet Computer — backend, database, auth, files and frontend as Motoko canisters. Customers sign in with **Internet Identity** to run and manage their migrations.

### Auth origin

The app is served from `https://migrator.mingleberrymedia.com` and does **not** pin a `derivationOrigin`, so that single origin is the one Internet Identity sees. No canister URL is listed because none serves the app.

### Logo

`mingle-cloud-migrator_logo.webp`, 128×128, 2.4 kB.

The mark is placed on a light tile instead of being shipped with transparency. The dapp card renders the logo on `bg-bg-primary_alt`, which follows the theme, and the lower half of this particular mark is near-black — on a dark background it would disappear. The tile keeps it readable either way; `rounded-xl` on the `img` still applies.

### Checks

Ran the two assertions from `dapps.test.ts` against the modified list: the logo resolves in `$lib/icons`, and every entry has at least one valid `authOrigins`. The JSON parses and the diff is limited to the appended entry.

Same contributor as #4204 (CLA already signed).


sea-snake:

PR #4209, opened 2026-08-12
test(be): regression coverage for DKIM From-header forgery in email recovery

sea-snake:
## Why
Email recovery could be tricked into treating a genuinely DKIM-signed message as coming from a *different* mailbox in the same signing domain, by carrying two mailboxes in the `From:` header — either as an address-list, or smuggled via an embedded `\r\n` that expands into a second canonical header line while the original signature stays valid. Because DKIM for a shared-domain provider (e.g. Gmail) only proves "some user at this domain sent this," an attacker holding an ordinary signed message from their own account at that domain could resolve the verified sender to a *victim's* address and complete anonymous recovery for the victim's anchor — account takeover with no victim interaction.

The exploitable path was closed in #4203, which routes `From:` extraction through the strict RFC 5322 single-mailbox parser (rejecting address-lists and multi-line values). That fix currently rests on a refactor with only light unit coverage; this PR adds explicit regression tests so it cannot silently regress.

## What
- **Unit** — `extract_from_address` rejects a CRLF-spliced `From:` with `AddressMismatch`; and a byte-equality assertion showing the spliced message reproduces the honest message's DKIM signed-input, documenting *why* the forged message kept a valid signature.
- **Integration (PocketIC)** — reproduces the full attack: anonymous `email_recovery_prepare_delegation` claiming the victim's address, plus a genuinely-signed email whose `From:` is spliced to resolve to the victim. Asserts the flow ends in `Failed(AddressMismatch)`, never reaches `RecoveryReady`, and mints no delegation. A negative control confirms honest single-mailbox recovery still succeeds.

Verified the integration test reaches `RecoveryReady` (i.e. fails) against the pre-fix build and passes against current `main`.

No production code changes — regression coverage only.

### DIFF
--- a/src/internet_identity/src/email_inbound/smtp.rs
+++ b/src/internet_identity/src/email_inbound/smtp.rs
@@ -1162,9 +1162,12 @@ mod tests {
         use serde_bytes::ByteBuf;
 
         // A DKIM-Signature covering From:To (order matters for the hash
-        // input). Tag values other than `h=` are irrelevant to
-        // build_header_hash_input, which only canonicalises the listed
-        // headers plus the signature header with `b=` blanked.
+        // input). Its other tag values (d=, s=, bh=, ...) are arbitrary
+        // only because the *same* DKIM-Signature value is reused for both
+        // the honest and forged messages: build_header_hash_input
+        // canonicalises the signature header itself (with `b=` blanked)
+        // into the signed bytes, so those tags do contribute — identically
+        // on both sides.
         let dkim_value = "v=1; a=rsa-sha256; c=relaxed/relaxed; d=example.com; \
                           s=sel; h=From:To; bh=MTIzNDU2; b=YWJj";
         let sig = crate::dkim::parse_dkim_signature(dkim_value).expect("parse DKIM-Signature");
--- a/src/internet_identity/src/email_inbound/smtp.rs
+++ b/src/internet_identity/src/email_inbound/smtp.rs
@@ -1121,6 +1121,105 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn extract_from_rejects_crlf_spliced_header() {
+        // A `From:` value carrying a second mailbox smuggled behind an
+        // embedded CRLF must not resolve to that second mailbox — it is
+        // not a single mailbox, so
+        // extraction fails closed. Before the single-mailbox parser
+        // this returned the *last* angle-bracket pair, i.e. the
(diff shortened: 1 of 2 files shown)


src/internet_identity/src/email_inbound/smtp.rs
```diff
@@ -1121,6 +1121,105 @@ mod tests {
...
+        // (asserted above), not in the signature check.
+        use internet_identity_interface::internet_identity::types::smtp::{
+            SmtpHeader, SmtpMessage,
+        };
+        use serde_bytes::ByteBuf;
+
+        // A DKIM-Signature covering From:To (order matters for the hash
+        // input). Tag values other than `h=` are irrelevant to
+        // build_header_hash_input, which only canonicalises the listed
+        // headers plus the signature header with `b=` blanked.
```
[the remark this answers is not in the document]
sea-snake: Good catch — you're right, the signature header is canonicalised into the hash input, so those tags do contribute (identically on both sides here, since the same DKIM-Signature value is reused). Fixed the comment in 5bf3c85.

src/internet_identity/tests/integration/email_recovery.rs
```diff
@@ -1784,6 +1784,245 @@ fn fake_dkim_dns_response(txt: &[u8]) -> Vec<u8> {
...
+        &env,
+        canister_id,
+        internet_identity_interface::internet_identity::types::email_recovery::EmailRecoveryGetDelegationArgs {
+            nonce: challenge.nonce,
+            session_key,
+            expiration,
+        },
+    )
+    .expect("get_delegation call failed")
+    .expect("delegation should be retrievable for honest recovery");
```
MRmarioruci, line 2025:
```suggestion
    .expect("delegation should be retrievable for honest recovery");
    
    verify_delegation(&env, user_key, &signed_delegation, &env.root_key().unwrap());
    assert_eq!(signed_delegation.delegation.pubkey, session_key);
    assert_eq!(signed_delegation.delegation.expiration, expiration);
```

MRmarioruci (approved):

src/internet_identity/tests/integration/email_recovery.rs
sea-snake: Good call — done in d870d3d. Bound the returned delegation, and it now verifies against the root key plus asserts `pubkey == session_key` and the matching expiration, so the control proves a valid, correctly-bound delegation rather than just an Ok. Re-ran the integration test green.

PR #4210, opened 2026-08-13
fix(fe): redirect ICRC-25 authorize from legacy domains to id.ai

sea-snake:
# Motivation

A dapp that points the new ICRC-25 transport at a legacy domain — for example `identityProvider: "https://identity.ic0.app/authorize"` instead of the default `https://id.ai/authorize` — kept the user on that domain for the entire sign-in. Everything appeared to work, so nothing surfaced the misconfiguration.

The passkey created that way is scoped to the primary origin's RP ID (`getRpId()` returns `id.ai` on a related origin), but it is recorded with `window.location.origin`, i.e. the legacy domain. On **Access methods** that passkey then shows a "Legacy" badge and the message "This legacy passkey is no longer usable and should be removed" — about the passkey the user is currently signed in with. Removing it is blocked at the same time, so the user is told to delete a working passkey and prevented from doing so.

`/authorize` was the one path excluded from the primary-origin redirect, on the assumption that redirecting would break the ICRC-29 channel. It doesn't. The client posts `icrc29_status` with a `*` target origin, filters incoming messages by source window rather than by origin, and *learns* the signer origin from the handshake response — so a redirect before the handshake completes is invisible to it, and the resulting principal is unchanged (derivation uses the dapp's origin and the II canister, not the domain serving II).

Redirecting *after* the handshake would break it: the client pins the signer origin at that point, ignores responses from the new one, and closes the window when its disconnect timeout elapses. The redirect previously ran in the root layout's `onMount`, which happens *after* the authorize layout establishes the channel in `$effect.pre` — so simply removing the exclusion would have been a race.

# Changes

- Moved the primary-origin redirect out of the root layout's `onMount` into the client `init` hook (`src/frontend/src/hooks.client.ts`), where it runs before any component mounts and therefore before the channel is established. Extracted it to `src/frontend/src/lib/utils/primaryOrigin.ts`.
- Removed `/authorize` from the list of paths kept on the current origin. Every other exclusion is unchanged:
  - `/` + `#authorize` — the legacy transport carries the request across its own redirect and bounces the response back to the legacy origin.
  - `/self-service` — links to itself on every related origin so passkey behaviour can be diagnosed per domain; redirecting would defeat its purpose.
  - `/vc-flow`, `/iframe/webauthn`, `/callback` — all verified to create no passkeys.
- The root layout keeps the page hidden while a redirect is in flight, as before.
- Dropped the `cli.id.ai` special case, which rewrote every hit on that host to `/cli` on the primary origin. The domain is intentionally retired, so its e2e test, the test's helper, and the dev server host entry go with it. `CLI_GENERIC_DERIVATION_ORIGIN` keeps its value — it is a derivation origin seed rather than a served domain, and changing it would change the principal of every generic CLI sign-in.

This stops new passkeys from being stamped with a legacy origin. It does not repair passkeys already created that way — that needs a separate change to rewrite the `origin` metadata on successful authentication at the primary origin, plus recording the RP origin rather than `window.location.origin` at creation time.

# Tests

Added an e2e case to `routes/authorize/legacy.spec.ts`, run against both legacy domains, that drives the ICRC-25 transport at `<legacy domain>/authorize` and asserts:

- the authorize window ends up on the primary origin,
- authorization completes and the dapp receives a valid principal,
- the newly created passkey's `origin` is the primary origin.

The second assertion is what catches a regression in the redirect timing: a client that never completes its handshake would hang rather than fail an assertion.

Removed `cli.spec.ts`'s "cli.id.ai redirects to id.ai/cli" test along with the behaviour it covered.

### DIFF
--- a/src/frontend/src/lib/utils/primaryOrigin.ts
+++ b/src/frontend/src/lib/utils/primaryOrigin.ts
@@ -36,7 +36,7 @@ export const redirectToPrimaryOrigin = (): boolean => {
   if (primaryOrigin === undefined || window.location.origin === primaryOrigin) {
     return false;
   }
-  const { pathname, search, hash, host } = window.location;
+  const { pathname, search, hash } = window.location;
   // Legacy AuthClient, the legacy transport carries the request across the
   // redirect itself and bounces the response back to this origin.
   if (pathname === "/" && hash === "#authorize") {
@@ -46,9 +46,7 @@ export const redirectToPrimaryOrigin = (): boolean => {
     return false;
   }
   const target = new URL(primaryOrigin);
-  // cli.id.ai serves only the CLI authorize entry point, every hit there
-  // lands on /cli on the primary origin, regardless of path.
-  target.pathname = host === "cli.id.ai" ? "/cli" : pathname;
+  target.pathname = pathname;
   target.search = search;
   target.hash = hash;
   redirecting = true;
--- a/src/frontend/src/lib/utils/primaryOrigin.ts
+++ b/src/frontend/src/lib/utils/primaryOrigin.ts
@@ -1,0 +1,57 @@
+import { getPrimaryOrigin } from "$lib/globals";
+
+// Paths that must keep running on the origin they were loaded from.
+const KEEP_ON_CURRENT_ORIGIN = [
+  // Legacy verifiable credentials
+  "/vc-flow",
+  // Support page, it links to itself on every related origin so passkey
+  // behaviour can be diagnosed per domain
+  "/self-service",
+  // WebAuthn iframe used for migration
+  "/iframe/webauthn",
+  // OpenID callback, the redirect_uri is registered per origin
+  "/callback",
+];
+
+let redirecting = false;
+
+/**
+ * Whether {@link redirectToPrimaryOrigin} has started a redirect, used to keep
+ * the page hidden instead of flashing this origin's render before it commits.
+ */
+export const isRedirectingToPrimaryOrigin = (): boolean => redirecting;
+
+/**
+ * Send the page to the primary origin when it was loaded from a related one.
+ *
+ * Called from the client `init` hook, before any component mounts, so it runs
+ * before the authorize layout establishes the ICRC-29 channel in `$effect.pre`.
+ * That ordering is load bearing: the client learns the signer origin from the
+ * handshake response, so a redirect before the handshake is invisible to it,
+ * while a redirect after it leaves the client pinned to the previous origin
+ * until its disconnect timeout closes the window.
+ */
+export const redirectToPrimaryOrigin = (): boolean => {
+  const primaryOrigin = getPrimaryOrigin();
+  if (primaryOrigin === undefined || window.location.origin === primaryOrigin) {
+    return false;
+  }
+  const { pathname, search, hash, host } = window.location;
+  // Legacy AuthClient, the legacy transport carries the request across the
+  // redirect itself and bounces the response back to this origin.
+  if (pathname === "/" && hash === "#authorize") {
+    return false;
+  }
+  if (KEEP_ON_CURRENT_ORIGIN.includes(pathname)) {
+    return false;
+  }
+  const target = new URL(primaryOrigin);
+  // cli.id.ai serves only the CLI authorize entry point, every hit there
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/utils/primaryOrigin.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+  const target = new URL(primaryOrigin);
+  // cli.id.ai serves only the CLI authorize entry point, every hit there
+  // lands on /cli on the primary origin, regardless of path.
+  target.pathname = host === "cli.id.ai" ? "/cli" : pathname;
+  target.search = search;
+  target.hash = hash;
+  redirecting = true;
+  window.location.replace(target);
+  return true;
```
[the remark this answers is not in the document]
sea-snake: Changed to `target.href` in 6b58f68f8 — agreed on consistency with the `.href` call site in `transport/legacy.ts`.

One correction though: this did type-check. `lib.dom.d.ts` declares `replace(url: string | URL): void`, and `npm run check` passed with 0 errors both before and after. So the change is for consistency, not to fix a type error.

vite.config.ts
```diff
@@ -80,7 +80,6 @@ export default defineConfig(({ command, mode }): UserConfig => {
                   {
                     hosts: [
                       "id.ai",
-                      "cli.id.ai",
```
MRmarioruci, line 83: Why was this removed?

sea-snake: It's mentioned in the PR description change list:

> Dropped the cli.id.ai special case, which rewrote every hit on that host to /cli on the primary origin. The domain is intentionally retired, so its e2e test, the test's helper, and the dev server host entry go with it. CLI_GENERIC_DERIVATION_ORIGIN keeps its value — it is a derivation origin seed rather than a served domain, and changing it would change the principal of every generic CLI sign-in.

TL;DR The frontend was adapted to serve the `cli.id.ai` domain and redirect it to `id.ai/cli` for backwards ICP CLI compatibility, but this was never used or needed in the end. So it's removed here to avoid unnecessary dead code.

MRmarioruci (approved):

src/frontend/src/lib/utils/primaryOrigin.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export const redirectToPrimaryOrigin = (): boolean => {
+  const primaryOrigin = getPrimaryOrigin();
+  if (primaryOrigin === undefined || window.location.origin === primaryOrigin) {
+    return false;
+  }
+  const { pathname, search, hash } = window.location;
+  // Legacy AuthClient, the legacy transport carries the request across the
+  // redirect itself and bounces the response back to this origin.
+  if (pathname === "/" && hash === "#authorize") {
```
aterga, line 42: Are we sure this path and hash apply _only_ to the legacy auth client? 

FYI, I'm still finding a lot of recently-built apps that direct users to /#authorize

src/frontend/src/lib/utils/primaryOrigin.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Whether {@link redirectToPrimaryOrigin} has started a redirect, used to keep
+ * the page hidden instead of flashing this origin's render before it commits.
+ */
+export const isRedirectingToPrimaryOrigin = (): boolean => redirecting;
+
+/**
+ * Send the page to the primary origin when it was loaded from a related one.
+ *
+ * Called from the client `init` hook, before any component mounts, so it runs
```
aterga, line 27: Not a great idea to summarize where the function is called from, since call sites are not and cannot be the responsibility of the callee.

Instead, explain in what cases this calling the function makes sense conceptually.

aterga (approved):

src/frontend/src/lib/utils/primaryOrigin.ts
sea-snake: Yes, only legacy AuthClient. Those apps are unfortunately not using latest AuthClient :(

src/frontend/src/lib/utils/primaryOrigin.ts
sea-snake: Given this function is intended to be used in a specific place and strongly tied to that flow, I think an exception to that rule is the better approach here.

PR #4219, opened 2026-08-17
feat(be): notification state, per-app consent, and device-signed VAPID JWTs

MRmarioruci:
Groundwork for push notifications: before II can notify anyone it has to record which apps a user allowed and which devices can be reached. This PR adds that state and its management API — no sealing, sending, or delivery yet. First of a stack.

## Changes

The candid gains the `notification_*` and `webpush_*` methods II's own frontend calls. Each is a thin `#[update]`/`#[query]` in `main.rs` that takes an `anchor_number`, authorizes it through the standard `check_authorization` gate, and delegates into the `notifications` module.

`notifications` splits the way the interface does — generic concerns at the top, channel-specific underneath. `notifications::consent` is channel-agnostic (a user consents to an *app*, not to a channel); `notifications::push` is the Web Push channel, split into `subscription` and `jwt_pool`. Each `pub fn` checks the feature flag, authorizes the anchor, then calls a storage-only helper taking the resolved anchor — so the storage logic is unit-testable off-canister. The storable types mirror that tree under `storage/storable/notifications/`.

State is four standalone stable maps (memory ids 33-36), following the account/application convention (hashed keys, additive on upgrade):
- **subscriptions** — `(anchor, sha256(endpoint))`, evict-oldest at 20 per anchor
- **consent** — `(anchor, sha256(origin))`; the value keeps the plaintext origin plus per-app metadata (granted-at, last-sent, muted) so the notifications page can list and mute each app without a second map
- **recipient index** — `principal -> anchor`, written at consent for the send path to resolve
- **jwt pool** — per device, kept out of the subscription row so listing devices stays cheap

A `notifications_enabled` flag in persistent state gates every method.

For the notifications page, `notification_consented_apps` returns each allowed app with its metadata and `notification_set_app_muted` toggles an app's `muted` flag.

A consent also records **which account** it was granted from. The recipient II notifies is derived from anchor and origin, but the service worker's pull credential is an *account* delegation, and the two are not interchangeable. In-session that costs nothing, because the sign-in knows the account it just used; a device that subscribes later has to mint a credential for an app consented earlier, and without the account it can only guess the default one — authenticating to the app as a different persona, which returns nothing and quietly degrades every notification for that app to the generic one.  `notification_grant_consent` takes it as a trailing optional argument and `notification_consented_apps` reports it back; `None` keeps meaning the default account, so rows written before this read back unchanged. Muting keeps the consent row and its seals but the send path skips a muted app (#4251), so a user can silence an app without having to re-grant it later.

Two design points a reviewer will hit in the code:

- **II holds no VAPID private key.** The IC has no P-256 signing primitive, so rather than a canister-held key each device generates its own VAPID keypair and uploads a batch of pre-signed JWTs; the send path (later PR) spends the one whose window covers now. Only the raw signatures are stored and the compact JWT is reassembled by templating, so the signing-input byte layout is a wire contract with the frontend.
- **Consent status is channel-shaped.** `notification_consent_status` returns a `NotificationConsentStatus` record (consent + which channels can currently reach the user), not a bool, so a second channel (email, SMS) becomes a `NotificationChannel` variant rather than a new method.

There is deliberately no sender-registration endpoint: a sending dApp authorizes its canisters by serving `/.well-known/ii-notification-senders`, which lands with the send path (where the outcall lives).

## Tests

17 unit tests: consent round-trip (incl. recipient index), subscribe/unsubscribe/revoke idempotency, evict-oldest at the cap, and the JWT pool (stored on subscribe, dropped with its subscription or on eviction, replaced on refresh, refused for an unknown endpoint), plus storable round-trips (now carrying the per-app metadata). The candid-compat test confirms the `.did` matches the Rust surface. PocketIC scenarios to follow.

<div align="right">Next: #4229</div>


sea-snake:
From description:
> There is deliberately no sender-registration endpoint: a sending dApp authorizes its canisters by serving /.well-known/ii-notification-senders, which lands with the send path (where the outcall lives).

I don't see this outcall in this PR, also I though the app FE would register canisters for a given user after the authentication flow, by calling the II canister directly?

src/internet_identity/src/storage/storable/application.rs
```diff
@@ -49,6 +49,12 @@ impl StorableOriginSha256 {
         let hash = sha256sum(origin.as_bytes());
         Self { hash }
     }
+
+    /// Lower bound of a per-anchor prefix range scan keyed by
+    /// `(anchor, origin_hash)` — see `notifications::consent::consented_origins`.
+    pub const MIN: Self = Self { hash: [0u8; 32] };
+    /// Upper bound of the same scan.
+    pub const MAX: Self = Self { hash: [0xffu8; 32] };
```
aterga, line 57: Nit. I'm not sure it's worth adding these constants.

src/internet_identity/src/storage/storable/notifications/consent.rs
```diff
@@ -0,0 +1,48 @@
...
+        let mut buffer = Vec::new();
+        minicbor::encode(self, &mut buffer).expect("failed to encode StorableNotificationConsent");
+        Cow::Owned(buffer)
+    }
+
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Self {
+        minicbor::decode(&bytes).expect("failed to decode StorableNotificationConsent")
+    }
+
+    const BOUND: Bound = Bound::Unbounded;
```
aterga: Why not use bounded serialization in this case?

src/internet_identity/src/storage/storable/notifications/push/subscription.rs
```diff
@@ -0,0 +1,68 @@
...
+        let mut buffer = Vec::new();
+        minicbor::encode(self, &mut buffer).expect("failed to encode StorablePushSubscription");
+        Cow::Owned(buffer)
+    }
+
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Self {
+        minicbor::decode(&bytes).expect("failed to decode StorablePushSubscription")
+    }
+
+    const BOUND: Bound = Bound::Unbounded;
```
aterga: This should also be bounded probably

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+pub mod consent;
+pub mod push;
+
+use crate::authz_utils::{check_authorization, check_authz_and_record_activity};
+use internet_identity_interface::internet_identity::types::AnchorNumber;
+
+/// Bounds the origin so `delegation::get_principal` can't trap on it.
+pub const MAX_ORIGIN_LEN: usize = 255;
+
+fn validate_origin(origin: &str) -> Result<(), String> {
```
aterga, line 21: Is this function complete enough?

Shouldn't we check for `https`, etc.?

(It's worth searching for other places where we validate origins and considering reusing or copying existing code)

src/internet_identity/src/notifications/consent.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    NotificationConsentStatus {
+        consented: Some(has_consent(anchor_number, origin)),
+        deliverable_channels: Some(channels),
+    }
+}
+
+/// Every origin the caller's anchor has consented to.
+pub fn consented_origins(anchor_number: AnchorNumber) -> Vec<FrontendHostname> {
```
aterga, line 188: This should be a method on Storage, I think

Ideally we should not nest state-accessing helpers like `authorize_query` into other helpers.

src/internet_identity/src/notifications/consent.rs
```diff
@@ -0,0 +1,194 @@
...
+pub fn consent_status(
+    anchor_number: AnchorNumber,
+    origin: FrontendHostname,
+) -> NotificationConsentStatus {
+    if !feature_enabled() || !authorize_query(anchor_number) {
+        return NotificationConsentStatus::default();
+    }
+
+    let mut channels = Vec::new();
+    if super::push::subscription::has_any_subscription(anchor_number) {
```
aterga:
```suggestion
    if has_any_subscription(anchor_number) {
```

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+    Ok(())
+}
+
+fn feature_enabled() -> bool {
+    crate::state::persistent_state(|s| s.notifications_enabled.unwrap_or(false))
+}
+
+/// Server-side kill switch; every entry point checks it first.
+fn ensure_enabled() -> Result<(), String> {
```
aterga:
```suggestion
fn check_enabled() -> Result<(), String> {
```

ensure is typically referring to stateful operations (e.g., if not enabled, then enable)

src/internet_identity/src/notifications/webpush/jwt_pool.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(candid::CandidType, serde::Deserialize, Clone, Copy, Debug, PartialEq, Eq)]
+pub struct JwtPoolStatus {
+    /// Windows still signed and unspent.
+    pub remaining: u32,
+    /// When the device minted this pool; window `i` expires at
+    /// `issued_at_ns + (i + 1) * window`.
+    pub issued_at_ns: Timestamp,
+}
+
+fn replace_jwt_pool(
```
aterga, line 27: Needs a doc string

src/internet_identity/src/notifications/push/jwt_pool.rs
```diff
@@ -0,0 +1,150 @@
...
+    storage_borrow_mut(|storage| {
+        // Refuse to strand a pool with no subscription to spend it on.
+        if storage.push_subscriptions_memory.get(&key).is_none() {
+            return Err("no subscription for that endpoint".to_string());
+        }
+        storage.push_jwt_pool_memory.insert(
+            key,
+            StorablePushJwtPool {
+                signatures: jwt_signatures,
+                issued_at_ns: jwt_issued_at_ns,
```
aterga: perfect match is the best match ((c) Daniel Wong)


```suggestion
                issued_at_ns,
```

aterga: ditto elsewhere

src/internet_identity/src/notifications/push/jwt_pool.rs
```diff
@@ -0,0 +1,150 @@
...
+    jwt_signatures: Vec<Vec<u8>>,
+    jwt_issued_at_ns: Timestamp,
+) -> Result<(), String> {
+    validate_jwt_pool(&jwt_signatures)?;
+    let endpoint_hash = StorableEndpointSha256::from_endpoint(endpoint);
+    let key = (anchor_number, endpoint_hash);
+
+    storage_borrow_mut(|storage| {
+        // Refuse to strand a pool with no subscription to spend it on.
+        if storage.push_subscriptions_memory.get(&key).is_none() {
```
aterga: Having this check may indicate that the underlying model isn't represented via Rust types precisely. 

* The function `replace_jwt_pool` sets a requirement for `storage.push_subscriptions_memory` (the global state) without actually operating on `storage.push_subscriptions_memory`.
* Instead, this function is operating on `storage.push_jwt_pool_memory`, which should be its operational footprint.

A function should not require / assume anything for the memory outside of its footprint.

src/internet_identity/src/notifications/webpush/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+//! VAPID JWT pool. Shared validators and bounds live here.
+use std::ops::RangeInclusive;
+
+pub mod jwt_pool;
+pub mod subscription;
+
+/// Relay endpoints run ~200-300 bytes; capped at 1 KiB.
+pub const MAX_ENDPOINT_LEN: usize = 1024;
+const P256DH_LEN: usize = 65;
+const AUTH_LEN: usize = 16;
```
aterga, line 13: Each const deserves a doc string to explain why the value was picked (unless it's obvious)

src/internet_identity/src/notifications/push/subscription.rs
```diff
@@ -0,0 +1,302 @@
...
+    anchor_number: AnchorNumber,
+    endpoint: String,
+    p256dh: Vec<u8>,
+    auth: Vec<u8>,
+    vapid_public_key: Vec<u8>,
+    jwt_signatures: Vec<Vec<u8>>,
+    jwt_issued_at_ns: Timestamp,
+    now_ns: Timestamp,
+) -> Result<(), String> {
+    validate_param_len("endpoint", endpoint.len(), 1..=MAX_ENDPOINT_LEN)?;
```
aterga: Nit. Since there's so many validation steps, why not validate eagerly as much as we can, aggregate potential errors, and return `Result<(), Vec<String>>` to the caller to facilitate debugging / UX?

src/internet_identity/src/notifications/push/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+const VAPID_PUBKEY_LEN: usize = 65;
+/// Raw ECDSA P-256 signature: r‖s, 32 bytes each.
+const JWT_SIG_LEN: usize = 64;
+/// One JWT per validity window; 30 covers ~30 days. Bounded to cap the row.
+pub const MAX_JWT_POOL_LEN: usize = 30;
+/// Overflow evicts the oldest rather than rejecting: subscriptions are
+/// ephemeral, so eviction keeps the current device working.
+pub const MAX_SUBSCRIPTIONS_PER_ANCHOR: u64 = 20;
+
+fn validate_param_len(name: &str, len: usize, range: RangeInclusive<usize>) -> Result<(), String> {
```
aterga:
```suggestion
fn validate_param_len(observed_len_bytes: usize, allowed_range_bytes: RangeInclusive<usize>, dbg_name: &str) -> Result<(), String> {
```

src/internet_identity/src/notifications/push/subscription.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    storage_borrow_mut(|storage| {
+        let key = (anchor_number, endpoint_hash.clone());
+        let is_new_device = storage.push_subscriptions_memory.get(&key).is_none();
+
+        // A re-subscribe overwrites in place, so only a new device grows the count.
+        if is_new_device {
+            let range_start = (anchor_number, StorableEndpointSha256::MIN);
+            let range_end = (anchor_number, StorableEndpointSha256::MAX);
+            let existing: Vec<(StorableEndpointSha256, Timestamp)> = storage
```
aterga: existing _what_?

src/internet_identity/src/notifications/push/subscription.rs
```diff
@@ -0,0 +1,302 @@
...
+                .push_subscriptions_memory
+                .range(range_start..=range_end)
+                .map(|((_, hash), sub)| (hash, sub.created_at_ns))
+                .collect();
+
+            if existing.len() as u64 >= MAX_SUBSCRIPTIONS_PER_ANCHOR {
+                if let Some((oldest_hash, _)) = existing.into_iter().min_by_key(|(_, ts)| *ts) {
+                    let evicted = (anchor_number, oldest_hash);
+                    storage.push_subscriptions_memory.remove(&evicted);
```
aterga: It's a good idea to always explicitly check the outcome of .remove

src/internet_identity/src/notifications/push/subscription.rs
```diff
@@ -0,0 +1,302 @@
...
+                    // Drop the pool too — it's useless without its subscription.
+                    storage.push_jwt_pool_memory.remove(&evicted);
```
aterga: Ditto

src/internet_identity/src/notifications/push/subscription.rs
```diff
@@ -0,0 +1,302 @@
...
+                }
+            }
+        }
+
+        storage
+            .push_subscriptions_memory
+            .insert(key.clone(), subscription);
+        storage.push_jwt_pool_memory.insert(key, pool);
```
aterga: Should these be two separate memories, or should we merge them into one?

src/internet_identity/src/storage/storable/application.rs
MRmarioruci: They're used down the line in a few places so instead of inlining these everywere I thought constants would be nicer

src/internet_identity/src/storage/storable/notifications/consent.rs
MRmarioruci: Done. Bounded to 384 bytes (255-byte origin + timestamp + cbor map overhead).

src/internet_identity/src/storage/storable/notifications/push/subscription.rs
MRmarioruci: Done. Bounded to 1500 bytes, derived from the field caps.

src/internet_identity/src/notifications/mod.rs
MRmarioruci: Done. Added the https check, and noted the length bound matches delegation::check_frontend_length.

src/internet_identity/src/notifications/consent.rs
MRmarioruci: Done. Moved the scan into Storage::notifications_consented_origins, so authorize_query no longer nests inside another helper.

src/internet_identity/src/notifications/consent.rs
MRmarioruci: Done.

src/internet_identity/src/notifications/mod.rs
MRmarioruci: Done. Renamed to check_enabled.

src/internet_identity/src/notifications/webpush/jwt_pool.rs
MRmarioruci: Done.

src/internet_identity/src/notifications/push/jwt_pool.rs
MRmarioruci: Done.

src/internet_identity/src/notifications/push/jwt_pool.rs
MRmarioruci: Good point. Split it so replace_jwt_pool only touches the pool; the subscription requirement moved into refresh_subscribed_pool, a thin guard that stays testable off-canister.

src/internet_identity/src/notifications/webpush/mod.rs
MRmarioruci: Done. Added a doc string on each.

src/internet_identity/src/notifications/push/mod.rs
MRmarioruci: Done.

src/internet_identity/src/notifications/push/subscription.rs
MRmarioruci: Done. Renamed to existing_devices.

src/internet_identity/src/notifications/push/subscription.rs
MRmarioruci: Done. Capture the removed row and debug_assert it was present, since the key came from the scan just above.

src/internet_identity/src/notifications/push/subscription.rs
MRmarioruci: Same for the pool removal.

src/internet_identity/src/storage/storable/application.rs
MRmarioruci: I use them downstream on other PRs as well. Removing them from here we would have to inline them on about 4-5 other locations. This is cleaner but I get your point. I can remove them

src/internet_identity/src/notifications/push/subscription.rs
MRmarioruci: We could but I would avoid it.
THey share a key but they are written on different rythms. 
- Subcription is only written once in the beginning of the process
- JWTs is a window we have to maintain and refresh/add/remove more frequently. If we merge them then every refresh would rewrite the whole row. Not the end of the world but I do prefer this approach. Also if it turns out that we can use subnet signatures for VAPID then the whole JWT pool will be removed and this is easier to remove.

src/internet_identity/src/notifications/push/subscription.rs
MRmarioruci: Done. add_subscription now runs every validator and returns Result<(), Vec<String>>, so all the bad fields come back at once instead of just the first. Joined them at subscribe_device so the candid boundary stays Err : text, no IDL change.

MRmarioruci:
> From description:
> 
> > There is deliberately no sender-registration endpoint: a sending dApp authorizes its canisters by serving /.well-known/ii-notification-senders, which lands with the send path (where the outcall lives).
> 
> I don't see this outcall in this PR, also I though the app FE would register canisters for a given user after the authentication flow, by calling the II canister directly?

Right that the outcall isn't in this PR. It's on the next ones.

On the FE registering canisters:

I decided to follow the .well-known path for a few reasons.
1. caller() is only the canister principal. And from what I saw the most you can derive from it is <canister-id>.icp0.io URL, not the app's real origin, which can be any domain the app is served on. So we can't map caller() back to the origin ourselves; the origin has to declare its senders. Accepting it as a param I think is a no go security-wise. And with well-known we can also have them list multiple canisters for a given origin. Do you have any suggestions on this?

MRmarioruci:

Consent, the device subscription and the VAPID JWT pool are salvaged into the new consent feature, minus the recipient index and the sender binding, which belong to sending.

PR #4220, opened 2026-08-17
refactor(be): fold the discovery domain into the OpenID anchor lookup

sea-snake:
## What

`lookup_anchor_with_openid_credential` serves two different questions: resolving the anchor a login belongs to, and checking whether a credential is already registered. The first needs the discovery domain, the second does not, and the domain is currently carried around the call rather than being part of it.

- The lookup takes the discovery domain alongside the credential key.
- The registration uniqueness check becomes `is_openid_credential_registered`, returning a bool instead of an anchor number, which is all it ever used.
- `sso_gating::anchor_established_through_domain` was a hand-rolled lookup plus domain comparison; it folds into the shared lookup.
- `update_openid_credential` errors when the stored `sso_domain` disagrees with the incoming one, keeping the stored label consistent with what the lookup keys on.

One place now resolves an anchor from a verified JWT, with the domain part of that lookup rather than applied around it.

## Tests

Existing OpenID and SSO gating suites pass (482 integration, 660 unit), fmt and clippy clean.

### DIFF
--- a/src/internet_identity/src/main.rs
+++ b/src/internet_identity/src/main.rs
@@ -1515,11 +1515,13 @@ mod openid_api {
             Err(err) => return OpenIdResult::Err(err.into()),
         };
         // The verified credential already carries the SSO stable identifier, so
-        // the anchor write below reconciles the stable-id index — an existing
-        // anchor self-heals on a normal sign-in.
+        // the anchor write below reconciles the stable-id index.
         let prepared: Result<OpenIdPrepareDelegationResponse, OpenIdDelegationError> = async {
             let anchor_number = state::storage_borrow(|storage| {
-                storage.lookup_anchor_with_openid_credential(&openid_credential.key())
+                storage.lookup_anchor_with_openid_credential(
+                    &openid_credential.key(),
+                    discovery_domain.as_deref(),
+                )
             })
             .ok_or(OpenIdDelegationError::NoSuchAnchor)?;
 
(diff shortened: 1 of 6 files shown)


src/internet_identity/src/main.rs
```diff
@@ -1515,11 +1515,13 @@ mod openid_api {
             Err(err) => return OpenIdResult::Err(err.into()),
         };
         // The verified credential already carries the SSO stable identifier, so
-        // the anchor write below reconciles the stable-id index — an existing
-        // anchor self-heals on a normal sign-in.
+        // the anchor write below reconciles the stable-id index.
         let prepared: Result<OpenIdPrepareDelegationResponse, OpenIdDelegationError> = async {
             let anchor_number = state::storage_borrow(|storage| {
-                storage.lookup_anchor_with_openid_credential(&openid_credential.key())
+                storage.lookup_anchor_with_openid_credential(
```
MRmarioruci, line 1521: NoSuchAnchor now means two different things: "never seen this credential" and "seen it, but stamped with another domain". The frontend can only act on the first, authFlow.svelte.ts:674 routes it to signup, which then fails at check_openid_credential_is_unique because uniqueness is still domain-blind. So a domain mismatch walks the user through name entry and a captcha before dead-ending on "already registered on this or another anchor", with nothing pointing at the real fix.

Maybe this could get it's own variant? SsoNormalLoginRequired

MRmarioruci (approved):

src/internet_identity/src/main.rs
sea-snake: This will be resolved in a follow-up PR.

PR #4221, opened 2026-08-17
feat(fe): permissionless app metadata for the authorize flow

aterga:
# Motivation

The app name, description and logo shown on the authorize-flow screens ("Continue to _App_", the sign-in header, the redirect animation, the CLI header) are currently sourced from a curated `dapps.json` list shipped inside II itself. Apps not on that list fall back to a bare hostname, and getting on the list requires a change to this repository. This PR replaces that with a permissionless mechanism: any app can provide its own display metadata, with no client-library changes and no involvement from the II team.

# Changes

- **New well-known resource**: apps serve `/.well-known/ii-app-metadata` (JSON with optional `name`, `description`, `logo`) on their own origin, mirroring the existing `/.well-known/ii-alternative-origins` pattern. Works for any web origin, not just ICP canisters, since the file has no principal-derivation security semantics.
- **Fetched from the origin the identity is derived for** — the validated `derivationOrigin` when the authorization request carries one, the requesting origin otherwise. An app publishes once, on the origin its principals and its II accounts are bound to, and every alternative frontend origin it has certified as its own presents identically, with nothing to keep in sync. The screens keep displaying the origin the user signed in from as the trust anchor: where the two differ, the metadata's origin has listed the displayed one in its certified alternative-origins document, so no origin can present itself using metadata of an origin that hasn't vouched for it.
- **`$lib/utils/appMetadata.ts`** fetches and validates the document. Hardened like the alternative-origins fetch (no credentials, no redirects, 10 s budget spanning connection and body) plus:
  - **All-or-nothing validation.** A field that is present but doesn't meet the requirements rejects the whole document rather than being dropped on its own, and the reason is logged naming the field — a half-applied file looks like a working integration while quietly missing a name or logo on a screen the app can't inspect.
  - `name` ≤ 40 and `description` ≤ 120 Unicode code points, measured on the value as served, so the published JSON Schema is exactly what II enforces. Characters that can **reorder** text are refused (control characters, the bidi embeddings and overrides U+202A–U+202E, U+FEFF); the bidi marks U+200E/U+200F/U+061C, the isolates U+2066–U+2069 and the zero-width characters U+200B–U+200D are accepted, since mixed-direction and non-Latin names need them (RTL text ending in punctuation, Thai/Khmer line-break opportunities, Persian shaping, emoji sequences). Isolates must be balanced, so an app's string can never run past itself into II's own text, and a field must contain at least one visible character. Whitespace normalization is presentation only, applied after validation.
  - `logo` must resolve to the **same origin** as the document (which also rules out `data:`/`javascript:` schemes and keeps the sign-in private from third parties), be a raster image of an allowed content type, at most 1 MiB and 4096 px per axis.
  - **The logo is never rendered as served.** The response streams through a counting transform into a `Blob` (so the payload stays in the browser's blob store, never a JS buffer), is decoded with `createImageBitmap` — so `Content-Type` is a claim that gets checked, not trusted — then drawn once into a canvas, re-encoded at ≤ 512 px on its longest side and handed to the `` as a `blob:` URL. What renders is a still image II produced itself; `img-src` gains `blob:` accordingly. `image/svg+xml` is not accepted, because a vector image can't go through that step across the browsers this repo supports.
  - Every failure mode (missing file, CORS, timeout, invalid JSON, undecodable logo, …) yields `undefined` instead of an error, so metadata can never break sign-in. Failures of the logo *asset* cost only the logo, since a second network request can fail transiently.
  - The document itself is capped at 8 KiB.
- **`$lib/stores/app-metadata.store.ts`**: per-origin cached store. Resolves synchronously to the curated-list entry (so known dapps never flash an unbranded screen) and updates in place once the fetched document validates; a valid document replaces the curated entry wholesale (the app owns its presentation). The curated fallback is matched against the displayed origin as well as the derivation origin, since a curated entry names the origins an app signs in from and needn't include the one it derives from. It is a fallback only, never a metadata source, and can be removed once major dapps have migrated.
- **UI wiring**: `AuthorizeHeader` (which now also renders the app-provided description, with `dir="auto"` so its direction comes from its own content and stays isolated from the screen), `AuthWizardView`, `ContinueView`, `RedirectAnimationView`, the upgrade panel in `authorize/+page.svelte`, and `CliHeader` all read the store instead of `getDapps()`. The origin badge stays visible everywhere the metadata is shown, and keeps any scheme or port that distinguishes one origin from another (`originLabel`).
- **Spec**: new "App metadata" section in `docs/ii-spec.mdx` with the requirements, a JSON Schema that mirrors them (noting the one rule it can't express, isolate balance), CORS requirements and security considerations.

Not changed: the certified-attributes opt-in gate in `channelHandlers/attributes.ts` (a security decision, not display metadata), the legacy VC flow, and the dapps-explorer/manage screens, which still use the curated list.

Also note: no lingui message IDs were added or changed, so the translation catalogs are untouched.

# Tests

- New unit tests: `appMetadata.test.ts` (37 cases — hardened fetch options, byte caps enforced up front and while streaming, per-field validation and whole-document rejection, the bidi rules in both directions, same-origin logo enforcement including protocol-relative and scheme-downgrade attempts, content-type allowlist, decode failure, dimension cap, downscaling, timeout via fake timers) and `app-metadata.store.test.ts` (9 cases — fallback behaviour, wholesale replacement, per-origin caching, derivation-vs-displayed origin, subscriber updates), plus `originLabel` cases in `urlUtils.test.ts`. 60 tests across the three files.
- The published JSON Schema was validated against the documented behaviour (accepted and rejected samples, including RTL, Thai and emoji names).
- `npm run check` (tsc + svelte-check): 0 errors, warnings unchanged from `main`. `npm run lint:eslint`, `npm run format-check` and `cargo fmt --check`: clean.
- `npm test`: the new tests pass; the only failures in the full suite (`iiConnection.test.ts`, `findWebAuthnFlows.test.ts`, chrome-extension spec) are pre-existing in this environment and reproduce identically on `main` without this change.
- Existing playwright e2e assertions (e.g. "My Test Dapp account") remain valid via the curated-list fallback: the metadata fetch 404s against the test app and falls back exactly as before.

https://claude.ai/code/session_01ApQTUxtZxjzpFsoJqKCXxo

---
_Generated by [Claude Code](https://claude.ai/code/session_01ApQTUxtZxjzpFsoJqKCXxo)_


src/frontend/src/lib/components/ui/AuthorizeHeader.svelte
```diff
@@ -12,8 +12,10 @@
   const { class: className, origin, ...props }: Props = $props();
 
   const hostname = $derived(new URL(origin).hostname);
-  const dapps = getDapps();
-  const dapp = $derived(dapps.find((dapp) => dapp.hasOrigin(origin)));
+  // App-provided (permissionless) display metadata; the hostname badge below
+  // stays visible regardless, as the trust anchor the user can verify.
+  const metadataStore = $derived(getAppMetadataStore(origin));
```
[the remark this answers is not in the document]
aterga: Keeping this as is, deliberately: the invariant this PR enforces is that the metadata is always fetched from the **same origin the screen displays as the trust anchor** — the two must never diverge, otherwise a hostname badge would vouch for another origin's presentation. The consent view's header displays `effectiveOrigin`'s hostname (pre-existing behavior, unchanged by this PR), so its metadata is keyed to that same origin. Fetching from the channel origin while the badge shows the effective origin would break exactly the pairing you're pointing at.

I've updated the spec wording in debbb43 so it states the rule as "fetched from the origin displayed alongside it" rather than overclaiming `event.origin` everywhere. Whether the consent screen should display the channel origin instead of the effective origin is a fair question, but it predates this PR and would change what the badge itself shows, so I'd rather leave it to a separate discussion.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+  const contentType = (result.response.headers.get("content-type") ?? "")
+    .split(";")[0]
+    .trim()
+    .toLowerCase();
+  if (!APP_LOGO_CONTENT_TYPES.includes(contentType)) {
+    return undefined;
+  }
+  if (result.body.byteLength === 0) {
+    return undefined;
```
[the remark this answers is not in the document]
aterga: Addressed the user-visible half in c3108bf: every consumer now swaps a logo that fails to decode for the default icon via `onerror` (tracked per value, so a corrected logo still renders later) — no broken-image state, and the malformed-UTF-8 case on the metadata document itself now rejects via a fatal `TextDecoder`.

I deliberately did not add fetch-time decode probing or intrinsic-dimension caps. Two reasons:

1. **Threat model**: the logo only ever renders in the sign-in flow of the origin that served it — there is no cross-origin exposure, so a pixel-bomb or pathological SVG can only degrade the attacker's *own* sign-in popup, which they already fully control (they opened it and could equally serve a heavy page themselves). No security boundary is crossed; browsers additionally bound decode internally.
2. **Cost**: decode probing (`createImageBitmap`/`Image.decode`) has patchy cross-browser behavior for SVG specifically (the most common logo format), would move the very decode cost being worried about into the fetch path, and isn't observable in unit tests — a lot of machinery for a self-inflicted-only failure mode.

Excluding or transcoding SVG would meaningfully hurt legitimate adoption, so I'd rather keep the allowlist + 256 KiB cap + render-time fallback. Happy to revisit if the team wants a stricter posture.

---

aterga:
Note on the red Canister tests check: the failures are `e2e-playwright (mobile, 6_6)` / `(mobile, 3_6)` — the legacy-domain popup test (`legacy.spec.ts:52`) and the gated-SSO 1-click test (`sso.spec.ts:1129`) timing out on mobile emulation. These are pre-existing flakes, not caused by this PR: the [main-branch run for `21ab317`](https://github.com/dfinity/internet-identity/actions/runs/31816394951) — this PR's exact base commit, without any of these changes — failed with the same two jobs, and both tests pass on desktop and passed on this branch's earlier waves (`fd6faf9`, `c3108bf`) with identical relevant code. Everything else (Rust, frontend checks, all canister integration tests, the other 11 e2e shards) is green on every commit of this PR. I'm re-running the affected jobs until the flake clears.

---

aterga:
@copilot thanks — the last two rounds landed three findings that were all worth acting on, including one real bug. Responses below, all pushed.

**Round 6 — ZWNJ/ZWJ stripped by the sanitizer ([`4ee96e9`](https://github.com/dfinity/internet-identity/commit/4ee96e9)).** You were right, and this was a genuine bug rather than a nit. `-` swallowed U+200C and U+200D, neither of which is a control or bidi character: ZWNJ drives correct shaping in scripts such as Persian, and ZWJ holds emoji sequences together, so `Acme 👩‍💻` silently decayed into `Acme 👩💻`. The range is now `` — zero-width space plus LRM/RLM — and there's a regression case for both joiners. I checked it actually guards: it fails against the old range and passes against the new one.

**Round 5, item 1 — app name shown without its trust anchor ([`dcf8827`](https://github.com/dfinity/internet-identity/commit/dcf8827)).** Confirmed by reading the wrappers: `UpgradeSuccessView` renders no origin at all and `SsoNormalLoginRequired` shows only the SSO org name, so on those screens a permissionless name had nothing vouching for it, and during attribute consent the header shows the effective/derivation origin instead. I took your second suggested option: `upgradePanel` now takes `showsAppOrigin` and only names the app in the wrappers whose header shows that same channel origin (the sign-in wizard and the continue screen); upgrade-success, the SSO fail-safe and attribute consent keep the generic "This app has moved…" copy.

**Round 5, item 2 — untested abort path ([`dcf8827`](https://github.com/dfinity/internet-identity/commit/dcf8827)).** Added two fake-timer cases: a hanging origin has its signal aborted and falls back once the budget elapses, and the signal is still live one millisecond before it — so the timeout can't be removed or mis-scoped without a failure. The budget is now exported as `APP_METADATA_FETCH_TIMEOUT_MILLIS` so the tests assert against the real value.

Still standing by the one thing I declined earlier ([here](https://github.com/dfinity/internet-identity/pull/4221#discussion_r3797727929)): fetch-time logo decode probing and intrinsic-dimension caps. A pathological logo only degrades the sign-in popup of the origin that served it, and probing would penalise SVG, the most common legitimate logo format. The render-time `onerror` fallback plus the content-type allowlist and 256 KiB cap cover the user-visible failure.

On the red Canister tests check: the failure is `sso.spec.ts:1129` (gated non-sub SSO), which fails identically on `main` at this PR's base commit `21ab317` — see [that run's shard 3](https://github.com/dfinity/internet-identity/actions/runs/31816394951). Pre-existing flake, re-running.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,239 @@
...
+  redirect: "error",
+  headers: {
+    Accept: accept,
+  },
+  // do not send cookies or other credentials
+  credentials: "omit",
+  signal:
+    typeof AbortSignal.timeout === "function"
+      ? AbortSignal.timeout(FETCH_TIMEOUT_MILLIS)
+      : undefined,
```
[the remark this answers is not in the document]
aterga: Fixed in [`debbb43`](https://github.com/dfinity/internet-identity/commit/debbb43): switched to `AbortController` + `setTimeout`, the pattern `doh.ts` and `ssoDiscovery.ts` already use for exactly this reason, and kept the timer armed through the body read so a slow origin can't stall past the budget. [`dcf8827`](https://github.com/dfinity/internet-identity/commit/dcf8827) adds two fake-timer tests over that path.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,239 @@
...
+    // whitespace ones (\t \n \v \f \r), which the next step collapses into
+    // regular spaces (stripping them instead would join adjacent words).
+    .replace(
+      // eslint-disable-next-line no-control-regex
+      /[\u0000-\u0008\u000e-\u001f\u007f-\u009f\u200b-\u200f\u202a-\u202e\u2066-\u2069\ufeff]/g,
+      "",
+    )
+    .replace(/\s+/g, " ")
+    .trim();
+  if (cleaned.length === 0 || cleaned.length > maxLength) {
```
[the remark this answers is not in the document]
aterga: Fixed in [`debbb43`](https://github.com/dfinity/internet-identity/commit/debbb43): the limits now count code points (`[...cleaned].length`), matching the documented `maxLength` semantics, with a test asserting a 40-emoji name is accepted and 41 rejected.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,239 @@
...
+  const contentType = (response.headers.get("content-type") ?? "")
+    .split(";")[0]
+    .trim()
+    .toLowerCase();
+  if (!APP_LOGO_CONTENT_TYPES.includes(contentType)) {
+    return undefined;
+  }
+  const buffer = await response.arrayBuffer();
+  if (buffer.byteLength === 0 || buffer.byteLength > MAX_APP_LOGO_SIZE) {
+    return undefined;
```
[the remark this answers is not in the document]
aterga: Fixed in [`debbb43`](https://github.com/dfinity/internet-identity/commit/debbb43): both the document and the logo now go through one capped reader — oversize `Content-Length` rejected up front, then byte-counted chunk by chunk with the stream cancelled the moment the cap is crossed, mirroring `readCapped` in `authCallbacks.ts`. [`fd6faf9`](https://github.com/dfinity/internet-identity/commit/fd6faf9) also cancels the body on the early-return paths so a non-200 or declared-oversize response stops downloading immediately.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,239 @@
...
+    const contentLength = response.headers.get("content-length");
+    if (
+      contentLength !== null &&
+      Number(contentLength) > MAX_APP_METADATA_SIZE
+    ) {
+      return undefined;
+    }
+    const body = await response.text();
+    if (body.length > MAX_APP_METADATA_SIZE) {
+      return undefined;
```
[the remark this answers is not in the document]
aterga: Fixed in [`debbb43`](https://github.com/dfinity/internet-identity/commit/debbb43) alongside the logo cap: the shared reader streams and counts **bytes** (not UTF-16 units), cancelling at the cap, so the documented 8 KiB limit is now what's actually enforced.

---

docs/ii-spec.mdx
```diff
@@ -335,6 +335,64 @@ To prevent misuse of this feature, the number of alternative origins _must not_
...
+
+```json
+{
+  "name": "Example App",
+  "description": "A short tagline shown on the sign-in screen",
+  "logo": "/logo.png"
+}
+```
+
+Internet Identity fetches this document from the origin of the client application (the origin shown to the user, i.e. `event.origin` of the authorization request, not the `derivationOrigin`) when the authorization flow starts. When the document is missing or invalid, Internet Identity falls back to displaying the origin's hostname.
```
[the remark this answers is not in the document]
aterga: Fixed in [`debbb43`](https://github.com/dfinity/internet-identity/commit/debbb43): the spec now documents the transitional curated fallback and that a valid file replaces a curated entry wholesale. [`fd6faf9`](https://github.com/dfinity/internet-identity/commit/fd6faf9) also aligned the wording on which origin the document is fetched from (the one displayed alongside it) and clarified that the origin is shown as its hostname.

---

src/frontend/src/routes/(new-styling)/authorize/views/RedirectAnimationView.svelte
```diff
@@ -49,8 +50,8 @@
           </g>
         </svg>
         <img
-          src={dapp.logoSrc}
-          alt={$t`${dapp.name} logo`}
+          src={dapp.logo}
+          alt={$t`${dapp.name ?? hostname} logo`}
           class="absolute inset-1 size-[84px] rounded-[18px] object-cover"
```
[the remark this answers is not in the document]
aterga: Fixed in [`debbb43`](https://github.com/dfinity/internet-identity/commit/debbb43): the redirect screen now renders the app's hostname in a badge directly under the logo, so the permissionless logo keeps its trust anchor here too. The `window.location.hostname` line below it is unchanged — that one is deliberately II's own origin, in the "Powered by" attribution.

---

aterga:
@copilot round 7's three suppressed comments were all one point, and a fair one: the badge that vouches for the metadata rendered `new URL(origin).hostname`, so `https://example.com` and `https://example.com:8443` — different origins deriving different principals — presented the same anchor, and an `http` origin was indistinguishable from `https`. Fixed in [`4cd337f`](https://github.com/dfinity/internet-identity/commit/4cd337f).

I didn't go all the way to always showing the serialized origin, because that would put `https://` in front of every app name on every sign-in screen — a visible design change to a shared component, affecting every dapp, and the hostname-only badge predates this PR. Instead there's a new `originLabel` helper that shows the *minimum unambiguous* form:

- `https://example.com` → `example.com` (unchanged; the overwhelmingly common case)
- `https://example.com:8443` → `https://example.com:8443`
- `http://example.com` → `http://example.com`

So the familiar badge is preserved while nothing security-relevant is ever dropped. Applied to all three trust-anchor badges you'd flagged across rounds — authorize header, redirect animation, CLI header — with unit tests including the explicit assertion that the two example origins above don't collapse to the same label, and the spec updated to state the rule.

Separately: I'd left your round-1 threads fixed-but-unanswered, which made it look like they'd been ignored. Each one now has a reply naming the commit that fixed it, and the addressed ones are resolved. The two that remain open are the ones where I pushed back rather than changed the code — the `effectiveOrigin` question and logo decode probing — so a human can arbitrate them.

---

aterga:
@copilot round 8's point was right and is fixed in [`b339a81`](https://github.com/dfinity/internet-identity/commit/b339a81) — docs only, no behaviour change.

Two mismatches between the published contract and `cleanTextField`, both now stated explicitly:

- **Empty values.** The schema had no `minLength`, so `""` validated while the implementation drops it. Added `minLength: 1` to all three fields, and the prose now says a field that is empty *after normalization* is dropped (which `minLength` alone can't express, since `"   "` is length 3).
- **When the limits apply.** `maxLength` reads as a constraint on the raw value, but the code checks length after normalization. Worth noting the direction: because normalization only removes or collapses characters, the schema is *stricter* than Internet Identity — a `name` padded with whitespace that collapses under 40 is accepted by the code and rejected by the schema, never the reverse. The section now says so outright, marks the prose as normative and the schema as an authoring convenience, and documents the normalization order (strip → collapse → trim), that ZWNJ/ZWJ survive, and that an over-long field is dropped rather than truncated.

CI on the previous head [`4cd337f`](https://github.com/dfinity/internet-identity/commit/4cd337f) came back **fully green on the first attempt** — all 45 checks, every e2e shard on both projects, including the two that had been flaking.

On the review loop: rounds 5–8 each surfaced exactly one substantive item and they've been narrowing steadily — a real bug (ZWNJ/ZWJ), then a real UX/security gap (origin trust anchor), now documentation accuracy. All are fixed. Unless the next round turns up something of comparable substance, I'll treat the automated pass as converged and leave the PR to human review rather than keep cycling; the two open threads ([`effectiveOrigin`](https://github.com/dfinity/internet-identity/pull/4221#discussion_r3797361853), [logo decode probing](https://github.com/dfinity/internet-identity/pull/4221#discussion_r3797709939)) are deliberate pushbacks waiting on a human call.

---

docs/ii-spec.mdx
```diff
@@ -335,6 +335,69 @@ To prevent misuse of this feature, the number of alternative origins _must not_
 In order to allow Internet Identity to read the path `/.well-known/ii-alternative-origins`, the CORS response header [`Access-Control-Allow-Origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) must be set and allow the Internet Identity origin being used, for example `https://id.ai`, `https://identity.internetcomputer.org`, or `https://identity.ic0.app`.
 :::
 
+## App metadata
+
+Internet Identity displays the name and logo of the client application on its authorization screens (e.g. "Continue to _App_"). Any application can provide this metadata itself — permissionlessly, without being included in any curated list — by serving a JSON document on the URL `/.well-known/ii-app-metadata` of its origin:
```
sea-snake: In case there are multiple origins pointing at a single derivation origin, would this file be stored on the derivation origin or duplicated across the multiple origins? 

For example alternative origins is always on the ii derivation origin since it's a security constraint, while ii callbacks are on each individual origin since that relates to app logic.

I'd argue towards the derivation origin, it's the one single source of truth you're authentication towards, doing so avoids having multiple sibling apps showing different names/logos due to the maintenance burden of keeping those in sync.

docs/ii-spec.mdx
```diff
@@ -335,6 +335,69 @@ To prevent misuse of this feature, the number of alternative origins _must not_
...
+
+Requirements:
+
+- All fields are optional, and unknown fields are ignored. Fields that fail validation are ignored individually; a document without any valid field is ignored entirely (and a valid document replaces any curated fallback entry wholesale).
+- `name` and `description` are normalized before they are checked: control characters and bidirectional-override characters are stripped (the zero-width joiners U+200C and U+200D are preserved, since they are needed for correct shaping in scripts such as Persian and for emoji sequences), runs of whitespace are collapsed to single spaces, and the result is trimmed. A field that is empty after normalization is dropped. The normalized `name` must not exceed 40 characters and the normalized `description` must not exceed 120 characters, counted in Unicode code points; a field that exceeds its limit is dropped rather than truncated.
```
sea-snake: Normalizing is fine, but more importantly, the whole file should probably be rejected is the original input isn't passing requirements to begin with. This enables an app developer to catch the issue and fix the file instead of keeping a broken file around.

docs/ii-spec.mdx
```diff
@@ -335,6 +335,69 @@ To prevent misuse of this feature, the number of alternative origins _must not_
...
+- `logo` must be a URL (relative URLs are resolved against the application origin) pointing to an image _on the same origin_. Internet Identity downloads the logo (it is never hotlinked), so both the metadata document and the logo asset must be readable cross-origin (see the CORS note below). The logo must be served with one of the content types `image/png`, `image/jpeg`, `image/webp`, `image/gif`, `image/avif` or `image/svg+xml`, and must not exceed 256 KiB.
```
sea-snake: I wouldn't use `data:` URL's here, that would live in the JS heap and DOM which is something we should avoid polluting with external data.

Instead I'd recommend to use `blob:` URL's instead , potentially transcode the image with `createImageBitmap` if you want to make sure images are really images and make them non-animated before rendering in an `` tag.

Obviously rendering images inline e.g. svg string is completely out of scope and would be a major security risk.

docs/ii-spec.mdx
aterga: Agreed, and changed in [`2eb1146`](https://github.com/dfinity/internet-identity/commit/2eb1146).

The document is now fetched from the origin the identity is derived for: the validated `derivationOrigin` when the request carries one, the requesting origin otherwise. So an app publishes once, on the origin its principals — and its II accounts — are bound to, and every alternative origin it has certified as its own presents identically. Two things fell out of it that make me think it's the right call beyond the sync burden:

- The primary-account label on the continue screen (`My <app> account`) now comes from the same origin the accounts live under, so it no longer differs depending on which of an app's origins the user signed in from.
- It settles the one Copilot thread I'd left open ([`AttributeConsentView` passing `effectiveOrigin`](https://github.com/dfinity/internet-identity/pull/4221#discussion_r3797361853)): that screen is consistent by construction now, instead of being an exception to a rule.

Two decisions in there worth your eyes:

1. **The badge still shows the origin the user is signing in from**, not the derivation origin — that's the site they're actually on, and it's the pre-existing behaviour of that badge. Where the two differ, the origin serving the metadata has listed the displayed one in its certified `ii-alternative-origins`, so no origin can present itself with metadata from an origin that hasn't vouched for it. Also displaying the derivation origin is a UI change I didn't want to make unilaterally — happy to if you think the user should see it.
2. **No fallback to the requesting origin.** If an app declares a derivation origin and doesn't publish the file there, it gets no metadata even when its own origin serves one. That's the strict reading of "single source of truth" and I think it's right, but it does mean adding a `derivationOrigin` can silently drop a working file — easy to relax if you'd rather have the fallback.

One transitional wrinkle: the curated dapps list is still matched against the displayed origin as well as the derivation origin, because a curated entry names the origins an app signs in from and needn't include the one it derives from — without that, known dapps using a derivation origin would lose their fallback branding until they publish. That's fallback-only; it is never a metadata source.

---

docs/ii-spec.mdx
aterga: You're right, and it's a better contract. Changed in [`2eb1146`](https://github.com/dfinity/internet-identity/commit/2eb1146).

A field that is present but doesn't meet the requirements now rejects the whole document instead of being dropped on its own, and the reason is logged naming the field — otherwise "rejected" is just as invisible to the developer as "half-applied" was, since II silently falls back on a screen they don't control.

The length limits are now applied to the value **as served**, in code points, instead of after normalization, and characters that could make rendered text read differently from its content are rejected rather than stripped. So the published JSON Schema is now exactly what II enforces, and it lost the "deliberately stricter than II" caveat I'd had to write into the previous round: `minLength`/`maxLength` plus a `pattern` for non-blank and a `not`/`pattern` for the forbidden characters. Whitespace collapsing and trimming stay, but strictly as presentation, after validation — they can no longer rescue a value that breaks a rule.

One asymmetry I kept deliberately: failures of the logo **asset** (unreachable, wrong content type, too large, undecodable) still cost the app only its logo, not its name and description. That's a second network request whose failures aren't necessarily authoring mistakes — an app losing its name because its CDN 500'd for a moment seemed worse than rendering it without a logo. Invalid `logo` **values** in the document (not a string, not same-origin, unparseable) do reject the whole document, since those are exactly the mistakes a developer can see and fix.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,324 @@
...
+
+/** Maximum length of the app name in Unicode code points, after whitespace
+ *  normalization. */
+export const MAX_APP_NAME_LENGTH = 40;
+
+/** Maximum length of the app description in Unicode code points, after
+ *  whitespace normalization. */
+export const MAX_APP_DESCRIPTION_LENGTH = 120;
+
+/** Maximum size of the logo asset in bytes (256 KiB). */
```
sea-snake: Is this a reasonable size? I would double check what's a common size nowadays, it could very well be that most images are larger. The dapp logos stored in the II repo itself isn't a reliable source for that, we manually compressed and squeezed out every byte we could from those 😅

docs/ii-spec.mdx
aterga: Good call — switched in [`6d95347`](https://github.com/dfinity/internet-identity/commit/6d95347), and I took the transcode as well.

- **`blob:` instead of `data:`.** The logo reaches the `` as an object URL, so the bytes stay in the browser's blob store instead of being copied into the JS heap as a base64 string and into the DOM as an attribute value. `img-src` gains `blob:` in the frontend CSP; that doesn't make any remote asset loadable — only this origin's own scripts can mint an object URL, and it resolves entirely inside the browser.
- **Transcode.** `createImageBitmap` decodes what was downloaded, so `Content-Type` is now a claim that gets checked rather than trusted, and the bitmap is drawn once into a canvas and encoded again. What the page renders is therefore an image II produced itself: still (an animated source is flattened to its first frame), without whatever else rode along in the original container, rejected above 4096 px per axis on the way in (the 256 KiB byte cap doesn't bound intrinsic dimensions — a small file can declare enormous ones, and the decoded bitmap costs ~4 bytes a pixel), and scaled to at most 512 px on the longest side, which is the most the screens need on a high-density display.
- **No inline SVG** — agreed, never was on the table. But this does cost SVG entirely: `image/svg+xml` is out of the content-type allowlist, because a vector image can't go through `createImageBitmap` reliably across the browsers this repo supports, and keeping it as the one un-transcoded format would mean the guarantees above hold for every logo except the most common one. Apps with a vector logo serve a rasterized copy. That's a real adoption cost, so if you'd rather keep SVG as a pass-through (object URL, `` only, no transcode, no stillness guarantee) say so and I'll put it back.

This also lands what [the Copilot thread on decode probing and dimension caps](https://github.com/dfinity/internet-identity/pull/4221#discussion_r3797709939) asked for and I'd argued against. Your framing is the better one: my objection was about the threat model (a pathological logo only degrades the sign-in popup of the origin that served it), which ignored that the payload doesn't belong in the DOM and heap regardless — and once the bytes are re-encoded anyway, decode verification and the dimension cap come for free.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,324 @@
...
+      // Cancel the stream so the download stops now, not at GC time.
+      await response.body?.cancel().catch(() => undefined);
+      return undefined;
+    }
+    const declaredLength = response.headers.get("content-length");
+    if (declaredLength !== null && Number(declaredLength) > maxBytes) {
+      await response.body?.cancel().catch(() => undefined);
+      return undefined;
+    }
+    const body = await readBodyCapped(response, maxBytes);
```
sea-snake: You're reading the image body into the JS heap here, something I'd recommend to avoid, instead I'd just directly read it into a blob URL.

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -0,0 +1,324 @@
...
+
+/**
+ * Clean up an app-provided text field: normalize whitespace and strip
+ * control characters as well as bidi-override/invisible formatting
+ * characters, which could otherwise be used to visually spoof the sign-in
+ * screen. Returns `undefined` when the value is not a string, is empty after
+ * cleanup, or exceeds `maxLength` code points (no silent truncation: apps
+ * should notice and fix their metadata instead of shipping a clipped name).
+ */
+const cleanTextField = (
```
sea-snake: Are you sure stripping BIDI characters and such doesn't break multi language strings e.g. RTL languages with LTR strings?

docs/ii-spec.mdx
sea-snake: The label thing is fine, derivation origin is often an ugly URL that would just scare users, it's better to show the app URL that the user just came from.

src/frontend/src/lib/utils/appMetadata.ts
aterga: You're right that the bundled logos are a bad reference — I checked them and they're a median of **2.4 KiB**, largest 5.5 KiB across 111 entries, which is nothing like what an app exports. Raised to **1 MiB** in [`826858d`](https://github.com/dfinity/internet-identity/commit/826858d).

The cap also means less than it did before [`6d95347`](https://github.com/dfinity/internet-identity/commit/6d95347): now that the asset is decoded and re-encoded, it bounds the *download* only — what the page holds and renders is II's own copy at no more than 512 px on its longest side, whatever the source weighed. A 1024×1024 PNG with transparency straight out of a design tool is routinely a few hundred KiB and larger square exports pass a megabyte, so 1 MiB should stop being the reason an app loses its logo, and the cost of the headroom is bandwidth on a request that never blocks sign-in.

What actually bounds the work is the per-axis dimension limit (4096), since a small file can declare enormous dimensions and the decoded bitmap costs ~4 bytes a pixel — that one I left alone; the byte cap was never really doing that job.

---

src/frontend/src/lib/utils/appMetadata.ts
aterga: Fixed in [`130f460`](https://github.com/dfinity/internet-identity/commit/130f460).

The response now pipes through a counting `TransformStream` into a `Blob` instead of being read chunk-by-chunk into a `Uint8Array`, so the bytes accumulate in the browser's blob store and only one chunk at a time is ever a JS buffer. That blob then goes straight into `createImageBitmap`, so for the logo the encoded image is never a JS value at all — download to blob store, decode, re-encode, object URL.

I kept the streaming cap rather than `response.blob()` plus a size check, because the cap has to bound the transfer and not just the result (an origin can omit `Content-Length` or use chunked encoding). Erroring the transform aborts the source stream, so it still stops the download the moment the limit is crossed.

The metadata document is the one thing that still becomes a JS string: 8 KiB of JSON has to be parsed, and the fatal-UTF-8 decode is deliberate there.

---

src/frontend/src/lib/utils/appMetadata.ts
aterga: No — you're right, it did break them. Fixed in [`130f460`](https://github.com/dfinity/internet-identity/commit/130f460).

The set was far too broad. What it now refuses is only the characters that can **reorder** text: the embeddings and overrides U+202A–U+202E (plus control characters and U+FEFF). An override makes text render in an order other than the one it's written in, which is the actual spoofing primitive — a name that reads as something it doesn't contain — and the embeddings are deprecated in favour of isolates for the same reason.

Everything a mixed-direction name legitimately needs is accepted now:

- **U+200E / U+200F / U+061C** — zero-width marks that only affect where *neutral* characters (punctuation, digits) land at a direction boundary. This is precisely the `شرکت من!` case, where the `!` would otherwise follow the paragraph direction and end up on the wrong side, and the Hebrew-name-with-a-Latin-word case.
- **U+2066–U+2069** — the isolates, i.e. Unicode's recommended mechanism for embedding a run of unknown direction, rather than the deprecated embeddings.
- **U+200B–U+200D** — I'd also been wrong about U+200B: it's a line-break opportunity in Thai and Khmer, not just padding. (U+200C/U+200D were already allowed for Persian shaping and emoji sequences.)

Two guards came with that, since they're what makes the permissiveness safe:

1. **Isolates must be balanced.** An isolate only contains its contents while it's closed; an unbalanced one runs to the end of the paragraph, which for a name interpolated into one of II's own sentences means past the app's string. Balance is what makes "an app can affect its own name, never the screen around it" actually true.
2. **A field must have at least one visible character**, so a name made only of marks/zero-width characters is treated as absent rather than rendered as a blank.

And the presentation side: the app-provided description now renders with `dir="auto"`, so its direction comes from its own first strong character and it's isolated from the rest of the screen — which also removes the need for an author to reach for a mark at all in the common all-RTL case. Names interpolated into translated sentences can't be isolated per-substring without turning those strings into markup; with the reordering controls refused, the residual effect there is limited to where neutral characters sit.

Tests cover both directions now: the reordering controls rejected, and `شلوم`/`עברית`/Thai/nested-isolate names accepted unchanged. The published JSON Schema follows, and calls out isolate balance as the one rule it can't express.

---

sea-snake (approved):

PR #4224, opened 2026-08-18
docs: design for revocable app sessions

sea-snake:
Stacked on #4222, which supplies the account reference a session is stored on, its caps, the one write path every session change goes through, and the principal index. Documentation only.

An app delegation cannot be revoked for as long as it is valid, which is up to 30 days: the client holds a self-contained canister-signed artifact whose verification never consults the canister again. A user has no way to see where they are signed in, and no way to sign out.

Adds `docs/ongoing/revocable-app-sessions.md` and `docs/ongoing/revocable-app-sessions-spec.md`. The design puts a canister-side session behind the delegation: the record on the account reference, a chain rooted at a key the II frontend holds, refresh as a direct call from the app minting 5-minute delegations, revocation from either side, and a per-identity browser registry that groups a browser's sessions so they can be ended together.

Two things the specification now states that the design did not:

- A session is named by an id from a monotonic allocator that never reissues. That id is the only input to the session seed besides the account's own, which removes the collision argument the seed used to rest on — that no two records could share a browser and a creation round.
- A ceremony is one write. The browser registration, any browser the cap drops, that browser's sessions, the pruning, the reclaiming and the new record go together, so a sign-in that crosses the session cap is one atomic unit rather than two. Storage refuses a state above the cap; which live sessions give way to make room stays with the ceremony, because that is a judgement and not a storage rule.

`prepare_account_delegation` and `icrc34_delegation` are untouched: this adds methods and changes no existing shape.


docs/ongoing/revocable-app-sessions.md
```diff
@@ -0,0 +1,665 @@
...
+    App -->|"app delegation"| DC
+```
+
+Three things deliberately never happen: the dapp canister never talks to II, the app never goes through the II frontend to refresh, and the II frontend never holds or uses the app's key.
+
+### 3.2 One audience per method
+
+| Method | Audience | Authenticated as | § |
+| ------ | -------- | ---------------- | - |
+| `ii_session_delegation` (JSON-RPC) | app frontend | the authorize transport | 6.2 |
```
MRmarioruci: Why not name it app_session_delegation following the same convention as below?

MRmarioruci: Nit. Since you present the details below as well on the API section 3.3, this section seems redundant.

docs/ongoing/revocable-app-sessions.md
```diff
@@ -0,0 +1,665 @@
...
+1. Prune entries whose `valid_till` has passed.
+2. If the list is still at the cap, remove the **least recently used** entry: the smallest `last_refreshed`, falling back to `created_at` for a session that has never refreshed.
+3. Insert.
+
+Blocking would be the wrong failure: the user is trying to sign in on a new browser and the reason it cannot work is internal bookkeeping. Dropping the stalest session costs that browser a ceremony next time, which is the mildest possible outcome.
+
+Evicting on `last_refreshed` rather than `created_at` is what `last_refreshed` buys here beyond the UI. Ordering by creation would drop a months-old session still in daily use in favour of one created an hour ago and never touched again, which is exactly backwards.
+
+The cap also bounds the per-refresh match in §7.2, which walks the row's records recomputing seeds, so it wants enforcing at creation rather than being treated as advisory.
```
MRmarioruci: Don't really get this paragraph

docs/ongoing/revocable-app-sessions.md
```diff
@@ -0,0 +1,665 @@
...
+
+**The locator is not an argument. It arrives as caller info on the ingress message.** II already does this for the gated-SSO session: `openid/sso_bundle.rs` reads a canister-signed bundle off the call with `ic0::msg_caller_info_signer` and `ic0::msg_caller_info_data`, and `prepare_icrc3_attributes` gates `sso:<domain>` attributes on it (`main.rs:2303`). Sessions use the same mechanism with their own bundle.
+
+Reading it mirrors `read_certified_sso_bundle` (`openid/sso_bundle.rs:180`) step for step:
+
+1. `msg_caller_info_signer()` must be `ic_cdk::id()`. Anything else, or nothing attached, is authorization failing.
+2. Reject on size before allocating, as `MAX_SSO_BUNDLE_BYTES` does.
+3. Decode with a domain-separator prefix and length-prefixed fields.
+4. Check the bundle's own expiry.
```
MRmarioruci: What's the expiry of the bundle? I guess 5 minutes again?

docs/ongoing/revocable-app-sessions.md
sea-snake: Previous II specific RPC methods were already prefixed `ii_` (and `ii-`...), to clearly indicate they're RPC methods specific to Internet Identity, not e.g. wallets following standards like OISY. 

The `app_` canister method prefix is different in this regard, it scopes methods on the II canister itself to clearly illustrates that these methods are meant to be called by apps, not II's frontend. 

The methods being on the II canister, already scopes them in this regard, while RPC methods go over an abstract interface, standards and library that's used beyond II itself.

docs/ongoing/revocable-app-sessions.md
sea-snake: Bundles themselves don't have a concept like expiry as seen in delegations. With SSO we simply put a time stamp as value in the bundle to introduce such a concept.

Given that every single session delegation has a unique principal and bundles are bound to that same principal, the bundle won't need it's own expiry value.

docs/ongoing/revocable-app-sessions.md
sea-snake: Written into the spec now, so the next reader doesn't have to ask. Short version of what @sea-snake said: the two prefixes answer different questions. A canister method needs its audience in the name because every method sits on the same canister and nothing else says who is meant to call it, hence `app_` (following `mcp_`). An RPC method travels over a transport shared with other providers and standards, so what it has to say is which provider it belongs to — which is why every II-specific RPC method already carries `ii_`.

See "No method serves both frontends" in revocable-app-sessions-spec.md.

sea-snake: Fair, and the docs have since been split in a way that removes it: this PR is now a design doc plus a separate implementation spec, and the method-by-audience table only exists in one of them. The design doc no longer carries an API listing at all — it describes the approach, and the spec carries the shapes.

docs/ongoing/revocable-app-sessions.md
sea-snake: Rewritten — that paragraph was doing a bad job of a real point. It now reads:

> The cap also bounds the work every refresh does. A refresh arrives with no session id to look up, so the canister recomputes the seed of each session on the row until one derives to `caller()`. Ten records is therefore the ceiling on that search, which is why the cap is enforced when a session is created rather than treated as advisory.

The thing the original sentence skipped is that refresh has nothing to look a session up *by*: the app sends only its own key, so the canister finds the session by recomputing candidate seeds and comparing. The cap is what stops that search growing.

docs/ongoing/revocable-app-sessions.md
sea-snake: Good catch — this exposed a leftover. The doc said the bundle has no expiry in one place and still listed "check the bundle's own expiry" as a step in another, copied from the SSO bundle it mirrors. The step is gone, replaced by why there isn't one:

> There is no expiry step, unlike the SSO bundle, which carries a timestamp of its own. A bundle is bound to one session's principal, and the session record it names is read on every refresh, so its lifetime is already the session's.

Matches what @sea-snake said, and the implementation carries no expiry field either — the only thing in the bundle is the account principal.

PR #4226, opened 2026-08-18
fix(be,fe): scope account delegations with an optional targets allowlist

MRmarioruci:
Account delegations couldn't be restricted to specific canisters: the delegation's `targets` field was always empty, so any session delegated for an app could call every canister that identity can reach. This adds an optional allowlist so a delegation can be scoped to named canisters, which the IC enforces at the request boundary. Nothing scopes delegations yet — this is the capability the push pull credential (a later PR in this stack) needs to be restricted to just the sending dApp's canister.

Stacked on #4219.

## Changes

The candid `prepare_account_delegation` and `get_account_delegation` gain a trailing `targets : opt vec principal`. The thin `main.rs` wrappers pass it into `account_management`, where the principals' bytes are folded into the signed delegation message through the existing `delegation_signature_msg_with_permissions` — the same helper that already carries `permissions`. `add_delegation_signature` grows a `targets` param as a sibling to `permissions`; `get_account_delegation` reconstructs the identical message and echoes the targets in the returned `Delegation`. Prepare and verify must derive byte-identical messages, so both build the target bytes the same way.

Everything else passes no targets, so behavior is unchanged: the five other `add_delegation_signature` callers (session, MCP registration, OpenID, email recovery), the two legacy delegation endpoints and the MCP delegation flow all pass `None`, and the frontend account and CLI callers pass an empty list.

## Tests

A unit test pins that a non-empty `targets` changes the delegation signable, that the same targets produce the same signable, and that target order is part of it — so any drift in the composition (which would silently break target-scoped delegations against the replica) fails the build. The prepare-then-get round trip is PocketIC-tier and follows.

<div align="left">Previous: #4219</div>

### DIFF
--- a/src/frontend/src/lib/generated/internet_identity_idl.js
+++ b/src/frontend/src/lib/generated/internet_identity_idl.js
@@ -1028,6 +1028,7 @@ export const idlFactory = ({ IDL }) => {
           SessionKey,
           Timestamp,
           IDL.Opt(Permissions),
+          IDL.Opt(IDL.Vec(IDL.Principal)),
         ],
         [
           IDL.Variant({
@@ -1304,6 +1305,7 @@ export const idlFactory = ({ IDL }) => {
           SessionKey,
           IDL.Opt(IDL.Nat64),
           IDL.Opt(Permissions),
+          IDL.Opt(IDL.Vec(IDL.Principal)),
         ],
         [
           IDL.Variant({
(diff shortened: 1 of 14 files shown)


MRmarioruci:
Closing — superseded by the revocable-app-sessions design (docs/ongoing/revocable-app-sessions.md).

The push pull credential will be a notification session carrying an II-signed attribute bundle (the `read_certified_sso_bundle` / `prepare_sso_attr_bundle` machinery), presented as ingress caller-info via `AttributesIdentity`. Scoping comes from that bundle (II vouches the app principal), not a per-request `targets` param — and revocable-sessions restructures `prepare_account_delegation`/`get_account_delegation` (positional → record, `session` field) anyway, so this param would be reworked out. Per-origin principal derivation already prevents cross-app impersonation, so the standalone scoping this added was marginal.

Branch `fix/account-delegation-targets` kept for reference; delete when convenient.

PR #4229, opened 2026-08-18
feat(be): seal Web Push payloads (RFC 8291), eager at consent

MRmarioruci:
Second in the push-notifications stack. Before II can deliver a push, the payload has to be sealed so the relays (Apple/Google/Mozilla) can't read it. This PR adds RFC 8291 payload sealing, computed once when consent is granted (per origin) and reused for every send.

Stacked on #4219.

## Changes

`notifications::push::rfc8291` encrypts a payload under a device's `p256dh` + `auth` keys into the `aes128gcm` wire format browsers expect — ECDH to a fresh ephemeral key, HKDF, AES-128-GCM. `notifications::push::seal` orchestrates it: `grant_consent` (now async) seals every subscribed device of the anchor for the newly-consented origin, and `subscribe_device` (now async) reseals the new device for every consented origin. Both draw randomness the way the rest of the canister does — `ChaCha20Rng::from_seed(random_salt().await)` — and only reach for it when there is actually something to seal. The blobs live in a new stable map keyed `(anchor, sha256(endpoint), sha256(origin))` at memory id 37, so the send path (a later PR) just reads a ready blob rather than paying ~35M instructions per message.

The sealed payload is content-free: `{"o":"<origin>"}` — only the origin to pull for, no canister (dropped per the routing decision), so relays learn nothing and a forged ping can only name a consented origin. Seals are invalidated wherever the inputs change: re-subscribe reseals the device (its keys may have rotated), unsubscribe and cap-eviction drop the device's seals, and revoke-consent drops the origin's seals.

No candid change — the endpoints became async internally but their signatures are unchanged.

## Tests

Unit tests cover the RFC 8291 round trip (short and long payloads, invalid device key, oversized plaintext), the storable round trip, that the routing payload carries only the origin, and that drop-device / drop-origin invalidation removes exactly the right seals. The eager seal-then-send round trip needs `raw_rand`, so it lands as a PocketIC scenario with the rest of the send path.

<div align="left">Previous: #4219</div>

<div align="right">Next: #4251</div>


sea-snake:
@MRmarioruci Quick question, why is the origin the notification payload? To pull data, it does that from a canister right?

MRmarioruci:
> @MRmarioruci Quick question, why is the origin the notification payload? To pull data, it does that from a canister right?

Yes, the canister is where we pull the data from. But the service worker already knows which canister that is, because we cached it at consent time. We don't put it in the notification.

Mainly about security/trust.
If we did put the canister in the notification, an attacker (who has somehow intercepted both VAPID JWT and also knows about the device endpoint) could mint a notification for a legit origin but point it at their own canister. The SW would then pull from (and sign into, with the user's per-app credential) the attacker's canister.

So the flow is:

At consent we store, for this app: origin, the canister to pull from, and the pull credential.
At send we deliver only the sealed origin.
The SW looks up the canister + credential for that origin locally, and pulls

sea-snake:
@MRmarioruci How would the frontend know which notification id to pull?

MRmarioruci:
> @MRmarioruci How would the frontend know which notification id to pull?

1. Ping wakes up the service worker
2. Service worker pulls "all that's pending for me".
3. The SW dedups locally on the notification id. And only shows/renders the ones that haven't been shown before.
4. In the meantime if you get another 2 pings, they will just be ignored cause you are already showing the notifications using the first ping
5. A ticker will run on the client library to eventually remove the stored notifications on the app's side. Depending on urgency we could have something that clears every 5 minutes or so.

src/internet_identity/src/notifications/webpush/seal.rs
```diff
@@ -0,0 +1,249 @@
...
+/// service worker wakes on this ping and fetches it from the app directly.
+fn notification_payload(origin: &str) -> Vec<u8> {
+    serde_json::to_vec(&serde_json::json!({ "o": origin }))
+        .expect("serializing a notification payload cannot fail")
+}
+
+fn seal(rng: &mut ChaCha20Rng, p256dh: &[u8], auth: &[u8], origin: &str) -> Option<Vec<u8>> {
+    let auth: &[u8; rfc8291::AUTH_SECRET_LEN] = auth.try_into().ok()?;
+    rfc8291::encrypt(&notification_payload(origin), p256dh, auth, rng).ok()
+}
```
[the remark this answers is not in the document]
MRmarioruci: Fixed. add_subscription now validates that p256dh is a real SEC1 P-256 point, so an unsealable key is rejected with an error at subscribe rather than silently storing nothing. The remaining .ok() only covers the ~2^-256 invariant failures.

src/internet_identity/src/notifications/webpush/seal.rs
```diff
@@ -0,0 +1,249 @@
...
+
+/// Seal every subscribed device of `anchor` for `origin`. Called when consent
+/// is granted, so the send path finds a ready blob for each device.
+pub async fn seal_devices_for_origin(
+    anchor_number: AnchorNumber,
+    origin: &FrontendHostname,
+    now_ns: Timestamp,
+) {
```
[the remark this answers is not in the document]
MRmarioruci: Added. Split the eager-seal path into sync cores (seal_devices_into / reseal_device_into) so it is testable without raw_rand, then covered that consent seals every subscribed device under the right (anchor, endpoint, origin) key.

src/internet_identity/src/notifications/push/subscription.rs
```diff
@@ -62,10 +63,11 @@ pub(super) fn add_subscription(
...
             if existing.len() as u64 >= MAX_SUBSCRIPTIONS_PER_ANCHOR {
                 if let Some((oldest_hash, _)) = existing.into_iter().min_by_key(|(_, ts)| *ts) {
-                    let evicted = (anchor_number, oldest_hash);
+                    let evicted = (anchor_number, oldest_hash.clone());
                     storage.push_subscriptions_memory.remove(&evicted);
-                    // Drop the pool too — it's useless without its subscription.
+                    // Drop the pool and seals too — useless without the subscription.
                     storage.push_jwt_pool_memory.remove(&evicted);
+                    drop_device_seals_in(storage, anchor_number, &oldest_hash);
                 }
```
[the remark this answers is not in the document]
MRmarioruci: Added eviction_drops_the_evicted_device_seals: it seals the oldest device, forces the 21st subscribe, and asserts those seals are gone.

src/internet_identity/src/storage.rs
```diff
@@ -1599,6 +1620,94 @@ impl<M: Memory + Clone> Storage<M> {
...
+        &mut self,
+        key: (AnchorNumber, StorableEndpointSha256, StorableOriginSha256),
+        seal: StorablePushSeal,
+    ) {
+        self.push_seal_memory.insert(key, seal);
+    }
+
+    /// Drops every seal of one device. A device's seals are a contiguous range,
+    /// since the endpoint precedes the origin in the key.
+    pub fn remove_push_seals_for_device(
```
```suggestion
    pub fn remove_all_push_notification_seals_for_device(
```

src/internet_identity/src/storage.rs
```diff
@@ -1599,6 +1620,94 @@ impl<M: Memory + Clone> Storage<M> {
...
+            anchor_number,
+            endpoint_hash.clone(),
+            StorableOriginSha256::MIN,
+        );
+        let end = (
+            anchor_number,
+            endpoint_hash.clone(),
+            StorableOriginSha256::MAX,
+        );
+        let keys = self.push_seal_memory.range(start..=end).map(|(key, _)| key);
```
```suggestion
        let keys = self.push_seal_memory.range(start..=end).keys();
```

src/internet_identity/src/storage.rs
```diff
@@ -1599,6 +1620,94 @@ impl<M: Memory + Clone> Storage<M> {
...
+        self.remove_push_seals(keys.collect());
+    }
+
+    /// Drops every device's seal for one origin. The origin is the last key
+    /// component, so this scans the anchor's seals and filters.
+    pub fn remove_push_seals_for_origin(
```
MRmarioruci: Same as above - naming

src/internet_identity/src/notifications/webpush/rfc8291.rs
```diff
@@ -0,0 +1,391 @@
...
+        &salt,
+    );
+
+    // AES-128-GCM sealed body: plaintext || 0x02 || <no padding>.
+    // The 0x02 delimiter indicates "last record" per RFC 8188 §2.
+    let mut sealed_input = Vec::with_capacity(plaintext.len() + 1);
+    sealed_input.extend_from_slice(plaintext);
+    sealed_input.push(0x02);
+
+    let cipher = Aes128Gcm::new_from_slice(&cek).map_err(|_| EncryptError::AeadFailure)?;
```
MRmarioruci: Double check if there is an existing rust library

src/internet_identity/src/notifications/webpush/rfc8291.rs
```diff
@@ -0,0 +1,391 @@
...
+// The `p256` + `aes-gcm` crates surface a few APIs that are deprecated
+// as part of RustCrypto's transition to `generic-array` 1.x. The
+// replacements aren't yet stable across the versions II pins, so we
+// silence them for this module — cleanup rides on the ecosystem's
+// generic-array 1.x bump.
+#![allow(deprecated)]
+
+use aes_gcm::{
+    aead::{Aead, KeyInit, Payload},
+    Aes128Gcm, Nonce,
```
MRmarioruci, line 39: Analyze and justify that this is a good dependency.

src/internet_identity/src/notifications/webpush/seal.rs
```diff
@@ -0,0 +1,317 @@
...
+    storage: &mut Storage<DefaultMemoryImpl>,
+    rng: &mut ChaCha20Rng,
+    anchor_number: AnchorNumber,
+    origin: &str,
+    devices: &[(StorableEndpointSha256, Vec<u8>, Vec<u8>)],
+    now_ns: Timestamp,
+) {
+    let origin_hash = StorableOriginSha256::from_origin(&origin.to_string());
+    for (endpoint_hash, p256dh, auth) in devices {
+        if let Some(blob) = seal(rng, p256dh, auth, origin) {
```
MRmarioruci: Dont fail silently if the seal fails. Propagate the errors to the caller and handle them

MRmarioruci: Maybe an error message for the user if sealing fails for give device.

src/internet_identity/src/notifications/webpush/rfc8291.rs
MRmarioruci: it is

MRmarioruci:

RFC 8291 sealing is salvaged into the new consent feature as-is.

PR #4230, opened 2026-08-18
docs: one-pager tying the session designs together

sea-snake:
Documentation only.

Four stacked design docs is a lot to hand someone who just wants to know what is changing and why. Adds `docs/ongoing/shared-revocable-sessions.md`: the glanceable version, with diagrams for today and after, sibling-subdomain sharing, and a table of which methods are public, internal and untouched, linking into the detail — the four designs, their specifications, the scenarios they must satisfy in a browser, and the dashboards that would watch them.

The two properties it exists to make obvious: a stolen delegation is worth at most 5 minutes instead of up to 30 days, and a session cannot renew or clone itself, because creating one needs an access method a session does not have.


docs/ongoing/shared-revocable-sessions.md
```diff
@@ -0,0 +1,157 @@
...
+
+When a user signs in to an app, II gives it a signed statement that a key may act for the user until a stated expiry, for up to 30 days. Anything on the network can check that statement without asking II, which is what makes it cheap, and also means II cannot take it back. Signing out of the app clears the app's own storage and invalidates nothing. Separately, an app running as several subdomains cannot share one sign-in, so the user signs in on each and signing out of one leaves the others signed in.
+
+Three designs together replace that. The canister keeps a session for the account the user signed in with, and the app holds a chain rooted at it that can only be used to ask II for delegations. The app mints itself a 5-minute delegation whenever it needs one. Deleting the session stops further mints, so access ends within five minutes. Because the canister now records which browser each session came from, a user can see the browsers they are signed in from and sign one out of every app at once. And because subdomains that share a derivation origin resolve to the same account, they share one session, so signing in on one lets the others get delegations without another sign-in.
```
MRmarioruci: > The canister keeps a session for the account
Is this referring to the II backend or the app's backend canister?

sea-snake: The II canister. Fixed here and everywhere else the page said a bare "canister" — five places, including the summary line you flagged, the approach section, and the before/after table. The app's own canister never talks to II at all, so it was worth making unambiguous rather than leaving to context.

PR #4232, opened 2026-08-19
refactor(be): one read and write path for account reference lists

sea-snake:

The account reference list is written at six sites, and five counters plus a global gauge are derived from it at each one. Keeping them correct is a convention rather than a guarantee, and the PRs above this add two more write sites, a reverse index, and sessions on every reference.

**One write path.** `write_reference_list` is the only place a reference list is stored. It reads the previous list, diffs it against the new one, and derives every counter from that diff; `update_counters` and `AccountType` are gone. The deltas are defined over the list itself, so the counter is derived from what is stored rather than maintained beside it — and the repair path that existed because nobody could trust it, `rebuild_identity_account_counters`, goes with it. What a drifted counter would cost is written down in the test that replaces it.

**One read path.** `account_references` replaces `lookup_account_references`, `find_account_references` and `find_account_reference` — three overlapping readers, one of which returned the application number it had been handed. `account_reference` sits beside it for the single-reference case; answering `None` there *is* the ownership check, since an account belongs to whichever identity's list names it.

**One checked way to store a list.** `StorableAccountReferenceList` had an infallible `From<Vec<AccountReference>>`, so what may be stored was decided at each write site. `TryFrom` replaces it and the unchecked path no longer compiles; `Default` went too, since it handed out the empty list the constructor exists to refuse.

An empty list is a tombstone: it says every reference at the origin was moved away and its default account must never be derived again. Only a move may create one, and nothing moves accounts yet, so an empty list is a bug rather than an intent.

Non-emptiness is the *only* rule there, deliberately. Decoding stays infallible, so a rule added to that constructor also applies to every future write of an existing list — a stored list breaking one would become unwritable, which for a reference list means an identity locked out of that origin. Emptiness is the only rule that can be shown never to have been written: `create_additional_account` always pushes at least one reference, `create_default_account` always has one, and the read-modify-write paths write back what they read.

**The refusals that could leave half a write behind now run before it.** Returning `Err` on the IC commits every write that came before it. This is not yet a general property of the create paths — `allocate_account_number`, the account insert and the config write all still precede `write_reference_list`, so a refusal there would commit an allocated number and a stored account that nothing references. I traced the errors reachable from that call in these two paths and believe none of them can fire today: `Empty` is impossible because the list always has at least one entry, `OriginNotFoundForApplicationNumber` because `lookup_or_insert_application_number_with_origin` guarantees the list, and `AccountCounterOutOfBounds` needs a `u64` overflow on a `+1` delta. What remains is `ErrorUpdatingAccountCounter` from a stable-memory grow failure, which is the same exposure as before this PR. What *is* fixed is the set of refusals that were reachable:

- The application list is resolved, and the storable form built, before the counters move.
- `create_default_account` decides which reference the new number goes into up front. Its `MissingAccount` refusal used to leave an allocated account stored, and the config pointing at it, with no reference naming it.
- `update_account` read the account it was about to rename with an `expect` whose message said the `None` was impossible. It is reachable two ways — an identity naming an account it holds no reference to, and a second numberless update of a default the first one already named — and the integration tests asserted the trap. It answers `UpdateAccountError::Unauthorized` instead, which is what `prepare_account_delegation` already answers for the same read.
- The `last_used` stamp on the delegation path moved above `add_delegation_signature`/`update_root_hash` and propagates. Propagating it where it stood would have been worse than the `let _`: an error would have been reported for a delegation whose signature was already in the map.

**Writes are narrowed to what actually changed.** Renaming an account never touched its reference, but went through a helper that always wrote the list back, so every rename rewrote the whole `(anchor, application)` blob with byte-identical content. It writes only the account record now. The counter refusal also names which counter and count diverged, with the value stored and the delta that would not fit — the write is refused, so the counters keep the values that disagreed and the error is the only place that information exists.

The derived counters equal what the six call sites computed, site by site. What changes: the error paths above, and `update_account` refusing rather than trapping.

Tests: `reference_list_write_path_tests` (10), including written counters equalling a rebuild over a mixed workload, a write against an unknown application writing **nothing at all** even on a zero-delta write, and the two counts moving independently so a refusal names the one that failed. `account_reference_state_tests` (8) pins what an absent list, an empty list and a list holding references each mean across `read_account`, `list_accounts`, `create_additional_account` and `update_account` — a tombstone never regains a default, naming a default keeps its reference in place with its `last_used`, an identity holding no reference can neither rename nor stamp the account, and a rename writes only the record.


src/internet_identity/src/storage/storable/accounts_counter.rs
```diff
@@ -13,29 +13,14 @@ pub struct StorableAccountsCounter {
...
-                    .checked_add(1)
-                    .expect("overflow in stored_accounts"),
-                stored_account_references: self.stored_account_references,
-            },
+    pub fn increment_accounts(&self) -> Self {
+        Self {
+            stored_accounts: self
+                .stored_accounts
+                .checked_add(1)
+                .expect("overflow in stored_accounts"),
```
aterga: Would it be easy to making the code panic-free in this PR?

sea-snake: The field `stored_accounts` is a u64 value, it actually overflowing should be practically unreachable and cannot be considered a user error. 

Plumbing this unreachable error through to the candid response would introduce an error that the frontend can't handle and should never expect. Similar to the salt not being set.

There's no handling of the error either in the canister, a genuine overflow would not be resolvable since it would lead to an invalid count e.g. stop counting or resetting the count. 

I'm open to suggestions though :D

sea-snake: Update, seems like I can plumb it through to the generic `StorageError` which is already intended for these edge cases.

sea-snake: Done, both writes to the counter return now instead of trapping.

The overflow check moved into `allocate_account_number`, which already returns `Result<_, StorageError>`, with a new `AccountsCounterOverflow` variant. That let `increment_accounts` go away entirely, so `StorableAccountsCounter` is a plain storable type again like everything else under `storable/`, and the check sits on the line that produces the account number instead of one indirection away.

Worth being explicit about why it refuses rather than saturating, since saturating looks like the tidier way to be panic-free: `allocate_account_number` takes the account number *from* the counter, so at the ceiling saturating would hand out the same number forever, and since a named account's principal is `calculate_account_seed(account_number, origin)`, two accounts at one origin would collide on a principal. Refusing is the only safe option there, and there's a comment saying so.

One more I'd missed, which matters more than the overflow: `apply_reference_counter_deltas` wrote the global counter with an `expect`. That's a regression in this PR rather than existing behaviour, `main` maps the same write to `StorageError::ErrorUpdatingAccountCounter` in `update_counters`, so it does that again and the function returns `Result`.

There's now a test that sets `stored_accounts` to `u64::MAX` and asserts the error, so the "unreachable in practice" claim above is at least pinned by something. I couldn't find a way to make the stable cell's `set()` fail from a unit test, so that second path is unverified beyond compiling.

src/internet_identity/src/account_management.rs
```diff
@@ -362,9 +362,9 @@ pub async fn prepare_account_delegation(
     });
     update_root_hash();
 
-    // Update last used timestamp
     storage_borrow_mut(|storage| {
-        storage.set_account_last_used(anchor_number, origin.clone(), account_number, time());
+        let _ =
```
MRmarioruci: Should we propagate this error instead of discarding it? If updating last_used fails, the sign-in still succeeds but II may not record the account use correctly. What would this cause downstream?

src/internet_identity/src/storage.rs
```diff
@@ -1695,43 +1709,92 @@ impl<M: Memory + Clone> Storage<M> {
...
+        let key = (anchor_number, application_number);
+        let previous = self
+            .stable_account_reference_list_memory
+            .get(&key)
+            .map(Vec::<AccountReference>::from)
+            .unwrap_or_default();
+
+        let deltas = ReferenceListDeltas::between(&previous, &current);
+
+        self.stable_account_reference_list_memory
```
MRmarioruci, line 1691: The reference list is saved before the counters are updated. If the counter update fails, the list is saved but the counters are not. Can we make sure these changes succeed or fail together?

src/internet_identity/src/storage.rs
```diff
@@ -1617,17 +1621,25 @@ impl<M: Memory + Clone> Storage<M> {
                     }
                 }
 
-                let value = StorableAccountReferenceList::from_vec(account_references);
-
-                self.stable_account_reference_list_memory.insert(key, value);
+                self.write_reference_list(
```
```suggestion
          let Some(result) = result else {
             return Ok(None);
          };
                self.write_reference_list(
```
If no matching reference was found, account_references is unchanged here. Can we return Ok(None) instead of passing the unchanged list to write_reference_list?

src/internet_identity/src/storage.rs
```diff
@@ -2477,6 +2529,43 @@ impl<M: Memory + Clone> Storage<M> {
...
+            accounts: current_accounts - previous_accounts,
+            references: current_references - previous_references,
+        }
+    }
+
+    fn is_empty(&self) -> bool {
+        self.accounts == 0 && self.references == 0
+    }
+
+    fn apply(&self, accounts: u64, references: u64) -> (u64, u64) {
```
```suggestion
fn apply(
    &self,
    accounts: u64,
    references: u64,
) -> Result<(u64, u64), StorageError> {
    let accounts = accounts
        .checked_add_signed(self.accounts)
        .ok_or(StorageError::AccountCounterOutOfBounds)?;

    let references = references
        .checked_add_signed(self.references)
        .ok_or(StorageError::AccountCounterOutOfBounds)?;

    Ok((accounts, references))
}

// and on the call sites something like
let (stored_accounts, stored_account_references) =
    deltas.apply(existing_accounts, existing_references)?;
```

Maybe it would be better to return an error here instead of clamping the count to zero. If eviction removes a reference while the counter is already at zero, clamping hides the fact that the counter and stored references no longer match

src/internet_identity/src/storage/storable/accounts_counter.rs
sea-snake: Done, as described in the thread above — leaving a note on where it ended up.

`AccountsCounterOverflow` is raised by `checked_add(1)` on the line that produces the account number, and the `set` maps to `ErrorUpdatingAccountCounter`. `increment_accounts` is gone, so `StorableAccountsCounter` is a plain storable type again.

Worth stating why it refuses rather than saturating, since saturating is the tidier-looking way to be panic-free: the counter *is* the account number, so at the ceiling saturating would hand out the same number forever, and a named account's principal is `calculate_account_seed(account_number, origin)` — two accounts at one origin would collide on a principal. Refusing is the only safe answer there.

One path is still unverified beyond compiling: I could not find a way to make the stable cell's `set()` fail from a unit test, so the `ErrorUpdatingAccountCounter` arm has no test. The overflow arm does.

Only the minicbor encode/decode `expect`s remain, which every storable type in the repo has.

src/internet_identity/src/account_management.rs
sea-snake: Already fixed higher in the stack, in #4235 (`6968cf14c`). The call moved above the signature block and propagates:

```rust
storage_borrow_mut(|storage| {
    storage.record_account_use(anchor_number, origin.clone(), account_number, time())
})
.map_err(|err| AccountDelegationError::InternalCanisterError(err.to_string()))?;
```

To answer the downstream question directly, since it is the interesting part: `last_used` is display-only — it is on `AccountInfo` in candid and nothing in storage orders or reaps by it. What made the discard wrong was not the lost timestamp but the position: it ran *after* `add_delegation_signature` and `update_root_hash`, so an error there would have discarded work that had already succeeded. Moving it above them is what makes propagating correct rather than harmful.

src/internet_identity/src/storage.rs
sea-snake: Fixed. The rule is now that every fallible step runs before every write, in both functions:

- `write_reference_list` applies the counter deltas first and inserts the list after.
- `apply_reference_counter_deltas` computes all three counter values, then does the one write that can report a failure (the `StableCell::set`) before the two `insert`s that cannot.

So there is no `Err` path left after any write: either it refuses with nothing written, or every write lands.

Worth being explicit about why the ordering is the whole fix, because it is a bit counter-intuitive: on the IC, returning `Err` **commits** everything written so far — only a trap rolls the message back. So "succeed or fail together" is unreachable while a fallible step sits after a write, and it is `Result`-returning rather than trapping that creates the exposure. Ordering gets it without trapping.

For the record this was not a regression in this PR: `main` maps the same write to `ErrorUpdatingAccountCounter` in `update_counters`, so the hazard already existed — this PR moved it.

src/internet_identity/src/storage.rs
sea-snake: Already in, higher in the stack. #4235 (`6968cf14c`) rewrote `with_account_mut` — the two arms collapsed into one and the early return is there with a comment saying exactly your reason:

```rust
let Some(account_reference) = account_references
    .iter_mut()
    .find(|account_reference| account_reference.account_number == maybe_account_number)
else {
    // `f` never ran, so nothing was modified and writing the list back would
    // store the bytes it already holds.
    return Ok(None);
};
```

The saving is bigger than the call: `write_reference_list` re-reads the previous list, computes the deltas and rewrites the whole `(anchor, application)` blob. The deltas come out empty so the counter work early-returns, but the blob rewrite happened anyway — a stable write for a no-op, on a row that is rewritten whole on any change.

src/internet_identity/src/storage.rs
sea-snake: Taken, near enough verbatim — `checked_add_signed` with a new `StorageError::AccountCounterOutOfBounds`, and `apply` returns `Result`.

You undersold why it matters. The clamped value is not only recorded, it is *acted on*:

```rust
if application_references == 0 {
    self.remove_unreferenced_application(application_number, &application.origin);
}
```

So an under-run did not just hide a mismatch — a counter that had already drifted could saturate to zero, which reads as "no anchor references this application any more", and the row would be retired while other anchors still pointed at it. Refusing is the only safe answer once the counters and the lists have diverged.

It pairs with the reordering in the thread above: the refusal now arrives before anything is written, rather than midway through. There is a test that forces the divergence (`set_counters_for_testing`, then a write that drops a reference) and asserts both the error and that the stored list is untouched.

src/internet_identity/src/account_management.rs
aterga: As a rule of thumb, we should not drop any local values.

src/internet_identity/src/storage.rs
```diff
@@ -1590,23 +1590,27 @@ impl<M: Memory + Clone> Storage<M> {
...
+    ) -> Result<Option<T>, StorageError>
     where
         F: FnOnce(&mut StorableAccountReference, Option<&mut StorableAccount>) -> T,
     {
         match maybe_account_number {
             None => {
                 // We are looking for a synthetic account
-                let (key, mut account_references) =
-                    self.find_account_references(anchor_number, application_number)?;
+                let Some(((_, application_number), mut account_references)) =
```
aterga: Why does `find_account_references` return `application_number`, while that is also its input arg?

aterga: Why do we ignore the output `_`?

src/internet_identity/src/storage.rs
```diff
@@ -1617,17 +1621,25 @@ impl<M: Memory + Clone> Storage<M> {
...
             Some(account_number) => {
                 // Account should be stored, otherwise, it was removed and we'll return `None`.
-                let mut storable_account = self.stable_account_memory.get(&account_number)?;
-                let (key, mut account_references) =
-                    self.find_account_references(anchor_number, application_number)?;
+                let Some(mut storable_account) = self.stable_account_memory.get(&account_number)
+                else {
+                    return Ok(None);
+                };
+                let Some(((_, application_number), mut account_references)) =
```
aterga: If your business logic is written against a projection of `find_account_references`'s return type, create a wrapper function that returns just that projection and document it. Don't drop returned values, as that is a common source of coding errors.

src/internet_identity/src/storage.rs
```diff
@@ -1695,43 +1709,108 @@ impl<M: Memory + Clone> Storage<M> {
...
+            });
+        }
+
+        let application = self
+            .stable_application_memory
+            .get(&application_number)
+            .ok_or(StorageError::OriginNotFoundForApplicationNumber { application_number })?;
+
+        let key = (anchor_number, application_number);
+        let previous = self
```
aterga: Use more verbose variable names even for local variables.

src/internet_identity/src/storage.rs
```diff
@@ -1695,43 +1709,108 @@ impl<M: Memory + Clone> Storage<M> {
...
+            .stable_account_reference_list_memory
+            .get(&key)
+            .map(Vec::<AccountReference>::from)
+            .unwrap_or_default();
+
+        let deltas = ReferenceListDeltas::between(&previous, &current);
```
aterga: Ditto

src/internet_identity/src/storage.rs
```diff
@@ -2477,6 +2545,49 @@ impl<M: Memory + Clone> Storage<M> {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct ReferenceListDeltas {
+    accounts: i64,
+    references: i64,
```
aterga, line 2488: It is unusual to use signed integers in canister code, why did you decide to do so in this case?

src/internet_identity/src/storage.rs
```diff
@@ -2477,6 +2545,49 @@ impl<M: Memory + Clone> Storage<M> {
...
+                .filter(|reference| reference.account_number.is_some())
+                .count() as i64;
+            (stored, references.len() as i64)
+        }
+
+        let (previous_accounts, previous_references) = counts(previous);
+        let (current_accounts, current_references) = counts(current);
+
+        Self {
+            accounts: current_accounts - previous_accounts,
```
aterga: This can overflow. 

As a rule of thumb, please always use panic-safe code. 

Either saturating or checked arithmetic, whichever fits best.

src/internet_identity/src/storage.rs
```diff
@@ -2477,6 +2545,49 @@ impl<M: Memory + Clone> Storage<M> {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct ReferenceListDeltas {
```
aterga, line 2484: This type deserved documentation.

src/internet_identity/src/storage.rs
```diff
@@ -2477,6 +2545,49 @@ impl<M: Memory + Clone> Storage<M> {
...
+            accounts: current_accounts - previous_accounts,
+            references: current_references - previous_references,
+        }
+    }
+
+    fn is_empty(&self) -> bool {
+        self.accounts == 0 && self.references == 0
+    }
+
+    /// Refuses rather than clamping: an under-run means the counters and the stored
```
aterga, line 2533: Poorly worded doc comment

src/internet_identity/src/storage.rs
```diff
@@ -2477,6 +2545,49 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// lists have already diverged, and a clamped zero reads as "no anchor references
+    /// this application any more", which retires a row other anchors still point at.
+    fn apply(&self, accounts: u64, references: u64) -> Result<(u64, u64), StorageError> {
```
aterga: Consider a more verbose naming scheme, e.g.

```suggestion
    fn apply(&self, num_accounts: u64, num_references: u64) -> Result<(u64, u64), StorageError> {
```

aterga (dismissed):
TODO: review src/internet_identity/src/storage.rs more carefully once the current comments have been addressed.

src/internet_identity/src/storage.rs
sea-snake: They're deltas, not absolute numbers. Using signed integers is necessary to have negative numbers.

src/internet_identity/src/storage.rs
```diff
@@ -2508,6 +2619,12 @@ pub enum StorageError {
         application_number: ApplicationNumber,
     },
     ErrorUpdatingAccountCounter,
+    AccountsCounterOverflow,
+    AccountCounterOutOfBounds,
```
MRmarioruci: Could this error say which counter failed and include its current value and delta? Right now we cannot tell which stored count is inconsistent.

src/internet_identity/src/storage.rs
```diff
@@ -1638,13 +1650,15 @@ impl<M: Memory + Clone> Storage<M> {
                     }
                 }
 
-                let value = StorableAccountReferenceList::from_vec(account_references);
-
-                self.stable_account_reference_list_memory.insert(key, value);
+                self.write_reference_list(
```
MRmarioruci: this always writes both records, although callers only change one. set_account_last_used changes the reference, while update_existing_account changes the account. Could we use separate helpers and avoid rewriting unchanged stable-memory entries?

sea-snake: Currently refactoring these parts, optimizing for consistency and a single fn path over avoiding edge cases where data is unnecessarily rewritten. The cost of incorrectly written entries is significantly higher than unnecessarily rewritten unchanged entries.

src/internet_identity/src/account_management.rs
sea-snake: Done — and it moved as well as propagating, because propagating it where it stood would have been worse than the discard.

```rust
// Stamped before the delegation is signed. On the IC returning `Err` commits
// every write that came before it, so propagating a failure from here once the
// signature was in the map would report an error for a delegation that has
// already been issued. `Ok(None)` is not a failure: it means the row holds no
// reference to stamp, which is how a default account that is still derived rather
// than stored reads.
storage_borrow_mut(|storage| {
    storage.set_account_last_used(anchor_number, origin.clone(), account_number, time())
})
.map_err(|err| AccountDelegationError::InternalCanisterError(err.to_string()))?;
```

It sat *after* `add_delegation_signature` and `update_root_hash`. Returning `Err` on the IC commits everything written before it — only a trap rolls a message back — so a `?` in that position would have told the caller the delegation failed while its signature was already in the signature map and the root hash already updated. Moving it above them is what makes the propagation correct rather than harmful.

The `Option` is deliberately not an error: `Ok(None)` means the row holds no reference to stamp, which is how a default account that is still derived rather than stored reads. Only the `Result` propagates.

While checking the same rule elsewhere in this PR I found two more places it was broken, both now fixed: `create_default_account` refused with an account already allocated and stored, and `update_account` trapped on a read whose `None` is reachable. Both are in the description.

src/internet_identity/src/storage.rs
sea-snake: It shouldn't have, and the function is gone. It returned the key it was handed because its caller needed the `ApplicationNumber` back out of the `Option<ApplicationNumber>` it passed in — the `?` inside was doing the unwrapping, and the tuple was how the unwrapped value got out.

That is now the caller's job, done once, before the call:

```rust
let Some(application_number) = self.lookup_application_number_with_origin(&origin) else {
    return Ok(None);
};
```

The reader it replaced takes a plain `ApplicationNumber` and returns only the row:

```rust
/// What this identity's reference-list row at `application_number` holds.
///
/// The one place that maps storage onto [`ReferenceRow`], so no caller has to
/// remember that an absent row and an empty one mean opposite things.
fn reference_row(
    &self,
    anchor_number: AnchorNumber,
    application_number: ApplicationNumber,
) -> ReferenceRow
```

Your three comments here are really one, so the answer to all of them is in the reply to [the wrapper-function comment](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932876652).

sea-snake: Because `find_account_references` returned a key its caller already had, and the `_` was dropping the half of it that was redundant. The signature was the problem, not the call site — see the replies to [the return-type question](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932863214) and [the wrapper-function comment](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932876652). Both the function and the `_` are gone.

src/internet_identity/src/storage.rs
sea-snake: Taken, and it turned out to be the most useful comment on this PR — the projection wasn't the real problem, it was the symptom, and following it out led to a much bigger cleanup. The PR is retitled and its description rewritten; the short version:

A `(anchor, application)` row has **three** states, and they mean three different things:

- **absent** — nothing ever happened at this application, so a default account is still reconstructible;
- **empty** — a tombstone: every reference moved away, and that seed must never be derived again;
- **holding references** — the ordinary case.

All three arrived at callers as `Option<Vec<_>>`, which spells the first two — *opposites* — as one nullable vector. That is what let three overlapping readers each re-derive the distinction by hand, and it is why one of them wanted to return its own input back out.

So rather than a wrapper returning the projection, the row itself is now a type:

```rust
/// What one `(anchor, application)` reference-list row holds.
///
/// Deliberately not `Option<Vec<_>>`: absence and emptiness are opposites here, and
/// spelling them as one nullable vector is what lets a caller treat them alike.
/// ...
/// Matching is exhaustive, so a fourth state cannot be added without every reader
/// being made to say what it does about it.
enum ReferenceRow {
    /// No row.
    Untouched,
    /// A row holding nothing. A permanent tombstone.
    Tombstone,
    /// A row holding references.
    Held(HeldReferences),
}

/// The references stored against one row, never empty.
struct HeldReferences(Vec<AccountReference>);
```

`reference_row` is the single documented reader, replacing `lookup_account_references`, `find_account_references` and `find_account_reference`. Every caller matches exhaustively, so the tombstone arm is *stated* — `read_account` says out loud what it answers for a tombstone, `list_accounts` says it lists nothing, `create_additional_account` says a tombstone must not regain a default.

`HeldReferences` earns its keep on the write side: `write_reference_list` takes it, so the empty list it used to refuse at runtime cannot be handed to it at all, and `StorageError::EmptyAccountReferenceList` is gone with it.

`with_account_mut` split along the same line — `with_reference_mut` for the reference alone, `with_named_account_mut` for a named account and its record — which dropped the two unreachable `Option`s `update_existing_account` was unwrapping, and the `Some(Some(x))` pattern with it. `reference_row_state_tests` (7) pins all three states across the four public entry points.

src/internet_identity/src/storage.rs
sea-snake: Done. `write_reference_list`'s locals now say which list and which counters they mean:

```rust
fn write_reference_list(
    &mut self,
    anchor_number: AnchorNumber,
    application_number: ApplicationNumber,
    new_references: HeldReferences,
) -> Result<(), StorageError> {
    ...
    let previous_row = self.reference_row(anchor_number, application_number);
    let counter_deltas = ReferenceListDeltas::between(&previous_row, &new_references);
```

`current` → `new_references`, `previous` → `previous_row`, `deltas` → `counter_deltas`, and `ReferenceListDeltas::between`'s parameter took the same rename so the two agree. `app_num` in `create_additional_account` is `application_number` now too, matching every other call site.

src/internet_identity/src/storage.rs
sea-snake: Same rename — see [the reply above](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932909564). `previous` is `previous_row` and it is a `ReferenceRow` rather than a vector now, which is the more useful half of the change: the name and the type both say that an absent row and an empty one are not the same thing.

src/internet_identity/src/storage.rs
sea-snake: That answer was too thin for a fair question, so it is now written down where it belongs — the type carries it (also [the documentation comment](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932923714)):

```rust
/// How one write to a reference-list row moves the counters derived from it.
///
/// Signed because these are differences rather than totals: a write that drops a
/// reference has to move the counters down, and there is no unsigned way to say so.
/// Both are applied to `u64` totals by [`ReferenceListDeltas::apply`].
struct ReferenceListDeltas {
    /// Change in named accounts — references that carry an account number.
    accounts: i64,
    /// Change in references, named and tracked-default alike.
    references: i64,
}
```

The stored counters stay `u64`. `i64` never reaches storage: it exists only between the diff and `apply`, which folds it back into a `u64` with `checked_add_signed` ([the panic-safety comment](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932921728)). Retiring a row goes through `ReferenceListDeltas::removing` rather than a diff against an empty list, because there is no empty `HeldReferences` to diff against — a row holding nothing is a tombstone and stays.

src/internet_identity/src/storage.rs
sea-snake: Fixed, using both kinds where each fits.

**Checked where an overflow means the data is already wrong.** `apply` refuses rather than clamping, because the clamped value is not only recorded, it is *acted on*:

```rust
/// Refuses rather than clamping: an under-run means the counters and the stored
/// lists have already diverged, and a clamped zero reads as "no anchor references
/// this application any more", which retires a row other anchors still point at.
fn apply(&self, num_accounts: u64, num_references: u64) -> Result<(u64, u64), StorageError> {
    let num_accounts = num_accounts
        .checked_add_signed(self.accounts)
        .ok_or(StorageError::AccountCounterOutOfBounds)?;
    ...
```

Paired with the ordering fix, the refusal arrives before anything is written rather than midway through. There is a test that forces the divergence and asserts both the error and that the stored list is untouched.

**Saturating where the overflow cannot happen and a cast would hide it.** Counting a list into `i64`:

```rust
/// Saturating rather than `as`: a list long enough to overflow `i64` cannot
/// exist — `MAX_ANCHOR_ACCOUNTS` bounds it far below — and saturating says so
/// without a cast that would wrap silently if that ever stopped being true.
```

So no `as`, no bare `+`, and no arithmetic left that can panic on this path.

src/internet_identity/src/storage.rs
sea-snake: Agreed — documented, and both fields with it, since the field docs are where the two counters stop being interchangeable:

```rust
/// How one write to a reference-list row moves the counters derived from it.
///
/// Signed because these are differences rather than totals: a write that drops a
/// reference has to move the counters down, and there is no unsigned way to say so.
/// Both are applied to `u64` totals by [`ReferenceListDeltas::apply`].
struct ReferenceListDeltas {
    /// Change in named accounts — references that carry an account number.
    accounts: i64,
    /// Change in references, named and tracked-default alike.
    references: i64,
}
```

`between` and `removing` are documented too, and `between`'s doc carries the rule that is easy to get wrong: a tombstone counts as no references, which is right for the totals and is exactly why retiring a row must not go through it — a tombstone's row is still alive while holding nothing, so decrementing to zero on it would let the row be retired and a moved-away default be reconstructed at the same principal. `removing` exists for that case.

src/internet_identity/src/storage.rs
sea-snake: Rewritten. It said "updates the counters" and then listed what it didn't do, which is not what a reader needs from it. It now says what the function is for and states the rule that makes the ordering matter:

```rust
/// The single write path for an anchor's account reference list at one
/// application, including the counters derived from it.
///
/// Takes [`HeldReferences`], so it cannot be handed an empty list: that would be a
/// tombstone, which only a future move may create and which no path in this design
/// should reach by writing references it happens to have none of.
```

and inside it, why the two steps are in this order rather than the other:

```rust
// Counters before the list: returning an error after the list is written would
// commit one without the other. Past this line nothing can fail.
```

src/internet_identity/src/storage.rs
sea-snake: Taken verbatim:

```rust
fn apply(&self, num_accounts: u64, num_references: u64) -> Result<(u64, u64), StorageError> {
```

It returns `Result` now as well — see [the panic-safety comment](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932921728), and [the documentation comment](https://github.com/dfinity/internet-identity/pull/4232#discussion_r3932923714) for the type's docs.

src/internet_identity/src/account_management.rs
```diff
@@ -351,6 +356,17 @@ pub async fn prepare_account_delegation(
...
     let seed = account.calculate_seed();
 
+    // Stamped before the delegation is signed. On the IC returning `Err` commits
+    // every write that came before it, so propagating a failure from here once the
+    // signature was in the map would report an error for a delegation that has
+    // already been issued. `Ok(None)` is not a failure: it means the row holds no
+    // reference to stamp, which is how a default account that is still derived rather
+    // than stored reads.
+    storage_borrow_mut(|storage| {
+        storage.set_account_last_used(anchor_number, origin.clone(), account_number, time())
```
aterga: In an async function, `time()` should be explicitly justified — is this the time of the function call, or the time of this async message's execution? Often when all timestamps are intended to be the same within an async function, it's a good idea to lift the timestamp into a function argument and make it immutable.

src/internet_identity/src/storage.rs
```diff
@@ -1527,126 +1526,107 @@ impl<M: Memory + Clone> Storage<M> {
...
+    ///
+    /// The one place that maps storage onto [`ReferenceRow`], so no caller has to
+    /// remember that an absent row and an empty one mean opposite things.
+    fn reference_row(
         &self,
         anchor_number: AnchorNumber,
         application_number: ApplicationNumber,
-    ) -> Option<Vec<StorableAccountReference>> {
-        self.stable_account_reference_list_memory
+    ) -> ReferenceRow {
```
aterga: This PR seems to introduce a lot of new concepts, like reference rows. This isn't self-describing, so it should be avoided unless this concept is a design-level building block.

src/internet_identity/src/storage.rs
sea-snake: Done — it names the counter, the count, the value stored and the delta that would not fit:

```
the anchor 10000 stored accounts counter cannot move from 0 by -1, so it no longer agrees with the stored reference lists
```

```rust
AccountCounterOutOfBounds {
    counter: ReferenceCounter,
    count: ReferenceCount,
    stored: u64,
    delta: i64,
},
```

`ReferenceCounter` is `Anchor { anchor_number }` / `Global` / `Application { application_number }`, so it identifies the row too, and `ReferenceCount` is `Accounts` / `References`. `apply` now takes the counter it is moving and `apply_one` carries the single refusal path for both counts.

Your reasoning is why the fields matter rather than just being nicer: the write is refused *before* anything is stored, so the counters keep the values that disagreed with the lists and there is nothing left to read the cause off afterwards. The error is the only place that information exists.

Chasing this turned up a related bug at the call sites. The global counter was having *both* counts moved, but its account count is the account-number allocator — `allocate_account_number` owns it, and this path rewrites it unchanged:

```rust
let (_, global_references) = deltas.apply(
    global_counter.stored_accounts,
    global_counter.stored_account_references,
)?;
```

So it could refuse a write over a count nothing was going to store, and the discarded half of the pair was hiding it. That site applies only its reference count now, and the `let (_, ...)` is gone.

Two tests: the existing underflow test asserts the whole rendered message, and a new one drops a tracked default so the two deltas differ (0 and -1) and pins that the refusal names the count that actually failed. I checked both catch a regression — mislabelling the counter fails the first, and swapping which delta each count reads fails the second (it passed the first, since both deltas are -1 there).

src/internet_identity/src/storage.rs
sea-snake: That refactor has landed, and it took the separate helpers you asked for — because the types needed them, not just the writes:

```rust
/// Applies `f` to one of this identity's account references, and writes the row
/// back if it ran.
fn with_reference_mut<T, F>(...)
where F: FnOnce(&mut AccountReference) -> T

/// Applies `f` to one of this identity's named accounts and the reference to it,
/// and writes both back if it ran.
fn with_named_account_mut<T, F>(...)
where F: FnOnce(&mut AccountReference, &mut StorableAccount) -> T
```

`set_account_last_used` takes the first and so no longer touches the account record at all — that was the real waste you spotted: called with `Some(account_number)`, the old single path rewrote the `StorableAccount` on **every sign-in**, for a stamp that only ever changes the reference.

`update_existing_account` takes the second and still rewrites its row alongside the record. That one stays, for the reason given above: `f` gets both handles and the function cannot know which it touched, so telling them apart would mean a third helper carrying a read-only reference, and one unchanged row rewrite is the cheaper side of that trade.

Splitting them paid off beyond the writes, which is why it was worth doing here rather than deferring: `update_existing_account` was unwrapping `Some(Some(x))` and had two `?` on `Option`s that the single path made unprovable but that are unreachable once a named account is its own entry point. Both are gone. The record is also written only once the reference has granted access, so a caller with no reference to the account cannot rewrite the record of whichever identity does own it.

MRmarioruci (approved):

src/internet_identity/src/account_management.rs
sea-snake: Done — one read, bound immutably, passed to all four uses:

```rust
// One read, passed to everything below. `time()` is constant only within a single
// execution, and an `await` on an inter-canister call ends one — the continuation
// resumes with a later time. Reading it at each use would make "the same instant"
// rest on no `await` ever appearing between them, which is not a property to leave
// to where the calls happen to sit.
let now = time();
```

To answer the question directly: it is the time of *this message's execution*. There is one `.await` in the function and all four reads sat below it, so they were in the same execution and did return the same value — but only because no `await` happens to appear between them, which is not something the code said. That is the reason to bind it rather than an argument that the four agree. The comment that used to make that argument (*"`time()` is stable here (no await since it was read for `expiration`)"*) is deleted; the binding carries it now.

src/internet_identity/src/storage.rs
sea-snake: Taken — `ReferenceRow` and `HeldReferences` are gone, along with the vocabulary. The reader is back to the shape it describes:

```rust
/// This identity's account references at `application_number`, or `None` where it
/// has no row there at all.
fn account_references(
    &self,
    anchor_number: AnchorNumber,
    application_number: ApplicationNumber,
) -> Option<Vec<AccountReference>>
```

It is still the one reader, replacing `lookup_account_references`, `find_account_references` and `find_account_reference`, so the other three threads stay addressed — there is just no new concept to learn for it.

Worth saying why dropping it was better than defending it, because the concept *does* meet the test you set. #4222 has a table headed **Row state** listing all four cases, a state diagram with a `Tombstone` node, and a section anchored *tombstones must not decrement*. So it is design-level, and I would have replied with that.

What changed my mind is that the arm those types existed for has a one-PR lifespan. The only reader that needs to tell an *empty* row from a *non-empty row with no default* is `read_account`, because this PR still answers a synthetic default for the empty one (the `XXX WARNING`). #4234's entire job is flipping that, and after it both cases answer the same thing — so the special case is deleted rather than inverted, and the only state any reader still singles out is *absent*, which `Option` already carries. A type whose distinguishing arm the next PR removes is not worth the vocabulary.

The one rule that did need enforcing moved to where it cannot be bypassed rather than into a type callers pass around:

```rust
impl TryFrom<Vec<AccountReference>> for StorableAccountReferenceList
```

The infallible `From<Vec<_>>` is deleted, so the unchecked path no longer compiles, and `Default` went with it — it handed out the empty list the constructor exists to refuse. Non-emptiness is the only rule there, deliberately: decoding stays infallible, so a rule added here also applies to every future write of an *existing* row, and a stored row breaking one would become unwritable — an identity locked out of that origin. Emptiness is the only rule I can show was never written.

Two more things fell out of removing the abstraction, both in the PR description:

`remove_reference_list` accepted any non-empty row and trusted its caller to have filtered. It now refuses unless the row is exactly one live tracked default, which is the only thing prunable — the check belongs in the function doing the irreversible step. That caught an existing test that was pruning a row with a named account in it.

And the helper pair I had built (`with_reference_mut` / `with_named_account_mut`) collapsed to one plus a plain read, which surfaced a real defect: the rename path's closure never mutated the reference, but the helper always wrote the row back, so every account rename rewrote the whole `(anchor, application)` blob with byte-identical content. It writes only the account record now.

src/internet_identity/src/account_management.rs
```diff
@@ -322,6 +327,13 @@ pub async fn prepare_account_delegation(
             .ok_or(AccountDelegationError::Unauthorized(caller()))
     })?;
 
+    // One read, passed to everything below. `time()` is constant only within a single
+    // execution, and an `await` on an inter-canister call ends one — the continuation
+    // resumes with a later time. Reading it at each use would make "the same instant"
+    // rest on no `await` ever appearing between them, which is not a property to leave
+    // to where the calls happen to sit.
+    let now = time();
```
aterga, line 325: This is better. Why not lift this to the function arg? Because the caller also has multiple execution times.

MRmarioruci (approved):

src/internet_identity/src/account_management.rs
sea-snake: Both done, and the second one is what made the first one possible.

`prepare_account_delegation` takes `now: Timestamp` and the three call sites in `main.rs` pass `time()`, as does `prepare_session_delegation`.

That was blocked before by the salt: the function awaited `ensure_salt_set()`, so a `now` read before the await was not the same `now` as after it, and passing one in would have been passing in a lie. The await is gone — `sessions.rs`, `account_management.rs`, `session_delegation.rs` and `delegation.rs` have zero `.await` and zero `async fn` between them now — and `ensure_salt_set` is deleted rather than merely unused.

A fresh canister gets its salt from a zero-delay timer installed in `#[init]`, since deriving one needs raw randomness and no lifecycle hook can await. `init_salt` survives as an endpoint but no longer traps on "Salt already set": every canister has a salt from its install onwards, so a caller asking for one is asking a question that is already answered.

One thing still outstanding, which we have agreed and I have not done: `sessions.rs` takes its own `time()` in four places — `create_account_session`, `app_prepare_delegation`, `app_get_delegation`, `check_session`. Each is the first line of a body called straight from an endpoint, so `now` is still one value per message, but the rule should be uniform and `main.rs` should supply it there too.

src/internet_identity/src/storage/storable/account_reference_list.rs
```diff
@@ -25,13 +28,64 @@ impl Storable for StorableAccountReferenceList {
...
+    /// away and its default account must never be derived again. Only a move may
+    /// create one, and nothing moves accounts yet, so an empty list here is a bug
+    /// rather than an intent — storing it would deny an identity its default account
+    /// for good.
+    Empty,
+    /// One account cannot be held twice at one origin. Both references would derive the
+    /// same principal, so the index would hold one entry for two slots while the stored
+    /// account count claimed both — and nothing downstream could tell which reference a
+    /// read had answered from.
+    RepeatedAccountNumber(AccountNumber),
```
nmattia, line 51: And we store a list instead of a set because we may want to prune the oldest accounts, is that it?

src/internet_identity/src/storage/storable/account_reference_list.rs
```diff
@@ -46,14 +100,96 @@ impl From<StorableAccountReferenceList> for Vec<AccountReference> {
     }
 }
 
-impl From<Vec<AccountReference>> for StorableAccountReferenceList {
-    fn from(value: Vec<AccountReference>) -> Self {
-        StorableAccountReferenceList(
+/// The only way to build a list to be stored, so every write is checked. Deliberately
+/// `TryFrom` rather than `From`: an infallible conversion existed here before and the
+/// checks it lacked had to be remembered at each of the write sites instead.
```
nmattia, line 105: nit: sounds like an LLM's justification during development, not necessarily something worth keeping as a comment

src/internet_identity/src/storage/storable/account_reference_list.rs
```diff
@@ -46,14 +100,96 @@ impl From<StorableAccountReferenceList> for Vec<AccountReference> {
...
+        if value.is_empty() {
+            return Err(StorableAccountReferenceListError::Empty);
+        }
+
+        // A reference is identified by the account it names, and the numberless one names
+        // the tracked default — so one pass over `Option<AccountNumber>` catches a repeated
+        // number and a second tracked default alike.
+        let mut seen = BTreeSet::new();
+        for reference in &value {
+            if !seen.insert(reference.account_number) {
```
nmattia, line 119: Might be a little late in the design process, but I'm wondering if it wouldn't make sense to invest into a one-time migration such that both "default" and named accounts are treated the same. This would probably save time and reduce complexity (and potential for critical bugs) over the long term.

src/internet_identity/src/storage/storable/account_reference_list.rs
```diff
@@ -46,14 +100,96 @@ impl From<StorableAccountReferenceList> for Vec<AccountReference> {
...
+                return Err(match reference.account_number {
+                    Some(account_number) => {
+                        StorableAccountReferenceListError::RepeatedAccountNumber(account_number)
+                    }
+                    None => StorableAccountReferenceListError::RepeatedTrackedDefault,
+                });
+            }
+        }
+
+        Ok(StorableAccountReferenceList(
```
nmattia, line 129: nit (not introduced by this PR): should use `into_iter`

src/internet_identity/src/storage/storable/discrepancy_counter.rs
```diff
@@ -24,19 +24,4 @@ impl Storable for StorableDiscrepancyCounter {
...
-                    .expect("overflow in account_counter_rebuilds"),
-            },
-        }
-    }
-}
-
-pub enum DiscrepancyType {
-    AccountRebuild,
-}
+impl StorableDiscrepancyCounter {}
```
nmattia, line 27:
```suggestion
```

src/internet_identity/src/main.rs
```diff
@@ -2312,8 +2322,6 @@ mod attribute_sharing {
             get_account_for_origin(anchor.anchor_number(), origin.clone(), account_number)
                 .map_err(PrepareIcrc3AttributeError::GetAccountError)?;
 
-        state::ensure_salt_set().await;
```
nmattia, line 2315: nit: I think it'd make sense to do this separately (the whole `async` simplification)

src/internet_identity/src/storage.rs
```diff
@@ -1527,126 +1527,76 @@ impl<M: Memory + Clone> Storage<M> {
             .and_then(|application_number| self.stable_application_memory.get(&application_number))
     }
 
-    fn lookup_account_references(
+    /// This identity's account references at `application_number`, or `None` where it
+    /// has no account reference list there at all.
+    ///
+    /// The one reader, so no caller has to assemble the list from storage itself. The
+    /// `None` is only ever "no list": an empty list is a tombstone and means the
```
nmattia, line 1534: just to make sure I understand: as of right now, `None` is unexpected, correct?

src/internet_identity/src/storage.rs
```diff
@@ -1527,126 +1527,76 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// opposite, so the two must not be collapsed by a caller either.
+    fn account_references(
         &self,
         anchor_number: AnchorNumber,
         application_number: ApplicationNumber,
-    ) -> Option<Vec<StorableAccountReference>> {
+    ) -> Option<Vec<AccountReference>> {
```
nmattia, line 1540: For my own understanding: why does this not return a `StorableAccountReferenceList`?

src/internet_identity/src/storage.rs
```diff
@@ -1527,126 +1527,76 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// `Ok(None)` means there was nothing to apply `f` to and nothing was written:
+    /// the list is absent or a tombstone, or it holds no reference for this account —
+    /// see [`Self::account_reference`] for what that last case means.
+    fn with_account_reference_mut<T, F>(
         &mut self,
         anchor_number: AnchorNumber,
-        application_number: Option<ApplicationNumber>,
-        maybe_account_number: Option<AccountNumber>,
+        application_number: ApplicationNumber,
+        account_number: Option<AccountNumber>,
```
nmattia, line 1576: nit: would avoid the type blindness and explicitly introduce `DefaultAccount | Account<AccountNumber>`

src/internet_identity/src/storage.rs
```diff
@@ -1527,126 +1527,76 @@ impl<M: Memory + Clone> Storage<M> {
...
         f: F,
-    ) -> Option<T>
+    ) -> Result<Option<T>, StorageError>
```
nmattia, line 1578: For my own understanding: Is there any case where `Ok(None)` is not expected to be an error?

src/internet_identity/src/storage/storable/account_reference_list.rs
sea-snake: A set is unordered, ordering is something we want to store. Maybe a `BTreeSet` will work here indeed.

src/internet_identity/src/storage/storable/account_reference_list.rs
sea-snake: We did discuss this, but both accounts have their own caps and behavior, which would be lost if we had only one account type.

src/internet_identity/src/storage.rs
sea-snake: Empty vec is unexpected, a non-empty vec and none are both expected.

src/internet_identity/src/storage.rs
sea-snake: Storables structs are internal only, we have separate types for internal (storage) and external (logic) data, so changes in one won't affect the other vice versa.

src/internet_identity/src/account_management.rs
```diff
@@ -423,32 +420,23 @@ pub fn get_account_delegation(
...
 
-/// Checks whether the stored number of accounts as per the counter exceeds the maximum permitted number.
-/// If it does, it rebuilds the counter. If it still exceeds, it will return an error.
-fn check_or_rebuild_max_anchor_accounts(
-    storage: &mut Storage<DefaultMemoryImpl>,
+/// Refuses where this identity already holds as many accounts as it may.
+///
+/// Counted rather than repaired. There used to be a rebuild here for a counter that could
+/// drift, because it was maintained by hand at each write site; the write path derives it
+/// now, so there is nothing to repair and no reason to look twice.
```
nmattia, line 427: Not relevant going forward (sounds like LLM justification during development)

src/internet_identity/src/account_management.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
     update_root_hash,
 };
 use ic_canister_sig_creation::{signature_map::CanisterSigInputs, DELEGATION_SIG_DOMAIN};
-use ic_cdk::{api::time, caller};
```
nmattia, line 21: out of curiosity, why do we stop calling `time()`?

src/internet_identity/src/account_management.rs
```diff
@@ -796,7 +784,7 @@ fn should_update_default_account_for_origin() {
 }
 
 #[test]
-// This test is to make sure that the check_or_rebuild_max_anchor_accounts function correctly errors
+// This test is to make sure that the check_max_anchor_accounts function correctly errors
 // It should error when the counters are at or above max and argument 'first_time' is false
 fn should_fail_check_or_rebuild_when_not_first_time() {
```
nmattia, line 789: I think the name here is off and was referring to the `first_time` arg used for recursion

src/internet_identity/src/account_management.rs
```diff
@@ -813,57 +801,27 @@ fn should_fail_check_or_rebuild_when_not_first_time() {
             MAX_ANCHOR_ACCOUNTS as u64,
             MAX_ANCHOR_ACCOUNTS as u64,
         );
-        let res = check_or_rebuild_max_anchor_accounts(
-            storage,
-            anchor.anchor_number(),
-            MAX_ANCHOR_ACCOUNTS as u64,
-            false,
-        );
+        let res =
```
nmattia, line 804: nit: would also test for expected behavior for max - 1 and max + 1

src/internet_identity/src/state.rs
```diff
@@ -245,31 +245,29 @@ struct State {
...
+///
+/// A canister that already has a salt is the answer this asks for, not a fault, so it says
+/// so by returning rather than by trapping a caller who wanted exactly that.
+pub async fn set_salt_if_unset() {
+    if storage_borrow(|storage| storage.salt().is_some()) {
+        return;
     }
 
-    storage_borrow(|storage| {
+    let salt = random_salt().await;
```
nmattia, line 260: something to maybe keep in mind: previously `init_salt` was synchronous (for some definition of synchronous) when called after the salt was set; this is not the case anymore and since the endpoint is public it might be expensive if someone spams it. I haven't looked into canister performance in a while so maybe this is a non-issue.

src/internet_identity/src/main.rs
```diff
@@ -90,7 +93,7 @@ const ID_AI_ORIGIN: &str = "https://id.ai";
 
 #[update]
 async fn init_salt() {
```
nmattia, line 95: is this endpoint still necessary?

src/internet_identity/src/storage/storable/account_reference_list.rs
sea-snake: A `Vec`, and staying one. Uniqueness is already enforced where it matters — `TryFrom` is the only way to build a list that gets stored, and it rejects a repeated account number and a second tracked default — so a set buys no safety. A `BTreeSet` would also order references by account number, which throws away the insertion order that makes "oldest" mean anything. My "maybe a `BTreeSet` will work here" was the wrong instinct.

src/internet_identity/src/storage/storable/account_reference_list.rs
sea-snake: Agreed, and trimmed in #4318. The sentence about the infallible conversion that preceded it is a fact about history, and it is already recorded in the commit that made the change; what stays is the sentence that says what the code does — this is the only way to build a list to be stored, so every write is checked.

src/internet_identity/src/storage/storable/account_reference_list.rs
sea-snake: Keeping the two, but the answer I gave you at the time was thinner than it should have been: the caps are enforced by the write gate from the lists, not by the type, so "both have their own caps" was not the reason.

The reason that holds is the migration itself. Making them one type means minting a number for every derived default that exists, which turns an absent list — which today *means* "derive the default" — into stored rows for every origin any identity has ever visited. That is exactly the growth the tracked-default eviction exists to bound, so the seam is buying something rather than just being legacy.

src/internet_identity/src/storage/storable/account_reference_list.rs
sea-snake: Applied in #4318, at the `collect` rather than the loop you flagged: `value.iter().cloned()` was cloning every reference out of a `Vec` the function owns and drops. The dedup pass above it keeps borrowing, since it runs before the conversion and the vector is still needed afterwards.

src/internet_identity/src/storage/storable/discrepancy_counter.rs
sea-snake: Deleted, and the rest of it with it, in #4318. Nothing had incremented that counter since the rebuild it counted was removed in this PR, so `internet_identity_account_counter_discrepancy_count` was reporting whatever it held at that upgrade and could never move again — a gauge frozen at zero reads as "no discrepancies" rather than "nobody is looking". The accessor, the stable cell and the storable go too; memory index 18 is retired the way 7, 8, 25, 27, 28 and 30 already are.

src/internet_identity/src/main.rs
sea-snake: Fair, and conceded — the salt lifecycle moved across `main.rs` and `state.rs` inside a PR about account reference lists, which is why two of your other comments here are also about it. Too late to split this one, but the point stands for the stack still open.

src/internet_identity/src/storage.rs
sea-snake: The empty vec is the unexpected one; `None` and a non-empty vec are both expected and mean opposite things — `None` is "no list at this origin", an empty list is a tombstone saying every reference was moved away and the default must never be derived again. That is why the two must not be collapsed by a caller either.

src/internet_identity/src/storage.rs
sea-snake: Because the storable types stay inside `storage`. The logic layer has its own, and the conversion happens at the boundary, so a change to either the stored encoding or the domain type does not reach the other. It is the invariant the whole refactor rests on.

src/internet_identity/src/storage.rs
sea-snake: Not doing this, for the same reason as the default/named question above. `Option<AccountNumber>` is the key shape in the account-key storable and the Candid shape, so an enum would convert at both boundaries and buy type safety only in between. The double duty is also load-bearing in the good way: the dedup pass catches a repeated number and a second tracked default in one pass over `Option<AccountNumber>`, which two variants would split into two checks. (The function you commented on is gone either way — #4234 replaced it with `write_account`.)

src/internet_identity/src/storage.rs
sea-snake: Your instinct was right: if every `Ok(None)` was really an error, the `Option` did not belong there. `write_account`, which replaced that path in #4234, returns `Result<Account, StorageError>` with no `Option` at all — the "nothing to apply it to" case is a `StorageError` now rather than a silent `None`.

src/internet_identity/src/account_management.rs
sea-snake: Dropped in #4318. The paragraph explained an accepted cost that the test's own name already states — `a_drifted_account_counter_is_not_repaired_and_costs_the_identity_its_limit` — so all it added was the archaeology.

src/internet_identity/src/account_management.rs
sea-snake: Because `now` is a parameter now: it is read once per request at the endpoint and threaded down. Two things follow — every stamp within one request agrees, and the unit tests drive time without a canister clock, which is what lets the account-limit tests run in-process at all.

src/internet_identity/src/account_management.rs
sea-snake: Fixed in this same stack — it is `a_drifted_account_counter_is_not_repaired_and_costs_the_identity_its_limit` now, and `first_time` is gone with the recursion it belonged to.

src/internet_identity/src/account_management.rs
sea-snake: Both already covered, by construction: `should_fail_to_create_accounts_above_max` loops `0..=MAX_ANCHOR_ACCOUNTS` asserting `Ok` on every iteration below the cap and `AccountLimitReached` at it, which is max − 1 and max. Max + 1 is unreachable through the API — the write refuses at the cap — and its stand-in, a counter forced above the cap, is `a_drifted_account_counter_is_not_repaired_and_costs_the_identity_its_limit` via `set_counters_for_testing`.

src/internet_identity/src/state.rs
sea-snake: Checked, and it is a non-issue: `set_salt_if_unset` returns on `storage.salt().is_some()` *before* `random_salt().await`, so a call to a canister that already has a salt costs one ordinary update message and never reaches the randomness call. A future that completes without awaiting finishes in the same execution, so the `async` costs nothing by itself — and the endpoint was an unauthenticated `#[update]` before this PR too, so the spam surface is unchanged either way.

src/internet_identity/src/main.rs
sea-snake: Yes, and it is keeping its trap message too. The zero-delay timer is installed from `init` only, deliberately: an upgrade keeps the salt because stable memory is what an upgrade keeps, and the one canister that wants one afterwards — upgraded from a release that derived its salt on the first delegation and never issued one — has to be asked rather than given one. The integration tests also need it set deterministically rather than whenever a timer fires.

src/internet_identity/src/account_management.rs
nmattia: >  it is read once per request at the endpoint and threaded down. 

Isn't it equivalent to reading `time()`? I believe the ic system interface guarantees that the value will be the same (esp. since everything is synchronous now)

PR #4233, opened 2026-08-19
refactor(be): allocate application numbers from a monotonic cell

sea-snake:

`ApplicationNumber` is derived from `lookup_application_with_origin_memory.len()`, which is only safe while nothing is ever removed. Once an application can be removed, `len()` collides with a live number rather than merely reusing a retired one:

```
applications {0, 1, 2} exist        len() == 3
remove 1                            len() == 2
next new origin is assigned 2       already owned by a live application
```

Both origins would then share one account universe: every `(anchor, 2)` reference list and config list would belong to both.

Numbers now come from a monotonic `StableCell<u64>` at memory index 33. **The counter is what guarantees a number is never reissued**, because it only climbs — a number it has passed is not offered again even after that application's list is retired.

The counter cannot be the whole answer on its own, because it postdates the applications numbered before it existed. So allocation takes `max(counter, highest_stored + 1)`, and the map being keyed by `ApplicationNumber` means `last_key_value` gives that highest number exactly, holes in the lists or not. Nor can the floor stand alone: removing the highest list walks it backwards, which is what the counter is there to prevent.

Both steps are checked and refuse at the ceiling rather than saturating, for the reason `allocate_account_number` refuses — the number keys both the application list and the origin index, so reissuing one puts two origins on a single list. `allocate_application_number` and `lookup_or_insert_application_number_with_origin` return `Result` for that; all three production callers propagate.

`ApplicationNumber` never appears in `internet_identity.did`, so this is invisible to clients.

**Rollback needs a release note.** The cell at index 33 is exactly what a rollback discards, restoring `len()` as the allocator over a map that removals have punched holes in. Rolling forward recovers a safe allocator, since the cell can only be raised, but a collision created during the rollback window has already happened.

Tests: `application_number_allocator_tests` (7). The two the design exists for are `a_gap_below_the_highest_number_is_not_handed_out_again` — a hole in the lists while the counter is behind, the one state a list count gets wrong — and `a_removal_before_the_first_allocation_does_not_collide_with_a_live_number`. Both fail if the floor goes back to `len()`.


src/internet_identity/src/storage.rs
```diff
@@ -648,7 +657,21 @@ impl<M: Memory + Clone> Storage<M> {
...
             sso_stable_id_index_memory: StableBTreeMap::init(sso_stable_id_index_memory),
-        }
+        };
+        storage.seed_application_number_allocator();
+        storage
+    }
+
+    /// Existing application numbers are dense from zero, so the row count is the
+    /// first free number.
+    fn seed_application_number_allocator(&mut self) {
```
MRmarioruci: Could we make this a one-time migration instead? This is safe, but an explicit initialized state would make it clearer that `len()` is only for the old dense numbering.

MRmarioruci: on the other hand I don't mind this approach that much.. it's easy and cheap.. so up to you.

sea-snake: Keeping the approach, but you were right to doubt the invariant — it does not survive this stack, and the comment was the actual problem.

`#4232` adds `remove_unreferenced_application`, whose own doc says the number is never reissued. Once that starts firing, application numbers stop being dense: rows are deleted while the high-water mark keeps climbing, so `len()` becomes an undercount. What keeps this correct is `max(stored_counter, len())` — the counter has already passed `len()`, so the maximum picks it.

The old comment ("dense from zero, so the row count is the first free number") would therefore have been false as soon as reaping shipped, and anyone trusting it and simplifying to `len()` would reissue live numbers and collide origins. It now says the row count is a floor, exact only for data written before the counter existed, and that the maximum is what keeps allocation monotonic afterwards.

I left the `expect` — unlike the ones in the panic-free thread, this is `init` with no caller to return to.

src/internet_identity/src/storage.rs
```diff
@@ -648,7 +657,27 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// then on the count undershoots and only the stored counter is right — taking the
+    /// maximum is what keeps allocation monotonic across both.
+    fn seed_application_number_allocator(&mut self) {
+        let seeded = ApplicationNumber::max(
+            *self.next_application_number_memory.get(),
+            self.stable_application_memory.len(),
+        );
+        self.next_application_number_memory
+            .set(seeded)
+            .expect("failed to seed the application number allocator");
```
aterga: Unsafe unwrap, please get rid of it.

src/internet_identity/src/storage.rs
```diff
@@ -648,7 +657,27 @@ impl<M: Memory + Clone> Storage<M> {
                 sso_stable_id_index_memory.clone(),
             ),
             sso_stable_id_index_memory: StableBTreeMap::init(sso_stable_id_index_memory),
-        }
+        };
+        storage.seed_application_number_allocator();
+        storage
+    }
+
+    /// Seeds the allocator from whichever is higher: the stored counter, or the row
```
aterga: This doc string isn't good enough, I struggle to gain enough context to understand what this function is doing and why.

src/internet_identity/src/storage.rs
```diff
@@ -1508,6 +1537,14 @@ impl<M: Memory + Clone> Storage<M> {
         }
     }
 
+    fn allocate_application_number(&mut self) -> ApplicationNumber {
+        let new_number = *self.next_application_number_memory.get();
+        self.next_application_number_memory
+            .set(new_number + 1)
```
aterga: Overflowing arithmetic should be replaced with saturating or checked arithmetic in canister code.

aterga (dismissed):
PR description isn't self-contained. 

If apps can be removed, why not store the index of the next-free app ID rather than try to derive from the length of a collection?

src/internet_identity/src/storage.rs
```diff
@@ -647,7 +656,27 @@ impl<M: Memory + Clone> Storage<M> {
...
+    ///
+    /// The row count is a floor rather than the answer. It is exact only for data
+    /// written before this counter existed, where numbering was dense from zero.
+    /// Retiring an application removes its row without reissuing its number, so from
+    /// then on the count undershoots and only the stored counter is right — taking the
+    /// maximum is what keeps allocation monotonic across both.
+    fn seed_application_number_allocator(&mut self) {
+        let seeded = ApplicationNumber::max(
+            *self.next_application_number_memory.get(),
+            self.stable_application_memory.len(),
```
MRmarioruci: Could we seed this from the highest existing app number instead of using the row count? I understand we rely on the rows being dense but I don't love this.

sea-snake: Taken, and it turned out to be worth more than a tidier seed.

`stable_application_memory` is keyed by `ApplicationNumber`, so `last_key_value` gives the highest number outright — exact whether or not the rows have holes in them. The density assumption is gone rather than documented around:

```rust
let above_highest_stored = match self.stable_application_memory.last_key_value() {
    Some((highest, _)) => highest.checked_add(1).ok_or(StorageError::ApplicationsCounterOverflow)?,
    None => 0,
};
let new_number = ApplicationNumber::max(
    *self.next_application_number_memory.get(),
    above_highest_stored,
);
```

You were right to dislike it, and the reason is sharper than "I don't love it". The row count is wrong in exactly one state: a hole in the rows while the counter is behind. `{0,1,2}` with `0` removed and the counter at zero gives `len() == 2`, which is the number `https://c.com` still holds — two origins on one row, sharing its accounts and counters.

That state was unreachable in production, because seeding ran at load and any removal implied the counter was already ahead. But `a_removal_before_the_first_allocation_does_not_collide_with_a_live_number` was one line-order away from it: the removal happened *after* `from_memory`, so seeding had already read three rows. Swap those two lines and it collides.

The other half of it is that this removes the reason to seed eagerly at all. Eager seeding existed precisely so a removal could not shrink the count first; with an exact floor there is nothing to shrink, so the allocator takes the maximum where it allocates. `seed_application_number_allocator` is gone, and with it the constructor-time `expect` [aterga](https://github.com/dfinity/internet-identity/pull/4233#discussion_r3933194545) asked about and the doc [he](https://github.com/dfinity/internet-identity/pull/4233#discussion_r3933207901) could not follow — three of its four sentences existed to say when the row count could be trusted.

To be explicit about what still carries the guarantee, since it is not the new floor: the counter does, because it only climbs, so a number it has passed is never offered again even after the row is retired. `highest + 1` cannot do that job on its own — removing the highest row walks it backwards. It is only a floor for the applications numbered before the counter existed.

New test `a_gap_below_the_highest_number_is_not_handed_out_again` sets up the hole-plus-behind-counter state directly. Both it and the reordering-sensitive test above fail if the floor goes back to `len()`.

src/internet_identity/src/storage.rs
sea-snake: Gone — the write it guarded no longer happens.

It was not that `set` on a `StableCell<u64>` was likely to fail. It was that the call sat in `seed_application_number_allocator`, reached from `Storage::new` and `Storage::from_memory`, which return `Storage` rather than `Result` — so there was nowhere to report to, and adding an error channel to the constructor would have been a large change for an unlikely failure.

Seeding existed only because the floor was `stable_application_memory.len()`, which a removal can shrink, so it had to be read before any removal could happen. Following [MRmarioruci's suggestion](https://github.com/dfinity/internet-identity/pull/4233#discussion_r3934514456) the floor is now the highest stored number, which is exact regardless of holes, so there is nothing to read early. The allocator takes `max(counter, highest + 1)` where it allocates, and that is a function returning `Result`:

```rust
self.next_application_number_memory
    .set(new_number.checked_add(1).ok_or(StorageError::ApplicationsCounterOverflow)?)
    .map_err(|_| StorageError::ErrorUpdatingApplicationNumberAllocator)?;
```

So the cell is written in one place, and that place reports both ways it can fail. `Storage::new` writes nothing.

src/internet_identity/src/storage.rs
sea-snake: Fair, and it was a symptom rather than a wording problem — so it is gone rather than reworded.

Three of its four sentences existed to explain *when the row count could be trusted*: exact for data written before the counter existed, undershooting once a retirement removes a row, with `max` covering both. That is a doc apologising for the implementation, and no amount of rewriting would have given you the context you were missing, because the thing being explained was a workaround.

The floor is now the highest stored number ([MRmarioruci's suggestion](https://github.com/dfinity/internet-identity/pull/4233#discussion_r3934514456)), which is exact whether or not the rows have holes, so there is no condition left to describe and no seeding step to describe it on. What remains is on the allocator, and says what it hands out and why nothing can take it back:

```rust
/// Hands out an application number that no live application holds and no retired
/// one ever held.
///
/// The counter is what guarantees that: it only ever climbs, so a number it has
/// passed is never offered again even after the application's row is retired. Its
/// value is not the whole answer only because it postdates the applications
/// numbered before it existed, so the highest stored number is taken as a floor —
/// exact, unlike a row count, which a retirement leaves undershooting. It cannot be
/// the answer on its own either: removing the highest row walks it backwards.
///
/// Refuses at the ceiling rather than saturating. The number keys both the
/// application row and the origin index, so reissuing one would put two origins on
/// a single row and have them share its accounts and counters.
```

If that still leaves you reconstructing something, say which part and I will fix the doc rather than argue for it.

src/internet_identity/src/storage.rs
sea-snake: Done, checked rather than saturating — and worth saying why, since you offered both and saturating would have been unsafe here.

```rust
self.next_application_number_memory
    .set(new_number.checked_add(1).ok_or(StorageError::ApplicationsCounterOverflow)?)
    .map_err(|_| StorageError::ErrorUpdatingApplicationNumberAllocator)?;
```

At the ceiling, saturating hands out the same number forever. The application number keys both `stable_application_memory` and the origin index, so the next new origin would be inserted over a live application's row: two origins sharing one row, its `stored_accounts` and `stored_account_references`, and every `(anchor, number)` reference list and config row underneath it. That is the collision this whole PR exists to prevent, arrived at from the other end. Refusing is the only safe answer.

`allocate_account_number` made the same call for the same reason one function up, so this follows it with a sibling variant rather than inventing a policy.

The consequence is that `allocate_application_number` and `lookup_or_insert_application_number_with_origin` return `Result`. All three production callers propagate: `create_additional_account` and `create_default_account` already returned `Result<_, StorageError>`, and `set_default_account_for_origin` maps onto its existing `InternalCanisterError`. The `expect` on the line below is a `map_err` now, and the other one in this file went with the seeding step ([thread](https://github.com/dfinity/internet-identity/pull/4233#discussion_r3933194545)).

The floor the counter is combined with is also checked, for the same reason — see [MRmarioruci's thread](https://github.com/dfinity/internet-identity/pull/4233#discussion_r3934514456) for why that floor changed.

src/internet_identity/src/storage.rs
sea-snake: Following up rather than reopening, because my answer above is now wrong and I would rather not leave it as the last word on this thread.

I argued there that `max(stored_counter, len())` is what keeps allocation correct, and that the row count is a floor exact only for pre-counter data. The floor is no longer the row count. You raised the same smell again [further down](https://github.com/dfinity/internet-identity/pull/4233#discussion_r3934514456) and it was right both times: the map is keyed by `ApplicationNumber`, so `last_key_value` gives the highest number exactly, holes or not, and the density argument disappears instead of needing to be documented.

That also removed the eager seeding this thread was about, so the question you actually asked — one-time migration, or seed on load — is answered a third way: neither. There is no initialization step to make explicit, because the allocator reads both inputs where it allocates. Which is closer to what you were after with "an explicit initialized state would make it clearer", arrived at by deleting the state rather than naming it.

The `expect` I defended here on the grounds that `init` has no caller to return to is gone with it.

MRmarioruci (approved):

PR #4234, opened 2026-08-19
refactor(be): one read and one write for accounts

sea-snake:

This PR started as a one-line fix: `read_account` returned a synthetic default account for an **empty** reference list, the inverse of what that list means. Reviewing it turned up the same rule written three more times, so the fix is now the shape that makes it unrepeatable.

| List state | Meaning | Default account |
| --------- | ------- | --------------- |
| Absent | Nothing ever happened here | Reconstructible |
| Present, empty | Everything here was moved away | Must never be derived again |

## Reads normalise

`account_references` is the only place that knows absence from emptiness. An absent list becomes the derived default; a tombstone stays empty. No caller carries a missing-list policy any more, and the `is_empty()` branch this PR set out to delete has nowhere left to live.

## Writes take the whole thing

Three maps describe an identity's accounts at an origin — the reference list, the account records, the config naming the default — and each was written on its own, so they could drift. `write_account_state` writes all three together, diffing the list against the stored list for the counter deltas. Every fallible step runs before any write, so a refusal leaves nothing behind. A list the list already holds is not written, which is what lets a rename leave every reference where it stands without its caller knowing to skip.

## What is left

```rust
pub fn read_account(&self, key: &AccountKey) -> Option<Account>
pub fn list_accounts(&self, anchor: AnchorNumber, origin: &FrontendHostname) -> Vec<Account>
pub fn lookup_account_with_principal(&self, principal: Principal) -> Option<Account>
pub fn create_account(&mut self, anchor, origin, name) -> Result<Account, StorageError>
pub fn write_account(&mut self, account: Account) -> Result<Account, StorageError>
pub fn set_default_account(&mut self, anchor, origin, account_number) -> Result<(), StorageError>
```

`AccountKey` replaces `ReadAccountParams` and carries no capability — an account's signing seed only ever leaves `read_account`, which is the ownership check.

`write_account` takes the state to store, not a patch over it: what the account carries is what the list ends up holding. Renaming, naming the tracked default and recording a use are one read-modify-write.

Gone: `create_additional_account`, `create_default_account`, `update_account`, `update_existing_account`, `set_account_last_used`, `account_reference`, `with_account_reference_mut`.

## Three defects go with them

- **`create_additional_account` orphaned a record.** It stored the record before two fallible steps. On the IC `Err` commits what came before, so a refusal left a record holding a consumed account number that nothing referenced. The record now writes inside the gate, after everything that can fail. (Same defect as the one fixed in `create_default_account` on #4232; missed there.)
- **`get_account_for_origin(None)` bypassed the list read** — it returned `Account::synthetic` without touching storage, and the attribute endpoints called `calculate_seed()` on it, so they signed for a default the identity may have named or moved away. This is @MRmarioruci's question on this PR. Every account now comes out of `read_account`, and `Account::synthetic` is `#[cfg(test)]`, so the compiler enforces it rather than the code happening to comply.
- **`known_app_num` let the origin and the application number diverge.** `read_account` took the caller's number and never checked it named the origin beside it, so references could be read from one application while the seed came from the other origin. Two callers outside storage passed it. Storage resolves the origin itself now.

Naming the tracked default writes its config and its reference together, so a config naming an account no reference names can no longer be left behind.

## Tests

698 unit tests pass. The tests that pinned the old shapes are rewritten against the new surface rather than deleted, and the two that built an `Account` by hand now read first, like every real caller.


src/internet_identity/src/storage.rs
MRmarioruci: Is read_account the only path that needs this check? get_account_for_origin still returns a synthetic default for None, and the attribute endpoints use it to calculate the signing seed. Should that path use read_account

src/internet_identity/src/storage.rs
```diff
@@ -2060,11 +2060,10 @@ impl<M: Memory + Clone> Storage<M> {
             // The tracked default.
             None => match row {
                 ReferenceRow::Untouched => Some(synthetic_default()),
-                // XXX WARNING: a tombstone means the default moved away, so answering
-                // with a synthetic one lets its former owner reconstruct it at the same
-                // principal. Kept for now because refusing would lock out an identity
-                // that moved its default away and then reached the account limit.
-                ReferenceRow::Tombstone => Some(synthetic_default()),
+                // A tombstone means the default moved away. Answering with a
+                // synthetic one would let its former owner reconstruct it at the same
```
aterga: Has this been a bug all along, or is this a fix for the code introduced in the PR stack?

src/internet_identity/src/storage.rs
sea-snake: Yes, and it does now — that question is what turned this PR into the shape it is.

`get_account_for_origin` no longer short-circuits: the `None` case goes through `storage.read_account` like every other, so the tracked default is only returned where the identity's row still names it. An identity that named or moved its default away gets `NoSuchAccount` instead of a seed for an account it no longer has.

To make that stick rather than rely on nobody reintroducing it, `Account::synthetic` is now `#[cfg(test)]`. Production has no way to build a seed-bearing account outside `read_account`, which is the ownership check, so the compiler enforces this rather than the code happening to comply.

The PR description has been rewritten and the title changed — the one-line fix you reviewed is now a consequence of the read normalising, not a branch of its own.

src/internet_identity/src/storage.rs
sea-snake: All along — it is on `main` today, not something this stack introduced. `storage.rs` on `main` has the `if acc_ref_vec.is_empty()` arm returning a synthetic default, with an `XXX WARNING` above it saying it "would allow a user to regain access to their transferred default account".

It is only reachable once something can write an empty list, and nothing does yet, so it has never fired in production — which is why it survived as a warning rather than a bug report. #4232 made it unwritable from the write path; this PR removes the read that would have honoured it.

Fair warning that the PR has grown since you looked: the same absent-versus-empty rule turned out to be written in three more places, so it is now one normalising read and one write path rather than a deleted branch. Title and description are updated.

src/internet_identity/src/storage.rs
sea-snake: Correcting myself on one sentence above: "the compiler enforces this" is wrong.

`Account::synthetic` being `#[cfg(test)]` removes one constructor, but `Account::new`, `new_with_last_used` and `new_full` are all still `pub` and all build a seed-bearing account with `seed_from_anchor: None` — byte-identical to what `synthetic` built. So production can still construct a default account outside `read_account`; nothing stops it at compile time.

What is true is the part that matters for your question: `get_account_for_origin` no longer short-circuits, so every read of a default goes through `read_account` and its ownership check. That is a property of the call graph, held by review rather than by the type system.

Making it compiler-enforced would mean gating the seed derivation rather than the constructors, since it is `calculate_seed_with_salt` that turns an `Account` into a signing capability. That is a larger change than this PR, and I have noted it rather than folding it in here.

src/internet_identity/src/storage.rs
```diff
@@ -1576,238 +1508,466 @@ impl<M: Memory + Clone> Storage<M> {
...
+        let stored_anchor = self
+            .stable_anchor_account_counter_memory
+            .get(&anchor_number)
+            .unwrap_or_default();
+        let mut anchor_accounts = stored_anchor.stored_accounts;
+        let mut anchor_references = stored_anchor.stored_account_references;
+        let mut global_references = minting.global.stored_account_references;
+        for one in &validated {
+            let (accounts, references) = one.deltas.apply(
+                ReferenceCounter::Anchor { anchor_number },
```
MRmarioruci, line 1635: Where do we check that this anchor exists? A missing anchor gets empty counters here, so create_account can store an account and application that no identity owns. Could we reject it before minting anything?
```
if !self.stable_anchor_memory.contains_key(&anchor_number) {
    return Err(StorageError::AnchorNotFound { anchor_number });
}
```
Somewhere on the top should be fine.

src/internet_identity/src/storage.rs
```diff
@@ -1576,238 +1508,466 @@ impl<M: Memory + Clone> Storage<M> {
...
-        // An origin nothing has ever been stored under holds no reference to stamp,
-        // which is the same answer as a list that holds no reference for this account.
-        let Some(application_number) = self.lookup_application_number_with_origin(&origin) else {
-            return Ok(None);
-        };
+        // The account cap is a rule about the state this write leaves the identity in,
+        // not a question for a caller to ask first. Refusing here costs nothing, because
+        // nothing has been stored — which is the only reason a rule can live at the end of
+        // a write rather than in front of it.
+        if anchor_accounts > MAX_ANCHOR_ACCOUNTS {
```
MRmarioruci: Worth noting this refuses on the resulting count rather than on the change, so it also fires on writes that add nothing. Recording a use goes through here with a zero delta. Today the counter tops out at exactly the cap so nothing is reachable, but it makes lowering MAX_ANCHOR_ACCOUNTS lock existing identities out of sign-in rather than just capping them, which the pre-check shape did not.

src/internet_identity/src/storage.rs
```diff
@@ -1928,350 +1915,283 @@ impl<M: Memory + Clone> Storage<M> {
...
-                references
-            }
-        };
-
-        self.write_reference_list(anchor_number, application_number, references)?;
-
-        // Return created default account
-        Ok(Account::new_full(
+        self.write_account_state(
             anchor_number,
```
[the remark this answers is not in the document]
sea-snake: Right about the gap, though not about the route in. `set_default_account_for_origin` reads the account first and refuses with `NoSuchAccount`, and no code path removes an account, so no caller could reach the state you describe.

But the doc comment was wrong about *why*, and the underlying point stands: the default account is a value derived from the origin's account reference list, and it was the last such value maintained outside the write. Validation built the list and then, twelve lines later, took the caller's config exactly as given without ever relating the two.

Fixed in `b67df08e7`, by relating them — and by repairing rather than refusing:

- the target is the caller's config where it supplied one, and the stored config otherwise, because a caller that is not changing the default passes none, and "leave it alone" must not mean "leave it naming an account that has gone"
- where the list being written does not hold that target, the default moves to the first reference the list still has

Refusing would mean a future write that drops a reference gets rejected unless its caller also remembered to move the default — reintroducing the hand-maintenance at a new site. This way an account removal clears or moves the default for free.

`set_default_account` also stops building a config and moves one field of the stored one instead, since building one decides every field it leaves out.

src/internet_identity/src/storage.rs
sea-snake: You're right, and the session cap ten lines above already had the shape you're describing —
`(session_delta != 0).then(...)`, so it never refuses a write that did not move the count.
The account cap had no such guard, so the two sat side by side and disagreed.

Fixed in `def7645c4`: the account cap now refuses only where the write *grew* the count.

```rust
let accounts_delta: i64 = validated.iter().map(|one| one.deltas.accounts).sum();
if accounts_delta > 0 && anchor_accounts > MAX_ANCHOR_ACCOUNTS {
```

`> 0` rather than the neighbour's `!= 0`, because the neighbour has the same wart one step
further along: it would refuse a write that *removes* while the count is still over, which is
the one write that would fix being over. So the session cap got the same treatment in
`19bd01fa9` — with the guard on the refusal rather than on `session_count`, since that
value is also the number the write stores and a revocation still has to store where the
count moved to.

Neither state is reachable while the caps are fixed. Lowering one is what reaches them, and
then an identity above the new cap keeps its accounts and its sign-in, which is what a
lowered cap should mean.

src/internet_identity/src/storage.rs
sea-snake: It is checked, with exactly the error you reach for. `create_account`'s first act is
`self.read(anchor_number)?`, and `read` returns `StorageError::AnchorNotFound { anchor_number }`
for a missing anchor — before anything is minted, since `MintingState` only bumps in-memory
fields and the allocator moves in apply.

Worth saying it is not a check at that one call site, because that would be a thing to
remember at the next one. `write_account_state` takes `anchor: Anchor` **by value**, so no
write path can reach it without having read the identity record first — a missing anchor
refuses every write, not just this one.

The comment on that line makes your argument: "an identity that does not exist has nothing to
hold what is about to be written — the counters, the account reference lists and the session
count all key on a record that would not be there."

MRmarioruci (approved):

PR #4235, opened 2026-08-19
feat(be): track the default account on first use at an origin

sea-snake:

Signing in with an identity's default account at an origin it has never used records **nothing** — no timestamp, no reference list, no application list. That is the common case, so the canister only knows about accounts a user has explicitly named. Nothing is ever removed either, so applications would accumulate for the life of the canister.

This PR makes first use of the default account a stored fact, and bounds what that costs.

## Tracking is one deleted guard

The base PR refuses to store a reference list whose only content is the derived default: absence already says an identity has its default at an origin, so a list repeating that stores nothing. Recording *use* is what changes that — a timestamp has nowhere else to live.

So tracking is the removal of that refusal from `write_account_state`, and recording a use is the same `read_account` / set `last_used` / `write_account` every other write goes through. A tracked default is an existing shape, `AccountReference { account_number: None, last_used: Some(t) }`, so nothing new is stored and a named account still cannot be created from nothing. A default that was given away is not recreated: a tombstone stays a tombstone.

## Why tracking, eviction and removal are one PR

They cannot be deployed separately, and this was measured rather than assumed: with tracking alone, one anchor signing in at 2,000 origins creates 2,000 reference lists and 2,000 application lists with no cap and no reclamation. Tracking mints an application list on an operation with no per-anchor cap; eviction is what bounds that and the only thing that makes a reference count fall; removal is dead code without eviction, since a count that never falls never reaches zero. A prefix of the three grows stable memory without bound on an attacker-drivable path.

## Eviction

A cap of 500 lists holding nothing but a tracked default, screened by the cheap upper bound `stored_account_references - stored_accounts` with the scan as the authority. Sign-in **evicts rather than failing**, in bounded batches — and where the screen fires but the scan finds no candidate, which an identity holding 500 lists that each pair a default with a named account can do, it proceeds having evicted nothing.

Eviction is part of the write. Victims are chosen in validate, against the state the write is about to leave behind — the counters plus this call's deltas, with every origin the write touches excluded — and the removals go in the same batch. It used to hang off a second call made once the first had returned, which made a write that pushed an identity over the cap two atomic units: on the IC an `Err` from the second commits the first, so a list could go while its sessions stayed.

It takes the whole list and its config list, never leaves an empty list behind, and spares a default while a named account shares its list. The list the caller just wrote is excluded, since every message in a round reads the same `time()` and would otherwise tie. Non-destructive: it drops a timestamp, and the account returns at the same principal on next use.

## Removal

The predicate is the existing `StorableApplication.stored_account_references`, an accurate live count today because every insert is counted and nothing has ever been removed, so it only needed decrementing — which #4232 made a property of one function. No new field, no migration. Reaching zero removes the application list, and the origin-index entry when it still points at that number.

A removal reads the **stored** list rather than the normalising one: an absent list must not look prunable, or its counters would be decremented for a list that never existed.

## Tests

`default_account_tracking_tests` (8), `tracked_default_eviction_tests` (14), `application_removal_tests` (9). PocketIC drives 500 real sign-ins and asserts an evicted origin returns at the same `user_key`.


src/internet_identity/src/account_management.rs
```diff
@@ -159,6 +159,7 @@ pub fn set_default_account_for_origin(
     origin: FrontendHostname,
     account_number: Option<AccountNumber>,
 ) -> Result<AccountInfo, SetDefaultAccountError> {
+    check_frontend_length(&origin);
     let application_number = storage_borrow_mut(|storage| {
         storage.lookup_or_insert_application_number_with_origin(&origin)
```
MRmarioruci: What happens if this is called with a named account that does not exist at this origin? We create the application row here, then return `NoSuchAccount` below without creating an account reference. Could that leave the application behind forever cause the reaping will fail? Maybe this should happen where the account reference is created if that's the case.

sea-snake: Confirmed, and it is worse than the question asks — thank you for catching it.

The row is stranded exactly as you guessed: `remove_unreferenced_application` is reachable only from `apply_reference_counter_deltas`, which is reachable only from `write_reference_list`. This path never writes a reference list, so none of that runs and the row stays for the life of the canister.

The part beyond the question: `origin` is caller-supplied and `set_default_account` is only anchor-authenticated. `check_frontend_length` bounds each origin's length; nothing bounds the number of distinct origins. So any signed-in identity could mint an unreapable application row per call with a fabricated origin and a bogus `account_number` — and burn an application number each time, which is never reissued. Storage growth, not just an orphan.

Fixed by not writing anything until the operation is known to succeed: the named account is validated through the read-only `lookup_application_number_with_origin`, and `lookup_or_insert` moved below it. An origin with no application row cannot hold a named account, so that is `NoSuchAccount` without writing.

Two more sites had the same shape, found while fixing this one:

- **`create_default_account`** — user-reachable too, and it stranded more: an allocated account number, the stored account, the application row *and* the `AnchorApplicationConfig` written by `set_anchor_application_config`, all committed before the `MissingAccount` its own comment predicts. It now reads the reference list and checks for a default before allocating anything.
- **`create_additional_account`** — same shape but only reachable on a storage fault, since the sole error after its writes is `write_reference_list` failing. The allocation moved ahead of the writes so all three read the same way.

Covered by a unit test that the refusal leaves no account and no rewritten config, and an integration test asserting `internet_identity_total_application_count` is unchanged after a refused `set_default_account` — that metric is where a stranded row shows up.

src/internet_identity/src/storage.rs
```diff
@@ -1686,7 +1721,132 @@ impl<M: Memory + Clone> Storage<M> {
...
+            .into_iter()
+            .filter(|(application_number, _)| *application_number != just_written)
+            .collect();
+        if candidates.len() as u64 <= EVICTABLE_DEFAULT_ACCOUNTS_WATERMARK {
+            return Ok(());
+        }
+
+        candidates.sort_by_key(|(application_number, last_used)| (*last_used, *application_number));
+
+        let victims = u64::min(
```
MRmarioruci, line 1748: I think this stops at 451 not 450.. Should be fine.. cause I think you also have a test that asserts this too..

src/internet_identity/src/account_management.rs
```diff
@@ -172,19 +180,27 @@ pub fn set_default_account_for_origin(
...
+        storage.lookup_or_insert_application_number_with_origin(&origin)
+    });
+
     let config = AnchorApplicationConfig {
         default_account_number: account_number,
     };
 
     storage_borrow_mut(|storage| {
+        storage
+            .ensure_account_reference_list(anchor_number, application_number)
```
aterga: Nit

```suggestion
            .ensure_account_reference_list_exists(anchor_number, application_number)
```

src/internet_identity/src/storage.rs
```diff
@@ -2594,6 +2790,19 @@ impl ReferenceListDeltas {
         }
     }
 
+    /// What retiring a row holding `previous` does to the counters.
```
aterga: Let's get rid of the row terminology? 

Also, the function name isn't great. Removing sounds like a process, not an action. Who is removing what?

src/internet_identity/src/storage.rs
```diff
@@ -1656,7 +1691,142 @@ impl<M: Memory + Clone> Storage<M> {
...
+            .account_references(anchor_number, application_number)
+            .is_some()
+        {
+            return Ok(());
+        }
+
+        self.write_reference_list(
+            anchor_number,
+            application_number,
+            vec![AccountReference {
```
aterga: Nit

```suggestion
            vec![AccountReference::default()]
```

src/internet_identity/src/storage.rs
```diff
@@ -1656,7 +1691,142 @@ impl<M: Memory + Clone> Storage<M> {
...
+                account_number: None,
+                last_used: None,
+            }],
+        )?;
+        self.evict_idle_tracked_defaults(anchor_number, application_number)
```
aterga: What exactly is being evicted? (This should be apparent from the name of this function)

src/internet_identity/src/storage.rs
```diff
@@ -1656,7 +1691,142 @@ impl<M: Memory + Clone> Storage<M> {
...
+    }
+
+    /// Removes a reference-list row and everything derived from it.
```
aterga:
```suggestion
    /// Removes an account list for a given application and everything derived from it.
    ///
    /// Removing here means that ...
```

src/internet_identity/src/storage.rs
```diff
@@ -1664,15 +1693,81 @@ impl<M: Memory + Clone> Storage<M> {
...
+                // does not get to fail the write it is riding on.
+                let application = self.stable_application_memory.get(&application_number)?;
+                (!written_origins.contains(&application.origin)).then_some((
+                    last_used,
+                    application_number,
+                    application.origin,
+                ))
+            })
+            .collect();
+        if candidates.len() as u64 <= EVICTABLE_DEFAULT_ACCOUNTS_WATERMARK {
```
MRmarioruci: I think this shoukld return until candidates.len() reaches MAX_EVICTABLE_DEFAULT_ACCOUNTS, rather than the watermark? Defaults stored with named accounts are included in the first count but cannot be evicted, so we may start evicting before reaching the limit.

src/internet_identity/src/storage.rs
```diff
@@ -2624,21 +2900,41 @@ impl From<AccountReference> for AccountReferenceWrite {
...
 ///
-/// The config is `None` where the write leaves the stored one alone.
-pub type AccountReferenceListWrite = (Vec<AccountReferenceWrite>, Option<AnchorApplicationConfig>);
+/// `None` removes the account reference list, and the derived default comes back with it —
+/// which is the same state an absent list already denotes on the read side. `Some` with an
+/// empty list is the opposite: a tombstone, which nothing may create yet.
+///
+/// The inner config is `None` where the write leaves the stored one alone. A removal cannot
+/// carry one, because there is nowhere in the type to put it.
+pub type AccountReferenceListWrite =
```
MRmarioruci, line 3004: Could this be an enum?

MRmarioruci:
```suggestion
enum AccountReferenceListWrite {
    Remove,
    Store {
        references: NonEmptyAccountReferences,
        config: ConfigWrite,
    },
}

enum ConfigWrite {
    Keep,
    Set(AnchorApplicationConfig),
}
```
A bit more explicit and removes the nested Options

src/internet_identity/src/storage/storable/application.rs
```diff
@@ -17,6 +17,21 @@ pub struct StorableApplication {
...
+    /// separately this application would look unreferenced and be retired — and the next
+    /// visit would mint a fresh application number the tombstone no longer applies to,
+    /// handing the identity back the default it had moved away from.
+    ///
+    /// Absent from every application stored before this field existed, and `0` is the
+    /// truth for those: nothing could write an empty reference list when they were
+    /// written. `default` is what makes that absence decode rather than trap.
+    #[n(3)]
+    #[cbor(default)]
+    pub tombstones: u64,
```
MRmarioruci: Nit.. The more I look at this term the more time I have to put in in order to try and understand what it does .
Basically something was there and removed x nothing was ever there.
Maye: had_defaults or something simpler?

src/internet_identity/src/storage.rs
sea-snake: 451, and you're right that a test asserts it — `evicting_drops_the_least_recently_used_down_to_the_watermark`, where the `- 1` in

```rust
let evicted = MAX_EVICTABLE_DEFAULT_ACCOUNTS - 1 - EVICTABLE_DEFAULT_ACCOUNTS_WATERMARK;
```

is exactly the origin the triggering write touched, which is excluded from being a victim of its own write. So the pass trims the *candidates* to the watermark and that one survives on top.

No behaviour change, but you had to derive that from the code, so in `b5f7a49b9` the `- 1` says it, and `MAX_EVICTABLE_DEFAULT_ACCOUNTS` stops calling itself a "cap" — it is where eviction triggers, and the count comes to rest a little above the watermark rather than at either number.

src/internet_identity/src/account_management.rs
sea-snake: The function is gone rather than renamed — `ensure_account_reference_list` and `lookup_or_insert_application_number_with_origin` both have zero references now, and `set_default_account_for_origin` is a read followed by `storage.set_default_account` with no application lookup and no list-creation step of its own.

Worth saying why, because the nit was sitting on the seam that got removed: a caller had to remember to materialise a list before writing a config against it. The write owns that now — "an origin nothing is stored under is created here, application and all, with its counters already right", and a refusal cannot leave behind an application no counter will ever retire.

src/internet_identity/src/storage.rs
sea-snake: Both done.

Row terminology went with the rewrite — two occurrences left in `storage.rs` and they are the per-identity config row in the legacy migration note, a different subject. The comment you were reading now says "What retiring a *list* holding `previous` does to the counters".

And the name is `retiring` as of `b5f7a49b9`. Its own doc comment already said retiring, and retire is this file's word for it everywhere else: an application no counter will ever retire, retirement running off a write to account state, a list retired when a live tracked default is all it holds. Only the name said otherwise.

I kept the gerund rather than going to `for_retirement`, because the sibling constructor sets the reading — `ReferenceListDeltas::between(previous, new)` and `ReferenceListDeltas::retiring(previous)` are both "the deltas *of* doing this", with the type as the subject. Changing one and leaving the other would read worse.

src/internet_identity/src/storage.rs
sea-snake: Not taking this one, for two reasons.

The suggestion would not compile: `AccountReference` derives no `Default` (`Clone, Debug, Eq, PartialEq, Ord, PartialOrd`), unlike `AccountsCounter` right above it.

And the literal it sat on is gone anyway — the type gained a third field, `sessions`, and a `new(account_number, last_used)` constructor that fixes `sessions: vec![]`. So the substance, not writing the field list at call sites, is what happened, through `new` rather than `Default`.

I would rather not add `Default` either. `default()` at a call site says "whatever the defaults happen to be", and a field added later with a non-trivial default gets picked up silently. Here `account_number: None` is not incidental — it is what makes a reference the tracked default — and it belongs in view where it is written. The two remaining sites also mean different things despite identical construction: one is the tracked default, the other is an account about to be named whose number the write mints. So a named constructor would be wrong at one of them.

src/internet_identity/src/storage.rs
sea-snake: Fair, and the old name was worse than the behaviour. It is split in two now and both names answer it: `evictable_default_lists` ("Lists whose only reference is a tracked default") and `evictable_after` ("The tracked defaults this write leaves over the cap, as origins to remove").

What is evicted is a stored *list*, never an account. The list holds nothing but a tracked default, and an absent list normalises back to the derived default, so the only thing actually lost is the `last_used` stamp at the origin idle longest — plus that origin's sessions, which `validate_removal` now says explicitly. "Evict tracked defaults" read as losing accounts, which is presumably why it drew the question.

src/internet_identity/src/storage.rs
sea-snake: Done in `b5f7a49b9`, including the part you left as a prompt.

The "removing here means that ..." is `validate_removal`'s doc now: a list is retired only when a live tracked default is all it holds, an absent list has nothing to remove, an empty list is a tombstone whose removal would make the default reconstructible again, and a list holding named accounts would lose references nothing else records.

The "everything derived from it" half was genuinely missing and is the part worth having. Four things go, and only one of them is visible to a person: the list, the config keyed beside it, the reference and tombstone counters, and every session the list held — so evicting an idle origin's list signs that origin's sessions out. That is also why a list holding anything more than a tracked default is never a candidate.

It says "at one origin" rather than "for a given application", since the write speaks origins and an application number is storage's handle for one.

src/internet_identity/src/storage.rs
sea-snake: You're right, and it was costing more than an early pass. Fixed in `3d896709a`.

The trigger is now the cap:

```rust
if (candidates.len() as u64) < MAX_EVICTABLE_DEFAULT_ACCOUNTS {
    return Vec::new();
}
```

and the pass still trims down to the watermark, so the watermark is where it stops rather than where it starts.

Your diagnosis of the first count is exactly right, and it can't be fixed there: it comes from the anchor's counters, which know how many references and accounts an identity has and nothing about how they're spread across lists — so it counts every numberless reference, including the ones in lists that also hold named accounts, which `evictable_default_lists` never picks. Being an upper bound is all it needs to be, since its only job is keeping the scan off the sign-in path, and it's one-sided: it can send us into a scan we didn't need, never skip one we did.

Worth noting what the early pass actually cost, because it isn't just churn — an evicted list takes that origin's sessions with it. So an identity with, say, 460 default-only lists and 100 mixed ones was under the cap and getting signed out of apps anyway.

The test moved with it: it signs in one past the cap now, since the origin a write is touching is never a candidate for its own eviction, so the pass first runs when the *other* lists reach the cap.

src/internet_identity/src/storage.rs
sea-snake: It should, and the file argues your case two lines below the type — `ListWrite { Untouched, Stored, Removed }` and `ApplicationWrite { Untouched, Stored, Retired }` name exactly this on the apply side, while the input is four meanings spelled as nested `Option`s with nine lines of doc working out which `None` is which.

Not doing it in this stack, for one reason: it's entirely internal. `write_account_state` is private, as are `validate_account_state` and `apply_account_state`. The API is the `pub fn` on `Storage` — `create_account`, `write_account`, `set_default_account`, `create_session`, `record_session_use`, `revoke_session`, `revoke_browser_sessions` — and none of them exposes this shape. `AccountReferenceListWrite` is `pub` only to reach `storage/tests.rs`, a child module that would see it anyway. So there's no consumer to migrate and it can be done at any time.

Two things worth writing down for whenever it is, both of which came out of taking your question seriously:

`Removed` would be the wrong name. `validate_removal` retires the list, the config beside it, the counters and the sessions the list held — but only when a single numberless reference is all the list holds. Anything else returns `untouched`: a silent decline, not a refusal, because eviction rides on a sign-in and mustn't fail it. Something like `RetireIfIdle` says that; `Removed` promises removal and usually declines.

And that variant has no production caller. All seven build `Some((refs, config))`; eviction reaches removal by calling `validate_removal` directly, bypassing the type. The only construction of the outer `None` is a test helper — so the type's removal route and the real one are parallel, and only the unused one goes through the type.

src/internet_identity/src/storage/storable/application.rs
sea-snake: Your reading is right — the count is what separates *emptied* from *never existed*, and that's the part the name doesn't carry.

Renamed to `stored_tombstones` in `3d896709a`, which puts it in the same series as its neighbours `stored_accounts` and `stored_account_references`. Those say what is stored; `tombstones` named a concept and left you to work out it was a count of lists. The name on the wire is `#[n(3)]`, so nothing about the encoding moves.

Keeping the word itself, though. It appears 81 times across the module — `tombstone_for_testing`, `StorableAccountReferenceListError::Empty`, `validate_removal`, and six test names including `tombstoned_list_does_not_revive_the_default` — so renaming just the field would split the vocabulary, and it's the standard term for a marker kept so a deletion is remembered.

`had_defaults` would also be wrong on the facts: it isn't about defaults but about every account reference having moved away, and "had" reads as a boolean about the past where this is a count of lists tombstoned right now.

The real fix for "the more I look the more time I have to put in" is the doc, which explained the mechanism well over eleven lines and never opened with the plain version. It now leads with: how many identities hold an emptied list here, counted separately because an emptied list is invisible to the reference count.

MRmarioruci (approved):

PR #4237, opened 2026-08-19
feat(be): reap applications no anchor references any more

sea-snake:
# Motivation

# Changes

# Tests

---
Design: #4222 (`docs/ongoing/tracked-default-accounts.md`). Overview: #4230.

### DIFF
--- a/src/internet_identity/src/storage.rs
+++ b/src/internet_identity/src/storage.rs
@@ -2002,14 +2002,34 @@ impl<M: Memory + Clone> Storage<M> {
             application.stored_accounts,
             application.stored_account_references,
         );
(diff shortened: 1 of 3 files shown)


sea-snake:
Folded into #4235. These three cannot be deployed separately: tracking makes every sign-in at a new origin mint an application row from an operation with no per-anchor cap, eviction is what bounds that and the only thing that makes an application's reference count fall, and reaping is dead code without it. Verified rather than assumed — with tracking alone, one anchor signing in at 2 000 origins creates 2 000 reference rows and 2 000 application rows. Shipping a prefix would mean unbounded, attacker-drivable stable-memory growth, so they are now one PR.

PR #4238, opened 2026-08-19
feat(be): index accounts by the principal a dapp sees

sea-snake:
Design: #4222. Overview: #4230. Nothing reads the index yet, so this is complete on its own but not yet authoritative.

Anchors are reverse-indexed five ways; accounts are not. So given the principal a dapp sees, the canister cannot tell which anchor, application or account produced it, because the seed is hashed.

- A new map at memory index 34: `Principal -> StorableAccountLocator`, the `(anchor, application, account)` triple. The application number is load-bearing: for a default account the account number is `None`, so the triple is the only thing saying which origin it belongs to.
- Maintained by the same write path as the counters, so no call site has to remember.
- It **diffs values, not just keys**, which is the case the credential indexes do not have: materializing a default rewrites `None` to `Some(n)`, leaving the principal unchanged while the locator gains an account number. A key-only diff would see no change and leave a stale value.
- Removals are compare-and-delete — only if the stored locator still names this anchor — so a future account move cannot drop the recipient's live entry.
- Deriving a principal needs the salt, and `state::salt()` traps when unset, so the write checks it synchronously and returns a typed error instead. The salt is written exactly once and can never be unset or rotated, so that branch is reachable only on a **freshly installed** canister that has never issued a delegation.
- The account-mutating endpoints `await ensure_salt_set()` on entry, ahead of the cap check rather than between the check and the write, so check and write still land in one message and two concurrent calls cannot both pass the same check. Without it, `create_account` on a fresh canister returns `SaltNotSet`; on a live canister the await resolves immediately.

`calculate_anchor_seed` gains a salt-parameterised variant and `calculate_account_seed` is replaced by one, so storage can derive without reading global state, which is also what makes the index unit-testable.

Tests: `account_principal_index_tests` (8), including materializing a default updating the locator **under the same principal**, eviction and application removal leaving no entries behind, and a write without a salt writing nothing.


src/internet_identity/src/storage.rs
```diff
@@ -1899,11 +1919,101 @@ impl<M: Memory + Clone> Storage<M> {
 
         let deltas = ReferenceListDeltas::between(&previous, &current);
 
+        self.sync_account_principal_index(anchor_number, application_number, &previous, &current)?;
```
```suggestion
if previous.len() != current.len()
    || previous
        .iter()
        .zip(&current)
        .any(|(previous, current)| previous.account_number != current.account_number)
{
    self.sync_account_principal_index(
        anchor_number,
        application_number,
        &previous,
        &current,
    )?;
}
```
Nit. The derived principal only depends on the account/origin, so checking the account numbers first would avoid rebuilding the index on normal sign-ins.

src/internet_identity/src/main.rs
```diff
@@ -407,11 +407,12 @@ fn get_accounts(
 }
 
 #[update]
-fn create_account(
+async fn create_account(
     anchor_number: AnchorNumber,
     origin: FrontendHostname,
     name: String,
 ) -> Result<AccountInfo, CreateAccountError> {
+    state::ensure_salt_set().await;
```
MRmarioruci: Nit. Could we initialize the salt at install insteaf of lazily here from account operations? If we have two requests arriving before the salt is set, they can both start creating it, the first one saves it, and the second fails cause the salt exists. A bit racy..

src/internet_identity/src/main.rs
```diff
@@ -423,12 +424,13 @@ fn create_account(
...
 
 #[update]
-fn update_account(
+async fn update_account(
     anchor_number: AnchorNumber,
     origin: FrontendHostname,
     account_number: Option<AccountNumber>,
     update: AccountUpdate,
 ) -> Result<AccountInfo, UpdateAccountError> {
+    state::ensure_salt_set().await;
```
MRmarioruci: same as above if applicable.

src/internet_identity/src/main.rs
```diff
@@ -470,11 +472,12 @@ impl From<IdentityUpdateError> for SetDefaultAccountError {
 }
 
 #[update]
-fn set_default_account(
+async fn set_default_account(
     anchor_number: AnchorNumber,
     origin: FrontendHostname,
     account_number: Option<AccountNumber>,
 ) -> Result<AccountInfo, SetDefaultAccountError> {
+    state::ensure_salt_set().await;
```
MRmarioruci: ditto

src/internet_identity/src/storage.rs
sea-snake: Taken as written. Calling it a nit undersells it: this is the sign-in path, not an occasional one.

`sync_account_principal_index` ran on every `write_reference_list`, and derives a principal for every reference in both the previous and current lists — so 2N seed hashes plus 2N principal derivations per write. `record_account_use` writes on every sign-in to stamp `last_used`, and your premise holds exactly: `calculate_seed_with_salt` derives from the anchor, origin, account number and `seed_from_anchor`, none of which a `last_used` stamp touches. So the hottest write in the system was recomputing everything and discarding it as unchanged.

I checked the one thing that could make the guard wrong — a principal changing while the account numbers stay put, which needs `seed_from_anchor` to be mutable. It is not: written at exactly two sites, both at creation, never after. The only two closures reaching `with_account_mut` set `last_used` and `name`, and `name` is not in the seed either. Noting it here so the next person changing the seed knows this guard depends on it.

The sync also became infallible while I was in there: its two refusals (the salt and the origin) are hoisted into the caller, which already had the application row, so it cannot fail part-way through its own writes.

src/internet_identity/src/main.rs
sea-snake: Fixed, and taken further than "a bit racy" — the loser **traps**. The check happens before the await:

```rust
storage_borrow(|storage| { if storage.salt().is_some() { trap("Salt already set"); } });
let salt = random_salt().await;              // both messages get here
storage_borrow_mut(|storage| storage.update_salt(salt));   // update_salt traps if set
```

so the second to resume dies on a request that was perfectly valid.

Rather than making the race survivable, the initialisation is gone from these endpoints: a canister sets its salt once via `init_salt`, and `create_account`, `update_account` and `set_default_account` now refuse with `SaltNotSet` when it is absent. There is no interleaving point left, which is also why all three stop being `async`.

`StorageError::SaltNotSet` already existed — `sync_account_principal_index` in this PR uses it — so this is the same mindset applied one layer out.

Fallout worth knowing: these endpoints *were* the initialisation for some flows, so four integration suites were relying on `create_account` to set the salt for them. They now call `init_salt` at install, the way a deployment does.

src/internet_identity/src/main.rs
sea-snake: Covered by the reply on `create_account` above — `update_account` drops `ensure_salt_set` and refuses with `SaltNotSet` instead, and is no longer `async`.

src/internet_identity/src/main.rs
sea-snake: Same here — `set_default_account` drops `ensure_salt_set`, refuses with `SaltNotSet`, and is no longer `async`. See the `create_account` thread for the reasoning.

src/internet_identity/src/storage.rs
```diff
@@ -1896,22 +1927,133 @@ impl<M: Memory + Clone> Storage<M> {
...
+        let previous_references = previous_row.references();
+        let accounts_changed = previous_references.len() != new_references.as_slice().len()
+            || previous_references
+                .iter()
+                .zip(new_references.as_slice())
+                .any(|(previous, new)| previous.account_number != new.account_number);
+        // Resolved before anything is written, so a missing salt refuses here rather
+        // than half-way through.
+        let salt = if accounts_changed {
+            Some(*self.salt().ok_or(StorageError::SaltNotSet)?)
```
MRmarioruci, line 1953: Does this actually refuse before anything is written by the caller? create_additional_account and create_default_account store the account before reaching here, and set_default_account_for_origin may already have created the application

src/internet_identity/src/storage.rs
```diff
@@ -1896,22 +1927,133 @@ impl<M: Memory + Clone> Storage<M> {
...
+    fn sync_account_principal_index(
+        &mut self,
+        anchor_number: AnchorNumber,
+        application_number: ApplicationNumber,
+        origin: &FrontendHostname,
+        salt: &[u8; 32],
+        previous: &[AccountReference],
+        current: &[AccountReference],
+    ) {
+        let previous_entries =
```
MRmarioruci, line 2313: Do we need to derive both full lists for a one-account change? Adding the 500th account reads and hashes roughly 1k references. Could we compare the account numbers first and derive principals only for entries that were added, removed, replaced?

src/internet_identity/src/storage.rs
```diff
@@ -1896,22 +1927,133 @@ impl<M: Memory + Clone> Storage<M> {
...
+            self.account_principals(anchor_number, application_number, origin, salt, current);
+
+        for (principal, locator) in &previous_entries {
+            if current_entries.contains_key(principal) {
+                continue;
+            }
+            if self
+                .lookup_account_with_principal_memory
+                .get(principal)
+                .is_some_and(|stored| stored.anchor_number == locator.anchor_number)
```
MRmarioruci, line 2325: Why do we only compare anchor_number here? We already have the full old locator, so could we remove it only when the stored locator matches exactly

src/internet_identity/src/storage.rs
```diff
@@ -1915,6 +1936,25 @@ impl<M: Memory + Clone> Storage<M> {
             (ListWrite::Untouched, ReferenceListDeltas::default())
         };
 
+        // A derived principal is a function of the anchor, the origin and the account
+        // number, so a write that leaves every account number in place — a `last_used`
+        // stamp, which is every sign-in — cannot have changed one. Skipping the sync
+        // there keeps the hottest write in the system off a per-account hash.
+        let previous_references = stored.clone().unwrap_or_default();
```
MRmarioruci: Can we move stored here instead of cloning it? Everything above only borrows it, so let previous_references = stored.unwrap_or_default(); should work.

src/internet_identity/src/storage/storable/account_key.rs
```diff
@@ -0,0 +1,40 @@
...
+        let mut buffer = Vec::new();
+        minicbor::encode(self, &mut buffer).expect("failed to encode StorableAccountKey");
+        Cow::Owned(buffer)
+    }
+
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Self {
+        minicbor::decode(&bytes).expect("failed to decode StorableAccountKey")
+    }
+
+    const BOUND: Bound = Bound::Unbounded;
```
MRmarioruci:
```suggestion
    const BOUND: Bound = Bound::Bounded {
        max_size: 31,
        is_fixed_size: false,
    };
```

src/internet_identity/src/storage.rs
sea-snake: It does now, and structurally rather than by careful ordering. This was the failure mode the single write was built for — on the IC an `Err` commits everything written before it and only a trap rolls back — so it is worth spelling out what makes it true:

- `create_additional_account` and `create_default_account` are gone; zero references in `storage.rs` and `account_management.rs`
- every stable-memory write in this domain is inside apply — the accounts, the origin index, the lists, the config — or inside the two syncs apply calls. Nothing in the account-state path writes anywhere else
- both salt refusals are in validation, one in `validate_account_reference_list` and one in `validate_removal`
- the allocators do not touch storage: `MintingState` holds the next application number and the global counter in memory and only bumps those fields, so a refusal mints nothing
- and your `set_default_account_for_origin` point is the other half of this — `lookup_or_insert_application_number_with_origin` is gone, the write creates the application, and a refusal cannot leave behind an application no counter will ever retire

The contract is written down on `write_account_state` now, so the next person does not have to re-derive it.

src/internet_identity/src/storage.rs
sea-snake: Good catch, and the cost was worse than a thousand hashes — `account_principals` does a stable read per numbered reference too, so adding the 500th account was roughly a thousand reads, a thousand seed hashes and a thousand principal derivations in order to change one index entry.

Fixed in `d5ca19bab`, diffing on the account number as you suggested. The reason that is sound: a principal comes from the salt, the origin and the account the reference names, never from its name or its last use. `seed_from_anchor` is the only other thing a seed is taken from, and it is written where a number is minted and nowhere else, so it cannot move under a number that already exists. Same number, same principal, same locator — nothing to do.

Naming a tracked default still works for that reason rather than in spite of it: naming mints a number, so the key moves, both sides are derived, and the locator moves under a principal that does not. Four index tests cover it, including `materializing_a_default_updates_the_locator_under_the_same_principal`.

One deliberate trade. Deriving both lists re-asserted the entry of every account at the origin on every write, so a drifted entry was repaired by the next unrelated write; a number diff never revisits an unchanged account. The backfill sweep is the repair now — its completion is heap state, so it is forgotten at each upgrade and walks every list again. The doc comment says so, rather than leaving the next reader to find out.

src/internet_identity/src/storage.rs
sea-snake: Deliberate, and exact matching would be a regression — `seed_from_anchor` is why.

It is set at exactly one place, naming the tracked default, and the comment there says what it is for: "Its seed stays the anchor's, so the principal this identity already signs in with here is preserved." So one principal legitimately maps to two different locators across an account's life — `{anchor, app, None}` before naming and `{anchor, app, Some(n)}` after.

`StorableAccountKey::account_number` is therefore not part of the principal's identity, and an exact-match guard would be comparing a field that moves under a stable principal. Where a stored entry had drifted to the other form, exact matching would decline to remove it and leave a locator naming an account the list no longer holds.

The anchor check is not vestigial either: it is what stops one identity's write from deleting another identity's index entry, which unconditional removal would allow.

Leaving it as is, and more firmly than before — the change on your other thread drops the per-write re-assertion of every entry, which makes this removal path the only in-write repair for the writing anchor's own drifted entries.

src/internet_identity/src/storage.rs
sea-snake: Correct — nothing uses `stored` after that line; every later hit is a different identifier (`stored_number`, `application.stored_accounts`). Done in `7925ec331`.

There was a second clone twelve lines up that you didn't mention, in the `unchanged` comparison. That one can't be a move, since `stored` is still wanted below, but it doesn't need to copy either — it only needs to compare, and to build the derived default in the one branch that has no stored list:

```rust
let unchanged = match &stored {
    Some(stored) => &references == stored,
    None => references == Self::derived_default_references(),
};
```

Both are on the hot path: `record_session_use` is every sign-in, and it comes through here with a list that can hold up to `MAX_ANCHOR_ACCOUNTS` references, each owning a `Vec<SessionRecord>` — so the copy is deeper than a vector of pointers. Two of them per sign-in before this.

src/internet_identity/src/storage/storable/account_key.rs
sea-snake: Taken, and your 31 was exactly right for the map encoding. Landed in `7925ec331` as an array instead, at 28:

```rust
#[cbor(array)]
...
const BOUND: Bound = Bound::Bounded { max_size: 28, is_fixed_size: false };
```

Array because a map's integer keys buy the ability to add a field and still decode records written before it — which authoritative state needs (`application.rs` has a test for exactly that) and a derived index doesn't: every entry here is reconstructible from the account reference lists, and the backfill sweep already does it.

Checking what the bound actually buys turned up two things I'd have got wrong by assuming:

V2 nodes write each entry at its *actual* length, so bounding doesn't pad a typical entry out to its maximum. And `is_fixed_size` is read for **keys** only — for values the `u32` length prefix is written unconditionally — so a fixed-size value would buy nothing, and this is a value.

What it does buy is page sizing. `BTreeMap::new` computes `PageSize::Value(max_node_size * 3 / 4)` only when key *and* value are bounded, falling back to `DEFAULT_PAGE_SIZE = 1024` otherwise. `Principal` is already `Bounded { 29 }`, so declaring this one is what unlocks it: 614 bytes instead of 1024, about 40% off. The session index got the same treatment at 41 → 721.

Two consequences worth recording. It's now-or-never — the page size is computed once, when the map is created, so this was only free because neither index exists on `main` yet. And a generous bound isn't free either: every byte of `max_size` costs about 8 bytes of page, so the number has to be the true maximum. There's a test pinning it, since exceeding a declared bound panics at write time and a fourth field should break a test rather than a canister.

MRmarioruci (approved):

PR #4240, opened 2026-08-19
feat(be): sweep existing account references into the principal index

sea-snake:

Account references written before #4238 have no index entry. A miss that could mean either "unknown principal" or "not yet indexed" is not something authorization can be built on, and the design turns a miss into a denial — which is why this is a sweep and not lazy fill.

- `backfill_account_principal_index_batch(cursor, batch_size)` returns the next cursor and is done when a batch examines fewer keys than it asked for. Same shape as `migrate_sso_credentials_batch`, removed in #4192 once complete.
- Driven by an interval timer that clears itself on completion, started from both `init` and `post_upgrade`. With nothing to index, the first batch reports completion immediately.
- **Insert only.** A list derives exactly the entries the write path would have written for it, so a batch that runs twice writes the same values.
- A batch before the salt is set indexes nothing and leaves the sweep running, which only a fresh install can reach; it resumes on the next batch once a salt exists.
- A hidden `account_principal_index_backfill_status` query returns `(indexed_entries, is_done)` for monitoring.

Batch size is 2,000 keys, matching previous migrations. The work is measurable rather than estimated: `internet_identity_total_account_references_count` is exactly the number of entries to write.

Tests: `account_principal_index_backfill_tests` (5), including resumption across three batches and a repeated sweep writing nothing new. PocketIC installs the **previous release**, creates three accounts, upgrades, lets the timer run, and asserts the sweep indexed exactly 6 entries.


src/internet_identity/src/storage.rs
```diff
@@ -1926,6 +1926,75 @@ impl<M: Memory + Clone> Storage<M> {
...
+                &references,
+            ) {
+                if self.lookup_account_with_principal_memory.get(&principal)
+                    == Some(locator.clone())
+                {
+                    continue;
+                }
+                self.lookup_account_with_principal_memory
+                    .insert(principal, locator);
+                outcome.indexed += 1;
```
MRmarioruci: In the previous migrations we use to collect errors, I am not sure it's needed for this one. If for consistency we wanna do that, then please add.

sea-snake: Added — and there was something real being swallowed, so thank you for asking rather than assuming.

The `continue` on a missing application row is the one that matters:

```rust
let Some(origin) = self.stable_application_memory.get(&application_number)... else {
    continue;   // silently
};
```

With only an `indexed` count, a run that passed over every row looked identical to a run with nothing to do. `AccountPrincipalIndexBackfillOutcome` now carries `skipped`, the driver accumulates it, `account_principal_index_backfill_status` returns `(indexed, skipped, is_done)`, and the completion log reports both.

That state is reachable, which is why it is worth counting rather than ignoring: `remove_unreferenced_application` fires when `stored_account_references == 0`, and on this branch that value came from a saturating add. A counter that under-ran clamped to zero, the application row was retired, and any reference lists still pointing at it became rows this sweep would silently walk past. #4232 now refuses instead of clamping, so the state should not arise — the count is there to say so if it ever does.

I left the salt-absent early return as it is. It reports `is_done: false` so the sweep resumes, which is what the timer comment already says it relies on; it now has a line saying an unset salt means the canister has not finished starting up rather than the backfill having finished.

src/internet_identity/src/main.rs
```diff
@@ -843,6 +844,85 @@ fn initialize(maybe_arg: Option<InternetIdentityInit>) {
     if let Some(openid_configs) = config.openid_configs {
         openid::setup(openid_configs);
     }
+
+    init_account_principal_index_backfill_timer();
```
MRmarioruci, line 854: Could you add a comment so that we remember to remove this afterwards.

src/internet_identity/src/main.rs
```diff
@@ -843,6 +844,85 @@ fn initialize(maybe_arg: Option<InternetIdentityInit>) {
...
+fn account_principal_index_backfill_status() -> (u64, u64, bool) {
+    (
+        ACCOUNT_PRINCIPAL_INDEX_BACKFILL_INDEXED.with_borrow(|indexed| *indexed),
+        ACCOUNT_PRINCIPAL_INDEX_BACKFILL_SKIPPED.with_borrow(|skipped| *skipped),
+        ACCOUNT_PRINCIPAL_INDEX_BACKFILL_DONE.with_borrow(|done| *done),
+    )
+}
+
+fn run_account_principal_index_backfill_batch() {
+    if ACCOUNT_PRINCIPAL_INDEX_BACKFILL_DONE.with_borrow(|done| *done) {
```
MRmarioruci, line 884: Maybe we could put the timer cleanup here as well. If this ever reaches/is triggered then it would mean that the timer ran again?? so making sure it clears again. Or we could just have it here and skip having it in two places

MRmarioruci (approved):

src/internet_identity/src/main.rs
sea-snake: Added in `47abea1bd`, and worth being precise about *when*, because "afterwards" is later than it looks:

```rust
// TEMPORARY: remove once every deployment has upgraded through a build that carries
// this sweep. Not "once it has run" — its completion flag is heap state, so it runs
// again after every upgrade, and a canister that skipped this build entirely would
// never have indexed the accounts it already held.
```

The condition is "every deployment has passed *through* a build containing it", not "it has finished once" — `ACCOUNT_PRINCIPAL_INDEX_BACKFILL_DONE` is a thread-local, so an upgrade forgets it and the sweep walks every list again.

The comment also flags something that has to be dealt with at the same time, and would otherwise be easy to miss: this sweep is now the *only* thing that repairs a drifted index entry. #4238 stopped re-asserting the entries a write did not move — it derives principals for the accounts whose numbers changed and leaves the rest alone — explicitly on the grounds that the sweep comes back around after every upgrade. So deleting the timer deletes that repair, and either something replaces it or the reasoning behind that change has to change with it.

src/internet_identity/src/main.rs
sea-snake: You've spotted a real hole, and your reading is exactly right: reaching that branch means the timer fired *after* completion, so the clear in the `is_done` block didn't take — an id that outlived its slot, or a second timer installed over the first. Returning quietly left it firing every second forever, doing nothing, and treating the evidence of a failed cleanup as a no-op.

Fixed in `47abea1bd` — it repeats the cleanup there.

On "just have it here and skip having it in two places": I kept both, for one reason. Clearing *only* on the early return means every completed sweep necessarily fires one more time before stopping — the `is_done` block would set the flag, the timer would fire again a second later, and only then shut down. Completion should stop the timer where completion happens. So the `is_done` block stays the normal path and the early return is the belt-and-braces for a timer that should no longer exist.

The two do share one description of what stopping means, though, so it isn't duplicated logic:

```rust
fn clear_account_principal_index_backfill_timer() {
    ACCOUNT_PRINCIPAL_INDEX_BACKFILL_TIMER_ID.with_borrow_mut(|id_slot| {
        if let Some(timer_id) = id_slot.take() {
            ic_cdk_timers::clear_timer(timer_id);
        }
    });
}
```

src/internet_identity/src/main.rs
sea-snake: Correcting myself on the second half of that reply — the part about a repair was wrong, and it was my invention rather than something the code justified.

There is no repair. This sweep is a one-time migration to index the accounts that predate the index. Its running again after every upgrade is an accepted *cost* of keeping the completion flag on the heap, not a guarantee anything is entitled to lean on — and it is machinery meant to be deleted, so an argument resting on it rests on something designed to disappear.

What #4238 actually gives up by no longer deriving both lists is redundancy, not repair: re-asserting every entry on every write would have papered over a bug in `sync_account_principal_index`, but it was never fixing a drift that something else causes. Nothing causes one — apply cannot fail, every refusal having happened before the first store, so an index entry is written with its list or neither is. Correctness rests on the sync function, which is where it rests for any index.

So deleting the timer costs nothing beyond the migration itself, and the coupling I claimed between the two PRs does not exist. Both comments now say that: the note here in `66b2d67c0`, and `sync_account_principal_index`'s doc on #4238 in `18f27c791`.

The first half stands: the removal condition is "every deployment has upgraded through a build carrying this", not "it has finished once".

PR #4241, opened 2026-08-19
feat(be): add a session record to the account reference schema

sea-snake:
Design: #4224, on the storage of #4222. Overview: #4230. The schema only — #4266 is what writes a session.

An app delegation is unrevocable for as long as it is valid, which is up to 30 days: the client holds a self-contained canister-signed artifact whose verification never consults the canister again. The fix needs somewhere to put a long-lived, revocable record, and #4222 gives a capped, evictable per-account store to hang it on.

A session is `(created_at, valid_till, last_refreshed, browser_id, read_only)` on the account reference. Nothing else. Putting it there rather than in a map of its own means it inherits the per-anchor caps that already bound references, so revoking, expiring and evicting reuse machinery that exists.

- The field is `Option`, so references written under the previous schema decode unchanged, and an empty list is not stored.
- Only `last_refreshed` is mutable, which is why it is the one field that will not feed the session's seed: a mutable input would change the session's principal every time it was stamped.
- **A list carries no eviction exemption for holding a session.** The list is what makes an app visible in settings, so sparing it would leave the user access they cannot see, and a session nobody can find is a session nobody can revoke. What an eviction costs is a ceremony, not an account: the account is computed, so it returns at the identical principal on the next sign-in.
- `reclaim_order` ranks dead sessions first, then live ones on `last_used + (last_used - created_at)` — recency extended by how long the session stayed in service, so an app in weekly use outranks one opened once yesterday. #4267 is what enforces a cap with it.

`AccountReference::new` replaces the struct literals, so adding the field did not spread across every construction site.

Tests: `session_tests` (11), including a reference written before sessions existed decoding with none, a list holding a session being evictable like any other, and a flood of unused sessions failing to displace a used one.


src/internet_identity/src/storage/storable/session_record.rs
```diff
@@ -0,0 +1,65 @@
...
+    #[n(1)]
+    pub valid_till_ns: StorableTimestamp,
+    #[n(2)]
+    pub max_idle_ns: StorableDuration,
+    #[n(3)]
+    pub last_refreshed_ns: Option<StorableTimestamp>,
+    #[n(4)]
+    pub device_id: StorableSessionDeviceId,
+    #[n(5)]
+    pub read_only: bool,
```
MRmarioruci, line 24: I’m thinking ahead to a session with more specific permissions, for example, an app may be allowed to call one write action, such as orders:create or whatever, but not read orders or make any other changes. That is neither full access nor read-only, so this boolean cannot represent it. If sessions later carry scopes, we would still need to write a made-up read_only value for every new scoped session. Would it be cleaner to make this optional now, so scopes can take over when needed?

sea-snake: Given that read_only is a property of the delegation, enforced by the IC itself, instead of application code, I think those two concerns are isolated from one another.

Likely, scoped permissions will be enforced in application code and thus rely on sender_info attributes.

src/internet_identity/src/storage/account.rs
```diff
@@ -40,6 +40,71 @@ pub struct AccountsCounter {
...
+    ///
+    /// One question rather than two, because a caller has no use for the halves
+    /// apart: a session past its lifetime and one nobody has used for longer than
+    /// it was allowed are equally over. Asking separately is how a caller ends up
+    /// checking one and forgetting the other.
+    ///
+    /// Idleness is measured from the last mint, or from creation where nothing has
+    /// minted yet, so a session abandoned immediately after sign-in is bounded like
+    /// any other.
+    pub fn is_over(&self, now: Timestamp) -> bool {
```
MRmarioruci: could we use a more explicit name here?? is_expired_or_idle or something..

src/internet_identity/src/storage/account.rs
```diff
@@ -40,6 +40,71 @@ pub struct AccountsCounter {
...
+        if self.valid_till_ns <= now {
+            return true;
+        }
+        let last_used = self.last_refreshed_ns.unwrap_or(self.created_at_ns);
+        now.saturating_sub(last_used) >= self.max_idle_ns
+    }
+
+    /// How long this session stayed in service: the span from its creation to the last time
+    /// its app asked for a delegation. Bounded by the session's own lifetime.
+    pub fn demonstrated_use(&self) -> u64 {
```
MRmarioruci:
```suggestion
    pub fn age_at_last_refresh_ns(&self) -> u64 {
```

src/internet_identity/src/storage/account.rs
```diff
@@ -40,6 +40,71 @@ pub struct AccountsCounter {
...
+            .map_or(0, |refreshed| refreshed.saturating_sub(self.created_at_ns))
+    }
+
+    /// What the caps reclaim on, ascending: dead sessions first, then live ones by how
+    /// recently used, extended by how long they stayed in service.
+    ///
+    /// The extension is what separates an app in weekly use from one opened once and
+    /// abandoned, which recency alone gets backwards — the abandoned one was touched more
+    /// recently. `browser_id` only makes the order total.
+    pub fn reclaim_order(&self, now: Timestamp) -> (bool, Timestamp, BrowserId) {
```
MRmarioruci: reclaim_sort_key. and a quick comment "it returns the key used to sort sessions for reclamation" should be fine

MRmarioruci (approved):

src/internet_identity/src/storage/storable/session_record.rs
```diff
@@ -0,0 +1,65 @@
+use crate::storage::account::SessionRecord;
```
MRmarioruci: nit. we could probably re-export these from storable

```
// storable/mod.rs
pub use browser::StorableBrowser;
pub use browser_id::StorableBrowserId;
```
and then
```
use crate::storage::storable::{StorableBrowser, StorableBrowserId};
```

MRmarioruci (approved):

src/internet_identity/src/storage/account.rs
sea-snake: Renamed to `is_expired_or_idle`. The doc comment had to open with "on either bound" to make up for the name, which was the sign — it now just explains why it is one question rather than two.

src/internet_identity/src/storage/account.rs
sea-snake: Renamed, but not to that: the function returns `last_refreshed − created_at`, a span rather than an age, so `age_at_last_refresh_ns` would name something it does not answer. It is `time_in_service_ns` now — your suffix and plainer wording, applied to what it actually returns.

src/internet_identity/src/storage/account.rs
sea-snake: Renamed to `reclaim_sort_key`. Left the doc as it is rather than adding the one-liner: it already says what the key is and why the last field breaks the tie, so a summary line above it would only repeat that.

src/internet_identity/src/storage/storable/session_record.rs
sea-snake: Left as it is. `storable.rs` re-exports nothing today — forty-odd `pub mod` lines and no `pub use` — and the submodules are what namespace these types. Flattening them into one namespace makes every storable type need a globally unique name, and doing it for two of them would split the convention.

PR #4242, opened 2026-08-19
feat(be): register the browser a session was created from

sea-snake:
Design: #4224. Overview: #4230. The registry only — #4264 proves and rotates a browser's key, #4249 renders the list.

Sessions live per account, so a user who wants to sign one browser out has nothing to name it by. A browser registry groups a browser's sessions across apps so they can be revoked together.

- `StorableAnchor` gains `browsers` (field 7) and `next_browser_id` (field 8), both `Option` so existing anchors decode cleanly. Each entry is `{id, description, created_at, last_used, current_browser_key, next_browser_key}`, capped at 20 because the anchor blob is read on nearly every authenticated path.
- `current_browser_key` is the key the browser last proved with and `next_browser_key` the successor it announced then. Either resolves the entry, and presenting the successor promotes it, retiring the key it replaced. That is what makes a browser profile copied off disk stop working: the copy dies as soon as the real browser signs in again.
- Accepting both is what makes a lost response harmless. The browser advances only once a sign-in succeeded, so an unanswered call leaves it proving with the key the entry still holds rather than looking like a new machine.
- A successor another browser of the same anchor already holds is refused (`SuccessorAlreadyInUse`), so one public key belongs to at most one entry and resolving a presented key never depends on list order. #4264 adds the stronger rule, proof of possession of the successor.
- The id never changes across rotations, so rotating costs a session nothing: sessions record the id, not the key. Ids come from a monotonic per-anchor counter and are never reused, so a flood of registrations leaves a permanent gap in the sequence rather than a list that looks untouched.
- At the cap the least recently used entry is dropped rather than the registration failing, and its id is returned so the caller can end that browser's sessions too.
- Eviction orders on `last_used`, not `created_at`. Clearing browser storage loses the key, so every wipe enrols a fresh entry and the wiping browser always holds the newest `created_at` — under enrolment order it would never be its own victim, and twenty wipes would evict twenty browsers the user actually signs in from.
- **`description` is what the browser reported, in tokens rather than a name to show.** `BrowserBrand`, `OperatingSystem` and `FormFactor` are variants, plus the hardware model where a client can name one; each variant carries an `Other : text` for something this list does not name, because an unrecognised browser is worth seeing rather than worth hiding behind a generic label. Tokens because products get renamed — "Chrome OS" became "ChromeOS", "Mac OS X" became "macOS" — so the wording a user reads lives in the frontend (#4249), where changing it reaches every stored record at once. The canister stores these and never interprets them.
- **A description is fixed at registration.** An entry that is advanced keeps what it was created with, so what a browser reports is a fact about a registration rather than about the last sign-in. A browser reporting something else is one this anchor has not seen, and the client is the party that decides so, by presenting a key pair no entry holds (#4271) — which is the path that already registers.
- Entries ride on `identity_info` alongside `mcp_config`, via a new `BrowserInfo` candid type in an `opt` field, backwards compatible in both directions.

Tests: `browser_tests` (19) covering registration and reuse, rotation including a lost response and ten rotations keeping one id, the one-key-one-entry invariant, and the cap dropping the least recently used — including twenty storage wipes interleaved with use of one browser leaving that browser listed. `browser_description` round-trips every token through storage, including the unrecognised ones, so adding a variant on one side without the other fails there rather than mapping to something else. PocketIC asserts an anchor from the previous release decodes with the new field absent.


src/internet_identity/src/storage/storable/session_device.rs
```diff
@@ -0,0 +1,37 @@
...
+    #[n(1)]
+    pub name: String,
+    #[n(2)]
+    pub created_at: Timestamp,
+    #[n(3)]
+    pub last_used: Timestamp,
+    #[cbor(n(4), with = "minicbor::bytes")]
+    pub key: Vec<u8>,
+    #[cbor(n(5), with = "minicbor::bytes")]
+    pub pending: Vec<u8>,
```
MRmarioruci:
```suggestion
    pub next_key: Vec<u8>,
```

src/internet_identity/src/storage/anchor.rs
```diff
@@ -627,6 +714,77 @@ impl From<(AnchorNumber, StorableFixedAnchor, Option<StorableAnchor>)> for Ancho
...
+    /// At the cap the least recently used records are dropped, and their ids returned so
+    /// the caller can end their sessions too.
+    pub fn resolve_session_device(
+        &mut self,
+        key: PublicKey,
+        next_key: PublicKey,
+        name: String,
+        now: Timestamp,
+    ) -> Result<(SessionDeviceId, Vec<SessionDeviceId>), SessionDeviceError> {
+        let holder = |candidate: &PublicKey| {
```
MRmarioruci:
```suggestion
        let device_id_for_key = |candidate: &PublicKey| {
```

MRmarioruci:
```suggestion
let device_index_for_key = |key: &PublicKey| {
    self.session_devices
        .iter()
        .position(|device| device.key == *key || device.next_key == *key)
};

let key_index = device_index_for_key(&key);
let next_key_index = device_index_for_key(&next_key);

if next_key_index.is_some() && next_key_index != key_index {
    return Err(SessionDeviceError::SuccessorAlreadyInUse);
}

if let Some(index) = key_index {
    let device = &mut self.session_devices[index];
    device.key = key;
    device.next_key = next_key;
    device.last_used = now;
    return Ok((device.id, vec![]));
}
```
We look up next_key twice and key once more before updating. It's not a huge deal cause the list is max 20 items but maybe we could do something like this..

src/internet_identity/src/storage/anchor.rs
```diff
@@ -627,6 +714,77 @@ impl From<(AnchorNumber, StorableFixedAnchor, Option<StorableAnchor>)> for Ancho
...
+            self.session_devices
+                .iter()
+                .find(|device| device.key == *candidate || device.pending == *candidate)
+                .map(|device| device.id)
+        };
+        if holder(&next_key).is_some() && holder(&next_key) != holder(&key) {
```
MRmarioruci:
```suggestion
        if key == next_key {
          return Err(SessionDeviceError::SuccessorMatchesCurrent);
        } 
        if holder(&next_key).is_some() && holder(&next_key) != holder(&key) {
```
Something like this would also prevent non-rotated keys. Kind of unlikely but if we ever got into this situation then someone could indefinitely rotate using the same key

src/internet_identity/src/storage/storable/session_device.rs
sea-snake: Renamed, and taken further than the field: the same key went by three names — `device_key` on the wire, `key` in the record, and a bare `key`/`next_key` parameter in between. It is `current_device_key` and `next_device_key` everywhere now, including the candid field, the regenerated TS bindings and the frontend call site. `pending` is gone with it.

The signature fields followed as `current_device_key_signature`. The two signature domain constants keep their byte strings, since those are the wire contract and renaming them would invalidate every signature.

src/internet_identity/src/storage/anchor.rs
sea-snake: Renamed to `device_index_for_key` — it returns a position now rather than an id, per your other suggestion.

sea-snake: Taken. It was slightly worse than one extra lookup: the condition read

```rust
if holder(&next_key).is_some() && holder(&next_key) != holder(&key) {
```

so `holder(&next_key)` ran twice on top of the separate `find` for `key` — four passes to answer two questions. Both indices are bound once now and the update indexes in place:

```rust
let current_index = device_index_for_key(&current_device_key);
let next_index = device_index_for_key(&next_device_key);
if next_index.is_some() && next_index != current_index {
    return Err(SessionDeviceError::SuccessorAlreadyInUse);
}
```

src/internet_identity/src/storage/anchor.rs
sea-snake: Good catch — this was a real hole, not a hypothetical. Fixed as `SessionDeviceError::SuccessorMatchesCurrent`, refused before anything is written.

Two things I found confirming it:

The signature check does not stand in the way. Both required signatures are made by the key being presented, so a caller announcing its own key satisfies them honestly — only the equality check was missing.

And there was a test asserting the behaviour you flagged, `a_browser_that_never_rotates_keeps_working`. I checked the client before touching it: `withBrowserProof` generates a fresh successor on every sign-in, so no legitimate browser ever sends an equal pair, which means that test only ever covered the abuse case. It is now `a_browser_cannot_name_itself_its_own_successor`, plus one for an already-registered browser trying to stop rotating, plus an integration test (`should_refuse_a_successor_equal_to_the_key_presented`) covering the wire path with both signatures valid.

Worth noting for anyone falsifying this later: the integration tests run against the prebuilt wasm, so removing the guard from source and re-running `cargo test` leaves the test passing. It only fails once the wasm is rebuilt.

src/internet_identity/src/storage/anchor.rs
```diff
@@ -38,11 +40,87 @@ pub struct Anchor {
...
+pub struct Browser {
+    pub id: BrowserId,
+    /// The browser's own public key, DER-encoded. What the entry is looked up by.
+    pub current_browser_key: PublicKey,
+    /// The successor the browser announced at its last sign-in, also accepted as a proof.
+    pub next_browser_key: PublicKey,
+    /// What this browser reported about itself when it registered.
+    pub description: BrowserDescription,
+    pub created_at: Timestamp,
+    pub last_used: Timestamp,
```
MRmarioruci, line 95: what if we allowed the user to name his devices? with a custom_name field? This would probably help him in the UI.. but I am not quite sure if someone would care this much to name their devices.. By naming I mean I would put "Mario's work laptop" or something..

src/internet_identity/src/main.rs
```diff
@@ -1098,6 +1098,22 @@ mod v2_api {
             Some(stored_verified_emails)
         };
 
+        let stored_browsers: Vec<BrowserInfo> = state::anchor(identity_number)
+            .browsers()
+            .iter()
+            .map(|browser| BrowserInfo {
```
MRmarioruci: This could also be another method inside the anchor browsers_info(). Would avoid having every consumer create a map and only leak necessary info.

src/internet_identity/src/main.rs
```diff
@@ -1098,6 +1098,22 @@ mod v2_api {
...
+                id: browser.id,
+                description: browser.description.clone(),
+                created_at: browser.created_at,
+                last_used: browser.last_used,
+            })
+            .collect();
+        let browsers = if stored_browsers.is_empty() {
```
MRmarioruci: and maybe that way you could also avoid this check, returns None/Some

src/internet_identity/internet_identity.did
```diff
@@ -1006,6 +1006,47 @@ type IdentityAuthnInfo = record {
...
+// real answer: the browsers exposing no client hints are the ones this is least sure of.
+type FormFactor = variant { Desktop; Mobile; Tablet; Unknown };
+
+// What a browser reported about itself when it registered. Self-reported, so it is
+// something the user reads to recognise their own browser rather than evidence about
+// where a session came from. The canister stores these and never interprets them.
+type BrowserDescription = record {
+    brand : BrowserBrand;
+    os : OperatingSystem;
+    form_factor : FormFactor;
```
MRmarioruci, line 1036: are we sure we are always gonna get this info? if not then make it opt?

MRmarioruci: actually maybe just make the whole record opt?? we don't know if some random browser with an issue reports jiberish here..

src/internet_identity/src/storage/storable/browser.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub created_at: Timestamp,
+    #[n(3)]
+    pub last_used: Timestamp,
+    #[cbor(n(4), with = "minicbor::bytes")]
+    pub current_browser_key: Vec<u8>,
+    #[cbor(n(5), with = "minicbor::bytes")]
+    pub next_browser_key: Vec<u8>,
+}
+
+impl Storable for StorableBrowser {
```
MRmarioruci: Is this Storable implementation needed? StorableBrowser is only encoded as part of StorableAnchor, so its bound is never used or checked directly. Could we remove this implementation and keep only Encode/Decode?

MRmarioruci (approved):

src/internet_identity/src/storage/anchor.rs
sea-snake: Good idea, but not in this stack. The record holds tokens and the wording is derived on read, which is what lets renaming a product rename every row that already exists; a stored custom name is the one string that cannot be re-derived, and it needs UI to set and edit. The derived name, the brand mark and the hardware model where the client reports one already answer "which machine is this", so I have written it down as a follow-up rather than built it.

src/internet_identity/src/main.rs
sea-snake: Agreed, done — `Anchor::browsers_info()`.

You put your finger on where the decision belongs: `Browser` carries `current_browser_key`, `next_browser_key` and `session_count` while `BrowserInfo` exposes only the id, the description and the timestamps, so `main.rs` was deciding what not to leak. That choice now sits next to the type that knows which fields are secret, and every consumer gets the shape rather than the map.

Done now rather than later because #4266 adds `session_count` to `BrowserInfo` for the devices page, so this code was going to be edited again one PR up either way.

src/internet_identity/src/main.rs
sea-snake: Yes — folded into the same method. `browsers_info()` returns `Option<Vec<BrowserInfo>>`, so the `is_empty() -> None` step at the call site is gone with it. `None` rather than an empty list is the shape the interface carries and no caller wants the difference.

src/internet_identity/internet_identity.did
sea-snake: `model` is already `opt text` — Android is the only client that names hardware, so it is absent most of the time.

Keeping `BrowserDescription` itself required, though. The types already absorb gibberish: `BrowserBrand` and `OperatingSystem` both carry `Other(text)` and `FormFactor` has `Unknown`, so a browser this canister does not recognise arrives as `Other("…")` rather than as nothing — and the devices page renders that verbatim on purpose, because a browser the user does not recognise is the row the page exists for. An `opt` record would add a state that cannot occur, since `prepare_account_session` requires one, and every reader would have to handle a `None` that never arrives.

What your instinct points at is real and is already handled a different way: every token is bounded at `MAX_BROWSER_TOKEN_BYTES` (64) and an over-long one is refused. So gibberish is capped, not optional.

src/internet_identity/src/storage/storable/browser.rs
sea-snake: Agreed, removed — `Encode`/`Decode` only.

Verified: `StorableBrowser` appears nowhere outside its own file except as a field of `StorableAnchor`, never as a `StableBTreeMap` or `StableCell` value, so nothing ever consults `BOUND: Bound::Unbounded`. It was a declaration no code reads. The precedent is in the same directory — `StorableBrowserDescription`, the nested storable beside it, has no `Storable` impl at all.

MRmarioruci (approved):

PR #4244, opened 2026-08-19
feat(be): record that a session is still in use

sea-snake:

"This browser used this app 3 minutes ago" against "5 weeks ago" is what makes a session list worth reading, and what lets someone spot a session they do not recognise **still being used** rather than merely still existing.

**Every refresh stamps**, and three consumers make coarsening it unattractive: the session cap orders live sessions on this field, so inside a coarsening interval every session would look equally idle; the browser registry cap orders on the same signal, where an hour is long enough to drop a browser in use; and a user-facing list that can be an hour stale does not answer the question it exists to answer.

**What the stamps cost, stated separately, because they are not the same write.** `last_refreshed_ns` and the reference's `last_used` ride free: the list blob is rewritten either way, so they change bytes already in flight. The browser registry's `last_used` does not — `stamp_browser_use` is a `read` plus a `write` of the whole anchor, which serialises every passkey, OpenID credential, recovery key and verified email, pre-checks the email-recovery binding against its reverse index, and re-syncs five reverse indices by diffing vectors, all to move one `u64`. So this PR adds **one anchor read-modify-write per five minutes of app use, per live session**, where before there was one per sign-in. That is accepted here rather than argued away: the registry cap orders on this field, and a browser in use dropping out of the list is worse than the write. The coarsening argument above is about `last_refreshed_ns` and the user-facing list, and does not apply to the device field.

The same write carries three stamps:

| Field | Lives on | Drives |
| ----- | -------- | ------ |
| `last_used` | the account reference | account eviction (#4235) |
| `last_refreshed` | the session record | the session cap (#4267) and the user-facing list |
| `last_used` | the device record | the registry cap (#4242) and the settings list |

Stamping the device is the only reason refresh touches the anchor: it authenticates by session chain and never runs `check_authorization`. What that write buys is a use signal a sign-in stamp cannot give the browser list.

Stable writes scale with `1/T` alongside the calls, so lowering the app-delegation TTL multiplies both.

Tests: `session_refresh_stamp_tests` (8), including a stamp for a session that is gone being refused rather than reporting success, and a refresh advancing the device's `last_used` while leaving its enrolment timestamp alone. PocketIC drives two refreshes a minute apart and reads the device back off `identity_info`.


src/internet_identity/src/sessions.rs
```diff
@@ -315,7 +316,19 @@ pub fn app_prepare_delegation(
     request: AppPrepareDelegationRequest,
 ) -> Result<AppPrepareDelegationResponse, AppSessionError> {
     let now = time();
-    let (account, session) = authorize_session(now)?;
+    let (locator, account, session) = authorize_session(now)?;
+
+    storage_borrow_mut(|storage| {
```
MRmarioruci: Technically this can return Ok(false) right, although highly unlikely. In case it does we don't handle that though and we still mint the delegation
```suggestion
    let stamped = storage_borrow_mut(|storage| {...
    if !stamped {
        return Err(AppSessionError::NoMatchingSession);
    }
```

src/internet_identity/src/storage/anchor.rs
```diff
@@ -731,6 +731,22 @@ impl Anchor {
         &self.session_devices
     }
 
+    /// Advances a device's `last_used`. Reports whether anything changed, so an unknown
+    /// device or a repeat inside one message costs no anchor write.
+    pub fn stamp_session_device_use(&mut self, device_id: SessionDeviceId, now: Timestamp) -> bool {
```
MRmarioruci: If i understand correctly, false means:
1. The device was not found
2. It's timestamp did not need updating

Maybe it's a bit cleaner if we return an enum
`enum DeviceStampResult {
    Updated,
    Unchanged,
    DeviceNotFound,
}`

src/internet_identity/src/sessions.rs
MRmarioruci, line 337: account_seed can fail and we did the stamping above..
```suggestion
let (locator, account, session) = authorize_session(now)?;
let seed = account_seed(&account)?;
let access = DelegationAccess::from_read_only(session.read_only);
let expiration = now
    .saturating_add(APP_DELEGATION_TTL_NS)
    .min(session.valid_till_ns);

// Then write the stamps.
```
Maybe this is better ordering.

src/internet_identity/src/storage.rs
```diff
@@ -2043,6 +2043,86 @@ impl<M: Memory + Clone> Storage<M> {
         Ok(remaining)
     }
 
+    /// Records that a session was used. Reports whether a session matched. The
+    /// reference's `last_used` rides on the same write.
+    pub fn stamp_session_refresh(
```
MRmarioruci: Could this take a session locator or handle instead of five separate identifier??

MRmarioruci: maybe `struct MatchedSession {
    locator: AccountLocator,
    account: Account,
    session: SessionRecord,
}`

```suggestion
    pub fn record_session_activity(
```

src/internet_identity/src/storage.rs
```diff
@@ -2043,6 +2043,86 @@ impl<M: Memory + Clone> Storage<M> {
...
+                anchor_number,
+                application_number,
+                *account_number,
+                std::slice::from_ref(session),
+            );
+        }
+        if !expired.is_empty() {
+            self.change_session_count(anchor_number, expired.len(), 0)?;
+        }
+        self.stamp_session_device_use(anchor_number, device_id, now)?;
```
MRmarioruci: The anchor stores both the session count and the browser’s last_used. When expired sessions are removed, we save the anchor once for the count and then load and save it again for the timestamp. Could we update both fields before writing it once?

src/internet_identity/src/sessions.rs
```diff
@@ -324,7 +324,10 @@ pub fn app_prepare_delegation(
     request: AppPrepareDelegationRequest,
 ) -> Result<AppPrepareDelegationResponse, AppSessionError> {
     let now = time();
-    let (account, session) = authorize_session(now)?;
+    let (key, account, session) = authorize_session(now)?;
+
+    storage_borrow_mut(|storage| storage.record_session_use(&key, now))
+        .map_err(|err| AppSessionError::InternalCanisterError(err.to_string()))?;
```
MRmarioruci: Maybe we could be more explicit about errors here?

src/internet_identity/src/sessions.rs
sea-snake: You are right, and it is worse than unlikely — `?` catches only the `Err`, so `Ok(false)` sailed past and an app delegation was minted for a session no list holds, which is the one outcome this design exists to prevent.

The fix is not a check at the call site, though: nobody makes a decision from that bool, so it should not be one. `record_session_use` returns `Result<(), StorageError>` with a `SessionNotFound` variant, and the case stops being ignorable because `?` handles it. Its three `Ok(false)` returns all meant "what you named is not there" — no application for the origin, no matching account reference, no matching session — and its doc justified them as "not a failure" for a revocation race no caller acts on.

`revoke_session` on #4245 got the same treatment, which closes the `.expect` hole you filed there.

src/internet_identity/src/storage/anchor.rs
sea-snake: The bool is already gone — it returns `()` — so there is no three-way outcome left to name and no enum.

Your comment did catch something, though: the doc still claimed the return value. It said "Reports whether anything changed, so an unknown device or a repeat inside one message costs no anchor write", which promises a value the function does not have and a saving the caller cannot make from a value it never receives. Rewritten for a function that returns nothing, and it says browser rather than device now.

src/internet_identity/src/sessions.rs
sea-snake: Agreed, reordered — `account_seed`, `access` and `expiration` all move above the stamp.

Your reasoning is the important part and it is now in the code: returning `Err` on the IC commits whatever was written before it, since only a trap rolls back, so a stamp above a refusal leaves the session recorded as used while the caller is told the call failed. All three depend on values already in hand, so moving them costs nothing.

The window is narrow — an unset salt means a canister that never initialised — so this is about the shape rather than a live bug. It is the principle the rest of the stack states outright: everything that can refuse does so before any of it is stored.

src/internet_identity/src/storage.rs
sea-snake: It does now — `record_session_use(&SessionLocator, now)`. It was `stamp_session_refresh` with the identifiers spread out when you read this.

The second half of what you asked for landed too, on the return side: `authorize_session` hands back an `AuthorizedSession { locator, account, session }` rather than a tuple. Named after the guarantee rather than the lookup, because one can only be held by having passed the principal lookup and the liveness check — so the type *is* the evidence rather than a bundle of three values.

sea-snake: Your `record_*` shape landed — it is `record_session_use` — and it stays that. Only the noun differs, deliberately: "activity" is taken in this canister for access-method activity and the active-anchor metrics (`activity_bookkeeping`, `update_activity_stats`), so `record_session_activity` would put one word on two unrelated mechanisms.

The doc comment you quoted was stale for a second reason too: "Reports whether a session matched" went when the bool became an error.

src/internet_identity/src/storage.rs
sea-snake: Already one write, and it says so in place.

You were reading `change_session_count(...)` followed by `stamp_session_device_use(...)`, each loading and writing the anchor. `record_session_use` now reads the anchor once, calls `anchor.stamp_browser_use(browser_id, now)` on the in-memory value and hands it to `write_account_state`, which folds the session count into the same write. The comment there is: "Stamped before the write rather than after, because the write is what stores the record. There is no second store."

That fell out of moving the counters into the write gate rather than from this thread, but it is the change you asked for.

src/internet_identity/src/sessions.rs
sea-snake: Agreed, and there is now something specific to be explicit about.

When you wrote this the stamp returned `Ok(false)` for "not there", so every error out of it really was internal. `record_session_use` returns `Result<(), StorageError>` now, per your other thread on this line, and one of its errors is a session that is not there — which at this call site means one revoked between `authorize_session` and the stamp. That is a race, not an internal fault, so it maps to `NoSuchSession`: the same answer the caller would have got a moment earlier. Everything else stays `InternalCanisterError`.

MRmarioruci (approved):

sea-snake:
Both suppressed findings from the Copilot review, answered — they land differently.

**`sessions.rs:383` — the missing PocketIC coverage.** It exists, one PR up. `should_advance_the_device_last_used_on_every_refresh` at #4245 asserts `created_at == last_used` at enrolment, advances 300 seconds, calls `app_prepare_delegation`, and reads the browser back through `identity_info` — exactly the boundary named, wiring included. This stack merges whole, so no change here.

**`storage/tests.rs:6585` — the test that doesn't test what it says.** Correct, and the test is now deleted rather than repaired. `storage_with_session()` calls `create_session`, which registers a browser through `resolve_browser`, so the test was a duplicate of the registered-browser refresh above it under a name claiming the opposite.

Constructing the orphan, as suggested, would pin a state that cannot occur. A browser is given up only at `MAX_BROWSERS` inside `resolve_browser`; the write gate sweeps that browser's sessions in the same write; and `sync_session_index` removes their principal-index entries along with them. So when the browser goes the session goes, and a later mint stops at `NoSuchSession` in `find_caller_session` before `record_session_use` is reached at all.

That also settles `stamp_browser_use`'s browser-not-found branch, which we looked at while here: it stays a no-op. It is unreachable for the reason above, and nothing observable would go wrong if it were reached — `browsers_info()` builds the devices list from registry entries, so a missing entry means no row rather than a stale one. The same `if` also covers the ordinary case of a repeat stamp inside one message, which must no-op.

PR #4245, opened 2026-08-19
feat(be): let an app sign its own session out

sea-snake:

Signing out of an app cleared browser state and invalidated nothing already issued. `app_revoke_session : () -> (variant { Ok; Err : AppSessionError })` removes the session record, so the app's access ends within one delegation lifetime instead of running to the original expiry.

- **No authorization check beyond the match** refresh already performs, because a caller cannot produce another session's principal and can only ever remove its own.
- **A session that is already gone is `Ok`**, so sign-out is idempotent: a client that retries, or signs out twice, gets the same answer without reasoning about whether its session was still there. Only a storage failure comes back as an error — the same error type its two siblings already return, so this is consistency with its own family rather than a new convention.
- It finds the session without the expiry check, so signing out of an already-expired session still clears the record.

The app deliberately cannot revoke anything else. "Sign out everywhere" is the II frontend's operation (#4246), not something a dapp can trigger.

Revocation latency is exactly the app-delegation TTL, by construction: revoking stops new mints and one already issued runs out. `mcp.rs` documents the same residue for its grants.

Tests: `session_removal_tests` (3) plus PocketIC coverage that a refresh working before sign-out returns `NoMatchingSession` after, three sign-outs in a list all succeed, an expired session is still signed out and its record and index entry go with it, and signing one browser out leaves another browser's session on the same account working.


src/internet_identity/src/sessions.rs
```diff
@@ -403,6 +403,37 @@ pub fn app_get_delegation(
...
+    storage_borrow_mut(|storage| {
+        storage.remove_session(
+            locator.anchor_number,
+            locator.application_number,
+            locator.account_number,
+            session.created_at_ns,
+            session.device_id,
+        )
+    })
+    .expect("failed to remove a session that was just matched");
```
MRmarioruci: Should we check the bool as well? expect only catches an Err, so Ok(false) still reports a successful sign-out. Since we just matched this session, false would mean it was left active.

src/internet_identity/src/storage.rs
```diff
@@ -1938,6 +1938,44 @@ impl<M: Memory + Clone> Storage<M> {
...
+                                    && session.created_at_ns == created_at
+                            })
+                    })
+            })
+            .unwrap_or(false);
+        if !present {
+            return Ok(false);
+        }
+
+        let dropped =
```
MRmarioruci: Could we remove the exact (device_id, created_at) match in one pass here? Right now we read the row once to check that exact session exists, then read it again and remove every session for the device. The one-session-per-browser rule makes that work today, but an inconsistent row could remove more than the caller matched, and the second read seems unnecessary.

src/internet_identity/tests/integration/sessions.rs
```diff
@@ -608,6 +609,76 @@ fn should_advance_the_device_last_used_on_every_refresh() -> Result<(), RejectRe
...
+    assert!(refresh(&env).is_ok());
+
+    app_revoke_session(&env, canister_id, session_principal)?;
+
+    assert_eq!(refresh(&env), Err(AppSessionError::NoMatchingSession));
+
+    Ok(())
+}
+
+#[test]
```
MRmarioruci, line 652: Could we add a test that lets the session expire before signing it out? This intentionally uses match_session instead of authorize_session so expired sessions still get cleaned up, but that behavior isn’t covered yet.

src/internet_identity/src/sessions.rs
```diff
@@ -403,6 +403,37 @@ pub fn app_get_delegation(
...
+}
+
+fn match_session() -> Result<(StorableAccountLocator, Account, SessionRecord), AppSessionError> {
```
MRmarioruci: Nit. the name of this function is a bit ambiguous.. or just add a small comment on top. Find caller matching session or something..

src/internet_identity/src/sessions.rs
sea-snake: Right on both counts, and the endpoint changed shape rather than gaining a bool check.

`app_revoke_session : () -> ()` was the odd one of its three siblings — `app_prepare_delegation` and `app_get_delegation` both return `variant { Ok; Err : AppSessionError }`, and that error already carries `NoMatchingSession` and `InternalCanisterError`, the two things a revoke can hit. So it returns that now: consistency with its own family, no new type. With an `Err` channel the trap has no job left.

One deliberate asymmetry: **sign-out is idempotent**, so a session that is not there is `Ok`, not `Err`. The caller wanted it gone and it is gone, and they cannot tell a pruned session from one that never existed — so only a storage failure comes back as an error. An app is never handed a refusal whose only sane response is to do nothing.

The `Ok(false)` hole itself is closed at the source: `revoke_session` returns `Result<(), StorageError>` now, per your thread on #4244.

src/internet_identity/src/storage.rs
sea-snake: The dangerous half was already fixed, and the rest is done now.

When you read this, the removal took *every* session for the device after matching on `(device_id, created_at)`, so a row holding two would lose both. It removes by `session_id` now, taking exactly what was matched — the id this stack added is what closed that.

The double walk you are pointing at existed only to produce `Ok(false)`, and with that becoming `SessionNotFound` it collapses: one `retain` removes the session and its own length change says whether there was anything to remove.

src/internet_identity/tests/integration/sessions.rs
sea-snake: Agreed, added — and you were right that it was uncovered. `should_refuse_a_refresh_once_the_session_has_expired` advances past expiry and checks a refresh is refused, but nothing signed an expired session out, so the one reason the two lookups differ was held in place by a comment: someone could have swapped it for `authorize_session` and CI would have stayed green.

`should_sign_out_a_session_that_has_already_expired` now lets the session expire, signs it out, and checks the browser's `session_count` falls to zero — so it asserts the record and index entry are gone, not merely unusable. Verified by making that swap: the test fails.

It also pins the narrower reason the sign-out path uses `find_caller_session` rather than `authorize_session`, which the comment there now gives: not that an absent session needs revoking, but that a session that is present and expired must not be refused.

src/internet_identity/src/sessions.rs
sea-snake: Agreed — `find_caller_session`, and one line of doc kept.

"Match" carried nothing; "find" is close to your own wording and pairs with the sibling — one finds a session, the other authorizes one — which now reads `authorize_session() -> AuthorizedSession` against `find_caller_session() -> (SessionLocator, Account, Session)`, the return types saying what the names say.

The doc line stays because no name carries *why* the non-authorizing one exists: an expired session is still the caller's to sign out. There is a test for that now too, from your other thread here.

MRmarioruci (approved):

src/internet_identity/src/storage.rs
```diff
@@ -3398,6 +3398,51 @@ impl<M: Memory + Clone> Storage<M> {
...
+        }
+        let anchor = self.read(key.anchor_number)?;
+        let (mut account_references, config) =
+            self.account_state_for_origin(key.anchor_number, &key.origin);
+
+        let write = account_references
+            .iter_mut()
+            .find(|write| write.account_reference.account_number == key.account_number)
+            .ok_or_else(not_found)?;
+        let sessions = &mut write.account_reference.sessions;
```
nmattia, line 3431: nit: would keep referring to `write.account_reference.sessions`. It's a bit more verbose but makes it clear that the `account_reference` is updated in-place in eg `sessions.retain(...)`.

PR #4246, opened 2026-08-19
feat(be): sign a browser out from the user's own settings

sea-snake:

An app can sign out only its own session (#4245), so "sign this browser out" has to come from somewhere the identity itself is authenticated. That is this PR: `revoke_browser_sessions`.

The two revocation surfaces are split by what each caller can prove:

| Caller | Authenticated as | May revoke | Names a session by |
| ------ | ---------------- | ---------- | ------------------ |
| The app | its own session chain | only its own | nothing; `caller()` is looked up in the session index |
| The II frontend | an identity access method | every session of one browser | `browser_id` |

## The user revokes a browser, not a session

A per-session revoke from settings was specified and is deliberately **not** here.

It would have named a session by `(origin, account, created_at)` — the creation time alone, because the request carries no `browser_id`. But authenticating a session requires the browser *and* the creation time together, precisely because a browser keeps its id across sign-ins. Revoking on the weaker half means two browsers that signed in during the same round are both signed out by a request that named one of them.

The narrower fix would be to add `browser_id` to the request. The better answer is that the II frontend is device-scoped: what a user picks in a settings list is a browser, and `revoke_browser_sessions` already ends everything it holds. Per-session sign-out is the app's own call, where the caller *is* the session and cannot name anyone else's.

## `revoke_browser_sessions`

Takes `(identity_number, browser_id)` and sweeps that identity's references **in one message**, writing only lists that hold that device's sessions.

- **All at once or not at all.** The sweep writes one list per affected application and adjusts the anchor's count after the loop, and on the IC an `Err` reply commits everything written before it — so reporting a storage failure would leave a browser signed out of some of its applications and not others. It traps instead, as `app_revoke_session` does for the same class of failure, and the message rolls back. `SessionRevokeError` therefore has one variant: the caller is either unauthorized, or the call succeeds, or the call is rejected.
- It **resolves** sessions by locator rather than through the principal index, but it still **removes the index entry of every session it drops** — that is what stops a revoked session's principal from staying resolvable.
- Sweeping eagerly is what keeps refresh cheap. Marking a device revoked and checking it during refresh would make revocation O(1) but add an anchor read to a call that otherwise never touches the anchor, and refresh happens every few minutes per active session.
- **The device record survives the sweep**, so a browser that has been signed out is still one the user recognises, and signing back in from it reuses the same id. Deleting a device record is a separate operation, not specified here.

There is no session listing method, deliberately. A flat "every session of this identity" call returns a list bounded only by the caps and mixes every origin together. The right decomposition is applications first, then sessions within one, and that wants designing alongside whatever lists applications. Neither exists yet.

## Tests

`session_revocation_tests` (3). PocketIC covers a whole browser signed out across two apps while a second browser keeps working **and the device record stays listed**, and a revoked session staying gone when the same browser signs in again.


src/internet_identity/src/sessions.rs
```diff
@@ -466,3 +467,35 @@ fn account_seed(account: &Account) -> Result<Hash, AppSessionError> {
     })?;
     Ok(account.calculate_seed_with_salt(&salt))
 }
+
+pub fn revoke_account_session(
+    request: RevokeAccountSessionRequest,
+) -> Result<(), SessionRevokeError> {
+    check_authorization(request.identity_number)
+        .map_err(|err| SessionRevokeError::Unauthorized(err.principal))?;
+    check_frontend_length(&request.origin);
```
MRmarioruci: I think this traps? we should probably avoid that.

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -846,3 +846,25 @@ pub enum AppSessionError {
...
+/// Signs one browser out of every app it is signed into.
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub struct RevokeDeviceSessionsRequest {
+    pub identity_number: IdentityNumber,
+    pub device_id: SessionDeviceId,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub enum SessionRevokeError {
+    Unauthorized(Principal),
```
MRmarioruci, line 879: Another case would be session doesn't exist?

src/internet_identity/src/sessions.rs
```diff
@@ -466,3 +467,35 @@ fn account_seed(account: &Account) -> Result<Hash, AppSessionError> {
...
+
+    storage_borrow_mut(|storage| {
+        storage.revoke_account_sessions(
+            request.identity_number,
+            &request.origin,
+            request.account_number,
+            request.created_at,
```
MRmarioruci: what happens if we include device_id in here as well? I don't love timestamps for uniqueness.

src/internet_identity/src/sessions.rs
MRmarioruci: I think there was something similar in prior PRs, where you have two browsers firing on the same round and this would revoke both sessions

src/internet_identity/src/sessions.rs
```diff
@@ -448,6 +449,20 @@ fn account_seed(account: &Account) -> Result<Hash, AppSessionError> {
     Ok(account.calculate_seed_with_salt(&salt))
 }
 
+pub fn revoke_browser_sessions(
+    request: RevokeBrowserSessionsRequest,
+    now: Timestamp,
+) -> Result<(), SessionRevokeError> {
+    check_authorization(request.identity_number)
+        .map_err(|err| SessionRevokeError::Unauthorized(err.principal))?;
```
MRmarioruci: Should this record activity for the access method that authorized the revoke?

src/internet_identity/tests/integration/sessions.rs
```diff
@@ -646,6 +647,106 @@ fn should_leave_another_browsers_session_alone() -> Result<(), RejectResponse> {
     Ok(())
 }
 
+#[test]
+fn should_sign_a_whole_browser_out() -> Result<(), RejectResponse> {
```
MRmarioruci, line 808: Could we add a test that calls this as another principal and checks that the session still works? Authorization is the main protection on this endpoint, but the integration test only covers the successful call.

src/internet_identity/src/sessions.rs
sea-snake: It did, and it does not any more.

The trap was `check_frontend_length(&request.origin)` inside a function returning `Result<(), SessionRevokeError>` — an error type that already carries `InternalCanisterError`, so there was an answer available and it was rejecting the message instead. It went two ways: `check_frontend_length` now has a sibling that returns an error rather than trapping, used wherever the response already carries a variant for the unreachable edges (see the #4268 thread on this); and this endpoint no longer takes an origin at all, since it revokes by `browser_id`, so the call is gone with the parameter.

src/internet_identity_interface/src/internet_identity/types.rs
sea-snake: Keeping the two variants. Revoking something already absent is success: the caller wanted it gone and it is gone, and they cannot tell a pruned session from one that never existed — so a third variant would hand the frontend a failure whose only sane response is to do nothing. Same reasoning as `app_revoke_session` on #4245, which swallows not-found into `Ok`.

There is a wrinkle particular to this endpoint, since it takes a `browser_id`: "does not exist" can also mean an id the identity has never held, which is a caller error rather than a race — and is today indistinguishable from a browser holding no sessions. Still not worth a variant: the devices page only ever sends ids it just read from `identity_info`.

src/internet_identity/src/sessions.rs
sea-snake: Both halves are resolved, and your instinct was right each time.

You were reading `revoke_account_sessions(identity_number, origin, account_number, created_at)`, keyed on a timestamp. The endpoint is now `revoke_browser_sessions(identity_number, browser_id, now)`: it names the browser rather than a moment, and individual sessions are identified by `session_id`, allocated per session and unique by construction.

This is the fourth thread in this stack tracing to the same root — a `created_at` timestamp standing in for identity, which is not unique within a round. One change answered all of them.

src/internet_identity/src/sessions.rs
```diff
@@ -427,3 +428,16 @@ fn account_seed(account: &Account) -> Result<Hash, AppSessionError> {
...
+    request: RevokeBrowserSessionsRequest,
+) -> Result<(), SessionRevokeError> {
+    check_authorization(request.identity_number)
+        .map_err(|err| SessionRevokeError::Unauthorized(err.principal))?;
+
+    storage_borrow_mut(|storage| {
+        storage.revoke_browser_sessions(request.identity_number, request.browser_id)
+    })
+    .map(|_| ())
+    .map_err(|err| SessionRevokeError::InternalCanisterError(err.to_string()))
```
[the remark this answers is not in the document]
sea-snake: The state behind this is real and worth naming, but the remedy is a log rather than a refusal.

`account_state` walks every reference list keyed by `(anchor_number, application_number)` and resolves each to an origin, because the write gate is keyed by origin. Where `stable_application_memory.get(&application_number)` misses, the `?` inside `filter_map` made the element vanish with no error and no log — so a browser-wide sweep could report success having skipped a list. A reference list for an application that does not exist is illegal, and this was where it disappeared quietly.

Refusing there would be worse than the hole, though: it would block signing out *every* browser for that identity, including the ones whose lists do resolve and whose sessions do mint. And the rows it would protect are inert — the mint path fails on the same lookup (`lookup_session_with_principal` → `lookup_account_with_principal` → `account_key_of` resolves the same application number), so sessions in an unresolvable list are unusable as well as unrevocable. There is no window in which those sessions authenticate anything.

So the sweep keeps its shape and the invariant violation is now logged in the `"ERROR: … invariant violated: …"` form this canister already uses. The change is on #4266, where `account_state` lives.

src/internet_identity/src/sessions.rs
sea-snake: Yes, and it does now — `check_authz_and_record_activity` instead of `check_authorization`.

What that records is about the access method rather than the request: `last_used` on the passkey, OpenID credential or email credential, which is what the access page shows, plus the active-anchor aggregation. Signing every browser out is an authenticated update by the identity's own holder, so the stamp is true and the access page would otherwise miss it. Twelve other endpoints in `main.rs` already use the recording form, and this path writes the anchor anyway, so it costs nothing extra.

src/internet_identity/tests/integration/sessions.rs
sea-snake: Agreed, added — and you are right that authorization is the whole protection here. `app_revoke_session` rests on a caller being unable to produce another session's principal; this endpoint takes `identity_number` as an argument, so nothing but the auth check stands between a stranger and signing every browser of any identity out. Nothing pinned it: the only `Unauthorized` assertion in the file was for `prepare_account_session`.

`should_refuse_to_sign_a_browser_out_for_another_principal` calls it as `principal_2()`, asserts `Unauthorized`, and then checks the session it aimed at still mints. Verified by dropping the check — the test fails.

MRmarioruci (approved):

src/internet_identity/src/sessions.rs
```diff
@@ -524,6 +525,28 @@ fn account_seed(account: &Account) -> Result<Hash, AppSessionError> {
...
+        IdentityUpdateError::StorageError(_, storage_error) => {
+            SessionRevokeError::InternalCanisterError(storage_error.to_string())
+        }
+    })?;
+
+    storage_borrow_mut(|storage| {
+        storage.revoke_browser_sessions(request.identity_number, request.browser_id, now)
+    })
+    .map(|_| ())
+    .map_err(|err| SessionRevokeError::InternalCanisterError(err.to_string()))
```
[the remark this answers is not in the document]
sea-snake: The mechanics are right and worth stating plainly: this endpoint now uses `check_authz_and_record_activity`, which writes the anchor before the sweep, and returning `Err` commits that write because only a trap rolls back. That prior write is new — it was a non-writing authorization check until this round.

Keeping the error rather than trapping, for three reasons.

**What commits is true.** The stamp records that the access method authenticated, which it did, independently of whether the sweep then succeeded. That is the same reasoning we settled on for `prepare_account_session`, where the activity stamp deliberately precedes validation: `last_used` means "this credential authenticated", not "the request it authenticated also succeeded".

**The all-or-nothing property is already held.** `storage.revoke_browser_sessions` is a single write by construction — its own comment says an `Err` cannot sign the browser out of some applications and report failure. So there is no partial revocation for a trap to protect against.

**Trapping would reverse two deliberate changes.** A trap was removed from this very function earlier in this round in favour of a typed error, on the grounds that a rejected message is not something a caller can read as a response. And dropping `InternalCanisterError`, as you also suggest, would leave a storage failure with no way to be reported at all — the variant is there so the frontend can tell "not authorized" from "something broke".

src/internet_identity/internet_identity.did
```diff
@@ -2028,6 +2041,8 @@ service : (opt InternetIdentityInit) -> {
     // its session was still there. An app can revoke only its own session.
     app_revoke_session : () -> (variant { Ok; Err : AppSessionError });
 
+    revoke_browser_sessions : (RevokeBrowserSessionsRequest) -> (variant { Ok; Err : SessionRevokeError });
```
nmattia, line 2044: nit: instead of making this part of the interface from day one, would it make sense to let the frontend do a little more work to start with (list the sessions & revoke them individually) to afford some more flexibility later on?

PR #4247, opened 2026-08-19
feat(fe): hand apps a session to re-issue their own delegations from

sea-snake:

Requires the canister API from #4243, which the stack order enforces.

# Motivation

The canister half of sessions is in place; this is how an app gets one.

# Changes

**A new JSON-RPC method, `ii_session_delegation`.** Namespaced rather than folded into `icrc34_delegation`, because an app that asks for a session but has not implemented refresh must not silently start receiving 5-minute delegations. With a session-only response, an app that cannot refresh never calls this, and `icrc34_delegation` keeps behaving exactly as today.

**The chain, and nothing else.** The canister signs the session identity to a non-extractable key this origin keeps in IndexedDB, keyed by `(identity, account, origin)`, and the frontend extends that chain to a key the app supplies. The app's hop carries `targets` restricted to the II canister: a developer guardrail rather than a defence, since a thief holding the chain can refresh, but it makes an app reaching for the session chain where it meant its app delegation fail immediately.

The result carries the chain alone. Nothing is attached to the calls an app then makes — the canister resolves the session from the signature on the call — so an app needs no agent extension of any kind.

**The browser key, and its rotation.** Sign-in presents a long-lived P-256 key kept in IndexedDB, one per identity so nothing stored links two of the user's identities to the same browser. Each sign-in generates a successor, signs with **both** keys, and advances to the successor once the canister accepts. So a key proves one sign-in and is then retired, and a copied profile cannot keep signing in unnoticed.

Three rules make that safe on the sign-in path:

- Advance only on success.
  A response that never arrives leaves both sides on the key the entry still holds.
- Store the key before the first call.
  The canister may have registered it, so returning with a different key would enrol this browser twice.
- Serialise with a web lock.
  Two sign-ins at once would leave whichever wrote last holding a key the canister never accepted.

**The consent duration is honoured, and an SSO cap still binds it.** An SSO organization caps how long its sign-ins stay valid, so a session sends a duration even when the user picked none, matching what the ICRC-34 path already does. Neither clamp has a unit test — the store this path needs is not stubbed in that file.

**The consent duration is honoured.** `valid_for` carries the lifetime the user picked at the consent screen, clamped by the canister, so a user choosing one hour gets an hour rather than the default thirty days.

**The name.** Derived from the user agent in its own module, ordered most-specific-token-first because every later token appears inside the earlier ones' agents. It names the device where a device word exists — `Safari on iPhone`, `Chrome on Mac`, `Chrome on Chromebook` — and uses the platform's device model where there is one, so an Android entry reads `Chrome on Pixel 5`. The label is coarse by construction, which is why the settings list identifies an entry by its id.

# Tests

`browser-key.store.test.ts` (14): both signatures verify under the domains the canister checks and neither verifies in the other's role; the announced successor is not the key in use; accepting a sign-in rotates to it and not accepting keeps the current key; each identity holds its own; clearing storage yields a key unrelated to either old one; concurrent sign-ins serialise; a browser without the lock API still signs in; the encoding is the 91-byte SPKI and 64-byte signatures the canister parses; and the device id is remembered on acceptance.

`describeBrowser.test.ts` (27): one per row of both ladders over real user-agent strings, the iPad and Mac pair separated only by `maxTouchPoints`, and the model path — used when present, ignored when empty or when the platform refuses, absent without the API, and dropped when it would exceed the canister's 128-byte limit.

`app-session.store.test.ts` (7) and `sessionDelegation.test.ts` (6) cover storage keyed by identity, account and origin, expiring sessions not served, a held session re-issued without a ceremony, and the chain carrying `targets` restricted to the II canister.

### DIFF
--- a/src/archive/archive.did
+++ b/src/archive/archive.did
@@ -71,6 +71,11 @@ type Operation = variant {
     add_name;
     update_name;
     remove_name;
+    // Registering the browser a session was created from. Once per browser per
+    // anchor; the self-reported name is redacted like an account name.
+    register_session_device : record {
+        name : Private;
+    };
     create_account : record {
         name : Private;
     };
--- a/src/canister_tests/src/api/archive.rs
+++ b/src/canister_tests/src/api/archive.rs
@@ -134,7 +134,8 @@ pub mod compat {
                 | Operation::AddEmailRecovery
                 | Operation::RemoveEmailRecovery
                 | Operation::AddVerifiedEmail
-                | Operation::RemoveVerifiedEmail => {
+                | Operation::RemoveVerifiedEmail
+                | Operation::RegisterSessionDevice { .. } => {
                     panic!("not available in compat type")
                 }
                 Operation::CreateAccount { name } => CompatOperation::CreateAccount { name },
--- a/src/canister_tests/src/api/internet_identity/api_v2.rs
+++ b/src/canister_tests/src/api/internet_identity/api_v2.rs
@@ -832,3 +832,37 @@ pub fn app_revoke_session(
     )
     .map(|_| ())
 }
+
+pub fn revoke_account_session(
+    env: &PocketIc,
+    canister_id: CanisterId,
+    sender: Principal,
+    request: RevokeAccountSessionRequest,
+) -> Result<Result<(), SessionRevokeError>, RejectResponse> {
+    call_candid_as(
+        env,
+        canister_id,
+        RawEffectivePrincipal::None,
+        sender,
+        "revoke_account_session",
+        (request,),
+    )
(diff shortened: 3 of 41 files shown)


sea-snake:
Superseded by a five-way split of this PR, each layer under 500 lines of implementation:

- #4269 — one queue for authorization-bearing requests
- #4270 — name the browser a session was created from
- #4271 — hold a browser key and rotate it on every sign-in
- #4272 — keep an app's session across a page load
- #4273 — hand apps a session to re-issue their own delegations from

The combined content is unchanged; the split was verified lossless against this branch's tree.

PR #4248, opened 2026-08-19
feat(be,fe): answer a silent re-auth without rendering anything

sea-snake:
Design: #4228, on the sessions of #4224. Overview: #4230. Sits on the frontend stack at #4249.

Sibling subdomains of one domain can share a sign-in: sign in on `chat.example.com` and `hr.example.com` is signed in too, sign out of one and the others follow. The client half is already specified in `@icp-sdk/auth` — a shared `derivationOrigin`, a domain-scoped cookie holding only a principal and an expiry, and a `/reauth` page running `prompt: 'none'` with a `hint`. This is only what II has to supply for that to work.

Re-issuing is a canister-checked mint like any other rather than an offline extension of a stored chain, so what the app receives stays revocable. The mint is unchanged; this adds one query, `check_session`, so the frontend can tell a record it still holds from a session the canister has already dropped.

**Three authorize-URL parameters.** `prompt` (`none` | `login`) and `hint` (a principal in text form) are preferences: both are read with `.catch(undefined)`, so a value that cannot be read degrades to an interactive sign-in rather than failing the request, and neither is ever a credential. They survive the round trip an interactive flow may take through an IdP, and are stripped from the address bar once consumed.

**`resumable` is not a preference, and it changes the interactive path.** It flips `storeAppSession` from unconditional to conditional — one line, and the one that makes the silent path fail closed for every app that never opted in, since a session that was never stored cannot be silently re-issued. It is therefore load-bearing for the security argument below in a way `prompt` is not: `prompt` decides whether a held session is *consulted*, `resumable` decides whether there is one to consult. It is also the only change here that alters what happens on an ordinary, non-silent sign-in.

**`prompt=none` renders nothing, ever.** Either the redirect carries a session chain or it carries `interaction_required`, which the client can tell apart from a real error and fall back from. Three things make that hold: the context is read before the channel is established, so the handler has it when it answers, ahead of anything that paints; the denial goes on the channel rather than through the error store, which drives a full-page view; and **every** failure path answers, invalid params and the catch-all around the whole handler included. Otherwise a storage or WebCrypto hiccup would paint II's error page inside what the user experiences as their own app's tab and send no response at all, hanging the client's `/reauth` instead of letting it take its fallback branch.

**`prompt=none` never creates a session**, because a session comes only from `prepare_account_session`, which requires an anchor access method that a silent request has not proved. So it can exercise authority and never obtain it — the same rule that stops a stolen chain spawning siblings.

**`hint` selects among the sessions this browser holds for the origin being authorized and can never name another origin's.** That is what makes it safe for the hint to come from a cookie the app can read and write: it selects from what II already holds for that origin, and holding the session is what confers anything. The derivation origin is validated before the session lookup, and the store filters on exact origin match. Several candidates with nothing to choose between them is `interaction_required`, not a guess.

**Silence is something an app asks for, twice.** A session is stored only where the sign-in that created it was marked `resumable`, and a stored session is consulted only under `prompt=none` — anything else, an absent `prompt` included, runs the ceremony. So a session this browser holds is never handed over without the user seeing a screen they did not ask for, and an app that never asked for resumability has nothing held for it at all.

Sibling sharing and sign-out propagation need no mechanism of their own: a shared `derivationOrigin` means every sibling resolves to one application and sessions live at `(anchor, application)`, so there is one record the siblings take turns re-issuing from — and `app_revoke_session` removes the record they all share, not merely the cookie.

`promptParams.ts` and `silentReauth.ts` sit inside the authorize route that uses them.

Tests: `promptParams.test.ts` (10) and `silentReauth.test.ts` (6) cover reading and degrading both parameters, the context surviving a resume, and selection — nothing held, one session, two personas asking rather than guessing, and hinted, unheld and ambiguous hints. `sessionDelegation.test.ts` gains a silent request answering rather than surfacing an unverified origin, and both `prompt=login` and an absent prompt running the ceremony with a session held.


src/internet_identity/internet_identity.did
```diff
@@ -1993,6 +1993,11 @@ service : (opt InternetIdentityInit) -> {
     // that retries, or that signs out twice, does not have to reason about whether its
     // session was already gone. An app can revoke only its own session.
     app_revoke_session : () -> ();
+    // Whether the calling session is still usable. For the II frontend's silent
+    // re-auth path, which must decide whether it can answer without rendering
+    // anything. Advisory: a query reply is not certified, and every mint enforces
+    // the same conditions regardless of the answer here.
+    check_session : () -> (bool) query;
```
MRmarioruci, line 2006: Nit. since we are returning just a boolean to indicate validity.
```suggestion
    session_is_valid : () -> (bool) query;
```
or something in this direction

src/internet_identity/src/sessions.rs
```diff
@@ -414,6 +414,13 @@ fn authorize_session(
     Ok(matched)
 }
 
+/// Whether the calling session is still usable, for the II frontend's silent re-auth
+/// path. Advisory: a query reply is not certified, and the refresh path enforces the
+/// same conditions on every mint regardless of the answer here.
+pub fn check_session() -> bool {
```
MRmarioruci, line 419: Do we need this extra check at all? The client immediately calls app_prepare_delegation after II returns the chain, and that call already tells us whether the session is gone. Could the silent client turn NoMatchingSession into InteractionRequiredError instead? That would avoid another public method, an uncertified pre-check, and the session can still disappear between this check and the mint anyway.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
```diff
@@ -170,10 +229,49 @@ export const handleSessionDelegationRequest =
...
+        const { prompt, hint, resumable } = get(authorizationPromptStore);
+        // Silence is something an app asks for. Anything else, an absent `prompt` included,
+        // runs the ceremony, so a held session is never handed over without the user
+        // seeing a screen they did not request.
+        const held =
+          prompt === "none" ? await appSessionsForOrigin(effectiveOrigin) : [];
+        const chosen = chooseSilentSession({ held, hint });
+
+        let usable = "session" in chosen ? chosen.session : undefined;
+        if (usable && !(await sessionIsLive(usable.record))) {
```
MRmarioruci, line 255: What happens if check_session fails because of a temporary network issue, or the query gives us a bad false? sessionIsLive turns every failure into false, and we then delete a valid stored session. Could we keep “session is gone” separate from “couldn’t check”, and only delete after a proper result?

src/frontend/src/routes/(new-styling)/authorize/silentReauth.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  if (hint !== undefined) {
+    const matched = held.filter((entry) => entry.accountPrincipal === hint);
+    if (matched.length === 1) {
+      return { session: matched[0] };
+    }
+    return {
+      denial:
+        matched.length === 0 ? "login_required" : "account_selection_required",
```
MRmarioruci, line 31: Could we check which sessions are still live before choosing? If storage contains one live session and one that was revoked elsewhere, chooseSilentSession sees two and returns account_selection_required without checking either. The stale record will block a silent sign n even though there is only one usable session.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
MRmarioruci, line 267: Is this afffected by the silent reauth round trip? If yes then this will probably inflate our metrics.. Maybe we could skip on prompt=none

MRmarioruci: Can we skip this for prompt=none? Silent re-auth never uses the identity picker, but this still fetches passkey credential IDs for every remembered identity. That adds unrelated canister queries to every silent request.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
```diff
@@ -26,14 +29,23 @@ import {
...
 import {
   Delegation,
   DelegationChain,
+  DelegationIdentity,
   ECDSAKeyIdentity,
 } from "@icp-sdk/core/identity";
 import type { PublicKey, Signature } from "@icp-sdk/core/agent";
+import { get } from "svelte/store";
+import {
+  chooseSilentSession,
```
MRmarioruci, line 46: Could chooseSilentSession live beside the channel handler or under $lib?

MRmarioruci (approved):

src/internet_identity/internet_identity.did
sea-snake: Leaving the name. No predicate-style method exists in this interface — everything is verb-first, and `check_session` sits directly alongside `check_captcha`, `app_revoke_session`, `prepare_account_delegation` and `get_account_session`.

"Is valid" also overclaims against the method's own doc: it's advisory ("a query reply is not certified, and every mint enforces the same conditions regardless of the answer here"), and what it reports is whether the session is still *usable*, which folds in idleness and revocation rather than validity alone.

Flagging that this is the last free moment for it — a public Candid method on an unmerged PR — so the decline is deliberate rather than deferred.

src/internet_identity/src/sessions.rs
sea-snake: Keeping it, because of where the caller sits. It's the `prompt: "none"` branch in `sessionDelegation.ts`, which drops a dead session to the `login_required` denial.

`prompt: "none"` forbids interaction, so II has to decide **before** it answers. Answer success and the II window is done: the app holds a chain whose first mint fails, and recovering means a second full trip through II after being told the first succeeded. With the check the app gets `login_required` from the request it actually made.

The alternative probe isn't free either. Without a dedicated method the only way II can test liveness is to mint through `app_prepare_delegation` itself — an update call, and mints are precisely what the canister counts as *use*, so probing would reset the session's idleness and corrupt the signal being asked about.

Both of your objections stand and the doc concedes them: it is another public method, and the race is real. The check converts the common case — a session revoked days ago from another device — rather than guaranteeing anything. The `NoMatchingSession` → `InteractionRequiredError` mapping is also only reachable for apps using our client; anything else gets the raw error.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
sea-snake: Nothing is deleted, which changes the shape of this. The code sets `usable = undefined`, and the comment at the call site says so: "The record stays. A session that is really gone leaves a record that is filtered out on read once it expires, and removing it would need a certified answer." A false costs the silent path only; the record survives and the next request re-checks it — there's a test for exactly that.

Separating "gone" from "couldn't check" wouldn't change behaviour either, because both have to end in not answering silently. Answering optimistically hands the app the dead chain the check exists to prevent, and if II can't reach the canister the app's own mint can't either — so optimism relocates the failure and loses the clean `login_required`.

The cost, stated plainly: a transient failure downgrades a silent re-auth to an interactive one, and self-corrects on the next request. The `catch` also covers permanent failures — malformed `chainJson`, a key that won't import — which correctly land on "not usable" too.

src/frontend/src/lib/stores/channelHandlers/silentReauth.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  if (hint !== undefined) {
+    const matched = held.filter((entry) => entry.accountPrincipal === hint);
+    if (matched.length === 1) {
+      return { session: matched[0] };
+    }
+    return {
+      denial:
+        matched.length === 0 ? "login_required" : "account_selection_required",
```
sea-snake: Yes, and this was a real defect — thank you. `chooseSilentSession` counted *records*, not live sessions, and the liveness check ran only on whatever it had already picked. So one live record beside one revoked elsewhere looked like two candidates and was refused as `account_selection_required`, without either being checked.

Fixed by ordering rather than by teaching the chooser about liveness: `held` is filtered by liveness first, then the choice runs over what's left. `chooseSilentSession` stays pure and its tests stay as they are, and the post-check is gone, so no session is checked twice.

Cost is one `check_session` per held session in parallel instead of one total — bounded by the accounts at that origin, and it only rises in the case that was previously answered wrongly. Hinted paths are unchanged: a hint matching a dead record yields `login_required` either way.

src/frontend/src/routes/(new-styling)/authorize/+layout.svelte
sea-snake: It is affected, and you're right about the direction — worse than volume, it changes what the metric means and puts a session that never had a chance to convert into the funnel denominator.

`prompt=none` isn't known at mount, so it can't be a check at the call site. But the discriminator already exists: `isReady` is exactly "an authorization screen is going to render", and the silent branch never reaches `createSession`, the only caller of `setRequestContext`, so it never turns true. The page view now fires from a once-only effect gated on `isReady`.

One behaviour change worth stating: a request that errors before `isReady` — unsupported browser, malformed request — also stops counting. Same correction, I think.

sea-snake: Agreed, and it's the same fix as the `pageView` thread just above — they're now one gated effect.

`prompt=none` isn't knowable at mount, but `isReady` is the same discriminator: it turns true when `setRequestContext` runs, which the silent branch never reaches. So the pre-fetch's one canister query per remembered identity no longer happens for a request that never uses the picker.

The interactive path loses only the gap between mount and request parse, which is over well before the user could act on it, so the pre-fetch keeps its purpose.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
sea-snake: Moved to `lib/stores/channelHandlers/`, beside `sessionDelegation.ts`.

The import was the argument for it: `} from "../../../routes/(new-styling)/authorize/silentReauth";`. Its only non-test importer is in `$lib` and no route uses it, so it was a `$lib` module that happened to sit under `routes/`. The test moved with it.

PR #4249, opened 2026-08-19
feat(fe): a Devices page for the browsers an identity is signed in from

sea-snake:
Design: #4224. Overview: #4230. Renders what #4246 revokes; #4248 stacks on top of this.

The canister knows which browsers an identity has signed in from and can end all of one browser's sessions at once. This is the screen for it: **Devices**, its own page in the manage nav, between Recovery and Settings.

- **Fed by `identity_info.browsers`**, so no extra call: the field rides on the same certified update the page's other data already comes from.
- **Browsers are grouped by the machine they run on** — one card, a heading per platform, most recently used group first and most recently used browser first within it, which is the order someone deciding what to sign out wants to read. The heading counts browsers and hedges the machines ("2 browsers on Mac device(s)"), because two identical laptops report the same thing and nothing sent can tell them apart.
- **The browser reading the page is marked in place**, by a pill on its own row rather than by a section of its own, so it keeps its position among that machine's other browsers. The frontend recognises itself by the browser id it stored at sign-in, because `identity_info` is signed by an access method and the canister cannot tell which browser is asking.
- **A row says what the machine is, not only what it is called.** The form factor picks the glyph on the group heading, and the brand picks the mark on the row — the six browsers that hold essentially all of the market (Chrome, Safari, Edge, Firefox, Samsung Internet, Opera), each from the vendor's own artwork. A browser outside the six keeps its name and shows no mark: an icon nobody recognises says less than the name written out.
- **The name is derived here, not read.** The record holds tokens — brand, system, form factor, and the hardware model where the client could name one (#4242) — and this page turns them into the words a user reads. So renaming a product, or teaching the client to recognise one, renames every row that already exists rather than only the browsers that sign in afterwards. The platform word is deliberately not the system's own name: nobody says they are on "Chrome OS" or "iPadOS", and the hardware wins where it is known, because "Chrome on Pixel 9" is what its owner recognises.
- **Signed out has two causes.** A browser holding no sessions reads as signed out rather than disappearing, so the row stays where the user last saw it — and so does one idle beyond thirty days, whatever its session count says. `MAX_SESSION_TTL_NS` is thirty days and `last_used` advances on every refresh, so every session such a browser held is past its absolute bound; the count includes expired records until some write prunes them.
- **Beyond ninety days a browser is not rendered at all.** Nothing can be signed out there, and keeping it buries the rows that can.
- **The page can never be empty**, because the browser reading it is always one of the rows — including when the canister holds no record for it, a browser signed in to Internet Identity but to no app yet. It describes itself and reads as signed out, last used never, first seen now.
- **Signing out asks first**, and says what it costs: every app opened from that browser asks for sign-in again, while this browser stays signed in to Internet Identity itself. The call is `revoke_browser_sessions`; the entry stays, so signing back in reuses it. Failures surface in a toast rather than silently.

Below `sm` the row stacks the meta under the name — the two columns and the button leave a 375px screen no room for a name beside them.

Tests: `browsers.test.ts` (33), including ordering on use rather than registration, the current browser marked among two entries sharing a name, nothing marked for an id this identity does not hold, the form factor each glyph stands for, grouping by platform with the most recently used group first, a browser past ninety days left out and one a day short of it kept, both causes of a signed-out row, `revoke_browser_sessions` errors mapped to messages, and every brand and platform word this page derives.


src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/SessionDevicesSection.svelte
```diff
@@ -0,0 +1,135 @@
+<script lang="ts">
```
MRmarioruci, line 1: I expect this to go onto it's own page so not reviewing yet..

src/frontend/src/routes/(new-styling)/manage/(authenticated)/settings/components/BrowsersSection.svelte
```diff
@@ -0,0 +1,135 @@
+<script lang="ts">
```
sea-snake: It did go onto its own page, and this PR is that page — **Devices**, in the manage nav between Recovery and Settings. `BrowsersSection.svelte` is gone; the code lives in `routes/(new-styling)/manage/(authenticated)/devices/`.

Worth knowing before you read it: the layout has since changed to the updated design. Browsers are grouped by the machine they run on in one card, the browser you are reading from is marked in place by a pill rather than lifted into a section of its own, and the inactive badge is gone — a browser idle beyond thirty days renders as signed out instead, since `MAX_SESSION_TTL_NS` is thirty days and nothing it holds can still be minted from. Beyond ninety days it is not rendered at all.

MRmarioruci (approved):

PR #4253, opened 2026-08-19
feat(fe): opt in to notifications during sign-in

MRmarioruci:
The backend stack (#4219→#4252) can accept, seal, and deliver notifications, but nothing lets a user turn them on. This adds the opt-in during sign-in, behind the off-by-default `PUSH_NOTIFICATIONS` flag.

A returning user is offered notifications during sign-in, before redirecting to the app, but only when the app asked for them. Opting in is best-effort and never blocks sign-in.

## Changes

Every authorize path routes through a shared gate (`maybeOfferOptIn`): when the flag is on, the app requested notifications (`iiNotifications` on its auth request, carried through to `notificationsRequestedStore`), push is supported, and an identity + origin are known, it shows `NotifOptInView` instead of authorizing immediately, stashing the exact authorization to resume. Without the app's request the screen never shows and sign-in authorizes straight through. "Not now" resumes it directly; "Allow" runs `enableNotifications` first, then resumes (a failure toasts and still authorizes — notifications never block sign-in). `enableNotifications` generates a device VAPID keypair (`vapidPool`), subscribes through the push service worker (`pushSubscription` + `static/push-sw.js`), signs a JWT pool whose bytes match the canister's reassembly, calls `webpush_subscribe_device` then `notification_grant_consent`, and finally mints a scoped account delegation (`pullCredential`) to a service-worker session key, stored in IndexedDB for the worker to pull with later.

The gate covers every path — the continue screen, upgrade-success, and the 1-click OpenID/SSO resume (which stops its redirect animation so the screen can show) — so a returning 1-click user still gets the chance to allow a requesting app. It self-guards on a selected identity, so a resume without one authorizes exactly as before. (The rare SSO normal-login fail-safe still authorizes directly.)

Interim choices, swapped when the surrounding designs land: the pull credential is a plain scoped delegation (not yet an AttributesIdentity/revocable session); the service worker is minimal and fetch-handler-free (it shows a generic notification until the pull logic arrives); and the canister to pull from is resolved later from the stored origin rather than recorded here.

## Tests

`vapidPool`: the exact claim bytes and key order of each window's JWT (the wire contract with the canister), plus keypair and full-pool shape. The browser-only paths — permission, `PushManager.subscribe`, the service worker, the delegation round-trip, and the opt-in firing on the 1-click resume — are covered by CI/e2e.

<div align="left">Previous: #4252</div>
<div align="right">Next: #4257</div>


MRmarioruci:

The opt-in is being reshaped: instead of a flag on the sign-in request it becomes its own JSON-RPC method, so an app can ask when notifications make sense rather than before it knows anything about the user. The consent screens themselves are reused.

PR #4255, opened 2026-08-19
refactor(be): additional input bounds

sea-snake:
Bound the characters a header field body may carry:

- C0 control characters are refused.
- HTAB is allowed (RFC 5322 §3.2.2).
- `CRLF` immediately followed by SP or HTAB is allowed as folding white space (§2.2.3); any other CR or LF is refused.
- One trailing CR/LF run is dropped before the scan, since the gateway can leave its terminating CRLF attached.

The check lives beside `validate_message` rather than inside it. An `Err` from that path becomes the `SmtpResponse` the gateway acts on, and `handle_smtp_request` reserves those for envelope and request shape — per-message outcomes answer `Ok` (module note in `email_inbound/smtp.rs`). So the inbound path runs the check and drops the message while still answering `Ok`, alongside the other silent-drop cases.

`validate_message` itself is unchanged, and errors keep the existing `smtp_err(SMTP_ERR_SYNTAX_ERROR, …)` shape.

## Tests

- `cargo test -p internet_identity_interface --lib` — 85 passed
- `cargo test -p internet_identity --bins` — 660 passed
- `scripts/test-canisters.sh email_recovery::` — 33 passed

New cases in `types::smtp::tests`: embedded bare LF, embedded CRLF without folding WSP, lone CR, folded value, value carrying a trailing CRLF, HTAB in value, and NUL / BEL / ESC.

`recovery_rejects_crlf_spliced_from_header` keeps asserting `Ok` on the wire and that recovery never completes. Since the spliced message is now dropped on arrival, the challenge is left pending rather than reaching `Failed(AddressMismatch)` — leaving it untouched also means an unauthenticated sender cannot cancel someone else's pending challenge.

### DIFF
--- a/src/internet_identity_interface/src/internet_identity/types/smtp.rs
+++ b/src/internet_identity_interface/src/internet_identity/types/smtp.rs
@@ -261,7 +261,6 @@ pub fn validate_message(message: &SmtpMessage) -> Result<(), SmtpResponse> {
                 ),
             ));
         }
-        validate_header_value_chars(&header.name, &header.value)?;
     }
     if message.body.len() > MAX_BODY_BYTES {
         return Err(smtp_err(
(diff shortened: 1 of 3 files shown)


MRmarioruci (approved):

src/internet_identity_interface/src/internet_identity/types/smtp.rs
```diff
@@ -275,6 +275,64 @@ pub fn validate_message(message: &SmtpMessage) -> Result<(), SmtpResponse> {
...
+            format!("Header '{name}' value contains a malformed CR/LF sequence"),
+        )
+    };
+    let mut i = 0;
+    while i < bytes.len() {
+        match bytes[i] {
+            b'\t' => i += 1,
+            b'\r' => {
+                if bytes.get(i + 1).copied() != Some(b'\n')
+                    || !matches!(bytes.get(i + 2).copied(), Some(b' ' | b'\t'))
```
aterga, line 317:
```suggestion
                    || !matches!(bytes.get(i + 2).copied(), Some(b' ') | Some(b'\t'))
```

PR #4257, opened 2026-08-19
feat(fe): dedicated notifications management page

MRmarioruci:
#4253 lets a user turn notifications on; this adds the surface to manage them afterwards — the device toggle, per-app muting, and revoke. Behind the off-by-default `PUSH_NOTIFICATIONS` flag.

## Changes

Notifications get their own `/manage/notifications` page with a sidebar entry, rather than a card buried in Settings; Settings keeps a link card that points at it. The page has two parts.

At the top, a device push toggle reflects whether *this* browser is registered — read through `readDeviceState()` (live subscription plus a matching stored key), so an orphaned subscription from a failed enable reads as off. Turning it on runs the enable flow (`enableDeviceNotifications`); turning it off unsubscribes this device (`unsubscribeDevice`). When the browser is blocked or the toggle can't take effect, the page shows the unblock steps instead of a dead switch.

Below it, the apps allowed to notify this identity, read from `notification_consented_apps` and listed by dApp name with when each was allowed and last sent. Each app has a mute switch — `notification_set_app_muted` — that stops delivery while keeping the grant (the send path skips a muted app, #4251), and a remove control behind a confirmation dialog. Removing optimistically drops the row, then `notification_revoke_consent` on the canister (which drops that app's seals); a failed revoke rolls the row back. Once consent is gone it drops the service worker's pull credential for that app — best-effort, and lazily imported so the credential store's IndexedDB open stays off the page. Revoke is decoupled from auth: signing out never reaches here, only this explicit action does.

## Tests

None added — the page is a thin wrapper over `notification_consented_apps` / `notification_set_app_muted` / `notification_revoke_consent` and the device enable/unsubscribe helpers, so it belongs to the e2e layer rather than a unit test.

<div align="left">Previous: #4253</div>
<div align="right">Next: #4258</div>


MRmarioruci:

The management page becomes its own feature, once consent has landed.

PR #4258, opened 2026-08-19
feat(fe): service worker pulls and renders notifications

MRmarioruci:
The stack can now consent, subscribe, seal, deliver a ping, and manage apps — but the ping still shows a placeholder. This makes the service worker real: it pulls the notification's content from the dApp and renders it, so a delivered ping becomes an actual notification. Behind the off-by-default `PUSH_NOTIFICATIONS` flag.

## Changes

The placeholder `static/push-sw.js` is replaced by a bundled SvelteKit service worker (`src/service-worker.ts`), registered on opt-in rather than every load (`kit.serviceWorker.register` is off) and with no `fetch` handler, so it stays off the auth origin's request path. On `push`, the browser has already decrypted II's sealed payload, so the worker reads the routing origin (`{"o":"<origin>"}`), loads that app's credential from IndexedDB, resolves the dApp canister from the origin's own discovery document, and pulls the pending notifications as the user's per-app identity — rendering each. With no usable credential, an expired one, or an unreachable canister it shows a generic, app-attributed notification; a `notificationclick` focuses an open tab for the app or opens one. To let the worker build an agent without the browser-only `globals`, the pull credential now also stores the IC host it was minted against.

## For review — provisional bits

- **Pull interface.** `ii_pending_notifications : () -> (vec record { title; body }) query` is a placeholder shape the worker expects; the client crate PR finalizes it.
- **Canister resolution.** The worker fetches `<origin>/.well-known/ii-notification-senders` at push time (cross-origin, so it needs the dApp to serve CORS). D-30.1 would prefer recording the canister at consent — a cleaner follow-up once the shape settles.
- **agent-js in a worker.** It bundles cleanly (`npm run build` emits `dist/service-worker.js`), but the authenticated pull is only exercisable against a live relay + dApp, so it is CI/manual-verified, not unit-tested. The generic fallback is the guaranteed path.
- **Module worker.** Registered with `{ type: "module" }` (Chrome 91+, Safari 16.4+, Firefox 111+).

The worker resolves the sending canister from the origin stored with the pull credential, which is canonicalized to the legacy gateway for principal stability and so is not necessarily served there. It looks for the well-known on each gateway, the same way the canister-side fetch does (#4251); without that the lookup fails on those origins and every notification falls back to the generic one.

<div align="left">Previous: #4257</div>


MRmarioruci:

The service worker moves to the content-pull feature, where it authenticates as a browser with the browser key-pair instead of holding a pull credential.

PR #4260, opened 2026-08-20
fix(fe): don't establish the authorize channel while redirecting to the primary origin

MRmarioruci:
`legacy.spec.ts:52` — a dapp pointing its ICRC-25 transport at a legacy domain — fails on mobile with `Target page, context or browser has been closed`, on `main` as well as here (e.g. [run 32347808452](https://github.com/dfinity/internet-identity/actions/runs/32347808452), both legacy domains, 3/3 attempts). #4210's redirect to the primary origin is supposed to happen before anything talks to the dapp, but nothing enforces that: `location.replace` only schedules the navigation, the app hydrates on the legacy document anyway, and the channel is established right there.

## Changes

`redirectToPrimaryOrigin`, in the client `init` hook, no longer redirects a document whose hash carries a redirect message from the primary origin. That leg of the legacy relay has to finish where the dapp opened it, since `endRedirectSession` needs the key `startRedirectSession` left in this origin's `sessionStorage`. It reads the two params the relay uses straight off the hash, and requires the primary origin as the sender so a crafted hash can't pin an authorize page to a legacy domain.

`getTransports` then only constructs the postMessage and URL transports on the primary origin — or when no primary origin is configured, which is also when nothing redirects. Both pin the signer origin the moment the client answers their handshake, so off-primary that hands the client a window which is about to be replaced. Restricting where they exist makes it impossible instead of a race the navigation usually wins. The legacy transport is unaffected: it carries its own request across the redirect.

## Tests

`primaryOrigin.test.ts` covers the redirect decision per leg: the bounced response stays, a plain `/authorize` redirects, a message from any other origin still redirects. `legacy.spec.ts:18` and `:52` cover the two flows end to end.

### DIFF
--- a/src/frontend/src/lib/stores/channelStore.ts
+++ b/src/frontend/src/lib/stores/channelStore.ts
@@ -53,9 +53,16 @@ type ChannelStore = Readable<Channel | undefined> & {
 
 const getTransports = (): Transport[] => {
   const primaryOrigin = getPrimaryOrigin();
+  // Off the primary origin this document is on its way there, and both of these
+  // transports pin the signer origin the moment the client answers their
+  // handshake — stranding it on a document that is being replaced. Only the
+  // legacy transport belongs here; it carries its request across the redirect.
+  const offPrimaryOrigin =
+    primaryOrigin !== undefined && window.location.origin !== primaryOrigin;
   return [
-    new PostMessageTransport(),
-    new UrlTransport(),
+    ...(offPrimaryOrigin
+      ? []
+      : [new PostMessageTransport(), new UrlTransport()]),
     new LegacyTransport(
       // Redirect requests and responses between related origins and primary origin
       primaryOrigin !== undefined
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/primaryOrigin.ts
```diff
@@ -42,6 +42,15 @@ export const redirectToPrimaryOrigin = (): boolean => {
   if (pathname === "/" && hash === "#authorize") {
     return false;
   }
+  // Same exchange, response leg: only this document can finish it, the redirect
+  // session that re-wraps the delegation lives in this origin's `sessionStorage`.
```
sea-snake: Good catch! The comment should be more clear this is specifically for the legacy transport return leg, in that regard the above condition is the outgoing leg and this is the return leg.

src/frontend/src/lib/stores/channelStore.ts
```diff
@@ -53,9 +53,16 @@ type ChannelStore = Readable<Channel | undefined> & {
 
 const getTransports = (): Transport[] => {
   const primaryOrigin = getPrimaryOrigin();
+  // Off the primary origin this document is on its way there, and both of these
+  // transports pin the signer origin the moment the client answers their
+  // handshake — stranding it on a document that is being replaced. Only the
+  // legacy transport belongs here; it carries its request across the redirect.
```
sea-snake: Maybe it's just me, while I understand the intent of this comment, it reads vague  :/

sea-snake (approved):

PR #4264, opened 2026-08-22
feat(be): verify a browser's key and its announced successor

sea-snake:
Design: #4224. Overview: #4230. The verifier only — its caller is the sign-in ceremony in #4268, which also removes the `#[allow(dead_code)]` this carries.

A session records which browser created it, so the user can see where they are signed in and end one browser's access everywhere at once. That is only worth anything if the browser named in a sign-in is the browser that made it; otherwise an attacker holding an access method could attribute their session to a browser the user recognises, and the settings list would lie. Ingress messages are public, so a key presented in one sign-in can be read off the wire by anyone.

`verify_browser_keys` checks two signatures before a sign-in may name a browser: the **current key** over the session key and the announced successor, and the **successor** over the session key and the current key. Each is made under its own domain separator, `ii-session-browser-key` and `ii-session-browser-successor`, so neither can be replayed in the other's role.

Requiring possession of the successor is what closes the wire-reading attack: announcing a key you do not hold is impossible without its private half, so a key read off the wire cannot be claimed when its owner next presents one.

**Verifying yields evidence, not a boolean.** It returns a `VerifiedBrowserKeys` whose fields are private and whose only constructor is this function, and `CreateSessionParams` requires one — so `create_session` registering a browser from these keys cannot depend on a caller having remembered to check. That is also why the module sits at the crate root rather than under `sessions`: storage must not depend upward on sessions, and a P-256 verifier that knows nothing of either is something both layers may use.

**One encoding per key.** P-256 only, no fallback. A key that does not parse, a signature of the wrong length, and an empty signature are each refused rather than skipped — and so is a compressed SEC1 point, although it is valid DER for the same key. `from_public_key_der` accepts both, so one private key had two spellings and `resolve_browser`, which compares the bytes it is handed, would have seen two browsers. WebCrypto emits only the uncompressed form, so nothing legitimate is turned away and nothing stored needs migrating.

Tests: `browser_key::tests` (11), one accepting case and ten refusals — including a successor signature replayed as the current one, a signature over another session key, one paired with a substituted successor, and a compressed key that `from_public_key_der` accepts.


Cargo.toml
```diff
@@ -49,7 +49,7 @@ rsa = "0.9.10"
 minicbor = "1.0.0"
 
 # DNSSEC verifier deps (PR 1b on docs/ongoing/email-recovery.md §7)
```
MRmarioruci: Unrelated, let's remove this comment.. not only about dnssec anymore

src/internet_identity/src/sessions/browser_key.rs
```diff
@@ -0,0 +1,283 @@
...
+/// hold it — without it, keys read off the wire could be planted as another browser's
+/// successor and claimed when that browser next presented one.
+pub fn verify_browser_keys(
+    current_browser_key: &PublicKey,
+    current_browser_key_signature: &[u8],
+    next_browser_key: &PublicKey,
+    next_browser_key_signature: &[u8],
+    session_key: &SessionKey,
+) -> bool {
+    verify(
```
MRmarioruci:
```suggestion
    let current_key_signature_valid = verify(
        current_browser_key,
        current_browser_key_signature,
        &signed_message(BROWSER_KEY_SIGNATURE_DOMAIN, session_key, next_browser_key),
    );
    let successor_key_signature_valid = verify(
        next_browser_key,
        next_browser_key_signature,
        &signed_message(
            SUCCESSOR_KEY_SIGNATURE_DOMAIN,
            session_key,
            current_browser_key,
        ),
    );

    current_key_signature_valid && successor_key_signature_valid
```

MRmarioruci (approved):

src/internet_identity/src/browser_key.rs
```diff
@@ -0,0 +1,283 @@
...
+use p256::ecdsa::{Signature, VerifyingKey};
+use p256::pkcs8::DecodePublicKey;
+
+/// Prefixed to the signed message so the browser key cannot be made to sign for another
+/// purpose by presenting a message from one.
+const BROWSER_KEY_SIGNATURE_DOMAIN: &[u8] = b"ii-session-browser-key";
+
+/// A different prefix for the successor's own signature, so neither signature can be
+/// replayed in the other's role.
+const SUCCESSOR_KEY_SIGNATURE_DOMAIN: &[u8] = b"ii-session-browser-successor";
```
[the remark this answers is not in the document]
sea-snake: The staleness is in the description, not the code. The canister, the test signer in `framework.rs` and `browser-key.store.ts` all use `ii-session-browser-key` and `ii-session-browser-successor`, so nothing produces a signature that fails to verify — the body was left behind by the device→browser rename. Rewritten to match: `verify_browser_keys`, the two `browser` separators, and `browser_key::tests`.

Cargo.toml
sea-snake: Agreed, removed. `p256` is the browser-key verifier this PR adds as well as a DNSSEC dependency, so the heading was wrong.

Removed rather than reworded, which matches the file: `serde_bytes`, `sha2`, `rsa` and `minicbor` carry no comment at all, and a note about *why* something was added belongs in the commit. `# Certification` below it stays — that is a grouping label, not a rationale.

src/internet_identity/src/sessions/browser_key.rs
sea-snake: Keeping the short-circuit, for a reason that is about cost rather than style: your version runs both verifications unconditionally, so a request whose first signature already failed still pays for a second P-256 verification. On a sign-in path `&&` stopping at the first failure is the right call.

The naming it would buy — which of the two crossed signatures each call checks — is already carried by the comment below the function explaining why the current key signs over the successor and vice versa. And the timing difference the short-circuit creates leaks nothing: both keys are public and travel on the wire.

The function did change shape here, though, for the reason in your #4266 thread: it returns a `VerifiedBrowserKeys` now instead of a `bool`.

MRmarioruci (approved):

PR #4265, opened 2026-08-22
feat(be): mint short-lived app delegations from a session

sea-snake:
Design: #4224. Overview: #4230. Depends on #4268 for the ceremony that creates a session.

This is the PR the feature turns on: after it, a delegation an app holds lasts five minutes and the session behind it can be revoked.

**`app_prepare_delegation` / `app_get_delegation`.** An app holds a session chain and asks for a delegation with it. What it gets is capped at five minutes and the cap is not requestable, so revoking the session ends access within one delegation lifetime.

**Finding the session from the call.** The index mapping a session's rooted principal to `{account_principal, session_id}` is written by #4266; this adds the two accessors that read it and the caller that needs them. The call names nothing and attaches nothing: `caller()` is looked up, the account principal resolves through the principal index from #4238, and the record is read from the list. A hit is itself the proof that the caller is that session, since only the holder of its key can arrive as that principal.

The account is named by principal rather than by locator on purpose: materialising a default account changes its locator and leaves its principal alone, so naming an account touches one index entry instead of every session on it.

**The entry names the session by its id.** A browser keeps its id across sign-ins, so an index entry keyed on the browser and outliving its session would authenticate its holder as whatever that browser created next — a revoked chain coming back to life. `session_id` is allocated per session and never reissued, so an entry can only ever resolve to the one session it was written for. Every path that destroys a session drops its index entry in the same write; the id is what makes a stale entry fail closed regardless.

**The `get` re-derives the five-minute ceiling** rather than trusting the `expiration` it is handed, because longer-lived delegations exist over the same account seed. An account's own principal is absent from the session index, so an app delegation cannot mint its own replacement.

Tests: `integration/sessions.rs` (11), including a refresh refused once the session has expired, a session replaced by a new ceremony invalidating the previous chain, and the account principal minting nothing itself. Index accessor coverage lands in `account_principal_index_tests` and `session_creation_tests`.


src/internet_identity/src/sessions.rs
```diff
@@ -304,3 +306,120 @@ fn account_principal(
...
+    .map_err(|_| AppSessionError::NoMatchingSession)
+}
+
+/// Authenticates a refresh from `caller()` alone.
+///
+/// The session index is keyed by the principal a session's chain is rooted at, so a hit is
+/// itself the proof that the caller is that session: nothing is named in the request and
+/// nothing is attached to it.
+fn authorize_session(now: Timestamp) -> Result<(Account, SessionRecord), AppSessionError> {
+    let handle = storage_borrow(|storage| storage.lookup_session_with_principal(caller()))
```
MRmarioruci: Could we read the session handle and account locator in one storage borrow here? They are part of the same lookup anyways..

src/internet_identity/src/sessions.rs
```diff
@@ -304,3 +306,120 @@ fn account_principal(
...
+    // sign-ins: on the browser alone, an index entry that outlived its session would
+    // authenticate its holder as whatever that browser created next.
+    let session = sessions
+        .into_iter()
+        .find(|session| {
+            session.device_id == handle.device_id && session.created_at_ns == handle.created_at
+        })
+        .ok_or(AppSessionError::NoMatchingSession)?;
+    if session.is_over(now) {
+        return Err(AppSessionError::NoMatchingSession);
```
MRmarioruci, line 413: THis is more of an expired Session error right?

src/internet_identity/src/sessions.rs
```diff
@@ -304,3 +306,120 @@ fn account_principal(
...
+        || request.expiration > session.valid_till_ns
+    {
+        return Err(AppSessionError::NoMatchingSession);
+    }
+
+    let seed = account_seed(&account)?;
+    let access = DelegationAccess::from_read_only(session.read_only);
+    let permissions = access.permissions();
+
+    state::assets_and_signatures(|certified_assets, sigs| {
```
MRmarioruci, line 370: This part is pretty much witness_session_delegation if I am not mistaken. can we avoid the duplication and use it in here?

src/internet_identity/src/storage/storable/session_handle.rs
```diff
@@ -25,6 +26,12 @@ pub struct StorableSessionHandle {
     pub created_at: u64,
 }
 
+impl StorableSessionHandle {
+    pub fn account(&self) -> Principal {
```
MRmarioruci, line 26:
```suggestion
    pub fn account_principal(&self) -> Principal {
```

src/internet_identity/src/sessions.rs
```diff
@@ -313,6 +316,111 @@ fn account_principal(
...
+pub fn app_get_delegation(
+    request: AppGetDelegationRequest,
+) -> Result<SignedDelegation, AppSessionError> {
+    let now = time();
+    let (account, session) = authorize_session(now)?;
+
+    if request.expiration > now.saturating_add(APP_DELEGATION_TTL_NS)
+        || request.expiration > session.valid_till_ns
+    {
+        return Err(AppSessionError::NoMatchingSession);
```
MRmarioruci:
```suggestion
        return Err(AppSessionError::NoMatchingSessionDelegation);
```

src/internet_identity/src/sessions.rs
```diff
@@ -313,6 +316,111 @@ fn account_principal(
...
+/// Authenticates a refresh from `caller()` alone.
+///
+/// The session index is keyed by the principal a session's chain is rooted at, so a hit is
+/// itself the proof that the caller is that session: nothing is named in the request and
+/// nothing is attached to it.
+fn authorize_session(now: Timestamp) -> Result<(Account, SessionRecord), AppSessionError> {
+    let key = storage_borrow(|storage| storage.lookup_session_with_principal(caller()))
+        .ok_or(AppSessionError::NoMatchingSession)?;
+
+    let (account, session) = storage_borrow(|storage| {
```
MRmarioruci, line 459: Could storage resolve this directly to (Account, SessionRecord)? read_account and read_session both look up the application and decode the same reference list. The session index already has the application number, so this could be done in one pass.

src/internet_identity/src/sessions.rs
sea-snake: Already consolidated further than when you asked, so leaving it here.

The account and the session now share a borrow; only the index lookup is separate, and it has to be, because the key the second borrow uses comes from the first. Collapsing the chain into one closure is possible but buys nothing today: all three failures already map to `NoMatchingSession`.

What did change is the return — see your other thread on this function: it is an `AuthorizedSession` now rather than a tuple.

src/internet_identity/src/sessions.rs
sea-snake: Keeping `NoMatchingSession`, for two reasons.

A distinct variant is another case the frontend has to handle for a situation it does not need to tell apart — the remedy is a fresh sign-in either way. And it could not be told truthfully: an expired session can already have been pruned by an unrelated write, so "found but expired" and "not found" are the same observation half the time, and a distinct error would be wrong whenever the sweep got there first.

Not in tension with the `NoSuchDelegation` variant added on #4268: a missing *signature* for a session that is present is a different fact, and one the caller can act on by asking with the parameters that were signed.

src/internet_identity/src/sessions.rs
sea-snake: Keeping them apart, deliberately — they are two different delegation types.

A shared helper would take `targets` and `permissions` as parameters, and those two parameters are exactly where a session credential and an app delegation must differ: the session credential is scoped to Internet Identity and update-capable, an app delegation is unscoped and carries read-only. So one function is a footgun taken on purpose — miscalling it mints an app delegation scoped to II, or a session credential scoped to nothing.

The scoping change on #4268 makes that sharper rather than softer. Each site states its own facts and cannot be handed the other's.

src/internet_identity/src/storage/storable/session_handle.rs
sea-snake: Taken, and its sibling is renamed with it, because that was the actual problem: there were two `account()` methods returning two different types. This one returns a `Principal` and is now `account_principal()`, matching the field it reads; `SessionRecordKey::account()`, which returns an `AccountKey`, is now `SessionLocator::account_key()`.

src/internet_identity/src/sessions.rs
sea-snake: Same answer as your thread higher up on this function, and the shape did change.

The account and the session already share one storage borrow; only the index lookup is separate, and it has to be, because the key the second borrow uses comes from the first. What is new is that the three values come back as a type rather than a tuple — `AuthorizedSession { locator, account, session }`, which only `authorize_session` can construct, so holding one is the evidence that the principal lookup and the liveness check both happened. That also fixed the call site that discarded the key positionally with `let (_, account, session)`.

src/internet_identity/src/sessions.rs
sea-snake: Taken, though it arrives with a rename beside it.

Both of `app_get_delegation`'s refusals are `NoSuchDelegation` now, because both mean the same thing: the caller asked with an expiration `app_prepare_delegation` never returned. That is true of the ceiling guard above as well as the signature lookup here — `prepare` hands back `min(now + APP_DELEGATION_TTL_NS, session.valid_till_ns)` and `now` has only advanced since, so the value it gave out cannot exceed either bound by the time `get` runs. The guard can only fire on a value the caller made up.

Worth being explicit about what makes it worth a variant, since the neighbouring thread about an expired session got the opposite answer: this one is actionable and that one is not. A missing signature means prepare again and use what comes back; a session that is gone means sign in again, and an expired session cannot even be told from a pruned one.

The rename: `AppSessionError::NoMatchingSession` is now `NoSuchSession`. It was the same fact as its counterpart on `AccountSessionError`, in different words, sitting beside `NoSuchAccount` and `NoSuchDelegation` — the only difference being that this side matches the caller rather than being handed a session id, which is not worth a second vocabulary.

MRmarioruci (approved):

src/internet_identity/src/sessions.rs
```diff
@@ -343,6 +346,141 @@ fn get_account_principal_for_origin(
...
+/// requestable by the app.
+pub const APP_DELEGATION_TTL_NS: u64 = 5 * MINUTE_NS;
+
+pub fn app_prepare_delegation(
+    request: AppPrepareDelegationRequest,
+) -> Result<AppPrepareDelegationResponse, AppSessionError> {
+    let now = time();
+    let AuthorizedSession {
+        account, session, ..
+    } = authorize_session(now)?;
```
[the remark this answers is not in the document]
sea-snake: True of this PR read alone, and closed by the next one — this stack merges whole. `record_session_use` appears nowhere at #4265 and arrives at #4244, which stamps both `session.last_refreshed_ns` and the account reference's `last_used` in a single write, and does it *before* the mint so a later refusal cannot leave a stamp behind (returning `Err` on the IC commits what preceded it).

Your second point is the same cause rather than a separate race: once the prepare stamps, it pushes the idle deadline further out, so the following `get` cannot fall the other side of a boundary the prepare just moved. Without the stamp it could straddle, which is the defect #4244 removes.

src/internet_identity/src/sessions.rs
```diff
@@ -343,6 +346,141 @@ fn get_account_principal_for_origin(
     ))
 }
 
+/// The one window revocation cannot reach. Matches what MCP mints, and is not
```
nmattia, line 349: I understand MCP was introduced first but arguably its TTL is not very relevant here

src/internet_identity/src/sessions.rs
```diff
@@ -343,6 +346,141 @@ fn get_account_principal_for_origin(
...
+
+    let (account, session) = storage_borrow(|storage| {
+        Some((
+            storage.read_account(&locator.account_key())?,
+            storage.read_session(&locator)?,
+        ))
+    })
+    .ok_or(AppSessionError::NoSuchSession)?;
+
+    if session.is_expired_or_idle(now) {
```
nmattia, line 467: nit (not introduced by this PR): `...or_idle()` is confusing since we explicitly allow a session to idle for up to `max_idle_ns`, and it sounds like being idle for longer than `max_idle_ns` would expire the session, so I think it would make more sense to call this `.is_expired()`.

nmattia, line 467: for my own understanding: are sessions removed from storage after they've expired? if so, who's responsible for this?

PR #4266, opened 2026-08-22
feat(be): store a revocable session on the account reference

sea-snake:
Design: #4224. Overview: #4230. The writers have no caller until the ceremony in #4268, which removes the `#[allow(dead_code)]` annotations this adds — each at the PR that first calls the function it sits on. The per-identity cap arrives in #4267.

A session record on the account reference: created at, expires at, last refreshed, the browser that made it, and whether the user consented to queries only.

**Its identity.** `session_seed = H(salt, "session", account_seed, session_id)`, every field length-prefixed. The id rather than `created_at`, which is not unique within a round. Building on the account's own seed rather than on the identity, application and account numbers is what makes a session survive anything that leaves the account's principal unchanged — naming a default account is exactly that, since it gains a number and a name while still deriving from the identity it was conjured from. Had the numbers been inputs, naming an account would have signed the user out of every app using it.

Only `last_refreshed` is mutable, which is why the seed does not take it: a mutable input would change the session's principal every time it was stamped.

**Replacement, not reuse.** A ceremony from a browser that already holds a session at this account deletes it and mints a new one, so a copy of the old chain stops working at the user's next sign-in rather than at its expiry. Expired records on the list go in the same write, which is what keeps the design free of any sweep.

Read hardest: the seed derivation, since it decides what survives an account being renamed.

Tests: `session_creation_tests` (23), including a second browser getting its own session, the same browser's being replaced, expired records pruned on write, an account the identity does not hold refused, and the seed being unchanged by naming a default account.


src/internet_identity/src/storage.rs
```diff
@@ -1750,6 +1768,346 @@ impl<M: Memory + Clone> Storage<M> {
...
+        reference.sessions.retain(|session| {
+            if session.device_id == device_id {
+                dropped.push((account_number, session.clone()));
+                return false;
+            }
+            true
+        });
+
+        let session = SessionRecord {
+            created_at_ns: now_ns,
```
MRmarioruci: Do we rely on the timestamp for uniqueness of this record? If so, wouldn't it be better to have a distinct unique value?

src/internet_identity/src/storage/storable/anchor.rs
```diff
@@ -40,6 +40,11 @@ pub struct StorableAnchor {
     /// Monotonic per-anchor allocator for `browsers`. Ids are never reused.
     #[n(8)]
     pub next_browser_id: Option<StorableBrowserId>,
+    /// Live sessions this anchor holds, as a trigger for the session cap rather than a
```
MRmarioruci: Just keep the first part of the comment until the comma.. the rest the reader can understand, it's usage might not be limited to only this.

src/internet_identity/src/storage/storable/session_record.rs
```diff
@@ -22,6 +23,8 @@ pub struct StorableSessionRecord {
     pub browser_id: StorableBrowserId,
     #[n(5)]
     pub read_only: bool,
+    #[n(6)]
+    pub session_id: StorableSessionId,
```
MRmarioruci:
```suggestion
    pub id: StorableSessionId,
```

src/internet_identity/src/storage/account.rs
```diff
@@ -53,8 +53,54 @@ impl AccountReference {
...
+/// successor.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SessionRecordKey {
+    pub anchor_number: AnchorNumber,
+    pub origin: FrontendHostname,
+    pub account_number: Option<AccountNumber>,
+    pub session_id: SessionId,
+}
+
+impl SessionRecordKey {
```
MRmarioruci:
```suggestion
impl SessionKey {
```
everywhere basically no need to say Record. It's like saying object in JS or whatnot.

src/internet_identity/src/storage.rs
```diff
@@ -103,12 +103,15 @@ use identity_jose::jwk::Jwk;
 use internet_identity_interface::archive::types::BufferedEntry;
 
 use crate::delegation::{self, check_frontend_length};
+use crate::delegation::{calculate_session_seed_with_salt, canister_sig_principal};
```
MRmarioruci: just use the import right above for these two functions as well

src/internet_identity/src/storage.rs
```diff
@@ -1580,6 +1616,37 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// Everything this identity has stored, keyed the way a write takes it.
+    ///
+    /// The symmetry is the point: an operation that touches many origins reads this,
+    /// changes what it means to change, and writes it back. Nothing has to range over
+    /// storage itself, and no application number reaches the caller.
+    ///
+    /// Stored entries only. An origin nothing has been stored under is not in the map —
+    /// there are unboundedly many of those — so a caller that means to write at one
+    /// reaches for [`Self::account_state_for_origin`] and puts it there.
+    fn account_state(
```
MRmarioruci, line 1615:
```suggestion
    fn account_state(
    &self,
    anchor_number: AnchorNumber,
) -> Result<BTreeMap<FrontendHostname, AccountReferenceListWrite>, StorageError> {
    self.stable_account_reference_list_memory
        .range(
            (anchor_number, ApplicationNumber::MIN)
                ..=(anchor_number, ApplicationNumber::MAX),
        )
        .map(|((_, application_number), list)| {
            let application = self
                .stable_application_memory
                .get(&application_number)
                .ok_or(StorageError::OriginNotFoundForApplicationNumber {
                    application_number,
                })?;

            let account_references = Vec::<AccountReference>::from(list)
                .into_iter()
                .map(AccountReferenceWrite::from)
                .collect();

            Ok((
                application.origin,
                Some((account_references, None)),
            ))
        })
        .collect()
}
```

Then callers can use like
```
let mut state = self.account_state(anchor_number)?;
```
revoke_browser_sessions treats this as the complete session state, so skipping a list could leave some sessions behind while the call still succeeds.

src/internet_identity/src/storage.rs
```diff
@@ -2500,6 +2640,417 @@ impl<M: Memory + Clone> Storage<M> {
...
+                anchor_number,
+                name: String::new(),
+            })?;
+        let reference = &mut account_references[position].account_reference;
+        reference.last_used = Some(now_ns);
+
+        // A ceremony replaces whatever this browser held here, rather than reusing it: the
+        // copy of an old session's chain stops working at the user's next sign-in instead of
+        // at its expiry.
+        let mut dropped: Vec<(Option<AccountNumber>, SessionRecord)> = vec![];
```
MRmarioruci: I don't see this being used anywhere.. why do we keep it?

src/internet_identity/src/storage.rs
```diff
@@ -2500,6 +2640,417 @@ impl<M: Memory + Clone> Storage<M> {
...
+                dropped.push((account_number, session.clone()));
+                return false;
+            }
+            true
+        });
+
+        // After the checks that can refuse this ceremony, so a refused one does not burn
+        // an id. Ids need not be contiguous, so a later failure leaving a gap is fine;
+        // what must never happen is one being handed out twice.
+        let session_id = self.allocate_session_id()?;
```
MRmarioruci, line 2886: This immediately updates stable memory even on failure which would result in a consumed id in any case. Not a big issue but we could reserve and then allocate probably

src/internet_identity/src/storage.rs
```diff
@@ -3086,6 +3653,26 @@ impl<M: Memory + Clone> Storage<M> {
...
+// Constructed by the sign-in ceremony, which lands two PRs up.
+#[allow(dead_code)]
+pub struct CreateSessionParams {
+    pub anchor_number: AnchorNumber,
+    pub origin: FrontendHostname,
+    pub account_number: Option<AccountNumber>,
+    /// What the browser proves it holds, and the successor it announces. Its registry
+    /// entry, its id, and whatever the cap gives up to make room for it are all worked out
+    /// inside the write, so no caller states any of them.
+    pub current_browser_key: PublicKey,
```
MRmarioruci: Could verify_browser_keys return a VerifiedBrowserKeys value and have CreateSessionParams require that? create_session updates the browser registry from these keys, so its safety currently depends on every caller remembering to verify them first.

MRmarioruci (approved):

src/internet_identity/src/storage.rs
sea-snake: It did when you asked, and you were right: the record was keyed on `created_at`, which is not unique within a round. Sessions carry an allocated `session_id` now, and that is what the principal index, the reclaim sort key and revocation key on.

src/internet_identity/src/storage.rs
```diff
@@ -2624,6 +2764,412 @@ impl<M: Memory + Clone> Storage<M> {
...
+        // After the refusals above, so a ceremony that cannot happen registers nothing —
+        // the record reaches storage only through the write at the end.
+        let (browser_id, dropped_browsers) = anchor
+            .resolve_browser(current_browser_key, next_browser_key, browser_name, now_ns)
+            .map_err(StorageError::Browser)?;
+
+        // The whole of what the identity holds, not just this origin: a browser the
+        // registry gave up to make room for this one may hold sessions anywhere, and those
+        // have to go in the same write as the browser that held them.
+        let mut state = self.account_state(anchor_number);
```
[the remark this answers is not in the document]
sea-snake: Confirmed, and fixed. `create_session` now writes the origin it signs in at and nothing else. The sessions of a browser the registry gave up are still swept, but by the gate, which derives them from the registry the write carries — so no caller has to remember the consequence — and it adds only the origins that actually hold one.

No test caught this, which is the more interesting half: all fifteen tests in `tracked_default_eviction_tests` drive a one-origin write path (`record_use`, `set_default_account`, `write_at`), so eviction was thoroughly covered and the sign-in path was not covered for eviction at all. `a_sign_in_evicts_the_stale_defaults_too` covers it now, and fails without the change with 501 lists where the watermark plus the new one is 451.

src/internet_identity/src/storage.rs
```diff
@@ -2276,14 +2378,31 @@ impl<M: Memory + Clone> Storage<M> {
...
+                }
+                if sessions_changed {
+                    session_delta += self.sync_session_index(
+                        anchor_number,
+                        application_number,
+                        &origin,
+                        &salt,
+                        &previous_references,
+                        &current_references,
+                    );
```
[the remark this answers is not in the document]
sea-snake: Half right. Creation, revocation and an every-entry sweep all assert the index directly. Replacement did not: a second sign-in from the same browser drops the old record, and nothing checked the superseded principal stops resolving. `replacing_a_session_takes_its_principal_out_of_the_index` covers it, and I confirmed it catches the regression by breaking the removal — two principals where one is expected.

src/internet_identity/src/storage.rs
```diff
@@ -2624,6 +2764,412 @@ impl<M: Memory + Clone> Storage<M> {
...
+            .map_err(StorageError::Browser)?;
+
+        // The whole of what the identity holds, not just this origin: a browser the
+        // registry gave up to make room for this one may hold sessions anywhere, and those
+        // have to go in the same write as the browser that held them.
+        let mut state = self.account_state(anchor_number);
+        if !state.contains_key(&origin) {
+            let held = self.account_state_for_origin(anchor_number, &origin);
+            state.insert(origin.clone(), Some(held));
+        }
```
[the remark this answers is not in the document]
sea-snake: Fixed by the same change as the eviction issue above: the write reads the one origin it names, and the whole-identity scan happens only when a browser was actually given up. A registry below `MAX_BROWSERS` cannot have dropped one, so it is answered without even reading the stored anchor.

src/internet_identity/src/storage/tests.rs
```diff
@@ -5314,3 +5467,806 @@ mod session_record_tests {
...
+        let (mut storage, anchor_number) = storage_with_anchor();
+
+        // Two sessions on the browser that will be given up, so a count that outlived its
+        // entry would be visible rather than indistinguishable from a fresh one.
+        storage
+            .create_session(params(anchor_number, 7, 1_000))
+            .unwrap();
+        storage
+            .create_session(params_at(anchor_number, 7, 1, 1_000))
+            .unwrap();
```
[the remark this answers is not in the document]
sea-snake: Correct — the second sign-in replaced the first and the browser held one session where the comment claims two. It signs in at another origin now, and the setup is asserted (`counts(…).0.get(&0) == Some(&2)`) so it cannot quietly become one again.

src/internet_identity/src/storage/storable/anchor.rs
sea-snake: Half agreed. You are right that naming the session cap ties the field to one consumer when it is just a count, so that goes.

Keeping the tail, because it is the fact that made this comment wrong once already — it used to say "Live sessions", which it is not. The count includes expired records and can read higher than what is live, and a reader given only "Stored sessions this anchor holds" has to work out that "stored" is load-bearing. It now reads:

    Stored sessions this anchor holds. Expiry removes a session with no write to
    observe, so this can read higher than what is live until a write prunes them.

src/internet_identity/src/storage/storable/session_record.rs
sea-snake: Agreed on the storable — `pub id: StorableSessionId`. Free to do, too: the struct is `#[cbor(map)]` with `#[n(6)]`, so the integer index is what reaches disk and the Rust field name never does. No migration. It is also the convention next door, where `StorableBrowser` calls its own identifier `id`.

The domain type keeps `session_id`: `SessionLocator`, the reclaim sort key and the principal index all name it that in contexts where a bare `id` would be ambiguous, and unlike the storable they are read across modules.

src/internet_identity/src/storage/account.rs
sea-snake: Agreed that `Record` is wrong, and it is gone from both — though not to `SessionKey`, which is unavailable: it already means the public key type in `internet_identity_interface`, and that is the concept most easily confused with a storage address.

So: `SessionRecordKey` → **`SessionLocator`**, because `Record` is wrong twice there — it is not a record and not a stable-structure key, it is the address of a session, and the codebase already has `StorableAccountLocator`. And `SessionRecord` → **`Session`**, because it *is* the session; the name was free.

The storable follows: `StorableSession` in `storable/session.rs`, beside `StorableBrowser`/`browser.rs`.

src/internet_identity/src/storage.rs
sea-snake: Agreed, combined:

    use crate::delegation::{
        self, calculate_session_seed_with_salt, canister_sig_principal, check_frontend_length,
    };

You filed the same thing on #4268 — one fix answers both. Left the `stats::event_stats` pair two lines below alone: identical shape, but pre-existing and outside this diff, so folding it in would put an unrelated change in a session PR.

src/internet_identity/src/storage.rs
sea-snake: The state you are pointing at is real, but not refusing on it — logging it instead.

A reference list keyed to an application number that resolves to nothing is *illegal*, and this was the place it disappeared silently: the `?` inside `filter_map` made the element vanish with no error and no log.

Returning `Err` here would be worse than the hole, though. It would block signing out *every* browser for that identity, including the ones whose lists do resolve and whose sessions do mint. And the rows it would be protecting are inert: the mint path fails on the same lookup (`lookup_session_with_principal` → `lookup_account_with_principal` → `account_key_of` resolves the same application number), so sessions in an unresolvable list are unusable as well as unrevocable.

So `account_state` stays the one private method its three callers share — the revoke sweep, the dropped-browser sweep and the cap reclaim — keeps returning the origin-keyed map, and now logs the invariant violation before carrying on, in the `"ERROR: … invariant violated: …"` shape the canister already uses in 17 places.

src/internet_identity/src/storage.rs
sea-snake: It is called two PRs up, by `prepare_account_session` — the scaffolding pattern this stack uses when a function arrives before its caller.

But the promise was not being kept, and that is the part worth acting on: the `#[allow(dead_code)]` count rose through the stack — 11, 11, 12, 18, 18 — and never came back down, so at least four allowances outlived the PR that was meant to remove them. An `allow` that survives its reason silences a genuinely dead function for good.

So each one is now removed by the PR that first calls it, and the notes name the caller rather than a distance in PRs, which is checkable. One finding out of that audit, which I have not acted on and want your call on: `account_principal_of` (this file, just above) has **no caller anywhere in the stack** — one test at this PR calls it and #4267 replaces that test, so from #4267 up it is genuinely dead and the allowance is hiding it.

src/internet_identity/src/storage.rs
sea-snake: Keeping the allocation as it is, and fixing the comment instead.

A gap costs nothing — ids need not be contiguous and the counter has room for 2^64 — while reserve-then-commit opens the failure the counter exists to prevent: two ceremonies in flight reserving the same id, or a commit that never happens. The id is an input to the session seed, so reissuing one would let a revoked session's identity be reached a second time, which is also why `allocate_session_id` refuses at the ceiling rather than saturating.

Your objection did catch a false claim, though. The comment said "After the checks that can refuse this ceremony, so a refused one does not burn an id", which is only mostly true — the write below can still refuse. Reworded to say gaps are acceptable and reissue is not, without claiming nothing downstream can refuse.

src/internet_identity/src/storage.rs
sea-snake: Taken. `verify_browser_keys` returns a `VerifiedBrowserKeys` and `CreateSessionParams` requires it.

Your premise holds exactly as stated: the params took bare `PublicKey`s, `create_session` handed them to `resolve_browser` which registers or advances a registry entry, and nothing in storage checked the proof — only `prepare_account_session` calling the verifier first, which is a caller remembering. The type has private fields and no other constructor, so holding one *is* the evidence; Rust privacy is per-module, so that is airtight within the crate rather than a convention.

One thing had to move to make the layering work: `CreateSessionParams` lives in `storage` and the verifier lived in `sessions::browser_key`, and storage must not depend upward on sessions. So `browser_key.rs` is now `crate::browser_key` — a self-contained P-256 verifier that knows nothing of sessions or storage, which either layer may depend on. It belonged there anyway.

`browser_description` stays a plain field: it is self-reported and verifies nothing.

MRmarioruci (approved):

PR #4267, opened 2026-08-22
feat(be): cap sessions per identity

sea-snake:
Design: #4224. Overview: #4230. #4266 stores sessions; this bounds them.

Sessions are per (identity, application, account, browser), so without a bound an identity's stored set grows with every app it ever signs in to.

**Five hundred stored records per identity, expired ones included.** Counting what is stored rather than what is live is what makes the cap enforceable: a session expires with no write anywhere, so no counter can follow the live set — something would have to decrement at the moment of expiry, and nothing runs then. An expired record holds its slot until something reclaims it, and because it is the first thing reclaimed, a held slot is never taken from a session in use.

**Reaching the cap reclaims rather than refusing.** The user is trying to sign in, and the only thing that could refuse them is internal bookkeeping. Reclaiming walks the identity's lists and takes dead sessions first, then live ones by `last_used + (last_used - created_at)`: how recently the session was used, extended by how long it stayed in service. The extension is the point, because recency alone gets the common case backwards — an app opened once and abandoned yesterday was touched more recently than an app in weekly use last opened three days ago. It needs no constant of its own, since the span it adds is bounded by the session's own 30-day lifetime.

**What the cap does not do.** It puts no bound on a flood of sign-ins: standing rises with use, so a party that can provoke sign-ins and keep refreshing them can outrank sessions an identity has left idle. What stands in the way is the ceremony — creating a session needs an access method, so a flood costs one authenticated sign-in per session. A stronger bound is deliberately not decided here.

**Enforced against a recount, not a counter.** A `session_count` on the anchor decides whether to run at all, so a sign-in below the cap reads nothing extra. At the cap, the pass counts what the lists actually hold and admits the sign-in against that number. Every path that removes a session decrements the counter, but it is one number maintained by seven call sites, and the recount is what makes a missed decrement cost an extra pass rather than a slot past the cap. It reads every list rather than a bounded prefix, because a truncated scan would undercount and the undercount would become the counter.

Tests: `session_creation_tests` (5), including 620 consecutive sign-ins never exceeding the cap, a drifted counter being corrected rather than denying a sign-in, and reclaiming across two lists taking only the sessions it selected and never their namesakes in another list.


src/internet_identity/src/storage.rs
```diff
@@ -2001,6 +2152,12 @@ impl<M: Memory + Clone> Storage<M> {
             }
         };
 
+        // Reclaiming before the session is admitted rather than after it: the stored set
+        // never sits above the cap, not even for the rest of this message.
+        if !self.ensure_session_slot(anchor_number, now_ns)? {
```
MRmarioruci: What happens when this browser already has a session for this account? This sign-in replaces that session, so it does not need an extra slot. At the cap, though, we reclaim other sessions before finding that out. This is not a deal-breaker cause you're almost at the limit anyways.. but if we want to be explicit that this trigger at exactly 500 we can adjust with a check

src/internet_identity/src/storage.rs
MRmarioruci: What happens if removed is larger than session_count here? Clamping to zero makes the counter lower than what is actually stored. Since this counter decides when we recount, the rows could then go over 500 before cleanup runs. Could we return an error or recount the rows instead of hiding the mismatch?

src/internet_identity/src/storage.rs
```diff
@@ -2812,6 +2945,12 @@ impl<M: Memory + Clone> Storage<M> {
             }
         }
 
+        // Before the session is added, because the room has to exist for it: dead sessions
+        // everywhere, and where that is not enough the least used give way down to the
+        // watermark. Nothing is stored yet, so a state still over the cap after this
+        // refuses in the write below rather than half-way through.
+        Self::reclaim_sessions(&mut state, now_ns);
```
MRmarioruci: We are scanning every account list on every sign-in. state is built with account_state above, then this counts all its sessions even when the counter is well below 500. Could we check anchor.session_count first and only load the full state when we hit the cap or drop a browser? The normal path would then only need this origin.

MRmarioruci (approved):

src/internet_identity/src/storage.rs
sea-snake: This is gone rather than checked for. The cap moved into the write gate and is gated on a write that grows the stored count, so a sign-in that only replaces this browser's own session has a delta of zero and never reclaims. Nothing is lost by not pruning there: a write that does not grow the count cannot breach a cap on growth.

src/internet_identity/src/storage.rs
sea-snake: Recounting rather than erroring — refusing a sign-in over storage's own bookkeeping would be the wrong answer for the user. The arithmetic is in `i64` now and falls back to counting the lists whichever end it fails to fit, so the counter self-corrects in both directions instead of only when it read high.

Worth saying that the state is not reachable today: every session write derives its delta from the lists it changed, so nothing drifts the counter under what is stored. `an_under_counting_anchor_is_corrected_rather_than_clamped` pins the behaviour anyway.

src/internet_identity/src/storage.rs
```diff
@@ -1779,6 +1883,32 @@ impl<M: Memory + Clone> Storage<M> {
...
+        let session_delta: i64 = validated.iter().map(|one| one.session_delta).sum();
+        let session_count = (session_delta != 0).then(|| {
+            let moved = stored_sessions.saturating_add_signed(session_delta as i32);
+            if moved > MAX_SESSIONS_PER_ANCHOR {
+                self.stored_session_count(anchor_number)
+                    .saturating_add_signed(session_delta as i32)
+            } else {
+                moved
+            }
+        });
```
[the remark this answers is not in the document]
sea-snake: Fixed, though not for the reason given. The clamp is unreachable: the counter is derived from the list diff on every session write, so nothing can leave it below what the lists hold, and the `u32::MAX` end needs a counter near four billion that the cap makes impossible. The sloppiest part was the `session_delta as i32` cast, which wraps rather than saturates. The arithmetic is `i64` now and recounts whichever end it fails to fit, which covers all three.

src/internet_identity/src/storage/tests.rs
```diff
@@ -5839,31 +5859,226 @@ mod session_creation_tests {
...
+                "the counter parted ways with the lists after {sign_in} sign-ins"
+            );
+        }
+    }
+
+    #[test]
+    fn the_session_cap_reclaims_to_the_watermark() {
+        let (mut storage, anchor_number) = storage_with_anchor();
+        let _session = storage
             .create_session(params(anchor_number, 7, 1_000))
```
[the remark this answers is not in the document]
sea-snake: Right on both counts, and the doc comment describing the watermark was sitting on the other test, which is what the duplication looked like from outside. Rewritten to earn its name: it signs in until the identity is exactly at the cap, then asserts the next sign-in leaves the watermark plus one. The drift case keeps its own test, and gained a counterpart for the other direction.

src/internet_identity/src/storage.rs
```diff
@@ -2389,7 +2526,7 @@ impl<M: Memory + Clone> Storage<M> {
                     );
                 }
                 if sessions_changed {
-                    session_delta += self.sync_session_index(
+                    self.sync_session_index(
```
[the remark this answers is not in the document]
sea-snake: It returns `()` now. Making the discard explicit would have left a value no caller can use — the count comes from each validated write's own `session_delta`.

src/internet_identity/src/storage.rs
sea-snake: Already done, and your description is nearly word for word what landed — this comment predates the push.

`create_session` hands the gate one origin via `account_state_for_origin`; the gate checks the counter first (`stored_sessions + added > MAX_SESSIONS_PER_ANCHOR`) and only then calls `sessions_to_reclaim`, which is what walks every list; the dropped-browser sweep is likewise conditional and skips even reading the stored anchor while the registry is below `MAX_BROWSERS`; and `sessions_added` reads only the origins the write names.

Worth saying that the same reasoning found more than a cost. The whole-map write you were pointing at was also why tracked-default eviction could never run on a sign-in: every stored origin counted as written, so `evictable_after` filtered every candidate out. There is a regression test for that now.

MRmarioruci (approved):

PR #4268, opened 2026-08-22
feat(be): sign a session to the II frontend at sign-in

sea-snake:
Design: #4224. Overview: #4230. This is the caller #4266 and #4264 were waiting for, so it removes their `#[allow(dead_code)]` annotations.

**`prepare_account_session` / `get_account_session`.** The first creates a session and signs its delegation to the II frontend's key; the second witnesses it. A separate pair from `prepare_account_delegation`, not an option on it: both mint, but one proves a live session and identifies the account by its principal while the other proves an access method and names the anchor outright. Merging them would mean one method with two authorizers and two argument shapes, and would drag the frontend's internal surface into the public API.

**Everything that can refuse does so before anything is written**, cheapest first. The account check leads — an account the identity does not hold is the one failure a caller can provoke, and returning it after the writes would leave a browser registered for a sign-in that never happened — so a request that was never going to succeed does not pay for two P-256 verifications on the way to being told so. Nothing is revealed by that order: `check_authz_and_record_activity` above is the auth guard. The browser proof follows, using the verifier from #4264, and `create_session` takes the `VerifiedBrowserKeys` it produces rather than two byte strings.

**The session credential is scoped to Internet Identity.** It exists to mint app delegations, which is an update call on this canister and nothing else, so it is signed with `targets = [id()]` and can be presented nowhere else. Leaving that to the II frontend would have left it to the party holding the session key, who can decline to add it. `permissions` stays absent for the same reason it is set on app delegations: minting is an update call, so a read-only session that could not make one could not sign in to an app at all — read-only travels on the app delegation instead.

**`get_account_session` tells three failures apart.** No stored session for that id is `NoSuchSession`; a session that is present with no signature for the asked-for key and expiration is `NoSuchDelegation`, because the remedy is to ask with the parameters that were signed rather than to sign in again; a seed that will not derive is the salt being unset, which is `InternalCanisterError`. An over-long origin is that too, rather than a rejected message the caller cannot read as a response.

**The request carries what the browser is, not a label for it.** `browser_description` holds the tokens the registry stores (#4242), and each token a client writes for itself — an unrecognised brand, an unrecognised system, and the hardware model — is bounded at 64 bytes. The named variants carry no text, so a description of nothing but those is within the limit whatever it says. Refused rather than truncated: a cut-off token would put a value in the record that no parser ever produced.

**Every later failure traps rather than returning.** On the IC, returning an error commits state and only a trap rolls the message back, so once the browser registration is written a failure has to trap or a caller could be told "no" and still have a browser enrolled.

**`valid_for`** is the lifetime the user chose at consent, clamped by the canister to between 10 minutes and 30 days. Every ceremony creates, so it always applies: the replacement's expiry is measured from the ceremony that made it, and no session is renewed in place.

Tests: `integration/sessions.rs` (18) drives the real ceremony — creating and verifying a session, a request for another identity refused, the registry cap dropping the least recently used and ending its sessions, the key proof's rejections at the endpoint, rotation keeping the entry, a retired key returning as a new browser, and two browsers each keeping the description it registered with.


src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,306 @@
...
+        current_device_key_signature,
+        next_device_key_signature,
+        permissions,
+        valid_for,
+    } = request;
+
+    check_authz_and_record_activity(identity_number)?;
+    check_frontend_length(&origin);
+    if device_name.len() > MAX_DEVICE_NAME_BYTES {
+        return Err(AccountSessionError::InternalCanisterError(
```
MRmarioruci, line 86: Maybe there is a better error for these instead of InternalCanisterError.. A long device name comes from the request, it's a validation error not a canister error.

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,306 @@
...
+            "device name exceeds the limit".to_string(),
+        ));
+    }
+    if !verify_device_keys(
```
MRmarioruci, line 90:
```suggestion
    if !device_keys_are_valid(
```

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,306 @@
...
+        })
+    })
+    .is_none()
+    {
+        return Err(AccountSessionError::NoSuchAccount);
+    }
+
+    let mut anchor = state::anchor(identity_number);
+    // A rotating browser presents the successor it announced, so both values are known.
+    let known_device = anchor.session_devices().iter().any(|device| {
```
MRmarioruci: Could resolve_session_device tell us whether it registered a new browser? We repeat its current/next-key matching here just to decide whether to archive the event. Returning registered together with the device ID and dropped IDs would keep that decision in one place.

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+        &current_browser_key,
+        &current_browser_key_signature,
+        &next_browser_key,
+        &next_browser_key_signature,
+        &session_key,
+    ) {
+        return Err(AccountSessionError::InvalidBrowserKey);
+    }
+
+    let now = time();
```
MRmarioruci, line 133: Note to agent: Move these below the account check on 129. Let validations stay on top, state writing after that, group things logically otherwise its harder to follow the code

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+            account_number,
+            session_id,
+        })
+    })
+    .ok_or(AccountSessionError::NoSuchSession)?;
+
+    let seed = session_identity(identity_number, &origin, account_number, &session)
+        .map_err(|_| AccountSessionError::NoSuchSession)?;
+    let signed_delegation = witness_session_delegation(&seed, &session_key, expiration)
+        .ok_or(AccountSessionError::NoSuchSession)?;
```
MRmarioruci: Maybe NoSuchDelegation?

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+
+    Ok(GetAccountSessionResponse { signed_delegation })
+}
+
+fn witness_session_delegation(
```
MRmarioruci:
```suggestion
fn get_session_delegation(
```

MRmarioruci: or just session_delegation

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+    );
+    Ok(seed)
+}
+
+/// The principal an app sees for this account. A session handle names the account by this
+/// and never by the numbers behind it, which are II's alone.
+///
+/// The account is read rather than reconstructed: a materialized default derives from
+/// `seed_from_anchor`, which only the stored list carries.
+fn account_principal(
```
MRmarioruci:
```suggestion
fn account_principal_for_origin(
```

src/internet_identity/src/storage.rs
```diff
@@ -102,8 +102,8 @@ use ic_stable_structures::{
 use identity_jose::jwk::Jwk;
 use internet_identity_interface::archive::types::BufferedEntry;
 
+use crate::delegation::calculate_session_seed_with_salt;
```
MRmarioruci: combine with import below?

src/internet_identity/internet_identity.did
```diff
@@ -1050,6 +1050,82 @@ type BrowserInfo = record {
     session_count : nat32;
 };
 
+type PrepareAccountSessionRequest = record {
+    identity_number : UserNumber;
+    origin : FrontendHostname;
+    account_number : opt AccountNumber;
+    // The II frontend's own key. The app never sees this chain's private key.
```
MRmarioruci:
```suggestion
    // The II frontend's own public key.
```

src/internet_identity/internet_identity.did
```diff
@@ -1050,6 +1050,82 @@ type BrowserInfo = record {
...
+    session_key : SessionKey;
+    // What this browser is, for the user's session list. Taken only where this sign-in
+    // registers a browser: an entry that is advanced keeps what it was registered with,
+    // so a browser reporting something else presents a key pair no entry holds and
+    // registers under its own.
```
MRmarioruci:
```suggestion
```

src/internet_identity/internet_identity.did
```diff
@@ -1050,6 +1050,82 @@ type BrowserInfo = record {
...
+    account_principal : principal;
+};
+
+type GetAccountSessionRequest = record {
+    identity_number : UserNumber;
+    origin : FrontendHostname;
+    account_number : opt AccountNumber;
+    session_key : SessionKey;
+    expiration : Timestamp;
+    // The session prepare_account_session created, named exactly rather than searched for.
```
MRmarioruci:
```suggestion
    // The session prepare_account_session created.
```

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+        next_browser_key,
+        current_browser_key_signature,
+        next_browser_key_signature,
+        permissions,
+        valid_for,
+        max_idle,
+    } = request;
+
+    check_authz_and_record_activity(identity_number)?;
+    check_frontend_length(&origin);
```
MRmarioruci: We shouldn't trap, let's return a proper error

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+            ),
+        };
+        sigs.get_signature_as_cbor(&inputs, Some(certified_assets.root_hash()))
+            .ok()
+    })
+    .map(|signature| SignedDelegation {
+        delegation: Delegation {
+            pubkey: ByteBuf::from(session_key.to_vec()),
+            expiration,
+            targets: None,
```
MRmarioruci: This session credential is only used to call II. Could we sign it with II as its target and return the same target here? The frontend adds that restriction in its next delegation, but someone holding the session key can omit that frontend-created hop.

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+        current_browser_key,
+        next_browser_key,
+        current_browser_key_signature,
+        next_browser_key_signature,
+        permissions,
+        valid_for,
+        max_idle,
+    } = request;
+
+    check_authz_and_record_activity(identity_number)?;
```
MRmarioruci, line 98: Should we record activity before validating the input?

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,376 @@
...
+    check_frontend_length(&origin);
+    if !browser_description_within_limits(&browser_description) {
+        return Err(AccountSessionError::InternalCanisterError(
+            "browser description exceeds the limit".to_string(),
+        ));
+    }
+    if !verify_browser_keys(
```
MRmarioruci: I think we should normalize both browser keys before passing them to storage. from_public_key_der accepts compressed and uncompressed encodings of the same P-256 key, but resolve_browser compares the original bytes. The same private key can therefore look like two different keys and pass the rotation check

src/internet_identity/src/sessions.rs
sea-snake: Keeping `InternalCanisterError`, and I do not think the variant is mislabelled here.

The limits are set deliberately high: any legitimate input, and a great deal more, passes. A 64-byte browser token is not reachable by a normal client, so turning it into a typed error would put a case in the Candid surface that the frontend then has to handle for nobody's benefit. `InternalCanisterError` is specifically the bucket for those unreachable edges.

The trap you flagged separately at `:101` is a different matter and is fixed — see that thread.

src/internet_identity/src/sessions.rs
sea-snake: Your point held — it returned a `bool` and read as if it did something rather than answered something — but it is moot now, and by your own suggestion: the thread on #4266 asked for `verify_browser_keys` to return a `VerifiedBrowserKeys` that `CreateSessionParams` requires, and that is what landed. It returns evidence, not a predicate, so a predicate name would be wrong.

src/internet_identity/src/sessions.rs
sea-snake: Stale, and no archive work — the duplication is already gone. `sessions.rs` had a `known_device` that re-scanned `anchor.session_devices()`; it no longer touches the registry at all, because `storage::create_session` calls `resolve_browser` itself.

The archive entry that duplication served is gone too, and it stays gone: the archive is not actively maintained, so nothing on the session path writes to it and `create_session` need not report whether it registered a browser.

src/internet_identity/src/sessions.rs
```diff
@@ -1,5 +1,295 @@
...
+        sigs.get_signature_as_cbor(&inputs, Some(certified_assets.root_hash()))
+            .ok()
+    })
+    .map(|signature| SignedDelegation {
+        delegation: Delegation {
+            pubkey: ByteBuf::from(session_key.to_vec()),
+            expiration,
+            targets: None,
+            permissions: None,
+        },
```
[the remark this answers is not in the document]
sea-snake: Half of this is a false positive, and the other half is a real gap that is now fixed — though not the way the report frames it.

**The escalation claim does not hold.** A read-only session cannot mint an unrestricted app delegation: `app_prepare_delegation` and `app_get_delegation` both derive their access with `DelegationAccess::from_read_only(session.read_only)`, so the delegation an app actually holds carries `permissions = "queries"` for a read-only session. Read-only travels on the app delegation, which is where the app's calls go.

**The session credential must stay update-capable.** Its whole job is to call `app_prepare_delegation`, which is an update call. Giving it `permissions = "queries"` would mean a read-only session could not sign in to an app at all.

**What was genuinely missing was scope.** That credential only ever calls Internet Identity, and it carried no `targets`. It is now signed with the II canister id (`ic_cdk::api::id()`) as its only target, so it cannot be presented anywhere else. `add_delegation_signature` gained a `targets` parameter for it, and both the signing and the witnessing side take the targets from one helper — the two messages have to match byte for byte or the lookup finds nothing.

src/internet_identity/src/sessions.rs
sea-snake: Agreed, reordered — and not just moved below the account check.

All the validations were already above the writes; what changed is their order among themselves. The account check goes first because it is the cheapest and the likeliest legitimate failure, so a request that was never going to succeed does not pay for two P-256 verifications on the way to being told so. The one reason not to reorder on aesthetics alone — that a cheap-but-revealing check tells a caller whether an identity holds an account before they have proven anything — does not apply, because `check_authz_and_record_activity` above is the auth guard.

The `now`/`valid_till`/`access` computation moved below both as well: it is neither validation nor writing and was sitting between them.

src/internet_identity/src/sessions.rs
sea-snake: Agreed, and one of its siblings is reclassified too.

`get_account_session` was returning `NoSuchSession` for three different facts. It now says:

- `:222` no stored session for that id — `NoSuchSession`, unchanged and correct.
- `:228` the session exists and its seed derives, but no signature was ever added for that `(session_key, expiration)` — that is a missing **delegation**. New `NoSuchDelegation` variant, because the remedies differ: this one is answered by asking with the parameters `prepare_account_session` signed, not by signing in again, which is where `NoSuchSession` sent the frontend.
- `:226` the seed failing to derive, which means the salt is unset — an internal failure wearing a caller error, so `InternalCanisterError`.

src/internet_identity/src/sessions.rs
sea-snake: Agreed — `get_session_delegation`. It follows the convention the analogous functions already set (`get_session_delegation` in `session_delegation.rs`, `get_account_delegation` in `account_management.rs`), and `witness_*` appeared nowhere else in the canister.

The same name in another module turned out not to be a clash: nothing imports the existing one unqualified — `main.rs` calls it `session_delegation::get_session_delegation(...)` — so this needed no import or call-site changes.

Worth flagging what that near-collision is really about, since it will come up again: two unrelated things are both called a "session delegation" — the delegation a browser holds for Internet Identity itself (anchor-seeded, 30 days) and this stack's app sessions (account-seeded, per-origin, per-browser, revocable). `identity_delegation` is the name that fits the first. Not renaming it now, because `prepare_session_delegation` and `get_session_delegation` are live Candid methods gated by `check_candid_interface_compatibility`, so it needs aliases and a frontend pass; and the `b"session-delegation"` domain separator can never be renamed at all, since changing it changes every derived principal.

src/internet_identity/src/sessions.rs
sea-snake: Agreed on the shape, with the `get_` prefix kept: `get_account_principal_for_origin`.

`for_origin` says the thing a reader most needs to know — an account's principal differs at every origin, and that is the property the whole design rests on. `get_` says it reads the account from storage rather than deriving from what the caller already holds, which is why it returns a `Result` that can be `NoSuchAccount`. One call site.

src/internet_identity/src/storage.rs
sea-snake: Agreed, combined — same fix as your thread on #4266, which is where the change landed:

    use crate::delegation::{
        self, calculate_session_seed_with_salt, canister_sig_principal, check_frontend_length,
    };

src/internet_identity/internet_identity.did
sea-snake: Taken as written. "Own key" did not say which half, and once it says public, the sentence about the private key answers a question nobody asked.

src/internet_identity/internet_identity.did
sea-snake: Agreed in substance — kept the first sentence, dropped the rest.

It now reads "What this browser is, for the user's session list." The three lines about advanced entries and key pairs no entry holds were canister mechanics, not something a client integrating against this interface acts on.

src/internet_identity/internet_identity.did
sea-snake: Taken as written. "Named exactly rather than searched for" was why *we* chose an id over a lookup — our reasoning, not a client's instruction.

src/internet_identity/src/sessions.rs
sea-snake: Agreed, and the fix goes further than this call site.

`check_frontend_length` stays exactly as it is for its existing callers, and there is a second helper that answers instead of trapping. It is used only where the response already carries a variant for the unreachable edges, so nothing else in the tree moves: `prepare_account_session` and `get_account_session` here, and `prepare_account_delegation` and `get_account_delegation` in `account_management.rs`. The four in `vc_mvp.rs` and the one in `delegation.rs` are untouched.

Storage stopped trusting its callers for the same bound, which is the part worth having. `create_session`, the write gate, `account_state_for_origin` and `lookup_application_number_with_origin` had no check at all — an over-long origin was kept out only because this endpoint checked first, which is the pattern this stack has otherwise been removing. There is one choke point, so it is one check: the only place an origin becomes a stored one is where `validate_account_reference_list` mints an application for it, and that now refuses with a new `StorageError::OriginTooLong`. Every stored origin passes through there, which makes the endpoint check redundant rather than load-bearing.

One consequence worth naming: `should_handle_very_long_origins_with_sha256` asserted that a 20,000-byte origin round-trips through storage, which is exactly what no longer holds. It is now `should_handle_an_origin_at_the_limit_with_sha256` — the index is keyed by the origin's hash, which is what that test was really about — with a new test that an origin past the limit is refused rather than stored.

src/internet_identity/src/sessions.rs
sea-snake: Agreed, and this was the real gap in the pair of findings on this line — done.

The session delegation is signed with `targets = [ic_cdk::api::id()]` and returns the same, so it can only be presented to Internet Identity. Your reason is the one that decides it: the frontend adding the restriction is a restriction the party holding the session key can decline to add.

Two things that had to come with it. `add_delegation_signature` gained a `targets` parameter — the signature is added when the session is created and witnessed when the credential is fetched, and the two messages must agree byte for byte, so adding targets on one side alone would make every fetch return nothing. And both sides take the targets from one helper for that reason.

`permissions` deliberately stays absent, which also answers the ZeroPath finding on `:258`: minting an app delegation is an update call, so a read-only session that could not make one could not sign in to an app at all. Read-only travels on the app delegation, which `DelegationAccess::from_read_only` already puts there. App delegations stay unscoped — an app calls whatever canisters it likes.

src/internet_identity/src/sessions.rs
sea-snake: Leaving it — this is existing shared behaviour and moving it would make it mean something less true.

What `check_authz_and_record_activity` writes is about the access method that authenticated, not the request: `last_used` on the passkey/OpenID/email credential, which is what the access page shows, and the active-anchor aggregation behind the metrics. The passkey *did* authenticate at that moment whether or not the rest of the request parsed, so stamping it below validation would make `last_used` mean "last used on a request that also parsed" and would under-count active anchors.

It also cannot simply move: that call does the authorisation and returns the `IIDomain`, and 12 other endpoints in `main.rs` call it first thing. The cost is one anchor write on a refused request, by a caller who already holds a valid access method.

src/internet_identity/src/sessions.rs
sea-snake: Good catch, and the fix is a refusal rather than normalization — which makes two-entries-from-one-key impossible instead of merely unprofitable.

WebCrypto cannot produce anything but the uncompressed SEC1 point: `generateKey({name:"ECDSA", namedCurve:"P-256"})` plus `exportKey("spki")` writes `0x04 || X || Y` per RFC 5480, and the API offers no compressed option at all. So the frontend already sends exactly one thing, and the backend refuses the other instead of parsing it. `verify_browser_keys` now accepts a SPKI only where its point is the uncompressed form.

Two things worth saying while we are here. Both encodings *are* valid DER, so "we only accept DER" would have been the wrong answer to this. And nothing stored today can be holding a compressed key, because only WebCrypto clients exist — there is nothing to migrate. There is a test that a compressed key is refused although `from_public_key_der` accepts it, and it fails if the check is removed.

MRmarioruci (approved):

src/internet_identity/src/storage.rs
```diff
@@ -3367,8 +3373,6 @@ impl<M: Memory + Clone> Storage<M> {
     /// the only place that capability is handed out — a caller that holds one has been
     /// through the check above.
     pub fn read_account(&self, key: &AccountKey) -> Option<Account> {
-        check_frontend_length(&key.origin);
-
         let reference = self
```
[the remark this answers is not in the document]
sea-snake: Correct, and thank you — this is a real gap. Moving the origin bound to the write path's choke point left the reads behind: `read_account` never reaches the point where an application is minted, because nothing is minted. Confirmed that `get_default_account` and the attribute endpoints do no length check of their own, so an over-long origin flowed through, absence normalised to the derived default, and the canister handed out a principal for an origin it could never persist.

Fixed differently from your suggestion, though. Rather than restore the trap, the read now answers nothing: **an origin too long to store has no accounts under it**, so `read_account` returns `None`. That keeps storage answering rather than rejecting — the line the rest of this change is drawing — needs no new error variant, and holds for callers not yet written rather than for the ones that remembered.

Visible effect: `get_default_account` on an over-long origin answers `InternalCanisterError` where it used to trap. `an_origin_past_the_limit_holds_nothing_to_read` covers it, and also asserts that an origin *at* the limit still derives its default, so the bound is what separates them rather than the read path having stopped working. Verified by removing the guard — the test fails.

src/internet_identity/src/storage.rs
```diff
@@ -3383,8 +3387,6 @@ impl<M: Memory + Clone> Storage<M> {
         anchor_number: AnchorNumber,
         origin: &FrontendHostname,
     ) -> Vec<Account> {
-        check_frontend_length(origin);
-
         self.account_references_for_origin(anchor_number, origin)
```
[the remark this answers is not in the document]
sea-snake: Same gap, same fix — see the thread on `read_account` just above. `list_accounts` now returns an empty vec for an origin past the limit, so `get_accounts` and `mcp_get_accounts` answer empty rather than echoing a synthetic oversized origin. The two are one defect and the test covers both.

src/internet_identity/tests/integration/sessions.rs
```diff
@@ -0,0 +1,603 @@
...
+            origin: ORIGIN.to_string(),
+            account_number: None,
+            session_key: ByteBuf::from(vec![1; 32]),
+            expiration: prepared.expiration,
+            session_id: prepared.session_id,
+        },
+    )?
+    .unwrap();
+
+    verify_delegation(
```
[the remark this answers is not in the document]
sea-snake: Right, and worse than you put it — the test could not fail for the property it exists to guard. `verify_delegation` builds the message it checks *from the delegation the endpoint returned*:

```rust
if let Some(targets) = &signed_delegation.delegation.targets { ... }
```

so if both the signing and the witnessing side dropped the targets, the reply would carry `None`, the helper would build a message without them, and the signature would verify. Green test, unscoped credential.

Both fields are now asserted outright. Verified by making `session_delegation_targets()` return an empty vec: the test fails with "the session credential must be usable only against Internet Identity".

src/internet_identity/tests/integration/sessions.rs
```diff
@@ -0,0 +1,603 @@
...
+        principal_1(),
+        session_request_from(identity_number, &victim),
+    )?
+    .unwrap();
+
+    let attacker = BrowserKey::new(2);
+    let mut request = session_request_from(identity_number, &attacker);
+    request.next_browser_key = victim.successor().public_key();
+    request.current_browser_key_signature =
+        attacker.sign(&request.session_key, &request.next_browser_key);
```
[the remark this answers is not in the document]
sea-snake: Correct — it passed for the wrong reason. Only `current_browser_key_signature` was re-signed, so `next_browser_key_signature` was still made by `attacker.successor()`, the key proof failed first, and the assertion on `InvalidBrowserKey` was satisfied without `resolve_browser` ever being reached.

Re-signed with `victim.successor()` as you suggest, which a test can do and an attacker cannot — that is exactly why the two checks are layered. The doc comment was describing the wire-reader threat, which belongs to `should_refuse_a_successor_the_caller_cannot_prove`; it now says what this test is actually about.

One correction to the finding, for the record: `SuccessorAlreadyInUse` was not wholly uncovered — `should_refuse_a_successor_another_browser_holds_even_when_proven` already exercised it, which I confirmed by deleting the check and watching both tests fail. So this test was the redundant one rather than the only one, though your reading of why it passed was right.

src/internet_identity/src/sessions.rs
```diff
@@ -0,0 +1,406 @@
...
+    .map_err(|err| match err {
+        // Told apart from the rest because the browser can act on it: it is the only
+        // party holding the successor that does resolve.
+        StorageError::Browser(BrowserError::StaleBrowserKey) => {
+            AccountSessionError::StaleBrowserKey
+        }
+        StorageError::Browser(_) => AccountSessionError::InvalidBrowserKey,
+        // The account was checked above, so anything left is a broken storage invariant
+        // rather than a request this caller could have got wrong.
+        err => AccountSessionError::InternalCanisterError(err.to_string()),
```
nmattia, line 168: Small nit: might be a good idea to enumerate all the errors so that new variants don't necessarily fall in the catch-all

src/internet_identity/src/sessions.rs
```diff
@@ -0,0 +1,406 @@
...
+        identity_number,
+        origin,
+        account_number,
+        session_key,
+        expiration,
+        session_id,
+    } = request;
+
+    check_authorization(identity_number)?;
+    frontend_length_within_limit(&origin).map_err(AccountSessionError::InternalCanisterError)?;
```
nmattia, line 211: why is this an internal canister error?

src/internet_identity/src/sessions.rs
```diff
@@ -0,0 +1,406 @@
...
+            account_number,
+            session_id,
+        })
+    })
+    .ok_or(AccountSessionError::NoSuchSession)?;
+
+    // Not `NoSuchSession`: the session was just read. A seed that will not derive means
+    // the salt is unset, which is a canister that never initialised rather than anything
+    // this caller named.
+    let seed = session_identity(identity_number, &origin, account_number, &session)?;
```
nmattia, line 230: nit: would it make sense for `session_identity` to be infallible? or at least to differentiate the missing salt from other errors? thinking there may be other errors added later on in which case a failure here won't necessarily mean "salt not set"

src/internet_identity/src/sessions.rs
```diff
@@ -0,0 +1,406 @@
...
+/// It exists to mint app delegations, which is an update call on this canister and
+/// nothing else. Scoping it here rather than leaving it to the frontend to add is the
+/// difference between a restriction and a request: the party holding the session key is
+/// the party that would have to add it, and can decline to.
+///
+/// `permissions` stays absent for the same reason it is set on app delegations: minting
+/// is an update call, so a read-only session that could not make one could not sign in
+/// to an app at all. Read-only travels on the app delegation this credential mints.
+fn session_delegation_targets() -> Vec<Principal> {
+    vec![ic_cdk::id()]
```
nmattia, line 253: is this ever not this canister's ID? would it make sense to inline it?

src/internet_identity/src/sessions.rs
```diff
@@ -0,0 +1,406 @@
...
+/// and never by the numbers behind it, which are II's alone.
+///
+/// The account is read rather than reconstructed: a materialized default derives from
+/// `seed_from_anchor`, which only the stored list carries.
+fn get_account_principal_for_origin(
+    anchor_number: AnchorNumber,
+    origin: &FrontendHostname,
+    account_number: Option<AccountNumber>,
+) -> Result<Principal, AccountSessionError> {
+    let salt = storage_borrow(|storage| storage.salt().copied()).ok_or_else(|| {
```
nmattia, line 329: nit: I've seen `state::salt()` used as well, would it make sense to be consistent and either use `storage_borrow()...` everywhere or alternatively use `state::salt()` everywhere?

PR #4269, opened 2026-08-22
refactor(fe): one queue for authorization-bearing requests

sea-snake:
Design: #4224. Overview: #4230. Groundwork for #4273, which is the second caller.

The delegation handler serialises its own requests so that two consent screens cannot race. The session handler coming in #4273 also authorizes, and it has to share that queue rather than keep one of its own — two queues would let one request paint over the other's screen.

Lifts the queue out of the delegation handler into `serialize.ts` as `serializeAuthorizationRequest`. No change to what the delegation handler does with it.

No tests: `handleDelegationRequest` has no unit coverage today, and the extracted function is a move of the existing queue. Its first test arrives with the second caller in #4273.


MRmarioruci (approved):

src/frontend/src/lib/stores/channelHandlers/serialize.ts
```diff
@@ -0,0 +1,17 @@
...
+ * flow, the authorized account — so a dapp sending requests in parallel could otherwise
+ * race them against each other and have the user approve a screen naming one origin
+ * while another is answered.
+ */
+let queueTail: Promise<unknown> = Promise.resolve();
+
+export const serializeAuthorizationRequest = <T>(
+  run: () => Promise<T>,
+): Promise<T> => {
+  const next = queueTail.then(run);
```
MRmarioruci, line 14: Are we resetting the authorization state between queued requests? The queue stops them from running together, but authorizedStore stays set after the first one. The next request can then reuse that approval instead of waiting for the user, and #4273 adds session requests to this same queue. Could setRequestContext clear the previous authorization and flow?

MRmarioruci (approved):

src/frontend/src/lib/stores/channelHandlers/serialize.ts
sea-snake: No, and deliberately — these stores are UI state rather than request state.

`authorizationStore` and `authorizedStore` describe *this window*: the origin on screen, the flow chosen, the account and duration the user picked. A request doesn't get its own copy of them; it reads what the window currently shows. So nothing here is stale or leaking, and clearing them from inside a request handler would have a request reach in and erase what the user is looking at.

The intent is that the user is taken through one flow, makes their choices there, and subsequent requests **read those decisions** rather than prompting for different ones. A second request riding the first approval is the design.

Two things worth knowing, because they're what makes it load-bearing rather than incidental: the attributes handlers depend on exactly this — six `waitForStore(authorizedStore)` sites and no `setRequestContext` of their own, so they ride the delegation's approval — and `delegation.ts` waits on `attributeConsentResultStore` before re-reading the approval. The sharing between the two is intentional.

PR #4270, opened 2026-08-22
feat(fe): describe the browser a session was created from

sea-snake:
Design: #4224. Overview: #4230. Nothing imports this until the sign-in in #4273.

Settings shows which browsers hold a session so the user can end one they do not recognise, and each entry needs to be something they *will* recognise. The client is the only party that can see what it is running on, so it is the one that has to say.

`describeBrowser` resolves the four things the record holds (#4242) — a brand, an operating system, a form factor, and the hardware model where one is reported — rather than composing a label out of them.

**Resolving rather than naming is the point.** A composed label fixes the wording at the moment of sign-in: a browser that registered before "Chrome OS" became "ChromeOS" would read the old name for as long as its record existed, and the agent it was parsed from is not kept, so nothing could re-derive it. With tokens the wording lives where the list is read (#4249), so renaming a product — or teaching this table to recognise a browser it used to miss — reaches every stored record instead of only the browsers that sign in afterwards.

**Two sources, because neither covers the field alone.** The user agent carries the brand and the system on every engine, and its table stays ordered most specific first, because every later token appears inside the earlier ones' agents: matching `Chrome` first would claim every Chromium-based browser, and every Chromium agent carries `Safari/` as well. Client hints are Chromium-only — Safari and Firefox expose nothing at all — so they supply only what no agent can: the hardware model, which a reduced Android agent reports as the literal string `K`, and a form factor the browser states rather than one inferred. Only the plural `formFactors` is asked for; the singular shape Chromium shipped first is not, because only current browsers are a concern. A stated `Watch`, `XR`, `Automotive` or `EInk` resolves to `Unknown`: the canister names desktops, mobiles and tablets, and calling a watch a phone is worse than naming it as none of them.

**A browser that hides reads as the one it imitates.** Brave ships a plain Chrome agent on purpose and strips the hints that would give it away, so it reads as Chrome — which is what the agent says, and the honest answer rather than a gap. Asking `navigator.brave.isBrave()` would name it, at the cost of a check per browser this frontend has heard of, which is the hardcoded list this table exists to avoid. A browser that does name itself in its agent, Vivaldi or DuckDuckGo, travels as the token it gave rather than resolving to Chrome, without needing a row of its own.

An unrecognised browser or system is carried as the token that arrived, capped to what the canister accepts, because that row is the one this list exists to surface. `maxTouchPoints` is consumed here rather than stored: an iPad in desktop mode sends a Mac agent and exposes no hints, so it is the only thing that tells it from a Mac.

Tests: `describeBrowser.test.ts` (37) — the 20 real user-agent strings, one per rung of both ladders, including the iPad and Mac pair separated only by `maxTouchPoints`; the model used when present and absent when empty or refused; a stated form factor, the four that resolve to `Unknown`, and a tablet stated beside one of them; a browser that hides itself reading as the one it imitates; and an unrecognised token capped to 64 bytes.


MRmarioruci (approved):

src/frontend/src/lib/stores/channelHandlers/describeBrowser.ts
```diff
@@ -0,0 +1,82 @@
+/**
```
MRmarioruci, line 1: Maybe this fits better under lib/utils instead of here?

src/frontend/src/lib/utils/describeBrowser.ts
```diff
@@ -0,0 +1,140 @@
...
+          formFactor?: string;
+        }>;
+      };
+    }
+  ).userAgentData;
+  if (data === undefined) {
+    return {};
+  }
+  try {
+    const high = await data.getHighEntropyValues?.(["model", "formFactors"]);
```
MRmarioruci, line 209: I think the older API was formFactor, so maybe we can add that as well for support?

src/frontend/src/lib/utils/describeBrowser.ts
```diff
@@ -0,0 +1,140 @@
...
+  if (/Linux|X11/.test(agent)) return { Linux: null };
+  return { Other: capped(agent) };
+};
+
+const formFactorOf = (
+  agent: string,
+  system: OperatingSystem,
+  hints: { mobile?: boolean; formFactors?: string[] },
+): FormFactor => {
+  if (hints.formFactors?.includes("Tablet") === true) return { Tablet: null };
```
MRmarioruci, line 162: Watch, XR, Automotive, or EInk are also formFactors we could get, but I dont think we need to care about those, maybe just add a comment above documenting the decision to ignore these form factors?

src/frontend/src/lib/utils/describeBrowser.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Chromium-only — Safari and Firefox expose nothing — so they are used for the two things
+ * an agent cannot give: the hardware model, and a form factor the browser states rather
+ * than one inferred from its agent.
+ */
+
+/** The canister refuses a longer token, so a resolver never offers one. */
+const MAX_BROWSER_TOKEN_BYTES = 64;
+
+/** Ordered most specific first: every later token also appears in the earlier ones' agents. */
+const BRANDS: [RegExp, BrowserBrand][] = [
```
MRmarioruci, line 25: Should we do all of this manually ourselves or maybe use a library like Bowser?

src/frontend/src/lib/utils/describeBrowser.ts
```diff
@@ -0,0 +1,82 @@
+/**
```
sea-snake: Agreed, moved to `$lib/utils/describeBrowser.ts` with its test. Every sibling in `channelHandlers/` imports `Channel`/`JsonRequest` and handles a JSON-RPC method; this one parses a user agent and returns a `BrowserDescription`. It also has two callers now and one — the devices page — is not a channel handler at all.

Two things came out of looking at it while there.

The `Vivaldi` and `DuckDuckGo` rows are gone. They existed because both carry `Chrome/` and would otherwise be *labelled Chrome* — two arbitrary picks from a long list. A browser outside the six is now named by the product token it appends, one rule covering Vivaldi, Yandex and anything this frontend has never heard of. Brave and Arc ship Chrome's agent byte for byte on purpose, so they still read as Chrome, which is the honest answer. Versions are dropped: the canister fixes a description at registration, so one captured here would sit frozen at whichever build first signed in.

The fallback was worse than the special cases. An unrecognised browser's *name* was the entire user agent capped at 64 bytes, and the operating system's was too — which is how a test ended up asserting `curl/8.4.0` as an OS. The system now comes from the parenthesised block where an agent states it.

Tests rebuilt around what is under test rather than one six-wide positional tuple, now covering a fork that names itself, one that hides, and a platform none of the seven matches.

MRmarioruci (approved):

src/frontend/src/lib/utils/describeBrowser.ts
sea-snake: Right that it was the older name, and it turns out the code was half-committed to it: the singular is read at the fallback but never *requested* — `getHighEntropyValues` is only ever asked for `["model", "formFactors"]`, so `high.formFactor` could never arrive.

Rather than add it to the request: we only care about newer browsers, so the older spec isn't a concern. The dead fallback and its comment are gone, leaving `formFactors: high?.formFactors` and one spec, the current one.

src/frontend/src/lib/utils/describeBrowser.ts
sea-snake: We do get them, and it turned out to matter more than "don't care" — so they're now mapped to `Unknown`.

The canister's variant is `Desktop; Mobile; Tablet; Unknown`, so none of the four has a home. Left to fall through they took whatever `mobile` said: a watch reports `mobile: true` and read as **Mobile**, automotive and EInk report `mobile: false` and read as **Desktop**. They're neither, so `Unknown` is the honest answer.

The check sits after the `Tablet` hint, so a device stating both — an e-ink tablet — keeps the variant that exists. Four tests cover the mapping and one covers that precedence.

src/frontend/src/lib/utils/describeBrowser.ts
sea-snake: Keeping it here, because Bowser answers a different question than this file asks.

It returns free-form strings — `"Chrome"`, `"Microsoft Edge"`, `"macOS"` — while everything here has to land on the canister's closed variants (`BrowserBrand`, `OperatingSystem`, `FormFactor`) under a 64-byte token cap. The mapping layer, which is most of this file, survives either way, and a translation table from its strings to our variants is a second place to drift: a rename upstream becomes an unmatched case with no compile error. It also reads no client hints, so the model and stated form factor — the half a user agent cannot supply — stays ours regardless.

The libraries are also far larger than what we need: Bowser is 258KB unpacked for what two short regex tables and a token scan do here, and the frontend ships inside the II wasm, which CI caps.

src/frontend/src/lib/utils/describeBrowser.ts
```diff
@@ -0,0 +1,231 @@
...
+};
+
+/** Truncated on a character boundary, because the cap the canister enforces is in bytes. */
+const capped = (token: string): string => {
+  const encoder = new TextEncoder();
+  let capped = token;
+  while (encoder.encode(capped).length > MAX_BROWSER_TOKEN_BYTES) {
+    capped = capped.slice(0, -1);
+  }
+  return capped;
```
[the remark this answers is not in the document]
sea-snake: The mechanism is right: the loop drops one UTF-16 code unit at a time, so chopping the low half of a surrogate pair leaves a lone high surrogate that `TextEncoder` emits as U+FFFD. 61 ASCII bytes plus an emoji is 65, one slice reaches 64, the loop exits, and the token keeps a replacement character.

Not taking it. `capped` guards two values: the brand token parsed out of a user agent, and the client-hint model. Neither carries astral characters, so a code-point loop is cost for a case that does not arise.

src/frontend/src/lib/utils/describeBrowser.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+  ).userAgentData;
+  if (data === undefined) {
+    return {};
+  }
+  try {
+    const high = await data.getHighEntropyValues?.(["model", "formFactors"]);
+    return {
+      mobile: data.mobile,
+      formFactors: high?.formFactors,
```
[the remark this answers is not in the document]
sea-snake: The contradiction is real, and the PR description was the stale half — corrected rather than the code.

Requesting and normalising the singular was considered and declined: only current browsers are a concern, so the shape Chromium shipped first is not supported, and the fallback that read it was dead anyway — `getHighEntropyValues` was only ever asked for `["model", "formFactors"]`, so `high.formFactor` could never arrive. It is gone along with the claim that it was covered.

The description now also states what does happen to the exotic factors, which is the part worth being explicit about: a stated `Watch`, `XR`, `Automotive` or `EInk` resolves to `Unknown`, because the canister names desktops, mobiles and tablets and calling a watch a phone is worse than naming it as none of them. Test count corrected too — 37, not 27.

src/frontend/src/lib/utils/describeBrowser.ts
```diff
@@ -0,0 +1,231 @@
...
+    return { mobile: data.mobile };
+  }
+};
+
+export const describeBrowser = async (): Promise<BrowserDescription> => {
+  const agent = navigator.userAgent;
+  const hints = await highEntropyHints();
+  const os = systemOf(agent, navigator.maxTouchPoints);
+  return {
+    brand: brandOf(agent),
```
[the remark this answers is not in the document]
sea-snake: Right that the PR and the code disagreed; wrong about which one was wrong.

Reading Brave as Chrome is deliberate, and the test says so: "Brave ships Chrome's agent byte for byte, deliberately. There is no token to find and nothing to report but Chrome — which is the honest answer, not a gap."

`navigator.brave.isBrave()` would name it, at the cost of a vendor check per browser this frontend has heard of — which is the hardcoded list this table was rewritten to get rid of. A browser that names itself in its agent, Vivaldi or DuckDuckGo, already travels as its own token without needing a row.

So the description loses the `isBrave()` claim and the "Brave named where its agent does not" line, and states what the code does instead.

PR #4271, opened 2026-08-22
feat(fe): hold a browser key and rotate it on every sign-in

sea-snake:
Design: #4224. Overview: #4230. Nothing imports this until the sign-in in #4273.

The canister identifies a browser by a key it proves possession of (#4264), so the frontend has to hold one, and rotate it, so a key captured from one sign-in cannot claim that browser at the next.

A non-extractable P-256 keypair in IndexedDB, per identity, plus a successor generated alongside it. A sign-in signs with the current key and announces the successor; the successor is promoted only once the canister has confirmed the sign-in.

**Promoting after confirmation rather than before is what survives a lost response.** If the reply never arrives, the browser still holds the key the canister has, so its next attempt presents the same one rather than a successor the canister never saw — which the canister would treat as an unknown browser and enrol as a second entry for the same machine.

**Sign-ins are serialised with a web lock**, because two at once would leave whichever wrote last holding a key the canister never accepted. Where the Web Locks API is missing the calls run unserialised: not blocking sign-in on it is the deliberate trade, and the window is one concurrent sign-in on the same device.

**A browser that no longer matches what it registered as signs in as a new one.** The description it reported (#4270) is stored beside the key pair, and compared before anything is sent. A registered entry keeps the description it was created with — the canister ignores what a sign-in reports once an entry is being advanced — so a browser whose brand, system, form factor or model has changed would otherwise keep rotating an entry describing something it no longer is. Where it differs, this presents a fresh key pair, which no entry holds and which therefore registers under its own.

Deciding here rather than being told keeps that a purely local operation: nothing has been sent when the comparison happens, so the entry left behind is untouched and a retry hits the same state and takes the same branch. The description is written with the key pair and never on its own, so the comparison is always against what the canister was actually sent.

The private key never leaves IndexedDB and never appears in a delegation chain.

Tests: `browser-key.store.test.ts` (21), including the successor promoted only after the call succeeds, a lost response leaving the current key in place, concurrent sign-ins serialised, the store working with the lock API absent, and a changed description signing in on a fresh key pair while an unchanged one keeps rotating.


src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,160 @@
...
+const storageKey = (identityNumber: bigint): string =>
+  identityNumber.toString();
+
+const generate = (): Promise<CryptoKeyPair> =>
+  crypto.subtle.generateKey({ name: "ECDSA", namedCurve: "P-256" }, false, [
+    "sign",
+    "verify",
+  ]) as Promise<CryptoKeyPair>;
+
+const read = async (
```
MRmarioruci, line 58: Maybe we shouldn't mix errors with a missing key here. I would do undefined for no key found and return something else in case we wanna handle this with a retry..

src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,160 @@
...
+    return await idbGet<BrowserKeyRecord>(
+      storageKey(identityNumber),
+      BROWSER_KEY_STORE,
+    );
+  } catch {
+    return undefined;
+  }
+};
+
+const write = async (
```
MRmarioruci, line 71: Should the initial key write be allowed to fail silently? If the key isn’t persisted, this sign-in can still register a browser, but the next sign-in has no key and registers another one. Repeating that can fill the 20-browser limit and eventually evict other browsers’ sessions. Maybe we could have the first write be required before calling the canister, while the successor write after acceptance can stay best-effort.

src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,160 @@
...
+};
+
+export interface BrowserProof {
+  publicKey: Uint8Array;
+  nextPublicKey: Uint8Array;
+  signature: Uint8Array;
+  /** By the successor itself, so a key the browser does not hold cannot be announced. */
+  nextSignature: Uint8Array;
+  /** Rotates to the successor. Called once the canister has accepted the sign-in. */
+  accept: (deviceId: number) => Promise<void>;
```
MRmarioruci, line 110: Is there a specific reason for having the caller confirming that the canister accepted the proof? If not maybe we could have this in withBrowserProof and it owns the whole thing?

src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,286 @@
...
+ * it. It is replaced at every sign-in, so a copy of it taken off disk stops working as soon
+ * as this browser signs in again.
+ */
+interface BrowserKeyRecord {
+  keyPair: CryptoKeyPair;
+  /** The successor announced at the last sign-in, kept from before the call until that
+   *  sign-in is known to have been accepted. The canister reaches this browser's entry
+   *  only through the successor it announced, so losing this key while the canister kept
+   *  it would leave the browser unable to prove it is itself ever again. */
+  announced?: CryptoKeyPair;
```
MRmarioruci:
```suggestion
  announced_successor?: CryptoKeyPair;
```

src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,286 @@
...
+  one: BrowserDescription,
+  other: BrowserDescription,
+): boolean =>
+  token(one.brand) === token(other.brand) &&
+  token(one.os) === token(other.os) &&
+  token(one.form_factor) === token(other.form_factor) &&
+  (one.model[0] ?? "") === (other.model[0] ?? "");
+
+/**
+ * The record to sign in with, which is a fresh one where this browser no longer matches
```
MRmarioruci:
```suggestion
 * * Returns the stored record unless the browser description changed; in that case, rotates
 * to a fresh record so this sign-in registers as a new browser.
```

src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,286 @@
...
+ * what it registered as.
+ *
+ * A registered entry keeps the description it was created with, so a browser reporting
+ * something else is one the canister has not seen. Rather than ask for an entry to be
+ * changed, this presents a key pair no entry holds, which registers under its own — and
+ * because nothing has been sent yet, a browser that gets this far has not disturbed the
+ * entry it is leaving behind.
+ */
+const forDescription = async (
```
MRmarioruci, line 153:
```suggestion
const getOrRotateForDescription = async (
```

src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,286 @@
...
+  identityNumber: bigint,
+  description: BrowserDescription,
+): Promise<BrowserKeyRecord | undefined> => {
+  const stored = await read(identityNumber);
+  if (
+    stored?.description === undefined ||
+    sameDescription(stored.description, description)
```
MRmarioruci, line 160: Could the key chain decide when this becomes a new browser instead of the description? If the stored successor still matches, it is the same browser. If two copies announce different successors, the chain forks and the canister can register a new browser then. A description change alone is not reliable, and copied profiles can also have the same description. We could keep the key here and update the description when the canister accepts a valid rotation.
Do you think this would work?

src/frontend/src/lib/stores/browser-key.store.ts
sea-snake: Declining, though the two facts really are different.

Their consequence is not: a browser that cannot read also cannot write. Every one of these failures is the store being unreachable — Safari with cookies blocked, a partitioned context — rather than one operation misfiring, so whatever `read` reported, the caller still cannot persist a key and still signs in as a new browser. A third state would name a condition no caller can act on and no layer can repair.

There is no fallback store either: `localStorage` and `sessionStorage` are gated by the same policy. An in-memory fallback was considered and does not help — II signs in through a popup, a fresh JS context each time, so the key is gone before the next sign-in.

Worth noting `idb-keyval` retries on its own: it clears its cached database promise on failure, so the next call reopens. Transient failures already heal, and only the persistent case reaches this `catch`.

src/frontend/src/lib/stores/browser-key.store.ts
sea-snake: The scenario is real and I traced it: an unpersisted key means the next sign-in registers another browser, and past `MAX_BROWSERS` the canister gives up the least recently used one *and ends its sessions*.

Not taking the fix, because requiring the first write turns a degraded sign-in into no sign-in at all, for exactly the users whose storage is blocked. Sign-in has to keep working there. Traced end to end with storage unavailable, every failing step is a storage step and each is already caught: `read` → `undefined`, `generate` is WebCrypto and unaffected, `write` → caught, the canister call succeeds, `accept`'s write → caught. The silent catch is what keeps that path alive rather than an oversight.

So it stays, with the cost recorded: a storage-blocked user adds a browser per sign-in and can eventually cost a real browser its sessions. The only fix that does not break their sign-in is a ceremony that registers no browser at all, which is a canister change and not this PR's.

Also, for the record: the `StaleBrowserKeyError` retry is never reached in this state. That error means presenting a key the canister has already *retired*, and a freshly generated pair has never been seen at all.

src/frontend/src/lib/stores/browser-key.store.ts
sea-snake: Agreed, and it now does — `accept` is gone from `BrowserProof`.

The reason it was a callback is that the rotation needs `browser_id`, which exists only in the canister's reply, and `withBrowserProof` is generic in `T`. So instead of handing back something to call, it takes `browserIdOf: (value: T) => number`. The caller still supplies the id — it is the only one who has it — but a required argument is something `tsc` refuses to let it omit, whereas forgetting to call `accept()` compiled fine.

What forgetting cost, which is why this was worth doing: the record kept the key the canister had just retired, so every later sign-in paid a stale-key recovery round-trip; `currentBrowserId` stayed absent, costing the devices page its marker and `forgetIdentity` its canister call entirely; and `forDescription` had no stored description to compare, so a browser that changed what it reports went unnoticed.

The store's own tests now split on the line that matters: `signIn` is a call the canister answered, `attempt` one it did not. Removing the rotation fails six of them.

MRmarioruci (approved):

src/frontend/src/lib/stores/browser-key.store.ts
sea-snake: Taken as `announcedSuccessor` — camelCase, since every sibling here is (`keyPair`, `browserId`).

The point holds regardless of case: `announced` is an adjective with no noun, so the doc comment underneath had to supply the word ("The successor announced at the last sign-in"). With the noun in the name the comment is free to say why the key is retained instead of what it is.

Contained to this store and its test; the field has no callers outside the module.

src/frontend/src/lib/stores/browser-key.store.ts
sea-snake: Taken as the summary. The opening sentence was knotted — "The record to sign in with, which is a fresh one where this browser no longer matches what it registered as" — and your two lines say it plainly.

Kept the paragraph below them, because that part isn't in the code: why a mismatch rotates rather than asking for the entry to be amended, and why a browser that gets this far hasn't disturbed the entry it's leaving behind.

src/frontend/src/lib/stores/browser-key.store.ts
sea-snake: Leaving this one. The module's vocabulary is terse and prefix-free throughout — `read`, `write`, `generate`, `signed`, `exported`, `prepared`, `sameDescription` — and this would be the only `get*` and the only name that enumerates its branches.

The call site reads `const from = await forDescription(identityNumber, description)`: the record for this description, which is what it returns. Rotating is how it gets there when the stored one no longer fits.

The underlying complaint did land, though — on the comment rather than the name, and it's taken in your suggestion above.

src/frontend/src/lib/stores/browser-key.store.ts
sea-snake: It would work, and most of it already does — the chain decides browser identity today, and it's the canister's rule rather than something to move here. `resolve_browser` advances an entry only by the successor it awaits, refuses a key an entry holds but isn't awaiting with `StaleBrowserKey`, and registers a new browser for a key no entry holds.

Copied profiles resolve exactly as you describe, too. Both copies present the same key; the second is refused, promotes its stored successor and takes the entry; the original then has nothing left to promote and registers fresh. One extra row, no change needed.

What `forDescription` covers is the case the chain *cannot* see: same browser, chain intact, now reporting something different. `resolve_browser` takes the description only on registration — "an entry that is advanced keeps the description it was registered with" — so without the local rotation the row would show the old description forever.

Keeping the key and updating the description on a valid rotation is the part we won't do, and it's a security property rather than an oversight: a mutable description is a device list an attacker can repaint. Whoever can advance an entry could relabel it as the device the user recognises, and the row they'd have noticed stops being a new row. Immutable, an unfamiliar device can only ever appear as its own entry.

The cost of that, stated plainly: it can fire on one physical device. An iPad toggling "Request Desktop Site" reports `Macos` rather than `Ipados` and earns a second row.

src/frontend/src/lib/stores/browser-key.store.ts
```diff
@@ -0,0 +1,307 @@
...
+const write = async (
+  identityNumber: bigint,
+  record: BrowserKeyRecord,
+): Promise<void> => {
+  try {
+    await idbSet(storageKey(identityNumber), record, BROWSER_KEY_STORE);
+  } catch {
+    // A browser that cannot keep its key signs in as a new one next time, which the
+    // identity sees as a new browser rather than as a failure.
+  }
```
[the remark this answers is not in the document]
sea-snake: The scenario needs a read that succeeds and a write that fails, and that is not a state that occurs here. Every one of these failures is the store being unreachable — Safari with cookies blocked, a partitioned context — rather than one operation misfiring, so reads and writes fail together.

With the store unreachable, `read` returns `undefined`, `prepared` generates a fresh pair, and the canister registers a **new** browser. No existing entry is advanced, so nothing is stranded; that path is the accepted degradation, and it costs a row per sign-in rather than a sign-in.

The advance-without-response case the finding builds on is the one this design is *for*. `prepared` puts both halves on disk before the call, so a lost response leaves the browser proving a key the canister retired, receiving `StaleBrowserKey`, and promoting the successor it still holds. A failed final write in `attempt` is harmless for the same reason: disk still holds the pair `prepared` wrote, so the next sign-in proves `S`, is told it is stale, promotes `T` and recovers.

PR #4272, opened 2026-08-22
feat(fe): keep an app's session across a page load

sea-snake:
Design: #4224. Overview: #4230. Consumed by #4273, which stores a record, and #4248, which reads and discards one.

A session outlives the tab that created it. Without storing the chain, a sibling subdomain asking for the first time — or a user returning to a closed tab — would mean another ceremony, which is the cadence this design exists to remove.

Records keyed by identity, account and origin, in IndexedDB alongside the identity's own session.

A record within five minutes of its expiry is not treated as usable, so a chain is never handed over that dies mid-request. Five minutes is half the shortest session a caller can ask for, so a session near that floor is never answered from storage.

The purge that already runs when an identity is discarded now takes these records with it, at the eight places that discard one. Leaving them would keep a chain for an identity the browser has forgotten.

Three of the four exports have no caller here: `storeAppSession` gains one in #4273, `appSessionsForOrigin` and `discardAppSession` in #4248. `purgeAppSessions` is called by this PR.

Tests: `app-session.store.test.ts` (11): storage keyed by identity, account and origin; a record for another origin not served; an expiring record not served; and the purge removing them.


src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -119,6 +120,7 @@
     const identityNumber = $authenticatedStore.identityNumber;
     lastUsedIdentitiesStore.removeIdentity(identityNumber);
     void purgeSession(identityNumber);
+    void purgeAppSessions(identityNumber);
```
MRmarioruci: A bit racy I think.. Could we await both purges before calling window.location.replace? We start the IndexedDB cleanup and immediately replace the page, so the identity can disappear from the UI before its stored credentials have actually been removed

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
```diff
@@ -130,6 +132,7 @@
       $lastUsedIdentitiesStore.identities[`${identityNumber}`];
     lastUsedIdentitiesStore.removeIdentity(identityNumber);
     void purgeSession(identityNumber);
+    void purgeAppSessions(identityNumber);
```
MRmarioruci: Also maybe we can have one shared helper for bot purge functions

src/frontend/src/lib/stores/app-session.store.ts
```diff
@@ -0,0 +1,201 @@
...
+      ([key, record]) => [key, record.accountPrincipal],
+    ),
+  );
+  return (await readAll<AppSessionRecord>(APP_SESSION_STORE)).flatMap(
+    ([key, record]) => {
+      const parsed = parseKey(key);
+      if (parsed?.origin !== origin) {
+        return [];
+      }
+      if (record.expiresAtMillis - EXPIRY_MARGIN_MS <= now) {
```
MRmarioruci, line 147: We could also delete expired sessions instead of just ignoring them

src/frontend/src/lib/stores/app-session.store.ts
```diff
@@ -0,0 +1,201 @@
...
+
+interface SessionKey {
+  identityNumber: bigint;
+  accountNumber?: bigint;
+  origin: string;
+}
+
+/** Returns a copy, so a caller mutating the record cannot write back into the store
+ *  through the object IndexedDB handed us. */
+const normalize = <T extends object>(record: T): T => ({ ...record });
```
MRmarioruci, line 58: Do we need this extra copy? IndexedDB already returns a structured clone, so changing the returned object cannot update the stored record.

src/frontend/src/lib/stores/app-session.store.ts
```diff
@@ -0,0 +1,201 @@
...
+ * to sign in as — it just has to be signed in for.
+ */
+export interface AppAccountRecord {
+  accountPrincipal: string;
+}
+
+// Two databases rather than two object stores, because idb-keyval gives a database one
+// store and fixes it at creation.
+const APP_SESSION_STORE = createStore("ii-app-sessions", "sessions");
+const APP_ACCOUNT_STORE = createStore("ii-app-accounts", "accounts");
```
MRmarioruci, line 43: Do these need to be separate databases? Both records use the same key and are joined again on every read. Could we keep one record with accountPrincipal and an optional session? We could still remove only the credential, while avoiding two scans and the stores getting out of sync.
```
interface StoredAppState {
  accountPrincipal: string;
  session?: AppSessionRecord;
}
```

src/frontend/src/lib/stores/app-session.store.ts
```diff
@@ -0,0 +1,201 @@
...
+  );
+
+export const purgeAppSessions = async (
+  identityNumber: bigint,
+): Promise<void> => {
+  const prefix = `${identityNumber.toString()}:`;
+  await Promise.all(
+    [APP_SESSION_STORE, APP_ACCOUNT_STORE].map(async (store) =>
+      Promise.all(
+        (await readAll<unknown>(store))
```
MRmarioruci, line 191: This style is a bit hard to follow for a human with the nested Promise.alls, await, maps, and filters

src/frontend/src/routes/(new-styling)/authorize/views/ContinueView.svelte
```diff
@@ -262,7 +262,7 @@
               isCanisterError<SessionDelegationError>(err) &&
               err.type === "Unauthorized"
             ) {
-              void purgeSession(selectedIdentityNumber);
+              void forgetIdentity(selectedIdentityNumber);
```
MRmarioruci: Why do we call forgetIdentity here? This only means the stored II session can no longer authorize the request. We used to remove that session and fall back to normal authentication. forgetIdentity also removes every app session and account mapping, and it runs in the background while the new authentication starts. Should this stay purgeSession?

src/frontend/src/routes/(new-styling)/+page.svelte
MRmarioruci, line 126: What should Undo restore now? forgetIdentity may already have revoked the browser’s app sessions and removed their keys, but this callback only adds the identity back to the picker. Should we remove Undo here, or only start the irreversible cleanup after the undo window closes?

src/frontend/src/lib/stores/app-session.store.ts
```diff
@@ -0,0 +1,201 @@
...
+    return undefined;
+  }
+  const separator = key.indexOf(":");
+  const accountSeparator = key.indexOf(":", separator + 1);
+  if (separator === -1 || accountSeparator === -1) {
+    return undefined;
+  }
+  const accountPart = key.slice(separator + 1, accountSeparator);
+  return {
+    identityNumber: BigInt(key.slice(0, separator)),
```
MRmarioruci: Could this return undefined when either number is invalid? BigInt can throw here, so one malformed or older IndexedDB key breaks the whole origin lookup.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
sea-snake: Agreed, and worse than racy — `forgetIdentity` ends this browser's sessions *canister-side*, an update call, and `window.location.replace` on the next line tore the page down while it was in flight. So "sign out and remove" could leave the sessions live, with the apps refreshing against records the canister still held. That is the one thing removal exists to stop, and its own doc says so.

Awaited now. It cannot strand anyone: `forgetIdentity` swallows a canister failure and clears the local records either way.

Since an update call is seconds, the dialog had to show it. `SignOutConfirmation` takes a promise, shows a progress ring on the button and disables **both** — leaving by the other button would navigate out from under the revoke just the same. Modelled on `ReauthPrompt`, which already does exactly this.

src/frontend/src/routes/(new-styling)/manage/(authenticated)/+layout.svelte
sea-snake: Declining this one: they look alike but are not the same operation. `purgeSession` is a single `idbDel` by exact key; `purgeAppSessions` scans two stores and deletes every key belonging to the identity. A shared helper would have to be the scan, which makes the single-key delete slower and less exact. They are also already composed — `forgetIdentity` calls both, so there is one entry point and callers use it.

The real repeat, the per-store loop *inside* `purgeAppSessions`, is fixed under your readability comment below: one function per store rather than one generic helper, so re-keying one store cannot quietly re-key the other.

src/frontend/src/lib/stores/app-session.store.ts
sea-snake: Agreed, and taken wider than asked. My first objection — that a delete could fail — was empty: read and write fail together, so if `readAll` returned the record the delete will manage.

More to the point, `readAll` has already pulled the whole store into memory and only the origin filter stops the rest being examined. So expiry now runs *before* that filter, for every key rather than this origin's, which reaches records for origins the user never returns to — where they actually accumulate. Not awaited, using this file's existing `idbDel(...).catch(() => {})`, so a read costs no latency.

A key `sessionKey` could not have produced goes the same way: no reader reaches it and no purge matches it, so it would sit there forever.

`APP_ACCOUNT_STORE` is deliberately untouched — those records have no expiry, `appAccountsForOrigin` is contractually meant to list accounts whose session is gone, and they are wanted long term.

src/frontend/src/lib/stores/app-session.store.ts
sea-snake: Correct, and the doc comment asserted a mechanism IndexedDB does not have. Every read deserialises a fresh structured clone, so the object is already detached: mutating it cannot reach the record and a second read returns another new object. `normalize` is deleted along with its three call sites.

It was incomplete on its own terms too — `{ ...record }` is shallow, and `AppSessionRecord` holds a nested `CryptoKeyPair` the spread would not have isolated even if isolation were needed.

src/frontend/src/lib/stores/app-session.store.ts
sea-snake: Declining — the divergence this would make impossible is a supported state rather than drift. `AppAccountRecord`'s doc says a sign-in that asked not to be resumable "stores this and no session", and `appAccountsForOrigin` lists accounts "whether or not a session for it survived".

Their lifetimes are now explicitly different: expired sessions are swept on read (your comment above), account records never are, because they are wanted long term. Under one record that stops being a delete and becomes a read-modify-write that must not lose `accountPrincipal`. `discardAppSession` degrades the same way, from a single atomic `idbDel` to get→modify→put that can drop a concurrent `rememberAppAccount`.

They also fail differently on purpose: `storeAppSession` propagates, because a caller must know the credential did not persist, while `rememberAppAccount` swallows — "losing the mapping costs a hint its shortcut, not the sign-in". One record forces one policy on both.

The two-scan cost is nil at this size, and the two databases are not an accident: `idb-keyval` gives a database one store and fixes it at creation.

src/frontend/src/lib/stores/app-session.store.ts
sea-snake: Agreed — two nested `Promise.all`s with an `await` inside a `.map` callback, and only because the two stores were handled by mapping over an array of them.

Split into `purgeSessionsOf` and `purgeAccountsOf`, one per store, composed sequentially. A shared `purgeByPrefix(store, prefix)` was the obvious flattening and is the wrong one: separate data deserves separate purges, and a generic prefix helper becomes a footgun the moment either store is keyed differently.

They also stopped doing string surgery — the filter is `parseKey(key)?.identityNumber === identityNumber`, so the key format is stated once, in this file's own parser.

MRmarioruci (approved):

src/frontend/src/routes/(new-styling)/authorize/views/ContinueView.svelte
sea-snake: Agreed, and it was heavier than it looked. The trigger is a single `Unauthorized` from `get_default_account` on the stored II session — that session is dead and nothing else is — while `forgetIdentity` answers it by attempting `revoke_browser_sessions` and purging **both** stores. So a lapsed II session signed the user out of every app on this browser and destroyed the records that would have let them resume silently.

It was also futile: `actorForIdentity` builds the actor from that same dead session, so the revoke was itself `Unauthorized` and swallowed. What actually happened was the local purge — a UX loss with no security gain, since the apps' own chains keep working until they expire either way.

Now `purgeSession`, which drops the dead session and falls through to the interactive `authenticate()` two lines below that already ran. Same thing `actorForIdentity`'s own `catch` does for this condition.

Three call sites had this pattern, not one — `handleContinueDefault`, `handleEnableMultipleAccounts` and the default-account setter — so all three changed.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: Good question, and chasing it down changed the answer twice.

It restores the list entry and nothing else — but it turns out there was never anything else to restore, so the toast was the thing that was wrong.

`revoke_browser_sessions` goes through `check_authz_and_record_activity` → `check_authorization`, which is full auth only; the session-scoped principal is accepted by `check_session_authorization` and nowhere else. So a stored session delegation is answered `Unauthorized` and revokes nothing. The only caller that could succeed is the live authenticated identity — and `ManageIdentities` disables the remove button on the selected one (`disabled={identity.identityNumber === selected}`, tooltip "Sign out to remove"). From this dialog the remote revoke can never land.

So removal here is purely local, which is what makes Undo worth keeping: the case it exists for is someone removing the wrong entry while tidying the list, and that case signs nobody out.

Three changes rather than a new mechanism:

- The toast no longer claims apps were signed out, because they weren't.
- `forgetIdentity` is replaced by the `purgeSession` / `purgeAppSessions` calls that describe what actually happens, instead of a round trip guaranteed to fail and be swallowed.
- The `isCurrent` branch and the `selectIdentity` call inside the Undo handler are gone — both unreachable, and the comment above them credited the wrong guard ("the last identity cannot be removed", when what prevents it is the disabled button on the selected one).

Deferring the destructive half until the toast expires was considered and rejected: closing the tab inside those five seconds would leave sessions never revoked, which is worse than a partial Undo.

src/frontend/src/lib/stores/app-session.store.ts
sea-snake: It could, and it's worse than one bad key: the function is documented to return `SessionKey | undefined` and honours that for the shape checks, then calls `BigInt` twice unguarded.

Both purges run `parseKey(key)?.identityNumber` over every key in the store, and the expiry sweep runs it over every key on read — so one unreadable key took out the whole purge or the whole read, including the sweep that would have deleted it. The test added for this fails twice without the guard: once on its own assertion, once on the following test, which is exactly that cascade.

Guarded now, returning `undefined` as documented. No regex — same call as the `Nat64StringCodec` thread on #4273, where `BigInt`'s own grammar is the thing being narrowed rather than duplicated. Keys the parser rejects are already deleted on read, so a malformed key is cleaned up rather than sitting there forever.

src/frontend/src/lib/stores/session-delegation.store.ts
```diff
@@ -116,3 +118,37 @@ export const actorForIdentity = async (
...
+    browserId === undefined
+      ? undefined
+      : await actorForIdentity(identityNumber);
+  if (browserId !== undefined && actor !== undefined) {
+    try {
+      await actor.revoke_browser_sessions({
+        identity_number: identityNumber,
+        browser_id: browserId,
+      });
+    } catch {
```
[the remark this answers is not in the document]
sea-snake: Correct on the fact, and it led somewhere bigger: the call was being made where it could never succeed.

`revoke_browser_sessions` is gated on `check_authz_and_record_activity` → `check_authorization`, which is full authorization only — the session-scoped principal is accepted by `check_session_authorization` and nowhere else. Removing an identity from a list never holds that: `manage/(authenticated)` passes `selected` so the identity in use cannot be removed there, and every other identity in these lists is unauthenticated, so `actorForIdentity` builds an actor from a stored session delegation and the request comes back `Unauthorized`. A real round trip, guaranteed to fail, swallowed — while four toasts told the user their apps had been signed out.

One function was hiding this, because what it did depended on whether `authenticationStore` happened to hold the identity, which no caller passed in. Now there are two:

- `forgetIdentity` drops the local records and nothing else. Five callers: the identity lists on the landing page, `manage/(authenticated)`, `mcp`, `cli` and `authorize`.
- `revokeIdentity` ends this browser's sessions and then forgets. One caller: the awaited sign-out-and-remove on the manage page, where the user is authenticated as the identity and the revoke lands.

The result variant is inspected in `revokeIdentity`, so an `Err` no longer resolves like a success. Worth being straight about its weight: it changes nothing observable, because the local records are purged whether the canister refused, failed or succeeded — a test asserting otherwise passed with the check removed, so it was deleted rather than kept. What is tested is the part that can fail: `forgetIdentity` makes no canister call.

Surfacing success to callers is declined for the removal flow, because there is nothing true to surface — the four toasts drop the claim instead.

src/frontend/src/routes/(new-styling)/+page.svelte
```diff
@@ -91,26 +92,17 @@
       duration: 2000,
     });
   };
+  // Local only. `ManageIdentities` disables the remove button on the selected identity,
+  // so the identity removed here is never the one signed in — and revoking a browser's
+  // sessions needs full authorization, which a stored session delegation does not carry.
+  // Dropping the records is therefore the whole of it, which is what makes Undo worth
+  // offering: the identity comes back, and nothing was signed out to restore.
```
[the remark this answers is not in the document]
sea-snake: You are right, and this was a regression I introduced. The fallback was live code and I removed it as unreachable on a premise that is false: `ManageIdentities` is passed `selected` only by `manage/(authenticated)`, and deliberately — that is the one page where the user is signed in. The landing page, `mcp`, `cli` and `authorize` all omit it so that every identity there is removable, which is intended behaviour.

So the fallback is restored, along with the `selectIdentity` in the Undo handler, and the comment above it now explains the actual reason rather than crediting a guard that does not exist there.

Passing `selected` on the landing page was the alternative and was rejected: it would take away a capability that is meant to be there.

src/frontend/src/routes/(new-styling)/+page.svelte
sea-snake: Correcting two claims I made above, both wrong.

I said `ManageIdentities` disables the remove button on the selected identity. It does that only on `manage/(authenticated)`, which is the one page that passes `selected`. This page passes none, deliberately, so every identity here is removable — including the selected one.

Which makes the second claim wrong too: with the button live, removing the identity the user is authenticated as would give `actorForIdentity` the live actor, so the remote revoke *can* land in general. It still cannot land from **this** flow, but for a different reason than I gave — `manage/(authenticated)` blocks the authenticated identity, and every other identity in these lists is unauthenticated, so the actor is built from a stored session delegation, which full authorization refuses.

Two things followed from the mistake, both now fixed. Deleting the `isCurrent` fallback as unreachable was a regression: removing the selected identity left the selection naming a deleted entry and the page rendered as signed out with other identities present. It is restored, with its Undo restoration. And rather than inlining the two purge calls here, this now calls `forgetIdentity`, which was split so that it drops local records only, with `revokeIdentity` as the separate operation that also ends the browser's sessions.

PR #4273, opened 2026-08-22
feat(fe): hand apps a session to re-issue their own delegations from

sea-snake:
Design: #4224. Overview: #4230. Depends on #4272 for the record store, #4271 for the browser key, and #4270 for the browser description — this is what gives all three their production caller.

The canister side is complete after #4265, but nothing hands an app a session yet. This is the JSON-RPC method that does.

**`ii_session_delegation`** is what an app calls instead of `icrc34_delegation` when it wants a session rather than a long-lived delegation. It answers with the session chain, extended to the app's own key, and the app mints five-minute delegations from it with no further browser involvement.

The `ii_` prefix is deliberate: an RPC method travels over a transport shared with other providers and has to say which one it belongs to, which is why every II-specific RPC method carries it.

**`targets` on the chain is restricted to the II canister.** That is a developer guardrail, not a defence against a thief — a thief holding the chain can refresh with it either way. Revocability is the protection that matters, and saying otherwise in a security table would be dishonest.

**The consent duration is honoured**, and an SSO organization's cap still binds it: `valid_for` carries the lifetime the user chose, and an SSO identity sends a duration even when the user picked none, matching the ICRC-34 path.

**The request carries only a session public key and an optional derivation origin.** An app cannot ask for an access level or a lifetime, because both are the user's to decide at consent.

`icrc25_permissions` now reports `ii_session_delegation` as a granted scope so an app can discover the method. `icrc34_delegation` itself is unchanged.

Tests: `sessionDelegation.test.ts` (5) covers a request for another method being ignored and params carrying no session key being rejected. The success path has no unit test here, which is worth knowing rather than glossing: the harness does not stub the authentication store, so `createSession` never completes in it, and neither this path's SSO clamp nor the ICRC-34 one is covered. #4248 stubs enough to drive the handler to an answer and exercises the `targets` restriction and the held-session path. Review `createSession` by reading it.


src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
```diff
@@ -0,0 +1,297 @@
...
+import type { ChannelError } from "$lib/stores/channelStore";
+
+export const SESSION_DELEGATION_METHOD = "ii_session_delegation";
+
+const SessionParamsCodec = z.object({
+  sessionPublicKey: Base64ToPublicKeyCodec,
+  // How long the app is willing for the session to last. A ceiling rather than a
+  // request: what the user picks at consent wins, an SSO organization's cap
+  // narrows it further, and the canister clamps the result.
+  maxTimeToLive: z.optional(StringToBigIntCodec),
```
MRmarioruci, line 53: StringToBigIntCodec calls BigInt directly, which still throws inside safeParse, and this parsing happens before the try/catch. We then send no JSON-RPC error. Maybe a bit better would be
```
const MAX_U64 = 2n ** 64n - 1n;

export const Nat64StringCodec = z.codec(
  z.string().regex(/^\d+$/),
  z.bigint().min(0n).max(MAX_U64),
  {
    decode: (value) => BigInt(value),
    encode: (value) => value.toString(),
  },
);
```

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      id: 1,
+      method: "ii_session_delegation",
+      params: {},
+    });
+
+    expect(sent).toHaveLength(1);
+    expect(sent[0]).toMatchObject({ id: 1, error: { code: -32602 } });
+    expect(onError).toHaveBeenCalledWith("invalid-request");
+  });
+});
```
MRmarioruci, line 97: Could we add at least one success-path test here? The current tests only cover an ignored method and invalid params, so none of the main flow, the canister arguments, stored record, browser-key acceptance, or returned chain, is exercised .

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
```diff
@@ -0,0 +1,297 @@
...
+  // ceiling that applies when the picker offered nothing.
+  const requested = authorized.maxTimeToLive ?? requestedMaxTimeToLive;
+  const validFor =
+    ssoSessionMaxAgeNs !== undefined &&
+    (requested === undefined || requested > ssoSessionMaxAgeNs)
+      ? ssoSessionMaxAgeNs
+      : requested;
+
+  const key = { identityNumber, accountNumber, origin: effectiveOrigin };
+  const iiKey = await ECDSAKeyIdentity.generate({ extractable: false });
```
MRmarioruci, line 241: Nit.. a lot of keys here.. kind of hard to understand what is what. maybe sessionSigningKey r something?

MRmarioruci: or maybe it's better to add a comment to why another keypair is needed at all. Like "let's II keep the session and later extend it to a new key from another tab or subdomain?"

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
```diff
@@ -0,0 +1,322 @@
...
+        session_id: prepared.session_id,
+        expiration: prepared.expiration,
+      })
+      .then(throwCanisterError),
+  );
+
+  const canisterChain = DelegationChain.fromDelegations(
+    [
+      {
+        delegation: new Delegation(
```
MRmarioruci, line 313: We drop the target the canister signed here. prepare_account_session signs this delegation for the II canister, and targets is part of the signature, so rebuilding it without targets makes the chain invalid. Could we update transformSignedDelegation to preserve delegation.targets[0] and reuse it here?

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
sea-snake: Confirmed exactly as you describe, and I checked it against the zod we ship: `BigInt("abc")` throws a raw `SyntaxError` **out of** `safeParse`. The parse sits above the handler's `try`, so a malformed duration skipped the invalid-params branch entirely and the app got no reply at all.

Taken, with the guard inside the codec rather than in front of it. A regex would be a second grammar beside `BigInt`'s own, free to drift from it; instead `decode` catches and reports:

```ts
ctx.issues.push({ code: "invalid_format", format: "nat64", input: value, message: "…" });
return z.NEVER;
```

which makes `safeParse` return `success: false` and the existing branch send the error. Your bounds are in the output schema, and they earn their place — `BigInt` alone accepts `"-1"` and `""`.

`StringToBigIntCodec` stays for the encode-side use further down, where the input is ours. A test with a non-numeric `maxTimeToLive` asserts the reply, and fails with the raw `SyntaxError` if the codec is swapped back.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.test.ts
sea-snake: Agreed — the handler's body had no unit coverage at all, so a regression could have sent the wrong canister arguments, skipped the browser-key rotation, persisted the wrong record or returned a chain the app cannot use, and nothing here would have noticed.

Added, covering the canister arguments, the stored record and the returned chain in one test rather than four sharing an elaborate setup.

One thing that had to go first: the test mocked `$lib/utils/iiConnection` as the identity function, purely to keep that 1078-line module out of its graph — which meant it **disabled** the legacy-domain remap it was standing in for, so any success-path test built on it would have asserted an `effectiveOrigin` that never runs. `remapToLegacyDomain` now lives in `urlUtils`, beside the two constants it already used, and is re-exported from `iiConnection` for its callers there. The mock is gone and the real thing runs.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
sea-snake: Both, and the second suggestion was the more valuable one.

The worst offender turned out to be `key` — a *storage locator* sitting a line above two things that really are keys. `sessionSigningKey` would have made it worse, because the request params already carry `sessionPublicKey`, which is the **app's** key, while this one goes to the canister as `session_key`: "session key" already meant two things depending on the side. So the names now say whose key each is — `recordKey`, `iiSessionIdentity`, `iiSessionPublicKey`, and the request's `sessionPublicKey` unchanged.

And your guess about why the pair exists is close, but the reason is security rather than convenience. Over a redirect there is no browser-verified origin identifying the requester, so the canister must not certify a delegation toward a key the request supplied: a malicious page could otherwise have II authenticate for another domain's derivation origin toward the attacker's own key, then read the finished delegation out of certified chain state. Signing to a key held only here makes the on-chain half inert, and the usable chain is completed by a hop only II can make. Same reason `url.ts` gives for its intermediate-key middleware — which this method has to reproduce, because `interceptDelegation` covers `icrc34_delegation` and passes everything else through.

Keeping the pair in the session record is what also lets a later silent re-auth resume the session, which is a consequence rather than the reason. That is what the comment now says.

MRmarioruci (approved):

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
sea-snake: Correct, and this one had already bitten us in the wild — it produced a `ProtocolError` from the test app on staging C: `the signature tree doesn't contain sig/…/… path`.

Worth being precise about the damage, because it's not a type mismatch: `targets` is part of the message the canister signed, so a delegation rebuilt without them hashes to something no signature in the tree covers, and **every** call the app makes with that chain is refused.

`targets[0]` is carried into the `Delegation` now, with a comment saying why it can't be dropped.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
```diff
@@ -0,0 +1,341 @@
...
+const SessionParamsCodec = z.object({
+  sessionPublicKey: Base64ToPublicKeyCodec,
+  // How long the app is willing for the session to last. A ceiling rather than a
+  // request: what the user picks at consent wins, an SSO organization's cap
+  // narrows it further, and the canister clamps the result.
+  maxTimeToLive: z.optional(Nat64StringCodec),
+  // How long the session may go unminted before the canister ends it. A ceiling
+  // like `maxTimeToLive`: the canister clamps it to between 10 minutes and the
+  // session's own granted length, and applies its own default where absent.
+  maxTimeToIdle: z.optional(Nat64StringCodec),
```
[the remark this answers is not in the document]
sea-snake: The premise is right — `docs/ongoing/revocable-app-sessions-spec.md` declares `params: { sessionPublicKey, icrc95DerivationOrigin? }`, and its lifetime section has the II frontend supplying `valid_for` from the user's pick and the SSO cap, with no app-supplied number anywhere.

Taking the second option: the contract is revised rather than the parameters removed, and #4224 is still open, so it's amended in place rather than shipped around.

Where the finding overstates: both values only ever narrow. What the user picks at consent wins over what the app asked for, an SSO organization's cap narrows that further, and the canister clamps the result. An app still cannot obtain a longer-lived session than the user and their organization allow, which is the property the original contract was protecting — and the ceiling semantics were already accepted earlier in this stack (`requestedMaxTimeToLiveStore`, the duration picker capped by the app's request).

Nothing here touches ICRC-34, for the record: `ii_session_delegation` is II's own method and the document amended is our own. The spec now declares both parameters and carries a section stating the narrowing order.

src/frontend/src/lib/utils/transport/utils.ts
```diff
@@ -135,6 +135,38 @@ export const StringToBigIntCodec = z.codec(z.string(), z.bigint(), {
...
+      } catch {
+        ctx.issues.push({
+          code: "invalid_format",
+          format: "nat64",
+          input: value,
+          message: "expected a nat64 as a decimal string",
+        });
+        return z.NEVER;
+      }
+    },
```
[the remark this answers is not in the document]
sea-snake: Confirmed against the zod we ship: `BigInt("")` and `BigInt(" ")` are `0n`, `BigInt("+1")` is `1n`, `BigInt("0x10")` is `16n`, and the nat64 bounds reject none of them.

The consequence is worth naming, since these are durations: the canister clamps `valid_for` to at least 10 minutes, so a malformed `maxTimeToLive` silently became a 10-minute session instead of the invalid-params error the app could act on.

Taken as `/^\d+$/` before converting. That also makes the `try`/`catch` dead — `BigInt` cannot throw on a string of digits — so the check replaces it rather than joining it. The bounds stay in the output schema, which is what still rejects `"-1"` and anything above 2^64-1.

Six inputs are covered by a table-driven test at the handler level; five of the six fail without the check.

PR #4281, opened 2026-08-24
fix(fe): fall back to the gateway twins when fetching app metadata

aterga:
# Motivation

App metadata (#4221) is fetched from the origin the app's identity is derived for. When that is a canister gateway origin, `remapToLegacyDomain` has already collapsed it onto `ic0.app`, so a canister derives one principal whether the user arrived via `ic0.app`, `icp0.io` or `icp.net`.

A canister need not be served on all three. The staging test app answers `400 client_domain_canister_mismatch` at `.ic0.app` while serving the document at `.icp0.io`. The fetch failed, the failure was silent by design, and the screens fell back to the curated entry, so an app that had published the file correctly still showed as unbranded with nothing to indicate why.

Principal derivation needs that remap. The metadata document does not: it has no bearing on derivation, and the spec is explicit that it carries no principal-derivation security semantics.

# Changes

- **`appMetadataOrigins(origin)`** inverts the remap: for `https://<sub>.ic0.app` it yields that origin followed by the `icp0.io` and `icp.net` twins, preserving a `.raw` label. Any other origin (a custom domain, localhost) yields just itself. All three twins resolve to the same canister, so this widens **where** the document may be served, never **whose** document is used.
- **`fetchAppMetadata`** tries those origins in turn and stops at the first that answers.
- **The same-origin rule for the logo now follows the document**: it is validated against whichever origin actually served it, so a twin's document may reference a logo on that twin, and one pointing back at the remapped origin is rejected exactly as any other cross-origin reference is.
- **`fetchCapped` reports the status it saw** rather than collapsing every failure to `undefined`, which is what lets the caller distinguish the two kinds of failure below.
- **Both directions of the gateway-origin mapping now share one regex** (`GATEWAY_ORIGIN_REGEX` in `urlUtils`): `remapToLegacyDomain` normalizes onto `ic0.app` with it and `gatewayOriginTwins` inverts it, so the set of gateway domains and the shape of a canister subdomain are stated once. `remapToLegacyDomain`'s behaviour is unchanged.
- **The spec documents the new discovery contract** (`docs/ii-spec.mdx`, App metadata): which origins are asked and in what order, which statuses settle the question and which move on, and that the logo's same-origin requirement is relative to the domain that served the document rather than the one that was requested.

## When a twin is tried

Only when the origin did not answer for the app:

| Outcome at an origin | Next |
| --- | --- |
| `200`, document valid | done, metadata applied |
| `200`, document invalid or unparseable | done, no metadata: the origin answered |
| `404` | done, no metadata: the canister says it has no such file |
| `400`, `5xx`, other non-`200` | try the next twin |
| no response (network error, CORS rejection, redirect, timeout) | try the next twin |

The `404` case is what keeps this cheap. All three domains resolve to the same canister, so a canister that says "no such file" has answered for every one of them, and the ordinary case of an app that publishes nothing still costs exactly one request. Only an app in the broken position above pays for extra attempts.

# Tests

11 new cases in `appMetadata.test.ts` (48 total in the file):

- `appMetadataOrigins`: twins for a remapped origin, `.raw` preserved, no twins invented for a custom domain.
- Fallback to `icp0.io` on the exact reported failure (`400 client_domain_canister_mismatch` at `ic0.app`, document at `icp0.io`), fallback on to `icp.net`, and fallback when the first origin cannot be reached at all.
- No twin tried on `404`, and none tried when a document is served but fails validation, each asserting the request count.
- Every origin failing yields no metadata after three attempts.
- A twin's logo resolves against that twin; a twin document whose logo points at the remapped origin is rejected.

Verification:

- `npx vitest run src/frontend/src/lib/utils/appMetadata.test.ts`: 47 passed, 1 failed. The failure is `should fetch a same-origin logo and render it from a blob url`, which asserts `expect.any(Blob)` against a Blob from a different realm. It reproduces identically on `main` with this change stashed (36 passed, 1 failed), so it is a pre-existing sandbox issue and not a regression here.
- `npx vitest run src/frontend/src/lib/utils/validateDerivationOrigin.test.ts`: 25 passed, covering the `remapToLegacyDomain` rewrite.
- `eslint --max-warnings 0`, `prettier --check` and `tsc --project tsconfig.all.json --noEmit` are clean on the changed source files.
- The e2e suite was not run locally (it needs a replica and the built canisters); CI covers it.

# Note on sequencing

I understand the plan was to ship fuller support for this in about a week. This is the narrow fix for the reported symptom only, so it should be easy to drop or fold in if it overlaps with that work.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xb4gLgdkcBhNJdByGQhiLc)_


src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -54,6 +54,41 @@ export interface AppMetadata {
   logo?: string;
 }
 
+/** The canister gateway domains that `remapToLegacyDomain` collapses onto
+ *  `ic0.app` so a canister derives one principal across all of them. */
+const GATEWAY_TWIN_DOMAINS = ["icp0.io", "icp.net"] as const;
+
+const LEGACY_GATEWAY_ORIGIN_REGEX =
```
aterga: Can we reuse a priorly defined regex?

aterga: Done in `ef5c72bd4`.

`GATEWAY_ORIGIN_REGEX` now lives in `urlUtils` and captures both the canister subdomain and the gateway domain, so one pattern serves both directions: `remapToLegacyDomain` normalizes onto `ic0.app` with it, and the new `gatewayOriginTwins` inverts it. `appMetadataOrigins` is down to the origin followed by its twins, with no regex of its own.

`urlUtils` rather than `iiConnection` because `appMetadata` is a leaf module with no local imports, and this way `remapToLegacyDomain` keeps its home and none of its call sites move.

`remapToLegacyDomain` behaviour is unchanged. It used to match only the two remapped domains and pass everything else through; it now matches all three and passes `ic0.app` through, which is the same function. I checked the old and new implementations against each other over `.raw` labels, subdomain-suffix lookalikes such as `x.icp0.io.evil.com`, non-default ports, paths, uppercase input and the empty string: identical on every case. `validateDerivationOrigin.test.ts`, which covers the derivation path that depends on it, is 25/25.

---

src/frontend/src/lib/utils/appMetadata.ts
```diff
@@ -54,6 +54,41 @@ export interface AppMetadata {
...
+ * client_domain_canister_mismatch` at `<id>.ic0.app`. Asking only the remapped
+ * origin would lose the metadata of every app in that position.
+ *
+ * So the remap is inverted here and the twins are used as fallbacks. All three
+ * resolve to the same canister, so this widens where the document may be
+ * served, never whose document is used.
+ */
+export const appMetadataOrigins = (origin: string): string[] => {
+  const subdomain = origin.match(LEGACY_GATEWAY_ORIGIN_REGEX)?.groups
+    ?.subdomain;
```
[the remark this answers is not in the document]
aterga: The facts here are right, and the second of the two remedies is what `ef5c72bd4` does: `gatewayOriginTwins` expands **any** of the three gateway domains, not just `ic0.app`. So `--app <id>.icp0.io` now yields `[icp0.io, ic0.app, icp.net]` and does get the twin fallback. This comment predates that push.

I would not take the first remedy, though. `CliHeader` uses the single `appOrigin` prop for both the metadata store and the origin badge:

```svelte
const metadataStore = $derived(getAppMetadataStore(appOrigin) ...);
const hostname = $derived(originLabel(appOrigin));
```

Passing `remapToLegacyDomain(...)` into it would therefore change what is displayed, so a user running `icp identity link web --app <id>.icp0.io` would see `<id>.ic0.app` in the badge: an origin they never typed and are not signing in from. The origin badge is the one value in that screen a user can actually verify, so it has to stay the origin they came from. Making the metadata origin match the derivation origin exactly would mean splitting the prop in two (display origin and metadata origin), which is a bigger change than this fix needs now that the fallback covers it.

One real residue worth recording: the CLI keys the metadata store under the unremapped origin while `/authorize` keys it under the remapped one, so the same app can occupy two cache entries and fetch the document twice in a session. Both entries resolve to the same canister and the same document, so this costs a request rather than correctness.

---

sea-snake (approved):
I think just `icp.net` was probably sufficient, but PR is fine nontheless.

PR #4285, opened 2026-08-26
test(fe): drive the session scenarios in a browser

sea-snake:
Drives the scenarios from [`docs/ongoing/session-test-scenarios.md`](https://github.com/dfinity/internet-identity/pull/4282) in a real browser: 21 of the 36, in 24 tests — the ones a browser can reach without a second device or a wall-clock month.

The test app is what makes them observable. Its session panel reports the account an app's canisters see, the session key II resolves a session from, what the session and the app delegation have left, how many times the delegation has been replaced, and a log of what the client library did — so a scenario is a handful of assertions on that panel rather than a canister-call transcript. `.github/versions/test-app` therefore moves to `prerelease-2026-08-26-sessions-2`, a prerelease built from the client library's session branches. It is a prerelease deliberately: `/releases/latest` skips prereleases, so nothing else picks it up.

## What to read hardest

`fixtures/testApp.ts` — a page object for the app, so the specs say `signIn`, `replaceDelegation`, `ageDelegation` rather than naming elements. Sign-in waits for the panel to say "signed in": the provider's window closes before the app has minted and stored anything, so waiting on the window alone races the result.

`routes/authorize/app-sessions.spec.ts` — the scenarios, grouped the way the document groups them, each citing the requirement ids it covers.

## The rest of the diff

`vite.config.ts` and `utils.ts` add three hosts under `nice-name.com`, so the sibling-domain scenarios have siblings to share a session across. One canister serves every host the dev server maps there.

The existing authorize specs pick the legacy protocol explicitly. The test app's sign-in button now defaults to ICRC-25 and sessions, and a test that does not say which protocol it wants would silently change what it exercises.

## Coverage

Two tests carry no scenario id, because the document has no scenario for what they assert: that one identity is a different account at each app, and that a silent re-issue keeps the account or fails in one way. They cover requirements the document cites rather than scenarios it lists.

STAY-3 found a real defect while it was being written, in the signer transport rather than here: a channel reported itself open until the heartbeat noticed the window had gone two seconds later, so an immediate retry sent into the closed window and no new one opened. That is fixed in `@icp-sdk/signer` 5.6.3 ([icp-js-signer#53](https://github.com/dfinity/icp-js-signer/pull/53)), which the pinned prerelease builds against, so the test retries the way a user would — with no wait.

The 15 scenarios not here need a second device, a month of wall-clock, or a mainnet deploy.

## What is verified, and what is not

The 24 scenarios and the shared-fixture fix were run locally against this prerelease, and the specific failures the fixture fix addresses (`sso.spec.ts:93`, `index.spec.ts:267`, all of `legacy.spec.ts`) were confirmed one at a time. A full sequential `routes/authorize` sweep is still going as I open this; two `consent.spec.ts` failures in it are not yet explained, and I will follow them up here rather than hold the PR for them.

### DIFF
--- a/.github/versions/test-app
+++ b/.github/versions/test-app
@@ -1,1 +1,1 @@
-release-2026-08-03
+prerelease-2026-08-26-sessions-2
--- a/src/frontend/tests/e2e-playwright/fixtures/authorize.ts
+++ b/src/frontend/tests/e2e-playwright/fixtures/authorize.ts
@@ -1,5 +1,6 @@
 import { IDL } from "@icp-sdk/core/candid";
 import { Principal } from "@icp-sdk/core/principal";
+import { readCanisterId } from "@dfinity/internet-identity-vite-plugins/utils";
 import { test as base, expect, type Page } from "@playwright/test";
 import { II_URL, toBase64 } from "../utils";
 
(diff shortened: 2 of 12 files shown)


sea-snake:
Superseded by a stack, one PR per group of `docs/ongoing/session-test-scenarios.md`, so each file explains itself without the reader holding the scenario ids in their head:

- #4286 — the test app pin and the fixtures (the base every layer below rests on)
- #4287 — signing in
- #4288 — staying signed in
- #4289 — more than one tab, and sibling subdomains
- #4290 — ending a session
- #4291 — using many apps and browsers
- #4292 — what the owner sees
- #4293 — coming back later

Same 24 scenarios, same content: `git diff` between this branch's tip and the top of the stack is empty. Closing this one.

PR #4286, opened 2026-08-26
test(fe): give the e2e a test app that holds a session

sea-snake:
The layer the session scenarios above this one are built on: a test app that holds a session, and fixtures that can read it.

`.github/versions/test-app` moves to `prerelease-2026-08-26-sessions-2`, built from the client library's session branches. Its session panel is what makes a session observable — the account an app's canisters see, the key II resolves the session from, what the session and the app delegation have left, how many times the delegation has been replaced, and a log of what the library did. It stays a prerelease deliberately: `/releases/latest` skips prereleases, so nothing moves onto it on its own.

## What to read hardest

`fixtures/testApp.ts` — the app addressed by what it does rather than by which element does it, so a scenario reads as `signIn`, `replaceDelegation`, `ageDelegation`. `signedInApp` is that app once a sign-in has landed: `afterEach` runs before the sign-in fixture's own teardown, so the provider's window may still be closing and the app may still be minting.

`fixtures/authorize.ts` — the same problem, in code that predates this. Four readers took `#principal` as soon as the provider's window closed, waiting for the app only over the redirect transport, on the grounds that the window flow "already has it". Over the session protocol nothing has been minted by then, so the read returned `""` and `Principal.fromText` rejected it with a message about a checksum rather than about a race. `waitForSignInResult` does that wait once for all four — bounded and tolerant, because a sign-in that produced nothing and a test that never opened the app are both outcomes those readers report rather than fail on.

## The rest

The existing authorize specs now say which protocol they mean. The app's sign-in button defaults to ICRC-25 and sessions, so a test that stays silent would quietly change what it exercises.

`vite.config.ts` and `utils.ts` add three hosts under `nice-name.com`, so the sibling-subdomain scenarios have siblings to share a session across. One canister serves every host the dev server maps there.

`app-sessions/helpers.ts` is shared by the layers above and has no caller here.


sea-snake:
Thanks — three of these were real and are fixed:

- **Mis-indented `useLegacyProtocol` in `session-delegation.spec.ts` and `manage-identity.spec.ts`.** Correct, and it was what `frontend-checks` was failing on. Formatted.
- **`authorizedPrincipal` still calling `Principal.fromText` on an empty `#principal`.** Correct, and it would have reintroduced exactly the checksum error the wait exists to avoid: the wait is bounded, so a sign-in that produced nothing leaves the element present and blank, and `principal === null` does not cover `""`. Now empty is treated as absent.
- **The orphaned `authorizeWithUrl` doc comment.** Correct — the helper was inserted between the comment and its function. Moved above both.

One more, found while fixing these rather than in review: the attribute round-trip fills a textbox named `II canister id (signer)`, and the pinned test app labels that field `II canister id:`. A `fill` on a locator matching nothing waits with no default action timeout, so every consent test died at the test timeout inside `afterEach` — 11 of them locally. Fixed here too.

The panel's sentinel strings (`-`, `none held`, `none`) are also gone from the scenarios: `TestApp` now answers questions (`expectHoldsDelegation`, `accountPrincipal`, `expectSharesSession`) and its locators are private.

src/frontend/tests/e2e-playwright/fixtures/authorize.ts
```diff
@@ -208,9 +233,7 @@ export const test = base.extend<{
   },
```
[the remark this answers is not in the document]
sea-snake: Fixed. `principal === null` did not cover `""`, and since the wait is bounded that is exactly what a sign-in producing nothing leaves behind. Empty is now treated the same as absent, so `Principal.fromText` never sees it.

src/frontend/tests/e2e-playwright/utils.ts
```diff
@@ -70,6 +75,19 @@ export const authorize = (
  * @param authenticate The method that will be called within authorize page
  * @returns Authenticated principal
  */
+/**
+ * Selects the raw postMessage protocol, which creates no session.
```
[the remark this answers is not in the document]
sea-snake: Fixed — `useLegacyProtocol` now sits above the `authorizeWithUrl` doc comment rather than between it and its function.

src/frontend/tests/e2e-playwright/fixtures/testApp.ts
```diff
@@ -0,0 +1,365 @@
...
+  async returnToTab(): Promise<void> {
+    await this.page.bringToFront();
+    await this.page.evaluate(() =>
+      document.dispatchEvent(new Event("visibilitychange")),
+    );
+  }
+
+  /** Moves this page past the point where its delegation is due. */
+  async ageDelegation(duration = "05:30"): Promise<void> {
+    await this.page.clock.install();
```
MRmarioruci: Can we install the clock before the refresh timer is created? This runs after sign-in, so the existing timer still uses real time and `fastForward` does not fire it. That means the idle test can pass without testing the timer. Maybe install the clock, reload the signed-in app, then fast-forward.

src/frontend/tests/e2e-playwright/fixtures/testApp.ts
```diff
@@ -0,0 +1,365 @@
...
+      await Promise.all(
+        databases.map(
+          ({ name }) =>
+            new Promise((resolve) => {
+              if (name === undefined) return resolve(undefined);
+              const request = indexedDB.deleteDatabase(name);
+              request.onsuccess =
+                request.onerror =
+                request.onblocked =
+                  () => resolve(undefined);
```
MRmarioruci: Can we fail if deleting IndexedDB is blocked or errors? We currently resolve in both cases, so `clearSiteData` can return while the browser key is still stored.

src/frontend/tests/e2e-playwright/fixtures/testApp.ts
```diff
@@ -0,0 +1,365 @@
...
+
+  /** Asks the provider to re-issue without rendering anything. */
+  async silentReauth(): Promise<void> {
+    await this.page.getByRole("button", { name: "Silent re-auth" }).click();
+  }
+
+  /** Makes a real canister call as whoever the app is acting as. */
+  async whoAmI(): Promise<void> {
+    await this.page.getByRole("button", { name: "Who am I?" }).click();
+    await expect(this.page.locator("#whoamiResponse")).not.toHaveText(
```
MRmarioruci: Can `whoAmI` assert that the response matches `accountPrincipal()`? Any returned principal passes now, including the anonymous principal, so this does not prove the app is acting as its account.

src/frontend/tests/e2e-playwright/fixtures/testApp.ts
sea-snake: You were right, and it was worse than the timer not firing: `install()` also
starts the page's clock at the Unix epoch, so a delegation the canister issued
for this year was never due and ageing it moved nothing at all.

Both fixed, and in one place rather than two. `installTestAppClock` now runs
before the page is navigated — anchored at the real time, and resumed so the
panel's poll keeps ticking — and it is called by both paths into the test app.
Leaving it to one of them was the other half of this: the fixture-driven
scenarios fast-forwarded a page whose clock had never been installed, where the
base is zero and the jump means nothing. That is why all three
"staying signed in" scenarios were passing without exercising a replacement.

src/frontend/tests/e2e-playwright/fixtures/testApp.ts
sea-snake: Fixed, and it turned up a real problem rather than only a missing check: with
the delete rejecting instead of resolving, the scenario failed with
`deleting IndexedDB 'auth-client-db' was blocked` — a database cannot be deleted
while a connection to it is open, and the client holds one for as long as the
page lives.

So `clearSiteData` no longer asks the page to delete its own storage. It asks
the browser, through `Storage.clearDataForOrigin`, which is what "clear site
data" does — then reads the databases back and fails if any survived.

src/frontend/tests/e2e-playwright/fixtures/testApp.ts
sea-snake: Fixed: `whoAmI` now reads the account first and asserts the response equals it,
so the anonymous principal — which is what a call made with no delegation
reaches the canister as — no longer passes.

It earned its keep immediately: it is what "an app left open and untouched"
asserts now that HOLD-2's second half is covered, and a reply-only check would
have gone on passing while nothing was being proved.

MRmarioruci (approved):

PR #4287, opened 2026-08-26
test(fe): run the signing-in session scenarios in a browser

sea-snake:
The "Signing in" scenarios of [`docs/ongoing/session-test-scenarios.md`](https://github.com/dfinity/internet-identity/pull/4282) — FIRST-1 and FIRST-3 — plus the per-app account, which the designs promise and no scenario there names.

What a sign-in leaves behind: an account the app acts as, a delegation to act with straight away rather than at the next foreground, and one sign-in per browser rather than one per attempt.

FIRST-1 lets `authorizePage` perform the ceremony, so its body is the provider's side and the panel is read in `afterEach`. The other two sign in twice, which that fixture does not do, so they drive the app themselves.


sea-snake:
Fixed. `textContent() ?? ""` was masking exactly what you describe — an empty session key would have made `not.toHaveText("")` pass for the wrong reason. `TestApp.sessionKeyPrincipal()` now asserts the app has one and returns a `string`, and `expectSessionKeyOtherThan(before)` does the comparison, so there is no nullable value in the scenario at all. Same treatment for the account read on the other line.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/signing-in.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      identities[0].identityNumber,
+      signInWithIdentity,
+    );
+    await testApp.open();
+    await testApp.signIn(authenticate);
+    const before = await testApp.sessionKey.textContent();
+
+    await testApp.signIn(authenticate);
+
+    await expect(testApp.sessionKey).not.toHaveText(before ?? "");
```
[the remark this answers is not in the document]
sea-snake: This has since been removed, it's now abstracted away by the test app fixture.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/signing-in.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      identities[0].identityNumber,
+      signInWithIdentity,
+    );
+    await testApp.open();
+    await testApp.signIn(authenticate);
+    const before = await testApp.sessionKeyPrincipal();
+
+    await testApp.signIn(authenticate);
+
+    await testApp.expectSessionKeyOtherThan(before);
```
MRmarioruci: Can we check that settings still has one browser entry after the second sign-in? A different session key only proves that the key rotated. This would still pass if the second sign-in created a duplicate browser entry, which is what FIRST-3 is meant to catch.

sea-snake: Agreed, and the session key turned out to be the wrong thing to read anyway:
the panel no longer reports it, so the check was resting on a readout that has
since gone.

FIRST-3 now asserts what the scenario claims — one entry for that pairing — by
counting the browsers settings lists. Read from a second browser, because a
browser reading its own list finds itself under no Sign out button: signing out
the browser in front of you goes through the identity's own sign-out instead.

MRmarioruci (approved):

PR #4288, opened 2026-08-26
test(fe): run the staying-signed-in session scenarios in a browser

sea-snake:
The "Staying signed in" scenarios of [`docs/ongoing/session-test-scenarios.md`](https://github.com/dfinity/internet-identity/pull/4282) — HOLD-1, HOLD-2, HOLD-3, HOLD-4 and HOLD-6 — and the silent re-issue an app can ask for itself.

An app delegation lasts five minutes and a session lasts hours, so an app left open outlives what it was handed. These are the scenarios about that gap being invisible: a replacement happens when a delegation is due and something used it, costs nothing when the tab is idle, keeps the account it had, and never puts a screen in front of the user.

Ageing the delegation is `page.clock`, so the boundary is reached without waiting for it.


sea-snake:
Fixed, and you are right about why: counting the open pages before and after says nothing about a window that opened and closed again, which is the only thing a scenario about rendering nothing is asking. `expectNothingOpens` now records the context's `page` events for the duration of the action and asserts none arrived:

```ts
await signedInApp.expectNothingOpens(() => signedInApp.replaceDelegation());
```

`openWindows` is gone.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/staying-signed-in.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      await expect(signedInApp.account).not.toHaveText("-");
+      const account = await signedInApp.account.textContent();
+      const windowsBefore = signedInApp.openWindows;
+
+      await signedInApp.replaceDelegation();
+
+      // MINT-16: the account does not move. USE-6: nothing is rendered to do it.
+      await expect(signedInApp.account).toHaveText(account ?? "");
+      await expect(signedInApp.delegation).not.toHaveText("none held");
+      expect(signedInApp.openWindows).toBe(windowsBefore);
```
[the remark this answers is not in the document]
sea-snake: Fixed, and for the reason you give: counting says nothing about a window that opened and closed again, which is the only thing this scenario asks. `TestApp.expectNothingOpens` records the context's `page` events for the duration of the action and asserts none arrived, and `openWindows` is gone.

```ts
await signedInApp.expectNothingOpens(() => signedInApp.replaceDelegation());
```

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/staying-signed-in.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  test.describe("an app left open and untouched replaces nothing", () => {
+    test.afterEach(async ({ signedInApp }) => {
+      await signedInApp.expectNoDelegationReplacements();
+
+      // MINT-5 and MINT-14: the scheduled refresh cancels unless the delegation
+      // it would replace signed a request, so an idle tab spends nothing.
+      await signedInApp.ageDelegation();
+      await signedInApp.page.waitForTimeout(1000);
+
+      await signedInApp.expectNoDelegationReplacements();
```
MRmarioruci, line 42: Can we use the app after the idle period, for example with `whoAmI()`? HOLD-2 says it still works when the user comes back. This only checks that no delegation was replaced while it was idle.

sea-snake: Added: the scenario is used after the idle stretch, with `whoAmI()`, which now
asserts the canister saw the app's account rather than merely that something
replied.

Worth saying that this was passing for a second reason too — the page's clock
was never anchored, so nothing was ageing at all. Both halves of HOLD-2 are now
actually exercised.

MRmarioruci (approved):

PR #4290, opened 2026-08-26
test(fe): run the session-ending scenarios in a browser

sea-snake:
The "Ending a session" scenarios of [`docs/ongoing/session-test-scenarios.md`](https://github.com/dfinity/internet-identity/pull/4282) — EXIT-1, EXIT-3, EXIT-5 and EXIT-6 — and the silent re-issue that has nothing left to answer from.

Access that can be ended is the point of the design, so this is where it gets ended: by the app signing out, and by the identity's owner signing a whole browser out from settings.

The settings scenario is the one to read: END-5 lets an app keep working until the delegation it holds expires, so nothing shows the moment access is revoked, and it is the next mint that discovers the session is gone. The test therefore ages the delegation and asks for a replacement rather than expecting an instant answer.


src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/ending-a-session.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+          context,
+          identities[0].identityNumber,
+          signInWithIdentity,
+        );
+        await settings
+          .getByRole("button", { name: "Sign out" })
+          .first()
+          .click();
+        await expect(settings.getByText("Signed out")).toBeVisible();
+        await settings.close();
```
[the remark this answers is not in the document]
sea-snake: Fixed. Both sites now go through `signOutFirstBrowser`, which waits 30s for settings to confirm — ending a sign-in is a canister call, so the 5s default was the wrong budget. The locator moved with it: `listedBrowsers(settings)` is the row-per-browser locator, so the three specs that read that list no longer spell out `getByRole("button", { name: "Sign out" })`.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/ending-a-session.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      // transport opens a channel either way, and SIL-1 is about screens, which
+      // the redirect transport the silent design targets is what makes
+      // checkable.
+      await signedInApp.expectSilentReauthFailed();
+    });
+
+    test("picks an identity and continues", signInAsFirstIdentity);
+  });
+
+  test.describe("signing the browser out from settings ends the app's access", () => {
```
MRmarioruci: Should this sign into two apps before signing the browser out? With one app this only proves that one session was revoked. EXIT-3 is meant to prove that the browser-wide action removes every app session for that browser.

sea-snake: Yes — it now signs into two apps before the browser is signed out, and both are
checked.

It also split in two, because the two cases reach it differently. Signing out
the browser you are using is not something the list offers: that row carries no
button, and its route is the identity's own sign-out, where forgetting is what
revokes. So there is now one scenario for this browser, through that route with
two apps signed in, and one for another browser, through the list's button from
a second context.

MRmarioruci (approved):

PR #4292, opened 2026-08-26
test(fe): run the owner-view session scenario in a browser

sea-snake:
SHOW-2 of [`docs/ongoing/session-test-scenarios.md`](https://github.com/dfinity/internet-identity/pull/4282).

Ending access is only useful if the owner can see what there is to end, so a sign-in has to show up in the identity's own settings.

This hook is the one place that writes the wait out rather than taking `signedInApp`: it reads only the settings page and never the panel, so there is nothing to hang the fixture off, and the list exists only once the sign-in has reached the canister.


sea-snake:
Keeping this one, and the comment explaining it — requesting `signedInApp` here leaves a destructured binding this hook never reads, which `@typescript-eslint/no-unused-vars` rejects (the rule only exempts names matching `^_`). This hook asserts against the identity's settings page and never touches the panel, so there is nothing to hang the fixture off.

Every other hook in the stack does take `signedInApp`, so this is the one exception rather than a pattern.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/owner-view.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+        // Stated here rather than taken from `signedInApp`, which this hook has
+        // nothing else to read: the row exists only once the sign-in has reached
+        // the canister.
+        await testApp.waitUntilSignedIn();
+
+        const settings = await openSettings(
+          context,
+          identities[0].identityNumber,
+          signInWithIdentity,
+        );
```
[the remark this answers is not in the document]
sea-snake: Keeping this one, with the comment that explains it. Requesting `signedInApp` here leaves a destructured binding the hook never reads — this hook asserts against the identity's settings page and never touches the app's panel — and `@typescript-eslint/no-unused-vars` rejects that (it only exempts names matching `^_`, and calling it unused would misrepresent something load-bearing).

Every other hook in the stack does take `signedInApp`; this is the single exception rather than a pattern, which is what the comment says.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/owner-view.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+        await expect(
```
MRmarioruci: Can we assert the browser name and last-used value too? SHOW-2 promises both, but any visible Sign out button passes this test.

sea-snake: Added: it now asserts the brand name the list shows and the last-used value,
which is what SHOW-2 promises.

Last-used reads "Right now" for a browser that has just signed in — the stamp is
only five-minute precise, so anything younger than one grain reads that way
rather than as a duration.

MRmarioruci (approved):

PR #4293, opened 2026-08-26
test(fe): run the coming-back session scenarios in a browser

sea-snake:
The "Coming back later" scenarios of [`docs/ongoing/session-test-scenarios.md`](https://github.com/dfinity/internet-identity/pull/4282) — STAY-2, STAY-3 and STAY-4.

What a browser keeps between visits, and what it must not: clearing the site's data is a clean start, an abandoned sign-in can be tried again from the same browser, and two identities used from one browser never see each other's account.

STAY-3 is the one that found a defect while it was being written, in the signer transport rather than here: a channel reported itself open until the heartbeat noticed the window had gone two seconds later, so an immediate retry sent into the closed window and opened no new one. Fixed in `@icp-sdk/signer` 5.6.3 ([icp-js-signer#53](https://github.com/dfinity/icp-js-signer/pull/53)), which the pinned prerelease builds against, so the test retries with no wait.

STAY-1 needs the browser closed and reopened, which a test cannot do to itself.


src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/coming-back.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+test.describe("coming back later", () => {
+  test.use({ authorizeConfig: SESSION_SIGN_IN });
+
+  test.describe("clearing the site's data is a clean start", () => {
+    test.afterEach(async ({ signedInApp }) => {
+      await signedInApp.clearSiteData();
+      await signedInApp.reload();
+
+      await expect(signedInApp.state).toHaveText("no session");
+    });
```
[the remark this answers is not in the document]
sea-snake: Fixed, and you are right that it mattered: a domain announces a session to its subdomains in a **cookie**, so a "clean start" that left cookies alone was not testing what the scenario says. `clearSiteData` now clears `sessionStorage` and calls `context.clearCookies()` alongside `localStorage` and IndexedDB.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/coming-back.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      await testApp.signOut();
+      await expect(testApp.state).toHaveText("no session");
+
+      await testApp.signIn(
+        continueAs(identities[1].identityNumber, signInWithIdentity),
+      );
+
+      await expect(testApp.account).not.toHaveText(first ?? "");
+    });
```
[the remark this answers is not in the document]
sea-snake: Already covered, though not where you were looking: `expectAccountOtherThan` asserts *both* that the account differs from the one given and that it is not the placeholder, so a placeholder reading fails rather than passes. That is why the scenario reads as one call now instead of two assertions.

```ts
async expectAccountOtherThan(principal: string): Promise<void> {
  await expect(this.account).not.toHaveText(principal);
  await expect(this.account).not.toHaveText(NO_ACCOUNT);
}
```

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/coming-back.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+test.describe("coming back later", () => {
+  test.use({ authorizeConfig: { protocol: "icrc25" } });
+
+  test.describe("clearing the site's data is a clean start", () => {
+    test.afterEach(async ({ signedInApp }) => {
+      await signedInApp.clearSiteData();
+      await signedInApp.reload();
+
+      await signedInApp.expectSignedOut();
```
MRmarioruci: Can this sign in again after the reload? STAY-2 says clearing site data signs the app out and that the next sign-in works. We only test the first half.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/coming-back.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      await signInWithIdentity(authPage, identities[0].identityNumber);
+    });
+    await testApp.expectSignedOut();
+
+    // DEV-13: the browser persisted its key before the call, so a second attempt
+    // is recognised as the same browser rather than blocked by the first.
+    await testApp.signIn(
+      continueAs(identities[0].identityNumber, signInWithIdentity),
+    );
+    await testApp.waitUntilSignedIn();
```
MRmarioruci, line 63: Can we also check that settings still has one browser entry? STAY-3 says the retry is recognized as the same browser. A successful retry alone would also pass if it created a duplicate.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/coming-back.spec.ts
sea-snake: Added, and it found something: signing in again does not work after a reload,
because clearing the site's data takes the app's own configuration with it —
which identity provider to use was typed into the form. A reload sent the retry
to the default provider, which is mainnet.

So the scenario now opens the app again rather than reloading it, which is what a
clean start actually leaves behind: a first-time visitor. Then it signs in, and
that is asserted.

src/frontend/tests/e2e-playwright/routes/authorize/app-sessions/coming-back.spec.ts
sea-snake: Added: settings is read after the retry and must list exactly one browser for
that identity, from a second browser so the entry carries a button to be counted.

Which is the half that would have caught a duplicate — a retry that succeeded
proved only that it was not blocked.

MRmarioruci (approved):

PR #4295, opened 2026-08-26
feat(be): notifications_enabled install arg

MRmarioruci:
The notifications feature has a server-side kill switch (`persistent_state.notifications_enabled`) but nothing could set it, so every notification entry point stayed disabled. This adds the install arg that turns it on, and the staging-deploy support to pass it. Stacked on the tip so the whole feature can be enabled on a staging canister for testing; the arg itself belongs further down the stack and can move to #4219 when the stack is finalized.

## Changes

`notifications_enabled : opt bool` joins `InternetIdentityInit`, applied and read back like the other deploy flags — `opt true` enables the feature, omitting it on upgrade preserves the stored value. It only writes the persistent-state cell, which already carries the field, so there is no stable-memory layout change. `scripts/deploy-common.bash` gains a `--notifications-enabled <true|false>` flag that emits the field in the backend install arg, guarded against double-setting via the extra-args reject-list.

## Tests

Covered by the existing `check_candid_interface_compatibility` test (the `.did` matches the Rust surface). The deploy flag was exercised by dry-running the arg builder both ways and confirming the reject-list guard fires.


MRmarioruci:

The notifications_enabled install argument is folded into the consent feature, so the kill switch ships with the first thing it gates.

PR #4301, opened 2026-08-28
fix(fe): handle invalid session delegation records

### LINKED ISSUE #4163: Authorize flow fails with expiresAtMillis null error
Image

When signing in with one specific Internet Identity on iOS, the authorize flow fails with:

> Unexpected error  
> null is not an object (evaluating 'r.expiresAtMillis')

Signing out and using a different Internet Identity works normally on the same device.

MRmarioruci:
## Summary

- validate session-delegation records loaded from IndexedDB with Zod
- treat `undefined` as a missing record and purge every other invalid stored value
- add regression tests covering `null` and malformed records

Fixes #4163

## Tests

- `npx vitest --config ./vitest.config.ts --run src/frontend/src/lib/stores/session-delegation.store.test.ts --exclude .claude/**`
- `npx eslint --max-warnings 0 src/frontend/src/lib/stores/session-delegation.store.ts src/frontend/src/lib/stores/session-delegation.store.test.ts`
- `npx prettier --check src/frontend/src/lib/stores/session-delegation.store.ts src/frontend/src/lib/stores/session-delegation.store.test.ts`
- `npx tsc --project ./tsconfig.all.json --noEmit`

### DIFF
--- a/src/frontend/src/lib/stores/session-delegation.store.ts
+++ b/src/frontend/src/lib/stores/session-delegation.store.ts
@@ -5,6 +5,7 @@ import {
   del as idbDel,
 } from "idb-keyval";
 import { get } from "svelte/store";
+import { z } from "zod";
 import { Actor, ActorSubclass, HttpAgent } from "@icp-sdk/core/agent";
 import type { _SERVICE } from "$lib/generated/internet_identity_types";
 import { idlFactory as internet_identity_idl } from "$lib/generated/internet_identity_idl";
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/stores/session-delegation.store.ts
```diff
@@ -74,6 +74,11 @@ export const actorForIdentity = async (
     return undefined;
   }
 
+  if (record === null) {
```
sea-snake: Instead of purging for specifically `null`, it should purge when the retrieved value is not specifically `SessionDelegationRecord` or `undefined`. I'd recommend to use Zod for this.

sea-snake (approved):

PR #4316, opened 2026-09-09
fix(be): legacy register checks the pubkey it binds

MRmarioruci:
The legacy `register` ran no uniqueness check on the device pubkey it binds, and its caller check
accepts an arbitrary temp key in place of the device principal, so the pubkey it binds need not be
one the caller holds.

## Changes

`register` now calls `check_pubkey_is_not_used`, the same guard `add`, `update` and `replace`
already run, and traps on a conflict as those do. `verify_caller_is_device_or_temp_key` still
accepts a temp key, which is unavoidable for a WebAuthn credential because it cannot sign canister
calls, but a device with no credential id signs for itself, so `register` now requires the caller
to be that device's principal.

The frontend does not call `register`; it is reachable only over Candid.

## Tests

Registering a pubkey another anchor already uses fails, and a device without a credential id
cannot be registered behind a temp key. Unit tests, clippy and the integration build are green
locally.

### DIFF
--- a/src/internet_identity/src/anchor_management/registration.rs
+++ b/src/internet_identity/src/anchor_management/registration.rs
@@ -1,4 +1,6 @@
-use crate::anchor_management::{activity_bookkeeping, post_operation_bookkeeping};
+use crate::anchor_management::{
+    activity_bookkeeping, check_pubkey_is_not_used, post_operation_bookkeeping,
+};
 use crate::state;
 use crate::state::ChallengeInfo;
 use crate::storage::anchor::Device;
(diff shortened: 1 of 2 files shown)


MRmarioruci:
Closing. With both indexes known complete for pre-existing anchors (#3490 populated the recovery-phrase and credential-id indexes, #3659 the passkey pubkey index, #3713 retired the machinery) and all three first-writer-wins, legacy `register` cannot capture an index entry from anyone: every slot it would claim is already taken. What was left is an invariant rather than an exploit, and the possession half of this PR would tighten a public Candid endpoint in a way that could break a third-party client registering a raw keypair behind a temp key, for no benefit. Not worth it. The branch stays if we want the uniqueness half later.

PR #4317, opened 2026-09-09
feat(be): sweep recovery keys that predate the recovery-phrase index

MRmarioruci:
The recovery-phrase principal index went in behind the write path (#3487), so it holds entries
only for anchors written since. The passkey pubkey index had the same shape, was swept once
(#3659) and the sweep retired (#3713); this one never was. So the uniqueness guard that reads it
cannot see a recovery key held by an anchor untouched since the index landed.

## Changes

`initialize` starts a timer, from both `init` and `post_upgrade`, that drives
`Storage::sweep_recovery_phrase_principal_index_batch`. Each batch walks anchors and indexes
every recovery principal whose slot is free, which is
`sync_anchor_with_recovery_phrase_principal_index` against an empty previous state — the shape
`force_sync_all_indices` used for the passkey index. A slot held by a different anchor is counted
and logged rather than overwritten, because the sweep cannot tell a duplicated phrase from a claim
by an anchor that does not own the key, so it reports and changes nothing.

Progress and counters live in `PersistentState` rather than the heap. #4312 is currently retiring
the account principal index sweep partly because its completion flag was heap state, so every
upgrade forgot it and walked every list again; this one resumes instead.
`recovery_phrase_index_sweep_status` reports entries indexed, collisions and completion. It is
hidden, so the Candid interface does not change.

The counters are worth as much as the writes: `indexed` is how many recovery principals predated
the index, and `collisions` is how many were already claimed by another anchor. Neither number is
observable any other way.

Like its predecessors this is temporary and wants retiring in a follow-up once every deployment
has upgraded through it and it reports done.

## Tests

Unit tests cover indexing unindexed anchors, resuming across batches, leaving a claim by another
anchor alone, idempotence, and the empty and finished cases. An upgrade test asserts the sweep
reports completion rather than ticking for the life of the canister; it deliberately does not
assert an entry count, since the previous release already indexes as it writes and a sweep that
finds everything correct reports nothing indexed. That assertion is what #4312 had to remove.

774 unit tests, clippy and the integration build are green locally.

### DIFF
--- a/src/canister_tests/src/api/internet_identity.rs
+++ b/src/canister_tests/src/api/internet_identity.rs
@@ -142,6 +142,22 @@ pub fn lookup(
     query_candid(env, canister_id, "lookup", (anchor_number,)).map(|(x,)| x)
 }
 
+pub fn lookup_caller_identity_by_recovery_phrase(
+    env: &PocketIc,
+    canister_id: CanisterId,
+    sender: Principal,
+) -> Result<Option<types::IdentityNumber>, RejectResponse> {
+    call_candid_as(
+        env,
+        canister_id,
+        RawEffectivePrincipal::None,
+        sender,
+        "lookup_caller_identity_by_recovery_phrase",
+        (),
+    )
+    .map(|(x,)| x)
+}
+
 pub fn get_anchor_credentials(
     env: &PocketIc,
     canister_id: CanisterId,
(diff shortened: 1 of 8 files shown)


MRmarioruci:
Closing: redundant. The recovery-phrase principal index was backfilled by #3490, a timer-driven batch migration released alongside the index itself in #3487, which read and rewrote every anchor from the range low bound up to the last allocated at migration start. #3713 later removed that machinery after it completed. I had read only the final state of `force_sync_all_indices`, which #3659 had narrowed to the passkey index precisely because the recovery migration was already done, and wrongly concluded the recovery index had never been swept. There is nothing for a second sweep to find.

PR #4336, opened 2026-09-16
feat(be): store which apps may notify an identity

MRmarioruci:
An app cannot reach a user once they leave its site, and II is the only party that knows which browsers belong to one identity. First of eight. This one records only that an app may notify an identity: nothing sends, and nothing here knows how a notification would reach a browser.

## Changes

`notification_grant_consent` authorizes the caller, folds the origin to the spelling consent is keyed by, and looks up the application the identity reached through it. No application means the identity has never signed in there, refused as `NoSuchSession`, because notifications address the account principal an identity holds at an app and there is nothing to authorize until that exists. With one in hand the grant sets `notifications_consented_at_ns` on the `(anchor, application)` config, reading the record and replacing that one field as every other writer of it must. `notification_revoke_consent` clears the same field and `notification_consent_granted` reads it.

An origin is compared against its own `Url::origin().ascii_serialization()`, so credentials, path, query, fragment, a trailing slash, an explicit `:443` and an uppercase host are rejected rather than trimmed, and gateway twins fold onto the legacy domain. Keying by application then makes a second row for one app impossible rather than unlikely: a wrong canonicalization can only fail to find the application and refuse.

`notifications_enabled_origins` is the deploy-time gate. Absent or empty, every method above refuses; entries enable those origins only.

No new stable memory. The row is the per-app config at index 20, reclaimed with the identity's account list at that origin, and that list is already bounded per anchor by the account registry's own eviction, so consent settles around 500 rows per identity. Eviction goes by account activity rather than grant age, and fails safe either way: an evicted app loses its permission and is offered it again on the user's next visit.

## Tests

890 unit tests pass in release, clippy clean on wasm32 and host. Coverage: the consent round-trip, gateway twins folding to one row, refusing an app the identity has never reached and an origin this deployment does not notify for, authority spellings no browser can serialize, a grant leaving the default account in the same record alone, and consent surviving an upgrade.


src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+    let Some(authority) = origin.strip_prefix("https://") else {
+        return invalid("origin must be an https:// URL");
+    };
+    if authority.contains(['/', '?', '#']) {
+        return invalid("origin must not carry a path, query or fragment");
+    }
+    let (host, port) = match authority.split_once(':') {
+        Some((host, port)) => (host, Some(port)),
+        None => (authority, None),
+    };
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in 5730735b6, and 29cd38fc0 then replaced the explicit check with a comparison against `Url::origin().ascii_serialization()`, which rejects userinfo along with every other spelling a browser wouldn't produce.

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+fn subdomain_of(origin: &str) -> Option<&str> {
+    let host = origin.strip_prefix("https://")?;
+    if host.contains('/') {
+        return None;
+    }
+    GATEWAYS
+        .iter()
+        .find_map(|gateway| host.strip_suffix(gateway))
+        .filter(|subdomain| !subdomain.is_empty())
+}
```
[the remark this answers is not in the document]
MRmarioruci: Right. Fixed in 5730735b6 by deleting the second implementation: `consent_origin` calls `remap_to_legacy_domain` from the attributes path, which is the helper the frontend regex mirrors. `leaves_a_deeper_subdomain_alone` pins `foo.bar.icp0.io` staying put.

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        Err(NotificationError::Disabled)
+    }
+}
+
+/// Authorize an update via the standard activity-recording gate. Takes the
+/// anchor as an argument (not a caller reverse-lookup) so it works for
+/// OpenID-only identities too.
+fn authorize_update(anchor_number: AnchorNumber) -> Result<(), NotificationError> {
+    check_authz_and_record_activity(anchor_number)
+        .map_err(|err| NotificationError::Unauthorized(err.to_string()))?;
```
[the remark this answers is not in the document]
MRmarioruci: Good catch, and the consequence is sharper than a wrong error type: the trap is an existence oracle, which contradicts what `consent_status` documents about not being probeable. Fixed in 5730735b6. Both wrappers check the anchor exists through a non-trapping read first, and a missing anchor gets the same `Unauthorized` text a wrong caller does, so the two can't be told apart.

src/internet_identity/internet_identity.did
```diff
@@ -2206,4 +2225,12 @@ service : (opt InternetIdentityInit) -> {
 
     // Looks up identity number when called with a recovery phrase
     lookup_caller_identity_by_recovery_phrase : () -> (opt IdentityNumber);
+
+    // ===== Notifications =====
+    notification_grant_consent : (anchor_number : UserNumber, origin : text) -> (variant { Ok; Err : NotificationError });
+    notification_revoke_consent : (anchor_number : UserNumber, origin : text) -> (variant { Ok; Err : NotificationError });
+    // Whether this app may notify this identity. False for a disabled deployment
+    // or an unauthorized caller, so neither can be probed with it.
+    notification_consent_status : (anchor_number : UserNumber, origin : text) -> (bool) query;
```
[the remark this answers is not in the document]
MRmarioruci: The description was stale rather than the code. Listing and mute were deferred deliberately in 8477ca291, to land with the screen that drives them. Description rewritten.

src/internet_identity/src/notifications/consent.rs
```diff
@@ -0,0 +1,202 @@
...
+    origin: FrontendHostname,
+    now_ns: Timestamp,
+) -> Result<(), NotificationError> {
+    let origin = consent_origin(&origin)?;
+
+    storage_borrow_mut(|storage| {
+        let application_number = storage
+            .notification_application(anchor_number, &origin)
+            .ok_or(NotificationError::SessionMissing)?;
+        storage.set_notification_consent(anchor_number, application_number, Some(now_ns));
```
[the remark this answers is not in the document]
MRmarioruci: Also the stale description. The cap stopped being needed when consent moved onto the per-app config: the row is reclaimed with the identity's account list at that origin, and that list is already bounded per anchor by the account registry's own eviction. So it settles around 500 rather than 50, and by account activity rather than grant age. Both are stated in the description now.

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+    let (host, port) = match authority.split_once(':') {
+        Some((host, port)) => (host, Some(port)),
+        None => (authority, None),
+    };
+    if host.is_empty() {
+        return invalid("origin has no host");
+    }
+    if port.is_some_and(|port| port.is_empty() || !port.bytes().all(|b| b.is_ascii_digit())) {
+        return invalid("origin port must be numeric");
```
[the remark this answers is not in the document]
MRmarioruci: Genuine, thanks, both of those passed. Fixed in 29cd38fc0 by comparing the input against `Url::origin().ascii_serialization()` rather than parsing the authority by hand. That also rejects `:443` and an uppercase host, which a browser spells differently and which would each have keyed a second row for one app. `rejects_an_authority_no_browser_can_serialize` covers them.

src/internet_identity/src/storage.rs
```diff
@@ -2798,6 +2798,78 @@ impl<M: Memory + Clone> Storage<M> {
...
+    pub fn set_notification_consent(
+        &mut self,
+        anchor_number: AnchorNumber,
+        application_number: ApplicationNumber,
+        consented_at_ns: Option<Timestamp>,
+    ) {
+        let mut config = self.lookup_anchor_application_config(anchor_number, application_number);
+        config.notifications_consented_at_ns = consented_at_ns;
+        self.stable_anchor_application_config_memory
+            .insert((anchor_number, application_number), config);
```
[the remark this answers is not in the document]
MRmarioruci: Stale description again. Consent has no map of its own, it's a field on the anchor-application config at index 20, which is also why there's no migration: the record is `#[cbor(map)]`, so an absent tag reads as `None`, and that's the right answer for every row written before the field existed. `should_keep_consent_across_an_upgrade` covers the upgrade. Index 37 is webpush subscriptions on the next PR.

src/internet_identity/internet_identity.did
```diff
@@ -2206,4 +2223,12 @@ service : (opt InternetIdentityInit) -> {
 
     // Looks up identity number when called with a recovery phrase
     lookup_caller_identity_by_recovery_phrase : () -> (opt IdentityNumber);
+
+    // ===== Notifications =====
+    notification_grant_consent : (anchor_number : UserNumber, origin : text) -> (variant { Ok; Err : NotificationError });
```
sea-snake: Similar to the `prepare_account_session` method, let's have a record for arguments. (sam applies to other methods below)

src/internet_identity/internet_identity.did
```diff
@@ -2206,4 +2223,12 @@ service : (opt InternetIdentityInit) -> {
...
+    notification_revoke_consent : (anchor_number : UserNumber, origin : text) -> (variant { Ok; Err : NotificationError });
+    // Whether this app may notify this identity. False for a disabled deployment
+    // or an unauthorized caller, so neither can be probed with it.
+    notification_consent_status : (anchor_number : UserNumber, origin : text) -> (bool) query;
```
sea-snake:
```suggestion
    notification_consent_granted : (anchor_number : UserNumber, origin : text) -> (bool) query;
```

Status would suggest like a label e.g. boolean, while this is simply granted yes/no.

src/internet_identity/internet_identity.did
```diff
@@ -2206,4 +2223,12 @@ service : (opt InternetIdentityInit) -> {
 
     // Looks up identity number when called with a recovery phrase
     lookup_caller_identity_by_recovery_phrase : () -> (opt IdentityNumber);
+
+    // ===== Notifications =====
+    notification_grant_consent : (anchor_number : UserNumber, origin : text) -> (variant { Ok; Err : NotificationError });
+    notification_revoke_consent : (anchor_number : UserNumber, origin : text) -> (variant { Ok; Err : NotificationError });
+    // Whether this app may notify this identity. False for a disabled deployment
+    // or an unauthorized caller, so neither can be probed with it.
```
sea-snake: No need to mention these implementation details in comments, the comments should either not be needed (meaning of methods is clearly implicit) or there for the explicit details a caller actually needs.

src/internet_identity/internet_identity.did
```diff
@@ -1732,6 +1736,19 @@ type ListAvailableAttributesError = variant {
     AuthorizationError : principal;
 };
 
+// Why a notification call was refused.
+type NotificationError = variant {
```
sea-snake: I would split this for both grant and revoke, they could have different variants over time. Alternatively, have a single grant/revoke method with a single error.

src/internet_identity/internet_identity.did
```diff
@@ -1732,6 +1736,19 @@ type ListAvailableAttributesError = variant {
...
+    Disabled;                          // this deployment has notifications off
+    Unauthorized : text;
```
sea-snake: Should match other unauthorized errors.

```suggestion
    Unauthorized : principals;
```

src/internet_identity/internet_identity.did
```diff
@@ -1732,6 +1736,19 @@ type ListAvailableAttributesError = variant {
     AuthorizationError : principal;
 };
 
+// Why a notification call was refused.
+type NotificationError = variant {
+    Disabled;                          // this deployment has notifications off
```
sea-snake: I would remove`Disabled`, use `InternalCanisterError : text;` for it since a frontend can't handle this variant anyway beyond showing it to the user in a toast. We need that variant anyway for other unhandled/unexpected errors e.g. storage errors.

src/internet_identity/internet_identity.did
```diff
@@ -1732,6 +1736,19 @@ type ListAvailableAttributesError = variant {
...
+    Unauthorized : text;
+    InvalidOrigin : text;              // not a bare https://host[:port]
+    InvalidSubscription : vec text;    // every rejected field, not just the first
+    // Nothing here to act on: no consent for that origin, or no subscribed endpoint.
+    NotFound;
+    // The identity has never signed in at the app, so there is no application for a
+    // consent to hang off. Apart from NotFound because signing in there clears it.
+    SessionMissing;
```
sea-snake: Let's use the variant introduced in `AccountSessionError`, this applies to `notification_grant_consent` not to `notification_revoke_consent` since in that scenario a noop is a happy succes.
```suggestion
    NoSuchSession;
```

src/internet_identity/internet_identity.did
```diff
@@ -1732,6 +1736,19 @@ type ListAvailableAttributesError = variant {
...
 };
 
+// Why a notification call was refused.
+type NotificationError = variant {
+    Disabled;                          // this deployment has notifications off
+    Unauthorized : text;
+    InvalidOrigin : text;              // not a bare https://host[:port]
+    InvalidSubscription : vec text;    // every rejected field, not just the first
+    // Nothing here to act on: no consent for that origin, or no subscribed endpoint.
+    NotFound;
```
sea-snake: I'd remove `NotFound` variant, if something to be revoked cannot be found, that means the grant is gone and that's all fine to begin with. No need to have the frontend get an error in that scenario, it already being revoked somehow earlier (or wiped by session limit) means this method is a noop success.

src/internet_identity/internet_identity.did
```diff
@@ -1732,6 +1736,19 @@ type ListAvailableAttributesError = variant {
     AuthorizationError : principal;
 };
 
+// Why a notification call was refused.
+type NotificationError = variant {
+    Disabled;                          // this deployment has notifications off
+    Unauthorized : text;
+    InvalidOrigin : text;              // not a bare https://host[:port]
```
sea-snake: I'd use `InternalCanisterError` instead of having a separate `InvalidOrigin` since this validation is handled by the frontend up front already anyway, the frontend is not going to be wired up this validation anyway and will show a toast anyway, so `InternalCanisterError` is sufficient.

src/internet_identity/internet_identity.did
```diff
@@ -1732,6 +1736,19 @@ type ListAvailableAttributesError = variant {
...
+    InvalidSubscription : vec text;    // every rejected field, not just the first
```
sea-snake: Same as with `InvalidOrigin`, this is not handled by the frontend anyway (just an error toast ) so it can also be handled by `InternalCanisterError` instead.

src/internet_identity/internet_identity.did
```diff
@@ -334,6 +334,10 @@ type InternetIdentityInit = record {
     // set/clear pattern as `dnssec_config`: null keeps the previously stored
     // value, `opt null` clears it, `opt opt "https://..."` sets it.
     mcp_official_url : opt opt text;
+    // Server-side kill switch for the notifications feature. null / `opt false`
+    // (the default) disables every notification endpoint; `opt true` enables
+    // them. Omitting it on upgrade keeps the stored value.
+    notifications_enabled : opt bool;
```
sea-snake: Let's instead have a list of origins (e.g. `https://oisy.com`) that can send notifications, `opt vec text` to match same approach as in `openid_configs`, send null to not touch, empty list to disable and list with entries to enable/update.

This enables us to rollout notifications to early adopter apps one by one. 

```suggestion
    notifications_enabled_origins : opt vec text;
```

src/internet_identity/src/main.rs
```diff
@@ -322,6 +323,29 @@ fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
...
 }
 
+// ---- Notifications: called by II's frontend / service worker ----
+
+#[update]
+fn notification_grant_consent(
+    anchor_number: AnchorNumber,
+    origin: FrontendHostname,
+) -> Result<(), notifications::NotificationError> {
+    notifications::consent::grant_consent(anchor_number, origin)
```
sea-snake: Authorization should be done first upfront in main.rs, not in the notification module!

src/internet_identity/src/main.rs
```diff
@@ -322,6 +323,29 @@ fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
...
+}
+
+#[update]
+fn notification_revoke_consent(
+    anchor_number: AnchorNumber,
+    origin: FrontendHostname,
+) -> Result<(), notifications::NotificationError> {
+    notifications::consent::revoke_consent(anchor_number, origin)
```
sea-snake: Authorization should be done first upfront in main.rs, not in the notification module!

src/internet_identity/src/main.rs
```diff
@@ -322,6 +323,29 @@ fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
...
+}
+
+#[query]
+fn notification_consent_status(anchor_number: AnchorNumber, origin: FrontendHostname) -> bool {
+    notifications::consent::consent_status(anchor_number, origin)
```
sea-snake: Authorization should be done first upfront in main.rs, not in the notification module!

src/internet_identity/src/storage.rs
```diff
@@ -2798,6 +2798,73 @@ impl<M: Memory + Clone> Storage<M> {
...
+    /// reclaimed with the identity's account list at the origin and nothing else.
+    pub fn notification_application(
+        &self,
+        anchor_number: AnchorNumber,
+        origin: &FrontendHostname,
+    ) -> Option<ApplicationNumber> {
+        let application_number = self.lookup_application_number_with_origin(origin)?;
+        self.stored_account_references(anchor_number, application_number)
+            .map(|_| application_number)
+    }
```
sea-snake: This method should go, storage receives origins for applications and returns origins, the application number is an internal concern for storage, it should not be leaked outside of it, so this method returning `Option<ApplicationNumber>` is not acceptable.

src/internet_identity/src/storage.rs
```diff
@@ -2798,6 +2798,73 @@ impl<M: Memory + Clone> Storage<M> {
...
+        &mut self,
+        anchor_number: AnchorNumber,
+        application_number: ApplicationNumber,
+        consented_at_ns: Option<Timestamp>,
+    ) {
+        let mut config = self.lookup_anchor_application_config(anchor_number, application_number);
+        config.notifications_consented_at_ns = consented_at_ns;
+        self.stable_anchor_application_config_memory
+            .insert((anchor_number, application_number), config);
+    }
```
sea-snake: I'd instead update the `lookup_anchor_application_config` method to return `Option<AnchorApplicationConfig>` (instead of a default config, now that we have default accounts always tracked, it no longer needs to be returning a default value). Then rename the method to `read_anchor_application_config` instead and also add a `write_anchor_application_config` that returns a StorageError if the application config does not exist yet (no session created yet).

This all keeps the storage layer abstract (only cares about application configs) without having knowledge about notification business logic.

src/internet_identity/src/main.rs
```diff
@@ -322,6 +323,29 @@ fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
     anchor_management::lookup_caller_identity_by_recovery_phrase(caller)
 }
 
+// ---- Notifications: called by II's frontend / service worker ----
+
+#[update]
+fn notification_grant_consent(
```
sea-snake: As mentioned in did file, arguments should be a single struct, then for validation we can have a single method that turns this into a `ValidXYZ` struct. Then the business logic can accept this validated struct only, knowing it moves forward with validated data.

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,216 @@
+//! Notifications: who may notify a user, and how they're reached.
```
sea-snake: See: https://doc.rust-lang.org/reference/items/modules.html

I'd recommend a `/notifications.rs` and `/notifications/validation.rs`, the former than can export the latter. Keeping the main module file role in deciding what needs to be made public and possible have some functions that interact between multiple submodules.

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+//!
+//! Each `pub fn` checks the feature flag and authorizes the anchor, then delegates to a
+//! storage-only helper taking the resolved `anchor_number`, which keeps the storage logic
+//! testable off-canister.
+
+pub mod consent;
+
+use crate::authz_utils::{
+    check_authorization, check_authz_and_record_activity, IdentityUpdateError,
```
sea-snake: As mentioned in main.rs, needs to go.

src/internet_identity/src/notifications/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+        ));
+    }
+    check_authz_and_record_activity(anchor_number)
+        .map_err(|err| NotificationError::Unauthorized(err.to_string()))?;
+    Ok(())
+}
+
+/// Read-only authorization; a query must not record activity.
+fn authorize_query(anchor_number: AnchorNumber) -> bool {
+    anchor_exists(anchor_number) && check_authorization(anchor_number).is_ok()
```
sea-snake: check_authorization already implies an anchor exists 😂

src/internet_identity/src/notifications/consent.rs
```diff
@@ -0,0 +1,191 @@
...
+use internet_identity_interface::internet_identity::types::{
+    AnchorNumber, FrontendHostname, Timestamp,
+};
+
+/// Records `consented_at_ns` as `origin`'s consent for `anchor_number`, or withdraws it
+/// with `None`.
+///
+/// A grant needs an application to hang off and is refused without one. A withdrawal is
+/// not: consent that cannot exist is already withdrawn.
+fn set_consent(
```
sea-snake: This method should receive an already validated input so all validation checks can go, also it should just read the application config, change it and write it.

src/internet_identity/src/notifications/consent.rs
```diff
@@ -0,0 +1,191 @@
...
+                );
+                Ok(())
+            }
+            None if consented_at_ns.is_some() => Err(NotificationError::SessionMissing),
+            None => Ok(()),
+        },
+    )
+}
+
+pub(crate) fn has_consent(anchor_number: AnchorNumber, origin: FrontendHostname) -> bool {
```
sea-snake: This should just read the application config and return is consent is there or not, no validation needed either since input is already validated.

src/internet_identity/internet_identity.did
MRmarioruci: Done. `NotificationGrantConsentRequest` / `NotificationRevokeConsentRequest` / `NotificationConsentGrantedRequest`, and the webpush methods take records too.

src/internet_identity/internet_identity.did
MRmarioruci: Renamed to `notification_consent_granted`.

src/internet_identity/internet_identity.did
MRmarioruci: Cut.

src/internet_identity/internet_identity.did
MRmarioruci: Split into `NotificationGrantConsentError` and `NotificationRevokeConsentError`.

src/internet_identity/internet_identity.did
MRmarioruci: Done, `Unauthorized : principal`.

src/internet_identity/internet_identity.did
MRmarioruci: Gone, it folds into `InternalCanisterError` now.

src/internet_identity/internet_identity.did
MRmarioruci: Done, on grant only. Revoke treats it as a happy no-op.

src/internet_identity/internet_identity.did
MRmarioruci: Gone. Revoking what was never granted returns Ok.

src/internet_identity/internet_identity.did
MRmarioruci: Gone, validation failures come back as `InternalCanisterError`.

src/internet_identity/internet_identity.did
MRmarioruci: Same, folded into `InternalCanisterError`.

src/internet_identity/internet_identity.did
MRmarioruci: Done, `notifications_enabled_origins : opt vec text` with the null/empty/list semantics you describe.

src/internet_identity/src/main.rs
MRmarioruci: Moved. `check_authz_and_record_activity` runs in main.rs before the module is reached.

src/internet_identity/src/main.rs
MRmarioruci: Moved up to main.rs.

src/internet_identity/src/main.rs
MRmarioruci: Moved up to main.rs, same as the other two.

src/internet_identity/src/storage.rs
MRmarioruci: Gone. The application number stays inside storage now.

src/internet_identity/src/storage.rs
MRmarioruci: Done. `read_anchor_application_config` returns `Option` and there is a `write_anchor_application_config` that errors when no application exists yet.

src/internet_identity/src/main.rs
MRmarioruci: Done. One record per method, and validation hands the business logic a `ValidatedXYZ`.

src/internet_identity/src/notifications/mod.rs
MRmarioruci: Restructured: `notifications.rs` exporting `notifications/validation.rs`, and `notifications/webpush.rs` doing the same for its own submodules.

src/internet_identity/src/notifications/mod.rs
MRmarioruci: Gone with the others.

src/internet_identity/src/notifications/mod.rs
MRmarioruci: Ha. Dropped.

src/internet_identity/src/notifications/consent.rs
MRmarioruci: Done. It takes the validated request and just reads the config, changes the one field, writes it back.

src/internet_identity/src/notifications/consent.rs
MRmarioruci: Done, it is only a read now.

src/internet_identity/src/main.rs
```diff
@@ -323,6 +328,54 @@ fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
...
+    request: NotificationGrantConsentRequest,
+) -> Result<(), NotificationGrantConsentError> {
+    let ValidatedNotificationGrantConsentRequest {
+        anchor_number,
+        origin,
+    } = request.try_into()?;
+    check_authz_and_record_activity(anchor_number)
+        .map_err(|_| NotificationGrantConsentError::Unauthorized(caller()))?;
+
+    notifications::grant_consent(anchor_number, origin, ic_cdk::api::time())
```
sea-snake: The function should explicitly take a ValidatedNotificationGrantConsentRequest to avoid the possibility of it ever being called with unvalidated data.

src/internet_identity/src/main.rs
```diff
@@ -323,6 +328,54 @@ fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
...
+    request: NotificationRevokeConsentRequest,
+) -> Result<(), NotificationRevokeConsentError> {
+    let ValidatedNotificationRevokeConsentRequest {
+        anchor_number,
+        origin,
+    } = request.try_into()?;
+    check_authz_and_record_activity(anchor_number)
+        .map_err(|_| NotificationRevokeConsentError::Unauthorized(caller()))?;
+
+    notifications::revoke_consent(anchor_number, origin)
```
sea-snake: Same point about validation as above here here.

src/internet_identity/src/notifications/validation.rs
```diff
@@ -0,0 +1,276 @@
...
+use crate::delegation::frontend_length_within_limit;
+use internet_identity_interface::internet_identity::types::attributes::remap_to_legacy_domain;
+use internet_identity_interface::internet_identity::types::{
+    AnchorNumber, FrontendHostname, NotificationConsentGrantedRequest,
+    NotificationGrantConsentError, NotificationGrantConsentRequest, NotificationRevokeConsentError,
+    NotificationRevokeConsentRequest,
+};
+use url::Url;
+
+pub struct ValidatedNotificationGrantConsentRequest {
```
sea-snake, line 18: Make sure these structs cannot be constructed in any other way than a tryFrom with their unvalidated counter parts.

sea-snake (approved):

src/internet_identity/src/main.rs
MRmarioruci: Done. `grant_consent` takes the validated request and destructures it in the signature, so there is no version of it that accepts loose fields.

src/internet_identity/src/main.rs
MRmarioruci: Same here, it takes the validated request too.

src/internet_identity/src/notifications/validation.rs
MRmarioruci: They each carry a private marker field now, so the `TryFrom` impls in this module are the only thing that can stamp one. A struct literal anywhere else fails to compile. Applied the same idiom to the Web Push requests on #4339 and #4340.

PR #4337, opened 2026-09-16
feat(fe): ask for notification consent over its own JSON-RPC method

MRmarioruci:
Stacked on #4336 — the frontend compiles against bindings generated from that PR's `.did`.

An app asks Internet Identity for permission to notify one of its users, over a JSON-RPC method of its own rather than a flag on the sign-in request, so it can ask at a moment the user can make sense of.

## Changes

`channelStore` registers `handleNotificationConsentRequest` beside the session one. The handler self-filters on the method and the feature flag, refuses a request that may not paint with an interaction-required error, then signs the user in through the existing authorize ceremony and hands off to `NotifOptInView` through a context store, the same shape attribute consent uses. What the app is told comes from reading `notification_consent_status` back rather than from what the screen believes, so the answer and the stored state cannot drift.

`subscribeAndRegisterDevice` mints a VAPID key, signs a pool of JWTs and sends both as one record. `reconcileDeviceNotifications` runs on authenticated boot and tops the pool up from its length and issue time, since a pool is spent by elapsed time and never shrinks.

The method is off by default: `PUSH_NOTIFICATIONS` gates both the handler and the scope II advertises.

## Tests

`npm run check` 0 errors, lint clean. 25 new unit tests across the handler, the opt-in screen resolver and the device reconcile, including that a silent request never paints, that the feature flag gates the method, and that a full-length pool which has nearly elapsed is topped up. `npm test` fails 63, matching the baseline on this branch's base.


MRmarioruci:
Superseded: this stack was re-cut into six smaller PRs and this branch is now the last of them, so its base had to move from feat/notification-consent to feat/notification-consent-screen. GitHub refuses a base change on a PR already bound into a stack, so reopening instead. The branch and its commit are unchanged.

PR #4339, opened 2026-09-16
feat(be): store a browser's Web Push subscription

MRmarioruci:
Stacked on #4336. A browser registers once and is then shared by every app the identity allows, so subscriptions and consent share nothing but the anchor. Nothing sends yet.

## Changes

`set_webpush_subscription` is authorized by the browser key the caller signs with rather than by an access method: that key says which browser the call comes from, which is what the row is keyed by, and a caller signing with no key this identity is signed in from is refused as `InvalidBrowserKey`. It takes one record rather than positional arguments, because the fields are same-typed blobs that would be silently swappable in a call, and it reports every invalid field at once instead of failing on the first.

It stores the relay endpoint, the VAPID public key the browser minted the subscription with, and the pool of JWTs the browser signed itself: Web Push requires ES256 on P-256 and a canister cannot sign that today. The pool rides on the subscription rather than a record of its own, because it shares that row's key and lifetime and a subscription without one could not receive anything, which is why both arrive in one call.

The subscription hangs off the browser's entry in the identity's browser registry, so it needs no stable memory of its own and inherits the registry's cap and eviction. Signing a browser out clears the subscription and leaves the entry, so signing back in is not a new browser.

`remove_webpush_subscription` is authorized by the identity's access method instead, so a browser that is lost or left behind can be silenced from another one.

The endpoint is capped at 512 bytes and required to parse as an absolute `https` URL, and the VAPID key is checked against the curve rather than only for length. The canister POSTs to the endpoint to wake the browser and a relay rejects a push whose key it cannot parse, so either would otherwise surface only as a browser that silently never receives anything.

The browser's `p256dh` and `auth` keys are not asked for or stored. They exist to encrypt an RFC 8291 payload, and the push carries no body.

## Tests

902 unit tests pass in release, clippy clean on wasm32 and host. Coverage: subscribe and unsubscribe round-tripping, a browser that rotates its endpoint keeping one subscription, two browsers of one identity registering on their own entries, a browser the registry does not list registering nothing, and through Candid: a key a sign-in retired, an access method that is not a browser, and every invalid field reported at once.


src/internet_identity/src/browser_key.rs
```diff
@@ -121,6 +125,27 @@ fn signed_message(domain: &[u8], session_key: &SessionKey, other_key: &PublicKey
...
+    signature: &[u8],
+    endpoint: &str,
+    jwt_issued_at_ns: u64,
+) -> bool {
+    let issued_at = jwt_issued_at_ns.to_be_bytes();
+    let mut message =
+        Vec::with_capacity(WEBPUSH_SIGNATURE_DOMAIN.len() + endpoint.len() + issued_at.len());
+    message.extend_from_slice(WEBPUSH_SIGNATURE_DOMAIN);
+    message.extend_from_slice(endpoint.as_bytes());
+    message.extend_from_slice(&issued_at);
```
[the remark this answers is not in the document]
MRmarioruci: Right, and the doc comment claiming a replay re-pins its own pool was just false. Fixed in af00e2bac: the message now covers the anchor, the VAPID key, the pool and the endpoint. The endpoint goes last as the only variable-length part and the two blobs are hashed rather than appended, so neither can borrow bytes from its neighbour whatever length arrives before validation.

Worth noting this found something worse than the finding. Every refusal test in that module was vacuous, because `subscribe_via_proof` was only ever used in negative assertions and the positive helper bypasses the proof. They all passed against a verifier that rejects everything. There is now a control that fails when the binding is removed.

src/internet_identity/src/notifications/webpush/subscription.rs
```diff
@@ -0,0 +1,318 @@
...
+        browser_id,
+        endpoint,
+        vapid_public_key,
+        jwt_signatures,
+        jwt_issued_at_ns,
+    } = subscription;
+
+    // Report every invalid field at once rather than failing on the first.
+    let errors: Vec<String> = [
+        validate_param_len(endpoint.len(), 1..=MAX_ENDPOINT_LEN, "endpoint"),
```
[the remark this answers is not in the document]
MRmarioruci: Agreed, fixed in af00e2bac. It is parsed and has to be an absolute https URL with a host, keeping the length check ahead of the parse. The canister POSTs to this to wake the browser, so anything else is a row nothing can be delivered to.

src/internet_identity/src/storage.rs
```diff
@@ -215,6 +217,8 @@ const NEXT_APPLICATION_NUMBER_MEMORY_INDEX: u8 = 33u8;
 const LOOKUP_ACCOUNT_WITH_PRINCIPAL_MEMORY_INDEX: u8 = 34u8;
 const LOOKUP_SESSION_WITH_PRINCIPAL_MEMORY_INDEX: u8 = 35u8;
 const NEXT_SESSION_ID_MEMORY_INDEX: u8 = 36u8;
+// Notification indexes, appended after the current max (36)
+const WEBPUSH_SUBSCRIPTIONS_MEMORY_INDEX: u8 = 37u8;
```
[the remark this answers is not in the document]
MRmarioruci: The code is right here and the description was stale, sorry. Main's highest index is 36, so 37 is the next free one. Description fixed.

src/internet_identity/src/storage/anchor.rs
```diff
@@ -747,6 +747,17 @@ impl Anchor {
...
 
+    /// The browser holding this key, in either slot.
+    ///
+    /// Either slot, because a sign-in stores the key it was reached by and the successor
+    /// it announced, while the browser keeps only the successor.
+    pub fn browser_by_key(&self, key: &PublicKey) -> Option<BrowserId> {
+        self.browsers
+            .iter()
+            .find(|browser| browser.current_browser_key == *key || browser.next_browser_key == *key)
+            .map(|browser| browser.id)
```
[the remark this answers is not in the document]
MRmarioruci: Good catch, fixed in af00e2bac. I checked both halves: the frontend does overwrite its keypair with the successor once a sign-in is accepted, and this was the function's only caller, so the current slot bought nothing and kept a copied key usable until the next sign-in. Matching the successor alone now, with the unit fixture flipped to match.

src/internet_identity/internet_identity.did
```diff
@@ -1749,6 +1749,20 @@ type NotificationError = variant {
     SessionMissing;
 };
 
+// What a browser uploads when it registers for Web Push. No p256dh or auth key, since
```
sea-snake: Nit, no need to mention past implementation facts in code comments e.g. "No p256dh or auth key", it's not a diary of this AI, the AI will be gone when it's session ends, nobody cares about the AI's life during that session, "it" will not be missed or remembered.

src/internet_identity/internet_identity.did
```diff
@@ -1749,6 +1749,20 @@ type NotificationError = variant {
...
+// the push carries no body to encrypt.
+type SubscribeDeviceRequest = record {
```
sea-snake:
```suggestion
type SetWebPushSubscriptionRequest = record {
```

src/internet_identity/internet_identity.did
```diff
@@ -2231,4 +2245,8 @@ service : (opt InternetIdentityInit) -> {
     // or an unauthorized caller, so neither can be probed with it.
     notification_consent_status : (anchor_number : UserNumber, origin : text) -> (bool) query;
 
+    // Register the device push subscription and a pool of JWTs so that we don't have to store a VAPID key-pair in the canister.
+    webpush_subscribe_device : (SubscribeDeviceRequest) -> (variant { Ok; Err : NotificationError });
```
sea-snake:
```suggestion
    set_webpush_subscription : (SetWebPushSubscriptionRequest) -> (variant { Ok; Err : SetWebPushSubscriptionError });
```

Also note that error should be specific to the method, the method caller is the current browser key pair, so the error should have a variant that indicates the browser key pair is not valid, do _not_ use the `Unauthorized` variant since that's for anchor methods authenticated by an access method (which a browser key is _not_).

src/internet_identity/internet_identity.did
```diff
@@ -1749,6 +1749,20 @@ type NotificationError = variant {
...
+type SubscribeDeviceRequest = record {
+    anchor_number : UserNumber;
+    endpoint : text;                   // the relay URL the browser was issued
+    vapid_public_key : blob;           // uncompressed SEC1 P-256, echoed to the relay as k=
+    jwt_signatures : vec blob;         // one raw ECDSA signature per validity window
+    jwt_issued_at_ns : nat64;
+    // Which browser this is: the key names an entry in the identity's browser registry,
+    // and the signature over endpoint || jwt_issued_at_ns shows the caller holds it.
+    browser_key : PublicKey;
+    browser_key_signature : blob;
```
sea-snake: `browser_key` and `browser_key_signature` can go since the caller() principal is coming from the browser key pair, so we can just loop over the anchor's browsers and find a matching browsers key (derived to principal) there.

src/internet_identity/src/main.rs
```diff
@@ -325,6 +325,21 @@ fn lookup_caller_identity_by_recovery_phrase() -> Option<IdentityNumber> {
 
 // ---- Notifications: called by II's frontend / service worker ----
 
+#[update]
+fn webpush_subscribe_device(
+    request: notifications::webpush::subscription::SubscribeDeviceRequest,
+) -> Result<(), notifications::NotificationError> {
+    notifications::webpush::subscription::subscribe_device(request)
```
sea-snake: Input validation should map to a `ValidatedXYZ` struct and business logic should only accept that validated struct. Also authorization (checking if caller() matches a browser key for the anchor should be done here, probably needs a new authorization util method for it.

src/internet_identity/src/browser_key.rs
```diff
@@ -16,6 +17,10 @@ const BROWSER_KEY_SIGNATURE_DOMAIN: &[u8] = b"ii-session-browser-key";
 /// replayed in the other's role.
 const SUCCESSOR_KEY_SIGNATURE_DOMAIN: &[u8] = b"ii-session-browser-successor";
 
+/// A third prefix, so a sign-in signature cannot be presented as a Web Push
+/// registration or the other way round.
+const WEBPUSH_SIGNATURE_DOMAIN: &[u8] = b"ii-webpush-subscription";
```
sea-snake: Since the frontend will sign the whole request with the browser key pair, the method caller() will have a principal that matches a given browser's key in the anchor. So all changes in this file (`browser-key.rs`) can go, we're not validating the browser signature like this for the notification subscription endpoint. The session endpoints were a single exception here since we needed multiple signatures to sign the request, we should not copy that exception into notification authorization.

src/internet_identity/src/storage/anchor.rs
```diff
@@ -747,6 +747,19 @@ impl Anchor {
         &self.browsers
     }
 
+    /// The browser holding this key as the successor it announced.
+    ///
+    /// The successor slot only. A browser overwrites its stored keypair with that
+    /// successor as soon as a sign-in is accepted, so the retired key in the other slot
+    /// is one no browser still holds, and matching it would keep a copied key usable
+    /// until the next sign-in.
+    pub fn browser_by_key(&self, key: &PublicKey) -> Option<BrowserId> {
```
sea-snake: This probably becomes browser_by_principal I suppose given that the browser key comes in as the caller() principal.

src/internet_identity/src/notifications/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
+//! The Web Push channel: one subscription per device plus its device-signed VAPID JWT
```
sea-snake: Nit: same as with the notification module, this should be "/webpush.rs" + "/webpush/xyz.rs".

src/internet_identity/src/notifications/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+//! Nothing here encrypts, because the push carries no body: the browser's `p256dh` and
+//! `auth` keys are neither asked for nor stored. What remains is RFC 8292 VAPID, which
+//! authenticates II to the relay either way.
+use std::ops::RangeInclusive;
+use url::Url;
+
+pub mod subscription;
+
+/// Relay endpoints run ~200-300 bytes; capped at 1 KiB.
+pub const MAX_ENDPOINT_LEN: usize = 1024;
```
sea-snake: Let's make it 512, discussed in the team we accept this as safe enough length. We can increase it later on if we find out browsers are incompatible.

src/internet_identity/src/storage/storable/notifications/webpush/jwt_pool.rs
```diff
@@ -0,0 +1,24 @@
...
+pub struct StorableWebPushJwtPool {
+    /// Raw ECDSA P-256 signatures (r‖s, 64 bytes each), in window order.
+    /// `ByteVec` so each signature encodes as a compact CBOR byte string rather
+    /// than an array of integers (the latter ~doubles high-entropy bytes).
+    #[n(0)]
+    pub signatures: Vec<ByteVec>,
+    /// Wall-clock time (ns) the device minted this pool. Window `i` expires at
+    /// `issued_at_ns + (i + 1) * window_ns`.
+    #[n(1)]
+    pub issued_at_ns: Timestamp,
```
sea-snake: There's a storable variant for this one.

```suggestion
    pub issued_at_ns: StorableTimestamp,
```

src/internet_identity/src/storage/storable/notifications/webpush/subscription.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::borrow::Cow;
+
+/// One browser's push subscription, keyed `(anchor, browser_id)` against the browser
+/// registry the anchor already holds. The device's VAPID JWT pool rides on the same row,
+/// sharing its key and lifetime.
+///
+/// Keying by the registry's id rather than by the endpoint is what makes the endpoint
+/// mutable: a browser that re-creates its push subscription keeps its row.
+///
+/// No `p256dh` or `auth` key, since the push carries no body to encrypt.
```
sea-snake: > // No `p256dh` or `auth` key, since the push carries no body to encrypt.

This is yet another diary of yours, nobody cares about your session "life story", I'm sorry Claude.

src/internet_identity/src/storage/storable/notifications/webpush/subscription.rs
```diff
@@ -0,0 +1,119 @@
...
+#[derive(Encode, Decode, Clone, Debug, PartialEq)]
+#[cbor(map)]
+pub struct StorableWebPushSubscription {
+    /// Denormalised so a range scan's value carries its owner.
+    #[n(0)]
+    pub anchor: StorableAnchorNumber,
```
sea-snake: This can go since `StorableWebPushSubscription` should be a property within a browser.

src/internet_identity/src/storage.rs
sea-snake: WEBPUSH_SUBSCRIPTIONS_MEMORY_INDEX should go, let's just store the subscription in the browser storable and an optional property. This was a team decision, the additional data overhead in every browser storable is an accepted choice by the team.

src/internet_identity/src/storage.rs
```diff
@@ -3236,10 +3265,24 @@ impl<M: Memory + Clone> Storage<M> {
...
 
         self.write_account_state(anchor, now, state)?;
+        self.remove_webpush_subscriptions(anchor_number, &BTreeSet::from([browser_id]));
 
         Ok(revoked)
     }
 
+    /// Drops the Web Push subscriptions of browsers this identity no longer lists. A row
+    /// keyed by a browser that is gone is an endpoint the user can no longer revoke.
+    fn remove_webpush_subscriptions(
```
sea-snake: The `remove_webpush_subscriptions` function is no longer needed, storage shouldn't be bothered with subscription logic, it should just simply read and write browsers that happen to have a web push subscription property.

src/internet_identity/internet_identity.did
MRmarioruci: Cut, and swept the others out too.

src/internet_identity/internet_identity.did
MRmarioruci: Renamed.

src/internet_identity/internet_identity.did
MRmarioruci: Done, with an `InvalidBrowserKey` variant instead of `Unauthorized`. Removal and the status query still take `Unauthorized`, since both get called from a different browser than the one they act on and so cannot present its key. Shout if you want those browser-keyed too and I will scope removal to the current browser.

src/internet_identity/internet_identity.did
MRmarioruci: Both gone. The browser is resolved by matching `caller()` against the anchor's browser keys.

src/internet_identity/src/main.rs
MRmarioruci: Done. Validation produces a `ValidatedXYZ`, and there is a new `check_browser_authorization` in `authz_utils.rs` that resolves the caller to one of the anchor's browsers.

src/internet_identity/src/browser_key.rs
MRmarioruci: Reverted, `browser_key.rs` is untouched now. Only the test framework kept a helper, to derive a principal from a browser key.

src/internet_identity/src/storage/anchor.rs
MRmarioruci: Yes, it is `browser_by_principal` now.

src/internet_identity/src/notifications/webpush/mod.rs
MRmarioruci: Done, `webpush.rs` over `webpush/subscription.rs` and `webpush/validation.rs`.

src/internet_identity/src/notifications/webpush/mod.rs
MRmarioruci: 512 it is.

src/internet_identity/src/storage/storable/notifications/webpush/jwt_pool.rs
MRmarioruci: Switched to `StorableTimestamp`.

src/internet_identity/src/storage/storable/notifications/webpush/subscription.rs
MRmarioruci: Gone, sorry. Went through the rest of them as well.

src/internet_identity/src/storage/storable/notifications/webpush/subscription.rs
MRmarioruci: Gone, it hangs off the browser now.

src/internet_identity/src/storage.rs
MRmarioruci: Done. The memory index is gone and the subscription is an optional field on `StorableBrowser`.

src/internet_identity/src/storage.rs
MRmarioruci: Gone. Storage just reads and writes browsers now.

src/internet_identity/src/notifications/webpush/subscription.rs
```diff
@@ -0,0 +1,179 @@
...
+        browser_id,
+    }: ValidatedRemoveWebPushSubscriptionRequest,
+) -> Result<(), RemoveWebPushSubscriptionError> {
+    let anchor = read_anchor(anchor_number)
+        .map_err(RemoveWebPushSubscriptionError::InternalCanisterError)?;
+    write_subscription(anchor, browser_id, None)
+        .map_err(RemoveWebPushSubscriptionError::InternalCanisterError)
+}
+
+pub(super) fn read_anchor(anchor_number: AnchorNumber) -> Result<Anchor, String> {
```
sea-snake: Nit, why does this method exist and why is it public?

MRmarioruci: No good reason. Two-line wrapper, one caller, same file. Inlined it in 396a302 and made `write_subscription` private while I was in there.

src/internet_identity/src/notifications/webpush.rs
```diff
@@ -0,0 +1,131 @@
...
+
+mod subscription;
+mod validation;
+
+pub use subscription::{remove_subscription, set_subscription};
+pub use validation::{
+    ValidatedRemoveWebPushSubscriptionRequest, ValidatedSetWebPushSubscriptionRequest,
+};
+
+#[cfg(test)]
```
sea-snake: Why is there only test stuff in this non test file? I'm confused.

sea-snake (approved):
Pre-approving, left a question on one file.

src/internet_identity/src/notifications/webpush.rs
MRmarioruci: Fair, it read as a test file. The fixtures moved to `webpush/fixtures.rs`, so the module root is the module list and the exports and nothing else.

MRmarioruci: Both sibling test modules read those fixtures, so they need a common ancestor and the module root was the only scope that works. Right scope, wrong file though. Moved them to `webpush/fixtures.rs` in 775bade, same as dnssec, dkim, dmarc and storage do. `webpush.rs` is 16 lines now.

src/internet_identity/src/notifications/webpush/subscription.rs
MRmarioruci: Correction on the SHA above: the branch was rebased after I posted and that commit is gone. Same change is 6859def.

PR #4340, opened 2026-09-16
feat(be): report what a browser registered for Web Push

MRmarioruci:
Stacked on #4339. A pool covers about thirty days, so a browser that stays in use has to replace it before it runs out, and something has to tell it when.

## Changes

`get_webpush_subscription_status` answers the endpoint a browser is registered with, how many windows its pool covers and when that pool was minted. The endpoint is what lets a browser tell its own registration from one another identity's re-subscribe left behind. The other two say when to sign the next pool: a count of unused signatures would sit at thirty forever, since a signature is picked by elapsed window and never removed. It is authorized by the browser key the caller signs with, like the write it reconciles against, so the key that says which browser is asking is also the answer to which one to report and the request carries no browser id to disagree with it. It answers `None` both for a deployment that does not notify and for a caller that is no browser of this identity, so neither can be probed with it.

Replacing a pool goes through `set_webpush_subscription` rather than a method of its own. The same endpoint means a top-up, so the registration keeps the age it had and the pool has to move forward; one that is not newer is refused with `StaleJwtPool`. Any other endpoint is a new registration and starts over.

## Tests

907 unit tests pass in release, clippy clean on wasm32 and host. Coverage: what a registration reports, a status answered to no caller but the browser itself, a top-up keeping the registration it lands on, a rotated endpoint starting fresh, a pool that is not newer being refused, a subscription surviving an upgrade, and one dropped when its browser is signed out or evicted.


src/internet_identity/src/notifications/webpush/jwt_pool.rs
```diff
@@ -0,0 +1,201 @@
...
+        .map_err(|err| NotificationError::InvalidSubscription(vec![err]))?;
+    storage_borrow_mut(|storage| {
+        let key = (anchor_number, browser_id);
+        let Some(mut subscription) = storage.webpush_subscriptions_memory.get(&key) else {
+            return Err(NotificationError::NotFound);
+        };
+        subscription.jwt_pool = Some(StorableWebPushJwtPool {
+            signatures: signatures.into_iter().map(ByteVec::from).collect(),
+            issued_at_ns,
+        });
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in c6c593d24, comparing against the stored pool and refusing anything not newer. Added the test you describe, and checked it fails with the guard deleted.

The refresh proof had the same replay gap as subscribe, so it now covers the pool it installs and the stored VAPID key too.

sea-snake (changes requested):
Let's just use the existing set web push subscription method to replace the jwt pool too, no need for additional methods in the candid. This also simplifies logic and validation to a single path for both new subscriptions and updating them.

src/internet_identity/src/notifications/webpush/jwt_pool.rs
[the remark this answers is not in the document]
MRmarioruci: Guarded, though the shape changed since: sea-snake asked for the refresh method to go, so `set_webpush_subscription` now handles both and refuses a pool that is not newer than the one the endpoint already holds. Unit and integration tests for the newer-then-stale order.

src/internet_identity/internet_identity.did
```diff
@@ -1793,6 +1796,20 @@ type RemoveWebPushSubscriptionError = variant {
     InternalCanisterError : text;
 };
 
+type GetWebPushSubscriptionStatusRequest = record {
+    anchor_number : UserNumber;
+    browser_id : nat32;
```
sea-snake: Browser id isn't needed if this call is authenticated with the browser key.

src/internet_identity/src/main.rs
```diff
@@ -347,6 +348,22 @@ fn set_webpush_subscription(
...
+/// `None` for a deployment that does not notify and for an unauthorized caller, so
+/// neither can be probed with it.
+#[query]
+fn get_webpush_subscription_status(
+    request: GetWebPushSubscriptionStatusRequest,
+) -> Option<WebPushSubscriptionStatus> {
+    let Ok(validated) = ValidatedGetWebPushSubscriptionStatusRequest::try_from(request) else {
+        return None;
+    };
+    let Ok((anchor, _)) = check_authorization(validated.anchor_number) else {
```
sea-snake: I thought this call was made from a service worker, wouldn't it then need to be authenticated by browser key caller() instead of access method?

src/internet_identity/internet_identity.did
MRmarioruci: Right, dropped. The request is just `anchor_number` now and the browser comes from the key the call is signed with.

src/internet_identity/src/main.rs
MRmarioruci: It runs from the page on authenticated boot rather than a worker, but the point holds either way: it is the browser asking about its own registration, so it is authorized by `check_browser_authorization` now, same as the write it reconciles against. The reconcile was signing the read with the identity actor and the write with the browser key, which is where the redundant browser id came from. Frontend follows on #4341 and #4342.

sea-snake (approved):

PR #4342, opened 2026-09-16
feat(fe): decide which notification screen a request should open

MRmarioruci:
Stacked on #4341. Which of four screens an app's request should open, and what committing the user's answer does. The screens are #4345, and nothing renders them until #4343.

## Changes

`resolveOptInScreen` reads this browser's state and whether the app is already allowed: a first-timer gets the full pitch, an already-subscribed browser is only asked about the app, a browser allowed elsewhere is offered this device, and a browser whose permission was denied gets unblock guidance instead of a prompt it cannot raise. An app the user declined recently is skipped rather than asked again.

`enableNotifications` subscribes the browser and then records consent, in that order, so a refusal at the browser prompt leaves no consent row behind for a browser that cannot receive anything. `allowApp` is the same without the subscription, for a browser that only needs to allow one more app. Both grant by asking and reacting to `NoSuchSession` rather than ensuring a session up front, so `mintApplicationSession` runs only where the identity has never reached the app: a second session at one it already holds would drop the one the app is still using.

`notificationDiagnostics` works out which browser this is and what its user would have to do by hand to unblock it, and `NotifUnblockSteps` renders that. They stay together because the guidance is only as good as the detection behind it.

## Tests

`npm run check` 0 errors, lint clean, 21 unit tests covering each screen the resolver can pick, the order the opt-in commits in, and a permission reset to `default` falling through to a screen that asks for it back.


src/frontend/src/lib/components/notifications/NotifUnblockSteps.svelte
```diff
@@ -0,0 +1,87 @@
...
+      case "firefox":
+        return [
+          $t`Click the lock icon in the address bar.`,
+          $t`Under Permissions, clear the "Blocked" setting for Send Notifications.`,
+          $t`Reload the page and try again.`,
+        ];
+      case "safari":
+        return [
+          $t`Open Safari, then Settings, then Websites, then Notifications.`,
+          $t`Find id.ai in the list and set it to Allow.`,
```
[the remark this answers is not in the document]
MRmarioruci: fixed in 13f5096b9. The steps now name `window.location.hostname`, so they match whichever alias the user is actually on.

src/frontend/src/lib/components/notifications/NotifUnblockSteps.svelte
```diff
@@ -0,0 +1,87 @@
...
+      aria-hidden="true"
+    />
+  </summary>
+  <div class="border-border-tertiary border-t px-4 py-3">
+    {#if browser === "ios"}
+      <p class="text-text-tertiary text-sm">
+        <Trans>
+          On iPhone and iPad, notifications work only when the site is added to
+          the Home Screen. Open Internet Identity in Safari, tap Share, then Add
+          to Home Screen, and open it from there to enable notifications.
```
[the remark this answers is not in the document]
MRmarioruci: Agreed, it's unreachable. Dropped the branch in b773ab156 along with the `ios` BrowserKind and the UA sniff. An iPhone now falls through to the safari steps, which are also wrong for it, but nothing can reach them until push works there at all. iOS UX is its own work block.

src/frontend/src/lib/utils/notifications/notificationState.ts
```diff
@@ -0,0 +1,62 @@
...
+    typeof Notification !== "undefined" ? Notification.permission : "denied";
+  if (!supported) {
+    return { supported, permission, subscribed: false };
+  }
+  const subscription = await currentDeviceSubscription();
+  const stored = await loadVapidKey();
+  const subscribed =
+    subscription !== undefined &&
+    stored !== undefined &&
+    stored.endpoint === subscription.endpoint;
```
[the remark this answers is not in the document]
MRmarioruci: fixed in f89e6d614. The resolver now requires a canister row for this identity (browser id plus `webpush_jwt_pool_status`); the local check stays only to tell a retry whether it has an endpoint to reuse.

That made `new-device` reachable for a browser that is already subscribed, where enabling would have re-subscribed and orphaned every other identity's row, so a161a92cf routes enabling through a reuse-or-mint step.

src/frontend/src/lib/utils/notifications/enableNotifications.ts
```diff
@@ -0,0 +1,110 @@
...
+type GrantArgs = {
+  identityNumber: bigint;
+  origin: string;
+  actor: ActorSubclass<_SERVICE>;
+};
+
+export type EnableNotificationsResult =
+  { status: "enabled" } | { status: "permission-denied" };
+
+export const enableNotifications = async ({
```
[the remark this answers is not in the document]
MRmarioruci: covered in f744b6ec9: denied, dismissed, and the subscribe-before-consent ordering. The ordering test fails if you swap the two awaits.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
```diff
@@ -289,11 +290,16 @@ export const handleSessionDelegationRequest =
...
 
-        const created = await createSession(
+        // Registered while it runs, so a notification consent arriving in the same
+        // ceremony waits for this session rather than creating one that would drop it.
+        const created = await trackSessionCreation(
           effectiveOrigin,
-          params.maxTimeToLive,
-          params.maxTimeToIdle,
-          resumable === true,
+          createSession(
```
sea-snake, line 297: Just an idea maybe something like `createSessionCached` that caches the first call for a given set of arguments and then subsequent calls will return from that cache. This avoids message handlers overwriting sessions without having them to wait for one another either.

src/frontend/src/lib/utils/notifications/notificationDiagnostics.ts
```diff
@@ -0,0 +1,101 @@
...
+  });
+};
+
+export const wasDeclinedRecently = (origin: string): boolean => {
+  const at = read().declinedFor?.[origin];
+  return at !== undefined && Date.now() - at < DECLINE_COOLDOWN_MS;
+};
+
+/** Best-effort engine sniff, only ever used to pick which unblock steps to show. */
+export const detectBrowser = (): BrowserKind => {
```
sea-snake: This browser detection can all go, we already have JS utils for this used to create a browser description when creating a session.

src/frontend/src/lib/utils/notifications/enableNotifications.ts
```diff
@@ -0,0 +1,114 @@
...
+ */
+const grantConsent = async (args: GrantArgs): Promise<void> => {
+  if (await granted(args)) {
+    return;
+  }
+
+  // A sign-in already under way is about to mint the application, so waiting for it
+  // leaves the session the app asked for in place. Asked again only where there was
+  // something to wait for.
+  if (await awaitSessionCreation(args.origin)) {
```
sea-snake, line 79: Frontend logic code should not await session creation, frontend code should solely write the frontend decision state (e.g. permission granted, origin consented etc) and then a new handler should wait for those and actually do everything needed similar to how delegations and attributes are implemented as channel handlers.

src/frontend/src/lib/utils/notifications/notificationState.ts
```diff
@@ -0,0 +1,90 @@
...
+  return {
+    supported,
+    permission,
+    subscribed,
+    registered: subscribed && (await hasCanisterRow(identityNumber, actor)),
```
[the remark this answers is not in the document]
MRmarioruci: Good catch, this was real. The status query now answers the endpoint, and both the opt-in screen and the boot reconcile compare it against the one the browser holds, so a row another identity's re-subscribe left behind gets repaired instead of trusted.

src/frontend/src/lib/utils/notifications/pushSubscription.ts
```diff
@@ -18,8 +18,8 @@ export const isPushSupported = (): boolean =>
   "Notification" in window;
 
 /** Prompts for notification permission; resolves true only on an explicit grant. */
-export const requestNotificationPermission = async (): Promise<boolean> =>
-  (await Notification.requestPermission()) === "granted";
+export const requestNotificationPermission =
+  (): Promise<NotificationPermission> => Notification.requestPermission();
```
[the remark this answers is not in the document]
MRmarioruci: Comment fixed.

src/frontend/src/lib/utils/notifications/notificationDiagnostics.ts
MRmarioruci: Gone, it uses the existing `describeBrowser` now.

src/frontend/src/lib/utils/notifications/enableNotifications.ts
MRmarioruci: Not done yet, and I would rather not half-do it. Today `enableNotifications` orchestrates prompt then device registration then consent, so moving to a handler that watches decision state is a restructure of this PR plus #4343 and #4345 together, not a patch. Checking with Mario whether that lands in this stack or with the manage screen.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
MRmarioruci: Holding this one for the same reason. `createSessionCached` replaces the `trackSessionCreation` / `awaitSessionCreation` pair, but if the handler ends up owning session creation the pair disappears anyway, so I would rather do the two together than change it twice.

src/frontend/src/lib/utils/notifications/enableNotifications.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (await granted(args)) {
+      return;
+    }
+  }
```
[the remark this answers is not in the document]
MRmarioruci: Real, and not fixed. The check and the mint are not one atom, so a concurrent delegation can slip between them. sea-snake suggested caching session creation per argument set, which would close it, and that is tangled up with restructuring this flow as a channel handler. Keeping it open until that call is made.

src/frontend/src/lib/utils/notifications/enableNotifications.ts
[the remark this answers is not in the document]
MRmarioruci: Covered now: `enabling notifications` tests that the device is registered before consent is recorded, plus the denied short-circuit and a dismissed prompt.

src/frontend/src/lib/utils/notifications/enableNotifications.ts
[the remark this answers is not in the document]
MRmarioruci: Correcting my earlier reply on this one: I said it was real and unfixed, and having traced it properly, it is not reachable.

Every channel handler runs through the single queue in `channelHandlers/serialize.ts`, and the consent handler holds that queue for the whole ceremony: `runConsentCeremony` awaits the screen settling from inside its own `serializeAuthorizationRequest` block, and the screen only settles after the enable call has finished. There are exactly two callers of `prepare_account_session`, `sessionDelegation`'s own (inside its serialized block) and `mintApplicationSession` (reachable only from that screen). So nothing can run between the check and the mint.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
MRmarioruci: Had a proper look at this and I think the pair should go rather than be replaced. Everything already runs through the one queue in `serialize.ts`, so by the time a consent block asks the canister there cannot be a session creation in flight, and if `NoSuchSession` comes back there is nothing to wait for. That makes `trackSessionCreation` / `awaitSessionCreation` look like they are buying nothing rather than needing a cache.

Happy to delete them, here or as a follow-up. Which would you prefer?

src/frontend/src/lib/utils/notifications/enableNotifications.ts
MRmarioruci: Agreed on the shape, but I would rather not do it in this stack.

What was pushing for it was the session race Copilot flagged on this same file, and that turns out not to be reachable: `serialize.ts` runs the handlers one at a time and the consent handler holds the queue across the whole ceremony, so the view cannot interleave with a delegation request. That leaves this as a refactor spanning #4342, #4343 and #4345 rather than a fix, and I would rather land the stack and do it with the manage screen, where the same handler has more to own.

Shout if you would rather it block this PR.

src/frontend/src/lib/stores/channelHandlers/sessionDelegation.ts
MRmarioruci: Deleted them in the end rather than caching. Nothing was waiting on them once the queue argument held up.

src/frontend/src/lib/utils/notifications/mintApplicationSession.ts
```diff
@@ -0,0 +1,65 @@
...
+ * hands nobody the delegation it creates.
+ *
+ * Read-only and as short as the canister allows, since nothing holds it. A later real
+ * sign-in replaces it at no cost, because consent rides the per-app config rather than
+ * the session.
+ *
+ * Only for an origin the identity has never reached: called for one it has, it would
+ * replace a session the app is still using.
+ */
+export const mintApplicationSession = async ({
```
sea-snake: This doesn't seem deduped, the `prepare_account_session` method would invalidate any prior session if there's already one made from the current session sign-in flow. So we'd probably need a single session creation method that returns the result promise from the first invocation for any subsequent invocations, this could be a method in the session handler or somewhere else it makes sense (it's not notification specific.

MRmarioruci: Nothing gets clobbered: the grant runs first and only mints on `NoSuchSession`, so there is no session there to replace. Both callers sit inside `serializeAuthorizationRequest` so they cannot interleave, and when consent goes first the sign-in drops the throwaway inside its own `create_session` on that account.

A cache cannot dedupe them either: consent generates a session key and drops it, the sign-in passes the app's own, so the two results are not interchangeable.

It is a workaround though. Consent wants the application row, not a session. Cleanest fix is letting `notification_grant_consent` create that row, which drops `mintApplicationSession` and `NoSuchSession` both. That is the storage minting path, so better on its own than folded in here.

sea-snake (approved):

PR #4343, opened 2026-09-16
feat(fe): let an app ask for notification consent

MRmarioruci:
Stacked on #4342. Seventh of eight; with this the flow runs end to end behind the flag.

An app asks over a method of its own rather than a flag on the sign-in request, so it can ask at a moment the user can make sense of instead of deciding before it knows anything about them. The cost is that an app on the legacy transport cannot reach it, since that transport emits one `icrc34_delegation` request under a fixed id and rejects a response to any other.

## Changes

`channelStore` registers the handler beside the session one. It self-filters on the method and the feature flag, refuses a request that may not paint with an interaction-required error before anything else happens, then signs the user in through the existing authorize ceremony and hands off to the consent screen through a context store, the same shape attribute consent uses.

What the app is told comes from reading `notification_consent_granted` back rather than from what the screen believes, so the answer and the stored state cannot drift, and an app that was already allowed gets the right answer without the screen having to decide.

The scope is advertised only while the flag is on. An app that saw it otherwise would call a method every handler ignores and wait for an answer that never comes.

## Tests

`npm run check` 0 errors, lint clean, 11 unit tests including that a silent request never paints, that a malformed one never paints either, and that the flag gates the method. `npm test` fails 63, matching the baseline on this branch's base.


src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
```diff
@@ -0,0 +1,148 @@
...
+        jsonrpc: "2.0",
+        id: requestId,
+        error: {
+          code: INTERACTION_REQUIRED_ERROR_CODE,
+          message: "Interaction required",
+        },
+      });
+    };
+
+    const parsed = NotificationConsentParamsCodec.safeParse(request.params);
```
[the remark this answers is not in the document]
MRmarioruci: fixed in ec6f3d7f7 with `params ?? {}`, plus a handler test for the no-params shape. The test harness was always supplying `{}`, which is why nothing caught it.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
```diff
@@ -0,0 +1,148 @@
...
+  await waitForStore(authorizedStore);
+  const { identityNumber, actor } = await waitForStore(authenticationStore);
+
+  notificationConsentStore.setContext({
+    effectiveOrigin,
+    appName: undefined,
+    identityNumber,
+    resolveActor: () => Promise.resolve(actor),
+  });
+  await waitForStore(notificationConsentSettledStore);
```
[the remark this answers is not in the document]
MRmarioruci: fixed in b31eb98ab. The status is read from `authenticationStore` once the screen settles rather than from the capture.

Not adding restart machinery for the switch itself. Worth noting the actor is shared and has its identity replaced in place, so `identityNumber` was the only half that could actually drift.

src/frontend/src/routes/(new-styling)/authorize/+page.svelte
```diff
@@ -639,6 +647,20 @@
   {/if}
 {/snippet}
 
+{#snippet notificationConsentContent()}
+  {#if $notificationConsentStore !== undefined}
+    {#key $notificationConsentStore}
+      <NotifOptInView
```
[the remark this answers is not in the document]
MRmarioruci: fixed in bda225f46 on #4345. `NotifOptInView`'s mount now catches and falls to the failed screen, so the request is always answered.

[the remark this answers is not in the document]
MRmarioruci: Fixed. The recovery probe is best-effort now, so it always reaches the failed screen and settles the request.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
```diff
@@ -0,0 +1,161 @@
...
+          derivationOrigin: params.icrc95DerivationOrigin,
+        });
+        if (validation.result === "invalid") {
+          onError("unverified-origin");
+          return;
+        }
+
+        const effectiveOrigin = remapToLegacyDomain(
+          params.icrc95DerivationOrigin ?? channel.origin,
+        );
```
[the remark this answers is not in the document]
MRmarioruci: Fixed. The handler checks the origin is a bare https origin before the ceremony, so permission is never requested for one the canister would refuse. The scope is only advertised for such origins too.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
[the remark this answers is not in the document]
MRmarioruci: Checked this and there is no identity switcher rendered during the notification step, neither `panelWrapper` nor `NotifOptInView` has one, so the captured number cannot go stale under it. The ceremony also re-reads the authenticated identity when the screen settles.

src/frontend/src/lib/stores/channelHandlers/icrc25.ts
```diff
@@ -23,9 +26,14 @@ const supportedStandards = [
   },
 ];
 
-const scopes = [
+const scopes = (origin: string) => [
   { method: "icrc34_delegation" },
   { method: "ii_session_delegation" },
+  // Only while the feature is on, and only for an origin consent can be keyed by:
+  // an app that saw the scope would otherwise call a method that cannot succeed.
+  ...(get(PUSH_NOTIFICATIONS) && isNotifiableOrigin(origin)
```
sea-snake: I would get rid of the `isNotifiableOrigin(origin)` check and the method, any app origin can receive web push since II's own origin in the one that subscribes and receives them, not the app origin.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (!isSilent) {
+        onError("invalid-request");
+      }
+      return;
+    }
+
+    // Consent is the user's answer, not a cached artifact, so a request that may not
+    // paint is refused before anything else happens.
+    if (isSilent) {
+      await deny();
```
sea-snake: The deny variable seems only used here, so just inline that stuff here then.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
```diff
@@ -0,0 +1,176 @@
...
+          return;
+        }
+
+        const effectiveOrigin = remapToLegacyDomain(
+          params.icrc95DerivationOrigin ?? channel.origin,
+        );
+        // Before the ceremony, which asks the browser for permission and registers
+        // the device: an origin the canister cannot key consent by spends both and
+        // then fails at the grant.
+        if (!isNotifiableOrigin(effectiveOrigin)) {
```
sea-snake: This check is unnecessary, as mentioned above, the II origin is the one that receives notifications, not the app. And II supports authentication with non https apps, e.g. development localhost or chrome extension pages just fine.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
```diff
@@ -0,0 +1,176 @@
...
+ * state and covers an app that was already allowed.
+ *
+ * `authorizedStore` is not evidence of a session at the app: two of the three paths that
+ * set it write nothing to the canister. The grant handles that itself, signing in when
+ * the canister reports there is nowhere to record a consent.
+ */
+const runConsentCeremony = async (
+  effectiveOrigin: string,
+): Promise<boolean> => {
+  authorizationStore.setRequestContext(effectiveOrigin, undefined);
```
sea-snake: The `setRequestContext` method shouldn't be overriding a context that might already have been set by the session request message handler, make sure you e.g. check if it's already set or not.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
```diff
@@ -0,0 +1,176 @@
...
+  // Awaited for its ordering and not its value: the user has to have chosen an identity
+  // before a screen can ask them about notifying it.
+  await waitForStore(authorizedStore);
+  const { identityNumber, actor } = await waitForStore(authenticationStore);
+
+  notificationConsentStore.setContext({
+    effectiveOrigin,
+    identityNumber,
+    resolveActor: () => Promise.resolve(actor),
```
sea-snake: Nit: `resolveActor` seems a bit redundant here, whatever reads this could probably just access the authenticaotn store, also not sure why it's a Promise.resolve() right now when it could just be a value anyway.

sea-snake (approved):

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
MRmarioruci: Inlined.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
MRmarioruci: Right, `setRequestContext` writes the origin and `maxTimeToLive` together, so passing `undefined` erased whatever a session request had asked for. Added a `setRequestOrigin` that leaves the duration alone, with a test.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
MRmarioruci: Both true, and the caching it enabled was dead too: the thunk could never fail or change, so the "resolve once and keep" in the view was guarding nothing. The context holds the actor now and the screen takes it as a value.

src/frontend/src/lib/stores/channelHandlers/icrc25.ts
MRmarioruci: Agreed on the reasoning, the subscription is on II's origin and the app origin is only the consent key. But this check isn't inventing the restriction, it mirrors `canonical_origin` in `notifications/validation.rs`, which refuses any scheme other than https. Dropping it here just moves the failure past the permission prompt and the device registration to the grant, which is what it exists to avoid.

It is inconsistent, though: the sign-in path only checks `frontend_length_within_limit`, so II already mints sessions and application rows for `http://localhost:5173`. Only notifications are stricter, and `notifications_enabled_origins` is the real gate in a deployment anyway.

So the fix is on #4336: drop the scheme check from `canonical_origin` and this mirror goes with it. Worth flagging that it only buys localhost. A `chrome-extension://` origin still fails the bare-origin comparison, because a non-special scheme serializes to an opaque origin rather than to itself, so extension pages need more than this.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
MRmarioruci: Same as above: the restriction is `canonical_origin` in the canister, so this goes when that does, on #4336.

src/frontend/src/lib/stores/channelHandlers/icrc25.ts
MRmarioruci: Gone in 3f095b08. Deleted the module and its test, and `scopes()` no longer takes an origin.

src/frontend/src/lib/stores/channelHandlers/notificationConsent.ts
MRmarioruci: Dropped in 3f095b08, and the canister stopped requiring https in 92b7fe84, so localhost and extension pages go through now.

PR #4351, opened 2026-09-17
feat(fe): let message formats nest in the svelte lingui implementation

sea-snake:
Our svelte lingui implementation did not let message formats nest, in any direction, and every direction failed quietly. The browser-group heading is what surfaced it: Copilot flagged `2 przeglądarek` on [#4350](https://github.com/dfinity/internet-identity/pull/4350), and no translation could fix it, because the heading chose between two messages with a ternary on the count and that gives every language exactly two forms. Polish, Russian and Ukrainian need three or four.

## What did not nest

- `findPluralInCallExpression` took a category value only as a literal or template literal. A nested `$t` or `$plural` hit the `flatMap`'s empty return and vanished from the ICU message.
- `findTransInComponent` and `textInNode` turned any non-identifier expression child into a positional placeholder, so `<Trans>{$plural(…)}</Trans>` extracted as `{0}` with the forms lost and the call left as an opaque runtime value.
- `findTransInCallExpression` accepted `message` only as a literal, so wrapping a plural to attach a `context` did not resolve either.

## One resolver

A single recursive resolver now handles every position where a message can stand — a string, a template literal, `$t\`…\``, `$t({ … })`, and `$plural(value, { …categories })` with each category resolved through the same function. `$select` slots in beside them when it arrives.

An expression position is deliberately narrower than a message position: a plain string or template literal in a `<Trans>` mustache or a template interpolation is a runtime value and stays a placeholder, since only a message format contributes text of its own. Getting that wrong is what made `<Trans>Hi {name}, … contact {"support"}</Trans>` lose its `{0}` on the first attempt.

Descriptor fields stay on the outermost call. There is one catalogue entry, so `id`, `context` and `comment` on a nested node have nowhere to go.

## Both walkers, not just extraction

A nested format is already part of the enclosing message, so extraction would emit a second unused entry for it, and transpilation would rewrite it into its own `i18n.t` call inside a range the enclosing rewrite then overwrites. Each message now reports the ranges it absorbed and both walks skip them. The transpiler collects top-down for that and applies its edits right-to-left, which preserves the inner-before-outer order it had while walking bottom-up.

## A pre-existing transpiler bug, fixed here

`overwriteComponent` sliced a node's own text and then indexed into it with offsets taken from the original source. Once a message inside that node had been rewritten — a `$t` in one of its attributes — the text was a different length and the output was corrupt:

```svelte
<!-- before: invalid svelte, on main today -->
<a title={$t({ id: "vER{@render __children()}})}>there</a>
```

It slices by original offsets now and emits valid code. I confirmed the corruption is byte-identical on `main`, so it predates this change; nesting just makes an inner rewrite the normal case rather than a rarity.

## Positional keys

Unique within a message. Sibling plural categories are alternatives, so each still numbers from where the enclosing message left off — `{0}` in both arms of a standalone plural, exactly as before, so no existing id moves — and the enclosing message resumes past the highest any arm reached. `<Trans>{"a"} and {$plural(n, …)}</Trans>` gives `{0}` to the mustache and `{1}` to the categories instead of colliding.

## The heading

```svelte
<Trans context="Browsers are grouped by platform, which cannot tell one device from several, so the plural hedges with device(s) instead of naming a number">
  {$plural(count, {
    one: `# browser on ${platform}`,
    other: `# browsers on ${platform} device(s)`,
  })}
</Trans>
```

`lingui extract` turns that into one entry, `msgctxt` plus `msgid "{count, plural, one {# browser on {platform}} other {# browsers on {platform} device(s)}}"`, replacing the two ternary messages.

The `device(s)` hedge is deliberate and stays — it is what Google's own session-devices page does. The count of devices is genuinely unknown, since browsers group by platform and two of them may be two machines or one machine twice. What the catalogues showed is that the parenthetical does not survive translation on its own: all ten had resolved it to a definite plural, claiming several devices where the English claims nothing. The `context` is there so a translator knows to keep the ambiguity. Note that `msgctxt` feeds the message id, so rewording it re-orphans the translations — worth settling now rather than later.

## Tests

27 new cases, covering extraction and transpilation separately, since a message can extract correctly and still transpile to broken svelte:

- every nesting direction: plural in `<Trans>`, among text, inside a tag, as a `$t` descriptor message, inside a `$t` template literal, `$t` in a plural category, plural in a plural category
- `context` arriving from a `<Trans>` attribute and from a `$t` descriptor, with a plural nested inside
- exactly one message extracted per nested construct
- positional keys staying distinct across nesting
- an expression holding a string or a template literal still becoming a placeholder
- the attribute case above, pinning valid output

The 63 existing tests pass unmodified, so no id or message changed for anything already in the catalogues.

I checked each new behaviour can actually fail: stubbing the resolver to return nothing fails 18, ignoring the consumed ranges fails 13, and restoring the old offset arithmetic fails 1.

`src/frontend/src/lib/locales/*.po` are left alone for the translation bot to regenerate.


sea-snake:
Closing — these commits belong on the existing branch, not a new PR.

PR #4356, opened 2026-09-18
refactor(be): cap concurrent HTTPS outcalls per call site

sea-snake:
Adds a small `ConcurrencyLimiter`, used the same way as `SingleFlightCache`:
one `thread_local` instance per call site, each with its own budget, no central
registry. The DoH resolver, SSO discovery, and JWKS refresh now make their
outcalls through `concurrency::guarded(&LIMITER, || …)`.

Each site bounds how many outcalls it keeps in flight at once and frees the slot
the instant the outcall resolves (the permit releases on `Drop`), so a burst of
distinct lookups no longer fans out without limit. A permit stranded by a caller
that trapped after an `.await` — so its `Drop` never ran — is reclaimed once it
ages out, so a leak can't wedge the budget.

This is groundwork for the web-push delivery dispatch, whose batched fan-out
will hold its own limiter the same way.

Tests cover the limiter accounting (budget, release, stale reclaim) and the
`guarded` wrapper (runs and releases; refuses without running when full).


src/internet_identity/src/concurrency.rs
```diff
@@ -0,0 +1,260 @@
+//! Bounded-concurrency limiter for a shared resource, used the same way as
```
MRmarioruci, line 1: Maybe orthogonal but would be good if we had metrics for these, otherwise we are in the dark if they need raising

sea-snake: Yeah, left that out of scope for now, metrics and our dashboard based on those overall need a revisit for us to actually get the proper alerts in time.

src/internet_identity/src/doh/mod.rs
```diff
@@ -305,12 +319,22 @@ mod prod {
                 },
             ],
         };
-        match http_request_with_closure(request, DOH_CALL_CYCLES, transform_doh).await {
-            Ok((response,)) => {
+        // Bound the provider fan-out against this feature's outcall budget. A
+        // refusal reads as a transient fetch error — the caller's quorum
+        // tolerates a missing provider exactly as it does a network failure.
+        match crate::concurrency::guarded(&OUTCALL_LIMIT, || {
```
MRmarioruci: When the budget is full we hand back a fetch error, which looks exactly like a provider being down. The quorum shrugs that off, but the cache underneath doesn't: it records the lookup as failed, and then refuses to retry that name for 15s, then 30s, then 60s. So a second of saturation costs someone 15 seconds of stalled recovery on a name that was never unreachable, and it's the next caller who pays, not the one holding the slots. SSO discovery is the same shape with a 60s base.

"Too busy to try" and "tried and failed" want opposite answers: retry now versus back off. Could the check move up to where the cache decides whether to start a fill? It already has a no-penalty "not ready, ask again" answer that it gives whenever a fill is in flight, and reusing it for "at capacity" means nothing gets claimed and no name gets penalised. It'd also make the five-provider fan-out all-or-nothing instead of quietly losing whichever providers get polled last.

MRmarioruci (approved):

src/internet_identity/src/doh/mod.rs
sea-snake: Good catch — fixed in ab957b629.

A capacity refusal no longer surfaces as a fetch error. The single-flight cache gained a third fill outcome, `Abandoned`: when a fill can't take a concurrency slot it returns `Abandoned`, and the cache releases the in-flight marker without recording anything — no value, no failure, no backoff — so the name keeps whatever state it had and the next poll retries as soon as capacity frees. "Too busy to try" and "tried and failed" now take the paths you described.

Each fill also takes a single slot up front, so the five-provider fan-out is all-or-nothing rather than quietly losing whichever providers are polled last.

PR #4360, opened 2026-09-22
feat(be): hold accepted notifications in a bounded per-app queue

MRmarioruci:
A notification II accepts cannot go out on the spot: delivery is an HTTP outcall per device, and the subnet tolerates far fewer of those per second than an app can submit. Something has to sit between the two, absorb the difference, and tell a sender when it cannot take more. This is that queue, the unprocessed stage of the notification pipeline.

## Changes

`admission_queue.rs` is the mechanism, generic over what it holds and living at the crate root beside the other feature-agnostic utilities. `admit` answers `Stored`, `Folded` or `Full` per item, synchronously, folding a key it already holds or a group at its cap, and stamping the arrival time itself so a caller cannot age its own items in or out. `take_batch` hands entries on round-robin across tenants, one at a time, lanes ordering only within a tenant so nobody can jump another tenant's queue. Capacity is a ceiling per tenant below the ceiling overall, so a tenant arriving later finds room rather than a queue one burst already filled. The retry hint grows with how long the queue has been full and jumps to minutes once nothing is leaving at all, which is a failure further down rather than congestion. `snapshot` and `restore` carry it across an upgrade, rebuilding the indexes rather than trusting them to have survived.

`notifications/backlog.rs` is the notification side of that contract: a recipient, an app-chosen id and an RFC 8030 urgency, keyed by `(recipient, id)` because an id is unique only within one app and one user, grouped by recipient because each pending notification becomes an outcall per device, lanes being the four urgency levels, and the origin hash as the tenant.

Nothing calls any of it yet, hence the `dead_code` allows. The entrypoint and the dispatcher arrive in later PRs of this stack.

## Tests

33 unit tests: ordering, folding, round-robin and the cursor surviving across batches, both capacity ceilings, every branch of the retry hint, expiry at both ends, and snapshot round-trip. Ten of them call `assert_consistent`, which re-derives the counters and both indexes from the lanes, so any path that touches state has to leave them agreeing. The config's invariants are `const` assertions, so an incoherent one fails the build rather than a test.


src/internet_identity/src/admission_queue.rs
```diff
@@ -0,0 +1,1751 @@
...
+            // Before refusing for capacity, free expired entries across all senders.
+            // Once per call is enough because now_ns does not change.
+            if !swept && (self.is_full() || self.sender_at_cap(&sender)) {
+                self.discard_expired(now_ns);
+                swept = true;
+            }
+
+            // If the queue or this sender is full, request a retry. The caller may retry later, and the queue may have freed space in the meantime.
+            if self.is_full() || self.sender_at_cap(&sender) {
+                let retry_after_ms = self.retry_after_ms(self.at_capacity_since_ns, now_ns);
```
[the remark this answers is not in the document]
MRmarioruci: fixed in d6f7d710a. Each sender now keeps its own `at_cap_since_ns`, started at the first refusal and cleared on the next successful insert. It can't be started when the sender fills, because `sender_cap()` is `min(max_entries_per_sender, max_entries / active_senders)` and moves under the sender as others come and go.

src/internet_identity/src/admission_queue.rs
```diff
@@ -0,0 +1,1751 @@
...
+                backlog.insert(&sender.sender, entry);
+            }
+            if let Some(queue) = backlog.senders.get_mut(&sender.sender) {
+                let levels = queue.entries_by_priority.len();
+                queue.serving_priority = sender.serving_priority.min(levels.saturating_sub(1));
+                queue.turns_left = sender.turns_left;
+            }
+        }
+        backlog.next_sender = snapshot.next_sender;
+        backlog.at_capacity_since_ns = backlog.is_full().then_some(now_ns);
```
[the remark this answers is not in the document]
MRmarioruci: fixed in d6f7d710a. `insert` no longer sets `full_since_ns` at all, so nothing reconstructs a group clock from an entry's age. The clock starts at the first refusal, which after a restore is the first one post-upgrade.

src/internet_identity/src/admission_queue.rs
```diff
@@ -0,0 +1,1751 @@
...
+        let batch = self.remove_batch(limit, now_ns);
+        match store(&batch) {
+            Ok(stored) => Ok(stored),
+            Err(error) => {
+                for taken in batch {
+                    self.insert(&taken.sender, taken.entry);
+                }
+                self.nonempty_since_ns = nonempty_since_ns;
+                self.at_capacity_since_ns = at_capacity_since_ns;
+                self.last_taken_ns = last_taken_ns;
```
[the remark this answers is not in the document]
MRmarioruci: Half of this went away in d6f7d710a: `insert` no longer sets a group's `full_since_ns`, so a rollback can only lose a group clock, never restore a stale one, and the next refusal restarts it there.

The rotation part is deliberate. Restoring it would cost more than it buys: one sender loses at most one turn, and no entry is lost. The doc now says that instead of claiming exact recovery.

src/internet_identity/src/notifications/backlog.rs
```diff
@@ -0,0 +1,284 @@
...
+    pressure_cleared_below: 8_000,
+    // Stop trying to wake devices for entries older than five minutes.
+    discard_entries_after_ns: 5 * MINUTE_NS,
+    // An app resending keeps a notification alive, but never beyond this.
+    max_lifetime_ns: 15 * MINUTE_NS,
+    retry: RetryPolicy {
+        base_ms: 5_000,
+        ceiling_ms: 300_000,
+        when_stalled_ms: 600_000,
+        doubles_every_ns: 30 * SECOND_NS,
```
sea-snake, line 72: I see a mixture of ms and ns here.

src/internet_identity/src/notifications/backlog.rs
```diff
@@ -0,0 +1,284 @@
...
+        max_doublings: 6,
```
sea-snake, line 73: It makes more sense to declare a ceiling instead of having a max doubling. The doubling itself should probably be a backoff trait with a default impl similar to what we have in the single flight cache: https://github.com/dfinity/internet-identity/blob/main/src/internet_identity/src/single_flight_cache.rs#L75

src/internet_identity/src/notifications/backlog.rs
```diff
@@ -0,0 +1,263 @@
...
+const _: () = assert!(NOTIFICATION_BACKLOG.is_coherent());
+
+// A single app must stay below the pressure reset threshold.
+const _: () = assert!(
+    NOTIFICATION_BACKLOG.max_entries_per_sender < NOTIFICATION_BACKLOG.pressure_cleared_below
+);
+
+// Opening one slot must not reset the pressure timer.
+const _: () =
+    assert!(NOTIFICATION_BACKLOG.pressure_cleared_below < NOTIFICATION_BACKLOG.max_entries);
```
sea-snake, line 87: Never saw this syntax before, is this like a compile time check on our constants?

src/internet_identity/src/notifications/backlog.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn key(&self) -> Self::Key {
+        (self.recipient, self.notification_id)
+    }
+
+    fn group(&self) -> Self::Group {
+        self.recipient
+    }
+
+    fn priority(&self) -> usize {
+        self.urgency as usize
```
sea-snake, line 46: You're casting an enum to a number here, no idea what "high" is going to be as numeric value, also what happens if the enum order changes?

src/internet_identity/src/admission_queue.rs
```diff
@@ -0,0 +1,1790 @@
...
+    /// Items in the same group share a per-sender pending limit.
+    type Group: Copy + Eq + Hash;
+
+    /// Zero is treated as one priority level.
+    const PRIORITY_LEVELS: usize;
+
+    fn key(&self) -> Self::Key;
+    fn group(&self) -> Self::Group;
+
+    /// Lower values get more turns. Out-of-range values use the lowest priority.
```
sea-snake, line 30: Wouldn't it make sense to let the queueitem define how many turns it gets instead of having an opaque priority number with no clear meaning?

src/internet_identity/src/notifications/backlog.rs
MRmarioruci: Fixed in 54db231b9, it's all ns now: `base_ns` / `ceiling_ns` / `when_stalled_ns`, and `retry_after_ms` became `retry_after_ns: u64`. Nothing converts units anywhere anymore, since `Timestamp` is ns and the candid `Deferred` carries an absolute ns timestamp, so the ms was only ever an internal unit. Config reads better for it too: `base_ns: 5 * SECOND_NS`.

src/internet_identity/src/notifications/backlog.rs
MRmarioruci: You're right, `max_doublings` was dead: `5_000 << 6` is 320_000 against a 300_000 ceiling, so `.min(ceiling)` already clamped before it ever bound. Removed in 54db231b9. Overflow was never its job either, `checked_shl` and `saturating_mul` handle that.

Left the trait alone for now. `RetryBackoff` is indexed by consecutive failure count and this one by how long the pressure has lasted, so a shared trait would have two impls taking different inputs and sharing only the shift. Worth doing if a third caller turns up.

src/internet_identity/src/notifications/backlog.rs
MRmarioruci: Yes, exactly that. `const _: () = assert!(...)` is evaluated at compile time, so the build fails if the constants contradict each other: base above ceiling, a pressure threshold that can never clear, that kind of thing.

src/internet_identity/src/notifications/backlog.rs
MRmarioruci: Fixed in 54db231b9. The enum is declared in the same order as the candid `Urgency` now (least urgent first), and `priority()` is an explicit match instead of a cast, so reordering it is a compile error rather than a silent renumbering.

src/internet_identity/src/notifications/admission_queue.rs
```diff
@@ -0,0 +1,1790 @@
...
+    /// Items in the same group share a per-sender pending limit.
+    type Group: Copy + Eq + Hash;
+
+    /// Zero is treated as one priority level.
+    const PRIORITY_LEVELS: usize;
+
+    fn key(&self) -> Self::Key;
+    fn group(&self) -> Self::Group;
+
+    /// Lower values get more turns. Out-of-range values use the lowest priority.
```
MRmarioruci: Yes, changed in 02949e4. The item type now sets the turns itself: `const TURNS: &'static [usize]`, one entry per lane, most urgent first, and `priority()` only picks the lane. Notifications use `&[4, 3, 2, 1]` for High, Normal, Low and VeryLow.

Taking: apps go round-robin, one item each. Inside an app, a lane runs for its turns and then hands over, an empty lane gives up its turns, and inside a lane the nearest deadline goes first.

Admitting: a lane only takes what it can drain before its items expire. Its cap is the app's allowance × its turns ÷ the turns of the busy lanes. A lane counts as busy if it holds items, is in the batch being sent, or held items in the last 5 minutes. So VeryLow alone can use the whole allowance, and next to High it gets a fifth. The 5-minute memory stops VeryLow filling up the moment High briefly empties, and it's kept even when the app's queue empties completely.

An item with a shorter deadline of its own also has to fit behind what's due before it in its lane: with half the window left, it gets half the lane's places.

Past either limit the item is refused with a retry hint. The hint runs on that lane's own clock, so it keeps growing while the lane stays full even if other lanes keep admitting. Nothing already queued is evicted when the mix changes. A lane over its new share drains at the new rate, and whatever runs out of time comes from that app's own lower lanes.

The allowance itself is sized from the ticker in #4367: 200 every 10 s over the 5-minute window is 6,000, and one app may hold 70% of it.

src/internet_identity/src/notifications/admission_queue.rs
```diff
@@ -0,0 +1,1981 @@
...
+
+    fn next_expiry_ns(&self) -> Option<Timestamp> {
+        self.entries_by_priority
+            .iter()
+            .filter_map(|entries| entries.first_key_value().map(|((ts, _), _)| *ts))
+            .min()
+    }
+}
+
+pub(crate) struct AdmissionQueue<Sender: Clone + Ord, Item: QueueItem> {
```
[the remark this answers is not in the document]
MRmarioruci: Intentional. Carrying the backlog across an upgrade is parked as its own follow-up to keep this stack small, and the description no longer promises it. The queue lives in the heap, so an upgrade drops at most its five-minute window of accepted notifications.

src/internet_identity/src/notifications/admission_queue.rs
```diff
@@ -0,0 +1,1981 @@
...
+use std::collections::{BTreeMap, HashMap};
+use std::hash::Hash;
+use std::ops::Bound::{Excluded, Unbounded};
+
+pub(crate) trait QueueItem: Clone {
+    /// Deduplication key within one sender.
+    type Key: Clone + Eq + Hash + Ord;
+
+    /// Items in the same group share a per-sender pending limit.
+    type Group: Copy + Eq + Hash;
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in 53bbc29: the trait now says equal keys must name the same group. It holds for notifications, since the key is (principal, id) and a principal belongs to one identity.

src/internet_identity/src/notifications/admission_queue.rs
```diff
@@ -0,0 +1,1981 @@
...
+        let nonempty_since_ns = self.nonempty_since_ns;
+        let at_capacity_since_ns = self.at_capacity_since_ns;
+        let last_taken_ns = self.last_taken_ns;
+
+        let batch = self.remove_batch(limit, now_ns);
+        match store(&batch) {
+            Ok(stored) => Ok(stored),
+            Err(error) => {
+                for taken in batch {
+                    self.insert(&taken.sender, taken.entry);
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in 53bbc29. A failed batch now goes back through `place`, which leaves the sender and lane backoff clocks alone, with a test that the lane's hint keeps escalating across a failed store.

src/internet_identity/src/notifications/admission_queue.rs
```diff
@@ -0,0 +1,1981 @@
...
+                }
+                self.nonempty_since_ns = nonempty_since_ns;
+                self.at_capacity_since_ns = at_capacity_since_ns;
```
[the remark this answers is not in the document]
MRmarioruci: Fixed in 53bbc29. `take_batch` discards expired entries before it saves the clocks, so the sweep sits outside the rollback and the restored clocks match what's left. Test added for an all-expired take whose store fails.

PR #4362, opened 2026-09-22
feat(be): the method an app's backend calls to notify its users

sea-snake:
Adds `app_send_notification`, the method an app's backend canister calls to have II notify its users, plus the candid and Rust types it takes. The send path does not exist yet, so the method refuses every call with `InternalCanisterError "Not enabled"` — this PR is the interface only, for review before anything is built behind it.

```candid
app_send_notification : (SendNotificationArg) -> (
    variant { Ok : SendNotificationResponse; Err : SendNotificationError }
);
```

### What the shape says

A notification is content-free: the app holds the content and the recipient's device fetches it, authenticated as that recipient, so the app's content key is `(recipient, id)`.

Ids are the sender's and scoped to `(origin, recipient)`, so one id can go to many recipients as independent notifications — a broadcast doesn't force the app to mint a key per user. Keeping them recipient-scoped is also what stops an app from accidentally serving one user's payload to another.

`expires_at` and `urgency` map onto Web Push's `TTL` and `Urgency` headers and are forwarded as sent. Urgency orders an origin's entries only against its own, so marking everything `High` gains nothing.

The reply lists only what was **not** accepted: an unknown recipient, a recipient with no channel enabled, or an entry there was no room for, which carries the time to send it again. `Err` is reserved for the whole call being refused — the caller isn't a listed sender for the origin, the batch was larger than II will process, or II broke.

Acceptance is not a delivery guarantee: an accepted entry can still expire, be displaced by a more urgent entry from the same origin, or fail at the recipient's channel.

### Open, deliberately not settled here

- How the origin↔sender relation is verified. `SenderNotListed` assumes the origin publishes its sender canister ids at a well-known path; nothing reads that yet.
- Whether the push carries `(origin, id)` as an encrypted payload or is a bare tickle the service worker follows up on.
- The naming sits slightly apart from the `notification_*` family already on main (`NotificationGrantConsentRequest` and friends): the `app_` prefix marks the method as one apps call rather than II's own frontend. Happy to fold it into the existing naming if that reads better.


src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -1052,3 +1052,86 @@ pub struct WebPushSubscriptionStatus {
...
+    /// retention, which is also the ceiling for any value given here.
+    pub expires_at: Option<Timestamp>,
+    /// `None` means `Normal`.
+    pub urgency: Option<Urgency>,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub struct SendNotificationArg {
+    pub origin: FrontendHostname,
+    pub notifications: Vec<Notification>,
```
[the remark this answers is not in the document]
sea-snake: Fixed in a9221ad9 by writing the rule down rather than changing the shape: the batch applies in order, so a later entry for a `(recipient, id)` supersedes an earlier one in the same call, and the reply names each distinct pair at most once. That is your middle option, and it also means an input index in the response would have nothing extra to say — every entry the sender submitted maps onto exactly one key.

src/internet_identity_interface/src/internet_identity/types.rs
```diff
@@ -1052,3 +1052,86 @@ pub struct WebPushSubscriptionStatus {
...
+/// Everything the batch carried that `not_accepted` does not name was accepted.
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub struct SendNotificationResponse {
+    pub not_accepted: Vec<NotAccepted>,
+}
+
+#[derive(Clone, Debug, CandidType, Deserialize, Eq, PartialEq)]
+pub enum SendNotificationError {
+    /// The origin's sender list was read and does not list the caller.
+    SenderNotListed,
```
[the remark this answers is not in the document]
sea-snake: Documented in a9221ad9 rather than given a variant. The origin a sender passes is a fixed value in its own deployment, not something it varies per call, so a malformed or non-canonically spelled one is a deployment bug surfaced once during integration — it comes back as `InternalCanisterError` carrying the reason, which is also how the neighbouring `notification_*` endpoints already answer it (`notifications/validation.rs` maps every `notifying_origin` failure onto that variant). Adding an invalid-origin variant here would make this method the odd one out in its own family.

src/internet_identity/src/main.rs
```diff
@@ -417,6 +417,21 @@ fn notification_consent_granted(request: NotificationConsentGrantedRequest) -> b
...
+/// Takes a batch of content-free signals for the origin's own users, authorized by that
+/// origin listing the caller as one of its senders.
+///
+/// The send path does not exist yet, so every call is refused.
+#[update]
+fn app_send_notification(
+    _request: SendNotificationArg,
+) -> Result<SendNotificationResponse, SendNotificationError> {
+    Err(SendNotificationError::InternalCanisterError(
+        "Not enabled".to_string(),
```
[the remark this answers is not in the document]
sea-snake: Adding a temporary test to see if code remains "not implemented" is overengineering...

src/internet_identity/internet_identity.did
```diff
@@ -1806,6 +1806,69 @@ type WebPushSubscriptionStatus = record {
     issued_at_ns : nat64;   // window i expires at issued_at_ns + (i + 1) * window
 };
 
+// ===== Notifications sent by an app =====
+
+// Scoped to (origin, recipient): the same id sent to two recipients is two
+// notifications. The canisters sending for one origin share its id space.
+type NotificationId = nat64;
+
+// Orders an origin's entries against its own only, so marking everything High
```
MRmarioruci: we don't need to document this here.. this is information for integrators.

src/internet_identity/internet_identity.did
```diff
@@ -1806,6 +1806,69 @@ type WebPushSubscriptionStatus = record {
...
+// gains nothing. Web Push's levels, forwarded as sent.
+type Urgency = variant { VeryLow; Low; Normal; High };
+
+// A content-free signal: the app holds the content, the recipient's device
```
MRmarioruci: ditto. this is information about the overall design, no need to live in candid

src/internet_identity/internet_identity.did
```diff
@@ -1806,6 +1806,69 @@ type WebPushSubscriptionStatus = record {
...
+// fetches it. A re-send means that content changed, and replaces the entry
+// while it is still pending; once delivered the id is forgotten.
+//
+// Acceptance is not a delivery guarantee — an entry may expire, be displaced by
+// a more urgent one from the same origin, or fail at the channel.
+type Notification = record {
+    id : NotificationId;
+    recipient : principal;
+    // Dropped undelivered once passed. Null means II's default retention,
+    // which is also the ceiling.
```
MRmarioruci:
```suggestion
    // Drop from the send queue once reached. If left Null, II will enforce it's own defaults.
```

src/internet_identity/internet_identity.did
```diff
@@ -1806,6 +1806,69 @@ type WebPushSubscriptionStatus = record {
...
+type SendNotificationArg = record {
+    origin : FrontendHostname;
+    notifications : vec Notification;
+};
+
+type NotAcceptedReason = variant {
+    // No such recipient at this origin.
+    UnknownRecipient;
+    // The recipient has no notification channel enabled.
+    NoChannel;
```
MRmarioruci, line 1835: what does channel mean here? If we add emails for example in the future I think that's another channel?? or are you just referring to the subscription endpoint? If it's the subscription endpoint then NoSubscriptionEndpoint would be better.. or maybe NoWebPushSubscriptionEndpoint, so that we extend with other reasons if we add more methods in the future?

src/internet_identity/internet_identity.did
```diff
@@ -1806,6 +1806,69 @@ type WebPushSubscriptionStatus = record {
...
+
+type NotAccepted = record {
+    id : NotificationId;
+    recipient : principal;
+    reason : NotAcceptedReason;
+};
+
+// Anything not_accepted does not name was accepted.
+type SendNotificationResponse = record {
+    not_accepted : vec NotAccepted;
```
MRmarioruci, line 1847: Why only receive not_accepted?? I would probably return the original vector + a status.. See my PR. cause there might be things like retry_ms, folded or not etc.. and this is info the client needs to act uppon..

src/internet_identity/internet_identity.did
```diff
@@ -1806,6 +1806,69 @@ type WebPushSubscriptionStatus = record {
...
+};
+
+type SendNotificationError = variant {
+    // The origin's sender list was read and does not list the caller.
+    SenderNotListed;
+    // More entries than II will process in one call; nothing was enqueued.
+    // limit is fixed and never below an origin's queue capacity.
+    TooManyNotifications : record { limit : nat32 };
```
MRmarioruci, line 1853: Kind of the same.. we need to think about this interface a bit more, in combination with the backlog queue

src/internet_identity/internet_identity.did
sea-snake: Agreed, trimmed. What's left is only what can't be read off the types: that ids are scoped to (origin, recipient) and shared across an origin's canisters, since that decides how an app assigns them, and that a re-send replaces a pending entry, since that's the update mechanism. The rationale — urgency being origin-local, acceptance not being delivery, what the batch limit is for — is gone.

That does leave the app-facing API without a written explanation anywhere, so integrator documentation should land alongside the send path rather than in the did.

src/internet_identity/internet_identity.did
sea-snake: I've  trimmed all comments.

src/internet_identity/internet_identity.did
sea-snake: It's an abstraction over "no webpush subscription", this enables us to use other "channels" besides webpush in the future.

src/internet_identity/internet_identity.did
sea-snake: As discussed, returning accepted is not necessary, that's implied from accepted ones not being in not_accepted.

MRmarioruci (approved):

